PR #853, opened 2013-07-03
Have test_openssl.rb be explicit about security checks it adds

ratnikov:
Hey all,

With this commit I'm trying to get JRuby more explicit about the security checks it requires to run. There are three parts to this commit:

1) General framework of being able to set allowed security checks for tests.
2) General security checks that are assumed to be required to get JRuby running (in test/security_helper.rb)
3) Specific security checks required for test_openssl.rb to work.

Please let me know what you think. Assuming this gets merged in, I'd like to tackle test_load.rb next to provide some motivation to get rid of CompoundJar loader. ;)

### DIFF
--- a/test/org/jruby/security/SecurityManager.java
+++ b/test/org/jruby/security/SecurityManager.java
@@ -1,0 +1,82 @@
+package org.jruby.security;
+
+import org.jruby.Ruby;
+import org.jruby.RubyString;
+import org.jruby.runtime.builtin.IRubyObject;
+import java.security.Permission;
+import java.util.ArrayList;
+import java.util.List;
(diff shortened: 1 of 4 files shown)


headius:
@kares @ratnikov Is this still valid? We've since moved these files and updated to newer versions from MRI.

ratnikov:
As long the openssl still requires security manager checks, I think this commit is still valid. I believe my intention was to make sure that JRuby tests are conscious about what security manager checks are introduced.

kares:
Finally - moved, with some tunings, into jruby-openssl, thanks Dmitry

PR #887, opened 2013-07-15
Update site with developer, license, CI info. Update some deps and plugins.

jendave:
Update site with developer, license, CI info. Update some deps and plugins.


headius:
This PR set the pom-based licenses to GPL and LGPL 3.0. Those are incorrect; we are on GPL 2.0 and LGPL 2.1. Fixing on master and 1.7 branch.

headius:
See #4454 for issue filed about the incorrect licenses in poms.

PR #987, opened 2013-09-04
Add :chdir support for `exec`

BanzaiMan:
Rework the option processing, and add support for `:chdir`. Note that we are only dealing with fire-and-forget `exec`.


nirvdrum:
Is this still valid?

BanzaiMan:
It is probably still "valid" in the sense that this method needs to exist, but the implementation is faulty for the reasons that Tom raised.

Perhaps we can print a message about that.

In any case, this PR can be closed.

Issue #1216, opened 2013-11-11
Net::HTTP returns empty response body for HTTPS requests in 2.0 mode


rykov:
Gemfury customers are reporting problems using the service with jRuby in Ruby 2.0 mode.  We've tracked it down to an odd issue where Net::HTTP returns a successful response, but `response.body` returns an empty string for `https://` requests.

Here's a test program to reproduce:

```
require "net/https"

uri = URI.parse("https://rubygems.org/")
http = Net::HTTP.new(uri.host, uri.port)
http.use_ssl = true

request = Net::HTTP::Get.new(uri.request_uri)
response = http.request(request)
puts "#{response.code} - #{response.body.size}"
```

And the output (tested in 1.7.5 and 1.7.6):

```
$ jruby -S ./test.rb
200 - 9188
$ jruby --2.0 -S ./test.rb 
200 - 0
```

simi:
Hello @enebo. This is problem for 1.7.18 (2.0 version) again. Any chance to get it fixed?

I can fix this temporary with this diff:

```diff
diff --git a/lib/ruby/stdlib/net/http/response.rb b/lib/ruby/stdlib/net/http/response.rb
index b073169..59e856d 100644
--- a/lib/ruby/stdlib/net/http/response.rb
+++ b/lib/ruby/stdlib/net/http/response.rb
@@ -374,9 +374,7 @@ class Net::HTTPResponse
         dest.force_encoding(Encoding::ASCII_8BIT)
       end
       block = proc do |compressed_chunk|
-        @inflate.inflate(compressed_chunk) do |chunk|
-          dest << chunk
-        end
+        dest <<@inflate.inflate(compressed_chunk)
       end
 
       Net::ReadAdapter.new(block)
```

```
$ ruby -v
jruby 1.7.18 (2.0.0p598) 2014-12-22 625381c on OpenJDK 64-Bit Server VM 1.7.0_45-mockbuild_2013_10_16_17_51-b00 +jit [linux-amd64]
```

I've found https://github.com/jruby/jruby/commit/2cbe7e52a7e6a8e43a2066007547ad520b8623b7 also. Is there any chance to get that released in 1.7.x series?

I tried `jruby-openssl` `0.9.5` and `0.9.6`.

Script to reproduce:

```
$ jruby -ropen-uri -ropenssl -e 'p open("https://www.seznam.cz").size'
0
```

Workaround without mokeypatch:

```
[...]

PR #1611, opened 2014-04-05
Import specs from rubysl-zlib

LTe:
New zlib from ruby allows streams to be processed without huge
memory growth. There is specs for new API of #inflate and #deflate
methods.  Each of these methods may take a block which can process
one chunk of the stream.

Related to https://github.com/jruby/jruby/issues/1216

### DIFF
--- a/spec/ruby/library/zlib/adler32_spec.rb
+++ b/spec/ruby/library/zlib/adler32_spec.rb
@@ -1,4 +1,3 @@
-require File.expand_path('../../../spec_helper', __FILE__)
 require 'zlib'
 
 describe "Zlib.adler32" do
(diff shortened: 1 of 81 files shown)


chrisseaton:
@eregon am I right in thinking the library specs from rubysl-zlib should now be merged from your work on RubySpec? If so this can be closed.

eregon:
@chrisseaton Yes, except new specs added there.
I imported the [new specs](https://github.com/LTe/rubysl-zlib/commit/bb8eb9bf8fe67fe56fe12887b306ad523dd59f16) from rubysl-zlib in ruby/rubyspec.
They will be in jruby on the next spec merge.

PR #1860, opened 2014-07-25
Bash launcher refactor

mjc:
This is the beginnings of a refactoring for the bash launcher.  I intend to modernize it and fix the following bug.

Currently the JRUBY_OPTS and option parsing code requires options that affect some kinds of code to be earlier in the option list than other kinds.  For example:

```
JRUBY_OPTS="--profile.html --profile.out test.html --sample" bin/jruby test.rb
```
this returns: jruby: unknown option --sample

```
bin/jruby --profile.html --profile.out test.html --sample test.rb
```
as does this.

This means that some options cannot currently be specified in JRUBY_OPTS if it is at all possible that anything launched with JRUBY_OPTS set may use some other options.

This rewrite is still in progress, I'm sending the pull request in case I end up not having time to finish it.

### DIFF
--- a/bin/jruby.bash
+++ b/bin/jruby.bash
@@ -342,59 +342,59 @@ if [ "$nailgun_client" != "" ]; then
     exit 1
   fi
 else
-if [ "$VERIFY_JRUBY" != "" ]; then
-  if [ "$PROFILE_ARGS" != "" ]; then
-      echo "Running with instrumented profiler"
-  fi
+  if [ "$VERIFY_JRUBY" != "" ]; then
+    if [ "$PROFILE_ARGS" != "" ]; then
+        echo "Running with instrumented profiler"
+    fi
 
-  if [[ "${java_class:-}" == "${JAVA_CLASS_NGSERVER:-}" && -n "${JRUBY_OPTS:-}" ]]; then
-    echo "warning: starting a nailgun server; discarding JRUBY_OPTS: ${JRUBY_OPTS}"
-    JRUBY_OPTS=''
-  fi
+    if [[ "${java_class:-}" == "${JAVA_CLASS_NGSERVER:-}" && -n "${JRUBY_OPTS:-}" ]]; then
+      echo "warning: starting a nailgun server; discarding JRUBY_OPTS: ${JRUBY_OPTS}"
+      JRUBY_OPTS=''
+    fi
 
-  "$JAVACMD" $PROFILE_ARGS $JAVA_OPTS "$JFFI_OPTS" "${java_args[@]}" -classpath "$JRUBY_CP$CP_DELIMITER$CP$CP_DELIMITER$CLASSPATH" \
-    "-Djruby.home=$JRUBY_HOME" \
-    "-Djruby.lib=$JRUBY_HOME/lib" -Djruby.script=jruby \
-    "-Djruby.shell=$JRUBY_SHELL" \
-    $java_class $JRUBY_OPTS "$@"
-
-  # Record the exit status immediately, or it will be overridden.
-  JRUBY_STATUS=$?
-
-  if [ "$PROFILE_ARGS" != "" ]; then
-      echo "Profiling results:"
-      cat profile.txt
-      rm profile.txt
-  fi
-
-  if $cygwin; then
-    stty icanon echo > /dev/null 2>&1
-  fi
-
-  exit $JRUBY_STATUS
-else
-  if $cygwin; then
-    # exec doed not work correctly with cygwin bash
-    "$JAVACMD" $JAVA_OPTS "$JFFI_OPTS" "${java_args[@]}" -Xbootclasspath/a:"$JRUBY_CP" -classpath "$CP$CP_DELIMITER$CLASSPATH" \
+    "$JAVACMD" $PROFILE_ARGS $JAVA_OPTS "$JFFI_OPTS" "${java_args[@]}" -classpath "$JRUBY_CP$CP_DELIMITER$CP$CP_DELIMITER$CLASSPATH" \
       "-Djruby.home=$JRUBY_HOME" \
       "-Djruby.lib=$JRUBY_HOME/lib" -Djruby.script=jruby \
       "-Djruby.shell=$JRUBY_SHELL" \
-      $java_class $mode "$@"
+      $java_class $JRUBY_OPTS "$@"
 
     # Record the exit status immediately, or it will be overridden.
     JRUBY_STATUS=$?
 
-    stty icanon echo > /dev/null 2>&1
+    if [ "$PROFILE_ARGS" != "" ]; then
+        echo "Profiling results:"
+        cat profile.txt
+        rm profile.txt
+    fi
+
+    if $cygwin; then
+      stty icanon echo > /dev/null 2>&1
+    fi
 
     exit $JRUBY_STATUS
   else
-    exec "$JAVACMD" $JAVA_OPTS "$JFFI_OPTS" "${java_args[@]}" -Xbootclasspath/a:"$JRUBY_CP" -classpath "$CP$CP_DELIMITER$CLASSPATH" \
-      "-Djruby.home=$JRUBY_HOME" \
-      "-Djruby.lib=$JRUBY_HOME/lib" -Djruby.script=jruby \
-      "-Djruby.shell=$JRUBY_SHELL" \
-      $java_class $mode "$@"
+    if $cygwin; then
+      # exec doed not work correctly with cygwin bash
+      "$JAVACMD" $JAVA_OPTS "$JFFI_OPTS" "${java_args[@]}" -Xbootclasspath/a:"$JRUBY_CP" -classpath "$CP$CP_DELIMITER$CLASSPATH" \
+        "-Djruby.home=$JRUBY_HOME" \
+        "-Djruby.lib=$JRUBY_HOME/lib" -Djruby.script=jruby \
+        "-Djruby.shell=$JRUBY_SHELL" \
+        $java_class $mode "$@"
+
+      # Record the exit status immediately, or it will be overridden.
+      JRUBY_STATUS=$?
+
+      stty icanon echo > /dev/null 2>&1
+
+      exit $JRUBY_STATUS
+    else
+      exec "$JAVACMD" $JAVA_OPTS "$JFFI_OPTS" "${java_args[@]}" -Xbootclasspath/a:"$JRUBY_CP" -classpath "$CP$CP_DELIMITER$CLASSPATH" \
+        "-Djruby.home=$JRUBY_HOME" \
+        "-Djruby.lib=$JRUBY_HOME/lib" -Djruby.script=jruby \
+        "-Djruby.shell=$JRUBY_SHELL" \
+        $java_class $mode "$@"
+    fi
   fi
-fi
 fi
 
 # Be careful adding code down here, you might override the exit
--- a/bin/jruby.bash
+++ b/bin/jruby.bash
@@ -192,6 +192,100 @@ set_system_classpath() {
   fi
 }
 
+set_java_args_from_j_option() {
+  val=${1:2}
+  if [ "${val:0:4}" = "-Xmx" ]; then
+      JAVA_MEM=$val
+  elif [ "${val:0:4}" = "-Xms" ]; then
+      JAVA_MEM_MIN=$val
+  elif [ "${val:0:4}" = "-Xss" ]; then
+      JAVA_STACK=$val
+  elif [ "${val}" = "" ]; then
+      $JAVACMD -help
+      echo "(Prepend -J in front of these options when using 'jruby' command)"
+      exit
+  elif [ "${val}" = "-X" ]; then
+      $JAVACMD -X
+      echo "(Prepend -J in front of these options when using 'jruby' command)"
+      exit
+  elif [ "${val}" = "-classpath" ]; then
+      CP="$CP$CP_DELIMITER$2"
+      CLASSPATH=""
+      shift
+  elif [ "${val}" = "-cp" ]; then
+      CP="$CP$CP_DELIMITER$2"
+      CLASSPATH=""
+      shift
+  else
+      if [ "${val:0:3}" = "-ea" ]; then
+          VERIFY_JRUBY="yes"
+      elif [ "${val:0:16}" = "-Dfile.encoding=" ]; then
+          JAVA_ENCODING=$val
+      fi
+      java_args=("${java_args[@]}" "${1:2}")
+  fi
+}
+
+set_defines_from_x_options() {
+  val=${1:2}
+  if expr "$val" : '.*[.]' > /dev/null; then
+    java_args=("${java_args[@]}" "-Djruby.${val}")
+  else
+    ruby_args=("${ruby_args[@]}" "-X${val}")
+  fi
+}
+
+# Run with JMX management enabled
+enable_jmx() {
+  java_args=("${java_args[@]}" "-Dcom.sun.management.jmxremote")
+  java_args=("${java_args[@]}" "-Djruby.management.enabled=true")
+}
+
+# Don't launch a GUI window, no matter what
+enable_headless() {
+  java_args=("${java_args[@]}" "-Djava.awt.headless=true")
+}
+
+# Run under JDB
+enable_jdb() {
+  if [ -z "$JAVA_HOME" ] ; then
+    JAVACMD='jdb'
+  else
+    if $cygwin; then
+      JAVACMD="$(cygpath -u "$JAVA_HOME")/bin/jdb"
+    else
+      JAVACMD="$JAVA_HOME/bin/jdb"
+    fi
+  fi
+  java_args=("${java_args[@]}" "-sourcepath" "$JRUBY_HOME/lib/ruby/1.9:.")
+  JRUBY_OPTS=("${JRUBY_OPTS[@]}" "-X+C")
+}
+
+enable_fast_startup() {
+  JAVA_VM=-client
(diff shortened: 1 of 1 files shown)


jirutka:
Both `jruby.bash` and `jruby.sh` are really **very** bad (no offense, but… look at them). Half of these scripts doesn’t make any sense, obfuscates the purpose and/or is not even used.

I’m just writing JRuby package for Alpine Linux. After I had read the scripts I decided that instead of patching `jruby.sh`, I’ll rather rewrite it.

The new script is more clean, robust, POSIX compatible (i.e. can be run with Busybox’s ash or even dash), without Windows/cygwin garbage (there’s already a native launcher for Windows) and Truffle stuff (since it’s still experimental, it’s IMO not needed for regular users yet).
Since it doesn’t use Bash arrays and I didn’t want to make arguments handling even more complicated than it already is, it doesn’t handle arguments with spaces correctly (same as `jruby.sh`).
Beside that it should support all the options of `jruby.bash` (hopefully) with the same behaviour.

I wrote it specifically for Alpine Linux, so it has different defaults and omits some features (like Truffle). I had thought that I’ll modify it for general purpose and send PR, but then I found this two years old PR, yet still unresolved, so I’m not sure if there’s actually any interest in it.

There’s the new script, maybe it’ll be useful:

```sh
#!/bin/sh
# vim: set ts=4:
#
# This jruby start script is rewrite of (poor) upstream's jruby.bash.
# It's POSIX compatible, more clean, with sensible defaults for Alpine Linux
# and without Windows, OS X and Truffle stuff. Beside that it should support
# all the options of jruby.bash (hopefully) with the same behaviour.
#
# -----BEGIN HELP-----
#
# Environment variables:
#     CLASSPATH     Additional paths to be added to the end of the Java classpath.
#                   This may be overwritten by CLI argument -J-classpath or -J-cp.
#     JAVA_HOME     Path to directory where is JRE/JVM installed to use specific java/jdb binary,
#                   not the one on the PATH.
#     JAVA_OPTS     Default options to be passed to JVM.
#     JFFI_PATH     Path to directory with libjffi.so (default is /usr/lib/jffi).
#     JRUBY_HOME    Path to directory where JRuby is installed (default is /usr/share/java/jruby).
#     JRUBY_OPTS    Default JRuby command line arguments.
#     JRUBY_SHELL   Path to the system shell executable (default is /bin/sh).
#     PROFILE_ARGS  Arguments for the instrumented profiler.
#
# Deprecated environment variables (for compatibility):
#     JAVA_MEM      The -Xmx (maximum memory allocation) parameter for JVM (e.g. -Xmx128M).
#                   This may be overwritten by CLI argument -J-Xmx.
#     JAVA_STACK    The -Xss (stack size) parameter for JVM (default is -Xss2048k).
#                   This may be overwritten by CLI argument -J-Xss.
#     JAVA_VM       What JIT mode to use; -client, or -server (default is -server).
#                   This may be overwritten by CLI arguments --server, --client, and --dev.
#     JRUBY_PARENT_CLASSPATH  Classpath propagated from the parent JRuby process.
#
# All the environment variables are optional.
# -----END HELP-----
#
set -eu

readonly MAIN_CLASS_JRUBY='org.jruby.Main'
readonly MAIN_CLASS_NGSERVER='org.jruby.main.NailServerMain'
readonly MAIN_CLASS_NGCLIENT='org.jruby.util.NailMain'

readonly JRUBY_HOME="${JRUBY_HOME:-"/usr/share/java/jruby"}"
readonly JRUBY_CP="$JRUBY_HOME/lib/jruby.jar"
readonly JRUBY_SHELL="${JRUBY_SHELL:-"/bin/sh"}"
readonly JFFI_PATH="${JFFI_PATH:="/usr/lib/jffi"}"
readonly NAILGUN_CMD="$JRUBY_HOME/tool/nailgun/ng"
readonly PROFILE_ARGS="${PROFILE_ARGS:-}"


java_stack="${JAVA_STACK:-"-Xss2048k"}"
java_cmd="${JAVA_HOME:+"$JAVA_HOME/bin/"}java"

# Use the same classpath propagated from parent jruby, if provided.
classpath="${JRUBY_PARENT_CLASSPATH:-}"
if [ -z "$classpath" ]; then
	_excluded='jruby.jar jruby-truffle.jar jruby-complete.jar'

	# Add all JARs from $JRUBY_HOME/lib to the classpath, except $_excluded.
	classpath=$(find "$JRUBY_HOME/lib" -maxdepth 1 \
		-name '*.jar' $(printf '! -name %s ' $_excluded) -print0 \
		| xargs -0 printf '%s:')
	classpath="${classpath%?}"  # %? removes leading ":"
fi

# Add additional paths from $CLASSPATH variable, if provided.
classpath="$classpath${CLASSPATH:+":$CLASSPATH"}"

extra_cp="${CLASSPATH:-}"
java_opts="${JAVA_OPTS:-} ${JAVA_MEM:-} $java_stack -Djffi.boot.library.path=$JFFI_PATH"
java_vm="${JAVA_VM:-"-server"}"
main_class="$MAIN_CLASS_JRUBY"
nailgun_client='no'
ruby_args=''
verify_jruby='no'


# Split out any -J argument for passing to the JVM.
# Scanning for args is aborted by '--'.
set -- ${JRUBY_OPTS:-} $@
while [ $# -gt 0 ]; do
	case "$1" in
	# Stuff after "-J" goes to JVM.
	-J | -J-X)
		"$java_cmd" -help
		printf '\n(Prepend -J in front of these options when using "jruby" command)\n'
		exit 1
	;;
	-J-classpath | -J-cp)
		extra_cp="$2"
		shift
	;;
	-J-ea)
		verify_jruby='yes'
		java_opts="$java_opts ${1:2}"
	;;
	-J*)
		java_opts="$java_opts ${1:2}"
	;;
	 # Pass -X... and -X? search options through.
	 -X*\.\.\.|-X*\?)
		ruby_args="$ruby_args $1"
	;;
	 # Match -Xa.b.c=d to translate to -Da.b.c=d as a Java option.
	 -X*)
		val=${1:2}
		if expr "$val" : '.*[.]' > /dev/null; then
			java_opts="$java_opts -Djruby.$val"
		else
			ruby_args="$ruby_args -X$val"
		fi
	;;
	 # Match switches that take an argument.
	 -C | -e | -I | -S)
		ruby_args="$ruby_args $1 $2"
		shift
	;;
	 # Match same switches with argument stuck together.
	 -e* | -I* | -S*)
		ruby_args="$ruby_args $1"
	;;
	 # Run with JMX management enabled.
	 --manage)
		java_opts="$java_opts -Dcom.sun.management.jmxremote \
			-Djruby.management.enabled=true"
	;;
	 # Don't launch a GUI window, no matter what.
	 --headless)
		java_opts="$java_opts -Djava.awt.headless=true"
	;;
	 # Run under JDB (debug mode).
	 --jdb)
		java_cmd="${JAVA_HOME:+"$JAVA_HOME/bin/"}jdb"
		java_opts="$java_opts -sourcepath $JRUBY_HOME/lib/ruby/1.9:."
		ruby_args="$ruby_args -X+C"
	;;
	 --client | --server)
		java_vm="${1#?}"  # #? removes leading "-"
	;;
	 --dev)
		java_vm='-client'
		java_opts="$java_opts -XX:+TieredCompilation -XX:TieredStopAtLevel=1 \
			-Djruby.compile.mode=OFF -Djruby.compile.invokedynamic=false"
	;;
	 --sample)
		java_opts="$java_opts -Xprof"
	;;
	# Start up as Nailgun server.
	 --ng-server)
		main_class="$MAIN_CLASS_NGSERVER"
		verify_jruby='yes'
	;;
	# Use native Nailgun client to toss commands to server.
	 --ng-client | --ng)
		nailgun_client='yes'
	;;
	 # Warn but ignore.
	 --1.8 | --1.9 | --2.0)
		echo "WARNING: $1 is ignored"
	;;
	-h | -help | --help)
		"$java_cmd" -help
		# Print help from the top of this script.
		sed -En '/-{5}BEGIN HELP-{5}/,/-{5}END HELP-{5}/p' "$0" \
			| sed -E "s/^# ?//; 1d;\$d;"
		exit 0
	;;
	 # Abort processing on the double dash.
	 --)
		break
	;;
	 # Other opts go to ruby.
	 -*)
		ruby_args="$ruby_args $1"
	;;
	 # Abort processing on first non-opt arg.
	 *)
		break
	;;
	esac
	shift
done

# Remove possible duplications of some common JVM options.
for opt in -Xmx -Xms -Xss; do
	val=$(expr "$java_opts" : ".*$opt\([^ \t]*\).*" ||:)  # gets the later one
	if [ -n "$val" ]; then
		# Remove all occurrences of $opt and append the last one to the end.
		java_opts=$(echo "$java_opts" | sed "s/$opt[^ \t]*//g")
		java_opts="$java_opts ${opt}${val}"
	fi
done

classpath="$classpath${extra_cp:+":$extra_cp"}"
java_opts="$java_opts $java_vm
	-Djruby.home=$JRUBY_HOME
	-Djruby.lib=$JRUBY_HOME/lib
	-Djruby.script=jruby
	-Djruby.shell=$JRUBY_SHELL"

# Append the rest of the arguments.
ruby_args="$ruby_args $@"

# Put $ruby_args back into the position arguments $1, $2, ...
set -- $ruby_args


if [ "$nailgun_client" = 'yes' ]; then
	if [ ! -f "$NAILGUN_CMD" ]; then
		echo 'ERROR: ng executable not found' 1>&2
		exit 1
	fi
	exec "$NAILGUN_CMD" $MAIN_CLASS_NGCLIENT $@

elif [ "$verify_jruby" = 'yes' ]; then
	[ -n "$PROFILE_ARGS" ] && echo 'Running with instrumented profiler'

	set +e
	"$java_cmd" $PROFILE_ARGS $java_opts \
		-classpath "$JRUBY_CP:$classpath" $main_class $@
	exit_code=$?
	set -e

	if [ -n "$PROFILE_ARGS" ]; then
		echo 'Profiling results:'
		cat profile.txt
		rm profile.txt
	fi

	exit $exit_code

else
	exec "$java_cmd" $java_opts -Xbootclasspath/a:"$JRUBY_CP" \
		${classpath:+"-classpath $classpath"} \
		$main_class $@
fi
```

enebo:
@jirutka I believe you are missing some quoting for things like JRUBY_HOME containing a space?  This is a real issue since people run this script using bash/mingw on windows even if nearly unheard of on unixy systems.  In general quoting with paths are for this case.

Also does your script still work for the case mentioned in https://www.mail-archive.com/dev@jruby.codehaus.org/msg03551.html (what was once JRUBY-2699)?

I am happy you are making this cleaner and run on more environments but I think stuff removed will need some individual justifications and/or someone else with more experience with bash to verify things are cool.

@wayneeseguin @mpapis I don't know if you guys have any time but it would really help.  My bash knowledge is from decades of limping my .profile/.bashrcs along so I do not feel up to the task :)

mpapis:
I might try to review today evening, in worst case next week, but in anyway it would be best seen as a PR where I can comment of fork from to provide improvements

jirutka:
> I believe you are missing some quoting for things like JRUBY_HOME containing a space? This is a real issue since people run this script using bash/mingw on windows…

As I’ve explicitly stated, I dropped Windows support. Windows users usually doesn’t have mingw (do they?) and there’s already a native launcher, so I don’t think it’s worth to increase complexity of the script for this. Trying to support everything in single script is a road to hell.

> …even if nearly unheard of on unixy systems.

Well, then such users must use some other script or native launcher…

> Also does your script still work for the case mentioned in https://www.mail-archive.com/dev@jruby.codehaus.org/msg03551.html (what was once JRUBY-2699)?

Kind of… It requires `JRUBY_HOME` to be at predefined path (`/usr/share/java/jruby` on Alpine) or be provided by environment; `jruby` script is installed into `/usr/bin/jruby`.

> I am happy you are making this cleaner and run on more environments…

Well, I dropped support for Windows and OS X (although it’s not problem to add these few OS X specific stuff, and I use OS X too), so it’s actually opposite.

> I might try to review today evening, in worst case next week, but in anyway it would be best seen as a PR where I can comment of fork from to provide improvements

**I should remind that the script above is written for Alpine Linux, it was not intended to be adopted by upstream as-is.** However, if you’re interested and you’ll not insist on supporting Windows, then I can modify it (e.g. bring Truffle support etc.) and send PR.

enebo:
@jirutka heh.  You know I missed a couple of sentences in your original comment.  I misread that this was just what you intended to use on alpine linux.  So in that case, we may or may not want to consider merging your code and cleaning up the things we will still need in the original script.

So thanks for sharing and if we can get someone interested in merging/cleaning up the original script then that will be great.  At this point, I think the project should consider if we should consider dropping some level of support in this script like yours does.

@mpapis I rescind my request unless you want to help @jirutka find any potential issues with his launcher script.

jirutka:
@enebo …if we can get someone interested in merging/cleaning up the original script…

I’ll be happy to do it, if there’s interest from JRuby devs, but as I stated, I don’t want to struggle with support for Windows/mingw.

mpapis:
spaces in paths - OSX with a new disk - complete nightmare

jirutka:
> spaces in paths - OSX with a new disk - complete nightmare

I don’t understand…? What a new disk has common with this?

mpapis:
@jirutka spaces in paths happen when you add a new disk to OSX, this PR is to improve the scripts used in jruby - this might include OSX, so we need to account for spaces in paths.

jirutka:
> spaces in paths happen when you add a new disk to OSX

New disk is mounted under `/Volumes`. There are no spaces, unless you have spaces in your disk label… However, how many users install JRuby on second/external drive?

mpapis:
usually the home gets moved to the second drive, it contains spaces often enough to be a problem, and I guess users might want to install jruby in home directory?

jirutka:
Okay… you’re concerned about spaces just in `JRUBY_HOME`, not in other arguments, right?

mpapis:
the spaces are common concern, when writing shell scripts it's best to use arrays / param shifting to avoid space problems, from what I have seen so far in your script you just flatten down all params to single strings like with `ruby_args="$ruby_args $@"`. `sh` is very bad for handling space in parameters (quoting solves variables)

jirutka:
> the spaces are common concern

In Unix world, it’s not, because you just avoid spaces in paths and CLI arguments.

> when writing shell scripts it's best to use arrays

Arrays are not supported by ash, dash, … it’s a bash extension, not POSIX.

The problem is that arguments handling is very complicated in these jruby start scripts; it doesn’t handle just CLI arguments, but it merges them with `JRUBY_OPT` etc. Quoting doesn’t help in this case much. If you know some sane approach how to deal with it, please tell. I don’t.

headius:
@jirutka Thanks for looking at this? Here's my thoughts on the whole thing.

It would be really nice to have a better .sh script, so this is a good start. For platforms where bash is not available, we'll have a better option.

If this can support all behaviors of the bash script, it's worth making the move. However...

Past attempts to support all behavior of the bash script almost always ended up breaking because arguments could not be processed as an array. The complexity you found in our bash script has come about largely due to the complexity of argument processing, and arrays actually made that a lot cleaner and simpler.

The spaces-in-paths issue is probably the real killer here, since a sh script can only deal with strings and has to re-parse arguments. This is really the biggest reason we have to use bash arrays. And it's not sufficient to say "don't use spaces". People do, OSes do...it has to work, and this is not negotiable.

This is the primary reason why we have never been able to replace the bash script.

There may be a way to do this, though: make the shell script call bash when available.

Here's the launchers we currently have:

* jruby.sh
* jruby.bash
* jruby-launcher (native, built by gem install, shipped in Windows dist)
* mjruby (native based on mruby, pre-built binaries for a few platforms)

The native launchers were added to avoid problems with scripts, both the bash/sh scripts on *nix and the bat scripts on Windows (which were horrendous). We recommend people install them, but most do not (though I believe rvm does try to install jruby-launcher for all JRuby installs).

mjruby is expected to replace jruby-launcher for 9.1. Because it requires the mjruby build toolchain, however, it does not support all platforms that jruby-launcher does.

So...

We have jruby.bash and jruby-launcher, which are largely feature-complete but very cumbersom to maintain. We have jruby.sh which works in more places than jruby.bash, but which has feature gaps (e.g. dealing with spaces in arguments). And we have mjruby, which is as complete as jruby-launcher but only installs on a few systems.

Copying @jkutner.

I'm thinking this plan might work, and I'd like input:

1. Instead of jruby.bash being the default script, we'd use a jruby.sh that checks for bash and uses the "best" script.

   Benefits:
   * `jruby` command works everywhere that has sh.

   Issues:

   * triple-launching to get into JVM (but all execs)
   * problems for folks that get the sh version without realizing it (but they'd have to do something to even run JRuby now).
2. mjruby includes jruby-launcher, using it when there's no mjruby binary available.

   Benefits:

   * We start transitioning off jruby-launcher to a more maintainable codebase.
   * All platforms are still supported.

   Issues:

   * Lingering jruby-launcher code into the indefinite future
   * Difficulties knowing which launcher a user is reporting bugs for. Launchers should probably be modified to set a JVM property identifying themselves.

There's another possibility too. We could make jruby.sh the sole launcher in our .nix distribution, but have it recommend installing our launcher gem to get a better command line experience. That would install mjruby if supported, jruby-launcher if it can be built, or jruby.bash if bash is available. If none of those options work, `bin/jruby` would be updated to a version of jruby.sh that no longer recommends installing a better launcher.

enebo:
@headius I believe mjruby is being merged into jruby-launcher and not the other way around.

headius:
I'm removing the milestone from this.

We may need improvements for the bash script, this is true. And there are outstanding bugs. Some of those are fixed by going to the native launcher, but we will always need a working non-native launcher at least for bootstrapping.

All that said, this is a low priority for anyone working on JRuby proper right now. I think the whole PR needs a reboot, because it has become a hodge-podge of old patches and new discussions that are often working at cross purposes.

If folks out there want to see a better bash/sh script, open a new issue or PR and do the work. It's unlikely to come out of jruby core, since we all hate bash/sh and wouldn't know where to start.

PR #1987, opened 2014-09-19
sequester tmp directories

tduehr:
Moves tmp_\* directories under tmp/.


nirvdrum:
Has there been any decision on this one?

headius:
I still don't like the idea of having `tmp` committed to the repository with nothing but `.gitignore` in it. We should just fix the test to use a system-level temp location. That would be good enough, right?

tduehr:
I agree that would be better.

headius:
There was an additional problem with the test: it used `rm_f` instead of `rm_rf` to delete the `@tmpdir` in teardown. In other words, it didn't actually delete it. Both issues are fixed now.

PR #2025, opened 2014-10-07
do not change jruby commandline context classloader

mkristian:
putting the jruby-classloader as context classloader means that all attached "jars" can be found via
the Thread.currentThread.contextClassLoader. running jruby from other environments like servlets just
leave the context classloader as is. so this was a commandline only feature.

finally it keeps the classloader semantic uniform wether executed from commandline or inside a j2ee container
or osgi framework or embedded jruby in a java application.

maybe there is reason for this but this reason is not there with running any tests and any code depending on this feature of the commandline execution will cease to work when running in an environment where there runtime.getJRubyClassLoader is not the thread.contextClassLoader.

one more thing: image some jruby code loads some java which again uses the embedded JRuby via ScriptingContainer then you will chain instance of JRubyClassLoaders via the parent classloader, which is just asked for classloader problems. you all might have used ruby-maven which does exactly such JRuby -> Java -> JRuby thingy.

@enebo @ratnikov maybe one of you have a solid usecase why jruby commandline does such classloader tricks - please tell me !

### DIFF
--- a/core/src/main/java/org/jruby/Main.java
+++ b/core/src/main/java/org/jruby/Main.java
@@ -283,8 +283,6 @@ public class Main {
         }
 
         try {
-            doSetContextClassLoader(runtime);
-
             if (in == null) {
                 // no script to run, return success
                 return new Status();
@@ -449,18 +447,6 @@ public class Main {
                 throw re;
             }
             return false;
-        }
-    }
-
-    private void doSetContextClassLoader(Ruby runtime) {
-        // set thread context JRuby classloader here, for the main thread
-        try {
-            Thread.currentThread().setContextClassLoader(runtime.getJRubyClassLoader());
-        } catch (SecurityException se) {
-            // can't set TC classloader
-            if (runtime.getInstanceConfig().isVerbose()) {
-                config.getError().println("WARNING: Security restrictions disallowed setting context classloader for main thread.");
-            }
         }
     }
 


kares:
this makes sense, as otherwise setting the main thread's loader just adds to issues for different behaviour depending on how JRuby is run. still it came as a necessity to resolve loading issues (e.g. on Web-Sphere) http://jira.codehaus.org/browse/JRUBY-2495 probably no longer necessary due other load-service/class-loader changes ... but since this might be quite an invasive change I would vote for having an option to switch this kind of behaviour (which could be allowing change by default on 1.7 and maybe default to not doing any context loader changes on 9K)

kares:
for reference here's a commit that introduced the current `setContextClassLoader`: https://github.com/jruby/jruby/commit/59719b4a76807b178c21338bf74595dac91af836

mkristian:
the problem or why I find this confusing is: it ONLY happens when you run jruby via org.jruby.Main which means any code which using this feature works via org.jruby.Main  but fails to work via jruby-rack or anywhere you do not load jruby via the context classloader (like OSGi) or just any java application using ScriptingContainer

having a switch for 9k to bring back the old behaviour is OK. 1.7 branch not needed IMO

next question how should this all work when you have more then ONE runtime ? well that is a different question anyways.

mkristian:
even hibernate does work now with OSGi http://www.infoq.com/news/2013/07/hibernate-osgi which early use the context classloader to find its models defined in some xml descriptors (but it is while that I used this)

I see that there are frameworks which uses the context classloader to "find" things - maybe it is better to tell them there code works ONLY via jruby commandline and not with embedded jruby from the very beginning.

mkristian:
anyways I added tests which runs via commandline ```mvn -Prake -Dtask=test:jruby``` and non-commandline ```mvn -Pmain,test -Dinvoker.test=extended```

PR #2030, opened 2014-10-08
Missing test

tduehr:
adds missing tests (and excludes for prepend) for current MRI inheritance behavior that are missing upstream.


kares:
seems that this has been stale on MRI's end (for 2+ years): https://bugs.ruby-lang.org/issues/10248
this also now has merge conflicts. if MRI accepts these the tests will eventually reach JRuby as well ...
so I guess its safe to close (the bug can still link to the change-set since its refering to the fork-ed repo).

tduehr:
I'll create a new one to put them in the jruby specific tests for now then.

kares:
@tduehr sounds very good esp. if they already pass, thanks.

PR #2095, opened 2014-11-04
Fix #localtime and #gmtime when they are given numeric offsets

cheald:
In Ruby 2.0, `#localtime` gained the ability to parse offsets as numbers of seconds from UTC. This patch fixes that functionality (and makes the relevant tests pass) by making #localtime and #gmtime use a common arg parser to determine the time zone offset from the passed IRubyObject. Additionally, #localtime always modifies the receiver now per the documentation, rather than returning a new object if an offset
was passed.

`numExact` is a direct port of the same code in C. I suspect it can be improved, but I don't know the JRuby API well enough yet to say how that might be.

### DIFF
--- a/core/src/main/java/org/jruby/RubyTime.java
+++ b/core/src/main/java/org/jruby/RubyTime.java
@@ -213,21 +213,83 @@ public class RubyTime extends RubyObject {
         return parseTZString(runtime, zone);
     }
 
-    public static DateTimeZone getTimeZoneFromUtcOffset(Ruby runtime, String utcOffset) {
-        DateTimeZone cachedZone = runtime.getTimezoneCache().get(utcOffset);
-        if (cachedZone != null) return cachedZone;
-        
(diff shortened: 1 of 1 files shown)


enebo:
I am trying to backport this to 1.7 for 1.9 support and noticed this code is broken.   Our call and coercion methods don't work like MRIs so we will not return UNDEF but will instead raise NoMethodError.  A simple example:

```text
jruby -e 'Time.now.getlocal(Array.new)'
NoMethodError: undefined method `to_r' for []:Array
    getlocal at org/jruby/RubyTime.java:491
  __script__ at -e:1
system ~/work/jruby-1_7 jruby-1_7 *+|CHERRY-PICKING 735% mri22 -e 'Time.now.getlocal(Array.new)'
-e:1:in `getlocal': can't convert Array into an exact number (TypeError)
	from -e:1:in `<main>'
```

I will comment with my fix and merge back new version to master so you can see how I changed things to fix it.

PR #2187, opened 2014-11-12
Fix JsonProfilePrinter

dmarcotte:
`first` flag was incorrectly set, resulting in a leading comma for json profile output.

Also remove some calls to rspec's `have`, which has been removed from rspec since the profiler specs were last run.

@headius, with this the spec/profiler specs are green for 1.7.  Do you want to re-enable on Travis?


dmarcotte:
@headius just noticed this guy was unmerged... just slipped off your radar too?  (No worries if that's the case!)  Or is there something we want to address here?

Also: I just rebased to the latest `jruby-1_7`, so Travis should soon (hopefully...) confirm everything is still cool here.

PR #2221, opened 2014-11-21
Preserve ThreadGroup#list order with WeakIdentityLinkedHashSet

godfat:
Because MRI uses a linked list which preserves the order of the threads.
https://github.com/ruby/ruby/blob/2a754a733045da9965e88d1f31e650ea6b3f2b6c/vm_core.h#L973-L978
This is useful to assume the first thread in the list is the root of a
group, and saving values in the thread could be shared amongst the
thread group. If we ever have thread group local variable, this is not
really needed.

### DIFF
--- a/core/src/main/java/org/jruby/RubyThreadGroup.java
+++ b/core/src/main/java/org/jruby/RubyThreadGroup.java
@@ -29,6 +29,7 @@
  ***** END LICENSE BLOCK *****/
 package org.jruby;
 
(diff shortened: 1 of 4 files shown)


headius:
It turns out this behavior was not specified in Ruby, and MRI stopped keeping them in order with the 2.2 release. Until they decide to specify this behavior, we will wait to make this change.

PR #2233, opened 2014-11-24
Comparable#== raises SystemStackError when calling classes don't impleme...

mjgpy3:
...nt #<=>

Using recursive pattern


headius:
Finally circling back to PRs...

Test should be generally green on master now. This were in flux during November and December.

The patch is ok, but it doesn't pay attention to the recursion guard's "recur" argument (the boolean you have named `___`). recur == true should trigger the inspect output to insert special text indicating a recursion.

MSNexploder:
#3006 fixed that

kares:
closing since #3006 has been merged

Issue #2142, opened 2014-11-07
Struct#inspect does not produce the same output as MRI 2.2


headius:
Couple failing MRI tests due to subtle changes in Struct#inspect:

```
PPTestModule::PPCycleTest#test_struct [/Users/headius/projects/jruby/test/mri/test_pp.rb:140]:
<"#<struct Struct::S :...>\n"> expected but was
<"#<struct Struct::S a=1, b=#<struct Struct::S:...>>\n">.

TestStruct::SubStruct#test_inspect [/Users/headius/projects/jruby/test/mri/ruby/test_struct.rb:121]:
Expected /^#<struct a=#<struct #<.*?>:...>>$/ to match "#<struct #<Class:0x2ccc231d> :...>".
```

That last failure is repeated through two other forms of the struct tests.

PR #2285, opened 2014-12-06
fix Struct#inspect with recursion

lumeet:
Referring to jruby/jruby#2142 the details regarding differences between MRI and JRuby recursion guards are way beyond my understanding. If the code changes suck, I hope this PR at least explains what the problem is. :)

### DIFF
--- a/core/src/main/java/org/jruby/RubyStruct.java
+++ b/core/src/main/java/org/jruby/RubyStruct.java
@@ -51,6 +51,8 @@ import org.jruby.runtime.marshal.UnmarshalStream;
 import org.jruby.util.ByteList;
 import org.jruby.util.IdUtil;
 
+import java.util.concurrent.Callable;
+
 import static org.jruby.RubyEnumerator.enumeratorizeWithSize;
 import static org.jruby.runtime.Helpers.invokedynamic;
 import static org.jruby.runtime.Visibility.PRIVATE;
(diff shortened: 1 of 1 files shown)


headius:
PR looks fine to me. Will try it locally.

headius:
Well the changes look fine, but they don't make the test I reported in #2142 pass:

```
~/projects/jruby $ jruby test/mri/runner.rb test_pp.rb
Run options: 

# Running tests:

[ 7/20] PPTestModule::PPCycleTest#test_withinspect = 0.01 s      
  1) Failure:
PPTestModule::PPCycleTest#test_withinspect [/Users/headius/projects/jruby/test/mri/test_pp.rb:157]:
<"[<inspect:[...]>]\n"> expected but was
<"[<inspect:[<inspect:[...]>]>]\n">.
```

Seems like it's doubling up the "inspect:" part.

lumeet:
@headius Don't know how I missed that one. It also seems to be a completely other story, which extends beyond the beginner level, I think. I've been investigating the issue quite a bit but don't know how to continue. `Array#inspect` should perhaps be more like that in `Struct`, which I tried with little success.

As far as I know the first line should be affected by the latter:
https://github.com/jruby/jruby/blob/master/core/src/main/java/org/jruby/Ruby.java#L4094
https://github.com/jruby/jruby/blob/master/lib/ruby/stdlib/pp.rb#L164

This seems not to be the case and maybe it's the root of the problem? In MRI, the behavior is the same as in JRuby now if the latter line is left out.

I think I'm going to leave the last part of the issue to someone with a bit more experience of the subject.

kares:
adapted your commit to fit with master updates: https://github.com/jruby/jruby/commit/33bb8d2dc6bd7856965cc0bc8e9b815407439f76?w=1 ... `Struct#inspect` now behaves as MRI the pp test fails the same but as you already find out its a fix to be made elsewhere (not in `Struct`)

PR #2292, opened 2014-12-09
Remove unnecessary puts for GC::Profiler.total_time

zanker:
Calling `GC::Profiler.total_time` dumps the headers used by `GC::Profiler.result` when the `GC::Profiler::NotifiedProfiler` is being used. Given the method doesn't dump anything, it looks like this isn't needed and could be dropped.


headius:
You're right, looks like copypasta.

Issue #2182, opened 2014-11-12
Struct#inspect has ASCII encoding


obrok:
On ruby 2.1 and 1.9:

```ruby
irb(main):001:0> Struct.new(:a).new("ΆἅἇἈ").inspect
=> "#<struct a=\"ΆἅἇἈ\">"
irb(main):002:0> Struct.new(:a).new("ΆἅἇἈ").inspect.encoding
=> #<Encoding:UTF-8>
```

On jruby 1.7.16:

```ruby
irb(main):001:0> Struct.new(:a).new("ΆἅἇἈ").inspect
=> "#<struct a=\"\xCE\x86\xE1\xBC\x85\xE1\xBC\x87\xE1\xBC\x88\">"
irb(main):002:0> Struct.new(:a).new("ΆἅἇἈ").inspect.encoding
=> #<Encoding:ASCII-8BIT>
```

PR #2321, opened 2014-12-15
Fix for issue 2182 on jruby-1_7 : Struct#inspect with utf8 encode string member

k77ch7:
This commit fixes issue #2182 on jruby-1_7 branch. Inspect method of Struct with utf8 string member should return a utf8 string.


headius:
@k77ch7 Since your second commit corrects flaws (duplication of cat19 logic) in the first one, can you force push this all as a single commit please?

Thanks for your help!

k77ch7:
@headius Thanks for reviewing. I just pushed.

PR #2334, opened 2014-12-18
Replace array flattening with manual handling in java_import

### LINKED ISSUE #2324: Array#flatten! call in java_import breaks if we import a class with static to_ary
I attempted to java_import [org.jruby.util.TypeConverter](https://github.com/jruby/jruby/blob/master/core/src/java/org/jruby/util/TypeConverter.java) and ran into this error inside Array#flatten! logic:

```
ArgumentError: wrong number of arguments (0 for 2)
     flatten! at org/jruby/RubyArray.java:2839
  java_import at file:/Users/headius/projects/jruby/lib/jruby.jar!/jruby/java/core_ext/object.rb:20
       (root) at file:/Users/headius/projects/jruby/lib/jruby.jar!/jruby/kernel/jruby/type.rb:3
       (root) at file:/Users/headius/projects/jruby/lib/jruby.jar!/jruby/kernel.rb:1
       (root) at file:/Users/headius/projects/jruby/lib/jruby.jar!/jruby/kernel.rb:4
         load at org/jruby/RubyKernel.java:969
       (root) at :1
```

The cause: flatten attempts to coerce elements to Array by calling to_ary with no arguments, but TypeConverter defines a static utility method to_ary(ThreadContext, IRubyObject).

The associated lines of code are seen in the trace above. I think we should just back off flatten! call and do something dumber and more manual that just requires everything already be an Array. That will avoid any future issues with unusual method names getting called for coercion.

lumeet:
Fixed #2324 by getting rid of flatten! as hinted by @headius. I thought about writing a test for the method, too, as I couldn't find any. Too bad, I couldn't find a way to run a single test in tests/jruby either. Is that the right place for such tests? Help appreciated.

### DIFF
--- a/core/src/main/ruby/jruby/java/core_ext/object.rb
+++ b/core/src/main/ruby/jruby/java/core_ext/object.rb
@@ -16,8 +16,20 @@ class Object
     return other.java_class.assignable_from?(self.java_class)
   end
 
+  # Import one or many Java classes as follows:
+  #
+  #   java_import java.lang.System
+  #   java_import java.lang.System, java.lang.Thread
+  #   java_import [java.lang.System, java.lang.Thread]
+  #
   def java_import(*import_classes)
-    import_classes.flatten!
+    import_classes = import_classes.each_with_object([]) do |classes, flattened|
+                       if classes.is_a?(Array)
+                         flattened.push *classes
+                       else
+                         flattened.push classes
+                       end
+                     end
 
     import_classes.map do |import_class|
       case import_class


headius:
You could put the tests in test/jruby (they should be runnable directly, with only a few counterexamples) or in spec/regression. Former would probably be best since usually the latter is for Ruby behaviors (rather than JRuby behaviors) that aren't tested well in other suites.

FWIW, this will get exercised heavily by other Java integration tests too.

headius:
Two comments on the commit:

* If there's any *array-like* objects passed to java_import, this will cause them to no longer coerce. It seems highly unlikely that someone would have that case in real code, though, and I don't think we've ever officially supported such wildly-structured import lists.
* Formatting is a little weird in the patch. I can fix that on the way in.

We'll only make this change on master, since it does have a potentially-visible behavior change.

Issue #1487, opened 2014-02-06
Ruby 2.0: Object.const_get throws a "NameError: wrong constant name" when called with leading ::


slowjack2k:
Object.const_get throws a "NameError: wrong constant name" when called with leading ::

Example:

```ruby
module A 
  class B 
  end 
end

Object.const_get '::A::B'

#=> A::B with ruby 2.0
#=> NameError: wrong constant name with JRUBY_OPTS=--2.0 jruby-1.7.10
```

PR #2339, opened 2014-12-21
Fix Module.const_get for strings beginning with "::"

lumeet:
Yes, #1487 seems super easy.

### DIFF
--- a/core/src/main/java/org/jruby/RubyModule.java
+++ b/core/src/main/java/org/jruby/RubyModule.java
@@ -2816,6 +2816,10 @@ public class RubyModule extends RubyObject {
         String symbol = fullName;
         boolean inherit = args.length == 1 || (!args[1].isNil() && args[1].isTrue());
 
+        if (symbol.indexOf("::") == 0) {
+            symbol = symbol.substring(2);
+        }
+
         RubyModule mod = this;
         int sep;
         while((sep = symbol.indexOf("::")) != -1) {


headius:
I'm not sure this is the right fix. For example, the following case should get the top-level Foo, and I don't think it will in your patch:

```ruby
Foo = 1
module Bar
end
Bar.const_get('::Foo')
```

lumeet:
Hmm... am I missing something here?

test.rb:
```ruby
Foo = 1
module Bar
end
puts Bar.const_get('::Foo')

module A 
  class B 
  end 
end

put Object.const_get('::A::B')
```
Using MRI 2.2 and patched JRuby:
```shell
$ rbenv shell 2.2.0
$ ruby test.rb
1
A::B

$ bin/jruby test.rb
1
A::B
```

headius:
Object.const_get('::A::B') is equivalent to Object.const_get('A::B'). Your other case does appear to show it working right, so perhaps I reviewed it poorly. Will try it locally.

headius:
Ok, I have a case that better illustrates the problem:

```ruby
Foo = 1
module Bar; Foo = 2; end
Bar.const_get('::Foo')
```

Basically a leading :: should cause it to start searching at Object, rather than from the target class.

lumeet:
Now this is a perfect example of sticking to test cases instead of thinking. I'll take a look at this again. :)

headius:
@lumeet There are very likely test cases in MRI's suite (rake test:mri, test/mri/socket/**) or in RubySpec (rake test:ruby:fast, spec/ruby/core/socket/**) if you want to try getting them to run. If so, you can also remove tags/excludes from spec/tags or test/mri/excludes.

headius:
BTW, you can get access to Object with `runtime.getObject`.

headius:
Fixed via 5868779.

PR #2348, opened 2014-12-26
Implicit type conversion error + String#start_with? and #end_with? error

lumeet:
As of 2.0, MRI raises "no implicit conversion..." `TypeError` in certain situations. I've tried to make respective changes to `TypeConverter`. While not being sure it's 100%  right, at least the following also seems to behave correctly with the patch.
```bash
$ rbenv shell 2.0.0-p598
$ ruby -e '"a" + 1'
no implicit conversion of Fixnum into String (TypeError)

$ rbenv shell jruby-9.0.0.0-dev
$ ruby -e '"a" + 1'
TypeError: can't convert Fixnum into String
```
The String part addresses then `test_start_with?`and `test_end_with?` in test_string.rb


headius:
Thank you!

Issue #1675, opened 2014-05-01
casecmp on UTF-16LE encoded string produces ArrayIndexOutOfBoundsException


tommunist:
Having some trouble with UTF-16LE encoded strings in jruby...
Tried the following in jruby-1.7.9:
```ruby
>> a = 'ABC'.encode('UTF-16LE')
=> "ABC"
>> b = 'ABC'.encode('UTF-16LE')
=> "ABC"
>> b.casecmp(a)
```
which produced:
```ruby
Java::JavaLang::ArrayIndexOutOfBoundsException: -1
	from org.jruby.RubyString.multiByteCasecmp(RubyString.java:1661)
	from org.jruby.RubyString.casecmp19(RubyString.java:1636)
	from org.jruby.RubyString$INVOKER$i$1$0$casecmp19.call(RubyString$INVOKER$i$1$0$casecmp19.gen)
	from org.jruby.runtime.callsite.CachingCallSite.cacheAndCall(CachingCallSite.java:326)
	from org.jruby.runtime.callsite.CachingCallSite.call(CachingCallSite.java:170)
	from org.jruby.ast.CallOneArgNode.interpret(CallOneArgNode.java:57)
	from org.jruby.ast.NewlineNode.interpret(NewlineNode.java:105)
	from org.jruby.ast.RootNode.interpret(RootNode.java:129)
	from org.jruby.evaluator.ASTInterpreter.INTERPRET_EVAL(ASTInterpreter.java:95)
	from org.jruby.evaluator.ASTInterpreter.evalWithBinding(ASTInterpreter.java:184)
	from org.jruby.RubyKernel.evalCommon(RubyKernel.java:1156)
	from org.jruby.RubyKernel.eval19(RubyKernel.java:1119)
	from org.jruby.RubyKernel$INVOKER$s$0$3$eval19.call(RubyKernel$INVOKER$s$0$3$eval19.gen)
	from org.jruby.internal.runtime.methods.DynamicMethod.call(DynamicMethod.java:180)
	from org.jruby.runtime.callsite.CachingCallSite.call(CachingCallSite.java:70)
	from org.jruby.ast.FCallManyArgsNode.interpret(FCallManyArgsNode.java:60)
... 114 levels...
	from org.jruby.ast.BlockNode.interpret(BlockNode.java:71)
	from org.jruby.evaluator.ASTInterpreter.INTERPRET_METHOD(ASTInterpreter.java:74)
[...]

PR #2349, opened 2014-12-27
Fix for issue 1675 on jruby-1_7: String#casecmp on UTF-16LE encoded string

k77ch7:
This commit fixes issue #1675 on jruby-1_7 branch. I wrote a rspec test using the available encodings in addition to UTF-16LE. See *GH-1675_casecmp_on_UTF16LE_encoded_string_spec.rb* for details.


headius:
Looks good.

PR #2405, opened 2015-01-02
Module#include is public for a while now.

godfat:
Could verify this by running:

    ruby -ve 'Module.new.include Module.new'

We should probably have a test for this somewhere.


headius:
Sorry I missed this; we made the change on master already.

PR #2406, opened 2015-01-02
Introduce condition profiling for AndNode.

thomaswue:
Use full condition profiling including counters that inject probabilities for AndNode.


chrisseaton:
Travis failed, but it's an unrelated intermittent failure. Merging.

PR #2414, opened 2015-01-03
Truffle rbx integration

chrisseaton:
This is the first big push towards implementing the Rubinius primitive system. Before we were using the standard library `Struct` from Rubinius, which was pure Ruby code. Here I've pulled in `Rational` and `Time`, which is tightly integrated into the Rubinius primitive system (they're like native calls).

Doing this we've got the majority of the specs passing for both of these classes in about a day. The work we're doing here will also apply for some standard library classes which also uses these primitives.

This is quite a lot of code here, but you can skip most of the Ruby files. To help you understand what's going on, I've added one comment where Ruby code calls a primitive, and another where that's implemented in Java.

@eregon, @nirvdrum, @thomaswue please review

@headius, @enebo, if you want to keep up with what we're doing, this might be a good commit to cast an eye over, but don't feel obliged to spend your time reviewing our code


core/src/main/ruby/jruby/truffle/core/rubinius/kernel/bootstrap/time.rb
```diff
@@ -25,11 +25,10 @@
 # OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
 
 class Time
-  # TODO(CS)
-  #def self.now
-  #  Rubinius.primitive :time_s_now
-  #  raise PrimitiveFailure, "Time.now primitive failed"
-  #end
+  def self.now
+    Rubinius.primitive :time_s_now
```
chrisseaton: Primitive call here. Note the unusual semantics - fall through for cases not handled.

core/src/main/java/org/jruby/truffle/nodes/rubinius/TimePrimitiveNodes.java
```diff
@@ -0,0 +1,315 @@
...
+
+        public TimeSNowPrimitiveNode(TimeSNowPrimitiveNode prev) {
+            super(prev);
+        }
+
+        @Specialization
+        public RubyTime timeSNow(RubyClass timeClass) {
+            return new RubyTime(timeClass,
+                            TimeOperations.millisecondsToSeconds(System.currentTimeMillis()),
+                            TimeOperations.millisecondsToNanoseconds(TimeOperations.millisecondsInCurrentSecond(System.currentTimeMillis())));
```
chrisseaton: Primitive call implemented here.

core/src/main/java/org/jruby/truffle/translator/BodyTranslator.java
```diff
@@ -346,9 +351,124 @@ public RubyNode visitBreakNode(org.jruby.ast.BreakNode node) {
...
+        }
+
+        for (int n = 0; n < argumentsCount; n++) {
+            arguments.add(new ReadPreArgumentNode(context, sourceSection, n, MissingArgumentBehaviour.UNDEFINED));
+        }
+
+        return new CallRubiniusPrimitiveNode(context, sourceSection,
+                primitive.getFactory().createNode(context, sourceSection, arguments.toArray(new RubyNode[arguments.size()])),
+                environment.getReturnID());
+    }
```
chrisseaton: Primitive call replaced with a node here.

core/src/main/java/org/jruby/truffle/nodes/rubinius/CallRubiniusPrimitiveNode.java
```diff
@@ -0,0 +1,59 @@
...
+
+        if (value != null) {
+            // If they didn't fail it's a return in the calling method
+
+            throw new ReturnException(returnID, value);
+        }
+
+        // The code after a primitive call is a fallback, so it makes sense to cut it off
+
+        branchProfile.enter();
```
thomaswue: Using a ConditionProfile on the if statement is maybe more intuitive than using a BranchProfile. Also, the first comment could go after the if branch.

core/src/main/java/org/jruby/truffle/nodes/rubinius/RubyTimeToDateTimeNode.java
```diff
@@ -0,0 +1,73 @@
...
+
+        // The @is_gmt instance varibale is only for internal use so we don't need a full cast here
+
+        final boolean isGMT;
+
+        if (isGMTObject instanceof Boolean && ((boolean) isGMTObject)) {
+            isGMT = true;
+        } else {
+            isGMT = false;
+        }
```
thomaswue: A full cast should not be slower and makes the code work under less global assumptions? In any case, the partial cast could be extracted into a utility method. If it is internal use, can there be assertions about the valid expected values or can indeed be an arbitrary value in the isGMTObject local variable?

core/src/main/java/org/jruby/truffle/nodes/rubinius/StringPrimitiveNodes.java
```diff
@@ -0,0 +1,64 @@
...
+        }
+
+        public StringCheckNullSafePrimitiveNode(StringCheckNullSafePrimitiveNode prev) {
+            super(prev);
+        }
+
+        @Specialization
+        public boolean stringCheckNullSafe(RubyString string) {
+            for (byte b : string.getBytes().unsafeBytes()) {
+                if (b == 0) {
```
thomaswue: I think this condition could benefit from a 0/1 condition profile.

core/src/main/java/org/jruby/truffle/nodes/rubinius/TimePrimitiveNodes.java
thomaswue: The two calls to System.currentTimeMillis() should be reduced to one call saving the value to a local variable.

core/src/main/java/org/jruby/truffle/nodes/rubinius/TimePrimitiveNodes.java
```diff
@@ -0,0 +1,315 @@
...
+    }
+
+    @RubiniusPrimitive(name = "time_s_dup", needsSelf = false)
+    public static abstract class TimeSDupPrimitiveNode extends RubiniusPrimitiveNode {
+
+        @Child protected ReadHeadObjectFieldNode readIsGMTNode = new ReadHeadObjectFieldNode("@is_gmt");
+        @Child protected ReadHeadObjectFieldNode readOffsetNode = new ReadHeadObjectFieldNode("@offset");
+
+        @Child protected WriteHeadObjectFieldNode writeIsGMTNode = new WriteHeadObjectFieldNode("@is_gmt");
+        @Child protected WriteHeadObjectFieldNode writeOffsetNode = new WriteHeadObjectFieldNode("@offset");
```
thomaswue: It is unrelated to these changes, but the constructors for ReadHeadObjectFieldNode and WriteHeadObjectFieldNode seem good candidates for being extracted into factory methods.

core/src/main/java/org/jruby/truffle/nodes/rubinius/TimePrimitiveNodes.java
```diff
@@ -0,0 +1,315 @@
...
+        }
+
+        @Specialization
+        public RubyArray timeDecompose(VirtualFrame frame, RubyTime time) {
+            final DateTime dateTime = toDateTimeNode.toDateTime(frame, time);
+            final Object[] decomposed = decompose(dateTime);
+            return new RubyArray(getContext().getCoreLibrary().getArrayClass(), decomposed, decomposed.length);
+        }
+
+        @CompilerDirectives.TruffleBoundary
```
thomaswue: This method should probably not have a TruffleBoundary annotation. After inlining, Graal's partial escape analysis will reduce this method to almost no code and maybe avoid the allocation of the array altogether.

core/src/main/java/org/jruby/truffle/translator/BodyTranslator.java
```diff
@@ -346,9 +351,124 @@ public RubyNode visitBreakNode(org.jruby.ast.BreakNode node) {
...
+        }
+
+        currentCallMethodName = "privately";
+        privately = true;
+
+        try {
+            return (((org.jruby.ast.IterNode) node.getIterNode()).getBodyNode()).accept(this);
+        } finally {
+            privately = false;
+        }
```
thomaswue: This scoping with the privately variable could be described in a comment. What are you trying to achieve here?

thomaswue:
And as a general comment: This looks like awesome progress and the absolute right direction to go (i.e., implementing large parts of the standard library in Ruby itself). This gives a couple of hundreds passing specs in just one go :)!

core/src/main/java/org/jruby/truffle/nodes/rubinius/TimePrimitiveNodes.java
chrisseaton: Ah yes of course - how silly - fixed.

core/src/main/java/org/jruby/truffle/nodes/rubinius/CallRubiniusPrimitiveNode.java
chrisseaton: Fixed.

core/src/main/java/org/jruby/truffle/nodes/rubinius/RubyTimeToDateTimeNode.java
chrisseaton: There's an extra complication here in that `ReadHeadObjectFieldNode` is a lower level node and returns `null` for an instance variable that hasn't been set. Higher level nodes like `BooleanCastNode` won't accept a `null`. I'll have a think about how to fix this...

core/src/main/java/org/jruby/truffle/nodes/rubinius/StringPrimitiveNodes.java
chrisseaton: Fixed.

core/src/main/java/org/jruby/truffle/nodes/rubinius/TimePrimitiveNodes.java
chrisseaton: This is the only place in the the codebase where we have pairs of read-write nodes, so it's not a common pattern.

I'm not sure about the more general factory method comment. We've just got a simple constructor here - what the benefit of wrapping that in a factory?

core/src/main/java/org/jruby/truffle/nodes/rubinius/TimePrimitiveNodes.java
chrisseaton: This is the Joda DateTime class - some of the routines such as `getDayOfYear` do not look simple and might not be suitable for PE. I'll check with @eregon as he knows this library very well.

core/src/main/java/org/jruby/truffle/nodes/rubinius/RubyTimeToDateTimeNode.java
thomaswue: OK, understand. Either they need to return undefined (or whatever would be returned based on language semantics) or BooleanCastNode needs to accept null. The former seems to me the better solution.

core/src/main/java/org/jruby/truffle/nodes/rubinius/TimePrimitiveNodes.java
thomaswue: OK, then combining them is probably not worth it. I still think it would be nicer to have the name of the methods be "read" / "write" instead of "execute". The benefit of the factory would be that we are free to return subtypes (e.g., when profiling or debugging flags are enabled or when we decide to change the class hierarchy).

chrisseaton: I hadn't thought about renaming them to `read` instead of `execute`. I'll do that when we're merged and can refactor.

I remember now that JS has always had factory methods for all nodes. I'll look again at converting everything to using factory methods when I next do a big refactor. We can probably find some common patterns in the codebase now that would benefit from that.

core/src/main/java/org/jruby/truffle/nodes/rubinius/TimePrimitiveNodes.java
thomaswue: OK, I thought these would be simple getters. The Joda library looks indeed very powerful with lots of classes and abstractions. Do we have any alternatives or is most of the functionality really required?

core/src/main/java/org/jruby/truffle/nodes/core/BignumNodes.java
```diff
@@ -192,7 +192,7 @@ public RubyBignum pow(RubyBignum a, RubyBignum b) {
 
     }
 
-    @CoreMethod(names = "/", required = 1)
+    @CoreMethod(names = {"/", "__slash__"}, required = 1)
```
nirvdrum: Why the addition of `__slash__`?

chrisseaton: Rubinius calls it from here https://github.com/jruby/jruby/pull/2414/files#diff-1eabf0e4bde8dc8dce596aa4d27d7042R41. I didn't do the digging yet to figure out if it's supposed to mean something special - such as never monkey patched.

core/src/main/java/org/jruby/truffle/runtime/core/TimeOperations.java
```diff
@@ -0,0 +1,38 @@
...
+ */
+package org.jruby.truffle.runtime.core;
+
+public abstract class TimeOperations {
+
+    public static long secondsAndNanosecondsToMiliseconds(long seconds, long nanoseconds) {
+        return secondsToMiliseconds(seconds) + nanosecondsToMilliseconds(nanoseconds);
+    }
+
+    public static long secondsToMiliseconds(long seconds) {
```
nirvdrum: Typo.  Should be `secondsToMilliseconds`.

core/src/main/java/org/jruby/truffle/runtime/core/TimeOperations.java
```diff
@@ -0,0 +1,38 @@
...
+ *
+ * Eclipse Public License version 1.0
+ * GNU General Public License version 2
+ * GNU Lesser General Public License version 2.1
+ */
+package org.jruby.truffle.runtime.core;
+
+public abstract class TimeOperations {
+
+    public static long secondsAndNanosecondsToMiliseconds(long seconds, long nanoseconds) {
```
nirvdrum: Typo. Should be `secondsAndNanosecondsToMilliseconds`.

core/src/main/java/org/jruby/truffle/nodes/rubinius/StringPrimitiveNodes.java
```diff
@@ -0,0 +1,67 @@
...
+                    return false;
+                }
+            }
+
+            return true;
+        }
+
+    }
+
+    @RubiniusPrimitive(name = "string_to_f", needsSelf = false)
```
nirvdrum: I cross-referenced the Rubinius repo with the v2.4.1 tag just to see how this primitive stuff was implemented.  It looks like this one is commented out in vm/builtin/string.hpp.  It's referenced in kernel/bootstrap/type.rb, however.  In any event, if you could provide a pointer of what the equivalent in Rubinius is, that'd be very helpful.

core/src/main/java/org/jruby/truffle/nodes/rubinius/TimePrimitiveNodes.java
nirvdrum: The new java.time classes in Java 8 are based on a variation of Joda.  I'm not sure if that simplifies the complexity or not.  But if we used those instead we'd need to still be able to compile on Java 6 & 7 like the rest of JRuby.

core/src/main/java/org/jruby/truffle/nodes/rubinius/TimePrimitiveNodes.java
```diff
@@ -0,0 +1,316 @@
...
+
+        @Specialization
+        public RubyString timeStrftime(VirtualFrame frame, RubyTime time, RubyString format) {
+            return getContext().makeString(format(toDateTimeNode.toDateTime(frame, time), time.getNanoseconds(), format.getBytes()));
+        }
+
+
+        @CompilerDirectives.TruffleBoundary
+        public ByteList format(DateTime time, long nanoseconds, ByteList format) {
+            // TODO: converts everything to JRuby objects and back - should find a more direct way using ByteList
```
nirvdrum: This is entirely a personal thing, but I've found it very helpful to annotate TODOs with an author (initials or handle) and date.  Some will claim that git blame would be sufficient, but the moment someone else touches that line (e.g., moves it) both the author and the date look wrong unless you dig far enough back.  If we want to got this way, we could settle on a date format.  I've been annotating with a full date to avoid any ambiguity between US formats and whatever crazy thing you guys use.

core/src/main/java/org/jruby/truffle/translator/BodyTranslator.java
```diff
@@ -346,9 +351,139 @@ public RubyNode visitBreakNode(org.jruby.ast.BreakNode node) {
...
+            throw new UnsupportedOperationException("Rubinius.invoke_primitive must have at least an initial literal symbol argument");
+        }
+
+        final String primitiveName = ((org.jruby.ast.SymbolNode) node.getArgsNode().childNodes().get(0)).getName();
+
+        final RubiniusPrimitiveConstructor primitive = context.getRubiniusPrimitiveManager().getPrimitive(primitiveName);
+
+        final List<RubyNode> arguments = new ArrayList<>();
+
+        for (org.jruby.ast.Node arg : node.getArgsNode().childNodes().subList(1, node.getArgsNode().childNodes().size())) {
```
nirvdrum: It took me a minute to realize you were skipping the primitive name here by starting at 1.  Maybe it's more obvious than I realized, but a comment might help.

core/src/main/ruby/jruby/truffle/core/rubinius/kernel/bootstrap/time.rb
nirvdrum: Thanks.  I wasn't following how this works.

nirvdrum:
This looks great.  I had a few small comments, but otherwise thumbs up from me to merge.

core/src/main/java/org/jruby/truffle/nodes/rubinius/TimePrimitiveNodes.java
thomaswue: OK, let's not go there then for now. When we are passing all date/time specs, we can revisit the question.

core/src/main/java/org/jruby/truffle/nodes/rubinius/TimePrimitiveNodes.java
thomaswue: Yes, I agree that this is useful and could be a convention.

core/src/main/java/org/jruby/truffle/translator/BodyTranslator.java
thomaswue: In general, this looks like a very Ruby-style way of expressing this. Requesting an iterator and skipping with next over the first is more efficient in Java.

core/src/main/java/org/jruby/truffle/nodes/rubinius/TimePrimitiveNodes.java
eregon: @thomaswue I think we need most of the functionality for the Date/DateTime class (with proper GJChronology and it is even configurable to set when is the reform between Julian and Gregorian).

For Time, it is a more traditional UNIX timestamp (with arbitrary precision, so Joda-Time is not perfect here (it uses a long) but the limit is so large probably no one will ever care). A simple UNIX timestamp also makes it not very precise in years before 1970. Having a common interface probably simplifies things and ideally should not have any significant overhead.
Storing it only as a timestamp will need relatively complex computations to get the day, are you thinking to also store it as Y-m-d H:M:S or cache it? I guess working only with the decomposed form is very tedious and also slows for some operations.

core/src/main/java/org/jruby/truffle/nodes/core/BignumNodes.java
eregon: Probably their way to not be polluted by `mathn` which redefines `/`.
I wonder why `div` or `quo` is not sufficient though.

core/src/main/java/org/jruby/truffle/nodes/core/BignumNodes.java
```diff
@@ -603,6 +576,75 @@ public Object leftShift(RubyBignum a, int b) {
...
+        }
+
+        public EvenNode(EvenNode prev) {
+            super(prev);
+        }
+
+        @CompilerDirectives.TruffleBoundary
+        @Specialization
+        public boolean even(RubyBignum value) {
+            return value.bigIntegerValue().getLowestSetBit() != 0;
```
eregon: There's got to be a better way for this, but it seems even `BigInteger.intValue()` goes to crazy lengths ...

core/src/main/java/org/jruby/truffle/nodes/rubinius/CallRubiniusPrimitiveNode.java
```diff
@@ -0,0 +1,55 @@
...
+import com.oracle.truffle.api.utilities.ConditionProfile;
+import org.jruby.truffle.nodes.RubyNode;
+import org.jruby.truffle.runtime.RubyContext;
+import org.jruby.truffle.runtime.control.ReturnException;
+
+/**
+ * Node which wraps a {@link RubiniusPrimitiveNode}, providing the implicit control flow that you get with calls to
+ * Rubinius primitives.
+ */
+public class CallRubiniusPrimitiveNode extends RubyNode {
```
eregon: +1 for not having a boolean flag all around!

core/src/main/java/org/jruby/truffle/nodes/rubinius/RubyTimeToDateTimeNode.java
```diff
@@ -0,0 +1,73 @@
...
+
+    @CompilerDirectives.TruffleBoundary
+    private DateTime toDateTime(long seconds, long nanoseconds, boolean isGMT, Object offset) {
+        final DateTimeZone dateTimeZone;
+
+        if (isGMT) {
+            dateTimeZone = DateTimeZone.UTC;
+        } else {
+            dateTimeZone = org.jruby.RubyTime.getLocalTimeZone(context.getRuntime());
+        }
```
eregon: `offset` should be used here, no?
Also, `getLocalTimeZone()` might change due to the env var `TZ` being redefined (used in specs).

core/src/main/java/org/jruby/truffle/nodes/rubinius/TimePrimitiveNodes.java
```diff
@@ -0,0 +1,316 @@
...
+            final long milliseconds = System.currentTimeMillis();
+            return new RubyTime(timeClass,
+                            TimeOperations.millisecondsToSeconds(milliseconds),
+                            TimeOperations.millisecondsToNanoseconds(TimeOperations.millisecondsInCurrentSecond(milliseconds)));
+        }
+
+    }
+
+    @RubiniusPrimitive(name = "time_s_dup", needsSelf = false)
+    public static abstract class TimeSDupPrimitiveNode extends RubiniusPrimitiveNode {
```
eregon: How is this different than Kernel#dup?

core/src/main/java/org/jruby/truffle/nodes/rubinius/TimePrimitiveNodes.java
```diff
@@ -0,0 +1,316 @@
...
+            final int hour = dateTime.getHourOfDay();
+            final int day = dateTime.getDayOfMonth();
+            final int month = dateTime.getMonthOfYear();
+            final int year = dateTime.getYear();
+            final int wday = dateTime.getDayOfWeek();
+            final int yday = dateTime.getDayOfYear();
+            final Object isdst = getContext().getCoreLibrary().getNilObject();
+            final Object zone = getContext().getCoreLibrary().getNilObject();
+            return new Object[]{sec, min, hour, day, month, year, wday, yday, isdst, zone};
+        }
```
eregon: If we want any perf on Time using the Rubinius lib directly, we need to be able to PE this so only the interesting element of the array is computed. But I am thinking we should just define `day`,`month`, ... ourselves.

core/src/main/java/org/jruby/truffle/nodes/rubinius/TimePrimitiveNodes.java
```diff
@@ -0,0 +1,316 @@
...
+        public RubyString timeStrftime(VirtualFrame frame, RubyTime time, RubyString format) {
+            return getContext().makeString(format(toDateTimeNode.toDateTime(frame, time), time.getNanoseconds(), format.getBytes()));
+        }
+
+
+        @CompilerDirectives.TruffleBoundary
+        public ByteList format(DateTime time, long nanoseconds, ByteList format) {
+            // TODO: converts everything to JRuby objects and back - should find a more direct way using ByteList
+            final RubyDateFormatter rdf = getContext().getRuntime().getCurrentContext().getRubyDateFormatter();
+            return rdf.compileAndFormat(org.jruby.RubyString.newString(getContext().getRuntime(), format), false, time, nanoseconds, null).getByteList();
```
eregon: I did a lot of work on this, I'll need to check why I kept using a RubyString as input.

core/src/main/java/org/jruby/truffle/nodes/core/BignumNodes.java
chrisseaton: `__slash__` is used to define `div`.

core/src/main/java/org/jruby/truffle/nodes/core/BignumNodes.java
chrisseaton: BigInteger is not highly optimised, so this is probably the least of our concerns with the performance of Bignum. It is cached, however, as it's used internally for many other operations.

core/src/main/java/org/jruby/truffle/nodes/rubinius/RubyTimeToDateTimeNode.java
chrisseaton: Yeah, I glossed over both those issues as I couldn't quite understand them at at the time. I'm sure it will be caught when I go through the rest of the specs.

core/src/main/java/org/jruby/truffle/nodes/rubinius/StringPrimitiveNodes.java
chrisseaton: It's possible that Rubinius is happy to parse code with undefined primitives, where we through an error so I had to define all those used.

core/src/main/java/org/jruby/truffle/nodes/rubinius/TimePrimitiveNodes.java
chrisseaton: It sets the native fields of seconds and nanoseconds for one. Then it just seemed easier to directly copy the fields, but now I think about a simple call to `clone` might be better.

core/src/main/java/org/jruby/truffle/nodes/rubinius/TimePrimitiveNodes.java
chrisseaton: We could do something such as a return something that looks like an array but really stores a `DateTime` and lazily calculates these values.

core/src/main/java/org/jruby/truffle/nodes/rubinius/TimePrimitiveNodes.java
chrisseaton: Will do.

core/src/main/java/org/jruby/truffle/translator/BodyTranslator.java
chrisseaton: I'll change it.

core/src/main/java/org/jruby/truffle/nodes/rubinius/TimePrimitiveNodes.java
thomaswue: That sounds a lot of complexity. Are those calculations really so complicated? We could maybe also just make sure the compiler understands the concept of a function call that does not side-effect and can be dead code eliminated in case its value is not needed.

chrisseaton: Well the root assumption is that these methods will not PE - so I'll investigate if that's actually the case today.

Yes we could individually wrap the methods on `DateTime`, add the `@TruffleBoundary` annotation there, and then if we also had a `@SideEffectFree` annotation as well, that would solve it.

thomaswue: Yes, or maybe add an optional "canSideEffect" boolean on the TruffleBoundary annotation.

core/src/main/java/org/jruby/truffle/translator/BodyTranslator.java
```diff
@@ -2036,8 +2163,19 @@ public RubyNode visitPostExeNode(PostExeNode node) {
     public RubyNode visitRationalNode(RationalNode node) {
         final SourceSection sourceSection = translate(node.getPosition());
 
-        // TODO: implement Rational
-        return new FixnumLiteralNode.LongFixnumLiteralNode(context, sourceSection, node.getNumerator());
+        // Translate as Rubinius.privately { Rubinius.convert(a, b) }
```
eregon: Why privately?
And also why to use Rational.convert to create a Rational literal? Is this spec?

core/src/main/java/org/jruby/truffle/translator/BodyTranslator.java
```diff
@@ -2036,8 +2163,19 @@ public RubyNode visitPostExeNode(PostExeNode node) {
...
+
+        // TODO(CS): use IntFixnumLiteralNode where possible
+
+        final LexicalScope lexicalScope = environment.getLexicalScope();
+        final RubyNode moduleNode = new LexicalScopeNode(context, sourceSection, lexicalScope);
+        return new RubyCallNode(
+                context, sourceSection, "convert",
+                new ReadConstantNode(context, sourceSection, "Rational", moduleNode, lexicalScope),
```
eregon: Rubinius, no?

core/src/main/ruby/jruby/truffle/core/float.rb
```diff
@@ -8,8 +8,18 @@
...
 
-  NAN = 0.0 / 0.0
-  INFINITY = 1.0 / 0.0
-  EPSILON = 2.2204460492503131e-16
+  NAN        = 0.0 / 0.0
+  INFINITY   = 1.0 / 0.0
+  EPSILON    = 2.2204460492503131e-16
+  RADIX      = 2
+  ROUNDS     = 1
+  MIN        = 2.2250738585072014e-308
```
eregon: This might break a spec since we are considered as JRuby in RubySpec and the JVM seems to always consider subnormals.

spec/ruby/language/numbers_spec.rb
```diff
@@ -56,13 +56,13 @@
...
-    -042.should == Rational(-34, 1)
+    042r.should == Rational(34, 1)
+    -042r.should == Rational(-34, 1)
   end
 
   it "can be a binary literal with trailing 'r' to represent a Rational" do
-    0b1111.should == Rational(15, 1)
-    -0b1111.should == Rational(-15, 1)
+    0b1111r.should == Rational(15, 1)
+    -0b1111r.should == Rational(-15, 1)
```
eregon: Why is this in the diff? It was already there before, no?

core/src/main/java/org/jruby/truffle/translator/BodyTranslator.java
chrisseaton: I believe this is how Rubinius does it, but I actually couldn't find the code that does it. They've got a couple of layers of parser and translator and I can't find where it's done. Privately because `Rational#convert` is private - it's a Rubinius implementation detail.

core/src/main/java/org/jruby/truffle/translator/BodyTranslator.java
chrisseaton: It's a private method on `Rational`.

core/src/main/ruby/jruby/truffle/core/float.rb
chrisseaton: Ok, I'll keep my eye out for problems and we can set the constants in Java code if needs be. Let's not change it until we've seen a problem though. I think Ruby code should be the default unless there is a good reason otherwise.

spec/ruby/language/numbers_spec.rb
chrisseaton: No, this spec was apparently broken, and it happens to just pass anyway. I'd push it to RubySpec if it still existed!

eregon:
Looks good to me for compatibility!
For perf, we definitely need a few adjustments.
Can we now drop the `boolean rubiniusPrimitive` in RubyCallNode?
And in DispatchHeadNode?

core/src/main/java/org/jruby/truffle/nodes/rubinius/RubyTimeToDateTimeNode.java
eregon: Slight problem of only using `@offset` is you lose the zone name which is now preserved in Time, even across Marshal (but maybe Rubinius does not yet handle it).

chrisseaton:
Yeah, I'll remove those. If we're going to worry about performance we need a good benchmark first!

core/src/main/java/org/jruby/truffle/nodes/rubinius/TimePrimitiveNodes.java
eregon: In this case, I think the reasonable solution is to define Time#day as getDateTime().getDayOfMonth() (so make it a node).
Mostly we have a mismatch between the C API (decompose all, always) and the Java API (compute only one component). Let's not make an ORM :)

spec/ruby/language/numbers_spec.rb
eregon: It does, but pushing to the archive branch might be funny...

core/src/main/java/org/jruby/truffle/nodes/rubinius/TimePrimitiveNodes.java
eregon: Turns out RubyString is only used to keep the taint status, so should be easy to change.

PR #2422, opened 2015-01-04
Truffle-related clean up and simplifications.

thomaswue:
Please review, @chrisseaton. Thanks!


core/src/main/java/org/jruby/truffle/nodes/globals/WrapInThreadLocalNode.java
```diff
@@ -27,14 +27,14 @@ public WrapInThreadLocalNode(WrapInThreadLocalNode prev) {
...
+    public ThreadLocal<?> wrap(Object value) {
         return wrap(getContext(), value);
     }
 
-    public static ThreadLocal wrap(RubyContext context, Object value) {
+    public static ThreadLocal<Object> wrap(RubyContext context, Object value) {
         final RubyContext finalContext = context;
 
-        final ThreadLocal threadLocal = new ThreadLocal() {
+        final ThreadLocal<Object> threadLocal = new ThreadLocal<Object>() {
```
chrisseaton: Could use the diamond operator on the RHS of this: `new ThreadLocal<>()`

thomaswue: Was thinking the same, but the diamond operator is unfortunately not supported by javac for anonymous classes.

PR #2425, opened 2015-01-05
Further Truffle-related clean ups and simplifications.

thomaswue:
Please review, @chrisseaton @nirvdrum @eregon.


core/src/main/java/org/jruby/truffle/nodes/control/DoWhileNode.java
```diff
@@ -67,6 +65,7 @@ public boolean executeRepeating(VirtualFrame frame) {
                     break;
                 } catch (RedoException e) {
                     // Just continue in the while(true) loop.
+                    context.getSafepointManager().poll();
```
thomaswue: We do not need a safepoint if the block is executed only once. In general, doing the safepoint only *after* the first iteration reduces the number of safepoints compiled/executed.

core/src/main/java/org/jruby/truffle/nodes/literal/ArrayLiteralNode.java
```diff
@@ -108,19 +108,24 @@ public RubyArray executeArray(VirtualFrame frame) {
...
                     executedValues[n] = values[n].executeFloat(frame);
                 } catch (UnexpectedResultException e) {
-                    final Object[] executedObjects = new Object[n];
-
-                    for (int i = 0; i < n; i++) {
-                        executedObjects[i] = executedValues[i];
-                    }
-
-                    return makeGeneric(frame, executedObjects);
+                    return makeGeneric(frame, executedValues, n);
```
thomaswue: As a general style guide: We should try to move exceptional parts of methods in extra helpers. Makes them easier to read and improves startup/warmup time (HotSpot compilers can avoid inlining a method, but the base line compiler would never throw away code and the optimizing ones only when it is *never* executed).

core/src/main/java/org/jruby/truffle/runtime/core/RubyArray.java
```diff
@@ -244,8 +216,11 @@ public int getSize() {
     }
 
     public void setSize(int size) {
+        assert verifyStore(this.store, size);
```
thomaswue: This should also be a general style guide. Put long assertion code in extra helpers that return boolean "true" and are invoked themselves with "assert methodCall". Again improves readability but also startup/warmup.

core/src/main/java/org/jruby/truffle/runtime/util/ArrayUtils.java
```diff
@@ -17,6 +18,30 @@
...
+     * @param end the end index (exclusive), must be >= 0 and <= source.length and >= start
+     * @return a newly allocated array with the extracted elements and length (end - start)
+     */
+    public static Object[] extractRange(Object[] source, int start, int end) {
+        assert checkExtractRangeArgs(source, start, end);
+        int length = end - start;
+        Object[] result = new Object[length];
+        System.arraycopy(source, start, result, 0, length);
+        return result;
+    }
```
thomaswue: The method Arrays.copyOfRange is not suitable for partial evaluation. It contains an exception branch with String concatenation. While we usually expect this branch to be optimized away, we cannot guarantee that HotSpot profiling gives the expected results (because completely different code may execute and then "enable" the path). Additionally, the semantics of extractRange is more complicated than what we usually need - i.e., it will return an array of the same type as the source array, but we only need it to be an Object[] array. Overall, this is again a small change aimed at reducing startup/warmup times - the part about the exception branch can also have peak performance impacts.

core/src/main/java/org/jruby/truffle/runtime/core/RubyArray.java
nirvdrum: Do you have a pointer to something that would better explain how this change improves startup/warmup?

chrisseaton:
Perhaps we want an `ArrayUtils` class in `com.oracle.truffle.api.util`, that are all amenable to PE. Even if some of the cases like `System.arraycopy` work fine anyway, it would mean that people don't have to figure out if they're safe or not for every array utility method in the JDK. Maybe the same for `BigInteger` - a set of helper methods that add `@TruffleBoundary` on those that need them, but not those that don't.

core/src/main/java/org/jruby/truffle/nodes/control/DoWhileNode.java
eregon: But the RedoException branch is only taken when using the `redo`, keyword, no?
And that is quite rare, so there is no more safepoint in a typical loop with this code if I see right (or is it done with instrumentation by the LoopNode?).

chrisseaton: Yes `redo` is unusual - most loops will run this inner while loop just once per actual loop iteration. I think we need to poll the safepoint before the return as well. This duplication of safepoints might be why I put it at the head in the first place.

core/src/main/java/org/jruby/truffle/runtime/core/RubyArray.java
thomaswue: There are the following effects:
1) The interpreter will execute less # of bytecodes. The check of the assert boolean is not stripped by javac. So the interpreter will just go ahead and check the assert boolean all over.
2) The inlining heuristic of C1 relies on # bytecodes in a method. C1 will think this is a huge method while large parts of it immediately disappear when the asserts are removed by canonicalization.
3) C1 and also Graal will first parse the bytecodes, convert them into their IR, and then see that large parts of the IR is dead code. This means unnecessarily long compilation times.

The effect is not really large, but it adds up. For the asserts, one option is to strip them with a static tool. For general slow path code, this is not possible to do automatically.

eregon:
Can we merge this?

PR #2430, opened 2015-01-06
[Truffle] jt tool.

chrisseaton:
Would this be useful to people not as used to the code base as we are?

You should set `function jt { ruby tool/jt.rb $@; }` then run `jt help`.

@thomaswue @nirvdrum @eregon


nirvdrum:
I think as we get further along and have more people trying out Truffle it could be useful.

thomaswue:
Yes, this looks great! I can later test this and write down some documentation.

eregon:
@chrisseaton I see you are becoming quite a Rubyist now :)
Looks good to me!

PR #2443, opened 2015-01-09
Follow MRI implementation for RubyArray hash

### LINKED ISSUE #2437: Array hashing is in Ruby 1.8 mode
Different arrays, are hashing very similarly

```
>> [].hash
=> 0
>> [2].hash
=> 0
```

MRI's behaviour is different:

```
>> [].hash
=> 3252255281033687896
>> [2].hash
=> 3208161167426848866
```

Who828:
This fixes #2437 and follows MRI approach for RubyArray hash method,  however for consistency across the codebase I have made use of SipHashInline and PerlHash unlike MRI which makes use of MurMurHash.

### DIFF
--- a/core/src/main/java/org/jruby/RubyArray.java
+++ b/core/src/main/java/org/jruby/RubyArray.java
@@ -666,13 +666,18 @@ public class RubyArray extends RubyObject implements List, RandomAccess {
             public IRubyObject call(IRubyObject obj, boolean recur) {
                 int begin = RubyArray.this.begin;
                 long h = realLength;
+                long n;
                 if (recur) {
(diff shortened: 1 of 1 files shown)


headius:
Seems good to use the same hashing algorithm as for Strings. Let's go with it.

headius:
Bleh, sorry, I botched this review. I had to revert.

Don't use toString to get bytes for the string. I'm almost positive MRI does not convert objects to String before including them in hash calculation. I believe it just uses their natural hash() result, which may be why they don't use murmurhash (murmur is for hashing bytes).

Wanna take another shot at it? Use the MRI code for reference, and if it doesn't make sense to use murmur/sip/perlhash because we don't have bytes[], then just do what MRI does.

PR #2451, opened 2015-01-12
No need to synchronise a newly created local RubyArray

### LINKED ISSUE #2446: ConcurrentModificationException from ThreadGroup.list
while the `rubyThreadList` field is a fully synchronized collection, iterator might still throw a CME

https://github.com/jruby/jruby/blob/jruby-1_7/core/src/main/java/org/jruby/RubyThreadGroup.java#L149

there's probably a typo the line above as it synchronizes on a local variable : 
```java
 RubyArray ary = RubyArray.newArray(getRuntime());
  synchronized (ary) {
```

```
WeakHashMap.java:882:in `nextEntry': java.util.ConcurrentModificationException
	from WeakHashMap.java:915:in `next'
	from RubyThreadGroup.java:149:in `list'
	from RubyThreadGroup$INVOKER$i$0$0$list.gen:-1:in `call'
	from JavaMethod.java:280:in `call'
	from CachingCallSite.java:306:in `cacheAndCall'
	from CachingCallSite.java:136:in `call'
	from CallNoArgNode.java:60:in `interpret'
	from CallNoArgBlockPassNode.java:53:in `interpret'
	from NewlineNode.java:105:in `interpret'
	from BlockNode.java:71:in `interpret'
	from ASTInterpreter.java:112:in `INTERPRET_BLOCK'
	from Interpreted19Block.java:206:in `evalBlockBody'
	from Interpreted19Block.java:157:in `yield'
	from Block.java:142:in `yield'
	from Mutex.java:149:in `synchronize'
	from Mutex$INVOKER$i$0$0$synchronize.gen:-1:in `call'

```

only seen this under high actor/fiber usage (Celluloid) ... reproduced on **1.7.16**

Who828:
Possibly fixes #2446, we don't need synchronise for a locally created RubyArray .

### DIFF
--- a/core/src/main/java/org/jruby/RubyThreadGroup.java
+++ b/core/src/main/java/org/jruby/RubyThreadGroup.java
@@ -145,13 +145,14 @@ public class RubyThreadGroup extends RubyObject {
     @JRubyMethod
     public IRubyObject list(Block block) {
         RubyArray ary = RubyArray.newArray(getRuntime());
-        for (RubyThread thread : rubyThreadList) {
-            if (thread != null) {
-                ary.append(thread);
+        synchronized (rubyThreadList) {
+            for (RubyThread thread : rubyThreadList) {
+                if (thread != null) {
+                    ary.append(thread);
+                }
             }
         }
         return ary;
-
     }
 
     /**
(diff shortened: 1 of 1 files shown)


headius:
This is fine to remove but we also need to add correct synchronization, either against the thread list or against the whole ThreadGroup.

headius:
Or rather, to fix #2446 we need the appropriate synchronization. This PR alone does correctly remove useless synchronization on the Array :-)

Who828:
Oops! I did a little bit of digging and I found out that synchronizedSet must be manually synchronised while iterating http://docs.oracle.com/javase/7/docs/api/java/util/Collections.html#synchronizedSet(java.util.Set), so I have made the change.

headius:
Looks good to me!

headius:
Also applied to 1.7 in d99bfd9.

PR #2453, opened 2015-01-12
Made changes to RubyArray hash to not use PerlHash/SipHash

### LINKED ISSUE #2437: Array hashing is in Ruby 1.8 mode

Who828:
Fixes #2437, this time I have used RubyHash.hashValue with identityHash as salt to replicate what MRI does, hopefully hashValue is similar to MRI's rb_hash function (https://github.com/ruby/ruby/blob/7e8795e3935c9cbf4fe9ed24872a49d31a3a33cf/hash.c#L110).

Let me know if this works.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -13,7 +13,7 @@ before_script:
 
 jdk:
   - openjdk7
-  - oraclejdk8
+#  - oraclejdk8
 
 os:
   - linux
--- a/core/src/main/java/org/jruby/RubyArray.java
+++ b/core/src/main/java/org/jruby/RubyArray.java
@@ -58,10 +58,8 @@ import org.jruby.runtime.marshal.MarshalStream;
 import org.jruby.runtime.marshal.UnmarshalStream;
 import org.jruby.util.ByteList;
 import org.jruby.util.Pack;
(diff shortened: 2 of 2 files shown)


kares:
current RubyArray's hashing implementation looks like : 
```java
    @JRubyMethod(name = "hash")
    public RubyFixnum hash(ThreadContext context) {

        long h = Helpers.hashStart(runtime, realLength);

        h = Helpers.murmurCombine(h, System.identityHashCode(RubyArray.class));

        for (int i = 0; i < realLength; i++) {
            IRubyObject value = eltOk(i);
            RubyFixnum n = Helpers.safeHash(context, value);
            h = murmurCombine(h, n.getLongValue());
        }

        h = hashEnd(h);

        return runtime.newFixnum(h);
    }
```
... so I wonder is this PR still relevant/useful to consider updating?

headius:
@kares I think you're right.

Issue #2376, opened 2014-12-29
Enumerator#first when yielded arrays of 1 element bug.


headius:
From https://jira.codehaus.org/browse/JRUBY-7108

```ruby
Enumerator.new{|y| y << [42]}.first # => 42, should be [42]
```

headius:
Fixed by #2458.

PR #2458, opened 2015-01-13
Fixed yielded arrays of 1 element bug when using Enumerator#first

Who828:
Fixes an old issue #2376 with RubyYielder class.

This fix passes Rubyspecs and MRI specs, however let me know if there any changes required.


headius:
Good to get this fixed, but I hope we can figure out a lighter way to do it in the future. If I understand the logic, we're basically double-wrapping the value so that when unwrapped by yield logic it comes out unharmed. Will merge today.

nightscape:
Ah, ok, so this is the corresponding PR. Sorry, didn't find it at first (hint: you can search for a SHA in Github and it will find the corresponding PR).

Issue #2466, opened 2015-01-16
Package Truffle stdlib files into tarball


nirvdrum:
Since Truffle has a long ways to go before being widely useful, we've whitelisted the standard libraries we think are okay to load.  These are implemented by way of stubs in lib/ruby/truffle/mri and they basically do a require_relative back to the standard lib file in JRuby.  The problem is these stubs aren't currently being packaged so anyone using the tarballs as the source for their installation can't use even the whitelisted libs.

@mkristian -- Any help you can provide with packaging would be much appreciated.

mkristian:
to add those files to jruby-stdlib jar is easy and with it goes to lla the
other distribution forms.

BUT is it really the way to go: pack those "all" included ?

as far as I understand truffle is optional, so if I do not want to use
truffle I want "truffle" free artifacts. using ruby scripting in a java
even if it is only "puts 'hello world'" means I need to add 20MB or more
extra libraries to get it working.

rather I would like specifiy a classifier "truffle" to get the truffle and
NOT get the stuff truffle replaces (as far I understand this that would be
IR). unfortunately maven does pick only another jar with the classifier and
not another pom.xml (which different dependencies). but it could be done in
a similar manner as 'jruby-noasm' which collects the jruby-core with the
right classifier.

I am not talking about jruby-dist which is anyways "all inclusive" ;)

so jruby should become more modular in the way it gets distributed.

nirvdrum:
dist is the problem.  ruby-build uses the nightly tarballs (and the versioned ones upon release) for installation.  They currently don't include everything needed to run Truffle-backed JRuby.

chrisseaton:
We're working on some ideas to reduce the size of the Truffle code (it's generated code, so we're improving the generator), so it may not be quite as bad as 20MB. I'll let you know as we experiment.

As you say, jruby-dist (the binary tarballs, right?) and JRuby complete I think should include all backends, and I think that's always the way it's been understood to work. Truffle's a runtime configuration option, not an entirely separate product.

We're happy to work with you on some finer grained packaging as well. Personally, I don't see the issue as it's only a few MB and we don't have separate packages for indy and non-indy, but if you know of use cases I'll trust your judgement on that. Let us know if you need us to help. It's already possible to remove Truffle by brute force deleting the classfiles from the jar and users will get a nice error message.

Do we need to do anything to help you get our stdlib into the tarballs and complete?

mkristian:
[...]

PR #2476, opened 2015-01-17
truffle modular

mkristian:
this is a follow up of #2466 to propose a slightly different setup. it basically introduce a jruby-trufle artifact which can be easily exceluded using the jruby amven artifact. jruby-jars gem takes advantage of it as well, i.e. jruby-core jar there is already truffle free.


headius:
+1 looks good to me. I'd say we also remove it from jruby-complete, since anyone using jruby-complete can just add jruby-truffle to deps and it will be there.

headius:
This can merge now.

mkristian:
yes. beside this some cleanup is needed - readline, ripper, back to
SNAPSHOT version, etc

but this can go in as is.

headius:
I updated the patch and cleaned up some dependencies and force pushed to the PR branch.

We need to fix -X+T at command line before we merge this, so truffle remains functional at command line. We can do that either in the launcher (requires jruby-launcher work in C/++) or shade it into lib/jruby.jar.

mkristian:
shade it the truffle into lib/jruby.jar

mkristian:
merged manually

Issue #1571, opened 2014-03-22
Enumerator#feed


avit:
Documented here in MRI:

http://www.ruby-doc.org/core-2.1.1/Enumerator.html#method-i-feed

Currently raises a NoMethodError, tested in JRuby 1.7.9.

MSNexploder:
This works as expected in JRuby 9.x.x.x

PR #2486, opened 2015-01-20
Implemented Enumerator#feed method

Who828:
This PR implements Enumerator#feed and addresses #1571.  However the behaviour is not same as MRI because of the difference in implementation of next in JRuby and MRI.

Here is an example to demonstrate the difference,

**MRI**

```Ruby
a = [1, 2, 3]
m = a.to_enum(:map!)

m.next # a -> [1, 2, 3], returns 1
m.next # a -> [nil, 2, 3], returns 2
m.next # a -> [nil, nil, 3], returns 3
m.next # a -> [nil, nil, nil],  throws StopIteration exception and calls the yield/method

```

**JRuby**

```Ruby
a = [1, 2, 3]
m = a.to_enum(:map!)

m.next # a -> [nil, 2, 3], returns 1
m.next # a -> [nil, nil, 3], returns 2
m.next # a -> [nil, nil, nil], returns 3
m.next # a -> [nil, nil, nil],  throws StopIteration exception and doesn't call yield/method

```

So basically when you call next first time in MRI, it actually initialises the block but doesn't call it yet unlike JRuby. 

So because of this I haven't included test_feed specs in TestEnumerable (test:mri), as they fail because of the different order of evaluation of yield/method.

Though this PR does implement feed correctly for JRuby's next implementation, so I am raising it.

### DIFF
--- a/core/src/main/java/org/jruby/RubyEnumerator.java
+++ b/core/src/main/java/org/jruby/RubyEnumerator.java
@@ -490,7 +490,7 @@ public class RubyEnumerator extends RubyObject {
     @JRubyMethod(name = "next_values")
     public synchronized IRubyObject nextValues(ThreadContext context) {
         ensureNexter(context);
-
+        if (!feedValue.isNil()) feedValue = context.runtime.getNil();
         return RubyArray.newArray(context.runtime, nexter.next());
     }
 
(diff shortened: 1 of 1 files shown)


headius:
A few improvements you should do:

* If you must call getRuntime, cache the result. Don't keep calling it everywhere you need a runtime.
* Better: get it from ThreadContext.runtime (and cache it in a local for that too).
* nil is accessible more cheaply as context.nil

enebo:
And I said this in other PR about no curlies on single line ifs.  So since there are other things to change do that as well.

Who828:
Made the changes based on the feedback.

Issue #1719, opened 2014-05-29
Cannot require 'openssl' in a signed jruby jar as of jruby 1.7.3


jeffgran:
I cannot sign the jruby-complete jar and run it any more, after upgrading to the latest version. I "binary searched" to find where the problem first starts happening, and it's the upgrade to jruby 1.7.3 from 1.7.2 that breaks this.

The reason I need to sign the jar is because I am deploying my application via a JNLP (Java Web Start), and java security requirements dictate that I have to sign all of the jars. I have tried unjaring, signing everything (including the jopenssl.jar, the bouncycastle jars, etc) re-creating the jruby-complete jar and then signing it too, but that doesn't work either.

I very well may just be "doing it wrong" but can not figure out the right way. 

Under jruby 1.7.2:

    $ java -jar jruby-complete-1.7.2.jar -e 'require "openssl"; puts OpenSSL.inspect'
    OpenSSL
    $ cp jruby-complete-1.7.2.jar jruby-complete-1.7.2-signed.jar
    $ jarsigner -keystore my-keystore jruby-complete-1.7.2-signed.jar my-alias
    $ java -jar jruby-complete-1.7.2-signed.jar -e 'require "openssl"; puts OpenSSL.inspect'
    OpenSSL

Starting with 1.7.3 all the way up to 1.7.12:

    $ java -jar jruby-complete-1.7.12.jar -e 'require "openssl"; puts OpenSSL.inspect'
    OpenSSL
    $ cp jruby-complete-1.7.12.jar jruby-complete-1.7.12-signed.jar
    $ jarsigner -keystore my-keystore jruby-complete-1.7.12-signed.jar my-alias
    $ java -jar jruby-complete-1.7.12-signed.jar -e 'require "openssl"; puts OpenSSL.inspect'

java.lang.SecurityException: class "org.jruby.ext.openssl.OpenSSLReal$OpenSSLModule$INVOKER$s$0$0$errors"'s signer information does not match signer information of other classes in the same package
	at java.lang.ClassLoader.checkCerts(ClassLoader.java:952)
	at java.lang.ClassLoader.preDefineClass(ClassLoader.java:666)
	at java.lang.ClassLoader.defineClass(ClassLoader.java:794)
	at org.jruby.util.JRubyClassLoader.defineClass(JRubyClassLoader.java:118)
	at org.jruby.internal.runtime.methods.InvocationMethodFactory.endClass(InvocationMethodFactory.java:1455)
[...]

PR #2501, opened 2015-01-22
upgrade to jruby-openssl 0.9.6

jeffgran:
WRT this openssl issue: https://github.com/jruby/jruby/issues/1719

### DIFF
--- a/lib/pom.rb
+++ b/lib/pom.rb
@@ -29,7 +29,7 @@ KRYPT_VERSION = '0.0.2'
 # the versions are declared in ../pom.xml
 default_gems =
   [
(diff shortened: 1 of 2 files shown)


mkristian:
there is problem when people try to downgrade: https://github.com/jruby/jruby/issues/2502#issuecomment-71169983

krypt is gone BUT was never a declared depenency in jruby-openssl so any downgrade needs to install krypt as well to work (1.9 + 2.0 mode only).

mkristian:
the other branch is more complete, in respect that it is not needed to bundle the krypt gem anymore and some test fixes.

@jeffgran guess will merge soon - thanx for bringing it to surface

PR #2504, opened 2015-01-23
Fix ascii-only detection of symbol names.

### LINKED ISSUE #1328: Encoding of symbol literals does not respect the encoding of the source file
Hello.  JRuby does not seem to respect encoding marking at the top of source files when deciding the encoding of symbol literals.  In a file marked as UTF-8, all symbols get marked as US-ASCII even if they contain special characters.  Here is the code that reproduces the issue:

```ruby
# coding: UTF-8
puts :µ.encoding  # jruby says US-ASCII, MRI says UTF-8
puts :a.encoding  # jruby and MRI both say US-ASCII
```

It is dangerous to have an symbol like µ with its encoding set to US-ASCII because there are many common things you might do, like calling `inspect`, that result in "ArgumentError: invalid byte sequence in US-ASCII".

One workaround is to use a string literal followed by `.to_sym`.

Here is the output of my `jruby -v`:

```
jruby 1.7.9 (1.9.3p392) 2013-12-06 87b108a on Java HotSpot(TM) 64-Bit Server VM
1.7.0_07-b10 [Windows 8-amd64]
```

The MRI I compared this to is: "ruby 2.0.0p0 (2013-02-24) [x64-mingw32]".

Sorry if this is a duplicate; I did check every closed or open github issue tagged with "encoding" trying to avoid that.

cheald:
If you run the issue described in #1328 with `-J-Dfile.encoding=CP1252`
then the string length comparison fails. This smoothes over the issue,
but it is still possible that there are problems with bytes being read
with the JVM's file.encoding rather than the encoding specified in the
magic comment.

Closes #1328

### DIFF
--- a/core/src/main/java/org/jruby/ast/SymbolNode.java
+++ b/core/src/main/java/org/jruby/ast/SymbolNode.java
@@ -32,7 +32,6 @@
  ***** END LICENSE BLOCK *****/
 package org.jruby.ast;
 
-import java.awt.image.ByteLookupTable;
 import java.util.List;
 
 import org.jcodings.Encoding;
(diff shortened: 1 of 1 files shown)


enebo:
I would apply this since it would clear up the specific problem reported in #1328, but I am worried if I do then we will wallpaper over the real issue which is getting wrong encoding bytes and feeding them to UTF8 encoding method.  So leaving PR open for the time being until I can look at #1328.  Once we do find the base case we can apply this since you said this was normative way MRI checks.

PR #2508, opened 2015-01-23
upgrade to jruby-openssl 0.9.6

mkristian:
this is the old branch we used for testing 0.9.6 against jruby-1_7

will continue https://github.com/jruby/jruby/pull/2501 here.


mkristian:
merged manually

PR #2520, opened 2015-01-26
Move methods from BasicObject to Object

sferik:
In the process of debugging an [unrelated issue](https://github.com/avdi/naught/issues/65#issuecomment-71541149), I came across an inconsistency between JRuby and CRuby: the `#itself` and `#object_id` methods are defined on `BasicObject` instead of `Object`.

# WARNING #

i-have-no-idea-what-im-doing-5

I am not an expert :coffee: programmer nor am I particularly familiar with the JRuby codebase, however I’ve made an attempt to fix this issue myself, as it appears to be mostly a :scissors: and :spaghetti: job.

I’ve also attempted to add a test to prevent this behavior from regressing, however I may have put it in the wrong place, as JRuby appears to have multiple test suites.

I’ve also taken the liberty to remove the `Object#id` method, which was deprecated when Ruby 1.9 was released and removed from CRuby in version 1.9.3.

### DIFF
--- a/core/src/main/java/org/jruby/RubyBasicObject.java
+++ b/core/src/main/java/org/jruby/RubyBasicObject.java
@@ -1024,15 +1024,6 @@ public class RubyBasicObject implements Cloneable, IRubyObject, Serializable, Co
         return getRuntime().newFixnum(getObjectId());
     }
 
-    /** rb_obj_itself
-     *
-     * Identity method for the object.
-     */
-    @JRubyMethod
-    public IRubyObject itself() {
-        return this;
-    }
-
     /**
      * The logic here is to use the special objectId accessor slot from the
      * parent as a lazy store for an object ID. IDs are generated atomically,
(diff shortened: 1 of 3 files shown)


chrisseaton:
@enebo should this be cherry-picked onto 17?

sferik:
@chrisseaton @enebo The only commit to consider cherry-picking into the 1.7 branch would be https://github.com/jruby/jruby/commit/519fd7540b690c8de03b315f6d65ad407aaa4220, as https://github.com/jruby/jruby/commit/0ce0215d22c022869ef191341213e17e77018a17 is not backwards-compatible and https://github.com/jruby/jruby/commit/c2184b5cdbbb59c7f2a59af88f994e3fa76ea74f `Object#itself` was only added in Ruby 2.2.0.

enebo:
So yeah I think 519fd75 should be and I even think 0ce0215 should for --1.9 if it really was removed for 1.9.3 since that is what version -1.9 supports.  @sferik 'id' was removed during 1.9.3 or after?

sferik:
@enebo `Object#id` was removed before 1.9.3?

sferik:
Sorry, that sentence should end in a period, not a question mark.

PR #2522, opened 2015-01-27
Implement pack & unpacking ipv6 address in sockaddr

GennadySpb:


enebo:
@GennadySpb Should this also work in 1.9.3 or 1.8.7?  Should we be backporting this to JRuby 1.7.x?

GennadySpb:
@enebo Right now I have no idea. Need to patch, compile & test it.

PR #2526, opened 2015-01-27
Fixed a bug with local_variables ordering and duplicate variables

### LINKED ISSUE #2129: Block variable shadowing and local_variables not matching up right
Here's the cases from MRI:

```ruby
  def test_shadowing_local_variables
    bug9486 = '[ruby-core:60501] [Bug #9486]'
    x = tap {|x| break local_variables}
    assert_equal([:x, :bug9486], x)
  end

  def test_shadowing_block_local_variables
    bug9486 = '[ruby-core:60501] [Bug #9486]'
    x = tap {|;x| break local_variables}
    assert_equal([:x, :bug9486], x)
  end
```

And our results:

```
123) Failure:
TestVariable#test_shadowing_block_local_variables [/Users/headius/projects/jruby/test/mri/ruby/test_variable.rb:95]:
<[:x, :bug9486]> expected but was
<[:bug9486, :x, :x]>.

[5903/5997] TestVariable#test_shadowing_local_variables = 0.00 s              
124) Failure:
TestVariable#test_shadowing_local_variables [/Users/headius/projects/jruby/test/mri/ruby/test_variable.rb:89]:
<[:x, :bug9486]> expected but was
<[:bug9486, :x, :x]>.
```

Combination of parser and IR I would guess, but I have not investigated.

Who828:
Fixes #2129 by removing the duplicate variables and matching the order across different scopes to MRI.

### DIFF
--- a/core/src/main/java/org/jruby/RubyKernel.java
+++ b/core/src/main/java/org/jruby/RubyKernel.java
@@ -60,6 +60,7 @@ import org.jruby.runtime.CallType;
 import org.jruby.runtime.Helpers;
 import org.jruby.runtime.ThreadContext;
 import org.jruby.runtime.Visibility;
+import org.jruby.runtime.DynamicScope;
 import org.jruby.runtime.backtrace.RubyStackTraceElement;
 import org.jruby.runtime.builtin.IRubyObject;
 import org.jruby.runtime.load.IAutoloadMethod;
@@ -76,6 +77,7 @@ import java.io.ByteArrayOutputStream;
 import java.io.File;
 import java.util.ArrayList;
 import java.util.Arrays;
+import java.util.HashSet;
 import java.util.Map;
 
 import static org.jruby.RubyEnumerator.enumeratorizeWithSize;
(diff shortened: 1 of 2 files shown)


enebo:
@Who828 Do you know if this is also a problem in earlier versions of Ruby?

headius:
I filed the original bug, and I did not test anything but 9k. The bug was filed to capture an new spec failure when we last updated.

In 1.7.18, it does look broken:

```
$ rvm jruby-1.7.18 do ruby -e "a = 1; b = 2; 1.times {|;b| p local_variables}"
[:a, :b, :b]
```

headius:
@enebo So I'm not sure what to do here. This is a bug, MRI acknowledged it as a bug, but it was unfixed until 2.2. If we fix this on 1.7 branch, our local_variables behavior would not match 1.9.3.

enebo:
Then I think we leave it broken.  That tells me no one likely has depended on this not being unique.  I bet people who do this and care menually uniq the elements already.  They definitely would if they want it to keep working between versions.

PR #2530, opened 2015-01-28
Forgot to match the order of JRuby local variables spec to MRI

Who828:
Opening a new PR since last one was merged, but this fixes the order to match MRI.


headius:
Thank you :-)

PR #2546, opened 2015-02-01
Re-enblig memory fences

pitr-ch:
It was breaking `StampedVariableAccessor`, resulting in
not sinchronized access to instance variables
when `StampedVariableAccessor` was used.


headius:
Thanks, @pitr-ch!

headius:
Very glad you found this before we shipped 9k final :-)

pitr-ch:
You are welcome, but it was pure luck anyway :)

PR #2560, opened 2015-02-04
Fix uninitialized usage of File::Stat

lumeet:
* Fix File::Stat#struct when uninitialized
* Make other methods fail nicely unless initialized

The behavior is now the same as in MRI 2.2.


enebo:
@lumeet Is this uninitialized behavior possibly also a bug in our 1.8.7 and 1.9.3 support?  If so then this could be done for our jruby-1_7 branch as well.

lumeet:
@enebo Yes, I submitted a PR for 1.7 too: #2572.

PR #2568, opened 2015-02-05
Exit iterator when `RubyHash#any?` short-circuits

grddev:
Previously, if `any?` returned true early (or threw an exception) the hash would be marked as still iterating, and thus prevent updates.

Expressed in Ruby code, the following results in "can't add a new key into hash during iteration" errors:

```ruby
h = {1 => 2}; h.any? { true }; h[2] = 1
h = {1 => 2}; h.any? { raise } rescue nil; h[2] = 1
```

I would have liked to put them into tests, but wasn't completely sure were would be most appropriate.

### DIFF
--- a/core/src/main/java/org/jruby/RubyHash.java
+++ b/core/src/main/java/org/jruby/RubyHash.java
@@ -1892,25 +1892,29 @@ public class RubyHash extends RubyObject implements Map {
 
     private IRubyObject any_p_i(ThreadContext context, Block block) {
         iteratorEntry();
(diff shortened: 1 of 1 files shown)


enebo:
@grddev Could this be an issue for our 1.7 branch as well? for 1.8.7 or 1.9.3 support?

grddev:
As far as I can see the `Hash#any?` implementation was added in 2a2806e, which seems to exist in the master branch only.

enebo:
@grddev Great thanks for checking!  I also looked and I think this was added specifically to Hash and used to just fall back to Enumerable impl.

PR #2571, opened 2015-02-06
Truffle arrays

chrisseaton:
@nirvdrum @eregon think this is a sensible approach to starting to decompose the array nodes?


truffle/src/main/java/org/jruby/truffle/nodes/array/ArrayReadDenormalizedNode.java
```diff
@@ -0,0 +1,53 @@
...
+    public abstract Object executeRead(VirtualFrame frame, RubyArray array, int index);
+
+    @Specialization
+    public Object read(VirtualFrame frame, RubyArray array, int index) {
+        if (readNode == null) {
+            CompilerDirectives.transferToInterpreter();
+            readNode = insert(ArrayReadNormalizedNodeFactory.create(getContext(), getSourceSection(), null, null));
+        }
+
+        final int normalizedIndex = array.normaliseIndex(index);
```
nirvdrum: I'd pick either American or British English here.  Switching from 's' to 'z', vice versa, is something I don't want to have to think too hard about.

truffle/src/main/java/org/jruby/truffle/nodes/array/NewFixnumLowerNode.java
```diff
@@ -0,0 +1,55 @@
...
+import org.jruby.truffle.runtime.core.CoreLibrary;
+
+/**
+ * Passes through {@code int} values unmodified, but will convert a {@code long} value to an {@code int}, if it fits
+ * within the range of an {@code int}. Leaves all other values unmodified. Used where a specialisation only accepts
+ * {@code int}, such as Java array indexing, but we would like to also handle {@code long} if they also fit within an
+ * {@code int}.
+ */
+@NodeChild(value="child", type=RubyNode.class)
+public abstract class NewFixnumLowerNode extends RubyNode {
```
nirvdrum: Where is this used?  I don't see it wired up anywhere.

truffle/src/main/java/org/jruby/truffle/nodes/array/ArrayReadNormalizedNode.java
```diff
@@ -0,0 +1,84 @@
...
+    )
+    public Object readObjectInBounds(RubyArray array, int index) {
+        return ((Object[]) array.getStore())[index];
+    }
+
+    @Specialization(
+            guards="!isNormalisedInBounds"
+    )
+    public RubyNilClass readOutOfBounds(RubyArray array, int index) {
+        return getContext().getCoreLibrary().getNilObject();
```
nirvdrum: I'm confused by this case.  Negative values aren't considered in bounds, but shouldn't necessarily return nil.  I take it by going through PrimitiveArrayNodeFactory we shouldn't ever get a negative value here, but expecting the caller to always pass a positive value seems suspect.  Maybe an assert would be enough.

nirvdrum: Okay.  I miss that this was the normalized node.  I still think an assert might be good here.

truffle/src/main/java/org/jruby/truffle/nodes/core/ArrayGuards.java
```diff
@@ -73,4 +73,30 @@ public static boolean areBothObject(RubyArray a, RubyArray b) {
         return a.getStore() instanceof Object[] && b.getStore() instanceof Object[];
     }
 
+    // New names
```
nirvdrum: Are you going to remove the old guards then?

nirvdrum:
I hadn't ever looked at the ArrayIndexNode that this is replacing, but the changes look straightforward and make sense to me.  Good stuff.

truffle/src/main/java/org/jruby/truffle/nodes/core/ArrayGuards.java
```diff
@@ -73,4 +73,30 @@ public static boolean areBothObject(RubyArray a, RubyArray b) {
...
+
+    public static boolean isNullArray(RubyArray array) {
+        return array.getStore() == null;
+    }
+
+    public static boolean isIntegerArray(RubyArray array) {
```
eregon: What about isIntArray? So mapping from primitive type to guard name is just "is#{type.capitalize}Array".

truffle/src/main/java/org/jruby/truffle/nodes/core/ArrayNodes.java
```diff
@@ -596,96 +597,25 @@ public boolean equal(VirtualFrame frame, RubyArray a, Object b) {
...
-        public Object getFloat(RubyArray array, int index) {
-            int normalisedIndex = array.normaliseIndex(index);
-
-            if (normalisedIndex < 0 || normalisedIndex >= array.getSize()) {
-                return getContext().getCoreLibrary().getNilObject();
-            } else {
-                return ((double[]) array.getStore())[normalisedIndex];
+        @Specialization
+        public Object at(VirtualFrame frame, RubyArray array, int index) {
+            if (readNode == null) {
```
eregon: Since we are always going to use it I would create the node eagerly here.
Also see the assignment from prev above.

truffle/src/main/java/org/jruby/truffle/nodes/array/PrimitiveArrayNodeFactory.java
```diff
@@ -0,0 +1,32 @@
...
+     */
+    public static RubyNode read(RubyContext context, SourceSection sourceSection, RubyNode array, int index) {
+        final RubyNode literalIndex = new FixnumLiteralNode.IntegerFixnumLiteralNode(context, sourceSection, index);
+
+        if (index > 0) {
+            return ArrayReadNormalizedNodeFactory.create(context, sourceSection, array, literalIndex);
+        } else {
+            return ArrayReadDenormalizedNodeFactory.create(context, sourceSection, array, literalIndex);
+        }
+    }
```
eregon: This is very nice, much better than the duplicated nodes before.

eregon:
Looks good. I am not sure we need to differentiate normalized/denormalized (the terms confuses me a bit) as the branch profiling should do just fine but it might be worth to avoid negative indices where we don't expect them and express a raw access from a Ruby access.

I would add asserts as Kevin proposed to make it clearer what are the expectations.

What other set of primitives should we have?
I think right now getLength is not needed, but I think to these:
* get at index normalized / full (this PR)
* set at index
* slice from index (inc) to index (excl), but what should it return? Also, can excl be negative here? it can not always be pre-normalized, for instance see ArraySliceNode currently used for rest/post args. I would be in favor of not differentiate normalized/not here for the end index.

I think there is a possibility to map a slice with a range to (begin,len) or vice-versa (might be tricky for some weird cases). It would definitely help the duplication.

truffle/src/main/java/org/jruby/truffle/nodes/array/ArrayReadDenormalizedNode.java
chrisseaton: Yeah I should be using US English.

truffle/src/main/java/org/jruby/truffle/nodes/array/NewFixnumLowerNode.java
chrisseaton: Ah I put it in and then took it out again. It should go back in. It's a much simpler version of the current Fixnum lowering node.

truffle/src/main/java/org/jruby/truffle/nodes/core/ArrayGuards.java
chrisseaton: Yeah in time - they were designed before we could refer to individual arguments.

truffle/src/main/java/org/jruby/truffle/nodes/core/ArrayGuards.java
chrisseaton: Yeah that's a good idea.

truffle/src/main/java/org/jruby/truffle/nodes/core/ArrayNodes.java
chrisseaton: Ah but the node might never get executed!

chrisseaton:
Denormalized means might be a negative index. Normalized means after we have wrapped any negative indices to be an offset from the end. It's a bit more confusing than that, as wrapping might give us another negative index, but then it's out of bounds.

The reason to make them separate is that the denormalized node does the normalization in a single specialisation, and then hands off to the normalized node that has multiple specializations. Before, all the specializations had to do their own normalisation.

I'll merge this and keep experimenting with more array primitives. Getting all Array specs working is on my todo list, including with the new randomisation we have.

I'll add some more asserts.

truffle/src/main/java/org/jruby/truffle/nodes/core/ArrayNodes.java
eregon: Ah right, all nodes are instantiate by the CoreMethodNodeManager, so if the program does not use Array#at then we create the subnode for nothing!

chrisseaton: Maybe we should make CoreMethodNodeManager lazy in the first place? A single node that if executed replaces itself with the many nodes that form the body of a core method.

eregon: Yeah it could solve this but it's an extra replace for potentially all builtins (this would be quite nicer with `@Cache`).

PR #2579, opened 2015-02-08
Fix for issue 2182 on master : Struct#inspect with utf8 encode string member

k77ch7:
This commit fixes issue #2182 on master. Inspect method of Struct with utf8 string member should return a utf8 string. Please review, @enebo and @headius. Thanks!


enebo:
I think perhaps the jruby-1_7 merge brought most of this over and I had to manually make a change or two.  We generally do not let merges take so long, so sorry for confusion.

Issue #2512, opened 2015-01-24
keyword arguments incompatibilities


blanquer:
There are a couple of things that I'm seeing that still don't match the MRI behavior on method keyword arguments.

One is the arity, which returns different results ( but I believe that this might already be tracked in https://github.com/jruby/jruby/issues/2480 ?)

The other one is the definition of the keyword types "required" vs. not . For example, required arguments are not reported as such in the .parameters call.

Here's a quick snippet that shows both things, and the difference in results:

```ruby
class Foo
  def self.test( arg1: "default" , arg2: )
    puts "arg1: #{arg1.inspect}"
  end
end

m = Foo.method(:test)
puts "PARAMS: #{m.parameters.inspect}"
puts "ARITY: #{m.arity}"
```

```shell
bug:~ $ rvm 2.2.0
bug:~ $ ruby /tmp/test.rb
PARAMS: [[:keyreq, :arg2], [:key, :arg1]]
ARITY: 1
bug:~ $ rvm jruby-head
bug:~ $ ruby /tmp/test.rb
PARAMS: [[:key, :arg1], [:key, :arg2]]
ARITY: 0

```

Who828:
Hey, thanks for reporting the bug!
I have fixed parsing bit, so it identifies keyreq correctly but I still need to work on the arity part of the bug.

enebo:
Recent commits has solved remaining issue with this snippet (e.g. arity being reported as wrong).

PR #2584, opened 2015-02-10
Fixed arity issue with RubyMethod

### LINKED ISSUE #2480: Keyword arguments arity issue
Probably related to #2320. I get similar results with lambda keyword arguments and keyword arguments passed as a Hash instead of passing them explicitly.

```shell
$ jruby -v
jruby 9.0.0.0.pre1 (2.2.0p0) 2015-01-18 746583b Java HotSpot(TM) 64-Bit Server VM 25.25-b02 on 1.8.0_25-b17 +jit [linux-amd64]
```

```ruby
l = lambda  { |a:| a }
l.call(a: 1) # wrong number of arguments (1 for 0)
l.arity # 0 (1 in MRI 2.2.0p0)
```
After some debugging, I found out that in `RubyBlock#prepareArgs` the arity is checked on the `Signature`'s arity object (a "Fixed0"), but keyword args would only be considered in `Signature#checkArity`.

Not sure if this is related or a different issue:

```ruby
def m(a:, b:); a + b; end
args = { a: 1, b: 2 }
m(args) # ArgumentError: missing keyword: a
method(:m).arity # 0 (1 in MRI 2.2.0p0)

Who828:
Fixes #2480 and #2512 with RubyMethod by taking account of required kwargs.

### DIFF
--- a/core/src/main/java/org/jruby/ast/ArgsNode.java
+++ b/core/src/main/java/org/jruby/ast/ArgsNode.java
@@ -277,4 +277,27 @@ public class ArgsNode extends Node {
 
         return Node.createList(pre, optArgs, restArgNode, blockArgNode);
     }
+
+    public int getRequiredKeywordCount() {
+        if (hasRequiredKeywordArg()) return 1;
+        return 0;
+    }
+
+    private boolean hasRequiredKeywordArg() {
+        if (getKeywords() == null) return false;
+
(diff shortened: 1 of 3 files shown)


enebo:
@Who828 Looks like some rubyspecs failed from this fix...

Who828:
Darn, I missed in midst of OpenSSL NameError.
I have fixed it now, will open a new PR.

PR #2567, opened 2015-02-05
JRuby+Truffle: Keyword Arguments Optimization

fniephaus:
This is work in progress.

More details can be found here:
- https://github.com/jruby/jruby/issues/2541
- https://github.com/fniephaus/jruby/pull/1

Issue #2541, opened 2015-01-30
JRuby+Truffle: Keyword Arguments - "illegal recursive call"


fniephaus:
Hi folks,

If you run the following code with JRuby+Truffle:

```ruby
def m(i:)
    i
end

50000.times.each do
  (1..100_000).each do |i|
    m i:i
  end
end
```

you'll get something similar to this:

```
jruby: using Graal Truffle Runtime
[truffle] opt fail  block in each:./core/kabench.rb:6 <split-0-U> |Reason com.oracle.graal.nodes.util.GraphUtil$2: Found illegal recursive call to HotSpotMethod<Class.getCanonicalName()>, must annotate such calls with @TruffleBoundary! 
[truffle] opt fail  m:./core/kabench.rb:1 <split-0-U>             |Reason com.oracle.graal.nodes.util.GraphUtil$2: This code path should never be part of a compilation. [] 
[truffle] opt fail  Range#each(core):core: Range#each <split-0-U> |Reason com.oracle.graal.nodes.util.GraphUtil$2: This code path should never be part of a compilation. [] 
```

I assume that there's just a TruffleBoundary missing somewhere as suggested by the error message. But unfortunately, I'm unable to tell if that's really the case and if yes how to fix it.
Any ideas?

Regards,
Fabio

chrisseaton:
Our current implementation of keyword arguments doesn't compile - it only works in the interpreter. You're supposed to get a nice error message telling you this, due to the `notDesignedForCompilation` directives (e.g. https://github.com/jruby/jruby/blob/1095f6785e2b6f8f884b25d40e2074b52893667b/truffle/src/main/java/org/jruby/truffle/nodes/methods/arguments/ReadKeywordArgumentNode.java#L36) but for some reason the compiler here is failing before it gets a chance to report that error. That happens sometimes.

If you change `m` to a keyword-rest argument, as in `m(**i)`, you do get the nice compiler error.

The reason the keyword arguments don't compile is that we've taken shortcuts and used slow operations such as `verySlowToKeyValues` and Java's `equals` rather than setting up proper Ruby call sites yet.

fniephaus:
Closed, see #2567.

PR #2588, opened 2015-02-11
JRuby+Truffle: Keyword Arguments Optimization

fniephaus:
Hi @chrisseaton ,
this PR speeds up keyword arguments in JRuby+Truffle as discussed earlier (#2567).
- Argument nodes are moved from `RubyCallNode` to `DispatchNode`
- Keyword arguments are now being treated as normal named arguments
- The method argument node array is specific for receiver types
- In the arguments array, we use `MarkerNode` and `MarkerNode.Marker` to indicate that a method call with keyword arguments is optimized, i.e., the `HashLiteralNode` is expanded (removed and values extracted into the array)

Feedback and thorough review would be appreciated (and necessary).

Regards,
@matthias-springer, @jgraichen, @mswart and @fniephaus

==
More details can be found here:
- #2541, #2567
- https://github.com/fniephaus/jruby/pull/1


chrisseaton:
Thanks very much for this. I'll go through it today. @nirvdrum and @eregon can you also take a look? I would say straight away that you still have one failing language spec that needs to be looked at - it doesn't look too complicated though.

fniephaus:
We're working on it! Thanks for taking a look.

nirvdrum:
This is great news.  One general comment though, please replace all tab characters with 4 spaces.  We should try to keep the formatting consistent with the rest of the code.

fniephaus:
Thanks for the feedback, @nirvdrum. We'll fix the indent asap.
@matthias-springer and I will also work on the failing language spec tonight. We already know how to fix it.

truffle/src/main/java/org/jruby/truffle/nodes/RubyCallNode.java
```diff
@@ -34,21 +29,12 @@
 
     @Child private RubyNode receiver;
     @Child private ProcOrNullNode block;
-    @Children private final RubyNode[] arguments;
+    private final RubyNode[] arguments;
```
eregon: Why removing the `@Children` annotation here?

truffle/src/main/java/org/jruby/truffle/nodes/dispatch/CachedBoxedDispatchNode.java
```diff
@@ -54,7 +65,111 @@ public CachedBoxedDispatchNode(
...
+            if (argumentNodes.length > 0) {
+                hashNode = (HashLiteralNode) argumentNodes[argumentNodes.length - 1];
+            } else {
+                hashNode = HashLiteralNode.create(context, null,
+                        new RubyNode[0]);
+            }
+
+            List<String> restKeywordLabels = new ArrayList<String>();
+            for (int j = 0; j < hashNode.size(); j++) {
+               final String label = ((ObjectLiteralNode) hashNode.getKey(j)).execute(null).toString();
```
eregon: `ObjectLiteralNode.getObject()` should be used here instead of `execute(null)`.
Will the key always be an ObjectLiteralNode?

eregon:
Looks good, I had a quick look through the code.
What was the decision behind using a MarkerNode, compared to other alternatives?

matthias-springer:
We use `MarkerNode` to indicate that a call site is optimized, i.e., the Hash is expanded and keyword arguments can be read directly from the arguments array instead of the last argument which is a Hash. Now that I though about it again, I think it may be better to add another (boolean) flag to the beginning of the arguments array.

We removed the `@Children` annotation because we moved the argument nodes to the dispatch node. We should remove the filed from `RubyCallNode` entirely.

truffle/src/main/java/org/jruby/truffle/nodes/methods/MarkerNode.java
```diff
@@ -0,0 +1,22 @@
...
+    public MarkerNode(RubyContext context, SourceSection sourceSection) {
+        super(context, sourceSection);
+    }
+
+    @Override
+    public Object execute(VirtualFrame frame) {
+        return new Marker();
+    }
+
+    public static class Marker {}
```
eregon: What are the pros/cons of using the singleton pattern here?

eregon:
I see, having the (second) marker at the end seems natural as it would be where the Hash is in the non-optimized case. I don't know what is best in case of rest arguments.

truffle/src/main/java/org/jruby/truffle/nodes/dispatch/CachedBoxedDispatchNode.java
eregon:
```ruby
def m(k:)
end
m("a" => 45)
```
will fail for instance.

fniephaus:
@chrisseaton All tests are passing now. Feel free to give more feedback

truffle/src/main/java/org/jruby/truffle/nodes/dispatch/DispatchNode.java
```diff
@@ -136,5 +158,82 @@ public final Object execute(VirtualFrame frame) {
...
+        } else if (store instanceof Object[]) {
+            seenObjectInUnsplat = true;
+            return ArrayUtils.extractRange((Object[]) store, 0, size);
+        }
+
+        throw new UnsupportedOperationException();
+    }
+    
+    @ExplodeLoop
+    protected Object executeArguments(VirtualFrame frame, Object argumentOverride) {
```
matthias-springer: I am wondering whether there's a better way to implement this. Argument nodes for calls are now stored in `DispatchNode`, but there are still places where argument objects (no nodes, but objects) are passed when calling the dispatch head node (e.g. `ArrayNodes.inject`). That's why we are keeping this option (not storing the argument nodes in `DispatchNode` but passing objects as before) still open.

truffle/src/main/java/org/jruby/truffle/runtime/methods/SharedMethodInfo.java
```diff
@@ -35,6 +48,59 @@ public SharedMethodInfo(SourceSection sourceSection, LexicalScope lexicalScope,
...
+                        throw new UnsupportedOperationException(
+                                "unsupported keyword arg " + node);
+                    }
+                }
+            }
+        } else {
+            keywordArguments = null;
+        }
+        
+        this.arity = getArity(argsNode);
```
matthias-springer: Also not sure whether this is the correct way to do it: we added arity here so that we can access it in the dispatch node (get number of required args etc. to determine whether we should optimize/expand a keywords hash. If the last argument is a Hash but we are short on named arguments, we have to treat the Hash as a regular last argument instead of using it for keyword args. That case is not optimized.)

chrisseaton: Arity here is good.

truffle/src/main/java/org/jruby/truffle/nodes/dispatch/CachedUnboxedDispatchNode.java
```diff
@@ -91,6 +97,9 @@ public Object executeDispatch(
 
         switch (getDispatchAction()) {
             case CALL_METHOD: {
+            	argumentsObjects = executeArguments(frame, argumentsObjects);
+            	blockObject = executeBlock(frame, blockObject);
```
chrisseaton: Tabs here, for example.

chrisseaton:
Still a few tabs, but I'll merge it now so we can get on and run our CI benchmarks. Can you go back and fix the tabs and submit a new PR for that?

fniephaus:
@chrisseaton sure, will do. Thanks!

chrisseaton:
I'm afraid I had to revert these merges temporarily. We've got some regressions on our benchmarks after pulling your branch in, and we don't want to leave any regressions in if we can't fix them in a day as it may mask other regressions.

I've made a branch `truffle-kwargs` with your changes on them, and we'll try to fix the regressions there. Hopefully it's something minor as it looked like your work shouldn't impact if keyword arguments aren't being used. I'd take an initial look and will let you know what the problem might be, but this might not be until Tuesday.

fniephaus:
Hey @chrisseaton,
Do you have any news on this?

chrisseaton:
I'm going to wait a few days, as we are changing how the PE works in Truffle and probably doing a new release of Graal. It makes sense to wait for those before exploring the problem here.

I don't anticipate it taking long to fix. Do you need it to be merged into master as part of the requirements of your project?

fniephaus:
No, I don't think so. The code is already in the repository which should be enough I guess.
I didn't know about the new release of Graal, so I thought it might be better to fix it asap in order to keep the effort as low as possible.

chrisseaton:
Yeah we won't leave it more than a week max.

fniephaus:
No worries. Let us know if we can be of any help. Otherwise, we will leave this with you if that's ok :)

chrisseaton:
Did you have any benchmarks for testing that keyword arguments were properly optimised?

fniephaus:
Yes, sure we did, but they were very simple.

Here's one:

```python
def m(foo:, bar:)
  foo + bar
end

200.times.each do |times|
    beginning_time = Time.now
    (1..1000000).each do |i|
        m(foo: 1, bar: 2)
    end
    end_time = Time.now
    puts "Time elapsed #{(end_time - beginning_time)*1000} milliseconds"
end
```

and here's a more complex one:

```python
def m(foo:, bar:)
  foo + bar
end

def n(foo = 1, bar = 2)
  foo + bar
end

def o(foo, bar:)
    foo + bar
end

def p(a, b=2, c:)
    a + b + c
end

def q(foo:1, bar:2)
    foo + bar
end

def r(a, b=2, c=3, d:, e:5, f:)
    a+b+c+d+e+f
end

def s(a, b=2, c:, d:, **rest)
    a+b+c+d+rest[:e]
    #rest[:e]
end

def t(**r)
    r[:a]+r[:b]
end

def u(a, b)
    a + b[:a] + b[:b]
end

def assert(a, b)
if a != b
    puts "error: " + a.to_s + "!=" + b.to_s
end
end

200.times.each do |times|
    beginning_time = Time.now
    (1..1000000).each do |i|
        assert(n(1, 2), 3)
        assert(m(foo: 1, bar: 2), 3)
        assert(m({foo: 1, bar:2}), 3)
        assert(o(1, bar: 2), 3)
        assert(o(1, {bar:2}), 3)
        assert(p(1, c:3), 6)
        assert(q(foo:1, bar:2), 3)
        assert(q(foo:1), 3)
        #assert(q(), 3)
        assert(r(1,f:6,d:4), 21)
        assert(s(1,e:5, f:6,d:4,c:3), 15)
        assert(t(a:1, b:2, c:3), 3)
        assert(u(1, {a: 2, b: 3}), 6)
    end
    end_time = Time.now
    puts "Time elapsed #{(end_time - beginning_time)*1000} milliseconds"
end
```

chrisseaton:
Maybe I'll just convert some of the existing benchmarks to use kwargs.

fniephaus:
Haha, you probably should!

PR #2605, opened 2015-02-17
Fix File.read with UTF-16LE

lumeet:
```shell
File.open("a.txt", "w+") do |f|
  f.write "\xFF\xFE"
  f.close
end
File.read("a.txt", mode: "rb:bom|utf-8")
```
...results in...
```
<java.lang.ClassCastException: org.jruby.RubyNil cannot be cast to org.jruby.RubyFixnum>
```

A check for a null value seems to be missing after the BOM. 

The `test_truncate_wbuf` removal is just general cleanup. It doesn't have anything to do with the fix. I guess it's ok. :)


enebo:
@lumeet Sorry for leaving this so long but it no longer applies cleanly.  Can you update this PR?

lumeet:
@enebo I'm closing this. The bug is fixed by another commit and `test/mri/excludes/TestFile.rb` looks the same on master, too.

PR #2606, opened 2015-02-18
Test load service

mkristian:
this adds the missing pieces to avoid the legacy load-service.

the commit with passing the runtime to URLResource is actually not related to the load-service and will fix existing problems.

@ratnikov maybe you have some idea to avoid the double lookup on "." when it is also part of $LOAD_PATH: https://github.com/jruby/jruby/blob/a916d4563c16b317ccc409dcd12c9a09aede5d85/core/src/main/java/org/jruby/runtime/load/LibrarySearcher.java#L135

@headius I reverted the fix you "readded" to make jruby-openssl work on jruby-9k since this is not needed anymore for jruby-openssl-0.9.6. but let me know if support for jruby-openssl-0.9.5 is needed and I will revert the revert.

without legacy load-service there is no need for the https://github.com/jruby/jruby/blob/a916d4563c16b317ccc409dcd12c9a09aede5d85/core/src/main/java/org/jruby/util/CompoundJarURLStreamHandler.java which seems to cause some problems with newer JDK security permissions - see https://github.com/jruby/jruby-openssl/pull/29

before merging I would set the default to use the legacy load-service - to stay as is. but for running travis I used the oposite default.


core/src/main/java/org/jruby/runtime/load/LibrarySearcher.java
```diff
@@ -133,6 +132,20 @@ private FoundLibrary findResourceLibrary(String baseName, String suffix) {
           return findFileResource(baseName, suffix);
         }
 
+        // ruby does not load a relative path unless the current working directory is in $LOAD_PATH
+        if (baseName.startsWith("./") || loadService.loadPath.contains(".")) {
```
ratnikov: Is there a test that requires to skip ./ check if . is in load path? This seems rather strange to me, since I thought MRI treats ./ as a pretty special file index. Also, MRI (2.1.2 at least) seems to bypass LOAD_PATH, even if it contains .:

$ ruby -Ilib:. -e 'require "./foo"'
local

(the lib contains another foo.rb that would have printed 'in lib')

mkristian: ./ just bypasses the LOAD_PATH - yes
but if I do not find it with "." on the findFileResourceWithLoadPath then the classloader will find it. since the application classloader has the PWD (the one from the java world) on its classpath.

so the test of "." is needed to not fall into the classloader search and find it there. and the classloader finds always come with a "uri:classloader:" prefix which is not wrong but strange for files from the filesystem.

mkristian:
@ratnikov your suspicion was correct: https://github.com/jruby/jruby/blob/38f389298577402e9460a55461bdc3584333d437/core/src/main/java/org/jruby/runtime/load/LibrarySearcher.java#L145

and all tests pass :)

 **but** now I can
```jruby  -e 'require "t"```
where t.rb is on PWD and it gets found by the classloader. MRI can not require t.rb !

I will give it another trial and somehow add a test to it. the current test are failing since they do something like
```
Dir.chdir('test') do ....
```
and here the PWD from ruby is different from java (which does not change to 'test')

**Comment by mkristian** [comment]
the last failing test is "strange" - this one fails
```
bin/jruby "lib/ruby/stdlib/rake/rake_test_loader.rb" "test/jruby/test_load.rb" 
```
and which is the same execution used by mvn -Prake ... but this one without the rakd_test_loader succeeds:
```
bin/jruby "test/jruby/test_load.rb" 
```

anyways IMO not including the new load-service in an early state of a new major version makes it hard to switch any time later since we know people are OK with regression when jumping to a new major release but can be very unhappy wirh regressions on later point releases ;)

to I would drop the last commit and move **forward**

headius:
@mkristian It looks like this was not really merged, so I'm marking as Invalid/Duplicate.

PR #2618, opened 2015-02-24
[Truffle] Implemented Bignum#to_s when given a base

bjfish:
Tried to follow patterns in Fixnum. The core spec builds successfully.


chrisseaton:
Thanks very much

PR #2621, opened 2015-02-25
[Truffle] Implemented Symbol#upcase, added final in Symbol#swapcase

bjfish:
Reference String#upcase for implementation

### DIFF
--- a/truffle/src/main/java/org/jruby/truffle/nodes/core/SymbolNodes.java
+++ b/truffle/src/main/java/org/jruby/truffle/nodes/core/SymbolNodes.java
@@ -241,7 +241,28 @@ public abstract class SymbolNodes {
         public RubySymbol swapcase(RubySymbol symbol) {
             notDesignedForCompilation();
 
-            ByteList byteList = StringNodes.StringNodesHelper.swapcase(symbol.toRubyString());
+            final ByteList byteList = StringNodes.StringNodesHelper.swapcase(symbol.toRubyString());
+            return getContext().newSymbol(byteList);
+        }
+
+    }
+
+    @CoreMethod(names = "upcase")
+    public abstract static class UpcaseNode extends CoreMethodNode {
+
+        public UpcaseNode(RubyContext context, SourceSection sourceSection) {
+            super(context, sourceSection);
+        }
+
+        public UpcaseNode(UpcaseNode prev) {
+            super(prev);
+        }
+
+        @Specialization
+        public RubySymbol upcase(RubySymbol symbol) {
+            notDesignedForCompilation();
+
+            final ByteList byteList = StringNodes.StringNodesHelper.upcase(symbol.toRubyString());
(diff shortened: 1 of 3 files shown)


truffle/src/main/java/org/jruby/truffle/nodes/core/SymbolNodes.java
```diff
@@ -241,7 +241,28 @@ public SwapcaseNode(SwapcaseNode prev) {
         public RubySymbol swapcase(RubySymbol symbol) {
             notDesignedForCompilation();
 
-            ByteList byteList = StringNodes.StringNodesHelper.swapcase(symbol.toRubyString());
+            final ByteList byteList = StringNodes.StringNodesHelper.swapcase(symbol.toRubyString());
```
chrisseaton: Converting to a `RubyString` isn't ideal - it would be good if we could just use the `ByteList` directly. When we are using the JIT compiler the `RubyString` wrapper object would not be allocated, but in the interpreter it's a little bit of overhead.

bjfish:
@chrisseaton Thanks for the feedback. I made some fixes. Are there any more updates needed?

truffle/src/main/java/org/jruby/truffle/nodes/core/SymbolNodes.java
```diff
@@ -241,10 +240,63 @@ public SwapcaseNode(SwapcaseNode prev) {
...
+            final int length = byteList.length();
+            for (int i = 0; i < length; i++) {
+                final char c = byteList.charAt(i);
+                if ((c >= 'a') && (c <= 'z')) {
+                    byteList.set(i, c & 0x5f);
+                } else if ((c >= 'A') && (c <= 'Z')) {
+                    byteList.set(i, c ^ 0x20);
+                }
+            }
+            return byteList;
```
chrisseaton: Did you write this from scratch or did it come from the main JRuby code? I'd be surprised if they don't already have a routine we could copy for this.

bjfish: Check out this for upcase
https://github.com/jruby/jruby/blob/40784184eaa8a47630aa5de0ccc21f9488ae258e/core/src/main/java/org/jruby/RubySymbol.java#L378
https://github.com/jruby/jruby/blob/40784184eaa8a47630aa5de0ccc21f9488ae258e/core/src/main/java/org/jruby/RubyString.java#L1783

chrisseaton: The presence of `multiByteUpcase` https://github.com/jruby/jruby/blob/40784184eaa8a47630aa5de0ccc21f9488ae258e/core/src/main/java/org/jruby/RubyString.java#L1825 makes me think there might be some edge cases we aren't covering here.

However, we'll go by the specs for now.

PR #2625, opened 2015-02-26
Adding Examples of Running a Single Spec

bjfish:


chrisseaton:
You've used top level titles here - if you look at the rendered HTML you can see they're a bit big for these minor topics. Maybe use lower level titles?

bjfish:
@chrisseaton I've dropped this down to h4, thanks

PR #2626, opened 2015-02-26
[Truffle] Passing through encoding information from SymbolNode

bjfish:
to RubySymbol. Previously all were encoding as ASCII. A few string only methods are here to keep it backward compatible. I wasn't sure how to facture out the method calls to these.

I think SymbolNode could be considered to use ByteList instead of a String in the future.


nirvdrum:
There's a FindBugs failure, but I'm not sure it's related.  I'm going to merge this and then look into the matter.

PR #2630, opened 2015-02-27
embed internal fixes

kares:
while using JRuby's embed APIs in a slightly more advanced scenarios I run into several issues :

- `BiVariableMap` is not behaving as a valid Map instance (throwing NPEs on methods if not used "right")

- `ARGV` was mean to be kept (as the map is cleared) but unfortunately not correctly

added some tests covering intended (existing and fixed) functionality. trying to understand the details of context providers (and related embed variable impls) was a bit hard to read thus the did clean-up some (helped my brain to process the code). 

let me know if smt needs more work, targeting **jruby-1_7** - there are no (existing) API incompatibilities ...


kares:
rebased and commited on **jruby-1_7**

PR #2631, opened 2015-02-28
[Truffle] Including RegexpOptions in RubyRegexp 

bjfish:
to be able to return more correct options from the options method.

### DIFF
--- a/truffle/src/main/java/org/jruby/truffle/nodes/core/EncodingNodes.java
+++ b/truffle/src/main/java/org/jruby/truffle/nodes/core/EncodingNodes.java
@@ -24,16 +24,21 @@ import org.jcodings.specific.ASCIIEncoding;
 import org.jcodings.specific.UTF8Encoding;
 import org.jcodings.util.CaseInsensitiveBytesHash;
 import org.jcodings.util.Hash;
+import org.jruby.Ruby;
+import org.jruby.RubyObject;
 import org.jruby.runtime.encoding.EncodingService;
 import org.jruby.truffle.nodes.RubyNode;
 import org.jruby.truffle.nodes.coerce.ToStrNode;
 import org.jruby.truffle.nodes.coerce.ToStrNodeFactory;
 import org.jruby.truffle.runtime.RubyContext;
+import org.jruby.truffle.runtime.control.RaiseException;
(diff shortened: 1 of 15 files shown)


truffle/src/main/java/org/jruby/truffle/nodes/core/EncodingNodes.java
```diff
@@ -142,6 +147,98 @@ public Object isCompatible(RubyEncoding first, RubyEncoding second) {
...
+
+            Encoding compatibleEncoding = org.jruby.RubyEncoding.areCompatible(first, second);
+
+            if (compatibleEncodingProfile.profile(compatibleEncoding != null)) {
+                return RubyEncoding.getEncoding(compatibleEncoding);
+            } else {
+                return getContext().getCoreLibrary().getNilObject();
+            }
+        }
+
```
chrisseaton: Could we factor this out into a helper method that takes two `ByteList` objects?

truffle/src/main/java/org/jruby/truffle/nodes/core/EncodingNodes.java
```diff
@@ -217,6 +314,21 @@ public RubyEncoding defaultExternal(RubyEncoding encoding) {
             return encoding;
         }
 
+        @Specialization
+        public RubyEncoding defaultExternal(RubyString encodingString) {
+            notDesignedForCompilation();
+
+            final RubyEncoding rubyEncoding = RubyEncoding.getEncoding(encodingString.toString());
+            getContext().getRuntime().setDefaultExternalEncoding(rubyEncoding.getEncoding());
```
chrisseaton: This sets the default external encoding on the JRuby runtime. I'm not sure if we re-use that same value in Truffle. But since it passes the specs I won't worry too much.

nirvdrum:
Thanks! In the future, if not too much of a hassle, please split different method implementations into different commits.  We've had to revert method implementations in the past because of failing benchmarks and what not.  It's just much easier to cope with that if the commits are more scoped in nature.  I usually use "git add -p" to do it if I've been working in a file for a while.  But, if it gets to hairy, a holistic commit like this is fine -- we're pragmatic.

PR #2633, opened 2015-02-28
Warn on block for String#lines

lumeet:
"passing a block to String#lines is deprecated" is displayed when a
block is used. MRI does this since 2.0.


nirvdrum:
Looks good.  Thanks!

PR #2638, opened 2015-03-01
[Truffle] Implemented Regexp#match_start for strscan.

bjfish:
This passes the match_start_spec.rb which isn't in our repo currently: https://github.com/rubinius/rubinius/blob/58d397733b6ec1ae003b296ba4ab27a65e8629dc/spec/core/regexp/match_start_spec.rb

It would be nice to have more specs for this method to ensure correctness.


chrisseaton:
You can just copy and paste that spec straight into this repo.

bjfish:
@chrisseaton I'll be updating this to add search_from, I think this match_start is a little off and will update too.

PR #2640, opened 2015-03-02
[Truffle] Regex #casefold? #=== for Symbol

bjfish:


truffle/src/main/ruby/core/rubinius/common/regexp.rb
```diff
@@ -128,6 +128,10 @@ def encoding
     source.encoding
   end
 
+  def casefold?
+    (options & IGNORECASE) > 0 ? true : false
+  end
+
```
chrisseaton: This code came from Rubinius, right?

truffle/src/main/java/org/jruby/truffle/nodes/core/RegexpNodes.java
```diff
@@ -126,6 +126,13 @@ public Object match(RubyRegexp regexp, RubyString string) {
             return regexp.matchCommon(string, true, false) != getContext().getCoreLibrary().getNilObject();
         }
 
+        @Specialization
+        public Object match(RubyRegexp regexp, RubySymbol symbol) {
+            notDesignedForCompilation();
+
+            return regexp.matchCommon(symbol.toRubyString(), true, false) != getContext().getCoreLibrary().getNilObject();
```
chrisseaton: In these kinds of situations we probably want to factor out `matchCommon` to use `ByteList`, rather than converting `Symbol` to `String`.

truffle/src/main/ruby/core/rubinius/common/regexp.rb
bjfish: That's right

truffle/src/main/java/org/jruby/truffle/nodes/core/RegexpNodes.java
bjfish: This refactoring doesn't look straightforward because of the way it is implemented right now. The MatchData uses RubyString as source.

truffle/src/main/ruby/core/rubinius/common/regexp.rb
chrisseaton: I don't know if anyone told you, but we're trying to keep to version 2.4.1 of Rubinius, just to be consistent. We'll probably upgrade later on.

truffle/src/main/java/org/jruby/truffle/nodes/core/RegexpNodes.java
chrisseaton: Yeah, you're right. Disregard.

truffle/src/main/ruby/core/rubinius/common/regexp.rb
bjfish: That's good to know. Looks like this method is the same in 2.4.1

PR #2644, opened 2015-03-04
[Truffle] Adding String#subpattern in ruby

bjfish:


nirvdrum:
Looks good.  If you're going to be doing anything more with String, please coordinate with me on IRC.  I'm going through a lot of the class right now, as you may have seen.

nirvdrum:
Thanks again!

PR #2645, opened 2015-03-04
[Truffle] Adding Implementations for Symbol#slice specs.

bjfish:
String#num_bytes was added as a regular method because adding it as an attr_reader did not appear to work.

### DIFF
--- a/truffle/src/main/ruby/core/rubinius/bootstrap/string.rb
+++ b/truffle/src/main/ruby/core/rubinius/bootstrap/string.rb
@@ -76,4 +76,8 @@ class String
     raise PrimitiveFailure, "String#find_character primitive failed"
   end
 
+  def num_bytes
+    @num_bytes
+  end
+
 end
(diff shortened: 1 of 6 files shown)


truffle/src/main/ruby/core/rubinius/bootstrap/string.rb
```diff
@@ -76,4 +76,8 @@ def find_character(offset)
     raise PrimitiveFailure, "String#find_character primitive failed"
   end
 
+  def num_bytes
+    @num_bytes
```
chrisseaton: The reason you couldn't use this with an attribute reader is that we replace reads of `@num_bytes` in our translator with a method that reads it from the actual string object - see https://github.com/jruby/jruby/blob/4eb00454a2e6e257513fc4013764080f259e049a/truffle/src/main/java/org/jruby/truffle/translator/BodyTranslator.java#L1568-L1573

PR #2646, opened 2015-03-04
[Truffle] - String#capitalize! returns nil when no changes are made

pedroandrade:


chrisseaton:
Thanks!

If you're working on `String` methods you should coordinate with @nirvdrum on IRC, as he's working there full time and we don't want to conflict or duplicate work by mistake.

PR #2649, opened 2015-03-05
[Truffle] Bulk untagging passing specs

bjfish:


chrisseaton:
Wow. Did you say you wrote a script to help do this without the problem of things hanging?

I'll just let the tests run and then merge.

bjfish:
@chrisseaton yes, the script works well for things hanging because of the time outs it has. 

But it still has the problem of untagging specs that don't pass when the whole suite is run. I suspect there is some test pollution between tests when they are all run.

Here's a script that just runs each core directory with a 90 second time out (probably more useful in hindsight):
https://gist.github.com/bjfish/75d144f88c5870053fc2

Here's a link to the script that does jt untag, jt test, git checkout for each spec (about 4 hours):
https://gist.github.com/bjfish/eafa28b6bc82eb2299a3

chrisseaton:
The difficulty is that our system uses lots and lots of hidden state for optimisations. So running one test before another can change the state of some method which causes it to break, where it wouldn't have it if you hadn't run the other test. We're going to tackle this with some code for randomising some of that state.

PR #2656, opened 2015-03-06
[Truffle] Adding Unique IDs to neverDesignedForCompilation to trace

bjfish:
This is just an idea in case you like it @chrisseaton 

Add this method to RubyNode -
```
    public static void notDesignedForCompilation(String message) {
        CompilerAsserts.neverPartOfCompilation(message);
    }
```
 
Add Unique ID to notDesignedForCompilation method in All Methods -
```
    @Specialization(guards = {"!isRubyModule", "!isRubyBignum"})
    public Object alias(RubyBasicObject object) {
        notDesignedForCompilation("5c0619014d0a4e0f9880499ff32473e1");
   ```
Graal Output includes Unique IDs to trace which method caused the optfail
```
[truffle] opt fail         parse_string:json_test.rb:1068                              |Reason com.oracle.graal.nodes.util.GraphUtil$2: This code path should never be part of a compilation. [14342|Const(50ccd29854f84c51ab8b57d2c17deb1e)] 
[truffle] opt fail         parse_object:json_test.rb:1168                              |Reason com.oracle.graal.nodes.util.GraphUtil$2: This code path should never be part of a compilation. [29103|Const(f5e9acd064e245d5b35a193003490752)]
```

### CODE REVIEW

**Comment by chrisseaton** [comment]
I think having a reason is a good idea, and I can see the advantage if they were descriptive, but if it's just an identifier can't you get that information anyway from `-G:+TruffleExceptionsAreFatal` and `-G:+DumpOnError`?

**Comment by bjfish** [comment]
@chrisseaton I think with these options I could only trace the first optfail that occurs.

**Comment by eregon** [comment]
:+1: for adding the reason, but I think a descriptive comment is quite more helpful and less cryptic.
I do see the point of having immediately a unique identifier to find the source though, Graal/Truffle should really provide that if possible.
A good description of the reason would be quite nice asa  documentation to why that method is not designed for compilation since it is not always trivial.

**Comment by bjfish** [comment]
@eregon I agree something descriptive would be more useful (still should be a unique message though). 
I only used a UUID right now because it would take significant amount of time to write a description for the 500+ places notDesignedForCompilation is used.

**Comment by eregon** [comment]
Maybe we could just derive some description from the node class name or the particular specialization node? Or use the backtrace if the Truffle description with `-G:+DumpOnError` is not precise enough.

**Comment by bjfish** [comment]
@eregon also :+1: , I wish this was provided: "Graal/Truffle should really provide that if possible."

**Comment by nirvdrum** [comment]
Maybe I'm missing something, but if Truffle auto-generated the ID, how would you grep through the source for it?

**Comment by bjfish** [comment]
@eregon I tried using the node to get a backtrace but for some reason this was causing Graal to just hang when run

@nirvdrum 
```ruby
require 'securerandom'
files =  Dir.glob("truffle/**/*.java")
puts files.size
puts files.to_s
files.each do |filepath|
   text = File.read(filepath)
   replace = text.gsub(/notDesignedForCompilation\(\);/)  { |num| "notDesignedForCompilation(\"#{SecureRandom.uuid.gsub('-','')}\");" }
   File.open(filepath, "w") {|file| file.puts replace}
end
```

eregon:
@bjfish do you have a particular example at hand so we can compare the alternatives?

nirvdrum:
@bjfish That's us generating the ID, which I'm actually okay with.

bjfish:
@eregon Sorry, I don't think I have anything to compare. 

The dynamic stuff I've tried just causes Graal to hang / be really slow.

chrisseaton:
@bjfish we really appreciate this PR but we don't think it's right for the project. I think the unique identifiers in the source code look really out of place and really the emphasis should be on removing these `notDesignedForCompilation`s rather than finding ways to making working with them easier.

We do understand your use case though, and we'll look at adding an option to Graal that prints the backtrace that you get with `TruffleCompilationExceptionsAreFatal` but doesn't actually exit.

Since you have your script, you can actually keep running that locally and then just not committing the results if you want to.

PR #2667, opened 2015-03-09
[Truffle] Fixed class variable lookup from singleton classes by attaching their companion class to them.

nirvdrum:
@eregon I made the changes we discussed.  Any additional feedback would be great.

### DIFF
--- a/truffle/src/main/java/org/jruby/truffle/runtime/ModuleOperations.java
+++ b/truffle/src/main/java/org/jruby/truffle/runtime/ModuleOperations.java
@@ -243,6 +243,30 @@ public abstract class ModuleOperations {
             }
         }
 
+        // If singleton class, check attached module as well.
+        if (module instanceof RubyClass) {
+            RubyClass klass = (RubyClass) module;
+
+            if (klass.isSingleton() && klass.getAttached() != null) {
+
+                // Look in the attached module.
+                value = klass.getAttached().getClassVariables().get(name);
+
+                if (value != null) {
+                    return value;
+                }
+
+                // Look in the attached module's ancestors.
+                for (RubyModule ancestor : klass.getAttached().parentAncestors()) {
+                    value = ancestor.getClassVariables().get(name);
+
+                    if (value != null) {
+                        return value;
+                    }
+                }
+            }
+        }
+
         // Nothing found
 
         return null;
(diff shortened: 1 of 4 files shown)


truffle/src/main/java/org/jruby/truffle/runtime/ModuleOperations.java
```diff
@@ -243,6 +243,30 @@ public static Object lookupClassVariable(RubyModule module, String name) {
...
+                }
+
+                // Look in the attached module's ancestors.
+                for (RubyModule ancestor : klass.getAttached().parentAncestors()) {
+                    value = ancestor.getClassVariables().get(name);
+
+                    if (value != null) {
+                        return value;
+                    }
+                }
```
eregon: We should not check both ancestry chains. MRI lookup is at https://github.com/ruby/ruby/blob/badb4de72a08b199f1a63864531d47beafca818b/variable.c#L2355-L2384.
So, I would add the `if attached` above the ancestors lookup, and in that `if` do the lookup on the attached module and set module to the attached module.

truffle/src/main/java/org/jruby/truffle/runtime/core/RubyBasicObject.java
```diff
@@ -104,8 +104,13 @@ public RubyClass getSingletonClass(Node currentNode) {
...
 
-        metaClass = RubyClass.createSingletonClassOfObject(getContext(), logicalClass,
-                String.format("#<Class:#<%s:0x%x>>", logicalClass.getName(), getObjectID()));
+        if (this instanceof RubyModule) {
+            metaClass = RubyClass.createSingletonClassOfObject(getContext(), logicalClass, (RubyModule) this,
+                    String.format("#<Class:#<%s:0x%x>>", logicalClass.getName(), getObjectID()));
+        } else {
+            metaClass = RubyClass.createSingletonClassOfObject(getContext(), logicalClass, null,
+                    String.format("#<Class:#<%s:0x%x>>", logicalClass.getName(), getObjectID()));
+        }
```
eregon: Good! Maybe to avoid duplication only define a `attached` variable in the `if`.

eregon:
Looks good, please see my two and I think we'll have the perfect commit :smile:

truffle/src/main/java/org/jruby/truffle/runtime/ModuleOperations.java
nirvdrum: Thanks.  I thought it occurred afterwards.

nirvdrum:
Thanks again for the review.

PR #2670, opened 2015-03-09
[Truffle] Adding Array#cycle to array.rb.

bjfish:


spec/truffle/tags/core/array/cycle_tags.txt
```diff
@@ -1,11 +1,2 @@
 fails:Array#cycle iterates indefinitely when not passed a count
 fails:Array#cycle iterates indefinitely when passed nil
```
chrisseaton: How come these two don't pass?

bjfish: @chrisseaton The tests were hanging, not sure of the cause.

chrisseaton: ok np

PR #2680, opened 2015-03-10
[Truffle] Adding Array#repeated_permutations to array.rb

bjfish:
Is recursive ruby code okay?


bjfish:
Is it okay for me to replace your Array#rindex shim? @nirvdrum Should I do Array#reverse too?

chrisseaton:
Yes it's generally good to replace shim methods with real implementations.

nirvdrum:
Feel free to replace the `Array#reverse` shim.  I did it this way because the Rubinius definition doesn't pull in cleanly.  It seems like something that might be better served in Java anyway.  But have at it either way.

PR #2681, opened 2015-03-11
[Truffle] Adding Array#combination to array.rb

bjfish:
Tuple.pattern adapted from 
https://github.com/rubinius/rubinius/blob/master/vm/builtin/tuple.cpp

Please review.


bjfish:
I didn't call `reallocate`(see the comment there) in Array#fill because we don't use Tuple internally. Please review.

chrisseaton:
What do you mean by 'adapted form tuple.cpp'? I only see Ruby code here? Have you written some from scratch by copying the Rubinius C++ code into Ruby?

bjfish:
@chrisseaton  that is correct, ruby from scratch after looking at the c++ code. I'm not the best at reading c++ code so requested some review on that.

Also, in reverse! Using an in-place reverse instead of the `Rubinius.primitive :tuple_reverse` from tuple.rb

bjfish:
@chrisseaton here is the code for Tuple.pattern
```c++
  // @todo performance primitive; could be replaced with Ruby
  Tuple* Tuple::pattern(STATE, Fixnum* size, Object* val) {
    native_int cnt = size->to_native();

    if(cnt < 0) {
      Exception::argument_error(state, "negative tuple size");
    }

    Tuple* tuple = Tuple::create_dirty(state, cnt);

    for(native_int i = 0; i < cnt; i++) {
      // bounds checking is covered because we instantiated the tuple
      // in this method
      tuple->field[i] = val;
    }

    // val is referend size times, we only need to hit the write
    // barrier once
    tuple->write_barrier(state, val);
    return tuple;
  }
```
I thought that just translated to 
```ruby
def self.pattern(size, val)
  Array.new(size, val)
end
```

chrisseaton:
It does yeah. I was asking because I was worried about the licensing requirements if code had been copied from Rubinius C++ to Ruby - we'd have to mark that very clearly. But this is trivial code and hasn't really been copied - you've just deferred to an existing method of our implementation.

PR #2682, opened 2015-03-11
[Truffle] - Initial implementation of Ruby Random class.

lucasallan:
Hey @chrisseaton @nirvdrum could you guys take a look?
Thanks!

### DIFF
--- a/truffle/src/main/java/org/jruby/truffle/nodes/rubinius/RandomPrimitiveNodes.java
+++ b/truffle/src/main/java/org/jruby/truffle/nodes/rubinius/RandomPrimitiveNodes.java
@@ -1,0 +1,137 @@
+package org.jruby.truffle.nodes.rubinius;
+
+import com.oracle.truffle.api.CompilerDirectives;
+import com.oracle.truffle.api.dsl.Specialization;
+import com.oracle.truffle.api.source.SourceSection;
+import org.jruby.Ruby;
+import org.jruby.RubyFixnum;
+import org.jruby.RubyNumeric;
+import org.jruby.RubyRandom;
+import org.jruby.runtime.builtin.IRubyObject;
+import org.jruby.truffle.runtime.RubyContext;
+import org.jruby.truffle.runtime.core.RubyBasicObject;
+import org.jruby.truffle.runtime.core.RubyBignum;
+
+import java.math.BigInteger;
+import java.security.SecureRandom;
+import java.util.Random;
+
+/**
+ * Rubinius primitives associated with the Ruby {@code Random} class.
+ */
+
+public abstract class RandomPrimitiveNodes {
+
+    @RubiniusPrimitive(name = "randomizer_seed")
+    public static abstract class RandomizerSeedPrimitiveNode extends RubiniusPrimitiveNode {
+
+        public RandomizerSeedPrimitiveNode(RubyContext context, SourceSection sourceSection) {
+            super(context, sourceSection);
+        }
+
+        public RandomizerSeedPrimitiveNode(RandomizerSeedPrimitiveNode prev) {
+            super(prev);
+        }
+
+        @Specialization
+        public long randomizerSeed(RubyBasicObject random) {
+            notDesignedForCompilation();
+
+            return System.currentTimeMillis();
+        }
+
+    }
+
+    @RubiniusPrimitive(name = "randomizer_rand_float")
+    public static abstract class RandomizerRandFloatPrimitiveNode extends RubiniusPrimitiveNode {
+
+        public RandomizerRandFloatPrimitiveNode(RubyContext context, SourceSection sourceSection) {
+            super(context, sourceSection);
+        }
+
+        public RandomizerRandFloatPrimitiveNode(RandomizerRandFloatPrimitiveNode prev) {
+            super(prev);
+        }
+
+        @Specialization
+        public double randomizerRandFloat(RubyBasicObject random) {
+            notDesignedForCompilation();
+
+            return Math.random();
+        }
+
+    }
+
+    @RubiniusPrimitive(name = "randomizer_rand_int")
+    public static abstract class RandomizerRandIntPrimitiveNode extends RubiniusPrimitiveNode {
+
+        public RandomizerRandIntPrimitiveNode(RubyContext context, SourceSection sourceSection) {
+            super(context, sourceSection);
+        }
+
+        public RandomizerRandIntPrimitiveNode(RandomizerRandIntPrimitiveNode prev) {
+            super(prev);
+        }
+
+        @Specialization
+        public long randomizerRandInt(RubyBasicObject random, Integer limit) {
+            notDesignedForCompilation();
+
+            return RandomPrimitiveHelper.randomInt(getContext().getRuntime().getCurrentContext().getRuntime(), limit);
+        }
+
+        @Specialization
+        public long randomizerRandInt(RubyBasicObject random, Long limit) {
+            notDesignedForCompilation();
+
+            return RandomPrimitiveHelper.randomInt(getContext().getRuntime().getCurrentContext().getRuntime(), limit);
+        }
+    }
+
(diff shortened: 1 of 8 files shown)


truffle/src/main/ruby/core/rubinius/common/random.rb
```diff
@@ -0,0 +1,177 @@
+#
+# Extracted from https://github.com/rubinius/rubinius/blob/v2.4.1/kernel/common/random.rb
+#
```
chrisseaton: This needs to be the full Rubinius license. Can you copy from another file in `common`?

lucasallan: For sure. give me a minute.

truffle/src/main/ruby/core/rubinius/common/random.rb
```diff
@@ -0,0 +1,203 @@
+# Copyright (c) 2007-2014, Evan Phoenix and contributors
```
lucasallan: @chrisseaton Done!

truffle/src/main/java/org/jruby/truffle/nodes/rubinius/RandomPrimitiveNodes.java
```diff
@@ -0,0 +1,113 @@
...
+        public RandomizerSeedPrimitiveNodes(RubyContext context, SourceSection sourceSection) {
+            super(context, sourceSection);
+        }
+
+        public RandomizerSeedPrimitiveNodes(RandomizerSeedPrimitiveNodes prev) {
+            super(prev);
+        }
+
+        @Specialization
+        public Long randomizerSeed(RubyBasicObject random) {
```
chrisseaton: Should be just 'long'

truffle/src/main/java/org/jruby/truffle/nodes/rubinius/RandomPrimitiveNodes.java
```diff
@@ -0,0 +1,113 @@
...
+            return System.currentTimeMillis();
```
chrisseaton: Is this what Rubinius uses for the randomiser seed?

truffle/src/main/java/org/jruby/truffle/nodes/rubinius/RandomPrimitiveNodes.java
```diff
@@ -0,0 +1,113 @@
...
+        public RandomizerRandFloatPrimitiveNodes(RubyContext context, SourceSection sourceSection) {
+            super(context, sourceSection);
+        }
+
+        public RandomizerRandFloatPrimitiveNodes(RandomizerRandFloatPrimitiveNodes prev) {
+            super(prev);
+        }
+
+        @Specialization
+        public Double randomizerRandFloat(RubyBasicObject random) {
```
chrisseaton: Again - just `double`

truffle/src/main/java/org/jruby/truffle/nodes/rubinius/RandomPrimitiveNodes.java
```diff
@@ -0,0 +1,113 @@
...
+/**
+ * Rubinius primitives associated with the Ruby {@code Random} class.
+ */
+
+public abstract class RandomPrimitiveNodes {
+
+    @RubiniusPrimitive(name = "randomizer_seed")
+    public static abstract class RandomizerSeedPrimitiveNodes extends RubiniusPrimitiveNode {
+
+        private final ConditionProfile nullByteProfile = ConditionProfile.createBinaryProfile();
```
chrisseaton: You've got this `nullByteProfile` in all your nodes - is this left over from some copy and paste? You never use it.

truffle/src/main/java/org/jruby/truffle/nodes/rubinius/RandomPrimitiveNodes.java
```diff
@@ -0,0 +1,113 @@
...
+            super(context, sourceSection);
+        }
+
+        public RandomizerRandIntPrimitiveNodes(RandomizerRandIntPrimitiveNodes prev) {
+            super(prev);
+        }
+
+        @Specialization
+        public Long randomizerRandInt(RubyBasicObject random) {
+            return System.currentTimeMillis();
```
chrisseaton: Is this right? Random int by using the current time? Shouldn't it use a PRNG?

truffle/src/main/java/org/jruby/truffle/nodes/rubinius/RandomPrimitiveNodes.java
```diff
@@ -0,0 +1,113 @@
...
+        public RubyBignum randomizerGenSeed(RubyBasicObject random) {
+
+            BigInteger randomInt = RandomPrimitiveHelper.randomSeed(getContext().getRuntime().getCurrentContext().runtime);
+
+            return new RubyBignum(getContext().getCoreLibrary().getBignumClass(), randomInt);
+        }
+    }
+
+    static class RandomPrimitiveHelper {
+        private static final int DEFAULT_SEED_CNT = 4;
```
chrisseaton: Whhat does `DEFAULT_SEED_CNT` mean? Is this code copied from somewhere?

lucasallan: This is copied from JRuby `core/src/main/java/org/jruby/RubyRandom.java:198`.

truffle/src/main/java/org/jruby/truffle/nodes/rubinius/RandomPrimitiveNodes.java
lucasallan: Removing it...

truffle/src/main/ruby/core/rubinius/common/random.rb
```diff
@@ -0,0 +1,203 @@
...
+  end
+
+  # Generate a random Float, in the range 0...1.0
+  def random_float
+    Rubinius.primitive :randomizer_rand_float
+    raise PrimitiveFailure, "Randomizer#rand_float primitive failed"
+  end
+
+  # Generate a random Integer, in the range 0...limit
+  def random_integer(limit)
```
chrisseaton: You seem to ignore the limit here? I think the primitive should accept a limit parameter and use it.

chrisseaton:
It looks like you've gotten the idea of how to import code form Rubinius - but you have some failures in the core specs and FindBugs - see the Travis report. Also, maybe you were doing a trivial initial implementation, but you actually need to implement a PRNG using the seed in the randomiser primitives.

truffle/src/main/java/org/jruby/truffle/nodes/rubinius/RandomPrimitiveNodes.java
chrisseaton: That's fine - but we need a clear comment saying which file it came from, as otherwise we're moving it from a file containing a license and attribution to one that doesn't.

truffle/src/main/ruby/core/rubinius/common/random.rb
lucasallan: The `limit` doesn't seem to be used anywhere.

chrisseaton: Yeah, this is a subtle bit of the semantics of Rubinius. `Rubinius.primitive :randomizer_rand_int` is a bit like a `super` call, in that it implicitly sends the same arguments as the enclosing method had. So it's as if you had written `Rubinius.primitive :randomizer_rand_int(limit)` (but you can't actually write that). That's how the `limit` argument gets into the primitive.

lucasallan: It's done. Could you review again? @chrisseaton

chrisseaton:
Just waiting for Travis to finish...

PR #2685, opened 2015-03-11
[Truffle] Adding #to_int to Array#at paramters.

bjfish:
I removed lowerFixnumParameter because it wasn’t handling double.

I’m not sure if the guard is needed?

### DIFF
--- a/truffle/src/main/java/org/jruby/truffle/nodes/core/ArrayNodes.java
+++ b/truffle/src/main/java/org/jruby/truffle/nodes/core/ArrayNodes.java
@@ -21,12 +21,15 @@ import com.oracle.truffle.api.nodes.UnexpectedResultException;
 import com.oracle.truffle.api.source.SourceSection;
 import com.oracle.truffle.api.utilities.BranchProfile;
 
+import org.jruby.RubyObject;
 import org.jruby.runtime.Visibility;
 import org.jruby.runtime.builtin.IRubyObject;
 import org.jruby.truffle.nodes.CoreSourceSection;
 import org.jruby.truffle.nodes.RubyNode;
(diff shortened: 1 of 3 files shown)


truffle/src/main/java/org/jruby/truffle/nodes/core/ArrayNodes.java
```diff
@@ -446,10 +449,11 @@ public Object setIntegerFixnumRange(VirtualFrame frame, RubyArray array, RubyRan
 
     }
 
-    @CoreMethod(names = "at", required = 1, lowerFixnumParameters = 0)
```
bjfish: Safe to remove this?

truffle/src/main/java/org/jruby/truffle/nodes/core/ArrayNodes.java
```diff
@@ -470,6 +474,20 @@ public Object at(VirtualFrame frame, RubyArray array, int index) {
             return readNode.executeRead(frame, array, index);
         }
 
+        @Specialization(guards = {"!isInteger(arguments[1])"})
```
bjfish: Is Guard needed?

nirvdrum:
We could have `FixnumLowerNode` truncate, but that's probably not safe in every case.  I think you're going to need a new a `@Specialization` on `double` that does the correct conversion.  I think for the most part, Ruby truncates and then uses the resulting `Fixnum` for the index.

truffle/src/main/java/org/jruby/truffle/nodes/core/ArrayNodes.java
nirvdrum: Per my comment on the PR, I think this should stay and a new specialization for `double` added.

PR #2689, opened 2015-03-12
[Truffle] Trying to add #to_ary to Array#replace

bjfish:
@nirvdrum @chrisseaton Is this correct place to use a CreateCast? If not, why not?

Also trying to implement ToAryNode here.

### DIFF
--- a/truffle/src/main/java/org/jruby/truffle/nodes/coerce/ToAryNode.java
+++ b/truffle/src/main/java/org/jruby/truffle/nodes/coerce/ToAryNode.java
@@ -1,0 +1,66 @@
+package org.jruby.truffle.nodes.coerce;
+
+import com.oracle.truffle.api.CompilerDirectives;
+import com.oracle.truffle.api.dsl.NodeChild;
+import com.oracle.truffle.api.dsl.Specialization;
+import com.oracle.truffle.api.frame.VirtualFrame;
+import com.oracle.truffle.api.source.SourceSection;
+import org.jruby.truffle.nodes.RubyNode;
+import org.jruby.truffle.nodes.dispatch.CallDispatchHeadNode;
+import org.jruby.truffle.nodes.dispatch.DispatchHeadNodeFactory;
+import org.jruby.truffle.runtime.RubyContext;
+import org.jruby.truffle.runtime.control.RaiseException;
+import org.jruby.truffle.runtime.core.RubyArray;
+
+
+@NodeChild(value = "child", type = RubyNode.class)
+public abstract class ToAryNode extends RubyNode {
+
+    @Child private CallDispatchHeadNode toAryNode;
+
+    public ToAryNode(RubyContext context, SourceSection sourceSection) {
+        super(context, sourceSection);
+    }
+
+    public ToAryNode(ToAryNode prev) {
+        super(prev);
+    }
+
+    @Specialization
+    public RubyArray coerceRubyArray(RubyArray rubyArray) {
+        return rubyArray;
+    }
+
+    @Specialization(guards = "!isRubyArray")
+    public RubyArray coerceObject(VirtualFrame frame, Object object) {
+        notDesignedForCompilation();
+
+        if (toAryNode == null) {
+            CompilerDirectives.transferToInterpreter();
+            toAryNode = insert(DispatchHeadNodeFactory.createMethodCall(getContext()));
+        }
+
+        final Object coerced;
+
+        try {
+            coerced = toAryNode.call(frame, object, "to_ary", null);
+        } catch (RaiseException e) {
+            if (e.getRubyException().getLogicalClass() == getContext().getCoreLibrary().getNoMethodErrorClass()) {
+                CompilerDirectives.transferToInterpreter();
+
+                throw new RaiseException(
+                    getContext().getCoreLibrary().typeErrorNoImplicitConversion(object, "Array", this));
+            } else {
+                throw e;
+            }
+        }
+        if (coerced instanceof RubyArray) {
+            return (RubyArray) coerced;
+        } else {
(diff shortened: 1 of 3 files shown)


truffle/src/main/java/org/jruby/truffle/nodes/coerce/ToAryNode.java
```diff
@@ -0,0 +1,58 @@
...
+    }
+
+    public ToAryNode(ToAryNode prev) {
+        super(prev);
+    }
+
+    @Specialization
+    public Object coerceObject(VirtualFrame frame, Object object) {
+        notDesignedForCompilation();
+        if (object instanceof RubyArray) {
```
nirvdrum: This should be its own specialization so in the common case of just being an array, you get a very small method body that could be inlined.  You'll then need to guard the generic case.

nirvdrum:
CreateCast is fine here.  It's allowing you to normalize one argument of all specializations to a fixed type.  I'm not sure how frequently ToAryNode will be used otherwise, but that's fine.

truffle/src/main/java/org/jruby/truffle/nodes/coerce/ToAryNode.java
```diff
@@ -0,0 +1,58 @@
...
+import org.jruby.truffle.runtime.RubyContext;
+import org.jruby.truffle.runtime.control.RaiseException;
+import org.jruby.truffle.runtime.core.RubyArray;
+
+
+@NodeChild(value = "child", type = RubyNode.class)
+public abstract class ToAryNode extends RubyNode {
+
+    @Child
+    private CallDispatchHeadNode toAryNode;
```
nirvdrum: Minor style thing, but in every other case I've seen we have the annotation and the field declaration on the same line.

nirvdrum:
Your `ToAryNode` works fine in the case you've defined it for.  If you'd like to able to use it in other places to provide a simple way to call `to_ary` on an object and return a RubyArray, you'll need to add the abstract execute* methods like in ToStrNode.

truffle/src/main/java/org/jruby/truffle/nodes/coerce/ToAryNode.java
chrisseaton: To clarify, one specialisation should be `RubyArray`, and then other should be `Object` and then also guarded `!isRubyArray`. Otherwise you could use the `Object` one first and then keep happily using it when you get a `RubyArray` as the type fits.

bjfish:
@nirvdrum Thanks for all the feedback. I've updated the PR with a rebased commit that I believe addresses your comments.

truffle/src/main/java/org/jruby/truffle/nodes/coerce/ToAryNode.java
```diff
@@ -0,0 +1,59 @@
...
+            if (e.getRubyException().getLogicalClass() == getContext().getCoreLibrary().getNoMethodErrorClass()) {
+                CompilerDirectives.transferToInterpreter();
+
+                throw new RaiseException(
+                    getContext().getCoreLibrary().typeErrorNoImplicitConversion(object, "Array", this));
+            } else {
+                throw e;
+            }
+        }
+        return coerced;
```
nirvdrum: If you look at the other coercion nodes, they check the value of the return type.  I think you need to do that here as well, since someone could define `to_ary` to return something other than a `RubyArray`.  Here are the two cases I saw with `Array#replace`:

```
$ ruby -e 'class X; end; [].replace(X.new)'
-e:1:in `replace': no implicit conversion of X into Array (TypeError)
	from -e:1:in `<main>'
```
That's the missing `to_ary` implementation, which you've addressed in the code above.  But there's also:

```
$ ruby -e 'class X; def to_ary; :hello; end; end; [].replace(X.new)'
-e:1:in `replace': can't convert X to Array (X#to_ary gives Symbol) (TypeError)
	from -e:1:in `<main>'
```

That's the case you're missing.

truffle/src/main/java/org/jruby/truffle/nodes/coerce/ToAryNode.java
```diff
@@ -0,0 +1,59 @@
...
+    public ToAryNode(ToAryNode prev) {
+        super(prev);
+    }
+
+    @Specialization
+    public Object coerceRubyArray(RubyArray rubyArray) {
+        return rubyArray;
+    }
+
+    @Specialization( guards = "!isRubyArray")
```
nirvdrum: Minor issue, but this internal spacing is inconsistent with what we do elsewhere.  Just keep an eye out for that in the future.

nirvdrum:
@bjfish No problem.  If you can address my last comment about handling the case when `to_ary` returning the wrong type, I think this will be good to merge.

bjfish:
@nirvdrum I've added to return type check, updated the return type to RubyArray (ToStrNode returns RubyString) , and fixed the guard spacing.

truffle/src/main/java/org/jruby/truffle/nodes/coerce/ToAryNode.java
```diff
@@ -0,0 +1,66 @@
...
+    public ToAryNode(RubyContext context, SourceSection sourceSection) {
+        super(context, sourceSection);
+    }
+
+    public ToAryNode(ToAryNode prev) {
+        super(prev);
+    }
+
+    @Specialization
+    public Object coerceRubyArray(RubyArray rubyArray) {
```
nirvdrum: Good point on the return type.  This should be `RubyArray`, too.

bjfish:
@nirvdrum I fixed the other return type.

nirvdrum:
Looks great.

Issue #2591, opened 2015-02-12
double-quoted UTF8 hash key has the wrong encoding


k77ch7:
The following test.rb works on MRI Ruby 2.2.0,
but gives the wrong encoding and  ArgumentError on master.

test.rb
```
# encoding: utf-8
h = { "Ãa1":  true }
puts h.keys.first.encoding
puts h
```

MRI Ruby 2.2.0
```
$ ruby ~/development/test.rb
UTF-8
{:Ãa1=>true}
```

JRuby master
```
$ jruby ~/development/test.rb
US-ASCII
ArgumentError: invalid byte sequence in US-ASCII
     inspect at org/jruby/RubySymbol.java:241
     inspect at org/jruby/RubySymbol.java:230
     inspect at org/jruby/RubyHash.java:844
        to_s at org/jruby/RubyHash.java:909
        puts at org/jruby/RubyIO.java:2406
        puts at org/jruby/RubyKernel.java:542
  __script__ at /Users/k77ch7/test.rb:4
```

My env
```
jruby 9.0.0.0-SNAPSHOT (2.2.0p0) 2015-02-12 1c1838e Java HotSpot(TM) 64-Bit Server VM 25.5-b02 on 1.8.0_05-b13 +jit [darwin-x86_64]
```

Kagetsuki:
In order to not open a new issue I'll add here, but I've just found out jRuby is basically not meeting m17n compliance like Ruby 1.9+ does. Note:  
http://rosettacode.org/wiki/Unicode_variable_names#Ruby  

I have some code that needs to have hash keys that are UTF-8 and would love to run it on jRuby but it's a no go. Both 9k and 1.7.19 have the same issue and specifying code pages and language for Java environment fail to remedy the issue.

I'm pretty sure the root of the issue is both a mix of failure to recognize the nature of the issue (we're NOT talking about UTF-8 strings held in variables or file system paths) and perhaps unfamiliarity with the fact mainline Ruby supports UTF-8 variable/method/hash key/symbols etc. However, why the problem exists in the first place is likely due to how Java imposes the system/environment code set on whatever is being run.

headius:
This seems to be a parser issue and @enebo will probably have to tackle it. Would be nice to fix in 1.7.

@enebo: It seems like the encodings are not being handled properly for these new style keys :-(

Kagetsuki:
[...]

PR #2692, opened 2015-03-13
Fix for issue 2591 on master : double-quoted UTF8 hash key encoding

k77ch7:
This commit fixes issue #2591 on master. Double-quoted UTF8 hash key encoding should be UTF8. @enebo and @headius please review my PR. Thanks!


k77ch7:
Thanks for the review.

PR #2701, opened 2015-03-14
[Truffle] Returning nil for out of range Array#delete_at

bjfish:
Please review because I am not too familiar with contains and branch profile.

Maybe the profiles are not needed here?


chrisseaton:
Although it may look like the branch profiles are not needed because the code behind them is so simple, they are still doing something important - they are reducing the number of types that a value may be. If the compiler sees both the possibility of `nil` and say a `Fixnum`, it will have to handle both types from this point onward. If we put the `nil` behind a branch profile, the rest of the code can assume it has a `Fixnum`.

PR #2702, opened 2015-03-14
[Truffle] Refactoring Array#+ to handle more cases and adding #to_ary

bjfish:
I started to refactor this because I got the following error when adding #to_ary:

https://gist.github.com/bjfish/5c6e32592ecb1715a5fd

Please review.
Did I remove any specializations that were need?
Any other updates I should make here right now?


bjfish:
@chrisseaton did these specialization changes look alright? Thanks.

chrisseaton:
@bjfish here's a list of `Array` methods we need to get some more libraries working if you'd like to know that you're doing something that will be immediately useful https://gist.github.com/chrisseaton/bc4ceb5e1602b3f09634

PR #2710, opened 2015-03-16
[Truffle] Adding Array#slice! to array.rb

bjfish:


truffle/src/main/ruby/core/rubinius/api/shims/tuple.rb
```diff
@@ -28,6 +22,48 @@ def copy_from(other, start, length, dest)
 
 class Array
 
+  def copy_from(other, start, length, dest)
+    # Rubinius.primitive :tuple_copy_from
```
nirvdrum: Is there any reason to not implement the primitive?  At the very least, we should probably add it and just return `null` so it falls back to the Ruby code.

bjfish: @nirvdrum where are these primitives added? I'll add the null case.

Working with the array methods so far, I think there be a more general purpose primitive that we could create that could replace `copy_from`. `copy_from` for example doesn't allow growing or shrinking the array.

I think it would be something like this but in java:
```ruby
def splice_arrays( array, other, a, b, m, n)
  array[a,n] = other[m,n]
end
```

nirvdrum: We have the primitive node classes in `org.jruby.truffle.nodes.rubinius`, but surprisingly, we don't seem to have an `Array` one yet.  That's where I'd advocate putting it.

The primitive name itself forms the actual namespace, so it doesn't need to be tied to a particular class . . . that's just for organization. As an example, `Rubinius::Character` is a subclass of `String`, much like the `Tuple` and `Array` relationship.  But I recently added a "character" primitive to the `StringPrimitiveNodes` class.

bjfish:
@nirvdrum added the stubbed out primitive implementation

nirvdrum:
Thanks!

PR #2711, opened 2015-03-16
[Truffle] Moving Array#to_s to array.rb

bjfish:


truffle/src/main/ruby/core/rubinius/common/array.rb
```diff
@@ -612,6 +612,29 @@ def flatten!(level=-1)
...
+    return "[...]" if Thread.detect_recursion self do
+      each_with_index do |element, index|
+        temp = element.inspect
+        result.force_encoding(temp.encoding) if index == 0
+        result << temp << comma
+      end
+    end
+
+    Rubinius::Type.infect(result, self)
+    #result.shorten!(2) # MODIFIED We don't implement string shorten.
```
chrisseaton: Did you make this change? We'd much prefer shimming `shorten!` to do nothing rather than have this line commented out.

bjfish: @chrisseaton Yes, I made this change. I'm adding `MODIFIED` in the comments if I change things.

It looks like String#shorten! is only used in 2 places: io and here.  So, I'm a little hesitant to add it to the string public methods.

Alternatively, I could leave this as byteslice or rewrite this method a little. I think the concatenating part of this method is a little ugly. I would write it something like this:
```ruby
  def inspect
    return "[]".force_encoding(Encoding::US_ASCII) if @total == 0
    elements = Array.new(self.size + 2)    
    elements[0] = "["
    elements[self.size - 1] = "]"

    first_encoding = nil
    return "[...]" if Thread.detect_recursion self do
      each_with_index do |element, index|
        temp = element.inspect
        first_encoding = temp.encoding if index == 0
        elements[i] = temp
      end
    end
 
    result = elements.join(", ").force_encoding(first_encoding)
    Rubinius::Type.infect(result, self)
    result
  end
```
I'll do whatever you prefer: adding shorten!, leave as is, or rewriting.

chrisseaton: We'd rather you left inspect as exactly in the Rubinius code, and then implement an empty `shorten!` as a shim - like this https://github.com/jruby/jruby/blob/master/truffle/src/main/ruby/core/shims.rb#L59-L60.

It's possible that this will cause problems later when we shimmed methods that silently do nothing, but at least they're easy to search for and understand. It's worked well for us so far.

bjfish:
@chrisseaton fixed to use String#shorten!

PR #2714, opened 2015-03-17
[Truffle] Updates to Array#{map, map!}.

bjfish:


truffle/src/main/java/org/jruby/truffle/nodes/core/ArrayNodes.java
```diff
@@ -1876,20 +1944,35 @@ public RubyArray mapObject(VirtualFrame frame, RubyArray array, RubyProc block)
...
+        }
+
+        @Specialization
+        public Object eachEnumerator(VirtualFrame frame, RubyArray array, UndefinedPlaceholder block) {
+            if (toEnumNode == null) {
+                CompilerDirectives.transferToInterpreter();
+                toEnumNode = insert(DispatchHeadNodeFactory.createMethodCall(getContext()));
+            }
+
+            return toEnumNode.call(frame, array, "to_enum", null, getContext().getCoreLibrary().getMapBangSymbol());
```
chrisseaton: This is a very common pattern - maybe we should have an annotation on `@CoreMethod` called `toEnumIfNoBlock`?

truffle/src/main/java/org/jruby/truffle/nodes/core/ArrayNodes.java
```diff
@@ -1898,20 +1981,38 @@ public RubyArray mapInPlaceNull(RubyArray array, RubyProc block) {
         }
 
         @Specialization(guards = "isIntegerFixnum")
-        public RubyArray mapInPlaceFixnumInteger(VirtualFrame frame, RubyArray array, RubyProc block) {
+        public Object mapInPlaceFixnumInteger(VirtualFrame frame, RubyArray array, RubyProc block) {
+
```
chrisseaton: Please try to avoid these extra blank lines - we like them between basic blocks but not between the head of the method and the first bit of code. Sorry to be pedantic!

chrisseaton:
We've got spec failures here but they look unrelated to your work. I'll try and fix them and then re-run the specs here later.

truffle/src/main/java/org/jruby/truffle/nodes/core/ArrayNodes.java
bjfish: I see 90 occurrences of `returns an Enumerator` in the spec/ruby directory so I agree it would probably make a good annotation. I think this should be a different PR though.

chrisseaton: Yeah - do you want to take that on? If you look at the accessors of the other properties in `@CoreMethod` you should be able to find the code where they are inserted into the core method prelude.

bjfish: @chrisseaton yes, ill give it a try

bjfish:
@chrisseaton extra lines removed

PR #2715, opened 2015-03-17
implements strict base64 encoding

mkristian:
apparently the respective tests were excluded.

@headius could have a look - there is a lot of "duplication" but did not know how to avoid it. if this patch is OK I will also merge it to master.


headius:
:+1: Looks good.

PR #2720, opened 2015-03-18
[Truffle] Adding annotation to return an enumerator if no block is given to a method.

bjfish:


truffle/src/main/java/org/jruby/truffle/nodes/core/ReturnEnumeratorIfNoBlockNode.java
```diff
@@ -0,0 +1,49 @@
...
+        this.method = method;
+        this.methodName = methodName;
+    }
+
+    @Override
+    public Object execute(VirtualFrame frame) {
+        final RubyProc block = RubyArguments.getBlock(frame.getArguments());
+
+        if (block == null) {
+
```
chrisseaton: Sorry to be a pain, but we don't want a newline here...

truffle/src/main/java/org/jruby/truffle/nodes/core/ReturnEnumeratorIfNoBlockNode.java
```diff
@@ -0,0 +1,49 @@
...
+            if (toEnumNode == null) {
+                CompilerDirectives.transferToInterpreter();
+                toEnumNode = insert(DispatchHeadNodeFactory.createMethodCall(getContext()));
+            }
```
chrisseaton: ...and we do want a blank line here!

truffle/src/main/java/org/jruby/truffle/nodes/core/ReturnEnumeratorIfNoBlockNode.java
```diff
@@ -0,0 +1,49 @@
...
+            return toEnumNode.call(frame, RubyArguments.getSelf(frame.getArguments()), "to_enum", null, getContext().getSymbolTable().getSymbol(methodName));
+
+        } else {
+
```
chrisseaton: same

truffle/src/main/java/org/jruby/truffle/nodes/core/ReturnEnumeratorIfNoBlockNode.java
```diff
@@ -0,0 +1,49 @@
...
+            try {
+                return method.executeRubyBasicObject(frame);
+            } catch (UnexpectedResultException e) {
+                return e.getResult();
+            }
+
```
chrisseaton: same

truffle/src/main/java/org/jruby/truffle/nodes/core/CoreMethodNodeManager.java
```diff
@@ -210,6 +210,10 @@ private static RubyRootNode makeGenericMethod(RubyContext context, MethodDetails
         final CheckArityNode checkArity = new CheckArityNode(context, sourceSection, arity);
         RubyNode sequence = SequenceNode.sequence(context, sourceSection, checkArity, methodNode);
 
+        if (methodDetails.getMethodAnnotation().returnsEnumeratorIfNoBlock()) {
+            sequence = new ReturnEnumeratorIfNoBlockNode(methodDetails.getMethodAnnotation().names()[0], sequence);
```
chrisseaton: Using the first name here might not be correct. As it currently is, if you call `collect` without a block it will do `to_enum :map`, which isn't always going to be correct. We can handle that later, but can you put a comment in for now?

bjfish:
@chrisseaton I updated but wasn't sure if i got the adding and removing lines correct

truffle/src/main/java/org/jruby/truffle/nodes/core/ReturnEnumeratorIfNoBlockNode.java
```diff
@@ -0,0 +1,47 @@
...
+        super(method.getContext(), method.getEncapsulatingSourceSection());
+        this.method = method;
+        this.methodName = methodName;
+    }
+
+    @Override
+    public Object execute(VirtualFrame frame) {
+        final RubyProc block = RubyArguments.getBlock(frame.getArguments());
+
+        if (block == null) {
```
nirvdrum: I think we want a `ConditionProfile` here.  We've essentially collapsed two specializations into one to consolidate a common pattern.  But a given call site is apt to only use one form or the other.

bjfish:
@chrisseaton it would be nice if the style things could be checked with something like checkstyle

Also, I have had trouble using raiseIfFrozen or the IsFrozenNode in a method after the method is to_enum'ed. Any ideas? I might try to debug this later on.

truffle/src/main/java/org/jruby/truffle/nodes/core/ReturnEnumeratorIfNoBlockNode.java
```diff
@@ -0,0 +1,47 @@
...
+            if (toEnumNode == null) {
+                CompilerDirectives.transferToInterpreter();
+                toEnumNode = insert(DispatchHeadNodeFactory.createMethodCall(getContext()));
+            }
+
+            return toEnumNode.call(frame, RubyArguments.getSelf(frame.getArguments()), "to_enum", null, getContext().getSymbolTable().getSymbol(methodName));
+
+        } else {
+            try {
+                return method.executeRubyBasicObject(frame);
```
nirvdrum: This should probably be `execute` instead of `executeRubyBasicObject`.  I'm not sure how often it happens (maybe never), but this wrapper won't work if the specialization returns a primitive.

chrisseaton: Yes, sorry these notes are both correct and important but I already merged - can submit another PR?

bjfish: I'll submit a new PR

PR #2724, opened 2015-03-18
Socket.getaddrinfo reverse_lookup

kpamidon:
Added the reverse_lookup parameter to Socket.getaddrinfo. This was added in Ruby 1.9.

### DIFF
--- a/core/src/main/java/org/jruby/ext/socket/RubySocket.java
+++ b/core/src/main/java/org/jruby/ext/socket/RubySocket.java
@@ -287,7 +287,7 @@ public class RubySocket extends RubyBasicSocket {
         return SocketUtils.gethostbyname(context, hostname);
     }
 
-    @JRubyMethod(required = 2, optional = 4, meta = true)
+    @JRubyMethod(required = 2, optional = 5, meta = true)
     public static IRubyObject getaddrinfo(ThreadContext context, IRubyObject recv, IRubyObject[] args) {
         return SocketUtils.getaddrinfo(context, args);
     }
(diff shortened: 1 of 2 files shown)


headius:
This looks good. Maybe we can modify it to use Boolean instead of boolean and just have a separate method with no boolean parameter for the "null" behavior you have now?

kpamidon:
I'm not quite sure how to completely get rid of the Boolean, since the current code relies on the `buildAddrinfoList()` function to process the args array before sending the processed args on to the `AddrinfoCabllack()` function.

If the concern is over having the `getHostAddress(ThreadContext context, InetAddress addr)` function call the `getHostAddress(ThreadContext context, InetAddress addr, boolean reverse)` function and pass in a null for the last parameter, I can do what you suggest and leave the original `getHostAddress(ThreadContext context, InetAddress addr)` function alone; however, I can't see a way to get around then having to do something ugly like the following inside the `AddrInfoCallback()` to select the appropriate function call:
```
                    if (reverse == null) {
                        c[2] = runtime.newString(getHostAddress(context, address));
                    } else {
                        c[2] = runtime.newString(getHostAddress(context, address, reverse));
                    }
```

kpamidon:
I noticed I've made a serious error in doing a String comparison using the == operator in my initial Pull Request. I'm new to the Git paradigm; is the proper procedure here to make a brand new Pull Request, with the fixed code and close this one?

atambo:
Just amend your fix to your existing commit and do a force push to your socket_getaddrinfo_reverse_lookup branch and it will update on this pull request.

kpamidon:
Amending and force pushing worked like a charm, thanks.

PR #2725, opened 2015-03-19
[Truffle] Adding args to to_enum call in returnsEnumeratorIfNoBlock annotation.

bjfish:


truffle/src/main/java/org/jruby/truffle/nodes/core/ReturnEnumeratorIfNoBlockNode.java
```diff
@@ -41,4 +43,12 @@ public Object execute(VirtualFrame frame) {
         }
     }
 
+    private static <T> T[] unshift(T first, T[] array) {
```
bjfish: I think this could be done with less copying if a helper method was added to RubyArguments that unshifted this first argument when the user arguments are being extracted.

chrisseaton: Yeah - feel free to move this helper into RubyArguments and make something like `extractUserArgumentsWithUnshift`.

PR #2737, opened 2015-03-20
[Truffle] Implementing more of Array#pop.

bjfish:


bjfish:
@nirvdrum @chrisseaton  Is this much code repetition normal to get things implemented right now?

nirvdrum:
Array is a bit odd in having so many specialized stores, but looking at what you have, a fair bit could be pushed to a private helper method.

chrisseaton:
Yeah - the duplication here is pretty bad. What I think we will do is allocate an interface that hides the type, inside each specialisation, and hand that to a helper method. With PE the interface will compile away and we should get the same code. However I'd rather do a little bit of repetitive messy code while we work through the specs, and then at the end step back and think about what to de-duplicate.

truffle/src/main/java/org/jruby/truffle/nodes/core/ArrayNodes.java
```diff
@@ -2661,6 +2664,321 @@ public Object popObject(RubyArray array) {
...
+                final RubyArray result = new RubyArray(getContext().getCoreLibrary().getArrayClass(), Arrays.copyOfRange(store, array.getSize() - numPop , array.getSize()), numPop);
+                final long[] filler = new long[numPop];
+                System.arraycopy(filler, 0, store, array.getSize() - numPop, numPop);
+                array.setStore(store, array.getSize() - numPop);
+                return result;            }
+        }
+
+        @Specialization(guards = "isFloat", rewriteOn = UnexpectedResultException.class)
+        public RubyArray popFloatInBoundsWithNum(VirtualFrame frame, RubyArray array, int num) throws UnexpectedResultException {
+            if(num < 0){
```
chrisseaton: Really sorry to be so pedantic, but this should be `if (num < 0) {`, with the spaces in there.

bjfish: @chrisseaton Fixed it at 412f0ad8b98f6f1827926c5486aa0f56865a6fd8

PR #2745, opened 2015-03-23
[Truffle] Moving Array#zip to array.rb.

bjfish:
Could zip be done in ruby? How about insert too?

### DIFF
--- a/spec/truffle/tags/core/array/zip_tags.txt
+++ b/spec/truffle/tags/core/array/zip_tags.txt
@@ -1,6 +1,0 @@
-fails:Array#zip returns an array of arrays containing corresponding elements of each array
-fails:Array#zip fills in missing values with nil
-fails:Array#zip properly handles recursive arrays
-fails:Array#zip calls #to_ary to convert the argument to an Array
-fails:Array#zip uses #each to extract arguments' elements when #to_ary fails
-fails:Array#zip calls block if supplied
--- a/truffle/src/main/java/org/jruby/truffle/nodes/core/ArrayNodes.java
+++ b/truffle/src/main/java/org/jruby/truffle/nodes/core/ArrayNodes.java
@@ -4470,53 +4470,5 @@ public abstract class ArrayNodes {
 
     }
 
-    @CoreMethod(names = "zip", required = 1)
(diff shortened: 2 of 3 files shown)


chrisseaton:
`zip` is used in the inner loop of the `topaz-neural-net` benchmark https://github.com/jruby/bench9000/blob/master/lib/bench/benchmarks/topaz/neural-net.rb#L151, so needs to be fast.

You can see if this commit damages the performance of that benchmark by running `jt bench reference topaz-neural-net` with the old code, make your changes and then run `jt bench compare topaz-neural-net`. The margin for error is around ±10%, so maybe run it  a few times if it's closed. I'm going to guess it isn't close and the current code is much faster.

The same thing applies for `insert` - it's used by `classic-fannkuch-redux` so test in the same way.

We can't accept any code that damages the performance of these benchmarks, unless it's fixing an error in our implementation of Ruby. The options are to complete the implementation in Java, or find some way to run Java for the cases we need to be fast, and fallback to Ruby for everything else.

bjfish:
@chrisseaton  I'll check out the benchmarks, I got `jt bench `to run now, my problem was having GRAAL_BIN set to the graal bin directory and not the Graal java binary.

bjfish:
@chrisseaton 
Any idea how to resolve this warning below?
Also, should I be comparing the `4902.008372064519` or  `85.107286` or both? and is smaller or bigger better for the first value?
```
warning: jruby-9000-dev-truffle-graal topaz-neural-net never warmed up!
topaz-neural-net 4902.008372064519
took 85.107286s
```

chrisseaton:
The warning is ok to ignore here. It's just telling you that it has erred on the side of caution because it didn't see the benchmark performance stabilise as much as it would have hoped - that's also why it's run for so long - 85s.

chrisseaton:
You don't need to look at either of those numbers. `reference` stores the score in a file. When you run `compare` it will give you a % performance difference.

bjfish:
It was running 9.41% in the compare.

:warning: This isn't safe to merge because of a Graal error I need to fix in AtNode.

bjfish:
Closing due to performance regression

PR #2756, opened 2015-03-25
Allow File.open and File.new with 4 arguments

lumeet:
The methods now support filename, mode, permissions and options as
separate parameters.

The actual logic seems to be there already. For example, the following now has the same behavior as MRI 1.9+.
```ruby
File.open('test', 'w', 0777, cr_newline: true) { |f| f.puts 'test' }
```

### DIFF
--- a/core/src/main/java/org/jruby/RubyFile.java
+++ b/core/src/main/java/org/jruby/RubyFile.java
@@ -321,7 +321,7 @@ public class RubyFile extends RubyIO implements EncodingCapable {
     }
 
     // rb_file_initialize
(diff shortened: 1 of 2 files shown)


lumeet:
Ok, not quite then. I'll take a look at it again.

enebo:
@lumeet This removed convertToHash which will try and call to_hash if it is not an explicit Ruby Hash.  I think perhaps tests are working but I think this will now break:

```ruby
class Foo; def to_hash; {cr_newline: true }; end; end
File.open('test', 'w', 0777, Foo.new) { |f| f.puts 'test' }
```

lumeet:
@enebo It should be better now. Hopefully I haven't forgotten anything.

PR #2758, opened 2015-03-25
[Truffle] Implementing more of Array#{uniq,uniq!}

bjfish:


chrisseaton:
Could you express `uniq` in terms of `uniq!` in Ruby code, or vice-versa? Was one of them in Java already because some specialisation is exercised in benchmarks? If so could the other one be done in Ruby?

bjfish:
@chrisseaton it looks like `uniq` could be:
```
  def uniq(&block)
    dup.uniq!(&block) or dup
  end
```
I'll give this a try.

I previously tried the ruby `uniq!` and it didn't work as is right now. Probably some issue in IdentityMap.

It looks like the benchmarks use `uniq` in some places but it was never compiled because this was marked as `notDesignedForCompilation();`.

bjfish:
@chrisseaton `uniq` worked in ruby in terms of `uniq!`. I've updated the PR.

chrisseaton:
Try the whole thing as Rubinius if you can - it's quite a lot of Java code. You can look at what the benchmarks do by just searching in the `lib/benchmarks` directory - they don't use stdlib so it's all in there.

bjfish:
@chrisseaton I've updated this with the Rubinius verision with 2 added lines to modify the `array.store`

PR #2763, opened 2015-03-26
[Truffle] Adding more implementation of Array#initialize.

bjfish:


truffle/src/main/java/org/jruby/truffle/nodes/core/ArrayNodes.java
```diff
@@ -1648,6 +1702,10 @@ public RubyArray initialize(RubyArray array, int size, long defaultValue, Undefi
 
         @Specialization
         public RubyArray initialize(RubyArray array, int size, double defaultValue, UndefinedPlaceholder block) {
+            if (size < 0) {
```
chrisseaton: If you made an `isPositive` guard on all these specialisations, then the exception case could be a separate specialisation.

PR #2765, opened 2015-03-27
Truffle encoding

nirvdrum:
This is a bit large and I had to do some interesting things to init Rubinius properly.  I'd appreciate another set of eyes before I merge it.  @chrisseaton @eregon


truffle/src/main/java/org/jruby/truffle/nodes/core/EncodingNodes.java
```diff
@@ -501,48 +339,144 @@ public boolean isDummy(RubyEncoding encoding) {
...
+
+            final Encoding filesystemEncoding = getContext().getRuntime().getEncodingService().getLocaleEncoding();
+            final Object filesystemTuple = newTupleNode.call(
+                    frame,
+                    getContext().getCoreLibrary().getTupleClass(),
+                    "create",
+                    null,
+                    getContext().makeString("filesystem"),
+                    indexLookup(encodings, filesystemEncoding)
+            );
```
chrisseaton: Can you create some kind of `createTuple` helper here?

truffle/src/main/java/org/jruby/truffle/runtime/core/RubyEncoding.java
```diff
@@ -102,4 +102,9 @@ public RubyBasicObject allocate(RubyContext context, RubyClass rubyClass, Node c
 
     }
 
-}
\ No newline at end of file
+    @Override
+    public String toString() {
+        return getName().toString();
```
chrisseaton: Is this for debugging only?

nirvdrum: Yeah. If there's a better option, let me know. But I was looking at an array of 100 of them in the debugger and having the `toString` made it much easier to work with.

truffle/src/main/ruby/core/shims.rb
```diff
@@ -200,3 +200,35 @@ class IO
...
+    end
+
+    alias_method :default_internal_rubinius=, :default_internal=
+
+    def default_internal=(enc)
+      self.default_internal_rubinius = enc
+      self.default_internal_jruby = enc
+    end
+  end
+end
```
eregon: I like this

truffle/src/main/java/org/jruby/truffle/nodes/core/EncodingConverterNodes.java
```diff
@@ -11,72 +11,35 @@
...
-            }
-
-            return new RubyArray(getContext().getCoreLibrary().getArrayClass(), result, result.length);
-        }
-
-    }
-
-    @CoreMethod(names = "initialize", required = 2)
+    @RubiniusOnly
+    @CoreMethod(names = "initialize_jruby", required = 2, optional = 1)
```
eregon: make it private :wink: initialize is implicitly private but not when suffixed of course.

truffle/src/main/ruby/core/rubinius/api/shims/encoding.rb
```diff
@@ -0,0 +1,23 @@
...
+Encoding::TranscodingMap[:'UTF-16LE'][:'UTF-8'] = nil
+
+Encoding::TranscodingMap[:'UTF-32BE'] = Rubinius::LookupTable.new
+Encoding::TranscodingMap[:'UTF-32BE'][:'UTF-8'] = nil
+
+Encoding::TranscodingMap[:'UTF-32LE'] = Rubinius::LookupTable.new
+Encoding::TranscodingMap[:'UTF-32LE'][:'UTF-8'] = nil
+
+Encoding::TranscodingMap[:'ISO-2022-JP'] = Rubinius::LookupTable.new
+Encoding::TranscodingMap[:'ISO-2022-JP'][:'STATELESS-ISO-2022-JP'] = nil
```
eregon: Did you find out what uses these constants early? Is it `Encoding.default_external` or related in start-up code?

nirvdrum: One case was `Encoding::Converter::TranscodingPath` referencing `Encoding::TranscodingMap` at the top-level of its class definition.

eregon:
`truffle/src/main/ruby/core/rubinius/delta/converter_paths.rb` does not look pretty with its 7000+ assignments to a Hash. Do we have this map already somewhere in JRuby?
Also most of them are "go through UTF-8", not really sure of the gain to save that path in memory.

truffle/src/main/ruby/core/rubinius/common/encoding.rb
```diff
@@ -0,0 +1,614 @@
...
+      str = StringValue(str)
+
+      @replacement = str.encode(@destination_encoding)
+    end
+
+    class TranscodingPath
+      @paths = {}
+      @load_cache = true
+      @cache_valid = false
+      @transcoders_count = TranscodingMap.size
```
eregon: Ah here, I see. Could be worth to make a private class method `setup` that initialize to non-trivial values. But probably good enough for now.

nirvdrum: I was trying to avoid modifying their source altogether since it would make future updates easier.  But I'm open to other options.

eregon: Yeah, this is good for now I think.

eregon:
Startup suffers quite a bit though ...
```
$ for i in {1..10}; do; jt run -e 'puts :hi'; done
Before:
  4.46s user 0.14s system 160% cpu 2.872 total
  4.40s user 0.14s system 159% cpu 2.845 total
  4.45s user 0.13s system 159% cpu 2.875 total
After:
  10.18s user 0.22s system 169% cpu 6.145 total
  10.18s user 0.22s system 164% cpu 6.308 total
  9.91s user 0.24s system 164% cpu 6.166 total
```

chrisseaton:
Wow! Normally I'd say not to worry about it as it's not peak performance - but 10s is a bit insane. We risk people trying Truffle out getting a very bad impression. It might be the hash operations - we never resize remember - so I'll see if I can optimise the methods and specialisations used in `converter_paths.rb`.

nirvdrum:
It looks like that converter_paths load is to blame after all.  I'll work on fixing that now.  While big, it's a bit disconcerting that interpreted hash updates are that slow.  That might warrant additional investigation.

eregon:
Just parsing that file might be slow, Rbx precompiles to .rbc so that might motivate their approach.

eregon:
@nirvdrum Seems much better after 53683ba04b650253df58d0960d88c6471f3a1d2e:
```
  4.74s user 0.14s system 163% cpu 2.990 total
  4.68s user 0.15s system 161% cpu 2.999 total
  4.73s user 0.13s system 159% cpu 3.042 total
```

PR #2780, opened 2015-03-30
[Truffle] Don't taint from self in String#[](String).

nirvdrum:
@chrisseaton @eregon This is one of the proposed solutions for dealing with the need to skip tainting for certain specializations.  I don't particularly like using exceptions this way, but using a wrapper object would likely add much greater overhead.  The third alternative is to not use the CoreMethod option for tainting at all, but rather have each specialization that needs it use the `TaintResultNode` in place; this seems both error-prone (easy to forget one) and verbose.

I'm doing this as a PR to get feedback.  No hard feelings if you think this is terrible.

### DIFF
--- a/spec/truffle/tags/core/string/element_reference_tags.txt
+++ b/spec/truffle/tags/core/string/element_reference_tags.txt
@@ -1,4 +1,3 @@
 fails:String#[] with Range calls to_int on range arguments
 fails:String#[] with Range works with Range subclasses
 fails:String#[] with Regexp, index returns nil if there is no capture for the given index
-fails:String#[] with String taints resulting strings when other is tainted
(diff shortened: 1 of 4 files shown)


chrisseaton:
You could parameterise the exception - `DoNotTaint<RubyString>` - won't add anything but looks neater I think.

I don't mind the exception - they compile to a jump and using exception is idiomatic Truffle - unlike in normal Java code so no need to feel dirty. I guess an exception also simplifies the interaction with the DSL?

I think a wrapper object would also be fine.

eregon:
Looks good to me as the exception is contained in the TaintResultNode.

nirvdrum:
This was news to me, but generic classes can't extend java.lang.Throwable.

chrisseaton:
Ah yeah, I have come across that before. nm

PR #2782, opened 2015-03-30
[Truffle] Adding overflow specializations for Fixnum#abs.

### LINKED ISSUE #2781: [Truffle] Expected Absolute Value for Fixnum
Code Example:
```ruby
puts (-2**63).abs
```

Expected:
```
9223372036854775808
```
Actual
```
-9223372036854775808
```

bjfish:
Fixes #2781

### DIFF
--- a/truffle/src/main/java/org/jruby/truffle/nodes/core/FixnumNodes.java
+++ b/truffle/src/main/java/org/jruby/truffle/nodes/core/FixnumNodes.java
@@ -1395,7 +1395,7 @@ public abstract class FixnumNodes {
 
     }
 
-    @CoreMethod(names = { "abs", "magnitude" })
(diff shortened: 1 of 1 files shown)


chrisseaton:
Hmm... I'm not sure this is the best approach. You have a compare, and then a jump to deoptimize in the fastest case. A single `jo` to deoptimize would be tighter. The code for `Math.abs` is:

```java
    public static int abs(int a) {
        return (a < 0) ? -a : a;
    }
```

Perhaps we'd be better off writing that inline, and then just replacing the `-` with `ExactMath.subtract`?

bjfish:
@chrisseaton updated to use ExactMath

chrisseaton:
Thanks - take a look at `jt --asm` of the two to see why it's better this way.

PR #2789, opened 2015-03-31
[Truffle] Implementing Range in ruby.

bjfish:
Could range be done in ruby? 

There were two language specs that failed that I wasn’t sure how to fix.


chrisseaton:
At least `Range#inject`, `#map`, `#each`, `#to_a` and `#collect` are used in the fast path of benchmarks, so it will be tricky to get this right.

Try benchmarking this on matrix-multiply and spectral-norm to start with.

bjfish:
Reverted to some of the Java specializations for benchmarks. Looks like about 100% for matrix-multiply and 150% for spectral-norm (not sure why there is a speed up). The failing language specs now pass.

chrisseaton:
Sorry, this caused regressions on psd-imagemode-greyscale-combine-greyscale-channel and chunky-canvas-resampling-stpes-residues. Can you take a look at those? They must use something else that still needs to be in Java.

bjfish:
@chrisseaton The benchmark should be fixed here: https://github.com/jruby/jruby/commit/dc85f30e6b141f121bd46e6ef019e4d17d5ec983

Please give it a try when you can. Thanks.

chrisseaton:
Yeah that fixed it - thanks again.

PR #2797, opened 2015-04-02
Support respond_to_missing? in respondsTo() lookups

cheald:
RubyBasicObject assumes that if respond_to? is undefined, it can just fast-track
method lookups. We have to also check for respond_to_missing?; failure to do so causes
Javaland respondsTo() calls to fail when an object defines respond_to_missing? (despite
the documentation's admonishment to not use it)

The practical case for this is ActiveSupport::Duration in ActiveSupport 4.2.1

### DIFF
--- a/core/src/main/java/org/jruby/RubyBasicObject.java
+++ b/core/src/main/java/org/jruby/RubyBasicObject.java
@@ -597,19 +597,27 @@ public class RubyBasicObject implements Cloneable, IRubyObject, Serializable, Co
             // fastest path; builtin respond_to? which just does isMethodBound
             return getMetaClass().isMethodBound(name, false);
         } else if (!(respondTo.isUndefined() && respondToMissing.isUndefined())) {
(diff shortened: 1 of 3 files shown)


enebo:
@cheald close but it triggered at least one error:
```text
     [exec]   1) Failure:
     [exec] TestObject#test_implicit_respond_to_arity_3 [/home/travis/build/jruby/jruby/test/mri/ruby/test_object.rb:521]:
     [exec] [bug:6000].
     [exec] Expected Exception(ArgumentError) was raised, but the message doesn't match.
     [exec] <"respond_to? must accept 1 or 2 arguments (requires 3)"> expected but was
     [exec] <"null must accept 1 or 2 arguments (requires 3)">.
     [exec] 
     [exec] 6445 tests, 1100846 assertions, 1 failures, 0 errors, 16 skips
```

2 jobs failed so you should review that other job as well.

cheald:
Whoops. I shouldn't commit when tired. Fixing up.

PR #2806, opened 2015-04-06
[Truffle] Move Float#round to float.rb

bjfish:
I don’t believe this is used in any benchmarks but wasn’t 100% sure.


chrisseaton:
Can you try `chunky-canvas-resampling-steps-residues` and `chunky-canvas-resampling-steps` - they both use `#round`. Then also `chunky-canvas-resampling-nearest-neighbor` and `chunky-canvas-resampling-bilinear`, which use those methods.

bjfish:
@chrisseaton  The first benchmark I tried had regressed to 20% `chunky-canvas-resampling-steps-residues` so I added back the java specialization and a fallback to ruby specialization which brought the benchmark back to 100%.  I didn't try the other three benchmarks but I assumed they would be okay if the first regression was fixed.

PR #2807, opened 2015-04-06
New readme

chrisseaton:
I thought I could make some positive changes to the readme:

* Talk about it being a JVM implementation rather than a Java implementation
* Building instructions are all available in BUILDING.md so just point people there
* Provide instructions on how to install without building from source
* Warn people against using old versions from their package manager
* Add some positive statements about what we are trying to achieve
* Remove some dead links

What do people think?

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,9 +1,56 @@
-# JRuby -  An implementation of the Ruby language on the JVM
+# JRuby - an implementation of the Ruby language on the JVM
 
 Master: [![Build Status](https://travis-ci.org/jruby/jruby.png?branch=master)](https://travis-ci.org/jruby/jruby) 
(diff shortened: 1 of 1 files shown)


cheald:
First off: It's strangely hilarious to me that this failed the CI build :)

That said, I like the direction. I might almost go a step further and go out of the way to highlight some of the reasons that you might want to use JRuby over MRI, rather than just casually mentioning them in the "about" section. The README is what people see when they first encounter the project, so it should really be half information, half marketing copy, IMHO. 

I would also include information on bug reporting procedures, pull request protocol, and maybe even a link to the mailing list and IRC channel.  If someone has a problem with it, they shouldn't need to go further than the README to figure out how to get additional help.

chrisseaton:
Yeah, introducing some form of marketing copy was my goal. I think the current version isn't confident enough, if you see what I mean.

kares:
:+1:

donv:
:+1:

headius:
This was in pre2 bundle, so marking for pre2.

PR #2808, opened 2015-04-07
[Truffle] Adding fallback Specialization for Fixnum#<.

bjfish:
For example, `jruby -X+T -e 'puts 1 < Rational(“1.3”)’`


chrisseaton:
Can't we just do `ruby("b, a = math_coerce other, :compare_error; a < b")` - the separate fallback code and the extra primitive (which always just fails) seem like a lot of extra work for this.

bjfish:
@chrisseaton Yes, I've updated the PR.

PR #2811, opened 2015-04-08
[Truffle] Start to support the Rubinius FFI and POSIX calls

chrisseaton:
This is the start of the work that will enable us to use the Rubinius implementation of File, IO and Socket, which are large areas of missing functionality.

@thomaswue @nirvdrum @eregon


truffle/src/main/ruby/core/rubinius/common/file.rb
```diff
@@ -0,0 +1,126 @@
...
+  end
+
+  ##
+  # Deletes the named files, returning the number of names
+  # passed as arguments. Raises an exception on any error.
+  #
+  # See also Dir::rmdir.
+  def self.unlink(*paths)
+    paths.each do |path|
+      n = POSIX.unlink Rubinius::Type.coerce_to_path(path)
```
chrisseaton: We can call simple POSIX functions fairly directly.

truffle/src/main/ruby/core/rubinius/common/process.rb
```diff
@@ -28,8 +28,28 @@
...
+      @maxgroups = m
+    end
+  end
+
+  def self.groups
+    g = []
+    FFI::MemoryPointer.new(:int, @maxgroups) { |p|
+      num_groups = FFI::Platform::POSIX.getgroups(@maxgroups, p)
+      Errno.handle if num_groups == -1
+      g = p.read_array_of_int(num_groups)
```
chrisseaton: More complex POSIX functions that do things like write into a pointer you pass them are also supported, and then end up allocating native heap using `Unsafe`.

nirvdrum:
I haven't yet tried to run any of this, but my immediate concern is breaking Windows.  While there's an awful lot we don't yet fully support, JRuby+Truffle is substantially usable on Windows today.  I think jnr-posix will get us most of the way there and it may turn out we need to fill in some Win32 gaps there, but I wanted to raise the topic now before we get too far.  I'll try running the branch in Windows tomorrow to see where we're at.

truffle/src/main/ruby/core/rubinius/common/process.rb
nirvdrum: Do we only care about OpenJDK if not running in Graal?  JRuby generally has fallbacks if `Unsafe` isn't available.

truffle/src/main/ruby/core/rubinius/platform/ffi.rb
```diff
@@ -0,0 +1,287 @@
...
+
+  TypeSizes = Rubinius::LookupTable.new
+  TypeSizes[1] = :char
+  TypeSizes[2] = :short
+  TypeSizes[4] = :int
+  TypeSizes[8] = Rubinius::L64 ? :long : :long_long
+
+  # Load all the platform dependent types
+
+  # MODIFIED
```
nirvdrum: If we need to diverge from Rubinius, it'd probably be best to document why.

truffle/src/main/ruby/core/rubinius/common/process.rb
eregon: Which JRE do not have Unsafe?

eregon:
I commented per commit as this was easier.
Looks good, I'll merge it!

eregon:
I'll let you merge, in case you want toa dd additional commits before the merge.

chrisseaton:
@nirvdrum we're using jnr-posix to make these calls, and there is a Windows and generic Java implementation of most of them. For example `unlink` simply does `File.delete` on a JVM without JNI or on Windows. I hope that covers us for all the POSIX calls that make sense. I guess things like `lstat` aren't implemented for non-POSIX systems, but then what would it do on Windows anyway.

truffle/src/main/ruby/core/rubinius/common/process.rb
chrisseaton: I think Truffle requires it for the object model anyway, so the decision has been made for us.

truffle/src/main/ruby/core/rubinius/platform/ffi.rb
chrisseaton: Yes sorry I must have been in the middle of a thought here ands stopped mid-comment.

PR #2819, opened 2015-04-10
[Truffle] Add jt test mri command.

bjfish:
This is just an idea to add a command to run mri tests.

The big args array was derived from https://travis-ci.org/jruby/jruby/jobs/57843253

### DIFF
--- a/tool/jt.rb
+++ b/tool/jt.rb
@@ -149,6 +149,7 @@ module Commands
     puts '    --asm          show assembly (implies --graal)'
     puts '    --server       run an instrumentation server on port 8080'
     puts '    --igv          make sure IGV is running and dump Graal graphs after partial escape (implies --graal)'
+    puts 'jt test mri                                  run mri tests'
     puts 'jt test                                      run all specs'
     puts 'jt test fast                                 run all specs except sub-processes, GC, sleep, ...'
     puts 'jt test spec/ruby/language                   run specs in this directory'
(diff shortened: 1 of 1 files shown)


chrisseaton:
This is definitely what we want to do next, but I've got quite a few comments on how we want to do this:

* I'd like it to be under the `jt test` command, so it should be `jt test mri`.
* I'd like it to be able to specify a directory or a single file, so I can do `jt test mri test/mri/ruby/test_whileuntil.rb` or `jt test mri test/mri/ruby/test_whileuntil.rb`
* You're re-using JRuby's excludes there, which doesn't make any sense - can you create a full set of excludes for Truffle under `test/truffle/excludes` (may need to talk to @headius) about how to do that
* You might want to modify `jt tag`, `jt untag`, and `jt tag all` to work with these excludes if they see a MRI test path rather than a RubySpec path
* Do these tests actually work for us? Last time I checked we didn't run `test/unit` very well - if we do now that's excellent news
* When it works and we have everything that doesn't pass excluded, we should add it to Travis

This patch would still be useful with the first two points addressed.

bjfish:
@chrisseaton The first two points work now, `jt test mri` and `jt test mri test/mri/ruby/test_whileuntil.rb`
(I tested the command without `-X+T`).

I've ran a little bit of test/unit successfully. The first failure we'll run into here is:
```
/Users/brandonfish/Documents/jruby-mine/test/mri/lib/minitest/unit.rb:1389:in `(singleton-def)': undefined method `method_added' for #<Class:0x114> (NoMethodError)
	from /Users/brandonfish/Documents/jruby-mine/test/mri/lib/minitest/unit.rb:1388:in `TestCase'
	from /Users/brandonfish/Documents/jruby-mine/test/mri/lib/minitest/unit.rb:1242:in `Unit'
	from /Users/brandonfish/Documents/jruby-mine/test/mri/lib/minitest/unit.rb:746:in `MiniTest'
	from /Users/brandonfish/Documents/jruby-mine/test/mri/lib/minitest/unit.rb:12:in `<main>'
	from /Users/brandonfish/Documents/jruby-mine/test/mri/lib/test/unit.rb:5:in `<main>'
	from test/mri/runner.rb:7:in `<main>'
```

chrisseaton:
Great - I'll merge after pre2 is out.

tool/jt.rb
```diff
@@ -232,9 +233,64 @@ def run(*args)
...
+                 psych/test_tainted.rb psych/test_to_yaml_properties.rb psych/test_tree_builder.rb psych/test_yaml.rb psych/test_yamlstore.rb psych/json/test_stream.rb
+                 psych/nodes/test_enumerable.rb psych/visitors/test_depth_first.rb psych/visitors/test_emitter.rb psych/visitors/test_to_ruby.rb psych/visitors/test_yaml_tree.rb
+                 ripper/test_files.rb ripper/test_filter.rb ripper/test_parser_events.rb ripper/test_ripper.rb ripper/test_scanner_events.rb scanf/test_scanf.rb
+                 scanf/test_scanfblocks.rb socket/test_basicsocket.rb socket/test_nonblock.rb socket/test_socket.rb stringio/test_stringio.rb strscan/test_stringscanner.rb
+                 thread/test_queue.rb zlib/test_zlib.rb ruby/enc/test_big5.rb ruby/enc/test_cp949.rb ruby/enc/test_emoji.rb ruby/enc/test_euc_jp.rb ruby/enc/test_euc_kr.rb
+                 ruby/enc/test_euc_tw.rb ruby/enc/test_gb18030.rb ruby/enc/test_gbk.rb ruby/enc/test_iso_8859.rb ruby/enc/test_koi8.rb ruby/enc/test_shift_jis.rb ruby/enc/test_utf16.rb
+                 ruby/enc/test_utf32.rb ruby/enc/test_windows_1251.rb]
```
eregon: Large literals is almost never a good idea. Especially in a script file.

chrisseaton: Yes what we really want to do is add support for running `jt test ruby/**/test_.rb` or something like that.

eregon: Or do a proper Dir.glob or have a config file ala MSpec.

bjfish: @eregon  I think I tracked down how this is generated
```
./bin/jruby -X+T -e 'mri_test_files = File.readlines("test/mri.index").grep(/^[^#]\w+/).map(&:chomp).join(" ");puts mri_test_files'
```

eregon: Sweet, that might be a good replacement (we should have our own index at some point).

bjfish: @eregon BTW there are some workarounds required to apply in this patch if you want to try to use this command: https://gist.github.com/bjfish/1be84623da4a746f4d16

eregon: Yeah I saw that :)
Patching jruby-core files is not an option, so either we should have shims (for Process.waitall for instance) or we should implement the subset needed for the functionality.

PR #2822, opened 2015-04-12
[Truffle] Add pathname to the stdlib.

bjfish:


chrisseaton:
Just waiting for pre2 still.

I was going to say I'm so happy that all but two of the specs pass, but then I looked and there isn't a lot of spec coverage here anyway. Hopefully the MRI test suite will help us more here.

eregon:
Not so much in MRI tests either (except for pure string manipulation), and we would probably fail a lot since this is filesystem-related for many things.
I once wrote a pathname library with a pretty decent test coverage (also revealing alternative impls fail many of these):
https://github.com/eregon/path
I might integrate those into RubySpec one day.

PR #2829, opened 2015-04-13
[Truffle] Add a string specialization to Module#alias_method.

bjfish:


nirvdrum:
I suspect we'll need to deal with various combinations of these at some point.  And maybe ever coercible types.  But looks good.

nirvdrum:
If we do need to coerce, `SymbolOrToStrNode` might be a good starting point.

PR #2830, opened 2015-04-13
[Truffle] Move Regexp#=== to regexp.rb.

bjfish:
Is there a case where this is true?


nirvdrum:
It looks like Rubinius does a check on whether the object can coerce to a string and if not, returns false:

https://github.com/rubinius/rubinius/blob/v2.4.1/kernel/common/regexp.rb#L192-L198

PR #2834, opened 2015-04-14
[Truffle] Adding Long Integer specialization for Fixnum coerce

bjfish:


bjfish:
@nirvdrum Does this look right?

eregon:
Looks good!

nirvdrum:
This looks good.  You might want to handle the `BigInteger` case as well just to round it out.  Rubinius handles `Fixnum` and `Bignum`.  I don't see where the `String` and coercible variants come from.  I'll have to ask @chrisseaton about that.

eregon:
Yes, there should really not be a `String` case, but maybe Rubinius code requires it?

eregon:
Removed the `String` and `to_f` cases, they are handled by `Numeric#coerce`.

eregon:
And Bignum is not handled specially in MRI, so let's keep it as is.
```ruby
> 2.coerce (1<<75)
=> [3.777893186295716e+22, 2.0]
```

PR #2845, opened 2015-04-17
[Truffle] On demand freeze node

eregon:
@chrisseaton @nirvdrum What do you think?
Sounds very expensive?

The parent is needed otherwise replace cannot work correctly.
One alternative I see is to have a uber DebugNode stored in RubyContext with all subnodes we need. Of course that's profile pollution, but is it worse than duplicating logic and doing all these instanceof?

### DIFF
--- a/truffle/src/main/java/org/jruby/truffle/runtime/DebugOperations.java
+++ b/truffle/src/main/java/org/jruby/truffle/runtime/DebugOperations.java
@@ -23,10 +23,12 @@ import com.oracle.truffle.api.object.Location;
 import com.oracle.truffle.api.object.Property;
 import com.oracle.truffle.api.object.Shape;
 
+import org.jruby.truffle.nodes.objects.FreezeNode;
+import org.jruby.truffle.nodes.objects.FreezeNodeFactory;
+import org.jruby.truffle.nodes.objects.IsFrozenNode;
+import org.jruby.truffle.nodes.objects.IsFrozenNodeFactory;
 import org.jruby.truffle.runtime.backtrace.Backtrace;
 import org.jruby.truffle.runtime.core.RubyBasicObject;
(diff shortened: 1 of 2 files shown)


chrisseaton:
Seems fine as these are slow operations anyway.

Priority over making this fast is to make it go away by removing use of debug operations.

eregon:
The SourceSection is not handled correctly here, but it should not matter as these don't throw exceptions.

eregon:
@chrisseaton That is unfortunately not trivial for getSingletonClass().

nirvdrum:
I'm a fan of having a scratch space of sorts where we can attach and remove nodes as needed.  I guess that's basically the DebugNode idea, although even more on-demand.  Otherwise, this is fine too.

Since you're already using the abstract method form, couldn't you just make the node a field in DebugOperations?  Not a big deal either way, but it could reduce object churn and construction time.

chrisseaton:
Could we also do `ruby("o.freeze", "o", o)` - that will be even slower but simpler.

eregon:
> Since you're already using the abstract method form, couldn't you just make the node a field in DebugOperations? Not a big deal either way, but it could reduce object churn and construction time.

We need RubyContext to build it, and caching in a static member is always messy with multi-threading.

> Could we also do ruby("o.freeze", "o", o) - that will be even slower but simpler.

We would need a VirtualFrame (could be removed if we pass self explicitly), which the various caller don't have. And possibly a currentNode too (although the parser seems happy without).
It would be likely problematic when called early such as in CoreLibrary (could be fixed with some Ruby code) and really really slow.

eregon:
Can I merge this?

nirvdrum:
Go for it. If it turns out to be too slow, we can always go back to the old way.

PR #2848, opened 2015-04-17
[Truffle] Adding additional long handling for multiple specializations.

bjfish:
I believe somewhere while running the MRI tests an edge case is causing the AST to change to use some specialization which returns long frequently (possibly Fixnum#-). I’d like to add these specializations to handle long. I’ve added a TODO to indicate these should be reviewed later on for correctness.

### DIFF
--- a/truffle/src/main/java/org/jruby/truffle/nodes/rubinius/StringPrimitiveNodes.java
+++ b/truffle/src/main/java/org/jruby/truffle/nodes/rubinius/StringPrimitiveNodes.java
@@ -266,6 +266,27 @@ public abstract class StringPrimitiveNodes {
         }
 
         @Specialization
+        public Object stringByteSubstring(RubyString string, long index, int length) {
+            return stringByteSubstring(string, index, (long) length);
+        }
+
+        @Specialization
+        public Object stringByteSubstring(RubyString string, int index, long length) {
+            return stringByteSubstring(string, (long) index, length);
+        }
+
+        @Specialization
+        public Object stringByteSubstring(RubyString string, long index, long length) {
+            if (index > Integer.MAX_VALUE || index < Integer.MIN_VALUE) {
+                throw new RaiseException(getContext().getCoreLibrary().argumentError("index out of int range", this));
+            }
+            if (length > Integer.MAX_VALUE || length < Integer.MIN_VALUE) {
+                throw new RaiseException(getContext().getCoreLibrary().argumentError("length out of int range", this));
+            }
+            return stringByteSubstring(string, (int) index, (int) length);
+        }
+
+        @Specialization
         public Object stringByteSubstring(RubyString string, double index, double length) {
             return stringByteSubstring(string, (int) index, (int) length);
         }


nirvdrum:
Anything that is a `@CoreMethod` can use `lowerFixnumParameters`.  In the case of `MatchDataNodes`, the `lowerFixnumParameters` is present, but is pointing at the wrong parameter index.  Note that the index is 0-based and does not include `self`.  I messed this up originally and it looks like you found a couple lingering cases there.

chrisseaton:
`-Xtruffle.randomize_storage.array=true` can help you find these problems - it randomly choose a way to represent arrays whenever they are modified - including the rare-in-practice `long[]`.

nirvdrum:
I hate complicating this, but it looks on MRI, `Integer#downto` and `Integer#upto` work fine on values > 32-bit.  So casting here would be wrong.  @chrisseaton should we just effectively duplicate that code for (long, long), (long, int), and (int, long) permutations?

chrisseaton:
Yes casting `long` to `int` without checking the range is absolutely wrong - at least the first specialisation in this PR does that. It will silently truncate, which is the worst behaviour we could have.

Casting `int` to `long` is of course fine and often what we want to do.

Note that `lowerFixnumParameters` only lowers a `long` to an `int` if it is in that range. Otherwise it passes it through as a `long`.

bjfish:
@chrisseaton what would be the error to raise if the long is outside of the int range for a methods that only currently support int?

could the the same raise exception be thrown from a primitive node?

chrisseaton:
I think the key one is array indexing - as that is restricted to `int` by limitations of the JVM. JRuby seems to allow reads from an array with any index, and returns `nil`. Writes it gives an error that I'm not sure if MRI ever gives.

````
$ ~/.rbenv/versions/jruby-1.7.19/bin/ruby -e '[1, 2][0xffffffffffffff]'
$ ~/.rbenv/versions/jruby-1.7.19/bin/ruby -e '[1, 2][0xffffffffffffff] = 14'
ArgumentError: index too big
     []= at org/jruby/RubyArray.java:1408
  (root) at -e:1
````

For a `Bignum` MRI does give an error message.

````
$ ruby -e '[1, 2][10000000000000000000000000]'
-e:1:in `[]': bignum too big to convert into `long' (RangeError)
	from -e:1:in `<main>'
````

Note that you can index an array with some `Bignums`, as there are some values of the C type `long`, which is what MRI indexes arrays with, that do not fit into a `Fixnum` due to tagging.

Short answer - I think it depends on the method! I would concentrate on the specs for now - if the specs cover a case then fix it. Otherwise I am working on some stress test utilities that will help us explore this problem space more in the future.

**Comment by bjfish** [comment]
I've updated the PR to just fix String#byteslice https://github.com/jruby/jruby/issues/2843. This is used a lot and fixes a stackoverflow.

File: truffle/src/main/java/org/jruby/truffle/nodes/rubinius/StringPrimitiveNodes.java [hunk 1]
```diff
@@ -266,6 +266,33 @@ public Object stringByteSubstring(RubyString string, double index, UndefinedPlac
         }
 
         @Specialization
+        public Object stringByteSubstring(RubyString string, long index, int length) {
+            if (index > Integer.MAX_VALUE || index < Integer.MIN_VALUE) {
+                throw new RaiseException(getContext().getCoreLibrary().argumentError("index out of int range", this));
+            }
+            return stringByteSubstring(string, (int) index, length);
+        }
```
eregon: You could just call the (long, long) version to avoid duplication.

bjfish: I've updated this to avoid the duplication.

PR #2855, opened 2015-04-20
[Truffle] Complete, correct and optimised implementation of Array#pack

chrisseaton:
This is an implementation of the `Array#pack` method - it's like `printf` for binary data.

Until this patch, we have been using JRuby's pack implementation, but that was never going to be complete or correct, as it makes calls on objects and that will use JRuby's classes not Truffles. It was also never going to be optimised as it requires converting objects to JRuby and then back to Truffle.

This implementation compiles pack expressions to Truffle nodes, and so makes them available for compilation. It also includes an algorithm to try to recover loops in very long expressions that we found being used in the wild - a kind of reverse unrolling. The nodes are generally fairly linear, so it's not quite like a normal AST, but I've also decomposed the operations, so a single formatting directive becomes separate read, write and convert nodes, which all specialise independently.

For the first time I'm using the new DSL to cache the compiled expressions.

Also important is being able to PE through pack expressions. Pack is often used to convert a single value to a binary representation. The result is that for example we can constant fold this entire expression `[14].pack('C').getbyte(0) * 2`.

One final point is that I've gone to more lengths than usual to check the implementation is correct. It passes all 1596 specs of course, but I can also run the specs with compilation turned on, which we don't usually do, and with array storage randomisation.

```
$ jt build truffle test --graal -T-J-G:+TraceTruffleCompilation -T-Xtruffle.randomize_storage.array=true -T-J-G:+TruffleCompilationExceptionsAreFatal spec/ruby/core/array/pack
```

I'd like to also be able to test with specs run repeatedly and with a very low compilation threshold, but there's still a lot to be done to make that work.

Not for merge yet - we'll keep the `truffle-head` branch as just the API delta for now and merge this after the release of 0.7.

@thomaswue @eregon @nirvdrum @bjfish


chrisseaton:
When reviewing, just look for the new files and `ArrayNodes.java`. Everything else is pretty trivial.

core/src/main/java/org/jruby/util/Pack.java
```diff
@@ -542,7 +561,7 @@ private static ByteList encodes(Ruby runtime, ByteList io2Append,byte[]charsToEn
      * @param iLength The max number of characters to encode
      * @return the io2Append buffer
      **/
-    private static ByteList qpencode(ByteList io2Append, ByteList i2Encode, int iLength) {
+    public static ByteList qpencode(ByteList io2Append, ByteList i2Encode, int iLength) {
```
nirvdrum: It might be better to move this out to a `PackSupport` or `PackUtils` like we've done with `RubyString`.  While that was mostly to divorce from `org.jruby.RubyString` it was also done to prevent people thinking it was now a public part of the API.

core/src/main/java/org/jruby/RubyArray.java
```diff
@@ -60,28 +60,19 @@
...
 import java.lang.reflect.Array;
-import java.util.Arrays;
-import java.util.Collection;
-import java.util.Comparator;
-import java.util.IdentityHashMap;
-import java.util.Iterator;
-import java.util.List;
-import java.util.ListIterator;
-import java.util.RandomAccess;
+import java.util.*;
```
nirvdrum: I'd double-check this. But I think JRuby prefers all the imports listed out.

chrisseaton: Ah - there's not even any useful changes in this file. I'll just revert it. I had experimented with integrating our pack into JRuby which is why I was in this file, but there is a DSL limitation that makes it hard so I thought I'd leave that until later.

tool/jt.rb
```diff
@@ -131,8 +131,13 @@ def mvn(*args)
     sh 'mvn', *args
   end
 
-  def mspec(command, *args)
-    sh 'ruby', 'spec/mspec/bin/mspec', command, '--config', 'spec/truffle/truffle.mspec', *args
+  def mspec(env, command, *args)
+    env_vars = {}
+    mspec_env env_vars, command, *args
```
eregon: `env` is not used in the definition here and this will ignore the first argument. I think you meant to leave the arguments unchanged.

truffle/src/main/java/org/jruby/truffle/pack/nodes/PackNode.java
```diff
@@ -0,0 +1,183 @@
...
+    public int getSourceLength(VirtualFrame frame) {
+        try {
+            return frame.getInt(PackFrame.INSTANCE.getSourceLengthSlot());
+        } catch (FrameSlotTypeException e) {
+            throw new IllegalStateException(e);
+        }
+    }
+
+    /**
+     * Set the current position we are reading from in the source array.
```
eregon: Get

truffle/src/main/java/org/jruby/truffle/pack/nodes/PackNode.java
```diff
@@ -0,0 +1,183 @@
...
+import com.oracle.truffle.api.utilities.ConditionProfile;
+import org.jruby.truffle.pack.runtime.PackFrame;
+import org.jruby.truffle.pack.runtime.exceptions.TooFewArgumentsException;
+import org.jruby.util.ByteList;
+
+import java.util.Arrays;
+
+/**
+ * The root of the pack nodes.
+ * <p>
```
eregon: Parse error

truffle/src/main/java/org/jruby/truffle/pack/nodes/control/BackNode.java
```diff
@@ -0,0 +1,40 @@
...
+ */
+package org.jruby.truffle.pack.nodes.control;
+
+import com.oracle.truffle.api.CompilerDirectives;
+import com.oracle.truffle.api.frame.VirtualFrame;
+import org.jruby.truffle.pack.nodes.PackNode;
+import org.jruby.truffle.pack.runtime.exceptions.OutsideOfStringException;
+
+/**
+ * Moves the output position to a particular location - similar to seek
```
eregon: to the previous location

truffle/src/main/java/org/jruby/truffle/pack/runtime/PackFrame.java
```diff
@@ -0,0 +1,66 @@
...
+ */
+package org.jruby.truffle.pack.runtime;
+
+import com.oracle.truffle.api.frame.FrameDescriptor;
+import com.oracle.truffle.api.frame.FrameSlot;
+import com.oracle.truffle.api.frame.FrameSlotKind;
+
+public class PackFrame {
+
+    public static final PackFrame INSTANCE = new PackFrame();
```
eregon: `PackFrame.INSTANCE.getSourceSlot()` looks a bit confusing to me.
Would it make sense to use static members to clarify why there is only one instance? (i.e. This is the *descriptor* of a Frame in PackNodes, not the frame itself)

truffle/src/main/java/org/jruby/truffle/nodes/methods/ExceptionTranslatingNode.java
```diff
@@ -134,7 +134,11 @@ private RubyException translate(UnsupportedSpecializationException exception) {
                     }
                 }
             } else {
-                builder.append(value.getClass().getName());
+                if (value == null) {
+                    builder.append("null");
```
eregon: I met the same problem recently, but I did the check at line 103, which might to follow the flow better.

truffle/src/main/java/org/jruby/truffle/nodes/dispatch/UncachedDispatchNode.java
```diff
@@ -48,6 +51,7 @@ public UncachedDispatchNode(RubyContext context, boolean ignoreVisibility, Dispa
         callNode = Truffle.getRuntime().createIndirectCallNode();
         toSymbolNode = ToSymbolNodeFactory.create(context, null, null);
         toJavaStringNode = ToJavaStringNodeFactory.create(context, null, null);
+        metaClassNode = MetaClassNodeFactory.create(context, null, null);
```
eregon: Yeah, +1 for MetaClass nodes!

truffle/src/main/java/org/jruby/truffle/nodes/core/ArrayNodes.java
```diff
@@ -2615,159 +2618,143 @@ public CallTarget getCallTarget() {
...
-            }
+            return string;
+        }
 
-            if (result instanceof RubyString) {
-                return pack(frame, array, (RubyString) result);
-            }
+        @Specialization
+        public Object pack(VirtualFrame frame, RubyArray array, boolean format) {
+            return ruby(frame, "raise TypeError");
```
eregon: I'm not such a big fan of raising exception with the ruby() helper.
Notably raise is implemented rather late and the overhead is likely significant, while the gains minimal IMO.
But of course in this case I suppose a good ToStr node would do it instead of the extra specializations.

eregon:
It's a bit hard to grasp with the large diff so I think I'd better have a look in Eclipse instead.
This looks great in any case!

truffle/src/main/java/org/jruby/truffle/pack/nodes/read/ReadDoubleNode.java
```diff
@@ -0,0 +1,79 @@
+/*
```
eregon: I would group these "read primitive" nodes in a single file if that makes sense so to avoid *n* tiny files and ease possible refactoring.

bjfish:
@chrisseaton Here is a test with some pack usage found in protobuf/oga if you want to give this a try:
https://gist.github.com/bjfish/abda46b6505c0a7f48a3

Runs fine in jruby/ruby but saw some differences with truffle-pack branch.

truffle/src/main/java/org/jruby/truffle/nodes/core/ArrayNodes.java
```diff
@@ -2615,159 +2618,143 @@ public CallTarget getCallTarget() {
...
         }
 
-        // TODO CS 3-Mar-15 to be honest these two specialisations are a bit sneaky - we'll get rid of them ASAP
+        @Specialization(guards = "byteListsEqual(format, cachedFormat)")
+        public RubyString packCached(
+                VirtualFrame frame,
+                RubyArray array,
+                RubyString format,
+                @Cached("privatizeByteList(format)") ByteList cachedFormat,
+                @Cached("create(compileFormat(format))") DirectCallNode callPackNode) {
```
nirvdrum: What is `create` here?

core/src/main/java/org/jruby/RubyArray.java
chrisseaton: Fixed in f7661d7f95ac09a351bb00695ce6b24d7f1e0c62

tool/jt.rb
chrisseaton: Fixed in 0399ce5fe4dfe8d6af9c05b5795cc6e6ed9e1aff

truffle/src/main/java/org/jruby/truffle/pack/nodes/PackNode.java
chrisseaton: Fixed in 1cbc92cc36d96ccc7f8e6f9dc621faa711249d2d

truffle/src/main/java/org/jruby/truffle/pack/nodes/PackNode.java
chrisseaton: The `<p>`? That's a common notation in JavaDoc. I thought it was the correct way to do things?

truffle/src/main/java/org/jruby/truffle/pack/nodes/control/BackNode.java
chrisseaton: Fixed in fab18bbec4e62c0ca44f64fd88a671808388966f

truffle/src/main/java/org/jruby/truffle/pack/runtime/PackFrame.java
chrisseaton: Fixed in 63269314fac96870fdb714ca6380414301f2d4e5

truffle/src/main/java/org/jruby/truffle/nodes/methods/ExceptionTranslatingNode.java
chrisseaton: Fixed in c085e5747c368ea7fefb220768d71cedeed6de2f

truffle/src/main/java/org/jruby/truffle/pack/nodes/read/ReadDoubleNode.java
chrisseaton: The read nodes are a screen of code each - they're not that tiny. Do you mean the write nodes? They are pretty small.

truffle/src/main/java/org/jruby/truffle/nodes/core/ArrayNodes.java
chrisseaton: The DSL expands it into `DirectCallNode.create(compileFormat(format))`. There is a lookup order for where it finds methods in guards, and one place is static members of the parameters's class. It's not just copied and pasted into the DSL generated code - there is a specific lookup.

truffle/src/main/java/org/jruby/truffle/pack/nodes/PackNode.java
eregon: Good then, my mistake. I thought `<p>` would be similar to its HTML version but apparently it behaves a bit more like <br> and does not need to be closed.

truffle/src/main/java/org/jruby/truffle/pack/nodes/read/ReadDoubleNode.java
eregon: Indeed, my mistake, I meant integer write nodes. But it's also good to keep one per file for consistency then.

core/src/main/java/org/jruby/util/Pack.java
chrisseaton: Fixed in dccd8021d7aaeaeb7cb85a99a93791c51093768c

PR #2858, opened 2015-04-21
Fix rbenv install instructions

sferik:


mkristian:
thanx

headius:
This actually went out with pre2.

PR #2859, opened 2015-04-21
[Truffle] Add MRI test profile to pom files.

bjfish:


bjfish:
With workarounds (after current MRI test issues are fixed) `mvn -Ptruffle-mri-test` completes with:
```
     [exec] Finished tests in 43.266000s, 38.5291 tests/s, 3727.0836 assertions/s.
     [exec] 
     [exec] 1667 tests, 161256 assertions, 0 failures, 0 errors, 1
```

bjfish:
I just realized. I don't like this because every time the mri_truffle.index changes the pom.xml would need to be generated. Going to find another way.

PR #2866, opened 2015-04-22
[Truffle] New IO implementation

chrisseaton:
This reimplements Truffle's File and IO classes using Rubinius, on top of a few Rubinius primitives and some POSIX calls. There's now no reference to File or IO in the Java code at all.

This gives patch alone gives us 761 new passing specs (about 5% of the entire core specs), but it also opens the door for the rest of IO and things like Socket.

The primitives use `ruby()` a lot, mainly to read and write instance variables, which we might want to remove sooner rather than later.

There is also a lot of copying going on here. I think `puts` will copy at least four times. The Rubinius primitives also rely on stack allocation of large buffers being efficient, which does not translate well into Java.

@eregon @nirvdrum @thomaswue


truffle/src/main/java/org/jruby/truffle/nodes/rubinius/IOBufferPrimitiveNodes.java
```diff
@@ -0,0 +1,193 @@
...
+
+                    if (errno == Errno.ECONNRESET.intValue() || errno == Errno.ETIMEDOUT.intValue()) {
+                        // Treat as seeing eof
+                        bytesRead = 0;
+                        break;
+                    } else if (errno == Errno.EAGAIN.intValue() || errno == Errno.EINTR.intValue()) {
+                        //if (!state -> check_async(calling_environment))
+                        //    return NULL;
+                        //io -> ensure_open(state);
+                        getContext().getSafepointManager().poll(this);
```
eregon: If this actually works on STDIN, this will be awesome!

chrisseaton:
```ruby
STDIN.each_line do |line|
  puts line
end
```

With both tty and piped input works. Is that what you meant?

eregon: Yes, exactly :+1:

eregon: I mean, the ability to interrupt reading from STDIN is what interests me.

chrisseaton:
Merged, but still open for comments

truffle/src/main/java/org/jruby/truffle/nodes/core/ArrayNodes.java
```diff
@@ -1296,13 +1296,17 @@ public Object deleteAtNullOrEmpty(RubyArray array, int index) {
         private final BranchProfile nextProfile = BranchProfile.create();
         private final BranchProfile redoProfile = BranchProfile.create();
 
+        private final RubySymbol eachSymbol;
+
         public EachNode(RubyContext context, SourceSection sourceSection) {
             super(context, sourceSection);
+            eachSymbol = getContext().getSymbolTable().getSymbol("each");
```
nirvdrum: I think we use `getContext().newSymbol()` almost exclusively.  It might make future refactorings easier to stick with that.

truffle/src/main/java/org/jruby/truffle/nodes/core/KernelNodes.java
```diff
@@ -1562,7 +1567,12 @@ public boolean doesRespondTo(VirtualFrame frame, Object object, RubyString name,
 
         @Specialization
         public boolean doesRespondTo(VirtualFrame frame, Object object, RubySymbol name, UndefinedPlaceholder checkVisibility) {
-            return dispatch.doesRespondTo(frame, name, object);
+            return doesRespondTo(frame, object, name, false);
+        }
+
+        @Specialization
+        public boolean doesRespondTo(VirtualFrame frame, Object object, RubySymbol name, RubyBasicObject checkVisibility) {
+            return doesRespondTo(frame, object, name, false);
```
nirvdrum: This is a case where a one-off `@CreateCast` could keep the number of specializations down to 1, at the cost of having to box the boolean.

truffle/src/main/java/org/jruby/truffle/nodes/core/TrufflePrimitiveNodes.java
```diff
@@ -456,4 +456,24 @@ public RubyNilClass simpleShell() {
...
+            super(context, sourceSection);
+        }
+
+        public DebugPrintNode(DebugPrintNode prev) {
+            super(prev);
+        }
+
+        @CompilerDirectives.TruffleBoundary
+        @Specialization
+        public RubyNilClass debugPrint(RubyString string) {
```
nirvdrum: I was wondering why you would need this and then I remembered you were rewriting IO ;-)

truffle/src/main/java/org/jruby/truffle/nodes/rubinius/IOBufferPrimitiveNodes.java
```diff
@@ -0,0 +1,193 @@
...
+
+        @Specialization
+        public int fill(VirtualFrame frame, RubyBasicObject ioBuffer, RubyBasicObject io) {
+            int fd = (int) rubyWithSelf(frame, io, "@descriptor");
+
+            // TODO CS 21-Apr-15 allocating this buffer for each read is crazy
+            byte[] readBuffer = new byte[STACK_BUF_SZ];
+            int count = STACK_BUF_SZ;
+
+            if(left(frame, ioBuffer) < count) {
```
nirvdrum: Nitpick, but it looks like you changed your code style.  No space between the `if` and `(`.  I noticed it elsewhere as well.   Not a big deal.  Just a curious observation.

truffle/src/main/java/org/jruby/truffle/nodes/rubinius/IOPrimitiveNodes.java
```diff
@@ -43,4 +89,100 @@ public boolean fnmatch(RubyString pattern, RubyString path, int flags) {
...
+
+        public IOWritePrimitiveNode(IOWritePrimitiveNode prev) {
+            super(prev);
+        }
+
+        @Specialization
+        public int write(VirtualFrame frame, RubyBasicObject file, RubyString string) {
+            final int fd = (int) rubyWithSelf(frame, file, "@descriptor");
+
+            // We have to copy here as write starts at byte[0], and the ByteList may not
```
nirvdrum: Can you use `pwrite` to supply the offset and avoid the copy?

truffle/src/main/java/org/jruby/truffle/nodes/rubinius/StringPrimitiveNodes.java
```diff
@@ -1359,4 +1359,23 @@ private RubyString makeSubstring(RubyString string, int beg, int len) {
...
+            super(context, sourceSection);
+        }
+
+        public StringFromByteArrayPrimitiveNode(StringFromByteArrayPrimitiveNode prev) {
+            super(prev);
+        }
+
+        @Specialization
+        public RubyString stringFromByteArray(RubiniusByteArray bytes, int start, int count) {
+            // Data is copied here - can we do something COW?
```
nirvdrum: I'd make this a `TODO` if you want to revisit it.

truffle/src/main/java/org/jruby/truffle/runtime/RubyContext.java
```diff
@@ -309,6 +311,10 @@ public RubyString makeString(RubyClass stringClass, char string, Encoding encodi
         return makeString(stringClass, Character.toString(string), encoding);
     }
 
+    public RubyString makeString(byte[] bytes) {
```
nirvdrum: The number of these `makeString` cases is getting a bit high.  We may want to rethink where they live.

truffle/src/main/java/org/jruby/truffle/runtime/core/CoreLibrary.java
```diff
@@ -419,28 +389,9 @@ private void initializeConstants() {
...
-            altSeparator.freeze();
-
-            fileClass.setConstant(null, "ALT_SEPARATOR", altSeparator);
-        } else {
-            fileClass.setConstant(null, "ALT_SEPARATOR", nilObject);
-        }
-
-        fileClass.setConstant(null, "PATH_SEPARATOR", RubyString.fromJavaString(stringClass, File.pathSeparator));
-        fileClass.setConstant(null, "FNM_SYSCASE", 0);
+        rubiniusModule.setConstant(null, "UNDEFINED", rubiniusUndefined);
```
nirvdrum: I hope the separator stuff is handled nicely in Rubinius.  Without this, a lot of Windows simply won't work.

truffle/src/main/ruby/core/shims.rb
```diff
@@ -216,5 +193,16 @@ def self.synchronize(object)
...
+  DomainError = EDOM
+
+end
+
+module Math
+  DomainError = Errno::EDOM
+end
+
+$PROGRAM_NAME = $0
+$$ = Process.pid
```
nirvdrum: Does SVM support fork?

truffle/src/main/java/org/jruby/truffle/nodes/core/ArrayNodes.java
eregon: Didn't know about that. I would expect it to be `getContext().getSymbol()` since this will likely not create a new Symbol everytime.

truffle/src/main/java/org/jruby/truffle/nodes/core/KernelNodes.java
eregon: How to treat UndefinedPlaceholder though? We would need something to specify which default value it should take.

truffle/src/main/java/org/jruby/truffle/nodes/rubinius/IOBufferPrimitiveNodes.java
eregon: Yes, I noticed it as well. I suppose it's because this is transliterated from C++ code, but we should have sth checking against this.

truffle/src/main/java/org/jruby/truffle/nodes/rubinius/IOPrimitiveNodes.java
eregon: `pwrite` is about the offset of the fd, not the buffer though, isn't it?

truffle/src/main/java/org/jruby/truffle/nodes/core/KernelNodes.java
nirvdrum: Ahh, yeah.  I was thinking we could just treat it as an `Object`, but there isn't even a child node in that case, is there?

truffle/src/main/java/org/jruby/truffle/nodes/core/ArrayNodes.java
nirvdrum: It's perhaps a bit misleadingly named.  It just does `return symbolTable.getSymbol(name, ASCIIEncoding.INSTANCE);`, which will return a cached `Symbol` if it already exists or creates a new one and populates the cache otherwise.

truffle/src/main/java/org/jruby/truffle/nodes/rubinius/IOPrimitiveNodes.java
nirvdrum: You're right.  I misread the manpage.  Sorry for the false hope.

truffle/src/main/java/org/jruby/truffle/nodes/core/KernelNodes.java
eregon: It would work for the name but if you want to remove the UndefinedPlaceholder, BooleanCastNode has to handle it somehow.

truffle/src/main/java/org/jruby/truffle/nodes/rubinius/IOPrimitiveNodes.java
eregon: We really need a pointer to the middle of the array.
Maybe using the write version with a ByteBuffer can do that?

truffle/src/main/java/org/jruby/truffle/nodes/rubinius/IOBufferPrimitiveNodes.java
chrisseaton: No, I'm afraid it doesn't appear to work at the moment. Can you interrupt all reads in a process somehow?

truffle/src/main/java/org/jruby/truffle/nodes/core/ArrayNodes.java
chrisseaton: Fixed in 5c53e3c984459de85bf4eda0902c0f7775d72ab6

truffle/src/main/java/org/jruby/truffle/nodes/rubinius/IOBufferPrimitiveNodes.java
chrisseaton: Fixed in c770ac57d8de3432d03dd85e3fa8fad2d4acbd65

truffle/src/main/java/org/jruby/truffle/nodes/rubinius/IOPrimitiveNodes.java
chrisseaton: Maybe we need to modify jnr-posix to let us have a start offset.

truffle/src/main/java/org/jruby/truffle/runtime/RubyContext.java
chrisseaton: Maybe we should move all helpers into RubyNode where it really does simplify the code?

truffle/src/main/java/org/jruby/truffle/runtime/core/CoreLibrary.java
chrisseaton: I don't guarantee it is. I'm happy with some temporary Windows breakage - I know that's not ideal for you when you're using Windows, but progress today with for example IRB shows we're doing the right thing.

truffle/src/main/ruby/core/shims.rb
chrisseaton: I guess the real question you're asking is why am I setting this statically when it could change? Well that's why it's in `shims.rb` where there is no shame.

truffle/src/main/java/org/jruby/truffle/runtime/core/CoreLibrary.java
nirvdrum: Understood.  I was just raising it as an area we'll have to watch.  It may not be as simple as updating jnr-posix calls is all.

truffle/src/main/ruby/core/shims.rb
nirvdrum: Fair enough.

truffle/src/main/java/org/jruby/truffle/runtime/RubyContext.java
eregon: This would make them inaccessible from non-Node callers, which is undesirable though.

PR #2870, opened 2015-04-23
[Truffle] Add MRI tests to Travis builds and update the jt tool.

bjfish:


chrisseaton:
I'm going to let Travis run on this one.

bjfish:
@chrisseaton @nirvdrum @eregon I thought it would make sense for `jt test` to run all tests and then `jt test` would become `jt test specs`.

chrisseaton:
You've got it so that it's

```
jt test specs spec/ruby/language/while_spec.rb
```

But I think we can look at the path you supply to work out if it's MRI or RubySpec. We can tel the difference between

```
jt test spec/ruby/language/while_spec.rb
```

and

```
jt test test/test/mri/ruby/test_whileuntil.rb
```

automatically. Then `jt test` should probably run everything, `jt test specs` just specs, and `jt test specs fast` only the fast specs.

bjfish:
@chrisseaton updated jt per your comment

PR #2875, opened 2015-04-25
Fix typo in error message

jkutner:


jkutner:
This could be a problem for warbler's `runnable` feature, which uses `file:` path for jruby home :/

mkristian:
@jkutner did you see the error message on warbler ? if so there is no problem to remove this message IMO

jkutner:
@mkristian it seemed to "work" with the `file:/path/to/jar/...` prefix. But I was able to remove the warning by replace it with `uri:classloader:!/META-INF/...`

I say "work" because I'm stuck against another problem where `Gem::Specification` is returning `uri:classloader:...` for the executable script, but `RubyInstanceConfig.getScriptSource()` doesn't handle it.

mkristian:
I am happy to see that "uri:classloader:" works everywhere. but I am not sure where and how this ```RubyInstanceConfig.getScriptSource()```. could give me a bit more input. maybe we should continue this on new issue or so

jkutner:
here: https://github.com/jruby/jruby/pull/2876

PR #2894, opened 2015-04-30
Mimic MRI's openssl when jopenssl require fails.

voxik:
Some libraries (such as RubyGems [1]) rescues openssl load errors and
adjusts their behavior depending if the openssl is properly required.
However, prior this commit, require 'openssl' succeeded just fine, but
the subsequent requires failed, which makes RubyGems (and similar) to
fail unexpectedly.

[1] https://github.com/rubygems/rubygems/blob/master/lib/rubygems/security.rb#L14

### DIFF
--- a/lib/ruby/shared/openssl.rb
+++ b/lib/ruby/shared/openssl.rb
@@ -1,1 +1,5 @@
-require 'jopenssl/load'
+begin
+  require 'jopenssl/load'
+rescue LoadError => e
+  e.message.sub! 'jopenssl/load', 'openssl'
+end


mkristian:
that this file 'openssl.rb'  is in the jruby repo is probably not right. investigating . . .

mkristian:
as I thought any change in this file will be overwritten by the one we import during the build from jruby-openssl.gem

mkristian:
@voxik so any change in jruby-openssl will reflect on jruby. so next jruby-openssl version might have your patch.

voxik:
Perfect, thanks :)

PR #2919, opened 2015-05-08
Add another layer of try/finally around rescue/ensure to reset $!.

### LINKED ISSUE #2794: Load/require logic lacks path caching and indexing found in MRI
MRI has a number of caches and indexes that improve the performance of load/require searching. We lack most of those caches, or have poor implementations of them.

The two key bits I know of:
- The entries in the load path are made into absolute paths and held in a separate index used for file resolution.
- The resolved requires are absolutized for LOADED_FEATURES and a mapping is maintained between the require name and the absolute name. This index is updated whenever LOADED_FEATURES is seen to have changed.

We need to implement our loading logic so that it performs similar mappings, both to improve lookup performance and to better match MRI behavior with regards to full paths.

This may push to post-9k since it does not fix any known compatibility issues.

### LINKED ISSUE #2910: $! does not get cleared under certain circumstances
```
$ gem install leafy-metrics
Fetching: leafy-metrics-0.4.0-java.gem (100%)
  jar dependencies for leafy-metrics-0.4.0-java.gemspec . . .
      io.dropwizard.metrics:metrics-core:3.1.0
      io.dropwizard.metrics:metrics-graphite:3.1.0
      io.dropwizard.metrics:metrics-jvm:3.1.0
      org.slf4j:slf4j-simple:1.7.7
Successfully installed leafy-metrics-0.4.0-java
ERROR:  While executing gem ... (Psych::SyntaxError)
    (<unknown>): mapping values are not allowed here at line 13 column 151
```

headius:
Fixes #2910.

### DIFF
--- a/core/src/main/java/org/jruby/RubyMethod.java
+++ b/core/src/main/java/org/jruby/RubyMethod.java
@@ -206,13 +206,14 @@ public class RubyMethod extends AbstractRubyMethod {
         };
 
         BlockBody body;
+        String[] parameterList = JRubyLibrary.MethodExtensions.methodParameters(runtime, method);
         if (method instanceof IRMethodArgs) {
             Signature signature = ((IRMethodArgs) method).getSignature();
             body = CompiledBlockLight19.newCompiledBlockLight(signature,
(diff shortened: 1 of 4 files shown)


headius:
Merged to master.

Issue #2924, opened 2015-05-09
Issue in org.jruby.util.TypeConverter CheckType


cdwijayarathna:
I tried to use above mentioned method to check it a IRubyObject is a RubyHash or not,
Following is how I used it.
TypeConverter.checkType(context, opts, context.runtime.getHash());
When I give a Hash as opts, my irb interface I get following output.

TypeError: wrong argument type Hash (expected Hash)

I'm sure that here it should be "Wrong arguement type <something else>",
because in https://github.com/jruby/jruby/blob/master/core/src/main/java/org/jruby/util/TypeConverter.java#L300 it has used the reference class to (RubyModule t) to identify the type of the object ('x').So I believe #L300 should be x.someMethod(), not t.someMethod().

Other issue is, when I pass a Hash, to this it gives this TypeError. I think the issue is in https://github.com/jruby/jruby/blob/master/core/src/main/java/org/jruby/util/TypeConverter.java#L299,
x.getMetaClass().getNativeClassIndex() refers to "CLASS" while t.getClassIndex() refers to "HASH". 
I tried with other types as well and I didn't get true when types match at any time.

headius:
Reproduction in Ruby:

```ruby
runtime = JRuby.runtime
context = runtime.current_context
opts = {foo: 'bar'}
org.jruby.util.TypeConverter.check_type(context, opts, runtime.hash)
```

headius:
Chamila: I think you're probably right. Only one place in JRuby is using this method, so it must be stale or perhaps it never worked properly at all.

You should compare our version with MRI's version and make them match in a PR.

cdwijayarathna:
MRI implementation

    void
    rb_check_type(VALUE x, int t)
    {
        int xt;
    
        if (x == Qundef) {
    	rb_bug("undef leaked to the Ruby space");
        }
    
        xt = TYPE(x);
        if (xt != t || (xt == T_DATA && RTYPEDDATA_P(x))) {
    	const char *tname = rb_builtin_type_name(t);
	    if (tname) {
	        rb_raise(rb_eTypeError, "wrong argument type %s (expected %s)",
		         builtin_class_name(x), tname);
	    }
	    if (xt > T_MASK && xt <= 0x3f) {
	        rb_fatal("unknown type 0x%x (0x%x given, probably comes from extension library for ruby 1.8)", t, xt);
	    }
	    rb_bug("unknown type 0x%x (0x%x given)", t, xt);
        }
    }

[...]

PR #2930, opened 2015-05-11
Issue #2924

cdwijayarathna:


kares:
meh - lot of merges here ... could we have only the relevant commit with an explanation maybe :) ?

cdwijayarathna:
@kares  Please refer https://github.com/jruby/jruby/pull/2939,
Do I need to give explanation? I have provided detail of issue this PR address at Issue #2924

PR #2961, opened 2015-05-19
[Truffle] BigDecimal basic operations improvements, comparisons, to_s

pitr-ch:


lib/ruby/truffle/truffle/bigdecimal.rb
```diff
@@ -1,5 +1,15 @@
 # TODO require bigdecimal/*
 
+class Truffle::BigDecimal < Numeric
+  include Comparable
+
+  def ==(other)
+    (self <=> other) == 0
+  end
```
eregon: `include`-ing Comparable should already do this, but it might have some differences regarding exceptions rescued.

lib/ruby/truffle/truffle/bigdecimal.rb
```diff
@@ -1,5 +1,15 @@
 # TODO require bigdecimal/*
```
chrisseaton: What is in `bigdecimal/*`?

truffle/src/main/java/org/jruby/truffle/nodes/core/BigDecimalNodes.java
```diff
@@ -194,34 +256,199 @@ public SubNode(RubyContext context, SourceSection sourceSection) {
...
+    }
+
+    @CoreMethod(names = "/", required = 1)
+    public abstract static class DivOpNode extends BigDecimalCoreMethodNode {
+
+        public DivOpNode(RubyContext context, SourceSection sourceSection) {
+            super(context, sourceSection);
+        }
+
+        private Object mulBigDecimal(RubyBasicObject a, BigDecimal b) {
```
eregon: double check the method name :smile:

truffle/src/main/java/org/jruby/truffle/nodes/core/BigDecimalNodes.java
```diff
@@ -148,41 +176,75 @@ public AddNode(RubyContext context, SourceSection sourceSection) {
             super(context, sourceSection);
         }
 
-        @Specialization(guards = "isRubyBigDecimal(b)")
-        public Object add(RubyBasicObject a, RubyBasicObject b, int precision) {
+        private Object addBigDecimal(RubyBasicObject a, BigDecimal b, int precision) {
```
chrisseaton: We probably want a `@TruffleBoundary` on these helper methods. Java's `BigDecimal` is unlikely to partially evaluate well, and even if it did we probably don't want the operations inlined each time they're used.

truffle/src/main/java/org/jruby/truffle/nodes/core/BigDecimalNodes.java
```diff
@@ -194,34 +256,199 @@ public SubNode(RubyContext context, SourceSection sourceSection) {
...
+        }
+
+        @Specialization(guards = "isRubyBigDecimal(b)")
+        public Object division(RubyBasicObject a, RubyBasicObject b) {
+            return mulBigDecimal(a, getBigDecimalValue(b));
+        }
+
+    }
+
+    // TODO div,quo,divmod
```
chrisseaton: Please name and date any TODOs. I know you can get the same info from blame but it's nice to scan and see if someone is a month old or a year old.

truffle/src/main/java/org/jruby/truffle/nodes/core/BigDecimalNodes.java
```diff
@@ -236,7 +463,17 @@ public ToSNode(RubyContext context, SourceSection sourceSection) {
         @CompilerDirectives.TruffleBoundary
         @Specialization
         public RubyString toS(RubyBasicObject value) {
-            return getContext().makeString(getBigDecimalValue(value).toString());
+            BigDecimal bigDecimal = getBigDecimalValue(value);
+            boolean negative = bigDecimal.signum() == -1;
```
chrisseaton: We tend to mark locals as final unless there is a need for them to be mutable.

lib/ruby/truffle/truffle/bigdecimal.rb
pitr-ch: To go through https://github.com/pitr-ch/jruby/tree/bigdecimal/lib/ruby/stdlib/bigdecimal what should be loaded with `bigdecimal`. Doing it now to remove the todo.

truffle/src/main/java/org/jruby/truffle/nodes/core/BigDecimalNodes.java
pitr-ch: yep, sorry, It was just for me I forgot to remove it :/

pitr-ch:
Thanks for your comments, fixed.

lib/ruby/truffle/mri/bigdecimal/jacobian.rb
```diff
@@ -0,0 +1,87 @@
+#
```
eregon: We usually just do a require_relative to avoid the copy, see `lib/ruby/truffle/mri/abbrev.rb` for instance.
Or `require_relative '../../../stdlib/rubygems/' + File.basename(__FILE__)` if you do not want to hardcode the basename of the current file.

pitr-ch: great, thanks

Issue #2964, opened 2015-05-20
In windows ScriptingContainerTest fails


cshupp1:
In windows the test testGetCurrentDirectory in ScriptingContainer fails as follows:

```
org.junit.ComparisonFailure: expected:<C:[\Users\Cris\IdeaProjects\jruby\]core> but was:<C:[/Users/Cris/IdeaProjects/jruby/]core>
        at org.junit.Assert.assertEquals(Assert.java:115)
        at org.junit.Assert.assertEquals(Assert.java:144)
        at org.jruby.embed.ScriptingContainerTest.testGetCurrentDirectory(ScriptingContainerTest.java:1914)
```

The bug appears to be here (org.jruby.util.JRubyFile):

```
    public static String normalizeSeps(String path) {
        if (Platform.IS_WINDOWS) {
            return path.replace(File.separatorChar, '/');
        } else {
            return path;
        }
```

The order of replace is backwards, changing to:

```
   public static String normalizeSeps(String path) {
        if (Platform.IS_WINDOWS) {
            return path.replace('/', File.separatorChar);
        } else {
            return path;
        }
    }
```

With your permission my coding partner and I would like to use this simple issue to attempt our first contribution to JRuby. Is there a check in guide we should read?

Thanks,

Cris and Greg

cshupp1:
Oh the file we are proposing to change is org.jruby.util.JRubyFile

cshupp1:
Sadly, our proposed solution leads to a different test failing namely the 'testRequire' test in file 'org.jruby.test.TestKernel'.

We have a proposed fix for that too.  Consider the original method:

```
    public void testRequire() throws Exception {
        //reset the $loadTestvar
        eval("$loadTest = nil");
        assertEquals("failed to load the file test/loadTest", "0", eval("require '../test/loadTest'"));
        assertEquals("incorrectly reloaded the file test/loadTest", "", eval("require '../test/loadTest'"));

        assertEquals("incorrect value for $\" variable", "true", eval("print $\"[-1].end_with?('test/loadTest.rb')"));
    }
```

Changing to:

```
    public void testRequire() throws Exception {
        char s = File.separatorChar;
        //reset the $loadTestvar
        eval("$loadTest = nil");
        assertEquals("failed to load the file test/loadTest", "0", eval("require '../test/loadTest'"));
        assertEquals("incorrectly reloaded the file test/loadTest", "", eval("require '../test/loadTest'"));
[...]

PR #2973, opened 2015-05-22
This is a fix for bug ID 2964.  The URL to see the write up is here:

gpbowman-git:
https://github.com/jruby/jruby/issues/2964

On windows the mvn -Ptest fails due to incorrect substitution of '/' and '\'.

 See the write up for more details.

 Cris Shupp
 Greg Bowman


headius:
Thank you!

PR #2985, opened 2015-05-25
[Truffle] Add special values of BigDecimal

pitr-ch:
**not** ready for merge

### DIFF
--- a/lib/ruby/truffle/truffle/bigdecimal.rb
+++ b/lib/ruby/truffle/truffle/bigdecimal.rb
@@ -1,7 +1,67 @@
 class Truffle::BigDecimal < Numeric
   include Comparable
 
+  BASE = 10_000
+
+  SIGN_NEGATIVE_INFINITE = -3
+  SIGN_NEGATIVE_FINITE   = -2
+  SIGN_NEGATIVE_ZERO     = -1
+  SIGN_NaN               =  0
+  SIGN_POSITIVE_ZERO     =  1
+  SIGN_POSITIVE_FINITE   =  2
+  SIGN_POSITIVE_INFINITE =  3
+
+  def ==(o)
+    (self <=> o) == 0 rescue false
+  end
+
   alias_method :eql?, :==
+
+  def coerce(other)
+    [BigDecimal(other), self]
+    # TODO (pitr 28-may-2015): detect what cannot be coerced and raise?
+  end
+
+  # TODO (pitr 28-may-2015): compare with pure Java versions
+  def >(other)
+    unless comp = (self <=> other)
+      return false if nan? || (BigDecimal === other && other.nan?)
+      raise ArgumentError, "comparison of #{self.class} with #{other.class}"
+    end
+
+    comp > 0
+  end
+
+  def >=(other)
+    unless comp = (self <=> other)
+      return false if nan? || (BigDecimal === other && other.nan?)
+      raise ArgumentError, "comparison of #{self.class} with #{other.class}"
+    end
+
+    comp >= 0
+  end
+
+  def <(other)
+    unless comp = (self <=> other)
+      return false if nan? || (BigDecimal === other && other.nan?)
+      raise ArgumentError, "comparison of #{self.class} with #{other.class}"
+    end
+
+    comp < 0
+  end
+
+  def <=(other)
+    unless comp = (self <=> other)
+      return false if nan? || (BigDecimal === other && other.nan?)
+      raise ArgumentError, "comparison of #{self.class} with #{other.class}"
+    end
+
+    comp <= 0
+  end
+
+  def nonzero?
+    zero? ? nil : self
+  end
 end
 
 BigDecimal = Truffle::BigDecimal
--- a/lib/ruby/truffle/truffle/bigdecimal.rb
+++ b/lib/ruby/truffle/truffle/bigdecimal.rb
@@ -19,7 +19,6 @@ class Truffle::BigDecimal < Numeric
 
   def coerce(other)
     [BigDecimal(other), self]
-    # TODO (pitr 28-may-2015): detect what cannot be coerced and raise?
   end
 
   # TODO (pitr 28-may-2015): compare with pure Java versions
--- a/truffle/src/main/java/org/jruby/truffle/nodes/core/BigDecimalNodes.java
+++ b/truffle/src/main/java/org/jruby/truffle/nodes/core/BigDecimalNodes.java
@@ -16,18 +16,17 @@ import com.oracle.truffle.api.nodes.Node;
 import com.oracle.truffle.api.object.*;
 import com.oracle.truffle.api.source.SourceSection;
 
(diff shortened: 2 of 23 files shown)


truffle/src/main/java/org/jruby/truffle/nodes/core/BigDecimalNodes.java
```diff
@@ -110,21 +168,23 @@ public RubyBasicObject initialize(RubyBasicObject self, RubyBasicObject v) {
 
         @Specialization(guards = "isRubyString(v)")
```
chrisseaton: I would put `@TruffleBoundary` on this. The `v.toString()` is not going to compile well.

chrisseaton:
Looks great!

truffle/src/main/java/org/jruby/truffle/nodes/core/BigDecimalNodes.java
```diff
@@ -109,22 +168,25 @@ public RubyBasicObject initialize(RubyBasicObject self, RubyBasicObject v) {
         }
 
         @Specialization(guards = "isRubyString(v)")
+        @CompilerDirectives.TruffleBoundary
         public RubyBasicObject initializeFromString(RubyBasicObject self, RubyBasicObject v) {
-            // TODO (pitr 20-May-2015): add NaN, Infinity handling
+            // TODO (pitr 21-May-2015): add profile?
+            // TODO (pitr 25-May-2015): construction of positive and negative zero
             switch (v.toString()) {
```
eregon: How does this work? Is `toString()` redefined somehow for BigDecimal?

truffle/src/main/java/org/jruby/truffle/nodes/core/BigDecimalNodes.java
```diff
@@ -137,36 +199,58 @@ public AddOpNode(RubyContext context, SourceSection sourceSection) {
...
+                    (aType == Type.NEGATIVE_INFINITY && bType == Type.POSITIVE_INFINITY))
+                return createRubyBigDecimal(Type.NAN);
+
+            if (aType == Type.POSITIVE_INFINITY || bType == Type.POSITIVE_INFINITY)
+                return createRubyBigDecimal(Type.POSITIVE_INFINITY);
+
+            if (aType == Type.NEGATIVE_INFINITY || bType == Type.NEGATIVE_INFINITY)
+                return createRubyBigDecimal(Type.NEGATIVE_INFINITY);
+
+            // a,b are either normal or +-zero
```
eregon: `a or b`, right?

pitr-ch: changed to "// a and b have value NEGATIVE_ZERO or NORMAL". POSITIVE_ZERO was removed.

truffle/src/main/java/org/jruby/truffle/nodes/core/BigDecimalNodes.java
pitr-ch: In short: badly :disappointed:. Just saw https://github.com/pitr-ch/jruby/blob/bigdecimal/truffle/src/main/java/org/jruby/truffle/runtime/core/RubyBasicObject.java#L175-175 so I have to fix this. Thanks.

eregon: What you want is really `v` as a `RubyString`, right?
Then it makes sense to call `toString()` although it is not very efficient (good enough for now for debug output like this).

eregon: Ahh I missed the guard with Chris' note. I think it's clearer to have RubyString as the argument type in this case.

chrisseaton: `RubyString#toString` is the exception to that note. What we don't want to do is things like `to_s` as `toString` or `hash` as `hatched` - we don't want to map the Ruby API into the Java API like that (because those methods don't give us anywhere to store state for optimisation).

pitr-ch:
I had to add coerce specializations to other numerical classes, to make `>` work. If that's ok I'll finish rest of the comparisons and operations with special values to make this merge-able.

chrisseaton:
Looks good. Do you mean you'll add specialisations to classes like Float for comparing against BigDecimal? That doesn't sound right - as in MRI and Rubinius BigDecimal is a C extension and can't modify the other class methods like that. Look at the way Rational is handled - I think in some cases Fixnum > BigDecimal is done as BigDecimal <= Fixnum in some way.

eregon:
Yes `coerce` should be used for these cases and should already be handled in Fixnum/Bignum/Float.

eregon:
But I think @pitr-ch meant `BigDecimal < Float` for instance (in `<=>`), in which case it's fine to optimize that direction and probably even expected (to not need to call coerce and make them both BigDecimal).

pitr-ch:
I mean changes like https://github.com/jruby/jruby/pull/2985/files#diff-03ee9d17196ada218812f3319c5c3f0c. It was either raising always an exception or missing completely. So for `1 > BigDecimal("0.5")` it would raise argument error or missing node for FixnumNodes.Greater(int,BigDecimal). With this changes, it calls coerce on BigDecimal so  `1 > BigDecimal("0.5")` is translated to `BigDecimal(1) > BigDecimal("0.5")` and evaluated successfully. The other way `BigDecimal("0.5") > 1` is done usually ending up in `@Specialization(guards = "isNormal(a)") public int compare(RubyBasicObject a, int b) { return compareBigDecimal(a, getBigDecimalValue(b)); }`

eregon:
great!

pitr-ch:
This PR can be merged, assuming that I can ignore the two failing jobs on Travis.

lib/ruby/truffle/truffle/bigdecimal.rb
```diff
@@ -1,7 +1,67 @@
...
+
+  def ==(o)
+    (self <=> o) == 0 rescue false
+  end
+
   alias_method :eql?, :==
+
+  def coerce(other)
+    [BigDecimal(other), self]
+    # TODO (pitr 28-may-2015): detect what cannot be coerced and raise?
```
eregon: `BigDecimal()` should raise so that implementation should be fine.

chrisseaton:
Great - yeah those failures aren't our problem - do you want to merge it yourself? Might as well rebase instead of creating a merge commit.

eregon:
(it might be hard to rebase with the renames etc)

lib/ruby/truffle/truffle/bigdecimal.rb
pitr-ch: Right, thanks, I'll remove it.

pitr-ch:
Right I can push directly now :) I've already rebased this PR in the morning so I should not be more than few commits behind, rebase should be fine.

PR #3012, opened 2015-06-02
[Truffle] BigDecimal more features

pitr-ch:


truffle/src/main/java/org/jruby/truffle/nodes/core/BigDecimalNodes.java
```diff
@@ -676,36 +677,51 @@ public Object mult(RubyBasicObject a, RubyBasicObject b, int precision) {
...
+                        return createRubyBigDecimal(Type.POSITIVE_INFINITY);
+                    case 0:
+                        return createRubyBigDecimal(Type.NAN);
+                    case -1:
+                        return createRubyBigDecimal(Type.NEGATIVE_INFINITY);
+                    default:
+                        throw new RuntimeException();
+                }
+            } else {
+                // precision based on https://github.com/pitr-ch/jruby/blob/bigdecimal/core/src/main/java/org/jruby/ext/bigdecimal/RubyBigDecimal.java#L892-903
```
chrisseaton: Not sure think this kind of comment is a good idea - the link and line numbers will bit-rot very quickly.

Could you maybe format as:

```java
// Precision based on RubyBigDecimal#op_quo19_20 - see copyright and authorship there
```

truffle/src/main/java/org/jruby/truffle/nodes/core/BigDecimalNodes.java
```diff
@@ -713,9 +729,124 @@ public Object div(RubyBasicObject a, RubyBignum b) {
...
+                    }
+                case NEGATIVE_INFINITY:
+                    switch (getBigDecimalValue(b).signum()) {
+                        case 1:
+                            return createRubyBigDecimal(Type.NEGATIVE_ZERO);
+                        case 0:
+                        case -1:
+                            return createRubyBigDecimal(BigDecimal.ZERO);
+                    }
+            }
```
chrisseaton: This switch statement could be done as specialisations and guards - but that might be very verbose for this particular case.

pitr-ch: yeah, I was thinking about that. I not happy with the current state. When I go back to add the casting node I should be able to improve this without specialization explosion.

tool/jt.rb
```diff
@@ -148,11 +162,12 @@ def help
...
-    puts '    --asm          show assembly (implies --graal)'
-    puts '    --server       run an instrumentation server on port 8080'
-    puts '    --igv          make sure IGV is running and dump Graal graphs after partial escape (implies --graal)'
-    puts '    --jdebug       run a JDWP debug server on 8000'
+    puts '    --graal         use Graal (set GRAAL_BIN or it will try to automagically find it)'
+    puts '    --asm           show assembly (implies --graal)'
+    puts '    --server        run an instrumentation server on port 8080'
+    puts '    --igv           make sure IGV is running and dump Graal graphs after partial escape (implies --graal)'
+    puts '    --[j]debug      run a JDWP debug server on 8000'
+    puts '    --jexception[s] print java exceptions'
```
chrisseaton: Why the optional plural? I'm all for making JT as convenient as possible, but if this is a new argument can't we just pick one and go with it?

pitr-ch: It seem's to me that none of those (singular or plural) is a clear choice so I've added both. I can certainly use just one, which one would you choose?

chrisseaton: Exceptions, as it prints more than one?

pitr-ch: Yeah, r.g. in tests it can print many.

nirvdrum: In the future it'd probably be best to keep unrelated changes out as separate PRs.  I had been looking for the proposed JT changes and never would have thought to look at a BigDecimal pull request for them.

nirvdrum: FWIW, I like the options here.  The alternative is to raise a hard error on bad options.  I constantly supply `--debug` instead of `--jdebug` and don't realize I've messed up until the script has finished running.  I'd hate to pick the wrong `--jexception` form and think there were no exceptions, when in reality my bad option was just ignored.

eregon: But `--debug` is a valid Ruby option so it is conflicting!
`ruby --debug -e 'p $DEBUG' => true`
I like jdebug because it's clear it's for a debugger at the Java level, not the Ruby level.

pitr-ch: Ah you are right, I'll remove the alias.

PR #3017, opened 2015-06-04
Adopting to Redesigned Truffle Inter-operability API

jtulach:
I am in process of redesigning com.oracle.truffle.api.interop API. Once that gets into development version of Truffle, JRuby+Truffle will need to be adopted. This pull request shows what changes I had to do make things running.

### DIFF
--- a/truffle/src/main/java/org/jruby/truffle/nodes/core/TruffleInteropNodes.java
+++ b/truffle/src/main/java/org/jruby/truffle/nodes/core/TruffleInteropNodes.java
@@ -14,20 +14,10 @@ import com.oracle.truffle.api.CompilerDirectives;
 import com.oracle.truffle.api.CompilerDirectives.TruffleBoundary;
 import com.oracle.truffle.api.dsl.Specialization;
 import com.oracle.truffle.api.frame.VirtualFrame;
+import com.oracle.truffle.api.interop.ForeignAccess;
 import com.oracle.truffle.api.interop.TruffleObject;
+import com.oracle.truffle.api.nodes.Node;
 import com.oracle.truffle.api.source.SourceSection;
-import com.oracle.truffle.interop.messages.Argument;
-import com.oracle.truffle.interop.messages.Execute;
-import com.oracle.truffle.interop.messages.GetSize;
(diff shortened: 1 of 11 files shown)


truffle/src/main/java/org/jruby/truffle/nodes/core/TruffleInteropNodes.java
```diff
@@ -203,21 +193,21 @@ private static String slowPathToString(RubyString identifier) {
     @CoreMethod(names = "write_property", isModuleFunction = true, needsSelf = false, required = 3)
     public abstract static class WritePropertyNode extends CoreMethodArrayArgumentsNode {
 
-        @Child private ForeignObjectAccessNode node;
+        @Child private Node node;
 
         public WritePropertyNode(RubyContext context, SourceSection sourceSection) {
             super(context, sourceSection);
-            this.node = ForeignObjectAccessNode.getAccess(Write.create(Receiver.create(), Argument.create(), Argument.create()));
+            this.node = ForeignAccess.node(ForeignAccess.msgWrite());
```
chrisseaton: Are these node methods factories or something? The name is strange - being a noun rather than a verb.

chrisseaton:
This all looks fine. Do you want this merged now, or later?

jtulach:
My plan is to integrate the API changes on Monday. Then you'll need this PR.

truffle/src/main/java/org/jruby/truffle/nodes/core/TruffleInteropNodes.java
jtulach: Yeah, Christian Humer had similar comment. He wanted to rename node to createNode and msgXYZ to something else as well. The msgXYZ methods are however more like singleton accessors than factories.

eregon: I think as well `createNode` would be clearer if it is a factory method, `.node` could be like a getter, or possibly implies some caching (does it?).
Would an enum or constants make sense for the msgXYZ if they are singleton instances?

PR #3048, opened 2015-06-15
[Truffle] BigDecimal fixes

pitr-ch:


truffle/src/main/java/org/jruby/truffle/nodes/ext/BigDecimalNodes.java
```diff
@@ -1245,55 +1303,31 @@ public double toFSpecial(RubyBasicObject value) {
...
-    @CoreMethod(names = "split")
-    public abstract static class SplitNode extends BigDecimalCoreMethodNode {
-
-        final private ConditionProfile wasNormal;
-        @Child private CallDispatchHeadNode signCall;
-        @Child private CallDispatchHeadNode exponentCall;
-        @Child private ToIntNode signToInt;
-        @Child private ToIntNode exponentToInt;
-        @Child private IntegerCastNode signIntegerCast;
+    @CoreMethod(names = "unscaled")
```
eregon: If this method is not present in normal Ruby, it should be `private` (if possible or `protected` otherwise) and annotated with `@RubiniusOnly` (not so well named for this).

truffle/src/main/java/org/jruby/truffle/nodes/ext/BigDecimalNodes.java
```diff
@@ -999,39 +1006,91 @@ public boolean zeroSpecial(RubyBasicObject value) {
...
         @Specialization(guards = {
                 "isNormal(value)",
                 "isNormalZero(value)"})
-        public int signNormalZero(RubyBasicObject value) {
-            return 1;
+        public int signNormalZero(VirtualFrame frame, RubyBasicObject value) {
+            if (positiveZeroSign == null) {
+                CompilerDirectives.transferToInterpreterAndInvalidate();
+                positiveZeroSign = insert(createGetSignConstantNodeChain("SIGN_POSITIVE_ZERO"));
+            }
```
eregon: This seems a bit overkill for getting a constant, but it's not too easy as it's defined in Ruby.
The constants could be defined in Java, in which case we should have some hook to load BigDecimal when required.

pitr-ch: Could you elaborate on following?
> The constants could be defined in Java, in which case we should have some hook to load BigDecimal when required.

eregon: Having `private static final int SIGN_NEGATIVE_INFINITE = -3` in `BigDecimalNodes.java`, and have let's say a "setup" method / interface implementation / etc in the same file, doing like
`c = defineClass(BigDecimal); c.setConstant("SIGN_NEGATIVE_INFINITE",  SIGN_NEGATIVE_INFINITE)`.
It would solve the pre-loading problem and the name mismatch (`Truffle::BigDecimal`).

eregon: Similar to JRuby's `Library.load`.

pitr-ch: Sounds good, but I'll postpone that for later. I would like to get more spec passing first.

PR #3064, opened 2015-06-18
[Truffle] Rbx 2.5.6 upgrade

nirvdrum:
There's quite a bit here.  A lot of it is mechanical, but to summarize:

* Bootstrap files are largely kept the same they were
* Core files are 100% imports from Rubinius
* api/shims now provide method overrides for changes we used to make in place in the core files; api/shims require order has been altered suitably to make sure methods are overridden
* Any core changes that can't be overridden are wrapped with a Truffle.omit call with a reason why; this call is replaced with a literal nil at runtime
* BodyTranslator and ModuleTranslator were updated to provide parse time rewrites for some things we used to do manually
* Parse warnings are disabled in the lexer & parser rather than through source modifications
* Miscellaneous clean-ups along the way

I haven't yet untagged any specs this would make passing.  I'll do that after the merge if you guys don't have any objection.

Ping @chrisseaton, @eregon, @pitr-ch, @bjfish.


lib/ruby/truffle/rubysl/rubysl-timeout/lib/rubysl/timeout/timeout.rb
```diff
@@ -122,6 +122,7 @@ def self.watch_channel
 
   # Spin up the thread up front to avoid a thread collision problem spinning
   # it up lazily.
+  puts caller
```
bjfish: Remove debug puts?

nirvdrum: Thanks.  I had removed it, then needed it again, and forgot to remove it the second time :-/

chrisseaton:
Don't forget [Truffle] in issues and PRs.

truffle/src/main/ruby/core.rb
```diff
@@ -6,6 +6,14 @@
 # GNU General Public License version 2
 # GNU Lesser General Public License version 2.1
 
+module Truffle
+  # Utility method for commenting out part of Rubinius's implementation and detailing why.  It helps clearly
+  # demarcate things we want to omit from things Rubinius has left commented out and as such, should help compare
+  # diffs when upgrading Rubinius core files.
+  def self.omit(reason)
+  end
+end
```
chrisseaton: Is this just for documentation?

truffle/src/main/ruby/core/rubinius/common/main.rb
```diff
@@ -34,6 +35,22 @@ def include(*mods)
   # Truffle: Do not define #public, #private in Ruby since they need
   # to set the top-level frame visibility when called with no argument.
```
chrisseaton: Aren't we ok now we don't overwrite Java methods with Ruby ones? So are the comments below still required?

chrisseaton:
Looks good

truffle/src/main/ruby/core.rb
nirvdrum: Yeah. The two cases I wanted to handle are:

1) Delineating stuff Rubinius has commented out from stuff we have commented out.
2) Annotating things we've essentially deleted (just deleting them means that needs to be reevaluated each time we sync upstream).

So, it's essentially documentation for why we can't run something while keeping the diff between different Rubinius files minimal.  It's also easy to grep for.

truffle/src/main/ruby/core/rubinius/common/main.rb
nirvdrum: I'll have to look this one over again.  At the very least, they should be wrapped in `Truffle.omit` rather than commented out.

truffle/src/main/ruby/core.rb
chrisseaton: Sorry no I mean what's the point of this definition - it's handled in translator so we don't need Ruby code do we?

nirvdrum: Oh. I had it there before the translator stuff was written and it seemed more Ruby-like to keep it. I can delete it though.

nirvdrum: Re: "more Ruby-like" ... we could just have our own top-level methods that look more like keywords.  I figured this approach looked more like code written in Ruby and handling it in the translator was an optimization pass.

nirvdrum:
I merged, but I'm still open to feedback.

truffle/src/main/java/org/jruby/truffle/translator/BodyTranslator.java
```diff
@@ -415,6 +415,13 @@ public RubyNode visitCallNode(CallNode node) {
                 && ((org.jruby.ast.Colon2ConstNode) node.getReceiverNode()).getName().equals("Primitive")
                 && node.getName().equals("assert_not_compiled")) {
             return AssertNotCompiledNodeGen.create(context, sourceSection);
+        } else if (node.getReceiverNode() instanceof org.jruby.ast.ConstNode
+                && ((ConstNode) node.getReceiverNode()).getName().equals("Truffle")) {
+            if (node.getName().equals("omit")) {
+                // We're never going to run the omitted code and it's never used as the RHS for anything, so just
+                // replace the call with nil.
+                return new LiteralNode(context, sourceSection, context.getCoreLibrary().getNilObject());
+            }
```
eregon: We do need to extract some local vars in the Translator, it becomes really hard to read.

truffle/src/main/ruby/core/rubinius/alpha.rb
```diff
@@ -43,3 +43,13 @@ def include(mod)
   end
 
 end
+
+module Kernel
+
+  # Rubinius defines this method differently, using the :object_class primitive.  The two primitives are very similar,
```
eregon: What's the difference between the two?

nirvdrum: `:vm_object_class` takes the object to get the class for.  `:object_class` always uses `self`.

truffle/src/main/ruby/core/rubinius/api/shims/mirror.rb
```diff
@@ -0,0 +1,36 @@
...
+# redistribute it and/or modify it under the terms of the:
+#
+# Eclipse Public License version 1.0
+# GNU General Public License version 2
+# GNU Lesser General Public License version 2.1
+
+module Rubinius
+  class Mirror
+
+    # Implementation of the :module_mirror Rubinius primitive.
```
eregon: How does this relate to the RubiniusPrimitive implemented above?

eregon:
Looks good, thank you for the hard work!

It would be nice to have some simple stats on parse time for `core:/`. So we could see how these kind of changes affect it (Should be fairly minimal in this case but still interesting).

truffle/src/main/ruby/core/rubinius/api/shims/mirror.rb
nirvdrum: I'm not sure what you're asking.  But `:module_mirror` is used by Rubinius to get the mirror for an object.  We had an incomplete implementation in Java.  This is a complete implementation written in Ruby.  The Java one now just calls this code.

truffle/src/main/ruby/core/rubinius/alpha.rb
eregon: I see, would be good to have it as part of the comment for curious guys like me :)

truffle/src/main/ruby/core/rubinius/api/shims/mirror.rb
eregon: Do we still need the Java one then?

nirvdrum: I think Chris wants me to switch back to pure Java, so this may be moot.  But we would need the Java one to dispatch from the primitive manager currently.  The manager could be updated of course to avoid the extra level of indirection, but it would probably complicate things.

PR #3066, opened 2015-06-18
Create Truffle specific main function.

bbelleville:


chrisseaton:
Thanks for this. A couple of general comments:

* 4-spaces instead of tabs, please
* Comments with a capital letter (sorry to be picky)
* The licence header you need is:

```java
/*
 * Copyright (c) 2015 Oracle and/or its affiliates. All rights reserved. This
 * code is released under a tri EPL/GPL/LGPL license. You can use it,
 * redistribute it and/or modify it under the terms of the:
 *
 * Eclipse Public License version 1.0
 * GNU General Public License version 2
 * GNU Lesser General Public License version 2.1
 */
```

@enebo @headius this simpler Truffle-specifc `main` method is designed to support some research @bbelleville is doing with our ahead-of-time-compiler - we're not planning on separating from the real JRuby main method or launcher or anything like that apart from for this research.

truffle/src/main/java/org/jruby/truffle/Main.java
```diff
@@ -0,0 +1,105 @@
...
+                }
+                runtime.getGlobalVariables().set(
+                        "$" + entry.getKey().toString(), varvalue);
+            }
+
+            RootNode scriptNode = (RootNode) runtime
+                    .parseFromMain(filename, in);
+
+            // If no DATA, we're done with the stream, shut it down
+			if (runtime.fetchGlobalConstant("DATA") == null) {
```
eregon: I see tabs around here

truffle/src/main/java/org/jruby/truffle/Main.java
```diff
@@ -0,0 +1,105 @@
...
+        }
+
+        if (in == null) {
+            return;
+        } else {
+
+            // Global variables
+            IAccessor d = new ValueAccessor(runtime.newString(filename));
+            runtime.getGlobalVariables().define("$PROGRAM_NAME", d,
+                    GlobalVariable.Scope.GLOBAL);
```
eregon: Slightly long lines are OK if it makes the call simpler to read (like in this case)

truffle/src/main/java/org/jruby/truffle/Main.java
```diff
@@ -0,0 +1,105 @@
...
+            runtime.getGlobalVariables().define("$0", d,
+                    GlobalVariable.Scope.GLOBAL);
+
+            for (Iterator i = config.getOptionGlobals().entrySet().iterator(); i.hasNext();) {
```
eregon: Why is this not an enhanced foreach?

truffle/src/main/java/org/jruby/truffle/Main.java
chrisseaton: Yeah I went back and fixed.

eregon:
I'll address my comments in a commit.

truffle/src/main/java/org/jruby/truffle/Main.java
eregon: Because getOptionGlobals is a raw Map.

truffle/src/main/java/org/jruby/truffle/Main.java
eregon: Fixed by @chrisseaton in a later commit.

PR #3073, opened 2015-06-19
dir globs on uri classloader without the need of .jrubydir

mkristian:


mkristian:
@kares @ratnikov would you both have a look at those commits: https://github.com/jruby/jruby/commit/97e35b89b92361baf25661d39218e84b3f9eacb0 and https://github.com/jruby/jruby/commit/d02e9ed14c4552a1d7256933336591dbe841062c

which is basically switching off the old LoadService and removing obsolete code.

kares:
looks good to me, but I'm not proficient in covering all possible cases https://github.com/jruby/jruby/commit/97e35b89b92361baf25661d39218e84b3f9eacb0 will face in real-world ... it's "only" 9K thus experimenting should be fajn

PR #3076, opened 2015-06-22
[Truffle] Allow Rubinius primitives to be written entirely in Ruby

chrisseaton:
@eregon @nirvdrum @pitr-ch please review and sign off when you are done.


truffle/src/main/ruby/core/rubinius/api/shims/mirror.rb
```diff
@@ -31,6 +30,7 @@ def self.module_mirror(obj)
         end
       end
     end
+    Truffle::Primitive.install_rubinius_primitive self, :module_mirror
```
chrisseaton: I'm not sure how to get `self` in a `@CoreMethod` anymore - `onSingleton` and `isModuleFunction` both set `needsSelf` to false.

eregon: In this particular case, it's intended to not be obvious, since it would be likely wrong with a `module function` (defined on both instance and class level) as self could be either a module or the surrounding self.
If it is just a singleton/class method, then it might be right to want it and currently there is `constuctor` for this (which is literally `onSingleton + needsSelf`), which was appropriately named when I looked at current use-cases.

truffle/src/main/java/org/jruby/truffle/translator/BodyTranslator.java
```diff
@@ -495,21 +522,46 @@ private RubyNode translateRubiniusInvokePrimitive(SourceSection sourceSection, C
...
+            }
+
+            return new RubyCallNode(
+                    context,
+                    sourceSection,
+                    callConstructor.getMethod(),
+                    new LiteralNode(context, sourceSection, callConstructor.getModule()),
+                    null,
+                    false,
+                    arguments.toArray(new RubyNode[arguments.size()]));
```
eregon: Can we share this with above so the code is not duplicated?

truffle/src/main/java/org/jruby/truffle/translator/BodyTranslator.java
```diff
@@ -452,28 +453,54 @@ private RubyNode translateRubiniusPrimitive(SourceSection sourceSection, CallNod
...
+                    environment.getReturnID());
+        } else if (primitive instanceof RubiniusPrimitiveCallConstructor) {
+            final RubiniusPrimitiveCallConstructor callConstructor = (RubiniusPrimitiveCallConstructor) primitive;
 
-        return new CallRubiniusPrimitiveNode(context, sourceSection,
-                primitive.getFactory().createNode(context, sourceSection, arguments.toArray(new RubyNode[arguments.size()])),
-                environment.getReturnID());
+            final List<RubyNode> arguments = new ArrayList<>();
+
+            final int argumentsCount = callConstructor.getModule().getMethods().get(callConstructor.getMethod()).getSharedMethodInfo().getArity().getRequired();
```
eregon: Maybe split this in two lines, 1 for the method and 1 for the required args?

eregon:
Sounds like a nice addition.
Maybe we could take a (bound) Method as argument and passing the primitive name?
And additionally call the method directly (or its CallTarget) so even if it is overwritten the behavior stays identical.
This would allow to do something like
```ruby
Truffle::Primitive.install_rubinius_primitive :vm_object_class, Kernel.method(:class)
```
Which would solve the currently not very useful primitives just delegating to something else and possible monkey-patching issues.

truffle/src/main/ruby/core/rubinius/api/shims/mirror.rb
chrisseaton: Yes I thought that `self` was the module when it isn't. I think I will use some magic to get the self of the caller, since this slow path stuff.

truffle/src/main/java/org/jruby/truffle/translator/BodyTranslator.java
chrisseaton: It's only the arguments array that can be shared. I didn't do it originally as the body of the loop is a bit different, but I can probably do a second pass to handle that.

truffle/src/main/java/org/jruby/truffle/translator/BodyTranslator.java
chrisseaton: Will do.

chrisseaton:
Yes that's a good idea.

The main motivation for this was to start writing some primitives using Java interop. `vm_time` and `vm_gc` are both early candidates for this as they just do a single Java static call.

nirvdrum:
This looks good.  However, I am concerned about the ability to find these primitive methods.  While having a node that serves only to delegate to a RubyCallNode seems superfluous at face value, it does have the benefit of being consistent with primitives written in Java.  I.e., they can all be found in an expected location.

If we want to go this route, I propose introducing a new directory for implemented mirror methods.  `api/shims` may have made sense for that one case at that time, but I think could stand from better organization now.  This would still require searches in two places, but at least they're two well-known locations.

I would maybe even take it a step further and have the translator look at the source section to ensure Truffle::Primitive.install_rubinius_primitive can only be used from this directory, throwing an error otherwise.  That might be a bit heavy-handed (particularly for a lint tool), but it would help keep us honest.

chrisseaton:
A new directory for mirror methods? Do you mean a directory for primitive methods?

nirvdrum:
Sorry, yes, for primitives.  The mirror stuff I think could be tidied up a bit too, but that's a separate issue.

eregon:
I agree, but maybe a file is enough? It should be loaded just before alpha.rb I think.

nirvdrum:
A single file is fine with me.  I just don't want that to become unwieldly.  Ruby is much more concise than Java for this though, so it may not be a warranted concern.

eregon:
I'll do the change to make it take a Ruby Method.

pitr-ch:
:+1: looks good to me

eregon:
I merged with a few changes to accept (bound) Method.

I tried as well with UnboundMethod, but it does not fit for `vm_` primitives since their self is ignored, while the corresponding Kernel methods use `self`.

PR #3077, opened 2015-06-22
[Truffle] BigDecimal: add handling of ExceptionMode, Limit, Rounding

pitr-ch:
* Add coerce and cast node for BigDecimal
* Add create node (single entry point for BigDecimal creation)
* BigDecimalCoreMethodNode has helpers for: BigDecimal creation, getLimit
  getRoudingMode.
* BigDecimalCoerceNode is used to reduce number of Specializations in operation
  nodes
* Operation nodes have beter structure and share code
* add, sub, mult, div are handling special values
* All operations respect fiber local ExceptionMode, Limit, Rounding settings
* #round method added


spec/ruby/library/bigdecimal/round_spec.rb
```diff
@@ -187,4 +187,10 @@
       @n2_49.round(0, BigDecimal::ROUND_HALF_EVEN).should == @neg_two
     end
   end
+
+  it 'raise exception, if self is special value' do
+    lambda { BigDecimal('NaN').round }.should raise_error(FloatDomainError)
+    lambda { BigDecimal('Infinity').round }.should raise_error(FloatDomainError)
+    lambda { BigDecimal('-Infinity').round }.should raise_error(FloatDomainError)
+  end
```
pitr-ch: @eregon What are the requirements for adding specs?

nirvdrum: You can commit specs directly to the repository, but they should be in isolated commits.  That way @eregon can cherry-pick them back to the upstream repo.  It gets a lot harder if you've mixed the spec changes in with other changes unrelated to the rubyspec project.

truffle/src/main/java/org/jruby/truffle/nodes/ext/BigDecimalNodes.java
```diff
@@ -122,6 +133,28 @@ public static RubyBasicObject createRubyBigDecimal(RubyClass rubyClass, BigDecim
...
+            case 5:
+                return RoundingMode.CEILING;
+            case 6:
+                return RoundingMode.FLOOR;
+            case 7:
+                return RoundingMode.HALF_EVEN;
+        }
+
+        CompilerAsserts.neverPartOfCompilation();
+        throw new UnsupportedOperationException();
```
nirvdrum: I'd add a message in there to make debugging easier.

nirvdrum: If you want this to never be part of compilation, you might just want to add a `@TruffleBoundary` on the method.  But I suspect you only want the failure case to not be part of compilation. In which case I think you might want to just do a transfer.

truffle/src/main/java/org/jruby/truffle/nodes/ext/BigDecimalNodes.java
```diff
@@ -178,175 +217,327 @@ public static boolean isNan(RubyBasicObject value) {
             return getBigDecimalType(value) == Type.NAN;
         }
 
+        private void setupCreateBigDecimal() {
+            if (createBigDecimal == null)
+                createBigDecimal = insert(CreateBigDecimalNodeFactory.create(getContext(), getSourceSection(), null, null));
```
nirvdrum: I'll let @chrisseaton and @eregon weigh in, but I I think everything we've written up until now (that wasn't directly ported from elsewhere) always uses curly braces.  I'm personally not a fan of relying on indentation in a language that doesn't enforce it, and as such would vote in favor of always using braces.

truffle/src/main/java/org/jruby/truffle/nodes/ext/BigDecimalNodes.java
```diff
@@ -178,175 +217,327 @@ public static boolean isNan(RubyBasicObject value) {
...
+        }
+
+        protected RubyBasicObject createBigDecimal(VirtualFrame frame, Object value, RubyBasicObject self) {
+            setupCreateBigDecimal();
+            return createBigDecimal.executeCreate(frame, value, self);
+        }
+
+        private void setupLimitCall() {
+            if (limitCall == null)
+                limitCall = insert(DispatchHeadNodeFactory.createMethodCall(getContext()));
```
nirvdrum: You can't modify the graph when code is hot.  You should insert a `CompilerDirectives.transferToInterpreter();` before the insert line.  That should happen in each of the cases here that it's missing.

truffle/src/main/java/org/jruby/truffle/nodes/ext/BigDecimalNodes.java
```diff
@@ -178,175 +217,327 @@ public static boolean isNan(RubyBasicObject value) {
...
+        protected RoundingMode getRoundMode(VirtualFrame frame) {
+            setupRoundModeCall();
+            setupRoundModeIntegerCast();
+
+            return toRoundingMode(roundModeIntegerCast.executeInteger(frame,
+                    // TODO (pitr 21-Jun-2015): read the actual constant
+                    roundModeCall.call(frame, getBigDecimalClass(), "mode", null, 256)));
+        }
+
+        @Deprecated
```
nirvdrum: I think you're the only one that's written code using `BigDecimal` yet.  If the deprecation is helpful for you, by all means keep it.  But if not, just delete the deprecated code.

truffle/src/main/java/org/jruby/truffle/nodes/ext/BigDecimalNodes.java
```diff
@@ -178,175 +217,327 @@ public static boolean isNan(RubyBasicObject value) {
...
+            setBigDecimalValue(self, value);
+            return self;
+        }
+
+        // FIXME (pitr 21-Jun-2015): raise on underflow
+
+        private RubyBasicObject createWithMode(VirtualFrame frame, Type value, RubyBasicObject self,
+                                               String constantName, String errorMessage) {
+            if (modeCall == null)
+                modeCall = insert(DispatchHeadNodeFactory.createMethodCall(getContext()));
```
nirvdrum: Transfers here as well.

truffle/src/main/java/org/jruby/truffle/nodes/ext/BigDecimalNodes.java
```diff
@@ -354,126 +545,144 @@ public Object addSpecial(RubyBasicObject a, RubyBasicObject b) {
             else
                 return b;
         }
-
-        // TODO (pitr 28-may-2015): should it support coerce?
     }
 
-    @CoreMethod(names = "add", required = 2)
-    public abstract static class AddNode extends BigDecimalCoreMethodNode {
+    @CoreMethod(names = {"+"}, required = 1)
```
nirvdrum: Unfortunately, I don't think you were around when we discussed some code style stuff.  If there's only one method name, we've opted to omit the braces.  If the braces are used, the strings are space padded (`{ "+" }`).

truffle/src/main/java/org/jruby/truffle/nodes/ext/BigDecimalNodes.java
```diff
@@ -354,126 +545,144 @@ public Object addSpecial(RubyBasicObject a, RubyBasicObject b) {
...
-        @Specialization(guards = "isNormal(a)")
-        public Object add(RubyBasicObject a, double b, int precision) {
-            return addBigDecimal(a, getBigDecimalValue(b), precision);
-        }
+    }
 
-        @Specialization(guards = {"isNormal(a)", "isRubyBignum(b)"})
-        public Object addBignum(RubyBasicObject a, RubyBasicObject b, int precision) {
-            return addBigDecimal(a, getBignumBigDecimalValue(b), precision);
+    @CoreMethod(names = {"add"}, required = 2)
```
nirvdrum: Remove the braces here as well.

truffle/src/main/java/org/jruby/truffle/nodes/ext/BigDecimalNodes.java
```diff
@@ -354,126 +545,144 @@ public Object addSpecial(RubyBasicObject a, RubyBasicObject b) {
...
             super(context, sourceSection);
         }
 
-        @TruffleBoundary
-        private Object addBigDecimal(RubyBasicObject a, BigDecimal b, int precision) {
-            return createRubyBigDecimal(getBigDecimalValue(a).add(b, new MathContext(precision)));
+        @Specialization(guards = {
+                "isNormal(a)",
+                "isRubyBigDecimal(b)",
+                "isNormal(b)"})
```
nirvdrum: It may not matter much here, but the predicates are evaluated in order and short-circuit.  Assuming `isNormal` has uniform cost, both calls are either faster than or slower than the `isRubyBigDecimal` check and should probably be ordered as such.

truffle/src/main/java/org/jruby/truffle/nodes/ext/BigDecimalNodes.java
```diff
@@ -482,194 +691,113 @@ public NegNode(RubyContext context, SourceSection sourceSection) {
...
-            return createRubyBigDecimal(getBigDecimalValue(a).multiply(b));
-        }
+        private Object multBigDecimal(RubyBasicObject a, RubyBasicObject b, MathContext mathContext) {
+            final BigDecimal bBigDecimal = getBigDecimalValue(b);
 
-        @Specialization(guards = "isNormal(a)")
-        public Object mult(RubyBasicObject a, long b) {
-            return multBigDecimal(a, getBigDecimalValue(b));
-        }
+            if (zeroNormal.profile(isNormalZero(a) && bBigDecimal.signum() == -1))
```
nirvdrum: I don't think the profile makes much sense with the boundary attached to the method.  Should it be compiled?

truffle/src/main/java/org/jruby/truffle/nodes/ext/BigDecimalNodes.java
```diff
@@ -677,218 +805,297 @@ public Object multSpecial(RubyBasicObject a, RubyBasicObject b) {
...
             super(context, sourceSection);
         }
 
-        @TruffleBoundary
-        private Object mulBigDecimal(RubyBasicObject a, BigDecimal b, int precision) {
-            return createRubyBigDecimal(getBigDecimalValue(a).multiply(b, new MathContext(precision)));
+        @Specialization(guards = {
+                "isNormal(a)",
+                "isRubyBigDecimal(b)",
+                "isNormal(b)"})
```
nirvdrum: Same guard ordering comment.

truffle/src/main/java/org/jruby/truffle/nodes/ext/BigDecimalNodes.java
```diff
@@ -677,218 +805,297 @@ public Object multSpecial(RubyBasicObject a, RubyBasicObject b) {
 
     }
 
-    @CoreMethod(names = "mult", required = 2)
-    public abstract static class MultNode extends BigDecimalCoreMethodNode {
+    @CoreMethod(names = {"*"}, required = 1)
```
nirvdrum: Remove the braces here, too.

truffle/src/main/java/org/jruby/truffle/nodes/ext/BigDecimalNodes.java
```diff
@@ -677,218 +805,297 @@ public Object multSpecial(RubyBasicObject a, RubyBasicObject b) {
...
         }
 
         @TruffleBoundary
-        private Object div(RubyBasicObject a, BigDecimal b) {
-            if (normalZero.profile(b.signum() == 0)) {
-                switch (getBigDecimalValue(a).signum()) {
+        private Object divBigDecimal(RubyBasicObject a, RubyBasicObject b, MathContext mathContext) {
+            final BigDecimal aBigDecimal = getBigDecimalValue(a);
+            final BigDecimal bBigDecimal = getBigDecimalValue(b);
+            if (normalZero.profile(bBigDecimal.signum() == 0)) {
```
nirvdrum: Same comment about using profiles in methods with boundaries.

truffle/src/main/java/org/jruby/truffle/nodes/ext/BigDecimalNodes.java
```diff
@@ -677,218 +805,297 @@ public Object multSpecial(RubyBasicObject a, RubyBasicObject b) {
...
+                if (aType == Type.POSITIVE_INFINITY) return createBigDecimal(frame, Type.NEGATIVE_INFINITY);
+                else return createBigDecimal(frame, Type.POSITIVE_INFINITY);
 
             // a and b are only +-Infinity
-            return createRubyBigDecimal(Type.NAN);
+            return createBigDecimal(frame, Type.NAN);
+        }
+    }
+
+    @CoreMethod(names = {"/", "quo"}, required = 1)
```
nirvdrum: Spaces between the brace and the quotation mark.  We still have code that doesn't have the space.   It's not critical, but we're trying to normalize moving forward.

truffle/src/main/java/org/jruby/truffle/nodes/ext/BigDecimalNodes.java
```diff
@@ -1180,8 +1392,62 @@ public RubyBasicObject absSpecial(RubyBasicObject value) {
...
+                    throw new RaiseException(getContext().getCoreLibrary().
+                            floatDomainError("Computation results to 'Infinity'", this));
+                case NEGATIVE_ZERO:
+                    return value;
+                case NAN:
+                    CompilerDirectives.transferToInterpreter();
+                    throw new RaiseException(getContext().getCoreLibrary().
+                            floatDomainError("Computation results to 'NaN'(Not a Number)", this));
+            }
+            CompilerAsserts.neverPartOfCompilation();
```
nirvdrum: I don't think this is what you want.  I could be mistaken though.

truffle/src/main/java/org/jruby/truffle/nodes/ext/BigDecimalNodes.java
```diff
@@ -1180,8 +1392,62 @@ public RubyBasicObject absSpecial(RubyBasicObject value) {
...
+            super(context, sourceSection);
+        }
+
+        @TruffleBoundary
+        private BigDecimal round(RubyBasicObject value, int digit, RoundingMode roundingMode) {
+            final BigDecimal valueBigDecimal = getBigDecimalValue(value);
+            final int end = -valueBigDecimal.scale();
+            final int start = valueBigDecimal.precision() + end;
+            final int newPrecision = start - digit;
+            return valueBigDecimal.round(new MathContext(newPrecision, roundingMode));
```
nirvdrum: Small style suggestion, but adding some vertical spacing (e.g., between the declarations and the return statement) would help a lot with legibility.

truffle/src/main/java/org/jruby/truffle/nodes/ext/BigDecimalNodes.java
```diff
@@ -1379,4 +1645,92 @@ public int toISpecial(RubyBasicObject value) {
...
+        @Specialization(guards = {"isRubyBigDecimal(value)", "isNormal(value)"})
+        public BigDecimal doBigDecimal(RubyBasicObject value) {
+            return (BigDecimal) VALUE_PROPERTY.get(value.getDynamicObject(), true);
+        }
+
+        @Fallback
+        public RubyBasicObject doBigDecimalFallback(Object value) {
+            return nil();
+        }
+        // TODO (pitr 22-Jun-2015): How to better communicate failure without throwing
```
nirvdrum: Why don't you want to raise an exception?

truffle/src/main/java/org/jruby/truffle/nodes/ext/BigDecimalNodes.java
```diff
@@ -1379,4 +1645,92 @@ public int toISpecial(RubyBasicObject value) {
...
+        }
+
+        @Specialization(guards = "isRubyBignum(value)")
+        public BigDecimal doBignum(RubyBasicObject value) {
+            return new BigDecimal(BignumNodes.getBigIntegerValue(value));
+        }
+
+        @Specialization(guards = {"isRubyBigDecimal(value)", "isNormal(value)"})
+        public BigDecimal doBigDecimal(RubyBasicObject value) {
+            return (BigDecimal) VALUE_PROPERTY.get(value.getDynamicObject(), true);
```
nirvdrum: I'd look at the pattern @chrisseaton has been using with other basic objects.  He has a set of static methods in the `<ObjectType>Nodes` class.  E.g., `PointerNodes` has static methods for accessing `Pointer` fields from a basic object.

truffle/src/main/java/org/jruby/truffle/nodes/ext/BigDecimalNodes.java
```diff
@@ -1379,4 +1645,92 @@ public int toISpecial(RubyBasicObject value) {
...
+        }
+
+        public static BigDecimalCoerceNode create(RubyContext context, SourceSection sourceSection, RubyNode value) {
+            return BigDecimalCoerceNodeGen.create(context, sourceSection, value,
+                    BigDecimalCastNodeGen.create(context, sourceSection, null));
+        }
+
+        private void setupCreateBigDecimal() {
+            if (createBigDecimal == null)
+                createBigDecimal = insert(CreateBigDecimalNodeFactory.create(getContext(), getSourceSection(), null, null));
```
nirvdrum: You'll need a transfer here.

truffle/src/main/java/org/jruby/truffle/nodes/ext/BigDecimalNodes.java
```diff
@@ -1379,4 +1645,92 @@ public int toISpecial(RubyBasicObject value) {
...
+        }
+        // TODO (pitr 22-Jun-2015): How to better communicate failure without throwing
+    }
+
+    /**
+     * Coerces a value into a BigDecimal.
+     */
+    @NodeChildren({
+            @NodeChild(value = "value", type = RubyNode.class),
+            @NodeChild(value = "cast", type = BigDecimalCastNode.class, executeWith = "value")
```
nirvdrum: I'm not familiar with `executeWith`.  I guess I have some reading to do.

spec/ruby/library/bigdecimal/round_spec.rb
pitr-ch: Ok I'll split

lib/ruby/truffle/truffle/bigdecimal.rb
```diff
@@ -6,10 +6,68 @@ class Truffle::BigDecimal < Numeric
...
+  ROUND_HALF_UP   = 3
+  ROUND_HALF_DOWN = 4
+  ROUND_CEILING   = 5
+  ROUND_FLOOR     = 6
+  ROUND_HALF_EVEN = 7
+
+  def self.mode(key, value = nil)
+    raise ArgumentError, 'requires key to be Fixnum' unless key.is_a? Fixnum
+    if key == ROUND_MODE
+      Thread.current[:'BigDecimal.rounding_mode'] ||= 3
```
chrisseaton: Maybe we want thread-local variables that are private to Truffle to avoid them leaking into the application?

truffle/src/main/java/org/jruby/truffle/nodes/ext/BigDecimalNodes.java
pitr-ch: Ok IDE reconfigured to scream on me next time :)

chrisseaton: Yes, curly braces please. Our coding style comes from the Graal CheckStyle rules, which we don't enforce anymore but would still like to vaguely follow.

chrisseaton:
Wow BigDecimal is complicated. When I suggested it as something to start on I thought it would be just as simple as wrapping Java's BigDecimal.

truffle/src/main/java/org/jruby/truffle/nodes/ext/BigDecimalNodes.java
pitr-ch: Shouldn't there be `CompilerDirectives.transferToInterpreterAndInvalidate();`? I had it there but it seems that all the graph manipulation method have this call on the first line so I've thought I'll remove the duplication.

truffle/src/main/java/org/jruby/truffle/nodes/ext/BigDecimalNodes.java
pitr-ch: yeah it'll be removed soon when I fix `<=>` not to use it. I was just for me so I see the call easier.

truffle/src/main/java/org/jruby/truffle/nodes/ext/BigDecimalNodes.java
chrisseaton: But like exceptions that doesn't cover the `DispatchHeadNodeFactory.createMethodCall(getContext())` part, or the method call to `insert`. We want to transfer immediately when we find we have to and not have any of the slow path code in the compiled part.

We've been using just `transfer` in the caller. The `insert` does the `invalidate`.

truffle/src/main/java/org/jruby/truffle/nodes/ext/BigDecimalNodes.java
pitr-ch: `isNormal` requires `b` to be `RubyBigDecimal` so it could fail with the order switched. Is this ok or should `isNormal` be fixed to never fail?

truffle/src/main/java/org/jruby/truffle/nodes/ext/BigDecimalNodes.java
pitr-ch: Removed.

truffle/src/main/java/org/jruby/truffle/nodes/ext/BigDecimalNodes.java
pitr-ch: Good catch only the `BigDecimal#multiply` should be behind the boundary. Extracted.

truffle/src/main/java/org/jruby/truffle/nodes/ext/BigDecimalNodes.java
pitr-ch: Extracted.

truffle/src/main/java/org/jruby/truffle/nodes/ext/BigDecimalNodes.java
pitr-ch: Sure I'll fix it, I am :+1: for uniformity. So it should be `@CoreMethod(names = { "/", "quo" }, required = 1)`?

truffle/src/main/java/org/jruby/truffle/nodes/ext/BigDecimalNodes.java
pitr-ch: Sure added.

truffle/src/main/java/org/jruby/truffle/nodes/ext/BigDecimalNodes.java
pitr-ch: This branch of code should be never reached. I've used this since I found it elsewhere in our code. What is the preferred way how to handle this?

truffle/src/main/java/org/jruby/truffle/nodes/ext/BigDecimalNodes.java
pitr-ch: Good point, It was living in other file and I forgot to fix it after it was moved :/

truffle/src/main/java/org/jruby/truffle/nodes/ext/BigDecimalNodes.java
pitr-ch: This relates to `executeWith`, I can then use the nil value in `BigDecimalCoerceNode`:
```
@Specialization(guards = {"isRubyBigDecimal(value)", "isNil(cast)"})
public RubyBasicObject doBigDecimal(RubyBasicObject value, RubyBasicObject cast) {
    return value;
}
```

lib/ruby/truffle/truffle/bigdecimal.rb
pitr-ch: yeah but we found out with @eregon in Linz that CRuby leaks the value under the same exact name :) Would it still make sense that we'll hide it?

truffle/src/main/java/org/jruby/truffle/nodes/ext/BigDecimalNodes.java
pitr-ch: Ah right. I'll add transfers.

lib/ruby/truffle/truffle/bigdecimal.rb
chrisseaton: Ah no it's correct as it is then.

truffle/src/main/java/org/jruby/truffle/nodes/ext/BigDecimalNodes.java
chrisseaton: I think @nivrdrum just means to add a basic message to the exception.

truffle/src/main/java/org/jruby/truffle/nodes/ext/BigDecimalNodes.java
pitr-ch: Spaces added, I like it :) I thought it supposed to be no space.

pitr-ch:
@chrisseaton I though so too but at least I got to learn on it :)

Thanks for reviews and sorry for the monstrous commit, I've tried to split it but I've let it to snowball too much.

truffle/src/main/java/org/jruby/truffle/nodes/ext/BigDecimalNodes.java
pitr-ch: Hm, makes sense but I would be repeat the text quite often. What about a new Exception class for this case? Or is there already one?

truffle/src/main/java/org/jruby/truffle/nodes/ext/BigDecimalNodes.java
pitr-ch: Is there a reason why the generated `create` factory methods do not have 'transferToInterpreter` in them?

chrisseaton: I think maybe they could do, but I prefer to be really explicit with transfers. We don't want to let work leak from one side of a transfer to another by trying to be to magic about it. I prefer the explicit clean break that the directive gives us.

truffle/src/main/java/org/jruby/truffle/nodes/ext/BigDecimalNodes.java
nirvdrum: It's fine, although it might be worthwhile to codify that in a new guard, such as `isNormalRubyBigDecimal` that then does both the checks you care about.

truffle/src/main/java/org/jruby/truffle/nodes/ext/BigDecimalNodes.java
nirvdrum: I would like to see a message in the `UnsupporteOperationException` to help with debugging.  But the `neverPartOfCompilation` is the part I was questioning.  I didn't know it could float in a method like that.

spec/ruby/library/bigdecimal/round_spec.rb
eregon: The split is done automatically, but it's niver to split because then the commit message tend to be right and I don't have to remove [Truffle] and impl details unrelated to the spec.

truffle/src/main/java/org/jruby/truffle/nodes/ext/BigDecimalNodes.java
eregon: We want to omit the branch entirely, and inlining the `create` or `insert` might take some time or the inlining decision could go wrong.

truffle/src/main/java/org/jruby/truffle/nodes/ext/BigDecimalNodes.java
eregon: I'd rather keep natural order unless it is proven to be perf sensitive as it really improves readability to draw the lattice in my head.

eregon: That said in this case having `isRubyBigDecimal(b)` first is fine for me if other specializations start with that guard first too.

truffle/src/main/java/org/jruby/truffle/nodes/ext/BigDecimalNodes.java
eregon: What about having it in the `default` case? I think it would be easier to read + less IDE warnings.

eregon:
I think we should adopt some checkstyle so the code uniformity would be pretty clear. It does not have to be in the CI if we just all have plugins in our IDEs checking it.

truffle/src/main/java/org/jruby/truffle/nodes/ext/BigDecimalNodes.java
eregon: I would move this in the switch `default:` case.
In general I'm in favor of indentation-based control flow (and avoid `return` in the middle) :smile:

eregon:
The diff is too hard to read for me so I'll just check the branch locally. Looks good but there seems to be a lot of complexity and nested nodes which I did not expect to be needed.

eregon:
@pitr-ch I do not see why you need a BigDecimalCoreMethodNode and in particular why static methods cannot be used to create a a Ruby BigDecimal from a Java BigDecimal. Could you enlighten me?

truffle/src/main/java/org/jruby/truffle/nodes/ext/BigDecimalNodes.java
pitr-ch: `is(Normal|Special)RubyBigDecimal` introduced.

truffle/src/main/java/org/jruby/truffle/nodes/ext/BigDecimalNodes.java
pitr-ch: It think: since the code should not be ever reached it should also never be compiled, so it made sense to me and I've used it as I found it elsewhere in our code-base. I do not know if there is another reason. Yep, default branch sounds good.

pitr-ch:
@eregon about the create node vs static methods. I wanted a single place for BigDecimal creation which
- would handle all the possible types which may be used to create a BigDecimal
- would respect current exception mode, rounding, limit.

Exception mode, rounding, limit are read using nodes so it was then easier to create new node for the creation than bunch of static methods.

eregon:
But most operations will just create a Java BigDecimal or a special type, and they have their precision and rounding from their operands I think, right?
Looks a bit like a god node to me :smiley:

pitr-ch:
@eregon Hm you are right. It checks mostly only the exception_mode, It'll use rounding too when I add missing features to initialise. I could read the exception_mode and rounding in the node requiring BigDecimal creation passing the mode and rounding into the static methods, but I wanted to encapsulate it in one unit. It also allowed me to reuse the CastNode. Do you think strongly about this? I could change it but my feeling is that both have dis/advantages so I am not sure if it makes sense to change it. Are there advantages in having static methods I don't see? My assumption is that having small nodes is more flexible.

eregon:
@pitr-ch I did not see we need to call `BigDecimal.mode` or access the thread-local state to know the rounding mode. Then definitely merge as is.
It might be worth at some point to split into MegaCreateBigdecimalNode and CreateBigDecimalFromJavaNode.

chrisseaton:
Yeah let's get this merged. 1000 lines is a lot to have sitting on a branch for more than a few days.

pitr-ch:
yeah, I'll merge it, but first I have to figure out an performance issue in the current implementation. I think it's rounding operation which will hit in some tests (tagged) an loop which takes tens of minutes to finish. As a result tagging takes a very long time.

pitr-ch:
UnsupportedOperationError followup: I removed the never-part-of-compilation-assert (in BigDecimalNodes and in two other cases) and kept the current exception. The helper we've talked about is omitted. I've looked at the rest (about 100) UnsupportedOperation's usages. It would take quite long time to figure out where it is representing UnsupportedOperation and where it is representing an unreachable code-branch. So for now this code follows the same ambiguity we have in the rest of the code. If you think it would make sense to disambiguate I'll do that after this is merged.

PR #3084, opened 2015-06-26
[Truffle] BigDecimal

pitr-ch:


lib/ruby/truffle/truffle/bigdecimal.rb
```diff
@@ -181,6 +189,70 @@ def frac
...
+                       if [' ', '+'].include? format[0]
+                         format[0]
+                       else
+                         ''
+                       end
+                     when self < 0
+                       '-'
+                     when self == 0
+                       ''
+                     end
```
eregon: The indentation looks a bit strange to me.

lib/ruby/truffle/truffle/bigdecimal.rb
```diff
@@ -181,6 +189,70 @@ def frac
...
+                                  ['0',
+                                   '0' * exponent_value.abs + unscaled_value]
+                                else
+                                  after_dot = unscaled_value[exponent_value..-1]
+
+                                  [unscaled_value[0...exponent_value],
+                                   after_dot.empty? ? '0' : after_dot]
+                                end
+
+        prefix << add_spaces_to_s(before_dot, true, spaces) << '.' << add_spaces_to_s(after_dot, false, spaces)
```
eregon: I would use String interpolation for this, I find it easier to read and overhead it probably fairly low.

lib/ruby/truffle/truffle/bigdecimal.rb
```diff
@@ -181,6 +189,70 @@ def frac
...
+                                   '0']
+                                when exponent_value <= 0
+                                  ['0',
+                                   '0' * exponent_value.abs + unscaled_value]
+                                else
+                                  after_dot = unscaled_value[exponent_value..-1]
+
+                                  [unscaled_value[0...exponent_value],
+                                   after_dot.empty? ? '0' : after_dot]
+                                end
```
eregon: splat over a `case`statement with no expression seems too smart :smile:
What about just if/elsif with assigning the two variable in each clause?
```ruby
if exponent_value > unscaled_value.size
  before_dot = unscaled_value + '0' * (exponent_value - unscaled_value.size)
  after_dot = '0'
elsif exponent_value <= 0
  ...
```
(the variables need not be declared before as if/elsif assign in the parent scope)

lib/ruby/truffle/truffle/bigdecimal.rb
pitr-ch: Do you mean case/when or aligned = ?

lib/ruby/truffle/truffle/bigdecimal.rb
pitr-ch: I tend to prefer case over of/elsif in when there is more than two branches to distinguish it clearly from simple if. 

Multi-assign is used here to be DRY and "functional". In the current arrangement it is clear that whole block is about assigning those variables, but if the variables are in the case (or if/elsif) a reader has to read whole block to determine that it's about assigning of those variables and nothing else.

Anyway this is just stylistic issue, I don't mind changing it if you think this is too weird.

lib/ruby/truffle/truffle/bigdecimal.rb
pitr-ch: yeah I am not happy with this line, I've tried interpolation but `format` seems event better to me.

lib/ruby/truffle/truffle/bigdecimal.rb
eregon: the `=` is the clearest here.
Not sure for `case`/`when` but it might make sense to avoid anything that has visually so many levels of indentation (so `when` would have the same indent as the line before).

lib/ruby/truffle/truffle/bigdecimal.rb
eregon: But on the other hand you have to associate `array index 0` with `before_dot`.
There is also an assignment to `after_dot` which makes it even more complex.
Yeah I am a big fan of simplicity over DRY :wink: until it goes in its own helper method.

lib/ruby/truffle/truffle/bigdecimal.rb
eregon: `Kernel#format` ? I think the current version is better in that case (`format` is not idiomatic IMO).
Interpolation is a bit hard to read because there is no space between tokens, but it shows more clearly the order.

pitr-ch: see https://github.com/jruby/jruby/pull/3084/files#diff-8b366f91f8d86a2cbb3a9175eda4ee3bR221 I think it looks quite good. And it should be also quite performant thanks to our pack nodes, right?

truffle/src/main/java/org/jruby/truffle/nodes/ext/BigDecimalNodes.java
```diff
@@ -1108,15 +1110,58 @@ public Object divSpecialSpecial(VirtualFrame frame, RubyBasicObject a, RubyBasic
...
+            }
+            return super.div(frame, a, b, newPrecision);
+        }
+
+        @Specialization(guards = {
+                "isNormal(a)",
+                "isSpecialRubyBigDecimal(b)" })
+        public Object divNormalSpecial(VirtualFrame frame, RubyBasicObject a, RubyBasicObject b, NotProvided precision) {
+            if (getBigDecimalType(b) == Type.NEGATIVE_ZERO) {
+                throw new RaiseException(getContext().getCoreLibrary().zeroDivisionError(this));
```
nirvdrum: You should transfer before you throw here.

truffle/src/main/java/org/jruby/truffle/nodes/ext/BigDecimalNodes.java
```diff
@@ -1108,15 +1110,58 @@ public Object divSpecialSpecial(VirtualFrame frame, RubyBasicObject a, RubyBasic
...
+            } else if (getBigDecimalType(b) == Type.NAN) {
+                throw new RaiseException(getContext().getCoreLibrary().floatDomainError("Computation results to 'NaN'(Not a Number)", this));
```
nirvdrum: You should transfer before you throw here.

truffle/src/main/java/org/jruby/truffle/nodes/ext/BigDecimalNodes.java
```diff
@@ -1129,6 +1174,21 @@ public Object divNormalSpecial(VirtualFrame frame, RubyBasicObject a, RubyBasicO
         @Specialization(guards = {
                 "!isNormal(a)",
                 "isNormalRubyBigDecimal(b)" })
+        public Object divSpecialNormal(VirtualFrame frame, RubyBasicObject a, RubyBasicObject b, NotProvided precision) {
+            if (isNormalZero(b)) {
+                throw new RaiseException(getContext().getCoreLibrary().zeroDivisionError(this));
```
nirvdrum: Missing transfer.

truffle/src/main/java/org/jruby/truffle/nodes/ext/BigDecimalNodes.java
```diff
@@ -1129,6 +1174,21 @@ public Object divNormalSpecial(VirtualFrame frame, RubyBasicObject a, RubyBasicO
...
+            } else if (getBigDecimalType(a) == Type.NAN) {
+                throw new RaiseException(getContext().getCoreLibrary().floatDomainError("Computation results to 'NaN'(Not a Number)", this));
```
nirvdrum: Missing transfer.

truffle/src/main/java/org/jruby/truffle/nodes/ext/BigDecimalNodes.java
```diff
@@ -1129,6 +1174,21 @@ public Object divNormalSpecial(VirtualFrame frame, RubyBasicObject a, RubyBasicO
...
+            } else if (getBigDecimalType(a) == Type.POSITIVE_INFINITY || getBigDecimalType(a) == Type.NEGATIVE_INFINITY) {
+                throw new RaiseException(getContext().getCoreLibrary().floatDomainError("Computation results to 'Infinity'", this));
```
nirvdrum: Missing transfer.

truffle/src/main/java/org/jruby/truffle/nodes/ext/BigDecimalNodes.java
```diff
@@ -1136,11 +1196,176 @@ public Object divSpecialNormal(VirtualFrame frame, RubyBasicObject a, RubyBasicO
         @Specialization(guards = {
                 "!isNormal(a)",
                 "isSpecialRubyBigDecimal(b)" })
+        public Object divSpecialSpecial(VirtualFrame frame, RubyBasicObject a, RubyBasicObject b, NotProvided precision) {
+            if (getBigDecimalType(b) == Type.NEGATIVE_ZERO) {
+                throw new RaiseException(getContext().getCoreLibrary().zeroDivisionError(this));
```
nirvdrum: Missing transfer.

truffle/src/main/java/org/jruby/truffle/nodes/ext/BigDecimalNodes.java
```diff
@@ -1136,11 +1196,176 @@ public Object divSpecialNormal(VirtualFrame frame, RubyBasicObject a, RubyBasicO
...
+            } else if (getBigDecimalType(a) == Type.NAN || getBigDecimalType(b) == Type.NAN) {
+                throw new RaiseException(getContext().getCoreLibrary().floatDomainError("Computation results to 'NaN'(Not a Number)", this));
```
nirvdrum: Missing transfer.

truffle/src/main/java/org/jruby/truffle/nodes/ext/BigDecimalNodes.java
```diff
@@ -1136,11 +1196,176 @@ public Object divSpecialNormal(VirtualFrame frame, RubyBasicObject a, RubyBasicO
...
+        public Object divmodSpecial(VirtualFrame frame, RubyBasicObject a, RubyBasicObject b) {
+            final Type aType = getBigDecimalType(a);
+            final Type bType = getBigDecimalType(b);
+
+            if (aType == Type.NAN || bType == Type.NAN) {
+                return createArray(createBigDecimal(frame, Type.NAN), createBigDecimal(frame, Type.NAN));
+            }
+
+            if (bType == Type.NEGATIVE_ZERO || (bType == Type.NORMAL && isNormalZero(b))) {
+                throw new RaiseException(getContext().getCoreLibrary().zeroDivisionError(this));
```
nirvdrum: Missing transfer.

truffle/src/main/java/org/jruby/truffle/nodes/ext/BigDecimalNodes.java
```diff
@@ -1136,11 +1196,176 @@ public Object divSpecialNormal(VirtualFrame frame, RubyBasicObject a, RubyBasicO
...
+                return createArray(
+                        createBigDecimal(frame, type),
+                        createBigDecimal(frame, Type.NAN));
+            }
+
+            if (bType == Type.POSITIVE_INFINITY || bType == Type.NEGATIVE_INFINITY) {
+                return createArray(createBigDecimal(frame, BigDecimal.ZERO), createBigDecimal(frame, a));
+            }
+
+            throw new UnsupportedOperationException();
```
nirvdrum: While we have a lot of `UnsupportedOperationException`s without a message, I think we should really try to use a descriptive error message in all future work.

eregon: I agree, although just having the Java backtrace is usually quite enough to know where it comes from.

lib/ruby/truffle/truffle/bigdecimal.rb
eregon: Yeah it is nice! I think `String#%` is more Ruby-like but in this case the extra array literal would just add clutter.
`Array#pack` is for encoding stuff in binary, this is in text format so I don't think we have the same level of optimization yet.

pitr-ch:
Sorry about the missing transfers ( I've forgot again :angry: ). Added.

truffle/src/main/java/org/jruby/truffle/nodes/ext/BigDecimalNodes.java
pitr-ch: what about https://github.com/pitr-ch/jruby/commit/f1636466b3271cc50d85574bd7b2685d9d556d2f?

nirvdrum: The problem is it's usually a design decision to make something unsupported or there's an underlying reason why that is.  Without a comment or an error message, we're forever having to work out why it's unsupported when we come across it.  In cases like getting the metaclass or translating a JRuby exception, messages with the argument values can save a lot of time debugging.

pitr-ch: @nirvdrum I understand and agree, however these cases are all just a branch of code which should not be ever reached. So there is new sub class for that added to describe the meaning. What do you think?

pitr-ch:
Please wait with merge, 1 or 2 commits and BigDecimal is done.

truffle/src/main/java/org/jruby/truffle/nodes/ext/BigDecimalNodes.java
nirvdrum: @pitr-ch I commented on the other commit.  I like your proposed solution. It clearly conveys intent.

pitr-ch:
Ready for merge.

eregon:
Looks good!

Issue #1769, opened 2014-06-27
Inconsistency with MRI when adding a month to a date near Date::ITALY


robin850:
Hello,

There is a disparity between JRuby and MRI dealing with dates close to the calendar reform in Italy (i.e. 4th October of 1582). Using JRuby 1.7.13 and MRI 2.1.2:

``` ruby
>> require 'date'
=> true
>> date = Date.new(1582, 9, 9);

>> RUBY_ENGINE
=> "jruby"
>> (date >> 1).day
=> 19

>> RUBY_ENGINE
=> "ruby"
>> (date >> 1).day
=> 4
```

Since in Italy, October 5 to 14 don't exist in 1582, MRI is returning the day of the reform. However, it looks like JRuby is correctly adding a month to the original date but it's also adding 10 days.

The problem lies on the Joda side since [`#>>`](https://github.com/jruby/jruby/blob/8df1f563710f4c1c3afe9fb7a684e40f1a01e8b7/lib/ruby/2.1/date.rb#L1457) is relying on a `Joda::DateTime` object.

For the record, this disparity make the [following test](https://github.com/rails/rails/blob/9c492885d178b1a3d0de2f710a8a276db3d867b2/activesupport/test/core_ext/date_ext_test.rb#L147-L156) to fail in the Rails' test suite. Here's a failing test for your suite as RubySpec doesn't deal with the standard library anymore:

``` diff
diff --git a/test/mri/date/test_date_arith.rb b/test/mri/date/test_date_arith.rb
index 745ddbd..145d263 100644
--- a/test/mri/date/test_date_arith.rb
+++ b/test/mri/date/test_date_arith.rb
@@ -190,6 +190,15 @@ class TestDateArith < Test::Unit::TestCase
     assert_equal([2001, 2, 28], [d.year, d.mon, d.mday])
   end

+  def test_nex_month__with_calendar_reform
+    calendar_reform_italy = [1582, 10, 4]
+
+    d = Date.new(1582, 9, 9) >> 1
+    assert_equal(calendar_reform_italy, [d.year, d.mon, d.mday])
+    d = Date.new(1582, 9, 10) >> 1
+    assert_equal(calendar_reform_italy, [d.year, d.mon, d.mday])
+  end
+
   def test_next_year
     d = Date.new(2000,1,31).next_year(-1)
     assert_equal([1999, 1, 31], [d.year, d.mon, d.mday])
```

Have a nice day!

eregon:
[...]

Issue #2867, opened 2015-04-22
Date#next_year difference behvior with MRI


gazay:
> bug from AS rails tests

Seems to related to https://github.com/jruby/jruby/issues/1769

```ruby
# jruby master
irb(main):002:0> require 'date'
=> true
irb(main):003:0> Date.new(1581,10,10).next_year
=> #<Date: 1582-10-20 ((2299166j,0s,0n),+0s,2299161j)>

# ruby 2.2.2
irb(main):002:0> require 'date'
=> true
irb(main):003:0> Date.new(1581,10,10).next_year
=> #<Date: 1582-10-04 ((2299160j,0s,0n),+0s,2299161j)>
```

headius:
This could be simple so I'm marking for beginners.

raeoks:
It seems like a good bug to get started.
@eregon had already posted a solution at #1769, but he also mentioned that it's likely to be fairly slow.
@headius Let me know if you something like that in mind.

headius:
I'm not concerned about the performance if we can get a fix in for this for JRuby 9k RC2. @eregon or @raeoks, can you make a PR so we can wrap this up?

eregon:
I'll let @raeoks take care of it since he already worked on adding tests/specs.

eregon:
Fixed by #3157 and #3091.

eregon:
Thanks @raeoks for your contribution and the specs!

raeoks:
@eregon It's been a pleasure to work on JRuby. Next I'll move on to another `beginner` bug.

PR #3091, opened 2015-06-29
[#2867] modified Date#>> to take calendar reforms under consideration

raeoks:
- added tests for next month with calendar reform
- uses solution provided in #1769 comments
- also fixes issues #1769


kares:
@raeoks would you mind trying to target **jruby-1_7** so that this is fixed in JRuby 1.7.x as well?

@eregon since its mostly your work-around ... any thoughts on getting those in?

eregon:
These tests would be useful in RubySpec. You can just do a separate commit touching `spec/ruby`.

Well, if this behavior must be supported, I'd say there is not much choice.

raeoks:
I'll move test cases from `test/mri/date/test_date_arith.rb` to `spec/ruby/library/date`'s `add_month_spec.rb` and `next_year_spec.rb` files.
Also I'll be start working on creating a new pull request to fix same issue on jruby-1_7.

headius:
The changes look ok to me but I defer to @eregon on when and whether this should merge.

eregon:
@raeoks Could you move the test cases to specs? Or should I do it?

raeoks:
@eregon I'll do and doing it right now.

raeoks:
@eregon Should I create a new pull request resolving #2867 for Jruby 1.7?

eregon:
I guess yes but better ask @kares.
@raeoks Is this PR ready to be merged?

kares:
optimally, there would be only one PR targeting **jruby-1_7** (branch is periodically merged to master) ...

eregon:
I applied the commits to 1.7, 1.9-mode (1.8 Date is just so different and should not be affected the same way by this bug): 40e9140209d944f424c965ccdf8b2304b8c2aade and ec80a3209e54cb259c45851375d0a2c2fb7334d0.

I guess the merge in this case would not work so well (date.rb moved quite a bit), so I'd rather merge #3157 in master instead.

PR #3093, opened 2015-06-30
isOpen()/close() is racy. Simply close and ignore EBADF.

eam:


eam:
This fixes a race which can be hit within a few thousand iterations of:

```ruby
loop do |i|
  IO.popen4('echo', 'hi') { |_, _, out, _| out.read }
end
```

I believe the descriptor is occasionally closed between the test of isOpen() and the close(), triggering the exception.

headius:
I might feel better if we did *both*, actually, to avoid unnecessarily raising an error that gets ignored. The failed race would then be the only case that raises.

headius:
Thanks!

PR #3099, opened 2015-07-02
Uri like paths with pathname

mkristian:


jkutner:
Is this related to #3127 ?

mkristian:
first this is already merged to jruby-1_7 and master and part of the 1.7.21 and 9.0.0.0.rc2

it is related. the last weeks I was working on getting jruby launch a ruby application from an unexploded jar-file. basically how we pack the jar with https://github.com/jruby-gradle/jruby-gradle-plugin or with https://github.com/torquebox/jruby-maven-plugins/tree/master/jruby9

so all the problem fixed are related to having (with variations)
* GEM_HOME=uri:classloader:/
* GEM_PATH=uri:classloader:/
*  JARS_HOME=uri:classloader:/jars
* current working directory uri:classloader:/
* and having uri:classloader:/ on the LOAD_PATH

this works to the extend that you can pack up a rails application and run it from the jar (without unpacking it). the Pathname problem came up with such rails application.

PR #3107, opened 2015-07-06
[Truffle] Don't create threads in Main for Truffle

bbelleville:
@chrisseaton : This removes code that uses features that aren't supported for ahead of time compilation.


chrisseaton:
Merged, but I did something funny in git so GitHub doesn't see it.

eregon:
It might be worth leaving a TODO or since we might want to do that tearDown (possibly in a finally block instead?)

PR #3108, opened 2015-07-06
[Truffle] Function to initialize secureRandom

bbelleville:
@chrisseaton : This is to initialize ThreadContext.secureRandom. This change should not
affect the behavior of this class, but facilitates ahead of time
compilation. If this code is in a function it is easier to create a substitute when doing ahead of time compilation.


chrisseaton:
Merged, but I did something funny in git so GitHub doesn't see it.

PR #3120, opened 2015-07-08
Add some more tests for bignum pow

skmp:
I added tests for checking two more currently failing scenarios triggering ArithmeticExceptions instead of returning Infinity. 
I am not sure if this is the right place for adding these as these tests are taken over from MRI if I understand correctly. Can you provide some guidance here?


eregon:
This would be nice to have in rubyspec, in https://github.com/jruby/jruby/blob/master/spec/ruby/core/bignum/exponent_spec.rb for instance.

kares:
on master as https://github.com/jruby/jruby/commit/9a7dc0daf145e984a03c15173fbdfa6ae7429656 ... thanks @skmp

PR #3127, opened 2015-07-10
Added uri to path when using file protocol to chdir

jkutner:
See jruby/warbler#323

### DIFF
--- a/core/src/main/java/org/jruby/RubyDir.java
+++ b/core/src/main/java/org/jruby/RubyDir.java
@@ -86,6 +86,8 @@ public class RubyDir extends RubyObject {
 
     private final static Encoding UTF8 = UTF8Encoding.INSTANCE;
 
+    private static Pattern PROTOCOL_PATTERN = Pattern.compile("^(uri|jar|file|classpath):([^:]*:)?//?.*");
+
     public RubyDir(Ruby runtime, RubyClass type) {
         super(runtime, type);
     }
@@ -323,11 +325,8 @@ public class RubyDir extends RubyObject {
         checkDirIsTwoSlashesOnWindows(runtime, adjustedPath);
         String realPath = null;
         String oldCwd = runtime.getCurrentDirectory();
-        if (adjustedPath.startsWith("uri:")){
+        if (PROTOCOL_PATTERN.matcher(adjustedPath).matches()) {
             realPath = adjustedPath;
-        }
-        else if (adjustedPath.startsWith("file:")){
-            realPath = "uri:" + adjustedPath;
         }
         else {
             FileResource dir = getDir(runtime, adjustedPath, true);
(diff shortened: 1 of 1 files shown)


core/src/main/java/org/jruby/RubyDir.java
```diff
@@ -326,6 +326,9 @@ public static IRubyObject chdir(ThreadContext context, IRubyObject recv, IRubyOb
         if (adjustedPath.startsWith("uri:")){
             realPath = adjustedPath;
         }
+        else if (adjustedPath.startsWith("file:")){
+            realPath = "uri:" + adjustedPath;
+        }
```
mkristian: a path with ```!/``` is treated as jar uri whether this is a file: or jar: or jar:file: prefix or not. and this is the case where you want the uri: to add. but for ```uri:``` to work properly it needs ```new URL(adjustedPath).openStream()``` to work. i.e. in your case make sure there is jar:file: prefix

an path with file: onto the filesystem without the ```!/``` better just remove the file: prefix and let jruby use the filesystem.

but I do not understand why your current patch fixes the warbler problem

mkristian: I think we all to chdir to all paths following this pattern: https://github.com/jruby/jruby/blob/jruby-1_7/core/src/main/java/org/jruby/RubyFile.java#L670

it feels wrong to switch one pseudo-uri to another

jkutner:
No, neither 1.7.21 nor 1.7.22.dev fixed this. I tried removing the `file:` prefix, and got this error:

```
java.net.MalformedURLException: no protocol: /Users/jkutner/workspace/warbler-debug/warbler-debug.war!/WEB-INF/gems/bundler/gems/ruby-units-a777aeebcfbd/lib/ruby_units/version.rb
	at java.net.URL.<init>(URL.java:586)
	at java.net.URL.<init>(URL.java:483)
	at java.net.URL.<init>(URL.java:432)
	at org.jruby.util.URLResource.createRegularURI(URLResource.java:199)
	at org.jruby.util.URLResource.create(URLResource.java:187)
	at org.jruby.util.JRubyFile.createResource(JRubyFile.java:91)
	at org.jruby.util.JRubyFile.createResource(JRubyFile.java:102)
	at org.jruby.util.JRubyFile.createResource(JRubyFile.java:75)
	at org.jruby.RubyFileStat.setup(RubyFileStat.java:122)
	at org.jruby.RubyFileStat.newFileStat(RubyFileStat.java:97)
	at org.jruby.Ruby.newFileStat(Ruby.java:3342)
	at org.jruby.RubyFile.stat(RubyFile.java:894)
```

I will investigate the .jrubydir

mkristian:
the .jrubydir will not help here with this concrete bundler problem. just passing on the ```realPath = adjustedPath;``` should help and keeps things more consistent.

yes, I did not run the sample repo at the beginning.

jkutner:
Ah, i understand now. I will improve this

mkristian:
never saw the update - but saw your question on IRC :)

jkutner:
@mkristian thanks for your help with this!

PR #3140, opened 2015-07-15
[Truffle] returnIDs use Objects, Thread#raise, minor updates

pitr-ch:


truffle/src/main/ruby/core/rubinius/bootstrap/thread.rb
```diff
@@ -41,4 +41,36 @@ def self.kill(thread)
...
+    end
+
+    if self == Thread.current
+      Kernel.raise exc
+    else
+      exc.capture_backtrace! 2 unless exc.backtrace?
+      Truffle::Primitive.thread_raise self, exc
+    end
+  end
+
```
eregon: Where does this code come from? If it is modified it should be in api/shims.
If we can work with the original code, it is usually better, so in this case we could implement the `thread_raise` Rubinius primitive, right? (In ThreadPrimitiveNodes) See `VMPrimitiveNodes` for an example of Rubinius primitives nodes.

truffle/src/main/java/org/jruby/truffle/translator/ParseEnvironment.java
```diff
@@ -40,14 +39,8 @@ public void popLexicalScope() {
...
 
-    public long allocateReturnID() {
-        if (nextReturnID == Long.MAX_VALUE) {
-            throw new RuntimeException("Return IDs exhausted");
-        }
-
-        final long allocated = nextReturnID;
-        nextReturnID++;
-        return allocated;
+    public Object allocateReturnID() {
```
eregon: @chrisseaton might want to have a subtype for this, i.e.: `ReturnID`.

eregon:
Looks great, please address my 3 comments.

truffle/src/main/java/org/jruby/truffle/nodes/core/TrufflePrimitiveNodes.java
```diff
@@ -520,4 +521,28 @@ protected static boolean canLower(long value) {
 
     }
 
+    @CoreMethod(names = "thread_raise", required = 2, onSingleton = true)
+    public abstract static class RaiseNode extends CoreMethodArrayArgumentsNode {
+
+
```
chrisseaton: Please be careful about things like random extra newlines.

truffle/src/main/java/org/jruby/truffle/translator/ParseEnvironment.java
chrisseaton: Yeah I originally said Object, but you're right, we should have a class for it. So that for example it shows up well in allocation profiling.

pitr-ch: yeah good point.

truffle/src/main/ruby/core/rubinius/bootstrap/thread.rb
pitr-ch: The original Rubinius `Thread#raise` uses `Rubinius.lock` and other intrinsic `@alive` variable, checks if the Thread is alive etc. That's the reason why this method is modified, so moving to shims. Does it makes sense to use Rubinius primitive if the ruby wrapping code is different?

truffle/src/main/java/org/jruby/truffle/nodes/core/TrufflePrimitiveNodes.java
pitr-ch: ok, IDE formatter configuration updated.

truffle/src/main/ruby/core/rubinius/bootstrap/thread.rb
eregon: Yeah, I think having the Rubinius primitive still makes sense to preserve most of the original code (and also better organization, not visible so directly to user, etc).

pitr-ch: Done

pitr-ch: done

truffle/src/main/ruby/core/rubinius/api/shims/thread.rb
```diff
@@ -15,4 +42,43 @@ def recursive_objects
...
+    end
+
+    if self == Thread.current
+      Kernel.raise exc
+    else
+      exc.capture_backtrace! 2 unless exc.backtrace?
+      raise_prim exc
+    end
+  end
+
```
eregon: All of this should be in a separate file, because the loading order will load Rubinius' common/thread.rb after this according to [core.rb](https://github.com/jruby/jruby/blob/master/truffle/src/main/ruby/core.rb#L29).
The solution is to move it to a separate file, for instance named `thread_common.rb` or `thread_after_common.rb` and add that one after `common/thread,rb` in core.rb.
We should have a better organization for all of this.
cc @nirvdrum.

pitr-ch: Small correction `Thread#raise` is defined in `bootstrap/thread` we do not have it in our repo though, which was to reason why I've putted into existing `shims/thread`. You are right that it's error prone (if we later copy the `bootsrap`) and it would be better to have it loaded in extra file. Moving.

eregon: Ah so that's why the CI passes :) Indeed, the more clarity the better!

nirvdrum:
The comments were deleted when new changes were pushed, so I think I only have part of the conversation from email.  But, if you want `Thread.raise` from the Rubinius bootstrap file (and don't need to change it), just import it from the correct version of Rubinius.  The bootstrap files have been treated as additive, while the core files are 100% what upstream has and we exclude some stuff we can't use.

pitr-ch:
@nirvdrum I've used just part of the original method, that's why hence the discussion where to put it. The original method used internal implementation details. I did not import the original bootstrap method, should I?

nirvdrum:
@pitr-ch That's fine.  If you find yourself needing to only exclude part of a method `Truffle.omit` is an option as well (grep for current usages).

truffle/src/main/ruby/core/rubinius/api/shims/thread_common.rb
```diff
@@ -0,0 +1,77 @@
...
+# IMPLIED WARRANTIES OF MERCHANTABILITY AND FITNESS FOR A PARTICULAR PURPOSE ARE
+# DISCLAIMED. IN NO EVENT SHALL THE COPYRIGHT OWNER OR CONTRIBUTORS BE LIABLE
+# FOR ANY DIRECT, INDIRECT, INCIDENTAL, SPECIAL, EXEMPLARY, OR CONSEQUENTIAL
+# DAMAGES (INCLUDING, BUT NOT LIMITED TO, PROCUREMENT OF SUBSTITUTE GOODS OR
+# SERVICES; LOSS OF USE, DATA, OR PROFITS; OR BUSINESS INTERRUPTION) HOWEVER
+# CAUSED AND ON ANY THEORY OF LIABILITY, WHETHER IN CONTRACT, STRICT LIABILITY,
+# OR TORT (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY OUT OF THE USE
+# OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
+#
+# Modifications are subject to:
```
nirvdrum: Unless you've added any code (i.e., didn't just delete stuff from Rubinius), you really don't need this.  And even then, it's questionable for trivial changes.  There's no harm in having it here, but it's not usually a requirement.

nirvdrum:
FYI, I think what you did with the new shims file is fine.  As @eregon said, now that we have a pattern of a few use cases, we should tidy up.  Most of the files in api/shims really aren't shims for an API.

truffle/src/main/ruby/core/rubinius/api/shims/thread_common.rb
eregon: But a large parts of both methods below are part of Rubinius, so the license should be in this file, no?

eregon:
@nirvdrum Yeah, I think actually most of "API shims" are really just defined in bootstrap in proper Rubinius and it would make more sense to define it with our files at that same bootstrap stage.

truffle/src/main/ruby/core/rubinius/api/shims/thread_common.rb
nirvdrum: Yes.  The Rubinius license should be in tact.  I specifically meant the modification parts.

truffle/src/main/ruby/core/rubinius/api/shims/thread_common.rb
```diff
@@ -0,0 +1,77 @@
...
+# Copyright (c) 2014, 2015 Oracle and/or its affiliates. All rights reserved. This
```
nirvdrum: Unless you ported this from elsewhere, the code should not be copyrighted in 2014.

truffle/src/main/ruby/core/rubinius/api/shims/thread_common.rb
pitr-ch: There are few minor additions, so I'll keep it in this case.

Another possible approach to this could be to always copy the original method with the rubinius license and have the overrides in shims with truffle license. Then it would not be mixed and we would be able to say easily what is the difference. What do you think?

pitr-ch:
yeah, right now the `core.rb` serves as a source to find out where the file belongs. We could have `shims/` followed by the same structure of directories as rubinus has for its files.

Regarding this, PR can I merge?

truffle/src/main/ruby/core/rubinius/api/shims/thread_common.rb
pitr-ch: Good point it should be just 2015

chrisseaton:
> Regarding this, PR can I merge?

Not today! We're waiting for the 9k release.

chrisseaton:
@pitr-ch sorry couldn't merge this. Can you pull, fix and merge?

pitr-ch:
@chrisseaton np, after ci passes i'll merge

PR #3148, opened 2015-07-17
[Truffle] No more global lock!

eregon:
@chrisseaton @nirvdrum @pitr-ch 
This is the very simple commit to remove the GIL (cleanup will be done later when merged).
I tried a few trivial parallelism problems, a parallel `richards` benchmark and a few benchmarks from [NPB](http://www.nas.nasa.gov/publications/npb.html) ([Ruby version](http://www-hiraki.is.s.u-tokyo.ac.jp/members/tknose/)).

Multi-threaded programs need to be properly synchronized or they might crash in various ways (as in standard JRuby).

### DIFF
--- a/truffle/src/main/java/org/jruby/truffle/runtime/subsystems/ThreadManager.java
+++ b/truffle/src/main/java/org/jruby/truffle/runtime/subsystems/ThreadManager.java
@@ -24,7 +24,6 @@ import org.jruby.truffle.runtime.core.RubyBasicObject;
 import java.util.Collections;
 import java.util.Set;
 import java.util.concurrent.ConcurrentHashMap;
-import java.util.concurrent.locks.ReentrantLock;
 
 /**
  * Manages Ruby {@code Thread} objects.
@@ -33,10 +32,8 @@ public class ThreadManager {
 
     private final RubyContext context;
 
-    private final ReentrantLock globalLock = new ReentrantLock();
-
     private final RubyBasicObject rootThread;
-    private RubyBasicObject currentThread;
+    private final ThreadLocal<RubyBasicObject> currentThread = new ThreadLocal<RubyBasicObject>();
 
     private final Set<RubyBasicObject> runningRubyThreads = Collections.newSetFromMap(new ConcurrentHashMap<RubyBasicObject, Boolean>());
 
@@ -65,8 +62,7 @@ public class ThreadManager {
     @TruffleBoundary
     public void enterGlobalLock(RubyBasicObject thread) {
         assert RubyGuards.isRubyThread(thread);
-        globalLock.lock();
-        currentThread = thread;
+        currentThread.set(thread);
     }
 
     /**
@@ -77,13 +73,7 @@ public class ThreadManager {
      */
     @TruffleBoundary
     public RubyBasicObject leaveGlobalLock() {
-        if (!globalLock.isHeldByCurrentThread()) {
-            throw new RuntimeException("You don't own this lock!");
-        }
-
-        final RubyBasicObject result = currentThread;
-        globalLock.unlock();
-        return result;
+        return currentThread.get();
     }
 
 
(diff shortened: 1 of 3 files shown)


chrisseaton:
ha when I said open a PR I thought it would be massive - but I guess you've done all the work on master and then this is just the last step.

@nirvdrum can you test your tests using this branch? I'll try all the normal benchmarks.

@headius you may be interested in this

chrisseaton:
I wonder if since we already have a GIL here, it would make sense to keep it as a non-default option? Are there unsynchronised multi-threaded applications that work on MRI but not on JRuby?

eregon:
I think that's not an option as it would be unmanageable complexity to have both.

truffle/src/main/java/org/jruby/truffle/runtime/subsystems/ThreadManager.java
```diff
@@ -61,8 +58,7 @@ public RubyThread getRootThread() {
      */
     @TruffleBoundary
     public void enterGlobalLock(RubyThread thread) {
-        globalLock.lock();
-        currentThread = thread;
+        currentThread.set(thread);
```
nirvdrum: Should this method be renamed if there's no longer a global lock?

truffle/src/main/java/org/jruby/truffle/runtime/subsystems/ThreadManager.java
```diff
@@ -73,13 +69,7 @@ public void enterGlobalLock(RubyThread thread) {
...
     @TruffleBoundary
     public RubyThread leaveGlobalLock() {
-        if (!globalLock.isHeldByCurrentThread()) {
-            throw new RuntimeException("You don't own this lock!");
-        }
-
-        final RubyThread result = currentThread;
-        globalLock.unlock();
-        return result;
+        return currentThread.get();
```
nirvdrum: Same question as with `enterGlobalLock`: should this method be renamed if there's no longer a global lock?

truffle/src/main/java/org/jruby/truffle/runtime/subsystems/ThreadManager.java
eregon: Yes that method will go away in the cleanup phase once this is merged.
It has just been a lot more practical to keep a minimal diff until then.

pitr-ch:
This like is worrying me https://github.com/pitr-ch/jruby/blob/master/truffle/src/main/ruby/core/shims.rb#L165-165? Not sure how many problems we'll face in reality.

chrisseaton:
Yeah I don't think anyone is expecting the parallelism to just work perfectly straight away and there is clearly going to be a lot of work to make it correct. We need parallelism stress tests.

Obviously we should fix things as we find them (especially this one as it might impact on performance).

eregon:
@pitr-ch That one is very wrong, I was not aware of it. Shims should be to implement only a subset of the semantics, but not totally incorrect semantics like that one.
I agree with @chrisseaton, we cannot expect to be perfectly correct from day 1. I plan to try using JCStress tests for stress tests.

eregon:
@pitr-ch I fixed that one.

pitr-ch:
Nice :+1:

chrisseaton:
Sorry the merge is broken - can you fix and merge?

eregon:
Merged, here comes the way to parallel JRuby+Truffle!

A quick experiment on some Monte Carlo pi approximation shows we scale as good as C!
(even better with hyper-threading but that is strange)
monte_carlo_pi_c_ruby_16

chrisseaton:
> even better with hyper-threading but that is strange

Probably means you are waiting on fetching from RAM most of the time. Not sure how Monte Carlo pi works but if you can try a smaller working set it might fit into cache and so not scale was well with hyper threads.

PR #3157, opened 2015-07-21
Date next year difference

eregon:
From #2867, with commits rebased to avoid too much conflicts.


eregon:
@raeoks I just took your commits and organized them a bit differently.

PR #3167, opened 2015-07-22
normalize path (windows)

### LINKED ISSUE #3132: expandPathInternal Windows compatibility
MRI: File.realpath 'c:/test' => 'c:/test'
jRuby: File.realpath 'c:/test' => 'C:\ \test' (tested on 1.7.19 and 9.0.0.0 rc2)

It causes problems in Rails applications, because Rails.root path is invalid.

https://github.com/jruby/jruby/blob/da5dad6a960425375a2921c54fccd55971250076/core/src/main/java/org/jruby/RubyFile.java#L1592

ahorek:
fixes #3132

### DIFF
--- a/core/src/main/java/org/jruby/RubyFile.java
+++ b/core/src/main/java/org/jruby/RubyFile.java
@@ -1690,7 +1690,7 @@ public class RubyFile extends RubyIO implements EncodingCapable {
 
         if (canonicalize) {
             try {
-                realPath = new File(realPath).getCanonicalPath();
+                realPath = JRubyFile.normalizeSeps(new File(realPath).getCanonicalPath());
             } catch (IOException ioe) {
                 // Earlier canonicalization will have to do.
             }
--- a/test/jruby/test_file.rb
+++ b/test/jruby/test_file.rb
@@ -208,6 +208,12 @@ class TestFile < Test::Unit::TestCase
     def test_pathname_windows
       assert_equal(Pathname('foo.bar.rb').expand_path.relative_path_from(Pathname(Dir.pwd)), Pathname('foo.bar.rb'))
     end
+
+    # JRUBY-3132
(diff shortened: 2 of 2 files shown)


kares:
@ahorek seems good but without a test is a little less worthy, any chance you do a test-case or tell us what use-case (sample .rb script) has been failing for you that the change fixes? thanks

ahorek:
@kares - hi, most jruby tests don't work on windows and this is windows specific. Failing scenario is already described in the related issue.

spec/ruby/core/file/realpath_spec.rb
it "returns '/' when passed '/'" do
  File.realpath('/').should == '/'
end

=> for windows it should look something like this:
it "returns '/' when passed '/'" do
  if is_windows?
    File.realpath('/').should == 'C:/' # current behaviour 'C:\'
  else
    File.realpath('/').should == '/' # linux
  end
end

kares:
seems good ... @mkristian what do you think of: `File.realpath('/').should == 'C:/' # NOT 'C:\'`

mkristian:
the usual java vs. ruby paths. yes, this is OK. still a test which passes on windows as on unix would be great. one there we will have a CI running on windows and then it is important that we have a good test coverage for those windows only cases. test/jruby/test_file.rb is also a good place for such a test as this is in jruby direct control.

kares:
@mkristian thanks I wasn't sure since its been a while when a tried JRuby on Windows. 
shall wait a while if @ahorek decides to add some cases at `test/jruby/test_file.rb` or merge as is.

ahorek:
@kares - I've added very simple cases, is it ok? this fix should be also backported to JRuby 1.7.23

kares:
@ahorek looks good, we're in the process of merging jruby-1_7 into master, 
... after we shall merge these and likely even back-port to jruby-1_7. thanks.

Issue #3182, opened 2015-07-24
IO.puts is not thread safe


samwgoldman:
I have some logging code that looks like this:

``` ruby
def log(line)
  io.puts(line)
  io.fsync
end
```

This code is called from multiple threads.

In production, I've noticed lines that there are lines that look like this: `"#{line1}#{line2}\n\n"`.

I looked into it, and it looks like `puts` might have a race condition between writing the line and writing the subsequent newline. See https://github.com/jruby/jruby/blob/0be4796b08f027c5cdf8c7400ca0563237db5031/core/src/main/java/org/jruby/RubyIO.java#L2526.

Some repro code:

``` ruby
require 'thread'
8.times.map do
  Thread.new do
    1000.times do
      $stdout.puts "line"
      $stdout.fsync
    end
  end
end.map(&:join)
```

To test:

`ruby puts.rb | grep lineline`

/cc @cheald

Freaky:
MRI does exactly the same thing: https://github.com/ruby/ruby/blob/trunk/io.c#L7102

You've always had to do `io.write "line\n"` to avoid it.

cheald:
As Freaky points out, MRI does indeed exhibit this interleaving behavior.

That said, I would argue the behavior is surprising and probably should be fixed in MRI, too!

PR #3183, opened 2015-07-24
Use a single write for IO#puts

cheald:
This is a fix for #3182

Avoids interleaved writes across threads:

```
require 'thread'
8.times.map do
  Thread.new do
    1000.times do
      $stdout.puts "line"
      $stdout.fsync
    end
  end
end.map(&:join)
```

This can currently produce output like:

```
lineline


linelineline


lineline

lineline
```

After patching, output is consistently:

```
line
line
line
line
line
line
line
line
line
line
line
line
```

### DIFF
--- a/core/src/main/java/org/jruby/RubyIO.java
+++ b/core/src/main/java/org/jruby/RubyIO.java
@@ -2520,11 +2520,11 @@ public class RubyIO extends RubyObject implements IOEncodable {
             line = arg.asString().getByteList();
         }
 
-        write(context, maybeIO, line);
-
         if (line.length() == 0 || !line.endsWith(separator.getByteList())) {
-            write(context, maybeIO, separator.getByteList());
+            line.append(separator.getByteList());
         }
+
+        write(context, maybeIO, line);
     }
 
     private static IRubyObject inspectPuts(ThreadContext context, IRubyObject maybeIO, RubyArray array) {


headius:
This one came up on IRC today.

I'm not opposed to this, but currently we implement it the same way MRI does (nearly line-for-line). I'd like to know if there's a reason it's implemented that way, since there are even specs and tests that check for the separate writes.

headius:
Oh, I did think of one explanation: zero-allocation puts. If you have to stick the `\\n` on a buffer to do it as one write, you'll probably have to allocate a new buffer.

Our puts logic currently isn't zero-alloc, but it could be.

enebo:
@headius I think this has languished because of the fear of the unknown.  In most of our optimizations of reducing dyncalls we cannot observe a change in behavior.  The behavior @cheald fixes as was pointed out in original report is what MRI does as well (weird interleaved output).  So even if this is faster it will change behaviorally.    If we think no one will really want this behavior (and that is likely true), then we need to consider test suites mocking to make sure there are two write() calls.  The only way of getting around that would be to check if this is being called from builtin version of the method.  Seems like this is a ways beneath that point.  We could pass that and then remove the cost?

headius:
The cost of the two dynamic calls should be close to negligible at this point, since we added call-site caching. The only real benefit is performing the write in one go.

An alternative would be to lock the IO around the `puts` logic, but that would also need to check if `write` had been replaced (since there's no IO-locking semantics in MRI puts).

headius:
I don't think we'll do this. If you are using an IO across threads, it guarantees only that its internal logic and buffering will be thread-safe, not that any given call will be atomic (and indeed, even `write` calls might get turned into multiple native calls). If you want atomicity of groups of writes, including implicit cases like `puts`, you can introduced a lock or use the JRuby::Synchronized module.

Issue #3173, opened 2015-07-24
Can't reference the proc itself in a proc


yamam:
In a proc defined in a method, you can't reference the proc itself of local variable.

```ruby
def mk_proc
    proc1 = proc do
        p proc1
    end
end
mk_proc.call
```

```
> jruby-9.0.0.0/bin/jruby -v test.rb
jruby 9.0.0.0 (2.2.2) 2015-07-21 e10ec96 Java HotSpot(TM) Client VM 25.51-b03 on 1.8.0_51-b16 +jit [Windows 7-x86]
io/console not supported; tty will not be manipulated
nil

> jruby-1.7.21/bin/jruby -v test.rb
jruby 1.7.21 (1.9.3p551) 2015-07-07 a741a82 on Java HotSpot(TM) Client VM 1.8.0_51-b16 +jit [Windows 7-x86]
io/console not supported; tty will not be manipulated
#<Proc:0xd992b2@test.rb:2>

> ruby -v test.rb
ruby 2.0.0p576 (2014-09-19) [i386-cygwin]
#<Proc:0x2027282c@test.rb:2>
```

subbuss:
Seems fixed with 8f8081649a7ecb030b8da1b692ce425ce6edf908

subbuss:
This is not yet quite fixed.

subbuss:
Now fixed.

dirk:
@subbuss: Thanks for including detailed commit messages on those two commits; feel like I learned a lot from reading those and the diff. :smile:

subbuss:
@dirk I tend to write verbose commit messages :-) glad they were helpful.

PR #3187, opened 2015-07-25
Add MRI test for proc visibility

dirk:
For #3173 and to check against future regression.

This was broken in 9.0.0.0 but seems to be fixed in 9.0.1.0.


enebo:
Thanks. @headius or @eregon can we please merge this back over to MRI source?

dirk:
@enebo: Glad to help!

PR #3191, opened 2015-07-27
[Truffle] splat position; Thread#backtrace

pitr-ch:


eregon:
I would call *arg in an argument list a `rest` argument instead of `splat`.
Why do you want the caller to perform the computation of the index, could Arity do it if we just pass required, optional and whether there is a rest argument?

pitr-ch:
I think what you suggest is not enough because of post-splat arguments. The position can be smaller than required+optional. It could be pre, optional, splat?, post though.

eregon:
How so? The frame will contain the arguments in order, so for instance optional will be either the given value or the default one.

truffle/src/main/java/org/jruby/truffle/nodes/arguments/CheckArityNode.java
```diff
@@ -47,8 +47,8 @@ public void executeVoid(VirtualFrame frame) {
         final Object[] frameArguments = frame.getArguments();
         final int given;
         final RubyBasicObject keywordArguments;
-        
-        //TODO (MS): Check merge 
+
+        //TODO (MS): Check merge
```
eregon: Can you remove these unintentional changes?

pitr-ch: done, what's our policy on these changes, separate commits or don't? I feel these trailing spaces as malformed code.

eregon: ah sorry I misread the diff, I thought the diff added the extra spaces.
Removing neighboring style issues is fine.

eregon:
Looks good, bonus point if you can rebase on master :)

PR #3192, opened 2015-07-27
[Truffle] Add to loadedFeatures before loading

bbelleville:
For classloader and core features, add the feature to the collection of
loaded features before performing the load to break endless recursion in
the presence of cyclic dependencies.

This brings the behavior of core and classloader features in line with file features. I am encountering a cyclic dependency when doing ahead of time compilation. Since I include all ruby files in the image, features that would have been resolved as files instead are found with the classloader.

A quick way to expose this issue is to add `require_relative 'core.rb'` to truffle/src/main/ruby/core.rb.


eregon:
Sounds good, this is already how the normal files are loaded,
except that additionally the path is removed from $LOADED_FEATURES if there is an error.
Proper locking around require/load (which needs to be done) should also help to detect this recursion.

PR #3209, opened 2015-08-01
Correct a format string in Signal#trap.

### LINKED ISSUE #3208: Trapping reserved signals doesn't print a meaningful message
Let's trap a reserved signal:

```ruby
trap(:ILL) {}
```

Output from MRI:

```
ArgumentError: can't trap reserved signal: SIGILL
```

Output from JRuby 9.0.0.0

```
ArgumentError: malformed format string - %S
from org/jruby/RubyString.java:1168:in `%'
```

It seems like JRuby tries to print the error, but fails for some reason.

Freaky:
Fixes #3208: Trapping reserved signals doesn't print a meaningful
message

### DIFF
--- a/core/src/main/ruby/jruby/kernel/signal.rb
+++ b/core/src/main/ruby/jruby/kernel/signal.rb
@@ -4,7 +4,7 @@ module Signal
     sig = sig.to_s.sub(/^SIG(.+)/,'\1')
 
     if RESERVED_SIGNALS.include?(sig)
-      raise ArgumentError.new("can't trap reserved signal: SIG%S" % sig)
+      raise ArgumentError.new("can't trap reserved signal: SIG%s" % sig)
     end
 
     oldhandler, installed = if block
--- a/spec/ruby/core/signal/trap_spec.rb
+++ b/spec/ruby/core/signal/trap_spec.rb
@@ -88,6 +88,12 @@ describe "Signal.trap" do
     Signal.trap(:HUP, "IGNORE").should == "IGNORE"
   end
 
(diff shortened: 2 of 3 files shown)


rtyler:
The failure below from Travis doesn't look related, but would it be possible to add a test to ensure that #3208 is really fixed?

```
     [exec] 
     [exec] Failures:
     [exec] 
     [exec]   1) FFI::Function can wrap a blocking function
     [exec]      Failure/Error: expect(Time.now - time).to be >= 2
     [exec]      NameError:
     [exec]        uninitialized constant RSpec::Matchers::BuiltIn::Be
     [exec]      # ./spec/ffi/function_spec.rb:71:in `block in (root)'
     [exec] 
     [exec] Finished in 21.64 seconds
     [exec] 4843 examples, 1 failure
     [exec] 
     [exec] Failed examples:
     [exec] 
     [exec] rspec ./spec/ffi/function_spec.rb:65 # FFI::Function can wrap a blocking function
```

Freaky:
Certainly.  Shall I just add a new test to RubySpec?

dirk:
@Freaky: I'm guessing it should be added to the MRI `TestSignal` test case in [this file](https://github.com/jruby/jruby/blob/master/test/mri/ruby/test_signal.rb).

Freaky:
Did both.

kares:
picked the fix and test as 18df48a87559c1ab07c4e8b1015575e4104e8667 and 0adaddf41ac2625853be876602593f6b645ada83 on master ... Thanks Thomas!

PR #3218, opened 2015-08-03
add to_h to the class used for ENV

### LINKED ISSUE #3162: Different results for `to_h` and `to_hash` on `ENV`
Hi,

converting `ENV` to a `Hash` gives different results. It depends on which `to_`-method is used.

*MRI*

~~~ruby
ENV.object_id
# => 17981260
ENV.to_h.object_id
# => 22183040
ENV.to_hash.object_id
# => 22148380
~~~

*JRuby*

~~~ruby
ENV.object_id
# => 2042
ENV.to_h.object_id
# => 2042
ENV.to_hash.object_id
# => 2044
~~~

reuben-sutton:
That makes it so that it duplicates rather than returning the same object to make compatible with MRI

fixes #3162

### DIFF
--- a/core/src/main/java/org/jruby/RubyGlobal.java
+++ b/core/src/main/java/org/jruby/RubyGlobal.java
@@ -338,6 +338,12 @@ public class RubyGlobal {
             return getRuntime().newString("ENV");
         }
 
+        @JRubyMethod
+        public RubyHash to_h(){
+          return to_hash();
+        }
+
+
     }
 
     /**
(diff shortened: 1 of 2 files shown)


kares:
change seem to introduce a failure e.g. here https://travis-ci.org/jruby/jruby/jobs/73959532#L3277-L3284

reuben-sutton:
@kares So it does. Doesn't happen on oraclejdk8, so installing openjdk7 to see if I can reproduce there.

reuben-sutton:
hi @kares - just found the issue and committed a fix.

kares:
since this is rubyspecs your updating it needs a certain level of quality, you seem to set `ENV['foo']` for no reason ... either assert smt around it (maybe even change the key name to not collide with another spec) or simply remove the first/last lines. 

we also need the spec change to be in separate commits (so it can be easy ported upstream), thus please update your commits to be clean (one fixinig `to_h` another one adding a spec)

reuben-sutton:
Hi @kares, that's done. Thanks!

kares:
@reuben-sutton excellent, thanks! this should be good to go - we're just wait to hear about travis-ci (no worries about unrelated failures)

PR #3219, opened 2015-08-03
treat port nil as 0 for pack_sockaddr_in

### LINKED ISSUE #3156: Socket.pack_sockaddr_in fails for nil port (MRI doesn't)
JRuby raises a TypeError when the port is nil, whereas MRI treats nil the same way as zero.

```shell
> ruby-2.2.2 -rsocket -e 'p Socket.pack_sockaddr_in(nil, "127.0.0.1")'
"\x10\x02\x00\x00\x7F\x00\x00\x01\x00\x00\x00\x00\x00\x00\x00\x00"

> jruby-1.7.19 -rsocket -e 'p Socket.pack_sockaddr_in(nil, "127.0.0.1")'
TypeError: no implicit conversion from nil to integer
  pack_sockaddr_in at org/jruby/ext/socket/RubySocket.java:272
            (root) at -e:1
```

I'm using 1.7.19 in my example, but it seems the code in SocketUtils hasn't changed since 2012 in master, so I'm guessing it still applies to both 1.7.x and 9k. See https://github.com/jruby/jruby/blame/26d0b20ca324d082468bbcbc9d71709b142aa77b/core/src/main/java/org/jruby/ext/socket/SocketUtils.java#L131-L140

reuben-sutton:
Ensure compatability with MRI Ruby

closes #3156

### DIFF
--- a/core/src/main/java/org/jruby/ext/socket/SocketUtils.java
+++ b/core/src/main/java/org/jruby/ext/socket/SocketUtils.java
@@ -129,9 +129,13 @@ public class SocketUtils {
     }
 
     public static IRubyObject pack_sockaddr_in(ThreadContext context, IRubyObject port, IRubyObject host) {
-        int portNum = port instanceof RubyString ?
+        int portNum = 0;
+        
+        if(!port.isNil()){
+          portNum = port instanceof RubyString ?
                 Integer.parseInt(port.convertToString().toString()) :
                 RubyNumeric.fix2int(port);
+        }
 
         return Sockaddr.pack_sockaddr_in(
                 context,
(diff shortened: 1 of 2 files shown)


kares:
same here, please add the *spec/ruby/library/socket/shared/pack_sockaddr.rb* change in a separate commit

kares:
... also maybe you could instead open a PR targeting **jruby-1_7** ... otherwise we have to cherry-pick

reuben-sutton:
@kares not a problem - it's also an issue on master, so should I send two?

reuben-sutton:
Hi @kares - That's done. It affected both 9k and 1.7 so I added #3225 to backport.

kares:
great thanks, for the future if you only target 1_7 it will eventually end up in master as well with the merge
... for this one do not worry about it, thanks again.

PR #3222, opened 2015-08-04
Close any open selectors when tearing down a runtime.

bbrowning:
This fixes a runtime leak in embedded scenarios where a runtime may
never get garbage collected because it has a thread blocked on a
select call. This leak exists in JRuby 1.7.x as well, but for some
reason I'm seeing it more often in JRuby 9.x when running TorqueBox 3
integration tests. Perhaps there is more IO that uses select in 9.x
than 1.7.x?

Here's a gist with a fairly simple reproducer and results on 1.7.21, 9.0.0.0, and 9.0.1.0-SNAPSHOT with this patch applied - https://gist.github.com/bbrowning/3cbb3901d4facd7fc99f

### DIFF
--- a/core/src/main/java/org/jruby/util/io/SelectorPool.java
+++ b/core/src/main/java/org/jruby/util/io/SelectorPool.java
@@ -31,6 +31,7 @@ package org.jruby.util.io;
 
 import java.io.IOException;
 import java.nio.channels.Selector;
+import java.util.ArrayList;
 import java.util.List;
 
 import java.nio.channels.spi.SelectorProvider;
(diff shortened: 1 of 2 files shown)


mkristian:
@bbrowning does it makes sense to turn the reproducer from the gist into a test case ?

bbrowning:
Perhaps - my only hesitation was I didn't see an easy way to get around needing a sleep in the test case to make sure the runtimes have all actually been torn down by the time we check for threads hanging around.

kares:
+1 for doing this on **jruby-1_7** as well

bbrowning:
I've rebased this on top of JRuby master and turned the reproducer into a test case per the suggestion by @mkristian.

PR #3226, opened 2015-08-05
[Truffle] Limit the number of methods available for runtime compilation

bbelleville:
This pull request lowers the number of methods available for runtime
compilation from over 33,000 to less than 15,000. Some methods that
could not be annotated with TruffleBoundary were split in order to
introduce a boundary.

I used a static analysis tool to determine what methods were causing a large number of other methods to be included for runtime compilation and added Truffle boundaries to limit this.

### DIFF
--- a/truffle/src/main/java/org/jruby/truffle/nodes/core/ExceptionNodes.java
+++ b/truffle/src/main/java/org/jruby/truffle/nodes/core/ExceptionNodes.java
@@ -9,6 +9,7 @@
  */
 package org.jruby.truffle.nodes.core;
 
+import com.oracle.truffle.api.CompilerDirectives.TruffleBoundary;
 import com.oracle.truffle.api.dsl.Specialization;
 import com.oracle.truffle.api.nodes.Node;
 import com.oracle.truffle.api.source.SourceSection;
@@ -114,6 +115,7 @@ public abstract class ExceptionNodes {
             readCustomBacktrace = new ReadHeadObjectFieldNode("@custom_backtrace");
         }
 
+        @TruffleBoundary
         @Specialization
         public Object backtrace(RubyBasicObject exception) {
             if (readCustomBacktrace.isSet(exception)) {
--- a/truffle/src/main/java/org/jruby/truffle/nodes/core/ProcessNodes.java
+++ b/truffle/src/main/java/org/jruby/truffle/nodes/core/ProcessNodes.java
@@ -84,6 +84,7 @@ public abstract class ProcessNodes {
             return timeToUnit(time, unit);
         }
 
+        @TruffleBoundary
         @Specialization(guards = { "isThreadCPUTime(clock_id)", "isRubySymbol(unit)" })
         protected Object clock_gettime_thread_cputime(int clock_id, RubyBasicObject unit,
                 @Cached("getLibCClockGetTime()") LibCClockGetTime libCClockGetTime) {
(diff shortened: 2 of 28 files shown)


chrisseaton:
These all seem fine.

truffle/src/main/java/org/jruby/truffle/nodes/core/ExceptionNodes.java
```diff
@@ -114,6 +115,7 @@ public BacktraceNode(RubyContext context, SourceSection sourceSection) {
             readCustomBacktrace = new ReadHeadObjectFieldNode("@custom_backtrace");
         }
 
+        @TruffleBoundary
```
eregon: `asRubyStringArray()` should probably be the boundary, the rest should not have too many nodes.

truffle/src/main/java/org/jruby/truffle/nodes/core/ProcessNodes.java
```diff
@@ -84,6 +84,7 @@ protected Object clock_gettime_realtime(int clock_id, RubyBasicObject unit) {
             return timeToUnit(time, unit);
         }
 
+        @TruffleBoundary
```
eregon: This should be on `getLibCClockGetTime()` instead I think.

eregon:
It is a bit unfortunate that `ThreadLocal.get()` must be a boundary (probably only the slow path of it needs it). It might not matter much in practice though.
The rest looks fine, but it would be nice to reconsider the couple ones I highlighted.

chrisseaton:
`ThreadLocal.get()` is a map lookup so will also have to be boundary. It isn't just a native call.

eregon:
And with that I forgot:
very good job finding those @bbelleville!

eregon:
@chrisseaton if we get in `getEntryAfterMiss()` then yeah it is not pretty, but otherwise it looks all fine.
We should probably have a way to annotate existing JDK methods.

truffle/src/main/java/org/jruby/truffle/nodes/core/ProcessNodes.java
bbelleville: This one might actually be specific to ahead of time compilation, what causes the explosion is when the jnr code makes something reachable that is unsupported, calling error handling code. I'll look into this a little more.

bbelleville: I added support in the ahead of time compilation system for the unsupported feature that made error handling code reachable. We still need the TruffleBoundary here at least for ahead of time compilation so that it doesn't create a deopt target for the native clock_gettime.

truffle/src/main/java/org/jruby/truffle/nodes/core/ExceptionNodes.java
bbelleville: You're right, the tool shows that asRubyStringArray is the method that makes a lot reachable for compilation.

eregon: OK, I'll fix on master :smiley:

truffle/src/main/java/org/jruby/truffle/nodes/core/ProcessNodes.java
eregon: OK, thanks for looking at it!

PR #3228, opened 2015-08-06
Fix an embedded runtime memory leak in Java7ClassValue

bbrowning:
Commit 09f15f902ee14a848387b9f126cb731941a0a8be enabled the
Java7ClassValue implementation by default which exposed a runtime
memory leak in the embedded use case. This fixes that leak by wrapping
values stored in the java.lang.ClassValue instance with a
WeakReference.

### DIFF
--- a/core/src/main/java/org/jruby/util/collections/Java7ClassValue.java
+++ b/core/src/main/java/org/jruby/util/collections/Java7ClassValue.java
@@ -1,5 +1,9 @@
 package org.jruby.util.collections;
 
+import java.lang.ref.WeakReference;
+import java.util.ArrayList;
+import java.util.List;
+
 /**
  * A proxy cache that uses Java 7's ClassValue.
  */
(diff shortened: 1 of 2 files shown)


headius:
Related screenshot: https://dl.dropboxusercontent.com/u/9213410/class-values-leak.png

I've posted a question to the mlvm-dev list asking whether this rooting is expected. I was under the impression that when the ClassValue goes away, the value it references should be dereferenced, as in ThreadLocal.

bbrowning:
For reference, I'm seeing this leak when using:
```
$ java -version
openjdk version "1.8.0_51"
OpenJDK Runtime Environment (build 1.8.0_51-b16)
OpenJDK 64-Bit Server VM (build 25.51-b03, mixed mode)
```

bbrowning:
Even if this turns out to be a bug upstream in Java and it gets fixed, we'll still need to disable the usage of Java7ClassValue by default and/or apply this patch. Otherwise we're enabling an embedded runtime memory leak by default.

headius:
Will just disable to use of ClassValue for now, pending a fix from OpenJDK folks.

PR #3231, opened 2015-08-06
Remove the requirement for varargs when generating Class objects

### LINKED ISSUE #3206: Invocation of become_java! instances fails with "cannot make variable arity"
When experimenting with passing objects for invocation back and forth into Nashorn, I wrote the following code:

```ruby
class Example
  java_signature "java.lang.String hello()"
  def hello
    return 'hello world from Ruby'
  end
end

ex = Example.become_java!.new
# pass `ex` into Nashorn
```

From the JavaScript, which uses reflection for invocation, I invoked this `#hello` raises this error:
```
MemberName.java:852:in `makeAccessException': java.lang.IllegalAccessException: cannot make variable arity: rubyobj.Example.hello()String/invokeVirtual
```

@headius wrote [this patch](https://gist.github.com/headius/23ad5619bf804db574e7)  which, when compiled and tested, properly allows the invocation of this JRuby-originating instantiated Java class

rtyler:
This is effectively a patch written by @headius and tested locally and applied
by me. In my use-case this allows Java code to properly reflect back into a
Ruby-generated Class and invoke a method which takes no arguments.

Fixes #3206

### DIFF
--- a/core/src/main/java/org/jruby/RubyClass.java
+++ b/core/src/main/java/org/jruby/RubyClass.java
@@ -1391,7 +1391,7 @@ public class RubyClass extends RubyModule {
                 switch (methodEntry.getValue().getArity().getValue()) {
                 case 0:
                     signature = sig(IRubyObject.class);
-                    m = new SkinnyMethodAdapter(cw, ACC_PUBLIC | ACC_VARARGS, javaMethodName, signature, null, null);
+                    m = new SkinnyMethodAdapter(cw, ACC_PUBLIC, javaMethodName, signature, null, null);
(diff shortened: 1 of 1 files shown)


kares:
seems like this would be useful on **jruby-1_7** as well ?

rtyler:
@kares I agree, this was really @headius' patch from JRubyConf EU, so I figured he'd take it over at some point :)

kares:
#3206 has been fixed - a similar change came in with #3779 (in JRuby 9.1)

headius:
Hah, I totally forgot about this. Thanks @kares!

PR #3235, opened 2015-08-07
Rewrite PTY support

Freaky:
This reworks `PTY#spawn` to use `openpty(3)` (available on BSD, OS X and Linux) and `Process#spawn` instead of `forkpty(3)`, which as the name suggests, forks and is thus unsafe.  It also fixes use with multiple arguments.

It also adds `PTY#open` and `PTY#check`, and properly defines `PTY#getpty` as an alias to `PTY#spawn`.

It passes MRI's test_pty.rb, after a slight tweak to close a race condition:

```
Finished tests in 15.752000s, 0.9523 tests/s, 2.4759 assertions/s.
15 tests, 39 assertions, 0 failures, 0 errors, 0 skips

ruby -v: jruby 9.0.1.0-SNAPSHOT (2.2.2) 2015-08-01 6a9fcf8 OpenJDK 64-Bit Server VM 25.51-b03 on 1.8.0_51-b16 +jit [FreeBSD-amd64]
```

Needs testing and review, especially outside FreeBSD.

### DIFF
--- a/lib/ruby/stdlib/pty.rb
+++ b/lib/ruby/stdlib/pty.rb
@@ -1,6 +1,18 @@
 require 'ffi'
 
 module PTY
+  class ChildExited < RuntimeError
+    attr_reader :status
+
+    def initialize(status)
+      @status = status
+    end
+
+    def inspect
+      "<#{self.class.name}: #{status}>"
+    end
+  end
+
   private
   module LibUtil
     extend FFI::Library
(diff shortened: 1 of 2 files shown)


Freaky:
Still issues with this.  e.g. spawning off a copy of `yes` and closing both descriptors doesn't send it a SIGHUP.  `cat` works, though, or the test suite would hang.

Freaky:
Closing - committed in 4f8e6bde94007cae11e7f444f6db6a18f1d2682d

PR #3240, opened 2015-08-10
[Truffle] Updating to most recent version of Truffle API. Compiles, but does no…

jtulach:
…t implement TruffleLanguage.parse yet. 

Here are the syntactic changes needed to update to new version of Truffle API. However the bigger change is the *parse* method - it is supposed to be context-less - I have no idea how to implement it properly, the parsing seems to require reference to context right now.

### DIFF
--- a/truffle/pom.rb
+++ b/truffle/pom.rb
@@ -15,10 +15,10 @@ project 'JRuby Truffle' do
   repository( :url => 'http://lafo.ssw.uni-linz.ac.at/nexus/content/repositories/snapshots/',
               :id => 'truffle' )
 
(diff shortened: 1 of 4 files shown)


eregon:
Some parsing like in `a=42; eval("a")` needs the context and the surrounding frame, so for some cases it's needed. How do other languages handle this?

If we don't need the parent frame (parsing a whole file) then we use the context to reference core classes and such. Forcing them to do it without the context might not be very nice (does it mean everything depending on context has to be set up at the first `execute()`?).

pitr-ch:
@jtulach Could you add a `[Truffle]` prefix to the commit?

jtulach:
@eregon: In general the parsing of source shouldn't require any context at all. It's just about understanding the source syntax, not really resolving symbols, etc.

PR #3250, opened 2015-08-13
Fix homepage of jruby-jars.gemspec

frsyuki:
"Homepage" link shown at https://rubygems.org/gems/jruby-jars is 404 Not Found.


mkristian:
thanx :)

PR #3258, opened 2015-08-16
[Truffle] Use the OM more efficiently and create a DSL to help us do that

chrisseaton:
You will probably have to pull locally to review this, and even then it's going to be very hard as all `RubyBasicObject` and subclasses have been replaced with `DynamicObject`.

Shootout benchmarks all compile. Those that use arrays are around running at around 60%, but I think I know what this bug is. No speedups yet - but those that we might expected to speedup are those that currently have that regression.


headius:
The jnr-ffi commit above is unrelated...meant for #3285.

PR #3265, opened 2015-08-19
Updates for RSpec 3

benlovell:
Updates JI integration specs to RSpec 3 syntax. This was performed mostly via `transpec` but did require a modicum of manual fixups. Some of the key gotchas were around the semantics of `pending` and `skip` - especially the block form of `pending` which has since been deprecated. The `raise_error` matcher now requires an explicit error to reduce false-positives.

I made a few other minor cleanups and there were some flat-out broken specs that were fixed too.

There is still plenty of room for improvement in the tests that I'll be happy to soldier on with. 

The history isn't the cleanest so I'll squash this down to something sane once it's green but it should be good to go otherwise. :heart:


kares:
great, was hoping to have it on **jruby-1_7** but I guess it won't work in --1.8 ;( mode ... thus 9K it is!

kares:
changing `be_false`, `be_true`, `be_nil` to `be true` ... should handle most of the `spec:ji` failures

benlovell:
Of course now having upgraded to RSpec 3 there are other failures. I'll get those sorted now - they seem to be mostly around the `pending` change. :see_no_evil:

benlovell:
It seems those outstanding CI issues are unrelated to my changes, so I think this is good to go otherwise. I'll squish it down to something sensible when I get the OK.

benlovell:
@kares thanks for reviewing :heart_eyes: Those issues should be addressed now.

kares:
:purple_heart: :relieved: looks :rocket: :ship: **OK** to me ... do you want to merge as is or squish it :arrow_down: ?

p.s. also there's a :green_heart: [base](https://github.com/jruby/jruby/commit/cbc0e410445552d8bf9ba43c1625d46c79a412ad) if you're rebasing (but I do not think @enebo will mind)

benlovell:
@kares I rebased and squashed into one logical commit. :ship: :shipit: :heart_eyes:

enebo:
the spec ci run finished green so no need to wait for osgi++ to finish....thanks @benlovell this is massive!

PR #3269, opened 2015-08-20
[Truffle] Refactor some of the require system

chrisseaton:
* Fixes the problem we had loading stdlib from a jar.
* Allows you to modify where core is loaded from with `-Xtruffle.core.load_path=truffle/src/main/ruby` for better stack traces
* Some general tidy up


spec/truffle/tags/core/thread/backtrace/location/to_s_tags.txt
```diff
@@ -0,0 +1 @@
+fails:Thread::Backtrace::Location#to_s converts the call frame to a String
```
chrisseaton: I should have fixed these.

truffle/src/main/java/org/jruby/truffle/runtime/loader/SourceCache.java
```diff
@@ -0,0 +1,72 @@
...
+ */
+
+package org.jruby.truffle.runtime.loader;
+
+import com.oracle.truffle.api.source.Source;
+
+import java.io.IOException;
+import java.util.*;
+
+public class SourceCache {
```
eregon: Why do we cache file contents?
Normal use cases will only require/load a file once, isn't it? (and if not there should be some check if not changed on disk!)

chrisseaton: Mainly for the debugger. If I say I want to break on `foo.rb`, I need to know what files we have loaded that could fuzzily match that name.

What we have here isn't an overhead anyway as at the moment Truffle is also caching sources to some extent. They both probably need to get some eviction capability.

pitr-ch:
Nice. Would it make sense to have `Xtruffle.core.load_path=truffle/src/main/ruby` as default in `jt`.

chrisseaton:
Yes - I don't think that could ever mean you were testing in an inconsistent state so that should be fine - I'll add that.

Issue #1186, opened 2013-10-29
BigMath.exp and log missing


headius:
Need to implement these.

headius:
exp was added in 9.0.1.

Issue #3242, opened 2015-08-10
BigDecimal initializer does not mirror YARV behavior


garysweaver:
The BigDecimal [API doc](http://ruby-doc.org/stdlib-2.2.2/libdoc/bigdecimal/rdoc/BigDecimal.html#method-c-new) says the second argument (digits) for BigDecimal's initializer should be: "The number of significant digits, as a Fixnum. If omitted or 0, the number of significant digits is determined from the initial value." But, in reality, Ruby 2.2.2 seems to ignore the second argument unless it is higher than 18, at least on my system, e.g.
```
2.2.2 :001 > require 'bigdecimal'
 => true 
2.2.2 :002 > BigDecimal.new '31.0344830', 0
 => #<BigDecimal:7fbc4c1ee5c8,'0.31034483E2',18(18)> 
2.2.2 :003 > BigDecimal.new '31.0344830', 3
 => #<BigDecimal:7ffc0114ca80,'0.31034483E2',18(18)> 
2.2.2 :004 > BigDecimal.new '31.0344830', 18
 => #<BigDecimal:7ffc01083dd8,'0.31034483E2',18(18)> 
2.2.2 :005 > BigDecimal.new '31.0344830', 19
 => #<BigDecimal:7ffc01073e38,'0.31034483E2',18(27)> 
2.2.2 :006 > BigDecimal.new '31.0344830', 199
 => #<BigDecimal:7ffc01870c58,'0.31034483E2',18(207)>
2.2.2 :007 > BigDecimal.new '31.03', 0
 => #<BigDecimal:7ffc01031060,'0.3103E2',18(18)> 
2.2.2 :008 > BigDecimal.new '31.03', 1
 => #<BigDecimal:7ffc030935d8,'0.3103E2',18(18)> 
2.2.2 :009 > BigDecimal.new '31.03', 18
 => #<BigDecimal:7ffc03088e80,'0.3103E2',18(18)> 
2.2.2 :010 > BigDecimal.new '31.03', 19
 => #<BigDecimal:7ffc0307aba0,'0.3103E2',18(27)> 
2.2.2 :011 > BigDecimal.new '31.03', 199
 => #<BigDecimal:7ffc0306a250,'0.3103E2',18(207)> 
```

JRuby 9.0.0.0 on the other hand seems to try to set scale using some other mechanism that appears to be based on the value of the string used for initialization and from what I can tell, and seems to be ignoring the second argument:
```
[...]

PR #3271, opened 2015-08-20
bigdecimal & bigmath

ahorek:
Implemets:
bigdecimal#.to_r
bigmath#.exp

Fixes:
CMP/COERCE bigdecimals -> floats / rationals
Exceptions
Missing require
Tests, clean up

References:
#1186


kares:
you seem to have added a lot of new excludes ... those were passing before, right?
... also your commit messages could use some chit-chat :)

ahorek:
there are some minor cases when we don't match MRI:
1/ sometimes precision isn't limited
2/ special modes like BigDecimal::ROUND_MODE & BigDecimal::EXCEPTIONS are sometimes ignored / not implemented
3/ BigDecimal.pow works but it's extremely slow (see test_exception_overflow)
4/ MRI's BigDecimal can handle larger exponents, Java's BigInteger class can not
BigDecimal("10E1999999999") MRI => 0.1E2000000000, JRuby => Infinity
5/ #3242

nice to have:
BigMath.log / exp should be implemented natively like MRI did since Ruby 2.0.

all my changes are complatible with MRI 1.9.3, so backporting shouln't be a problem. I've also added comments and fixed broken tests.

enebo:
@ahorek can we trouble you for a second PR for jruby-1_7 branch?  I suspect we cannot cherry-pick this.

PR #3290, opened 2015-08-27
[Truffle] Allow for lazy parsing

bbelleville:
These changes make it possible to lazily parse methods in ahead of time compiled images. During image generation method bodies are elided from the core source files and stored in a separate data structure. When the file is initially parsed, the method body is replaced with a node to perform the lazy parsing. When this node is executed, it performs the parsing and then replaces itself with the code of the method.

### DIFF
--- a/truffle/src/main/java/org/jruby/truffle/translator/MethodTranslator.java
+++ b/truffle/src/main/java/org/jruby/truffle/translator/MethodTranslator.java
@@ -9,6 +9,9 @@
  */
 package org.jruby.truffle.translator;
 
+import java.util.ArrayDeque;
+import java.util.Deque;
+
 import com.oracle.truffle.api.CallTarget;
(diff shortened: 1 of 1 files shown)


eregon:
Is it lazy parsing or lazy translation? I would guess the latter, right?

bbelleville:
It does parsing and translation of method bodies lazily. During image generation I parse all ruby files, identify the method boundaries, and delete the method bodies. So a method like:
```
def sort!(&block)
    replace sort(&block)
end
```
Becomes:
```
def sort!(&block)
end
```
The actual code of the method body is stored in another data structure. When a file is parsed, if I have entries for the methods in that file, I insert a node that does lazy parsing as the method's body. This node gets the original code for the method, parses and translates it, and replaces itself with the result.

eregon:
Nice, merging!

PR #3291, opened 2015-08-27
[Truffle] Canonicalize input file name

bbelleville:
Currently, if the input file name is something like "./hello.rb", the assertion at https://github.com/jruby/jruby/blob/master/truffle/src/main/java/org/jruby/truffle/runtime/loader/SourceLoader.java#L37 will fail. Getting the canonical name for the file fixes this.


eregon:
Thanks!

PR #3296, opened 2015-08-28
[Truffle] Add tools for measuring startup time and memory consumption.

chrisseaton:
@pitr-ch @eregon @nirvdrum please review

### DIFF
--- a/core/src/main/java/org/jruby/Main.java
+++ b/core/src/main/java/org/jruby/Main.java
@@ -545,14 +545,16 @@ public class Main {
     }
 
     private static void printTruffleMemoryMetric() {
-        System.gc();
+        if (Options.TRUFFLE_METRICS_MEMORY_USED_ON_EXIT.load()) {
+            System.gc();
 
-        try {
-            Thread.sleep(1000);
-        } catch (InterruptedException e) {
+            try {
+                Thread.sleep(1000);
+            } catch (InterruptedException e) {
+            }
+
+            System.err.printf("allocated %d%n", ManagementFactory.getMemoryMXBean().getHeapMemoryUsage().getUsed());
         }
-
-        System.err.printf("allocated %d%n", ManagementFactory.getMemoryMXBean().getHeapMemoryUsage().getUsed());
     }
 
     private final RubyInstanceConfig config;
--- a/core/src/main/java/org/jruby/Main.java
+++ b/core/src/main/java/org/jruby/Main.java
@@ -546,7 +546,9 @@ public class Main {
 
     private static void printTruffleMemoryMetric() {
         if (Options.TRUFFLE_METRICS_MEMORY_USED_ON_EXIT.load()) {
-            System.gc();
+            for (int n = 0; n < 10; n++) {
+                System.gc();
+            }
 
             try {
                 Thread.sleep(1000);
--- a/core/src/main/java/org/jruby/Main.java
+++ b/core/src/main/java/org/jruby/Main.java
@@ -222,6 +222,7 @@ public class Main {
         }
 
         printTruffleTimeMetric("after-main");
+        printTruffleMemoryMetric();
     }
 
     public Status run(String[] args) {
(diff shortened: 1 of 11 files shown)


test/truffle/memory/minimum-heap.rb
```diff
@@ -0,0 +1,40 @@
...
+# Not run it fast - just run it at all.
+
+# For example:
+# $ ruby test/truffle/memory/minimum-heap.rb bin/jruby "-X+T -e 'puts 14'"
+
+tolerance = 5
+max_iterations = 100
+
+lower = 0
+upper = 4 * 1024
```
nirvdrum: Why 4 * 1024?

core/src/main/java/org/jruby/Ruby.java
```diff
@@ -549,7 +548,9 @@ public void runFromMain(InputStream inputStream, String filename) {
             return;
         }
 
+        Main.printTruffleTimeMetric("before-parse");
         ParseResult parseResult = parseFromMain(filename, inputStream);
+        Main.printTruffleTimeMetric("after-parse");
```
eregon: This is only to parse the main file, right? Maybe the name should reflect it.

eregon:
Looks good.
My initial thought was to do sth like printTime("next step name") + one printTime("END") at the end so there is no need for before/after pair and it's easier to make sure everything is counted overall. But this does not support nesting so this PR approach seems better.

test/truffle/memory/minimum-heap.rb
chrisseaton: It was just an initial size that was bound to be big enough. But now I think about it I can also search for the upper bound and not have one hard coded. I'll do that.

core/src/main/java/org/jruby/Ruby.java
chrisseaton: Will do.

chrisseaton:
@smarr please review if you have the time (especially https://github.com/jruby/jruby/pull/3296/files#diff-1c0dfa12324db9c68111f81e23be1ee6R1)

core/src/main/java/org/jruby/Main.java
```diff
@@ -531,6 +537,26 @@ protected static int handleRaiseException(RaiseException rj) {
...
+    public static void printTruffleTimeMetric(String id) {
+        if (Options.TRUFFLE_METRICS_TIME.load()) {
+            final long millis = System.currentTimeMillis();
+            System.err.printf("%s %d.%03d%n", id, millis / 1000, millis % 1000);
+        }
+    }
+
+    private static void printTruffleMemoryMetric() {
+        if (Options.TRUFFLE_METRICS_MEMORY_USED_ON_EXIT.load()) {
+            System.gc();
```
smarr: This seems to be problematic for multiple reasons.

Haven't yet read further in the changes, but, you are in the main method, and do a GC. So, I would expect, if there is no static state/memory leaks, after this GC, the used memory should be 0.
So, you might want to get the heap usage before the GC.

Also, System.gc() is really just a hint on HotSpot, you have to call it 3-4 times in succession to motivate HotSpot to actually do a collect (but, this then usually works).

core/src/main/java/org/jruby/Main.java
```diff
@@ -531,6 +537,26 @@ protected static int handleRaiseException(RaiseException rj) {
...
+
+            try {
+                Thread.sleep(1000);
```
smarr: Why do you sleep here? to give the GC a chance?

core/src/main/java/org/jruby/Main.java
smarr: Ok, I now saw that you parse verbose:gc, nice. Ok, but you probably still want to call System.gc() multiple times to make sure it triggers. Then, I would assume that the verbose:gc output gives you a sum of all the objects that made it into GCed memory.

Looks good to me!

chrisseaton: JRuby has a great deal of static state (even though it is actually re-entrant) so I still need to check heap usage right at the end. Now I think about it I'm not sure what the point is of the GC at the end, given that I need to check heap usage after it anyway.

pitr-ch: side note: there may be a reliable way how to force gc in JVMTI, http://docs.oracle.com/javase/7/docs/platform/jvmti/jvmti.html#ForceGarbageCollection. I did not use it though.

core/src/main/java/org/jruby/Main.java
chrisseaton: Without this GC I was actually getting collections after main returned, so I had no opportunity to get the final remaining heap size.

PR #3297, opened 2015-08-30
[Truffle] Shim to make it easier to use bundler.

chrisseaton:
@nirvdrum had a script to run bundler with JRuby and then patch up the files it produced to let us use that. This shim simplifies that. For example you can now do:

```
$ bundler install --standalone --path vendor
$ bin/jruby -X+T -r ./vendor/bundler/setup.rb app.rb
```

This is something that we could start suggesting people try out on their own apps at some point soon.

I was also thinking about taking this a couple of steps further. Can we autodetect and run that `setup.rb`? Can we provide a command bin/truffle-bundle that then lets you do just:

```
$ bin/truffle-bundle
$ bin/jruby -X+T app.rb
```

@nirvdrum @pitr-ch @eregon please comment

### DIFF
--- a/truffle/src/main/java/org/jruby/truffle/translator/BodyTranslator.java
+++ b/truffle/src/main/java/org/jruby/truffle/translator/BodyTranslator.java
@@ -979,14 +979,26 @@ public class BodyTranslator extends Translator {
         // Unqualified constant access, as in CONST
         final SourceSection sourceSection = translate(node.getPosition());
 
+        String name = ConstantReplacer.replacementName(sourceSection, node.getName());
+
+        /*
+         * We'd like to be able to use JRuby to run bundler to standalone-install gems required to run
+         * programs using Truffle. Bundler puts the JRuby engine name into the path, so to get around
+         * that we detect it asking for the engine, and pretend to be just JRuby.
+         */
+
+        if (name.equals("RUBY_ENGINE")
+                && sourceSection.getSource().getPath().endsWith("bundler/setup.rb")
+                && sourceSection.getStartLine() == 3) {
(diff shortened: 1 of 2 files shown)


truffle/src/main/java/org/jruby/truffle/translator/BodyTranslator.java
```diff
@@ -979,6 +979,20 @@ public RubyNode visitConstNode(org.jruby.ast.ConstNode node) {
...
+
+        /*
+         * We'd like to be able to use JRuby to run bundler to standalone-install gems required to run
+         * programs using Truffle. Bundler puts the JRuby engine name into the path, so to get around
+         * that we detect it asking for the engine, and pretend to be just JRuby.
+         */
+
+        if (name.equals("RUBY_ENGINE")
+                && sourceSection.getSource().getPath().endsWith("bundler/setup.rb")
+                && sourceSection.getStartLine() == 3) {
```
eregon: This seems fragile, I would remove the line check.

eregon:
Auto-loading of vendor/bundler/setup.rb seems bad, since many commands expect to run without the Bundler setup if it is not specified.
Would `bundler exec bin/jruby -X+T ...` work instead? It should set up the right environment to load the bundle.

pitr-ch:
I would suggest to wait with this PR until I submit mine PR with the gem. Both PR are trying to solve similar thing. Let's discuss then what'll be the best approach.

chrisseaton:
Yeah I forgot you were working on something similar. Better solutions welcome.

chrisseaton:
Closing in favour of #3298.

PR #3298, opened 2015-08-31
[Truffle] add app/gem runner

pitr-ch:
How to try with `activesupport`:
- install the gem
- `truffle --help` to see the help
- `truffle --verbose setup` (uses default bundled activesupport configuration) and sets the environment
- `truffle -v run test/core_ext/integer_ext_test.rb`

We can add config files for libraries we test, plus we get the same environment locally when testing. 

The bad path of the bundle issue is solved by symlink, https://github.com/jruby/jruby/pull/3298/files#diff-ebd53ce24fc30c47574135956e30702cR229.

### DIFF
--- a/tool/truffle/jruby_truffle_runner/bin/truffle
+++ b/tool/truffle/jruby_truffle_runner/bin/truffle
@@ -1,4 +1,0 @@
-#!/usr/bin/env ruby
-
-require_relative '../lib/jruby_truffle_runner.rb'
-JRubyTruffleRunner.new
--- a/tool/truffle/jruby_truffle_runner/bin/truffle
+++ b/tool/truffle/jruby_truffle_runner/bin/truffle
@@ -1,0 +1,4 @@
+#!/usr/bin/env ruby
+
+require_relative '../lib/jruby_truffle_runner.rb'
+JRubyTruffleRunner.new
--- a/tool/truffle/jruby_truffle_runner/gem_configurations/activesupport.yaml
+++ b/tool/truffle/jruby_truffle_runner/gem_configurations/activesupport.yaml
@@ -1,0 +1,31 @@
+---
+:global:
+  # default ../jruby/bin/ruby won't work since activesupport is one more dir deeper
+  :jruby_truffle_path: '../../jruby/bin/jruby'
+:setup:
+  :after:
+    # patches from truffle_bundle_path has to be loaded first
+    - 'echo "$:.unshift \"\#{path}/..\"" >> "#{options[:global][:truffle_bundle_path]}/bundler/setup.rb"'
+  :file:
+    shims.rb: |
+              require 'minitest'
+              # mock diff, it calls not fully supported Kernel#system
+              module Minitest
+                module Assertions
+                  def self.diff
+                    nil
+                  end
+                end
+              end
+              # mock load_plugins as it loads rubygems
+              def Minitest.load_plugins
+              end
+    bundler.rb: "module Bundler; def self.setup; end; end"
+    # mock method_source gem
+    method_source.rb: nil
+  # do not let bundler to install db gem group
+  :without:
+    - db
+:run:
+  :require:
+    - shims
--- a/tool/truffle/jruby_truffle_runner/jruby_truffle_runner.gemspec
+++ b/tool/truffle/jruby_truffle_runner/jruby_truffle_runner.gemspec
@@ -1,0 +1,17 @@
+Gem::Specification.new do |s|
+  s.name        = 'jruby_truffle_runner'
+  s.version     = '0.0.1'
+  s.platform    = Gem::Platform::RUBY
+  s.authors     = ['Petr Chalupa']
+  s.email       = ['git@pitr.ch']
+  s.homepage    = 'https://github.com/jruby/jruby'
+  s.summary     = 'Temporary JRuby+Truffle runner'
+  s.description = 'Until JRuby+Truffle is more complete, allows to run apps/gems simply on JRuby+Truffle.'
+
+  s.required_rubygems_version = '>= 1.3.6'
+
+  s.files        = Dir['{lib}/**/*.rb', 'bin/*', 'gem_configurations/*.yaml', 'LICENSE', '*.md']
+  s.require_path = 'lib'
+
+  s.executables = ['truffle']
+end
--- a/tool/truffle/jruby_truffle_runner/lib/jruby_truffle_runner.rb
+++ b/tool/truffle/jruby_truffle_runner/lib/jruby_truffle_runner.rb
@@ -1,0 +1,285 @@
+# Copyright (c) 2015 Oracle and/or its affiliates. All rights reserved. This
+# code is released under a tri EPL/GPL/LGPL license. You can use it,
+# redistribute it and/or modify it under the terms of the:
+#
+# Eclipse Public License version 1.0
+# GNU General Public License version 2
+# GNU Lesser General Public License version 2.1
+
+require 'optparse'
+require 'pp'
+require 'yaml'
+require 'fileutils'
+
+# TODO (pitr 01-Sep-2015): add pre stored run options combinations like -S irb, -I test
+
+class JRubyTruffleRunner
+  attr_reader :options
+
+  begin
+    assign_new_value   = -> (new, old) { new }
+    add_to_array       = -> (new, old) { old << new }
+    merge_hash         = -> ((k, v), old) { old.merge k => v }
+
+    # Format:
+    #   subcommand_name: {
+    #     :'option_name (also used as a key in yaml)' => [option_parser_on_method_args,
+    #                                                     -> (new_value, old_value) { result_of_this_block_is_stored },
+    #                                                     default_value]
+    #   }
+    OPTION_DEFINITIONS = {
+        global: {
+            verbose:             ['-v', '--verbose', 'Run verbosely (prints options)', assign_new_value, false],
+            help:                ['-h', '--help', 'Show this message', assign_new_value, false],
+            debug_port:          ['--debug-port PORT', 'Debug port', assign_new_value, '51819'],
+            debug_option:        ['--debug-option OPTION', 'Debug JVM option', assign_new_value,
+                                  '-J-agentlib:jdwp=transport=dt_socket,server=y,address=%d,suspend=y'],
+            truffle_bundle_path: ['--truffle-bundle-path NAME', 'Bundle path', assign_new_value, 'truffle_bundle'],
+            jruby_truffle_path:  ['--jruby-truffle-path PATH', 'Path to JRuby+Truffle bin/ruby', assign_new_value,
+                                  '../jruby/bin/jruby']
+        },
+        setup:  {
+            help:    ['-h', '--help', 'Show this message', assign_new_value, false],
+            after:   ['--after SH_CMD', 'Commands to execute after setup', add_to_array, []],
+            file:    ['--file NAME,CONTENT', Array, 'Create file in truffle_bundle_path', merge_hash, {}],
+            without: ['--without GROUP', 'Do not install listed gem group by bundler', add_to_array, []]
+        },
+        run:    {
+            help:       ['-h', '--help', 'Show this message', assign_new_value, false],
+            test:       ['-t', '--test', 'Do not use Truffle use plain JRuby', assign_new_value, false],
+            debug:      ['-d', '--debug', 'JVM remote debugging', assign_new_value, false],
+            require:    ['-r', '--require FILE', 'Files to require, same as Ruby\'s -r', add_to_array, []],
+            load_path:  ['-I', '--load-path LOAD_PATH', 'Paths to add to load path, same as Ruby\'s -I', add_to_array, []],
+            jexception: ['--jexception', 'print Java exceptions', assign_new_value, false]
+        },
+        clean:  {
+            help: ['-h', '--help', 'Show this message', assign_new_value, false]
+        }
+    }
+  end
+
+  begin
+    global_help = <<-TXT.gsub(/^ {6}/, '')
+      Usage: truffle [options] [subcommand [subcommand-options]]
+      Subcommands are: #{(OPTION_DEFINITIONS.keys - [:global]).map(&:to_s).join(', ')}
+
+      Allows to execute gem/app on JRuby+Truffle until it's more complete. Environment
+      has to be set up first with setup subcommand then run subcommand can be used.
+
+      Options can be set on commandline or in local directory in .truffle.yaml file.
+      Tha data in yaml file follow same structure as OPTION_DEFINITIONS what, its values
+      are deep-merged with default values, then command-line options are applied.
+
+      This tool contains default configurations for gems in gem_configurations directory.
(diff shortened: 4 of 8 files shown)


chrisseaton:
This looks like really good work and should be a great step forward in testing gems.

I can't figure out exactly how I'm supposed to use it though. How do I install the gem? Using my dev version of JRuby? How would it get onto my path?

I tried a couple of commands like this (without installing) and it didn't work:

```
chrisseaton@Chriss-MBP:~/Documents/ruby/psd.rb [master] $ ../jruby/bin/jruby ../jruby/tool/truffle/jruby_truffle_runner/bin/truffle --help
NameError: undefined local variable or method `help' for JRubyTruffleRunner:Class
                 method_missing at org/jruby/RubyBasicObject.java:1592
    block in JRubyTruffleRunner at /Users/chrisseaton/Documents/ruby/jruby/tool/truffle/jruby_truffle_runner/lib/jruby_truffle_runner.rb:25
                           call at org/jruby/RubyProc.java:308
  block in build_option_parsers at /Users/chrisseaton/Documents/ruby/jruby/tool/truffle/jruby_truffle_runner/lib/jruby_truffle_runner.rb:139
                           call at org/jruby/RubyProc.java:308
        block in parse_in_order at /Users/chrisseaton/Documents/ruby/jruby/lib/ruby/stdlib/optparse.rb:1396
                          catch at org/jruby/RubyKernel.java:1098
                 parse_in_order at /Users/chrisseaton/Documents/ruby/jruby/lib/ruby/stdlib/optparse.rb:1383
                         order! at /Users/chrisseaton/Documents/ruby/jruby/lib/ruby/stdlib/optparse.rb:1377
                          order at /Users/chrisseaton/Documents/ruby/jruby/lib/ruby/stdlib/optparse.rb:1369
                     initialize at /Users/chrisseaton/Documents/ruby/jruby/tool/truffle/jruby_truffle_runner/lib/jruby_truffle_runner.rb:107
                          <top> at ../jruby/tool/truffle/jruby_truffle_runner/bin/truffle:4
chrisseaton@Chriss-MBP:~/Documents/ruby/psd.rb [master] $ ../jruby/bin/jruby ../jruby/tool/truffle/jruby_truffle_runner/bin/truffle setup
/Users/chrisseaton/.rbenv/shims/bundle
No value provided for option '--without'
RuntimeError: command failed
       execute_cmd at /Users/chrisseaton/Documents/ruby/jruby/tool/truffle/jruby_truffle_runner/lib/jruby_truffle_runner.rb:268
  subcommand_setup at /Users/chrisseaton/Documents/ruby/jruby/tool/truffle/jruby_truffle_runner/lib/jruby_truffle_runner.rb:227
        initialize at /Users/chrisseaton/Documents/ruby/jruby/tool/truffle/jruby_truffle_runner/lib/jruby_truffle_runner.rb:119
             <top> at ../jruby/tool/truffle/jruby_truffle_runner/bin/truffle:4
```

Could you post a complete example of use for Active Support?

tool/truffle/jruby_truffle_runner/lib/jruby_truffle_runner.rb
```diff
@@ -0,0 +1,270 @@
+require 'optparse'
```
chrisseaton: Anything more substantial than a simple configuration file must begin with our standard copyright header - so the JRuby people are clear what we're putting into their repository if nothing else.

tool/truffle/jruby_truffle_runner/gem_configurations/activesupport.yaml
```diff
@@ -0,0 +1,24 @@
...
+    shims.rb: |
+              module Minitest
+                module Assertions
+                  def self.diff
+                    nil
+                  end
+                end
+              end
+    bundler.rb: "module Bundler; def self.setup; end; end"
+    method_source.rb: nil
```
chrisseaton: Does this create an empty file? How does the empty `method_source.rb` get used?

tool/truffle/jruby_truffle_runner/gem_configurations/activesupport.yaml
```diff
@@ -0,0 +1,24 @@
...
+  :without:
+    - db
+:run:
+  :require:
+    - shims
+  :load_path:
+    - test
```
chrisseaton: These are all a bit cryptic to me - can you explain why they're needed for Active Support?

tool/truffle/jruby_truffle_runner/jruby_truffle_runner.gemspec
```diff
@@ -0,0 +1,17 @@
+Gem::Specification.new do |s|
+  s.name        = 'jruby_truffle_runner'
+  s.version     = '0.0.1'
+  s.platform    = Gem::Platform::RUBY
+  s.authors     = ['Petr Chalupa']
+  s.email       = ['git@pitr.ch']
+  s.homepage    = 'http://github.com/jruby/jruby'
```
nirvdrum: May as well make this 'https' and avoid the redirect.

tool/truffle/jruby_truffle_runner/lib/jruby_truffle_runner.rb
```diff
@@ -0,0 +1,270 @@
...
+require 'pp'
+require 'yaml'
+require 'fileutils'
+
+# TODO stored after run combinations like -S irb, -I test
```
nirvdrum: We've been annotating TODOs with initials or GitHub username & date to better track TODO ownership.

tool/truffle/jruby_truffle_runner/lib/jruby_truffle_runner.rb
```diff
@@ -0,0 +1,270 @@
...
+
+class JRubyTruffleRunner
+  attr_reader :options
+
+  begin
+    assign_new_value   = -> (new, old) { new }
+    add_to_array       = -> (new, old) { old << new }
```
nirvdrum: I guess the GitHub file coloring can't work out the context here.  I thought that was mildly interesting.

pitr-ch:
Ah sorry, last minute changes strike again :(. `--help` option is broken, if you run just `truffle` without the `-h` option it will print help, it goes through different path.

tool/truffle/jruby_truffle_runner/lib/jruby_truffle_runner.rb
pitr-ch: fixed

tool/truffle/jruby_truffle_runner/gem_configurations/activesupport.yaml
pitr-ch: The nil is interpreted as a String which is intended. I needed to replace `method_source` gem with empty file. We have some bug though when parsing empty file hence the nil. I'll add comments to the yaml file.

tool/truffle/jruby_truffle_runner/gem_configurations/activesupport.yaml
pitr-ch: All the options here map directly to command-line options (but the help was broken). The test is required on LOAD_PATH when running test files, it's not required for active-support in general, I'll remove it.

tool/truffle/jruby_truffle_runner/lib/jruby_truffle_runner.rb
pitr-ch: yeah `-> ((k, v), old) { old.merge k => v }` doesn't work in Rubymine too.

pitr-ch:
To run the executable, I either do ` ruby ../../jruby/tool/truffle/jruby_truffle_runner/bin/truffle --help` or I install the gem:
- cd `tool/truffle/jruby_truffle_runner`
- `gem build jruby_truffle_runner.gemspec`
- `gem install jruby_truffle_runner-0.0.1.gem`
- cd `../../../../rails/activesupport`
- then `truffle --help` should work

pitr-ch:
Regarding usage in activesupport:
- clone rails
- `cd rails/activesupport`
- install jruby_truffle_runner on current Ruby implementation
- run `truffle setup`
- run e.g. `truffle run -- -I test test/core_ext/integer_ext_test.rb` (assumes jruby on path `../../jruby/bin/ruby`, can be changed in .truffle.yaml after setup step)

tool/truffle/jruby_truffle_runner/lib/jruby_truffle_runner.rb
```diff
@@ -0,0 +1,285 @@
...
+            after:   ['--after SH_CMD', 'Commands to execute after setup', add_to_array, []],
+            file:    ['--file NAME,CONTENT', Array, 'Create file in truffle_bundle_path', merge_hash, {}],
+            without: ['--without GROUP', 'Do not install listed gem group by bundler', add_to_array, []]
+        },
+        run:    {
+            help:       ['-h', '--help', 'Show this message', assign_new_value, false],
+            test:       ['-t', '--test', 'Do not use Truffle use plain JRuby', assign_new_value, false],
+            debug:      ['-d', '--debug', 'JVM remote debugging', assign_new_value, false],
+            require:    ['-r', '--require FILE', 'Files to require, same as Ruby\'s -r', add_to_array, []],
+            load_path:  ['-l', '--load-path LOAD_PATH', 'Paths to add to load path, same as Ruby\'s -I', add_to_array, []],
```
eregon: can the shorthand be -I as well?

tool/truffle/jruby_truffle_runner/lib/jruby_truffle_runner.rb
```diff
@@ -0,0 +1,285 @@
...
+      Subcommands are: #{(OPTION_DEFINITIONS.keys - [:global]).map(&:to_s).join(', ')}
+
+      Allows to execute gem/app on JRuby+Truffle until it's more complete. Environment
+      has to be set up first with setup subcommand then run subcommand can be used.
+
+      Options can be set on commandline or in local directory in .truffle.yaml file.
+      Tha data in yaml file follow same structure as OPTION_DEFINITIONS what, its values
+      are deep-merged with default values, then command-line options are applied.
+
+      Gem contains default configurations for gems gem_configurations directory.
```
eregon: what is `Gem`?

tool/truffle/jruby_truffle_runner/lib/jruby_truffle_runner.rb
```diff
@@ -0,0 +1,285 @@
...
+      end
+    end
+
+    eval_yaml_strings b
+  end
+
+  def eval_yaml_strings(value)
+    if String === value
+      begin
+        eval('"' + value.gsub(/\\#|"|\\/, '\#' => '\#', '"' => '\"', '\\' => '\\\\') + '"')
```
eregon: `'\#'` is translated to itself?

eregon:
Looks great, and it also raises the question what launcher do we want for testing/user/dev and I think it make sense to have sth slightly different for some of them.

tool/truffle/jruby_truffle_runner/bin/truffle
```diff
@@ -0,0 +1,4 @@
+#!/usr/bin/env ruby
+
+require_relative '../lib/jruby_truffle_runner.rb'
+JRubyTruffleRunner.new
```
eregon: This is not executable by default, can you chmod and commit that?

eregon: It should be renamed I think to match with the gem name.

tool/truffle/jruby_truffle_runner/gem_configurations/activesupport.yaml
```diff
@@ -0,0 +1,27 @@
+---
+:global:
+  # default ../jruby/bin/ruby won't work since activesupport is one more dir deeper
+  :jruby_truffle_path: '../../jruby/bin/ruby'
```
eregon: bin/**j**ruby

tool/truffle/jruby_truffle_runner/lib/jruby_truffle_runner.rb
pitr-ch: Yeah probably better.

tool/truffle/jruby_truffle_runner/lib/jruby_truffle_runner.rb
pitr-ch: yeah, it prevents '\\' to be applied to the `\` in `\#`, not very transparent I admit.

tool/truffle/jruby_truffle_runner/bin/truffle
pitr-ch: done;

I am not sure, `jruby_truffle_runner` executable is quite long :/

tool/truffle/jruby_truffle_runner/lib/jruby_truffle_runner.rb
eregon: For what do we need this? It seems the code in YAML should probably escapes itself if it needs to.

pitr-ch: To allow using other options in YAML. This is an example: https://github.com/pitr-ch/jruby/blob/master/tool/truffle/jruby_truffle_runner/gem_configurations/activesupport.yaml#L8-8. It'll insert a value from options to the command, `truffle_bundle_path` in this case.

pitr-ch:
- [ ] rename `truffle` executable to `jruby-truffle`

tool/truffle/jruby_truffle_runner/lib/jruby_truffle_runner.rb
eregon: How does it work with the "echo"?
For that particular case, could we just do it by adding a `-I #{options[:global][:truffle_bundle_path]}/bundler"`? This might be a reasonable thing to do by default.

pitr-ch: It has to be first directory in `$:`, if t's just added with `-I` then it ends up at the end of `$:`. But it could be default that the `bundler/setup.rb` adds the path always without extra configuration in yaml. Downside is that it won't be explicitly visible in the config file. Hm I'll add options to prepend load-path dirs to avoid the ugly echo but to keep the information in the config file.

chrisseaton:
Can we merge this and then keep working on master? It's already very useful.

pitr-ch:
Please check the last commit if you agree with the new names. After that we can merge.

chrisseaton:
Looks good to me.

tool/truffle/jruby_truffle_runner/lib/jruby_truffle_runner.rb
eregon:
```bash
$ ruby -I ahah -e 'puts $:.first'
.../ahah
```
So -I prepends the path normally :wink:

pitr-ch: Sorry I wrote it poorly. It ends up at the end of load paths we add which is wrong. All gem paths are added in `-r bundler/setup.rb` which is called after `-I` options. So it's prepended as you say but still it ends up after the gem paths prepended when bundler/setup.rb is required.

PR #3299, opened 2015-08-31
ObjectSpace specs

chrisseaton:
@eregon please review

### DIFF
--- a/spec/ruby/core/objectspace/each_object_spec.rb
+++ b/spec/ruby/core/objectspace/each_object_spec.rb
@@ -65,19 +65,28 @@ describe "ObjectSpace.each_object" do
     ObjectSpaceFixtures::to_be_found_symbols.should include(:local)
   end
 
-  it "finds an object stored in a local variable captured in a block" do
+  it "finds an object stored in a local variable captured in a block explicitly" do
     proc = Proc.new {
-      local_in_block = ObjectSpaceFixtures::ObjectToBeFound.new(:local_in_block)
+      local_in_block = ObjectSpaceFixtures::ObjectToBeFound.new(:local_in_block_explicit)
+      Proc.new { local_in_block }
+    }.call
+
+    ObjectSpaceFixtures::to_be_found_symbols.should include(:local_in_block_explicit)
+  end
+
+  it "finds an object stored in a local variable captured in a block implicitly" do
+    proc = Proc.new {
+      local_in_block = ObjectSpaceFixtures::ObjectToBeFound.new(:local_in_block_implicit)
       Proc.new { }
     }.call
 
-    ObjectSpaceFixtures::to_be_found_symbols.should include(:local_in_block)
+    ObjectSpaceFixtures::to_be_found_symbols.should include(:local_in_block_implicit)
   end
 
   it "finds an object stored in a local variable captured in a Proc#binding" do
     binding = Proc.new {
       local_in_proc_binding = ObjectSpaceFixtures::ObjectToBeFound.new(:local_in_proc_binding)
-      Proc.new{ }.binding
+      Proc.new { }.binding
     }.call
 
     ObjectSpaceFixtures::to_be_found_symbols.should include(:local_in_proc_binding)
--- a/spec/ruby/core/objectspace/each_object_spec.rb
+++ b/spec/ruby/core/objectspace/each_object_spec.rb
@@ -1,8 +1,6 @@
 require File.expand_path('../../../spec_helper', __FILE__)
 require File.expand_path('../fixtures', __FILE__)
 
-require 'weakref'
-
 describe "ObjectSpace.each_object" do
   it "calls the block once for each living, non-immediate object in the Ruby process" do
     class ObjectSpaceSpecEachObject; end
@@ -49,20 +47,20 @@ describe "ObjectSpace.each_object" do
 
   it "finds an object stored in a global variable" do
     $object_space_global_variable = ObjectSpaceFixtures::ObjectToBeFound.new(:global)
-    ObjectSpaceFixtures::to_be_found_symbols.should include(:global)
+    ObjectSpaceFixtures.to_be_found_symbols.should include(:global)
   end
 
   it "finds an object stored in a top-level constant" do
-    ObjectSpaceFixtures::to_be_found_symbols.should include(:top_level_constant)
+    ObjectSpaceFixtures.to_be_found_symbols.should include(:top_level_constant)
   end
 
   it "finds an object stored in a second-level constant" do
-    ObjectSpaceFixtures::to_be_found_symbols.should include(:second_level_constant)
+    ObjectSpaceFixtures.to_be_found_symbols.should include(:second_level_constant)
   end
 
   it "finds an object stored in a local variable" do
     local = ObjectSpaceFixtures::ObjectToBeFound.new(:local)
-    ObjectSpaceFixtures::to_be_found_symbols.should include(:local)
+    ObjectSpaceFixtures.to_be_found_symbols.should include(:local)
   end
 
   it "finds an object stored in a local variable captured in a block explicitly" do
@@ -71,7 +69,7 @@ describe "ObjectSpace.each_object" do
       Proc.new { local_in_block }
     }.call
 
-    ObjectSpaceFixtures::to_be_found_symbols.should include(:local_in_block_explicit)
+    ObjectSpaceFixtures.to_be_found_symbols.should include(:local_in_block_explicit)
   end
 
   it "finds an object stored in a local variable captured in a block implicitly" do
@@ -80,7 +78,7 @@ describe "ObjectSpace.each_object" do
       Proc.new { }
     }.call
 
-    ObjectSpaceFixtures::to_be_found_symbols.should include(:local_in_block_implicit)
+    ObjectSpaceFixtures.to_be_found_symbols.should include(:local_in_block_implicit)
   end
 
   it "finds an object stored in a local variable captured in a Proc#binding" do
@@ -89,7 +87,7 @@ describe "ObjectSpace.each_object" do
       Proc.new { }.binding
     }.call
 
-    ObjectSpaceFixtures::to_be_found_symbols.should include(:local_in_proc_binding)
+    ObjectSpaceFixtures.to_be_found_symbols.should include(:local_in_proc_binding)
   end
 
   it "finds an object stored in a local variable captured in a Kernel#binding" do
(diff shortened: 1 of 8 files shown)


truffle/src/main/java/org/jruby/truffle/runtime/object/ObjectGraph.java
```diff
@@ -115,9 +129,23 @@ private void visitObject(DynamicObject object, ObjectGraphVisitor visitor) throw
 
             // Visit all properties
 
-            for (Property property : object.getShape().getProperties()) {
+            for (Property property : object.getShape().getPropertyListInternal(false)) {
```
eregon: Do we still need this change with the one below?

truffle/src/main/ruby/core/thread.rb
```diff
@@ -16,6 +16,13 @@ def []=(symbol, value)
     __thread_local_variables[symbol] = value
   end
 
+  def thread_variable?(symbol)
+    __thread_local_variables.has_key? symbol
+  end
+
+  alias_method :thread_variable_get, :[]
+  alias_method :thread_variable_set, :[]=
```
eregon: `Thread#[]` is *fiber-local* get!

truffle/src/main/java/org/jruby/truffle/runtime/object/ObjectGraph.java
```diff
@@ -133,8 +161,11 @@ private void visitObject(Object object, ObjectGraphVisitor visitor) throws StopV
             visitObject(entry.getKey(), visitor);
             visitObject(entry.getValue(), visitor);
             visitObject(entry.getNextInLookup(), visitor);
-        } else if (object instanceof MaterializedFrame) {
-            visitFrame((MaterializedFrame) object, visitor);
+        } else if (object instanceof Frame) {
+            visitFrame((Frame) object, visitor);
+        } else if (object instanceof InternalMethod) {
+            final InternalMethod method = (InternalMethod) object;
+            visitObject(method.getDeclarationFrame(), visitor);
```
eregon: This is only needed in the case of a method defined by `define_method` with a proc, right?

spec/ruby/core/objectspace/fixtures.rb
```diff
@@ -32,4 +32,29 @@ def self.scoped(wr)
     return Proc.new { wr.write "finalized"; wr.close }
   end
 
+  class ObjectToBeFound
+    attr_reader :id
+
+    def initialize(id)
+      @id = id
```
eregon: Could this be renamed to `@name` instead?
`id` looks like `__id__`/`object_id`.

spec/ruby/core/objectspace/each_object_spec.rb
```diff
@@ -43,4 +46,127 @@ class ObjectSpaceSpecEachOtherObject; end
     # this is needed to prevent the new_obj from being GC'd too early
     new_obj.should_not == nil
   end
+
+  it "finds an object stored in a global variable" do
+    $object_space_global_variable = ObjectSpaceFixtures::ObjectToBeFound.new(:global)
+    ObjectSpaceFixtures::to_be_found_symbols.should include(:global)
```
eregon: Syntax: use `.` for "class" methods.

spec/ruby/core/objectspace/each_object_spec.rb
```diff
@@ -43,4 +46,127 @@ class ObjectSpaceSpecEachOtherObject; end
...
+    ObjectSpaceFixtures::to_be_found_symbols.should include(:hash_value)
+  end
+
+  it "finds an object stored in an instance variable" do
+    local = ObjectSpaceFixtures::ObjectWithInstanceVariable.new
+    ObjectSpaceFixtures::to_be_found_symbols.should include(:instance_variable)
+  end
+
+  it "doesn't find an object stored in a WeakRef that should have been cleared" do
+    weak_ref = WeakRef.new(ObjectSpaceFixtures::ObjectToBeFound.new(:weakref))
```
eregon: `GC.start` provides no guarantees, especially for conservative collectors. Does this pass in MRI?

spec/ruby/core/objectspace/each_object_spec.rb
```diff
@@ -43,4 +46,127 @@ class ObjectSpaceSpecEachOtherObject; end
...
+    }.call
+
+    ObjectSpaceFixtures::to_be_found_symbols.should include(:at_exit)
+  end
+
+  it "finds an object captured in finalizer" do
+    alive = Object.new
+
+    Proc.new {
+      local = ObjectSpaceFixtures::ObjectToBeFound.new(:at_exit)
```
eregon: :finalizer

spec/ruby/core/objectspace/each_object_spec.rb
```diff
@@ -43,4 +46,127 @@ class ObjectSpaceSpecEachOtherObject; end
...
+
+      ObjectSpace.define_finalizer(alive, Proc.new {
+        local
+      })
+    }.call
+
+    ObjectSpaceFixtures::to_be_found_symbols.should include(:at_exit)
+
+    alive.should_not be_nil
```
eregon: This should not be needed, upper we rely on the current locals to not go away.

truffle/src/main/java/org/jruby/truffle/runtime/object/ObjectGraph.java
```diff
@@ -115,9 +129,23 @@ private void visitObject(DynamicObject object, ObjectGraphVisitor visitor) throw
...
 
-            for (Property property : object.getShape().getProperties()) {
+            for (Property property : object.getShape().getPropertyListInternal(false)) {
                 visitObject(property.get(object, object.getShape()), visitor);
             }
+
+            // Visit specific objects that we're managing
+
+            if (RubyGuards.isRubyModule(object)) {
+                final ModuleFields module = Layouts.MODULE.getFields(object);
```
eregon: This could be extracted to `visitModule()`.

spec/ruby/core/objectspace/each_object_spec.rb
```diff
@@ -43,4 +46,127 @@ class ObjectSpaceSpecEachOtherObject; end
...
+    hash = {a: ObjectSpaceFixtures::ObjectToBeFound.new(:hash_value)}
+    ObjectSpaceFixtures::to_be_found_symbols.should include(:hash_value)
+  end
+
+  it "finds an object stored in an instance variable" do
+    local = ObjectSpaceFixtures::ObjectWithInstanceVariable.new
+    ObjectSpaceFixtures::to_be_found_symbols.should include(:instance_variable)
+  end
+
+  it "doesn't find an object stored in a WeakRef that should have been cleared" do
```
eregon: Can you move the `require 'weakref'` here?

eregon:
Nice specs! Please address my comments.

truffle/src/main/java/org/jruby/truffle/runtime/object/ObjectGraph.java
chrisseaton: Yes - `getProperties()` doesn't return `HiddenKeys` - took me ages to figure this out.

truffle/src/main/ruby/core/thread.rb
chrisseaton: Yes I did figure out there was something wrong here - but actually I think our problem is that `Thread#[]` currently isn't fibre-local for us. It's stored in `ThreadLayout` and there's nothing similar in `FiberLayout`.

truffle/src/main/java/org/jruby/truffle/runtime/object/ObjectGraph.java
chrisseaton: Yes I think so - when I wrote this I started to wonder why we have declaration frame in both `InternalMethod` and `ProcLayout` - and `define_method` is the reason.

spec/ruby/core/objectspace/fixtures.rb
chrisseaton: Done.

spec/ruby/core/objectspace/each_object_spec.rb
chrisseaton: Done.

spec/ruby/core/objectspace/each_object_spec.rb
chrisseaton: Done.

spec/ruby/core/objectspace/each_object_spec.rb
chrisseaton: Yes - I was thinking about a hypothetical implementation that does more global optimisations than us and sees that the variable holding `alive` can be re-used.

truffle/src/main/java/org/jruby/truffle/runtime/object/ObjectGraph.java
chrisseaton: Done.

spec/ruby/core/objectspace/each_object_spec.rb
chrisseaton: I've switched to a time-limited loop that GCs until the weakref is actually cleared.

spec/ruby/core/objectspace/each_object_spec.rb
chrisseaton: Done.

truffle/src/main/ruby/core/thread.rb
eregon: Yeah, we should have a similar `fiberLocals` in FiberLayout.

truffle/src/main/java/org/jruby/truffle/runtime/object/ObjectGraph.java
eregon: Which HiddenKey do we want to traverse? Before they were fields and only traversed if they were explicitly in visitObjectGraphChildren, right?

chrisseaton: We want to visit all fields, whether they're instance variables or they're internal fields (like `store` in `Array`) to find all objects. Things like `store` are currently `HiddenKey` so that they aren't mistaken for instance variables.

truffle/src/main/java/org/jruby/truffle/runtime/object/ObjectGraph.java
eregon: We could have define_method create a wrapper method around the Proc, but then it would also be fairly easy to forgot about traversing the Proc for ObjectSpace.

truffle/src/main/java/org/jruby/truffle/runtime/object/ObjectGraph.java
eregon: Right, sounds good and less error-prone

truffle/src/main/java/org/jruby/truffle/runtime/object/ObjectGraph.java
```diff
@@ -74,13 +79,26 @@ public void run(DynamicObject thread, Node currentNode) {
                     }
 
                     try {
-                        // We only visit the global variables from the root thread
-                        visitObject(context.getCoreLibrary().getGlobalVariablesObject(), visitor);
+                        // We only visit the global variables and other global state from the root thread
+
+                        if (thread == context.getThreadManager().getRootThread()) {
```
eregon: This will run for all fibers of the main thread. One way to only run for the initial Fiber is to compare with the root fiber of the main thread.

eregon: Or more simply since any of the thread could do this, the caller of visitRoots could be chosen to lookup these ones.

PR #3301, opened 2015-09-02
[Truffle] add RubyFiletypeDetector

pitr-ch:


truffle/src/main/java/org/jruby/truffle/RubyFiletypeDetector.java
```diff
@@ -0,0 +1,46 @@
...
+import java.io.FileReader;
+import java.io.IOException;
+import java.nio.file.Files;
+import java.nio.file.Path;
+import java.nio.file.spi.FileTypeDetector;
+import java.util.Optional;
+
+public final class RubyFiletypeDetector extends FileTypeDetector {
+
+    private final static String[] KNOWN_RUBY_FILES = new String[]{ "Gemfile", "Rakefile" };
```
mkristian: since we use Mavenfile for some projects (jruby-jars, jruby-openssl, etc) this "might" be another option. but my view of the world is biased  :)

truffle/src/main/java/org/jruby/truffle/RubyFiletypeDetector.java
```diff
@@ -0,0 +1,46 @@
...
+    private final static String[] KNOWN_RUBY_SUFFIXES = new String[]{ ".rb", ".rake", ".gemspec" };
+    private final static String RUBY_MIME = "application/x-ruby";
```
chrisseaton: You can get this constant from `RubyLangauge`.

truffle/src/main/java/org/jruby/truffle/RubyFiletypeDetector.java
```diff
@@ -0,0 +1,46 @@
...
+    @Override
+    public String probeContentType(Path path) throws IOException {
+        final Path fileNamePath = path.getFileName();
+        if (fileNamePath == null) {
+            return null;
+        }
+
+        final String fileName = fileNamePath.toString();
+
+        for (String candidate : KNOWN_RUBY_SUFFIXES) {
```
chrisseaton: Do you think people ever have `.RB` files? Do we need to normalise case?

chrisseaton:
Ping @jtulach

truffle/src/main/java/org/jruby/truffle/RubyFiletypeDetector.java
pitr-ch: What do you mean by `RubyLangauge`? I could not find a class like that, neither I found constant with `"application/x-ruby"`.

chrisseaton: Ah it's probably just on `truffle-head` - ignore me.

truffle/src/main/java/org/jruby/truffle/RubyFiletypeDetector.java
pitr-ch: There could be few eccentric individuals with `.RB`, let's be friendly.

eregon:
What is the purpose?

chrisseaton:
So GraalVM can be given a file with no more information and just run it.

truffle/src/main/java/org/jruby/truffle/RubyFiletypeDetector.java
```diff
@@ -0,0 +1,46 @@
...
+            if (fileName.equals(candidate)) {
+                return RUBY_MIME;
+            }
+        }
+
+        final Optional<String> firstLine = Files.lines(path).findFirst();
+        if (firstLine.isPresent() && firstLine.get().equals("#!/usr/bin/env ruby")) {
+            return RUBY_MIME;
+        }
+
```
jtulach: Nice, but maybe too strict. What about two spaces in between "env" and "ruby". What if somebody uses just "#!/usr/bin/ruby"...

pitr-ch: done

truffle/src/main/java/org/jruby/truffle/RubyFiletypeDetector.java
pitr-ch: done

eregon:
@pitr-ch No Java 8 method ;)

truffle/src/main/java/org/jruby/truffle/RubyFiletypeDetector.java
```diff
@@ -0,0 +1,49 @@
...
+        for (String candidate : KNOWN_RUBY_FILES) {
+            if (fileName.equals(candidate)) {
+                return RUBY_MIME;
+            }
+        }
+
+        final Optional<String> maybeFirstLine = Files.lines(path).findFirst();
+        if (maybeFirstLine.isPresent()) {
+            final String firstLine = maybeFirstLine.get();
+            if (firstLine.matches("^#! ?/usr/bin/(env +ruby|ruby).*")) {
```
eregon: It's probably best to compile the pattern only once.

truffle/src/main/java/org/jruby/truffle/RubyFiletypeDetector.java
```diff
@@ -0,0 +1,50 @@
+package org.jruby.truffle;
```
pitr-ch: I’ve forgot to ask, `RubyFiletypeDetector` is currently in `org.jruby.truffle` package, not the best place. What about `org.jruby.truffle.runtime.util` or other new `org.jruby.truffle.util`?

jtulach: The FiletypeDetector isn't Truffle-only, it might be useful for whole JRuby - so consider option of putting it to generic JRuby runtime.

chrisseaton: Yes org.jruby might be right. And the metadata would then need to go in the core jar obviously.

truffle/src/main/java/org/jruby/truffle/RubyFiletypeDetector.java
```diff
@@ -0,0 +1,50 @@
...
+    public String probeContentType(Path path) throws IOException {
+        final Path fileNamePath = path.getFileName();
+        if (fileNamePath == null) {
+            return null;
+        }
+
+        final String fileName = fileNamePath.toString();
+
+        for (String candidate : KNOWN_RUBY_SUFFIXES) {
+            if (fileName.toLowerCase(Locale.ROOT).endsWith(candidate)) {
```
nirvdrum: It's probably not a huge gain because most files will be '.rb', but maybe lift the `toLowerCase` call above the `for` loop to avoid repeated calculations.

truffle/src/main/java/org/jruby/truffle/RubyFiletypeDetector.java
pitr-ch: done.

PR #3311, opened 2015-09-04
[Truffle] Complain when running with an incompatible Graal version.

eregon:
@chrisseaton What should we do?

In the particular case of 0.8, there was no warning before this, and the program would run like on a stock JDK (because truffle is not on the bootclasspath I suppose).

We should probably also make the requirement of Graal 0.7 clearer in the wiki/README.


chrisseaton:
I previously removed this check because the graal.version property was lost in the reorganisation of the repositories but it does need to go back in and this code is fine for that.

eregon:
@chrisseaton Do you know which commit removed the check (or introduced it)?

chrisseaton:
It was removed in cad8f5e4fe722c7b8197af8fe8ca91d2cd85f30b.

PR #3315, opened 2015-09-09
[Truffle] allow Kernel#system and Process.spawn to be called with arr…

pitr-ch:
…ay of arguments

only :unsetenv_others is supported for now

### DIFF
--- a/truffle/src/main/java/org/jruby/truffle/nodes/core/TrufflePrimitiveNodes.java
+++ b/truffle/src/main/java/org/jruby/truffle/nodes/core/TrufflePrimitiveNodes.java
@@ -9,11 +9,8 @@
  */
 package org.jruby.truffle.nodes.core;
 
-import com.oracle.truffle.api.CompilerDirectives;
 import com.oracle.truffle.api.CompilerDirectives.TruffleBoundary;
 import com.oracle.truffle.api.Truffle;
-import com.oracle.truffle.api.dsl.NodeChild;
-import com.oracle.truffle.api.dsl.ShortCircuit;
 import com.oracle.truffle.api.dsl.Specialization;
 import com.oracle.truffle.api.frame.FrameInstance;
 import com.oracle.truffle.api.frame.FrameInstanceVisitor;
@@ -29,9 +26,6 @@ import org.jruby.RubyGC;
 import org.jruby.RubyString;
 import org.jruby.ext.rbconfig.RbConfigLibrary;
 import org.jruby.truffle.nodes.RubyGuards;
-import org.jruby.truffle.nodes.RubyNode;
-import org.jruby.truffle.nodes.ThreadLocalObjectNode;
-import org.jruby.truffle.nodes.objects.WriteInstanceVariableNode;
 import org.jruby.truffle.runtime.RubyArguments;
 import org.jruby.truffle.runtime.RubyCallStack;
 import org.jruby.truffle.runtime.RubyContext;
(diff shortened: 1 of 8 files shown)


truffle/src/main/java/org/jruby/truffle/nodes/core/TrufflePrimitiveNodes.java
```diff
@@ -565,4 +575,93 @@ public DynamicObject printInterleavedBacktrace() {
...
+        public SetGlobalStatusVariable(RubyContext context, SourceSection sourceSection) {
+            super(context, sourceSection);
+        }
+
+        abstract RubyNode getValue();
+
+        @ShortCircuit("value")
+        protected boolean needsValue() {
+            return false;
+        }
```
eregon: this seems a strange usage of `@ShortCircuit`.

truffle/src/main/java/org/jruby/truffle/nodes/core/TrufflePrimitiveNodes.java
```diff
@@ -565,4 +575,93 @@ public DynamicObject printInterleavedBacktrace() {
...
+
+        @Specialization
+        public DynamicObject set(VirtualFrame frame, boolean hasValue, DynamicObject value) {
+            if (writeNode == null) {
```
eregon: Why lazy? Eager sounds better, and a normal RubyNode instead of CoreMethodNode.
If lazy makes sense, please use `@Cached`.

eregon: See https://github.com/jruby/jruby/wiki/Developing-with-JRuby-Truffle

truffle/src/main/java/org/jruby/truffle/nodes/core/TrufflePrimitiveNodes.java
```diff
@@ -565,4 +575,93 @@ public DynamicObject printInterleavedBacktrace() {
...
+                CompilerDirectives.transferToInterpreter();
+                writeNode = insert(
+                        new WriteInstanceVariableNode(getContext(), getSourceSection(), "$?",
+                                new ThreadLocalObjectNode(getContext(), getSourceSection()),
+                                getValue(),
+                                true));
```
eregon: Otherwise, maybe `$? = some_value` could be allowed in `core:/`

truffle/src/main/java/org/jruby/truffle/nodes/core/TrufflePrimitiveNodes.java
pitr-ch: Afaik, It has to be DSL node because CoreMethods requires NodeFactory. Lazy is required since `getValue` would return `null` during initialisation.

truffle/src/main/java/org/jruby/truffle/nodes/core/TrufflePrimitiveNodes.java
pitr-ch: good idea, I'll try.

truffle/src/main/java/org/jruby/truffle/nodes/core/TrufflePrimitiveNodes.java
pitr-ch: yeah it is, but i do not know about another way how to disable value being executed twice.

truffle/src/main/java/org/jruby/truffle/nodes/core/TrufflePrimitiveNodes.java
pitr-ch: done.

eregon:
Looks good!

Issue #3313, opened 2015-09-07
Running threads stay running after teardown


stewartmatheson:
I would expect when I call tearDown on a ruby instance it would destroy any running ruby threads. Running threads however stay running which leads to memory leaks as the JVM will never GC objects with references in these running threads. This becomes a real issue when using servers such as torquebox that keep the same JVM running while destroying and creating ruby instances.

On a sidenote: I am looking for a way to get involved in the JRuby community. I have loved using JRuby for 2 years now and I am keen to give some of my time up to help out. I am having a bit of trouble figuring out where to start though. If this change is possible I would love it if some one could mentor me though this patch to get me started. Nothing too big just a quick chat session or even a 5 min skype call.

headius:
9k fix is sufficient, or are you running 1.7.x?

stewartmatheson:
We run 1.7.x under torquebox 3.

nirvdrum:
I've fixed similar issues on several gems when I was running 1.7.x. I've found the following approach works well:

https://github.com/wr0ngway/lumber/commit/4abd5a8e11ee086544bd98a7c03b3f9eb8fc1099

stewartmatheson:
Hey @nirvdrum I feel like jRuby should do this automaticlly. I would not trust a lot of MRI developers to think in an embedded way. https://github.com/mongodb/mongo-ruby-driver/blob/a639209ed055ff8486a8469d8c2ad924ed2049c7/lib/mongo/server/monitor.rb is the example that got me here. How many gems have you fixed up?

headius:
@stewartmatheson So to clarify, do you believe you just need the threads removed from the thread list?

My main concern here is that running threads may still reference the runtime and may attempt to access resources that have been shut down.

In a related situation, we have discussed various ways to keep the runtime alive even if the main thread exits but have never decided on one. For example, we might provide JRuby.main_thread_shutdown = false or something.

I'd like to understand your use case a bit better. Why is the runtime shutting down when there's still threads running that might access it? Are they external Java threads that were "adopted" into the runtime, or are they Ruby threads? Why can't you shut down the threads yourself?

nirvdrum:
@stewartmatheson I've had to fix at least a half dozen gems. While frustrating, everyone has been happy to accept the change at the very least. I'm hoping it ends up becoming a common pattern.

nirvdrum:
[...]

PR #3316, opened 2015-09-09
Remove running threads on tearDown

stewartmatheson:
See #3313 .

### DIFF
--- a/core/src/main/java/org/jruby/Ruby.java
+++ b/core/src/main/java/org/jruby/Ruby.java
@@ -3297,7 +3297,7 @@ public final class Ruby implements Constantizable {
             }
         }
 
-        getThreadService().disposeCurrentThread();
+        getThreadService().getRubyThreadMap().clear();
 
         getBeanManager().unregisterCompiler();
         getBeanManager().unregisterConfig();
--- a/core/src/test/java/org/jruby/runtime/TestThreadTearDown.java
+++ b/core/src/test/java/org/jruby/runtime/TestThreadTearDown.java
@@ -1,0 +1,43 @@
+package org.jruby.runtime;
+
+import junit.framework.TestCase;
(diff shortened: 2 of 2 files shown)


rtyler:
This looks reasonable to me, but what do I know :)

headius:
Oh, is this sufficient for #3313? I thought you would have expected JRuby to actually *kill* those threads, which I wasn't willing to do (and might not even be able to do, safely).

stewartmatheson:
Hey Charles,

Thanks for the response. I'm not %100 sure how to would "kill" threads. My test indicates that the threads are no longer running after calling tearDown so I was assuming that the patch was killing threads. Is there a better way to do this?

As for safely my thoughts are that tearDown was the same as killing a ruby process.  Developers working with MRI seem to be starting threads with no intent of stopping them safely. Mongo's new ruby driver is an example of this, they start a heartbeat thread that never dies.

headius:
@stewartmatheson It's possible to iterate over all threads and kill them now, using Ruby.getThreadService().getActiveRubyThreads(). Note that this will usually only give you the threads initiated from within that JRuby instance, but I think that's what you want.

For the record, I do not believe our Ruby.teardown should forcibly kill off user threads. We've actually had complains that teardown does *too much* in cases where people just allow the main thread to terminate but *expect* their other threads to keep running. We can't please everyone here.

However, I think for embedding purposes it's worth adding a *new* mechanism to "teardown and stop threads". This *might* be the default behavior for ScriptingContainer, but that's debatable.

Probably not going to get this until major work in 9.2. Help wanted!

headius:
I have created an updated version of this in #5390 that clears a few additional references, but it still does not do anything to kill threads. I am starting to suspect that's not a necessary condition to fix #3313?

PR #3325, opened 2015-09-14
"[2].to_enum.find_index(2.0)" returns 0 (to match MRI behavior)

alexdowad:
This is a bug which I found while testing one of my gems on JRuby. Please have a look!


kares:
very najs, I think @eregon would prefer having the *spec/ruby/core/enumerable/find_index_spec.rb* updated in a separate commit?

eregon:
Yes, that would be nicer to merge back upstream (essentially the commit name would reflect just the spec not the "fix" part)

alexdowad:
You got it. Have a look now.

Issue #3324, opened 2015-09-14
_id2ref should raise RangeError rather than return nil for no object found


chrisseaton:
JRuby:

```
irb(main):001:0> ObjectSpace._id2ref 545556
=> nil
```

MRI:

```
irb(main):001:0> ObjectSpace._id2ref 545556
RangeError: 0x00000000085314 is not id value
	from (irb):1:in `_id2ref'
	from (irb):1
	from /Users/chrisseaton/.rbenv/versions/2.2.3/bin/irb:11:in `<main>'
```

bitfurry:
@chrisseaton can specify the versions of Jruby and MRI ?
As for MRI 2.2.2 and Jruby 9.0.2.0 is not showing this error.

chrisseaton:
```
$ ~/.rbenv/versions/2.2.2/bin/irb 
irb(main):001:0> ObjectSpace._id2ref 545556
RangeError: 0x00000000085314 is not id value
  from (irb):1:in `_id2ref'
  from (irb):1
  from /Users/chrisseaton/.rbenv/versions/2.2.2/bin/irb:11:in `<main>'

$ ~/.rbenv/versions/2.2.3/bin/irb 
irb(main):001:0> ObjectSpace._id2ref 545556
RangeError: 0x00000000085314 is not id value
  from (irb):1:in `_id2ref'
  from (irb):1
  from /Users/chrisseaton/.rbenv/versions/2.2.3/bin/irb:11:in `<main>'

$ ~/.rbenv/versions/jruby-9.0.0.0/bin/ruby -X+O -S irb
irb(main):001:0> ObjectSpace._id2ref 545556
=> nil

$ ~/.rbenv/versions/jruby-9.0.1.0/bin/ruby -X+O -S irb
irb(main):001:0> ObjectSpace._id2ref 545556
=> nil

$ bin/jruby -X+T -S irb
irb(main):001:0> ObjectSpace._id2ref 545556
=> nil
```

bitfurry:
@chrisseaton I fixed the issue. Please feel to comment on my pull request.

PR #3327, opened 2015-09-15
Issue #3324

bitfurry:
The issue was that _id2ref function was retruing nil value. I fixed it to throw a RangeError. Please correct me if I am wrong at any point. 

jruby_pullrequest

### DIFF
--- a/core/src/main/java/org/jruby/RubyObjectSpace.java
+++ b/core/src/main/java/org/jruby/RubyObjectSpace.java
@@ -114,7 +114,7 @@ public class RubyObjectSpace {
                 return object;
             } else {
                 runtime.getWarnings().warn("ObjectSpace is disabled; _id2ref only supports immediates, pass -X+O to enable");
-                return runtime.getNil();
+                throw recv.getRuntime().newRangeError(String.format("0x%016x is not id value", longId));
             }
         }
     }
--- a/truffle/src/main/ruby/core/rubinius/common/object_space.rb
+++ b/truffle/src/main/ruby/core/rubinius/common/object_space.rb
@@ -35,8 +35,7 @@ module ObjectSpace
     if find_object([:object_id, Integer(id)], ary) > 0
       return ary.first
     end
-
-    return nil
+  return nil
   end
 
   def self.find_references(obj)
--- a/hs_err_pid364.log
+++ b/hs_err_pid364.log
@@ -1,0 +1,911 @@
+#
+# A fatal error has been detected by the Java Runtime Environment:
+#
+#  SIGBUS (0x7) at pc=0x00007f400a356582, pid=364, tid=139913045374720
+#
+# JRE version: Java(TM) SE Runtime Environment (7.0_80-b15) (build 1.7.0_80-b15)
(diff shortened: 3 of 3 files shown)


kares:
najs, would you mind cleaning-up the commits ? `git squash` them into one and maybe avoid touching *truffle/src/main/ruby/core/rubinius/common/object_space.rb* as the truffle part is unaffected ...

core/src/main/java/org/jruby/RubyObjectSpace.java
```diff
@@ -114,7 +114,7 @@ public static IRubyObject id2ref(IRubyObject recv, IRubyObject id) {
                 return object;
             } else {
                 runtime.getWarnings().warn("ObjectSpace is disabled; _id2ref only supports immediates, pass -X+O to enable");
-                return runtime.getNil();
+                throw recv.getRuntime().newRangeError(id+" This is not id value");
```
chrisseaton: Can you try to replicate the error message exactly? It should be `0x00000000000000 is not id value`, which you can probably make with `String.format('%016x is not id value', id)`.

truffle/src/main/ruby/core/rubinius/common/object_space.rb
```diff
@@ -35,8 +35,7 @@ def self._id2ref(id)
     if find_object([:object_id, Integer(id)], ary) > 0
       return ary.first
     end
-
-    return nil
+  return nil
```
chrisseaton: This is a valid commit, but it should be separate from this PR.

core/src/main/java/org/jruby/RubyObjectSpace.java
bitfurry: @chrisseaton I will sure.

truffle/src/main/ruby/core/rubinius/common/object_space.rb
bitfurry: Okay, I will create a seperate PR.

nirvdrum:
@bitfurry Thanks for the patch! In the future, please include the issue title or some other context in the commit message. We have a fair number of people that use the JRuby repository without constant access to GitHub. While "issue #3324" links great on GitHub site, it's far less usable from the command-line or non-GitHub mirrors.

chrisseaton:
The failure looks unrelated, and I'll merge now rather than pestering you about the unrelated line again.

Thanks very much for your contribution!

truffle/src/main/ruby/core/rubinius/common/object_space.rb
eregon: How is this valid?

chrisseaton: Hmm now you've said that yeah it doesn't make sense. Don't worry - I'll revert it.

bitfurry: @chrisseaton do I need to revert my whole commit or you will revert this part only ?

chrisseaton: Don't worry - I fixed it in 6b911f3a3af6c01adecee5a8ebc72eac81899deb.

bitfurry: Thanks :+1:

PR #3328, opened 2015-09-15
remove support for aliasing Java indexed getters as Ruby readers

### LINKED ISSUE #3262: jruby 9.0.0.0 Java::JavaLang::Character.name
>> Java::JavaLang::Character.name
ArgumentError: wrong number of arguments (0 for 1)
	from (irb):47:in `<eval>'
	from org/jruby/RubyKernel.java:979:in `eval'
	from org/jruby/RubyKernel.java:1099:in `catch'
	from org/jruby/RubyKernel.java:1099:in `catch'

kares:
fixes #3262 ... its a non-backward compatible change thus fired up here to decide

### DIFF
--- a/core/src/main/java/org/jruby/javasupport/binding/Initializer.java
+++ b/core/src/main/java/org/jruby/javasupport/binding/Initializer.java
@@ -182,9 +182,7 @@ public abstract class Initializer {
             if (javaPropertyName != null) {
                 if (rubyCasedName.startsWith("get_")) {
                     rubyPropertyName = rubyCasedName.substring(4);
-                    if (argCount == 0 ||                                // getFoo      => foo
-                            argCount == 1 && argTypes[0] == int.class) {    // getFoo(int) => foo(int)
-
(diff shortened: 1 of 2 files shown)


enebo:
I am willing to break this and see if anyone has ever happened across this syntax.  I know we never wrote about it in our book.  We also have no specs.  The original code was landed in 2007 along with our first JI JavaBean shorthand.  I am going to wait until 9.0.3.0 just so we can accidentally stumble over someone using this instead of doing it a day or two before release.

enebo:
Oh unless anyone else thinks this is being used...Obviously we do not want to break code but this is so unknown none of us knew it even existed...

rosenfeld:
My code used this since I thought it was supported on purpose and it broke when I upgraded JRuby to 9.0.5.0 from 1.7.11 I think. But that's okay, a few changes and I got it back to work. Just having some issues with latest RSpec since sometimes exception.backtrace is nil in JRuby for some reason...

Issue #2381, opened 2014-12-29
IO#rewind ... doesn't. (when backed by a ByteArrayInputStream).


headius:
From: http://jira.codehaus.org/browse/JRUBY-5891. Works in 1.7.18 and not on master.

See: https://gist.github.com/1043829 which includes a test case.

If you have a ByteArrayInputStream, then call #to_io on it, then you read from the stream and #rewind it, the offset into the stream doesn't change. I believe this is due to lseek dropping out the bottom of a conditional when it should perhaps throw an exception.

I am happy to provide a JUnit test and patch if folks agree with me that this is the wrong behavior. I just spent the better part of a day fixing broken behavior whose genesis was this issue.

kares:
seems to be working "correctly" on **9.0.1.0** :
```
data size before rewind: 100
Errno::EPIPE: Broken pipe - No message available
  rewind at org/jruby/RubyIO.java:1739
   <top> at bais_rewind.rb:10
```
... although I think with a little trick (by-passing `newChannel`) it would be possible to get ByteArrayInputStream seek-able - which I do found useful (will fire up a PR)

enebo:
@kares is this finished?

kares:
yep - closing

PR #3337, opened 2015-09-17
support (seekable) ByteArrayInputStream#to_io.rewind

kares:
except `FileInputStream` it seems quite useful to be able to rewind/seek a `byte[]` backed stream

this is not possible using `Channels.newChannel` as the returned channel (except for files) is not seekable
but using the 9K refactored RubyIO its fairly easy to play a trick by returning a custom channel instance ...

1.7.x remains behaving weirdly on `rewind` ... more notes at #2381


headius:
@kares I appreciate the attribution but the code at https://github.com/jruby/jruby/commit/6142ea3c062761bf9d8c926487a46b578aad4f58#diff-b60adbaec9c15033c4eb0231a09d211aR46 worries me. We can't incorporate code from GPL codebases like this, even if they're been massaged into place a little.

I'd suggest you look at the Android implementation of this logic, which is licensed Apache-2.0. We can incorporate them into a separate file with an Apache license header and it should be compatible with all three of our licenses.

Here's the Android 4.3 impl of Channels at equivalent read method: https://android.googlesource.com/platform/libcore/+/android-4.3_r1/luni/src/main/java/java/nio/channels/Channels.java#297

kares:
@headius thanks, the problematic part is now re-implemented as suggested.
tests fail due missing jnr-posix SNAPSHOT (currently, there's no green travis-ci build to rebase this off).

headius:
@kares I'll pull it down today, review and make sure it's building, and push it out.

headius:
Testing seems fine, I'll push shortly.

Last suggestions, if you get a chance (just throw them on master):

* Try to grab and setAccessible the fields you access during static initialization; if we can't do it, we need to fall back to the unseekable wrapper.
* Saving the fields will also be better perf rather than looking up and hitting setAccessible every time they're accessed.

PR #3345, opened 2015-09-23
[Truffle] Basics of yaml parsing

chrisseaton:
Don't merge - work in progress but looking for an initial review from any of @eregon, @pitr-ch or @nirvdrum please


truffle/pom.rb
```diff
@@ -10,6 +10,7 @@
 
               'jruby.basedir' => '${basedir}/..' )
 
+  jar 'org.yaml:snakeyaml:1.14'
```
chrisseaton: Not sure if we should add a concrete dependency like this. JRuby seems to lazily add it to the class path or something, but not sure where it gets it from.

truffle/src/main/java/org/jruby/truffle/nodes/control/ReturnNode.java
```diff
@@ -33,6 +33,8 @@ public ReturnNode(RubyContext context, SourceSection sourceSection, ReturnID ret
 
     @Override
     public Object execute(VirtualFrame frame) {
+        //ruby(frame, "Truffle::Primitive.debug_print 'HERE!!'; Truffle::Primitive.print_interleaved_backtrace");
+        //System.err.println(getEncapsulatingSourceSection().getShortDescription());
```
chrisseaton: Junk.

truffle/src/main/java/org/jruby/truffle/runtime/core/CoreLibrary.java
```diff
@@ -623,6 +635,13 @@ private void initializeConstants() {
         Layouts.MODULE.getFields(encodingConverterClass).setConstant(node, "XML_TEXT_DECORATOR", EConvFlags.XML_TEXT_DECORATOR);
         Layouts.MODULE.getFields(encodingConverterClass).setConstant(node, "XML_ATTR_CONTENT_DECORATOR", EConvFlags.XML_ATTR_CONTENT_DECORATOR);
         Layouts.MODULE.getFields(encodingConverterClass).setConstant(node, "XML_ATTR_QUOTE_DECORATOR", EConvFlags.XML_ATTR_QUOTE_DECORATOR);
+
+        Layouts.MODULE.getFields(objectClass).setConstant(node, "RUBY_REVISION", Constants.RUBY_REVISION);
```
chrisseaton: Junk

truffle/src/main/java/org/jruby/truffle/translator/BodyTranslator.java
```diff
@@ -1135,6 +1135,12 @@ public RubyNode visitDefsNode(org.jruby.ast.DefsNode node) {
     }
 
     protected RubyNode translateMethodDefinition(SourceSection sourceSection, RubyNode classNode, String methodName, org.jruby.ast.Node parseTree, org.jruby.ast.ArgsNode argsNode, org.jruby.ast.Node bodyNode) {
+        //
```
chrisseaton: Should be

```ruby
// Psych calls method names generated from class hierarchies. We have Truffle::Psych, so we need to adapt
```

lib/ruby/truffle/mri/yaml.rb
```diff
@@ -0,0 +1 @@
+require_relative '../../stdlib/' + File.basename(__FILE__)
```
eregon: Could we just have a symlink? I guess that might be a problem with the jar though.

truffle/src/main/java/org/jruby/truffle/nodes/RubyNode.java
```diff
@@ -214,13 +213,19 @@ public MemoryManager getMemoryManager() {
...
+    protected Object ruby(Frame frame, String expression, Object... arguments) {
         final MaterializedFrame evalFrame = setupFrame(frame, arguments);
         final DynamicObject binding = Layouts.BINDING.createBinding(getContext().getCoreLibrary().getBindingFactory(), evalFrame);
         return getContext().eval(expression, binding, true, "inline-ruby", this);
     }
 
-    private MaterializedFrame setupFrame(VirtualFrame frame, Object... arguments) {
+    private MaterializedFrame setupFrame(Frame frame, Object... arguments) {
         CompilerDirectives.transferToInterpreter();
         final MaterializedFrame evalFrame = Truffle.getRuntime().createMaterializedFrame(
```
eregon: Are these changes needed or can be removed? Otherwise you'll get some conflict :wink:

lib/ruby/truffle/mri/yaml.rb
chrisseaton: Yeah doesn't work in the jar.

truffle/src/main/java/org/jruby/truffle/nodes/RubyNode.java
chrisseaton: I'll merge in with whatever you have.

truffle/src/main/java/org/jruby/truffle/nodes/ext/psych/PsychParserNodes.java
```diff
@@ -0,0 +1,350 @@
...
+import org.yaml.snakeyaml.reader.ReaderException;
+import org.yaml.snakeyaml.reader.StreamReader;
+import org.yaml.snakeyaml.scanner.ScannerException;
+
+import java.io.ByteArrayInputStream;
+import java.io.InputStreamReader;
+import java.nio.charset.Charset;
+import java.util.Map;
+
+@CoreClass(name = "Truffle::Psych::Parser")
```
eregon: Could we just define this as Psych::Parser?
Then in CoreLibrary we would not load any of this by default, but only when "yaml" is required.
Also it could improve startup a bit :)

truffle/src/main/java/org/jruby/truffle/nodes/ext/psych/PsychParserNodes.java
```diff
@@ -0,0 +1,350 @@
...
+        }
+
+        @Specialization(guards = {"isRubyString(yaml)", "isRubyString(path)"})
+        public Object parse(DynamicObject parserObject, DynamicObject yaml, DynamicObject path) {
+            return doParse(parserObject, yaml, path);
+        }
+
+        @CompilerDirectives.TruffleBoundary
+        private Object doParse(DynamicObject parserObject, DynamicObject yaml, DynamicObject path) {
+            boolean tainted = (boolean) ruby("yaml.tainted? || yaml.is_a?(IO)", "yaml", yaml);
```
eregon: Ah I see to use ruby() behind a truffle boundary.

truffle/src/main/java/org/jruby/truffle/nodes/ext/psych/PsychParserNodes.java
chrisseaton: Yes we could probably make all core library nodes lazy - not just those for ext modules.

truffle/src/main/java/org/jruby/truffle/translator/BodyTranslator.java
```diff
@@ -1135,6 +1135,12 @@ public RubyNode visitDefsNode(org.jruby.ast.DefsNode node) {
     }
 
     protected RubyNode translateMethodDefinition(SourceSection sourceSection, RubyNode classNode, String methodName, org.jruby.ast.Node parseTree, org.jruby.ast.ArgsNode argsNode, org.jruby.ast.Node bodyNode) {
+        // Psych calls method names generated from class hierarchies. We have Truffle::Psych, so we need to adapt
+
+        if (methodName.startsWith("visit_Psych_Nodes_")) {
+            methodName = "visit_Truffle_Psych_Nodes_" + methodName.substring("visit_Psych_Nodes_".length());
+        }
+
```
eregon: Would this be needed if we defined as `Psych::SomeClass` directly?

chrisseaton: No.

truffle/src/main/java/org/jruby/truffle/translator/TranslatorDriver.java
```diff
@@ -167,7 +167,9 @@ private RubyRootNode parse(Node currentNode, RubyContext context, Source source,
 
         // Catch return
 
-        truffleNode = new CatchReturnAsErrorNode(context, truffleNode.getSourceSection(), truffleNode);
+        if (parserContext != ParserContext.EVAL) {
+            truffleNode = new CatchReturnAsErrorNode(context, truffleNode.getSourceSection(), truffleNode);
+        }
```
eregon: Indeed, although ReturnID might get messed up if there is a `return` inside eval, right?

eregon:
Looks good, where are the passing tests/specs?

truffle/src/main/java/org/jruby/truffle/nodes/ext/psych/PsychParserNodes.java
eregon: When would define the methods then? Here I'm thinking `require` should define the Psych module and its methods.
`Fixnum` should exist even if no one called any methods on a integer yet for instance.

chrisseaton:
This is enough to run the examples in the documentation. I'll get emitting working, then the full set of tests.

truffle/src/main/java/org/jruby/truffle/nodes/ext/psych/PsychParserNodes.java
chrisseaton: I think it's something that could be defined as part of the whole CoreMethod and CoreMethodNodeManager mechanism for all core methods, rather than just Psych. When you first look up methods on Fixnum, they get loaded.

Issue #3341, opened 2015-09-21
autoload & require deadlock


bignacio:
I run into an interesting problem when using require and autoload for the same classes.

I only tested it against JRuby 1.7.21 and quickly looking at the code I'd attributed the problem to combination of a global lock for autoload consts and the require lock.

Since autoload [at some point will call require](https://github.com/jruby/jruby/blob/a741a82fc802e483f7a8e798db34d7bc0824bb37/core/src/main/java/org/jruby/RubyModule.java#L3832) and try to [acquire a lock for a file](https://github.com/jruby/jruby/blob/a741a82fc802e483f7a8e798db34d7bc0824bb37/core/src/main/java/org/jruby/runtime/load/LoadService.java#L444), that same file can contain autoloaded classes that in turn will need to acquire the autoload lock.

I wrote a little, perhaps convoluted, sample code to reproduce the problem. It is gisted here https://gist.github.com/bignacio/7a947960113775477c72 and I copy it below

Save all the files into the same directory and run

$ jruby main.rb

as any good deadlock, it won't happen all the time so you can do something like this

$ while [ true ] ; do jruby main.rb  ; done

and wait untill it hangs.

You can confirm the dealock with jstack
Once the process is hung, get its pid and run

jstack -l < pid >

```ruby
# another_class_base.rb
module DeadlockTest
  class AnotherClassBase
  end
end

# file:  class_a.rb
module DeadlockTest
  autoload :AnotherClassBase, 'another_class_base'
  class ClassA
    class AnotherClass < AnotherClassBase
    end
  end
end

# file:  deadlock_test.rb
module DeadlockTest
  autoload :ClassA, 'class_a'
end

# file: main.rb
require 'thread'
require 'deadlock_test'

puts("Starting, JRuby version: #{JRUBY_VERSION}")
ta = Thread.new{
  puts("Thread A #{JRuby.reference(Thread.current).native_thread}\n")

  # This is not a usual thing for people to do but it can happen if an external component is loading
  # your ruby code
  # I first run into this problem in a storm topology (via RedStorm), which does
  # a similar thing. I'm sure there are other ways to reproduce the same problem
  # without a require but this
  require 'class_a'
}

tb = Thread.new{
  puts("Thread B #{JRuby.reference(Thread.current).native_thread}\n")

  o = DeadlockTest::ClassA.new
  puts('if you see this, try again :) ')
}

ta.join
tb.join
```

jstack output

```
Found one Java-level deadlock:
=============================
[...]

PR #3370, opened 2015-10-05
treat deadlocks as circular require - as MRI does it

mkristian:
#3341 has two simple tests which produces deadlocks on the current master branch.

this PR does fix it to the extend that I do not see deadlocks anymore.

@headius please review and I would some input on tests, i.e. it should be possible to run those tests from #3341 on CI but I am not if this makes sense


headius:
I'll have a look.

PR #3383, opened 2015-10-10
public org.jruby.Runtime.Profile.Builtin::MethodData

kml:
this makes building custom tooling easier

http://pastebin.com/XhNYdMtD

### DIFF
--- a/core/src/main/java/org/jruby/runtime/profile/builtin/InvocationSet.java
+++ b/core/src/main/java/org/jruby/runtime/profile/builtin/InvocationSet.java
@@ -27,7 +27,7 @@ package org.jruby.runtime.profile.builtin;
 
 import java.util.List;
 
-class InvocationSet {
+public class InvocationSet {
     
     final List<Invocation> invocations;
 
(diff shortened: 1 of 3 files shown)


kml:
Thanks to changing visiblity I can compile my modified copies of all default ProfilePrinters.
If it does not make sense to accept all those changes please ignore the PR.

kares:
originally those package-protected classes were hidden on purpose as they were not polished. 
but assuming no API review is about to happen soon its probably fine ... as a semi-official API :)

headius:
@kares was this closed by mistake?

kares:
@headius seems like it might have been ... re-opening (@BanzaiMan feel free to close with feedback)

kml:
From my perspective I don't need this right now.
I've done profiling using patched version of jruby and I'm not sure if anyone else need it.

kares:
these are still valid (even in 9K's context), although some of the stuff such as making fields public might be seen as internals and thus subject to change. believe there's no reason not to merge.

PR #3412, opened 2015-10-22
[Truffle] make j+tr part of the distribution

pitr-ch:


bin/jruby+truffle.bat
```diff
@@ -0,0 +1 @@
+echo Windows are not yet supported by jruby+truffle.
```
nirvdrum: English is weird. You really want to say "Windows is not" here because it's a singular item referring to the name of the operating system.

eregon: `is`, no?

bin/jruby+truffle
```diff
@@ -0,0 +1,5 @@
+#!/usr/bin/env ruby
+
+require 'jruby+truffle/runner'
+
+JRubyTruffleRunner.new
```
eregon: How is this going to work? you need a require_relative I think.

nirvdrum:
Looks good.

mkristian:
looks good for me as well. but I really would like to hear some comment from @headius  and @enebo about adding this to ./lib/ruby/stdlib

bin/jruby+truffle
pitr-ch: The runner file is on the same path as stdlib so it always available in jruby.

eregon: But what if I want to run it with MRI?

pitr-ch:
@enebo @headius: is this too invasive, or can it be merged?

enebo:
@pitr-ch is there any reason this cannot be in lib/ruby/truffle?  stdlib is mostly for files which are in MRI's stdlib.  I think we have some variants like jossl (although I think this is a mistake) but mostly those variants are working around differences between MRI and JRuby.

pitr-ch:
@enebo This tool is suppose to be loaded in JRuby it can prepare environment to run files with gems in JRuby+Truffle and then run files in the prepared environment. The tool always runs in JRuby it computes required options and then it executes JRuby+Truffle with them. If it is in lib/ruby/truffle it won't be visible to JRuby.

mkristian:
@pitr-ch as I understand it, the tool is executes exclusively via the bin/jruby+truffle

so instead of https://github.com/jruby/jruby/pull/3412/files#diff-0e9caaa4ef2b36ca7f4ea6308854ebd7R3
you could use something like
```
require File.join(JRuby.runtime.instance_config.jruby_home, 'lib/ruby/stdlib/jruby+truffle/runner')
```

pitr-ch:
@mkristian ah thanks for the suggestion. Sounds good, I'll update the PR.

pitr-ch:
@mkristian @enebo updated and ready for merge.

mkristian:
I think travis does not really run into this change. merging.

mkristian:
@pitr-ch thanx for all this shifting around a second time

pitr-ch:
No problem at all @mkristian. I was a better option.

PR #3420, opened 2015-10-23
Removing no longer needed import of a deprecated class

jtulach:
I've removed the TruffleVM from main development line of Truffle. Consider stop using it.


nirvdrum:
It's an unused import anyway. We should have removed this before.

PR #3421, opened 2015-10-23
Fix GzipWriter output

### LINKED ISSUE #1371: gzip issue with jruby and Rack::ETag
A few months ago I reported this issue against rack:

https://github.com/rack/rack/issues/571 (gzip issue with jruby and Rack::ETag)

It seems the problem may be with the way jruby is implementing gzip compression, which apparently is is doing some strange reuse of objects internally.

The bug report above includes a test case to reproduce and some analysis by @srawlins

In short the problem appears to be that rack's etag handler is concatenating parts that are rightfully assumed to be immutable but that are actually being modified in jruby's gzip implementation. Since this seems to be a jruby bug, the issue was closed a few days on the rack side.

dmarcotte:
GZIPOutputStream was passing one of its internal byte buffers to IOOutputStream#write, which in turn used those bytes to back the ByteList for its output.  So every subsequent time GZIPOutputStream wrote to that buffer, it overwrote the bytes behind the back of all the previous chunks of output.

Ensure that IOOutputStream is working on a copy of the bytes it gets from GZIPOutputStream to fix this issue.

Fixes #1371

Let me know if there's any questions, and thanks again!

### DIFF
--- a/core/src/main/java/org/jruby/ext/zlib/JZlibRubyGzipWriter.java
+++ b/core/src/main/java/org/jruby/ext/zlib/JZlibRubyGzipWriter.java
@@ -176,7 +176,25 @@ public class JZlibRubyGzipWriter extends RubyGzipFile {
         try {
             // the 15+16 here is copied from a Deflater default constructor
             Deflater deflater = new Deflater(level, 15+16, false);
-            io = new GZIPOutputStream(new IOOutputStream(realIo, false, false), deflater, 512, false);
+            final IOOutputStream ioOutputStream = new IOOutputStream(realIo, false, false) {
+                /**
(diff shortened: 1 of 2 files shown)


dmarcotte:
Just pushed a small extra change on this pull: b2dc3da.  Apologies for the noise.

dmarcotte:
Just saw Travis is showing red on this pull because the 1_7 branch went red while I was working on this.  See here for a green run of these changes: https://travis-ci.org/dmarcotte/jruby/builds/87085859

kares:
:point_up: very najs find - much appreciated!

dmarcotte:
Thanks @kares!

PR #3422, opened 2015-10-24
Fix Fixnum#pred overflow

grddev:
Without this `(-2**63+1).pred` returns a positive number. Given that the implementation hasn't changed (substantially) since the initial version in 2008, it seems all JRuby versions should be affected.


kares:
thanks, put [this one](https://github.com/jruby/jruby/commit/1c4c2962451a3ed3a8b9cb0eca740168e67fcf9f) onto jruby-1_7 it shall get to master with the branch merge

PR #3431, opened 2015-10-27
[Truffle] add Rubinius memory barrier and AtomicReference

pitr-ch:
Adding to support c-r synchronisation layer


eregon:
Looks good!
Just curious, why implementing methods directly instead of primitives?
For cheaper compilation?

pitr-ch:
`get_and_set` would be reimplemented whole anyway, the primitives are only used by AtomicReference class, so it seemed cleaner to implement all the methods in the same way.

PR #3458, opened 2015-11-12
Ruby 2.3's #dig

kares:
`[].dig(0)` and `{}.dig(:a)` as introduced by MRI [**2.3.0.preview1**](https://github.com/ruby/ruby/commit/29862685c0acf3a40c6b1f9e8780cbbd86cba658)

### DIFF
--- a/spec/ruby/core/array/dig_spec.rb
+++ b/spec/ruby/core/array/dig_spec.rb
@@ -1,0 +1,21 @@
+require File.expand_path('../../../spec_helper', __FILE__)
+
+describe "Array#dig" do
+
+  it "returns at with one arg" do
+    ['a'].dig(0).should == 'a'
+    ['a'].dig(1).should be_nil
+  end
+
+  it "recurses array elements" do
+    a = [ [ 1, [2, '3'] ] ]
+    a.dig(0, 0).should == 1
+    a.dig(0, 1, 1).should == '3'
+    a.dig(0, -1, 0).should == 2
+  end
+
+  it "raises without any args" do
+    lambda { [10].dig() }.should raise_error(ArgumentError)
+  end
+
+end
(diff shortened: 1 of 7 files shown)


eregon:
@kares Nice work and specs here!
Could you make a PR against ruby/rubyspec?
That would avoid a possible spec duplication and to test the specs on multiple MRI versions :smiley: 
I can then merge quickly the specs back to jruby master.

kares:
@eregon thx, I thought the work-flow is that you pick those eventually back to ruby/rubyspec.
so than I should remove those [rubyspec] commits from here ? ... also not sure how long this would be open before getting merged - maybe its only to be commited after 2.3 final as JRuby 9.1 will come out

eregon:
@kares For new feature specs I think it's more beneficial to PR against the ruby/rubyspec repo.
Specs for existing stuff is less critical and can more easily be merged later.
Also, this allows to tweak the spec while it's getting stable in MRI.

A guard of minimal version 2.3 will be needed anyway for this spec to pass on older versions, so this spec will only run when JRuby declares RUBY_VERSION to be "2.3.x" (it's possible to ignore such a guard with --unguarded while developing).

Yes, please remove them from here if you do the PR so there is no chance for conflict :)

spec/ruby/core/array/dig_spec.rb
```diff
@@ -0,0 +1,21 @@
...
+    a.dig(0, 0).should == 1
+    a.dig(0, 1, 1).should == '3'
+    a.dig(0, -1, 0).should == 2
+  end
+
+  it "raises without any args" do
+    lambda { [10].dig() }.should raise_error(ArgumentError)
+  end
+
+end
```
benlovell: Missing a terminating newline character here.

kares:
put the commits into **ruby-2.3** which should merge to master after 2.3 final when 9.1.0 is released

PR #3469, opened 2015-11-18
beyond JRuby's StandardErrorLogger

kares:
improvements and some new features for JRuby's internal logging
there seems to be a lot of commits but usually with little change - did not want to squash them as some of them have detailed messages.

... gist of changes : 

* `-Xlog.backtraces` [can get hard to read with multiple threads](https://github.com/jruby/jruby/commit/19bd5c520a9fb6f158d8ef091f12457fd57b9698) ... targeted for **jruby-1_7** #3403 as well

* `-Xlog.callers` is ~~double logged and~~ hard to read with a `to_s` array formatting - this is changed to use MRI backtrace formatting

* some Java and even JRuby libraries (e.g. daemonization) change `System.err` which causes the StandardErrorLogger to print "nowhere" if JRuby's classes are loaded before the stream change - we'll know allways read the `System.err` field

* "improved" default (StandardErrorLogger) logger format: `2015-11-04T11:29:41.759+01:00 [main] INFO SampleLogger : hello world` ... includes thread-name and level compared to current state

* possibility to use a delegating logger - with JUL and SLF4J impls provided. this will be quite useful in embed-ing scenarios. due this a `LoggerFactory.getLogger(Class)` is introduced and should be preferred of the `(String)` version so that one could potentially configure loggers easily e.g. for the whole `org.jruby.xxx` package. note that default logger still behave the same and only uses simple name of the class as the logger name


kares:
would it be OK to have these in for 9.1 // cc @enebo @headius  ?
fully understand logging is mostly intended "internal" as is but still some of it is already useful and other parts might be tuned as needed to be used in a "traditional" Java logging setup.

enebo:
So one comment is that the use of debug("foo {}", argument) should be a net
win for us performance wise versus our use of string concat when debugging
is disabled.

I was also worried when I saw some static final DEBUGs disappear but I see
no particular harm in any individual case (over time we have ended up with
expensive parameters to log statements kill performance even with debugging
off.  The only potential downside of some removed static final DEBUG checks
is increased codesize of the method since the debug call will now need to
get inlined out of existence by JVM.

@kares my main question is how could us including slf4j end up affecting
existing users?  If they are already using slf4j would this end up
potentially break their application?  If that is the case a second question
would be whether we can shade this or not?

-Tom

On Mon, Feb 22, 2016 at 10:58 PM, Karol Bucek <notifications@github.com>
wrote:

> would it be OK to have these in for 9.1 // cc @enebo
> <https://github.com/enebo> @headius <https://github.com/headius> ?
> fully understand logging is mostly intended "internal" as is but still
> some of it is already useful and other parts might be tuned as needed to be
> used in a "traditional" Java logging setup.
>
> —
> Reply to this email directly or view it on GitHub
> <https://github.com/jruby/jruby/pull/3469#issuecomment-187531640>.
>



-- 
blog: http://blog.enebo.com       twitter: tom_enebo
mail: tom.enebo@gmail.com

kares:
@enebo the old way of static final checks can stay here and there - removed some that are useful in embed (e.g. while running with jruby-rack its annoying to always have to look for System.out / err)

we do not include SLF4J ... its an optional dependency and would need to be manually added, a SLF4J logger impl (SLF4J only required to have it compiled) is provided but anyone willing to use it must add and setup SLF4J on their own and use the 'new' `-Djruby.logger.class=org.jruby.log.SLF4JLogger`

enebo:
@kares thanks for the explanation.  This all looked fine with what I saw.
@headius may want to take a gander due to it size.  Another set of eyes
will not hurt...

On Tue, Feb 23, 2016 at 9:13 AM, Karol Bucek <notifications@github.com>
wrote:

> @enebo <https://github.com/enebo> the old way of static final checks can
> stay here and there - removed some that are useful in embed (e.g. while
> running with jruby-rack its annoying to always have to look for System.out
> / err)
>
> we do not include SLF4J ... its an optional dependency and would need to
> be manually added, a SLF4J logger impl (SLF4J only required to have it
> compiled) is provided but anyone willing to use it must add and setup SLF4J
> on their own and use the 'new'
> -Djruby.logger.class=org.jruby.log.SLF4JLogger
>
> —
> Reply to this email directly or view it on GitHub
> <https://github.com/jruby/jruby/pull/3469#issuecomment-187735005>.
>



-- 
blog: http://blog.enebo.com       twitter: tom_enebo
mail: tom.enebo@gmail.com

kares:
@enebo thanks for the review, will wait for another :vertical_traffic_light:  - as I do not want to force this on you unless you are comfortable with the change (tried to make sure it works out-of-the box the same way as so far).

kares:
@headius how's this looking for you? (obviously will need a rebase of master)

kares:
@headius :8ball: still want this reviewed ... :thought_balloon: please

kares:
release is getting close ... should we still wait for @headius to get us a review ?

enebo:
@kares I believe @headius will not be online much until monday.  We will land then I guess.

kares:
merged in to have `-Xlog.callers` and `backtraces` consistent and usable in 9.1

PR #3481, opened 2015-11-20
implement Coverage.peek_result

tenderlove:
This patch implements `Coverage.peek_result` that was implemented in MRI
here:

  https://bugs.ruby-lang.org/issues/10816

I didn't add any tests because there are test in the MRI suite that cover this.  Also, I admit that I haven't run the JRuby tests but it does compile? ¯\_(⊙︿⊙)_/¯

### DIFF
--- a/core/src/main/java/org/jruby/ext/coverage/CoverageData.java
+++ b/core/src/main/java/org/jruby/ext/coverage/CoverageData.java
@@ -42,6 +42,10 @@ public class CoverageData {
         return coverage != null;
     }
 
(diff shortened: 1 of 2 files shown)


headius:
We run the MRI tests and will be updating them for 2.3 as we wrap up behavioral changes. In any case, the change is pretty small.

headius:
Merged to ruby-2.3 in ced60e86f0cdcb644bc6fc3d46b0373d654fe46a. We'll merge to master for 9.1.0.0.

PR #3485, opened 2015-11-22
[Ruby-2.3] - Implements Hash#fetch_values

lucasallan:


kares:
there's the **ruby-2.3** branch now (targeting 9.1.0.0), so you should put the commit there

lucasallan:
@kares thanks! Will do.

PR #3486, opened 2015-11-22
[Ruby-2.3] - Implements Hash#fetch_values

lucasallan:


kares:
:+1: ... if you're interested take a look at ruby/rubyspecs for specing the feature e.g. lately I spec-ed `dig` : https://github.com/ruby/rubyspec/commit/ce9326499ba1457bb74b7e39b97b166f4acce7ec

lucasallan:
@kares We already have rubyspecs for `fetch_values` https://github.com/jruby/jruby/blob/ruby-2.3/spec/ruby/core/hash/fetch_values_spec.rb
https://github.com/kachick/rubyspec/blob/master/core/hash/fetch_values_spec.rb

klebervirgilio:
:+1:

kares:
merged master into ruby-2.3 so it should get green ... believe you could have pushed there just fine :yellow_heart:

PR #3488, opened 2015-11-23
forcing a nil (no) backtrace with Java APIs

kares:
`RaiseException(Ruby runtime, RubyClass type, String msg, IRubyObject backtrace, boolean native)` constructor despite accepting an `IRubyObject` **backtrace** parameter (and not just `RubyArray`), passing a nil does not end up with the `RubyException#backtrace` being `nil` but rather an empty array which is not fortunate as in user-land its easy to dismiss and confusing.

like to use the constructor for optional "lightweight" exception raising such as jruby/jruby-openssl#55

p.s. I'd like to consider backporting these changes to **jruby-1_7** if there's no objections to that.

### DIFF
--- a/core/src/main/java/org/jruby/RubyException.java
+++ b/core/src/main/java/org/jruby/RubyException.java
@@ -21,7 +21,7 @@
  * Copyright (C) 2004 Stefan Matthias Aust <sma@3plus4.de>
  * Copyright (C) 2005 David Corbin <dcorbin@users.sf.net>
  * Copyright (C) 2006 Michael Studman <codehaus@michaelstudman.com>
- * 
+ *
  * Alternatively, the contents of this file may be used under the terms of
  * either of the GNU General Public License Version 2 or later (the "GPL"),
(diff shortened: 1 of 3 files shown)


kares:
the changes also slightly improved performance (when `backtrace` is not accessed) e.g.
```ruby
require 'hitimes'

def measure(count = 1_000)
  Hitimes::Interval.measure do
    count.times do
      begin
        raise RuntimeError.new
      rescue RuntimeError => e
        puts e.message if false
      end
    end
  end
end

print "warmup: "
puts measure(200_000).inspect
print "......: "
puts measure(500_000).inspect
```
**BEFORE:**
```
warmup: 12.830130032
......: 29.00416128

warmup: 12.668082653
......: 27.926976294

warmup: 12.572626931
......: 28.33092909
```
**AFTER:**
```
warmup: 11.538297404
......: 25.360850043

warmup: 11.461425197
......: 25.174530815

warmup: 12.311494166
......: 26.668306498
```

PR #3494, opened 2015-11-24
Proper support for UNIXSocket.for_fd

mrbrdo:
`for_fd` method


mrbrdo:
Regarding specs, the only test that covers this is for_fd_spec.rb, but it also uses the path method, which is not implemented on JRuby.

PR #3497, opened 2015-11-25
[Ruby-2.3] - Implements Numeric#positive? and Numeric#negative?

lucasallan:
RubySpecs:
  - https://github.com/jruby/jruby/blob/ruby-2.3/spec/ruby/core/numeric/negative_spec.rb
  - https://github.com/jruby/jruby/blob/ruby-2.3/spec/ruby/core/numeric/positive_spec.rb


enebo:
Thanks Lucas....keep 'em coming! :)

PR #3515, opened 2015-12-04
Truffle pr zlib

chrisseaton:
@eregon @nirvdrum @pitr-ch please review (it's all imported code though, so no point commenting on the actual code).


chrisseaton:
For something simple like `adler32` this is about 200x slower than using Java https://github.com/djberg96/pr-zlib/issues/3#issuecomment-161114787, so this is perhaps not the final code we'll use for zlib, but the functionality seems almost perfect for now.

nirvdrum:
It looks good to me. It's surprising the imported lib gets everything but CRC32 right, though.

eregon:
Could we keep the few methods we already implement?
Particularly the highly numeric ones?

chrisseaton:
Could do. I removed them as they aren't being used in anything high performance yet and I felt like less code to maintain was better until we did. What does everyone else think? Of course the Java code is still there if we want it now, or in the future.

chrisseaton:
I'll merge this as is. The Java equivalents are not friendly to AOT, so removing them for now simplifies both our Java codebase and the SVM substitutions. We can put them back as soon as we see a high-performance use case.

pitr-ch:
@chrisseaton did you have time to investigate why is it that much slower? Bad algorithms or we do not optimise it well?

chrisseaton:
I haven't looked into it in any depth - but I just wanted it for the completeness rather than the performance for now.

PR #3517, opened 2015-12-06
[ruby-2.3 feature #10769] add Enumerable#chunk_while, spec for it

cthulhua:
Implemented as #slice_when, but slice when block does not match rather than when it does.
Also added merged over ruby spec for it.

### DIFF
--- a/core/src/main/ruby/jruby/kernel/enumerable.rb
+++ b/core/src/main/ruby/jruby/kernel/enumerable.rb
@@ -91,6 +91,35 @@ module Enumerable
       end
     end
   end
+  
+  def chunk_while(&block)
+    raise ArgumentError.new("missing block") unless block
+    return Enumerator.new {|e| } if empty?
+    return Enumerator.new {|e| e.yield self } if length < 2
+
+    Enumerator.new do |enum|
+      ary = nil
+      last_after = nil
+      each_cons(2) do |before, after|
+        last_after = after
+        match = block.call before, after
+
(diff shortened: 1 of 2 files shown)


kares:
@eregon is it fine to have the rubyspec updates coming in like they are here?

cthulhua:
I saw that a similar thing was done for slice_when, so I added them.

eregon:
@kares @cthulhua Sorry, I missed your mention.
It would be better and simpler (for merging) to just let the regular rubyspec import take care of it.
But if it's the exact same change, then most likely there won't be a conflict (to be verified though).

cthulhua:
Ah OK - would you prefer that I remove the change? I wrote the rubyspec for this so I know there is no conflict - seems more hygienic to let the standard ruby spec import handle it though.

eregon:
@cthulhua If it's not much effort, this would be cleaner :smiley: 

No, we should just import latest rubyspec into this repo before the 2.3 merge
(@kares you can ask me, otherwise I do it every month).

cthulhua:
@eregon I removed the rubyspec change. I noticed that in ruby/ruby there was a test added for chunk_while (https://github.com/ruby/ruby/blob/trunk/test/ruby/test_enum.rb#L683) so i added it to the same place under test/mri, since that seemed to be what other folks did when they added ruby-2.3 features.

eregon:
Thanks :+1:

PR #3519, opened 2015-12-06
Truffle new pack parser

chrisseaton:
This is `Array#pack` reimplemented using Antlr instead of a hand-written lexer and parser. It's good software engineering, but it's also practice for developing a new new parser for Ruby in Antlr.

@nirvdrum @pitr-ch @eregon please review - @headius and @enebo for info as you asked about this at RubyConf.


eregon:
Looks good. Do we have any perf numbers compared to the manual parser?

chrisseaton:
For the parsing itself? No - I'll do that. That's not what it's designed for of course.

eregon:
Yes, I recall doing one lexer with jflex for Date#strftime(format) parsing and numbers were interesting.

chrisseaton:
It turns out it's quite a bit slower.

```java
final PackParser parser = new PackParser(this);

while (true) {
    final long start = System.nanoTime();

    for (int n = 0; n < 1_000_000; n++) {
        parser.parse("LI<BaL>Q<nnD@", false);
    }

    System.err.println((System.nanoTime() - start) / 1e9);
}
```

```
7.178332954
6.426043492
4.051578236
4.001246475
3.96417664
4.009482471
4.048091502
4.03034866
```

```java
final PackCompiler compiler = new PackCompiler(this, null);

while (true) {
    final long start = System.nanoTime();

    for (int n = 0; n < 1_000_000; n++) {
        compiler.compile("LI<BaL>Q<nnD@");
    }

    System.err.println((System.nanoTime() - start) / 1e9);
}
```

```
31.487802185
30.452111491
30.422754027
30.572044818
30.423045225
30.470363756
```

I'd still rather have this nice code in there for now though, and it's not putting me off trying a Ruby parser in Antlr4.

chrisseaton:
The Antlr dependency is 374 KB.

PR #3532, opened 2015-12-12
[ruby-2.3] Feature #10730: Implement Array#bsearch_index

cheald:
Implements Array#bsearch_index. The binary search is now a private method which returns an int
index, and the public #bsearch and #bsearch_index methods utilize it to return an appropriate
IRubyObject instance.

This is functionally identical to how the MRI changes were implemented.

### DIFF
--- a/core/src/main/java/org/jruby/RubyArray.java
+++ b/core/src/main/java/org/jruby/RubyArray.java
@@ -2111,24 +2111,45 @@ public class RubyArray extends RubyObject implements List, RandomAccess {
 
     @JRubyMethod
     public IRubyObject bsearch(ThreadContext context, Block block) {
-        Ruby runtime = context.runtime;
-
         if (!block.isGiven()) {
             return enumeratorize(context.runtime, this, "bsearch");
         }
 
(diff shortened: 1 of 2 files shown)


core/src/main/java/org/jruby/RubyArray.java
```diff
@@ -2111,24 +2111,45 @@ public IRubyObject index(ThreadContext context, Block block) {
...
     @JRubyMethod
     public IRubyObject bsearch(ThreadContext context, Block block) {
-        Ruby runtime = context.runtime;
-
         if (!block.isGiven()) {
             return enumeratorize(context.runtime, this, "bsearch");
         }
 
+        int rVal = bsearch_index_internal(context, block);
+        if(rVal == -1) {
```
nirvdrum: Not a big deal, but I'm curious why the two `if(rVal...` statements lack a space between the `if` and `(`, but the other `if` statements do not.

cheald: Heh, because I wasn't paying attention. Will fix it!

PR #3534, opened 2015-12-12
[ruby-2.3 Feature #11253] Add support for :flags option for IO#open

cheald:
Allows for setting arbitrary flags for sysopen. Also cleaned up some unnecessary `context.runtime` calls.


kares:
could you try rebasing please - there seems to be conflicts after another merge.

cheald:
Done. I think that should do it!

PR #3535, opened 2015-12-12
[ruby-2.3 Feature #11785] add `encoding:` optional argument to String.new

cheald:
Adds kwargs support to String#initialize such that it will accept an optional :encoding argument.

### DIFF
--- a/core/src/main/java/org/jruby/RubyString.java
+++ b/core/src/main/java/org/jruby/RubyString.java
@@ -1487,12 +1487,14 @@ public class RubyString extends RubyObject implements EncodingCapable, MarshalEn
         return this;
     }
 
+    // TODO: Fold kwargs into the @JRubyMethod decorator
+    final String[] validInitializeArgs = new String[]{"encoding"};
+
     @JRubyMethod(name = "initialize", visibility = PRIVATE, optional = 2)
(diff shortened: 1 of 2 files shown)


core/src/main/java/org/jruby/RubyString.java
```diff
@@ -1486,9 +1487,18 @@ public IRubyObject initialize19(ThreadContext context) {
         return this;
     }
 
-    @JRubyMethod(name = "initialize", visibility = PRIVATE)
-    public IRubyObject initialize19(ThreadContext context, IRubyObject arg0) {
-        replace19(arg0);
+    @JRubyMethod(name = "initialize", visibility = PRIVATE, optional = 2)
+    public IRubyObject initialize19(ThreadContext context, IRubyObject[] args) {
+        final String[] validArgs = new String[]{"encoding"};
```
cheald: I'm not quite sure what Java will do here - since it's a final String[], will it be optimized away, or is this going to result in an allocation per invocation? Should this be moved to an array initialized outside of the method?

kares:
picked onto ruby-2.3 with some tunings and maintained `initialize` compatibility: https://github.com/jruby/jruby/commit/a3c648dca8052f959377db68d3d3e060b4b0a072 ... thanks.

PR #3551, opened 2015-12-19
[Truffle] Antlr printf

chrisseaton:


truffle/src/main/java/org/jruby/truffle/format/parser/PrintfTreeBuilder.java
```diff
@@ -0,0 +1,240 @@
...
+import org.jruby.util.ByteList;
+import org.jruby.util.StringSupport;
+
+import java.util.ArrayList;
+import java.util.List;
+
+public class PrintfTreeBuilder extends org.jruby.truffle.format.parser.PrintfParserBaseListener {
+
+    // TODO CS 19-Dec-15 this is never used, but the functionality seems useful
+    public static final int PADDING_FROM_ARGUMENT = -2;
```
eregon: It is used in Time#httpdate (from the stdlib). But unfortunately there are no spec to test it. Would be great to add one.

truffle/src/main/java/org/jruby/truffle/nodes/core/KernelNodes.java
```diff
@@ -1947,11 +1947,11 @@ public Long block() throws InterruptedException {
 
     @CoreMethod(names = { "format", "sprintf" }, isModuleFunction = true, rest = true, required = 1, taintFromParameter = 0)
     @ImportStatic(StringCachingGuards.class)
-    public abstract static class FormatNode extends CoreMethodArrayArgumentsNode {
+    public abstract static class XFormatNode extends CoreMethodArrayArgumentsNode {
```
eregon: Why the renaming?

eregon: I would call it `SprintfNode`

eregon:
Looks good, there is still quite a bit of logic with switches in the TreeBuilder but it seems hard to avoid.

truffle/src/main/java/org/jruby/truffle/format/parser/PrintfTreeBuilder.java
chrisseaton: Accidentally deleted a comment from Benoit here saying it's using in `Time.httpdate` and that there should be a spec for it.

truffle/src/main/java/org/jruby/truffle/nodes/core/KernelNodes.java
chrisseaton: The renaming was a debugging thing. I'll use sprintf.

truffle/src/main/java/org/jruby/truffle/format/parser/PrintfTreeBuilder.java
```diff
@@ -0,0 +1,240 @@
...
+            case 'o':
+            case 'u':
+            case 'x':
+            case 'X':
+                final PackNode spacePaddingNode;
+                if (spacePadding == PADDING_FROM_ARGUMENT) {
+                    spacePaddingNode = ReadIntegerNodeGen.create(context, new SourceNode());
+                } else {
+                    spacePaddingNode = new LiteralIntegerNode(context, spacePadding);
+                }
```
eregon: I hoped this would go away with the parsing if we had sth like
('*' => ReadPaddingFromArg | \d+ => LiteralPadding | '' NoPadding) but it might be hard.
Padding from arg should be supported for all formats.

truffle/src/main/antlr4/org/jruby/truffle/format/parser/PrintfLexer.g4
```diff
@@ -0,0 +1,28 @@
...
+ANGLE_KEY   : '<' .*? '>' ;
+ZERO        : '0' ;
+NUMBER      : [1-9] [0-9]* ;
+SPACE       : ' ' ;
+PLUS        : '+' ;
+MINUS       : '-' ;
+STAR        : '*' ;
+DOLLAR      : '$' ;
+DOT         : '.' ;
+CURLEY_KEY  : '{' .*? '}' -> mode(DEFAULT_MODE) ;
```
nirvdrum: I don't know if it's a British spelling, but I've always seen it as "curly" (no 'e').

PR #3553, opened 2015-12-19
[Ruby 2.3] Concurrency improvements to Queue/SizedQueue 

thedarkone:
Context:

 * discussion on https://gitter.im/ruby-concurrency/concurrent-ruby,
 * https://bugs.ruby-lang.org/issues/11822.

The implementation is now based on Doug Lea’s `LinkedBlockingQueue`.

Concurrency wise, this is now dual locked (head/tail) linked queue,
since Ruby std lib queues have blocking operations a completely
lock-free queue is not practical endeavor.


headius:
Very nice! I had similar code at various times in the past week and kept getting pulled away. I also based mine on Doug's `LinkedBlockingQueue`. Will merge your impl over.

Issue #3554, opened 2015-12-20
Use of Unsafe in High Scale break on Android


donv:
Using jruby-jars 1.7.23 with Ruboto results in this exception:

https://travis-ci.org/ruboto/ruboto/jobs/97978906#L4896 :

```
W/System.err(  870): Caused by: java.lang.RuntimeException: Could not obtain access to sun.misc.Unsafe
W/System.err(  870): 	at org.cliffc.high_scale_lib.UtilUnsafe.getUnsafe(UtilUnsafe.java:25)
W/System.err(  870): 	at org.cliffc.high_scale_lib.NonBlockingHashMapLong.<clinit>(NonBlockingHashMapLong.java:88)
W/System.err(  870): 	... 91 more
W/System.err(  870): Caused by: java.lang.NoSuchFieldException: theUnsafe
W/System.err(  870): 	at java.lang.Class.getDeclaredField(Class.java:631)
W/System.err(  870): 	at org.cliffc.high_scale_lib.UtilUnsafe.getUnsafe(UtilUnsafe.java:21)
```

JRuby uses Unsafe already, so I am not sure what the problem is with High Scale.  Any help would be great.

mkristian:
another high scale issue is #3555

kares:
seems to me that this pretty much nails that we're better off extracting the `NonBlockingHashMapLong` class into our own and managing the missing Unsafe. // cc @headius

kares:
one solution is to port over the needed class with adjustments: https://github.com/jruby/jruby/issues/3560 ... and avoid packaging the .jar (which also resolves packaging patched JDK classes - #3555)

donv:
@kares Thanks!

Issue #3394, opened 2015-10-13
JRuby thread stuck on IntHashMap.rehash


banker:
I have a multi-threaded application running on `JRuby 9.0.1.0 (Java 1.8.0_45-b14)`. The application runs with 8 Ruby threads. One of those threads has been stuck on the `IntHashMap.rehash` method for the last 13 hours. The remaining threads are all functioning normally. Here is (most of) the stack trace for the thread in question as reported by `jstack`:

```
"Ruby-0-Thread-7: /data/apps/my_app/current/vendor/bundle/jruby/2.2.0/gems/my-gem-java/lib/java_executor_simple.rb:22" #21 daemon prio=5 os_prio=0 tid=0x00007f1f585bf800 nid=0x505e runnable [0x00007f1f0db99000]
   java.lang.Thread.State: RUNNABLE
        at org.jruby.util.collections.IntHashMap.rehash(IntHashMap.java:139)
        at org.jruby.util.collections.IntHashMap.put(IntHashMap.java:170)
        at org.jruby.java.dispatch.CallableSelector.matchingCallableArityOne(CallableSelector.java:118)
        at org.jruby.java.invokers.RubyToJavaInvoker.findCallableArityOne(RubyToJavaInvoker.java:282)
        at org.jruby.java.invokers.InstanceMethodInvoker.call(InstanceMethodInvoker.java:98)
        at org.jruby.runtime.callsite.CachingCallSite.callBlock(CachingCallSite.java:139)
        at org.jruby.runtime.callsite.CachingCallSite.call(CachingCallSite.java:145)
        at data.apps.my_app.releases.my_app_minus_03751e0c_minus_1caa0727.lib.my_app.application.invokeOther31:time(/data/apps/my_app/releases/my_app-03751e0c-1caa0727/lib/my_app/application.rb)
        at data.apps.my_app.releases.my_app_minus_03751e0c_minus_1caa0727.lib.my_app.application.RUBY$method$md?$0(/data/apps/my_app/releases/my_app-03751e0c-1caa0727/lib/my_app/application.rb:149)
        at java.lang.invoke.LambdaForm$DMH/885284298.invokeStatic_L7_L(LambdaForm$DMH)
        at java.lang.invoke.LambdaForm$MH/2130192211.invokeExact_MT(LambdaForm$MH)
        at org.jruby.internal.runtime.methods.CompiledIRMethod.call(CompiledIRMethod.java:131)
[...]

PR #3560, opened 2015-12-22
port NonBlockingHashMapLong from high-scale-lib

### LINKED ISSUE #3555: jruby-jars jruby-core should not include java/util/concurrent/ConcurrentHashMap.class
Starting with the inclusion of High Scale in JRuby 1.7.23, the jruby-jars gem includes java/util/concurrent/ConcurrentHashMap.class in the jruby-core jar.

This causes a security exception when an application tries to load the class outside the bootloader.

jruby-jars should not contain any packages in the `java` packages.

kares:
the high-scale lib has been added as a dependency fairly recently on 9K as well as 1.7.23
it was introduced to use `NonBlockingHashMapLong` as an alternative of `IntHashMap` (due #3394)

turns out that the library can not work without `Unsafe` (and thus broke Android support in 1.7.x see #3554) while JRuby allows to boot and run when `sun.misc.Unsafe` is not available. thus the proposal here is to port the `NonBlockingHashMapLong` over (the com.boundary:hish-scale-lib received no commit in more than a year) and tune it to be usable without `Unsafe` (fixing #3554)

this also resolves #3555 - packaging issue as .jar includes patch(es) of JDK class(es)

### DIFF
--- a/core/pom.xml
+++ b/core/pom.xml
@@ -207,11 +207,6 @@
       <artifactId>options</artifactId>
       <version>1.3</version>
     </dependency>
-    <dependency>
-      <groupId>com.boundary</groupId>
-      <artifactId>high-scale-lib</artifactId>
-      <version>1.0.6</version>
-    </dependency>
   </dependencies>
   <properties>
     <anno.sources>${project.basedir}/target/generated-sources</anno.sources>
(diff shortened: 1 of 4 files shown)


kares:
if this looks like a good "resolution" please review the CAS fall-back semantics for arrays such as : https://github.com/jruby/jruby/commit/1624fd4d989a8c220a39bf5d804637085bb48c4d#diff-0658ae080e35e981bf3e13528f2730c1R1398 ... if there's anything to be done better

thedarkone:
I'm just reviewing `1624fd4`. IMO yes, your replacement is as correct as Cliff's original code.

PS: other potential reviewers: Cliff's cas isn't on the same volatile field that is used for read side sychronizes-with order.

headius:
Looks ok to me on a quick read-through, and I have faith in @thedarkone's judgment. Merging.

PR #3581, opened 2016-01-06
Cache proxies of newly constructed Java objects

### LINKED ISSUE #3576: ConstructorInvoker isn't storing the proxy in the ObjectProxyCache
We are defining singleton methods on Java objects that had been freshly constructed. These objects get stored on the Java side, and then later retrieved again from the JRuby side, only to have their singleton methods missing. We are holding on to the original proxy from the constructed objects, so this shouldn't be a case of a garbage collected proxy object.

I started tracing JRuby with my scenario and found the code path I was hitting in `ConstructorInvoker` wasn't actually caching the `JavaProxy` in the `ObjectProxyCache`, while if I had gone through an `InstanceMethodInvoker` path, it would have been cached.

Below is a simple Minitest test case that fails, demonstrating the problem.

```java
// Test.java
public class Test {
    private static Test lastInstance;

    public Test() {
        lastInstance = this;
    }

    public static Test getLastInstance() {
        return lastInstance;
    }
}
```

```ruby
# test.rb
gem "minitest"
require "minitest/autorun"
Java::Default::Test.__persistent__ = true

class TestCaching < Minitest::Test
  def test_constructed_proxy_is_cached
    t = Java::Default::Test.new
    def t.testing; 42; end
    assert_equal 42, t.testing
    assert_equal t, Java::Default::Test.last_instance
    assert_equal 42, Java::Default::Test.last_instance.testing
  end
end
```

The results of running on `1.7.22` and `9.0.4.0`:
```
$ rvm jruby-1.7.22 do ruby test.rb
Run options: --seed 12875

# Running:

E

Finished in 0.006000s, 166.6667 runs/s, 333.3333 assertions/s.

  1) Error:
TestCaching#test_constructed_proxy_is_cached:
NoMethodError: undefined method `testing' for #<Java::Default::Test:0x53fdffa1>
    test.rb:11:in `test_constructed_proxy_is_cached'

1 runs, 2 assertions, 0 failures, 1 errors, 0 skips
```

```
$ rvm jruby-9.0.4.0 do ruby test.rb
Run options: --seed 9484

# Running:

E

Finished in 0.009849s, 101.5313 runs/s, 203.0626 assertions/s.

  1) Error:
TestCaching#test_constructed_proxy_is_cached:
NoMethodError: undefined method `testing' for #<Java::Default::Test:0x1a78dacd>
    test.rb:11:in `test_constructed_proxy_is_cached'

1 runs, 2 assertions, 0 failures, 1 errors, 0 skips
```

Some additional notes:
* In our codebase this works on `1.7.22`, but this test fails in both (in our case, more is happening, so maybe that extra stuff is triggering a working path in `1.7.22` somehow)
* If you drop `Java::Default::Test.__persistent__ = true` from the test, it starts passing in `9.0.4.0`, but then if you create another test method that does the same thing, the second test fails (presumably the code triggering the warning about persistence is somehow getting the proxy object to be cached?)

smellsblue:
This fixes #3576

I added a few related specs while I was at it (only the constructor one failed before my changes though).

I noticed the warning at `spec/READ_ME_FIRST.txt`, but wasn't sure if that applied to the `java_integration` specs, so if I should have added my spec elsewhere, please let me know.

### DIFF
--- a/core/src/main/java/org/jruby/java/invokers/ConstructorInvoker.java
+++ b/core/src/main/java/org/jruby/java/invokers/ConstructorInvoker.java
@@ -55,7 +55,7 @@ public final class ConstructorInvoker extends RubyToJavaInvoker {
         JavaConstructor constructor = (JavaConstructor) findCallable(self, name, args, args.length);
 
         final Object[] convertedArgs = convertArguments(constructor, args);
-        proxy.setObject( constructor.newInstanceDirect(context, convertedArgs) );
+        setAndCacheProxyObject(context, proxy, constructor.newInstanceDirect(context, convertedArgs));
 
         return self;
     }
@@ -66,7 +66,7 @@ public final class ConstructorInvoker extends RubyToJavaInvoker {
         JavaProxy proxy = castJavaProxy(self);
         JavaConstructor constructor = (JavaConstructor) findCallableArityZero(self, name);
 
-        proxy.setObject(constructor.newInstanceDirect(context));
+        setAndCacheProxyObject(context, proxy, constructor.newInstanceDirect(context));
 
         return self;
     }
@@ -79,7 +79,7 @@ public final class ConstructorInvoker extends RubyToJavaInvoker {
         final Class<?>[] paramTypes = constructor.getParameterTypes();
         Object cArg0 = arg0.toJava(paramTypes[0]);
 
-        proxy.setObject(constructor.newInstanceDirect(context, cArg0));
+        setAndCacheProxyObject(context, proxy, constructor.newInstanceDirect(context, cArg0));
 
         return self;
     }
@@ -93,7 +93,7 @@ public final class ConstructorInvoker extends RubyToJavaInvoker {
         Object cArg0 = arg0.toJava(paramTypes[0]);
         Object cArg1 = arg1.toJava(paramTypes[1]);
 
-        proxy.setObject(constructor.newInstanceDirect(context, cArg0, cArg1));
+        setAndCacheProxyObject(context, proxy, constructor.newInstanceDirect(context, cArg0, cArg1));
 
         return self;
     }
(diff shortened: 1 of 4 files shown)


headius:
Nice! I'll review tonight.

kares:
:evergreen_tree: :+1:

headius:
Finally getting back to this now.

headius:
Ok, this isn't quite right. This will unconditionally cache objects coming from all constructor calls whether the proxy cache is on or the class has been set to persistent.

If you look at InstanceMethodInvoker and trace through method.invokeDirect, after a long chain of calls (@kares we need to reduce this) it eventually gets back to Java.getInstance() which checks those flags and decides to cache or not.

headius:
This might be ok if we modify setAndCacheProxyObject to check those same flags.

smellsblue:
@headius thanks for taking a look! Sorry I didn't look deeply enough into the InstanceMethodInvoker path!

I'll take another pass and see if I can fix these problems, along with some more tests.

smellsblue:
@headius I'm wondering about something like this:

```java
    private void setAndCacheProxyObject(ThreadContext context, JavaProxy proxy, Object object) {
        proxy.setObject(object);

        if (Java.OBJECT_PROXY_CACHE || Java.getProxyClassForObject(context.runtime, object).getCacheProxy()) {
            context.runtime.getJavaSupport().getObjectProxyCache().put(object, proxy);
        }
    }
```

However, I notice there is a `RubyModule clazz` parameter to all the `call` methods... is that already the proxy class I seek? I'll dig in a bit more, but if you know off-hand, that would be a big help :-)

kares:
right, did not realize. will fix it unless @smellsblue spins up another PR.

headius:
@smellsblue I believe that `clazz` is the class on which we're calling `new`, so you should be able to check for the class proxy directly.

smellsblue:
@kares @headius I just created a new PR to both fix the bug and add some specs around proxy caching: #3602

@headius using the `clazz` parameter seems to work, so I went with it... if you think there might be a scenario where the `clazz` isn't the right `RubyClass`, please let me know, though I may need some guidance on how to check if it is the right `RubyClass` before fetching it.

PR #3593, opened 2016-01-12
Test indy invokers

headius:
This change enables MethodHandle-based invokers for *all* native method binding, eliminating our pregenerated INVOKER classes and avoiding code generation at runtime for third-party extensions.

For my quick tests, it boots in about the same time or less time than the generated invokers and existing AnnotationBinder-created populators, but I'm also deferring a substantial amount of work until methods are first called. This could be smart, or it could be just delaying the inevitable...but so far it seems to balance out well.

See commit fc4bcb5 for a more detailed discussion of the changes. This PR should not be merged until it is passing tests and we have decided to commit to this direction.

### DIFF
--- a/core/pom.rb
+++ b/core/pom.rb
@@ -138,19 +138,6 @@ project 'JRuby Core' do
                      :id => 'add-populators',
                      'sources' => [ '${anno.sources}' ] )
     end
-
-    plugin 'org.codehaus.mojo:exec-maven-plugin' do
-      execute_goals( 'exec',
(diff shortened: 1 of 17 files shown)


headius:
I forgot to mention this change also results in a 1.6MB reduction in the size of jruby.jar.

eregon:
:+1: on anything simplifying the build :)

kares:
would be interesting to have this in at some 9.2 -> maybe starting as an optional feature with a --switch

PR #3602, opened 2016-01-18
Fix cache-constructed-proxies

smellsblue:
Replicate the logic from org.jruby.javasupport.Java which determines whether an object should be cached.

I added some specs to ensure the caching is properly honored. If I comment out the new conditional, you can see that one of the new specs will correctly fail:
```
$ jruby -S rspec spec/java_integration/object/singleton_spec.rb spec/java_integration/types/wrapping_spec.rb
......F........

Failures:

  1) A Java method returning/receiving uncoercible Java types with persistence off doesn't cache the proxy when directly constructed
     Failure/Error: expect(object_proxy_cache.get(object)).to be_nil

       expected: nil
            got: #<Java::JavaLang::Object:0x38c6f217>
     # ./spec/java_integration/types/wrapping_spec.rb:74:in `block in (root)'

Finished in 0.103 seconds (files took 0.916 seconds to load)
15 examples, 1 failure

Failed examples:

rspec ./spec/java_integration/types/wrapping_spec.rb:72 # A Java method returning/receiving uncoercible Java types with persistence off doesn't cache the proxy when directly constructed
```

I wasn't sure if this was the right place to put these new specs, but it seemed like the most appropriate to me. Please let me know if I should move them.

### DIFF
--- a/core/src/main/java/org/jruby/java/invokers/ConstructorInvoker.java
+++ b/core/src/main/java/org/jruby/java/invokers/ConstructorInvoker.java
@@ -7,6 +7,7 @@ import org.jruby.Ruby;
 import org.jruby.RubyModule;
 import org.jruby.RubyProc;
 import org.jruby.java.proxies.JavaProxy;
+import org.jruby.javasupport.Java;
 import org.jruby.javasupport.JavaCallable;
 import org.jruby.javasupport.JavaConstructor;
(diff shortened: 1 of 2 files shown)


kares:
:clap: job Mike, thanks // cc @headius this should be the left-over from #3581

headius:
Looks good, thank you for your work on this!

PR #3605, opened 2016-01-19
Truffle ropes

nirvdrum:
This is the initial work in changing the storage strategy for Strings in JRuby+Truffle from ByteList to Rope. There was a fair bit of rote translation work here. I haven't spend any appreciable amount of time on optimizing anything yet. And Symbols are still backed by ByteList.

I tried to make the code work well for both JRuby backends. For simplicity I did pull in some Truffle annotations, but it should be easy to divorce the two.

@chrisseaton @pitr-ch @eregon Please review at your leisure.

@enebo @headius @lopex This work may be of interest to you.


nirvdrum:
I should note that all feedback is appreciated. But please be open to the answer "I didn't get to that yet."

test/truffle/compiler/pe/core/string_pe.rb
```diff
@@ -12,4 +12,24 @@
...
 
 example "'abc'.length", 3
+example "'こにちわ'.length", 4
+
+example "'abc'.bytesize", 3
+example "'こにちわ'.bytesize", 12
+
 tagged_example "'abc' == 'abc'", true # seems to fail sometimes
+example "x = 'abc'; x == x", true
+example "x = 'abc'; x == x.dup", true
```
chrisseaton: This is where the optimisation magic starts! This dup has been completely eliminated by the PE.

truffle/src/main/java/org/jruby/truffle/nodes/literal/StringLiteralNode.java
```diff
@@ -9,41 +9,33 @@
...
 
     @Child private AllocateObjectNode allocateObjectNode;
 
     public StringLiteralNode(RubyContext context, SourceSection sourceSection, ByteList byteList, int codeRange) {
         super(context, sourceSection);
         assert byteList != null;
-        this.bytes = byteList.bytes();
-        this.encoding = byteList.getEncoding();
-        this.codeRange = codeRange;
+        this.rope = RopeOperations.create(byteList.bytes(), byteList.getEncoding(), codeRange);
```
chrisseaton: We should de-dup here or in the caller so that literals in different places by the same bytes will be the same `LeafRope` instance.

truffle/src/main/java/org/jruby/truffle/nodes/rubinius/PosixNodes.java
```diff
@@ -354,10 +350,7 @@ public SetenvNode(RubyContext context, SourceSection sourceSection) {
         @CompilerDirectives.TruffleBoundary
         @Specialization(guards = { "isRubyString(name)", "isRubyString(value)" })
         public int setenv(DynamicObject name, DynamicObject value, int overwrite) {
-            final String nameString = RubyEncoding.decodeUTF8(StringOperations.getByteList(name).getUnsafeBytes(), StringOperations.getByteList(name).getBegin(), StringOperations.getByteList(name).getRealSize());
-            final String valueString = RubyEncoding.decodeUTF8(StringOperations.getByteList(value).getUnsafeBytes(), StringOperations.getByteList(value).getBegin(), StringOperations.getByteList(value).getRealSize());
-
-            return posix().setenv(nameString, valueString, overwrite);
+            return posix().setenv(decodeUTF8(name), decodeUTF8(value), overwrite);
```
chrisseaton: Are we assuming something that we shouldn't be when we `decodeUTF8`?

truffle/src/main/java/org/jruby/truffle/runtime/rope/ConcatRope.java
```diff
@@ -0,0 +1,106 @@
...
+
+    @Override
+    public byte[] getBytes() {
+        if (bytes == null) {
+            final byte[] leftBytes = left.getBytes();
+            final byte[] rightBytes = right.getBytes();
+
+            bytes = new byte[leftBytes.length + rightBytes.length];
+            System.arraycopy(leftBytes, 0, bytes, 0, leftBytes.length);
+            System.arraycopy(rightBytes, 0, bytes, leftBytes.length, rightBytes.length);
```
chrisseaton: For deep trees this is going to create a lot of intermediate arrays. Can't we allocate an array of the right size for the final string, and then pass it down into child nodes and get them to fill it in, rather than creating a new array each time?

truffle/src/main/java/org/jruby/truffle/runtime/rope/ConcatRope.java
```diff
@@ -0,0 +1,106 @@
...
+
+                final byte[] ret = new byte[length];
+                System.arraycopy(leftBytes, 0, ret, 0, leftBytes.length);
+                System.arraycopy(rightBytes, 0, ret, leftBytes.length, rightBytes.length);
+
+                return ret;
+            }  else {
+                return leftBytes;
+            }
+        }
```
chrisseaton: Same as `getBytes` - can't we allocate the final array once?

truffle/src/main/ruby/core/rubinius/api/shims/string_mirror.rb
```diff
@@ -0,0 +1,19 @@
...
+# Eclipse Public License version 1.0
+# GNU General Public License version 2
+# GNU Lesser General Public License version 2
+
+module Rubinius
+  class Mirror
+    class String < Mirror
+
+      def splice(starting_byte_index, byte_count_to_replace, replacement)
+        Rubinius.invoke_primitive :string_splice, @object, replacement, starting_byte_index, byte_count_to_replace
```
chrisseaton: Did we write this from scratch?

truffle/src/main/java/org/jruby/truffle/nodes/rubinius/PosixNodes.java
nirvdrum: We largely ignore configured internal and external encodings throughout. We need to fix that at some point.

truffle/src/main/java/org/jruby/truffle/runtime/rope/ConcatRope.java
nirvdrum: I've done very little with the byte[] thus far. Normally with ropes you'd just recursively walk the children, but that's not very PE friendly. I just did the dumbest thing to make it work for the time being.

truffle/src/main/java/org/jruby/truffle/runtime/rope/ConcatRope.java
nirvdrum: Possibly. But the intermediate ropes may very well be used elsewhere and thus require their own byte[] anyway. I don't think we want to get into children maintaining a reference to a parent allocated byte[].

truffle/src/main/ruby/core/rubinius/api/shims/string_mirror.rb
nirvdrum: Yes. It's a complete replacement for https://github.com/jruby/jruby/blob/cdd0eb583c7d30911b6f215238ebf62a8dfda608/truffle/src/main/ruby/core/rubinius/common/string_mirror.rb#L60-L91

nirvdrum: To clarify, I added a new primitive that Rubinius doesn't have. But I grouped it with the rest of the String primitives to keep things simple. Technically it should be a Truffle primitive. If you feel strongly about that, I can move it.

truffle/src/main/java/org/jruby/truffle/runtime/rope/ConcatRope.java
chrisseaton: Right I didn't think that the intermediate arrays are stored.

truffle/src/main/java/org/jruby/truffle/runtime/rope/ConcatRope.java
chrisseaton: Yeah ignore.

truffle/src/main/ruby/core/rubinius/api/shims/string_mirror.rb
chrisseaton: Should be fine.

chrisseaton:
Looks good to merge to me.

test/truffle/compiler/pe/core/string_pe.rb
eregon: How does it work? The rope is only shallow-copied? Then ropes are purely immutable?

truffle/src/main/java/org/jruby/truffle/nodes/core/InterpolatedStringNode.java
```diff
@@ -74,13 +74,14 @@ private Object concat(VirtualFrame frame, Object[] strings) {
...
 
+        // TODO (nirvdrum 11-Jan-16) Rewrite to avoid massively unbalanced trees.
         for (Object string : strings) {
             assert RubyGuards.isRubyString(string);
 
             if (builder == null) {
                 builder = dupNode.call(frame, string, "dup", null);
             } else {
-                builder = concatNode.call(frame, builder, "concat", null, string);
+                builder = appendNode.call(frame, builder, "append", null, string);
```
eregon: What is `append`?

truffle/src/main/java/org/jruby/truffle/nodes/core/StringGuards.java
```diff
@@ -22,42 +23,49 @@
...
-        return StringOperations.getByteList(string).getEncoding().isSingleByte();
+        return Layouts.STRING.getRope(string).getEncoding().isSingleByte();
     }
 
     public static boolean isValidOr7BitEncoding(DynamicObject string) {
         assert RubyGuards.isRubyString(string);
-        return StringOperations.isCodeRangeValid(string) || CodeRangeSupport.isCodeRangeAsciiOnly(StringOperations.getCodeRangeable(string));
+        final Rope rope = StringOperations.rope(string);
+
+        return (rope.getCodeRange() == StringSupport.CR_VALID) || (rope.getCodeRange() == StringSupport.CR_7BIT);
```
eregon: This could be a `&` if the bits are well chosen (or some boolean in the enum).

truffle/src/main/java/org/jruby/truffle/nodes/rubinius/StringPrimitiveNodes.java
```diff
@@ -125,16 +136,45 @@ public CharacterPrintablePrimitiveNode(RubyContext context, SourceSection source
...
-            final int codepoint = encoding.mbcToCode(bytes.unsafeBytes(), bytes.begin(), bytes.begin() + bytes.realSize());
+            final int codepoint = encoding.mbcToCode(rope.getBytes(), 0, rope.byteLength());
 
             return encoding.isPrint(codepoint);
         }
 
     }
 
+    @RubiniusPrimitive(name = "string_append")
+    public static abstract class StringAppendPrimitiveNode extends RubiniusPrimitiveNode {
```
eregon: What's the difference with `String#concat` ?

test/truffle/compiler/pe/core/string_pe.rb
chrisseaton: Yes, so `dup` allocates a tiny new wrapper around the same rope node and then the `==` is just a reference comparison in the fast path, which of course all PEs really well.

truffle/src/main/java/org/jruby/truffle/nodes/core/InterpolatedStringNode.java
nirvdrum: A Rubinius method/primitive. It does much of the same without some of the extra checking that `concat` needs to perform.

truffle/src/main/java/org/jruby/truffle/nodes/rubinius/StringPrimitiveNodes.java
nirvdrum: `String#concat` needs to deal with `Fixnum` values as well. Otherwise, they're essentially the same.

truffle/src/main/java/org/jruby/truffle/nodes/core/InterpolatedStringNode.java
eregon: But this means it's a public Ruby method. If it's not needed as such, it could be just a normal node.

truffle/src/main/java/org/jruby/truffle/runtime/layouts/StringLayout.java
```diff
@@ -24,19 +25,15 @@ DynamicObjectFactory createStringShape(DynamicObject logicalClass,
...
     boolean isString(DynamicObject dynamicObject);
     boolean isString(Object dynamicObject);
 
-    ByteList getByteList(DynamicObject object);
-    void setByteList(DynamicObject object, ByteList byteList);
-
-    int getCodeRange(DynamicObject object);
-    void setCodeRange(DynamicObject object, int codeRange);
+    Rope getRope(DynamicObject object);
+    void setRope(DynamicObject object, Rope byteList);
```
eregon: `Rope rope` here

truffle/src/main/java/org/jruby/truffle/runtime/rope/ConcatRope.java
```diff
@@ -0,0 +1,106 @@
...
+            } else {
+                leftBytes = left.extractRange(offset, length);
+            }
+
+            if (leftBytes.length < length) {
+                rightBytes = right.extractRange(0, length - leftBytes.length);
+
+                final byte[] ret = new byte[length];
+                System.arraycopy(leftBytes, 0, ret, 0, leftBytes.length);
+                System.arraycopy(rightBytes, 0, ret, leftBytes.length, rightBytes.length);
```
eregon: Maybe worth an helper in ArrayUtils for concatenating two arrays, the signature of arraycopy is not the most readable.

truffle/src/main/java/org/jruby/truffle/runtime/rope/ConcatRope.java
```diff
@@ -0,0 +1,106 @@
...
+            return true;
+        }
+
+        if (o instanceof ConcatRope) {
+            final ConcatRope other = (ConcatRope) o;
+
+            return left.equals(other.left) && right.equals(other.right);
+        }
+
+        return false;
```
eregon: Is this an speculative equals? If not, why does the case of having a leaf rope as `o` does not need to be considered?

truffle/src/main/java/org/jruby/truffle/nodes/core/InterpolatedStringNode.java
nirvdrum: Rubinius calls this method. We had it shimmed previously (https://github.com/jruby/jruby/blob/18866ad42627363aa331cf27d53896d6d7fa359b/truffle/src/main/ruby/core/shims.rb#L68-L72). I just implemented it as a primitive as Rubinius does. The new method lives in `bootstrap/string.rb`. This was necessary because I moved a lot of `String#concat` out to Rubinus, which uses `String#append` to implement the method. Our old shim would have been cyclical.

I could indeed just call the method rather than go through the full dispatch here. I've been avoiding that generally in case we need to run through taint or frozen checks.

truffle/src/main/java/org/jruby/truffle/runtime/rope/LeafRope.java
```diff
@@ -0,0 +1,66 @@
...
+    @Override
+    public byte[] getBytes() {
+        return bytes;
+    }
+
+    @Override
+    public byte[] extractRange(int offset, int length) {
+        assert offset + length <= bytes.length;
+
+        final int trueLength = Math.min(length, byteLength());
```
eregon: The min seems unneeded if the assert above holds.

truffle/src/main/java/org/jruby/truffle/runtime/layouts/StringLayout.java
nirvdrum: Good catch. Thanks.

nirvdrum: I'll grep for similar issues.

truffle/src/main/java/org/jruby/truffle/runtime/rope/LeafRope.java
```diff
@@ -0,0 +1,66 @@
...
+ */
+package org.jruby.truffle.runtime.rope;
+
+import org.jcodings.Encoding;
+
+import java.util.Arrays;
+
+public abstract class LeafRope extends Rope {
+
+    private final byte[] bytes;
```
eregon: Should be @CompilationFinal and not just final (so it indicates very byte is final as well).

truffle/src/main/java/org/jruby/truffle/runtime/rope/ConcatRope.java
nirvdrum: `equals` and `hashCode` in all the ropes is currently broken. I was trying to be clever initially and found too many cases where it falls apart. E.g., `'ab' + 'cd'` and `'a' + 'bcd'` will both be `ConcatRope` that should logically be equivalent, but currently fail because they don't have the same tree structure.

nirvdrum: That is to say, it's on my backlog of things to clean up.

truffle/src/main/java/org/jruby/truffle/runtime/rope/LeafRope.java
nirvdrum: Old code. The caller is now responsible for the bounds checking. I'll clean up.

truffle/src/main/java/org/jruby/truffle/runtime/rope/LeafRope.java
nirvdrum: I don't think it matters much currently because it's invoked through a polymorphic method and that seems to be an implicit boundary to Graal. I need to lift the field into the base `Rope` class, along with the accessor method.

I did try that and ran into some problems. I think it was because some ropes still lazily populate the byte[]. It's on my list to revisit this as part of working out whether to share the byte[] or eagerly allocate it. In either case, I want it to be a final field for all ropes and then I'll mark it `@CompilationFinal`.

truffle/src/main/java/org/jruby/truffle/nodes/core/InterpolatedStringNode.java
eregon: I would go for the having an AppendNode and using `executeStringAppend` here, it's much lighter weight.

nirvdrum: I'm going to defer this until after the merge. It's no worse off than what we had before and this PR is already getting rather long.

truffle/src/main/java/org/jruby/truffle/runtime/rope/RopeOperations.java
```diff
@@ -0,0 +1,146 @@
...
+            case StringSupport.CR_VALID: return new ValidLeafRope(bytes, encoding, characterLength);
+            case StringSupport.CR_BROKEN: return new InvalidLeafRope(bytes, encoding);
+            default: {
+                CompilerDirectives.transferToInterpreter();
+                throw new RuntimeException(String.format("Unknown code range type: %d", codeRange));
+            }
+        }
+    }
+
+    @CompilerDirectives.TruffleBoundary
```
pitr-ch: Didn't we make a convention to use just `@TruffleBoundary` ?

nirvdrum: Probably. I'll fix them. It's just how IntelliJ autocompletes.

truffle/src/main/java/org/jruby/truffle/runtime/rope/RopeOperations.java
```diff
@@ -0,0 +1,146 @@
...
+    public static Rope substring(Rope base, int offset, int byteLength) {
+        if (byteLength == 0) {
+            return withEncoding(EMPTY_UTF8_ROPE, base.getEncoding());
+        }
+
+        if (byteLength - offset == base.byteLength()) {
+            return base;
+        }
+
+        if (base instanceof SubstringRope) {
```
pitr-ch: Could this method be a node to cache the types and to avoid `instanceof`?

nirvdrum: Yes, but we also need to be able to create ropes outside the context of Ruby execution. I agree the `instanceof` are bad here. I plan on tidying this up as follow-on work.

pitr-ch:
:+1:

truffle/src/main/java/org/jruby/truffle/runtime/rope/RopeOperations.java
eregon: (FWIW instanceof on a leaf type is not more expensive than a field read & compare, but of course in a Node we could run the guards only for the cases we observe)

nirvdrum: Good to know. I think we're likely to see all cases except in extreme cases. But different call sites may only ever see a subset of patterns. So, I think the value comes with node splitting.

PR #3614, opened 2016-01-25
Support ExecJS in Truffle using Graal.js

chrisseaton:
You use it by requiring `truffle/execjs` after `execjs`. It uses Graal.js to run high-performance JavaScript, including with inlining between Ruby and JavaScript and back again, escape analysis between the two languages, etc etc. The same technology we used to do C extensions.

You will need `trufflejs.jar` from the GraalVM (https://github.com/jruby/jruby/wiki/Downloading-GraalVM)

For example:

```ruby
require "execjs"
require "truffle/execjs"

source = File.read('test/truffle/integration/execjs/coffeescript.js')
context = ExecJS.compile(source)
puts context.call("CoffeeScript.compile", "square = (x) -> x * x", bare: true)
```

```
bin/jruby -X-T bin/gem install execjs -v 2.6.0
JRUBY_OPTS='-J-cp trufflejs.jar' bin/jruby -X+T -Ilib/ruby/gems/shared/gems/execjs-2.6.0/lib test.rb
```


test/truffle/integration/execjs/coffeescript.rb
```diff
@@ -0,0 +1,32 @@
...
+# EXPRESS OR IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF
+# MERCHANTABILITY, FITNESS FOR A PARTICULAR PURPOSE AND
+# NONINFRINGEMENT. IN NO EVENT SHALL THE AUTHORS OR COPYRIGHT HOLDERS BE
+# LIABLE FOR ANY CLAIM, DAMAGES OR OTHER LIABILITY, WHETHER IN AN ACTION
+# OF CONTRACT, TORT OR OTHERWISE, ARISING FROM, OUT OF OR IN CONNECTION
+# WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE SOFTWARE.
+
+require "execjs"
+require "truffle/execjs"
+#require "open-uri"
```
nirvdrum: I was confused by this at first. I'd maybe just add a note above indicating you made changes to prevent a test from accessing a public URL.

nirvdrum:
This is awesome.

test/truffle/integration/execjs/coffeescript.rb
chrisseaton: I actually did it because we fail are inflating the gzipped response the URL gives us :)

PR #3615, opened 2016-01-26
[Truffle] Minimal Java POSIX

chrisseaton:
When we adopted Rubinius' kernel, we lost the ability to run without a real native implementation of `POSIX`. This is some work towards getting that ability back, by using knowledge of what Rubinius really tries to do with POSIX to provide a Java emulation that will work for us.

This PR gets us as far as loading, but not doing `puts`.

@nirvdrum


truffle/src/main/java/org/jruby/truffle/runtime/TruffleJavaPOSIX.java
```diff
@@ -0,0 +1,49 @@
...
+                    return 0;
+            }
+        }
+
+        return super.fcntlInt(fd, fcntlConst, arg);
+    }
+
+    @Override
+    public int getpid() {
+        return context.hashCode();
```
nirvdrum: This looks odd to me.

nirvdrum:
What does this gain us over just using the Java layer already in jnr-posix? Just faster turnaround?

chrisseaton:
The existing `JavaPOSIX` doesn't support calls like `fcntl`, but given we know what Rubinius is calling them for and what results will work for it, we can give them a meaningful implementation.

Basically we can't boot with `JavaPOSIX` but we can boot with this.

truffle/src/main/java/org/jruby/truffle/runtime/TruffleJavaPOSIX.java
chrisseaton: This is how JRuby does it. There doesn't appear to be any other way to get the PID, and at least the may deconflict between multiple contexts in the same VM.

headius:
Can we add what you need to JavaPOSIX? Do we really need to make another
POSIX abstraction?

chrisseaton:
I'm not sure all the changes I'll make will be safe for general use - they might be quite tied to what the Rubinius code runs. When I get everything I need running I'll reconsider and put stuff back into jar-posix where I can.

eregon:
So far it looks like most of this stuff (except probably getpid but then we could special-case that one) would be fine for JavaPOSIX, it's already doing this kind of "best-effort" implementation from what I recall.

truffle/src/main/java/org/jruby/truffle/runtime/TruffleJavaPOSIX.java
kares: there's an ugly way to read and parse from a management bean ... assuming that is fine to pull in jmx

chrisseaton: Yes thanks, but the point of this little project is to be totally pure Java and I think that bean name is implementation defined. We'll only use this in very limited cases - we normally require a native POSIX and won't run with the fallback.

PR #3619, opened 2016-01-27
Truffle ropes 2

nirvdrum:
@chrisseaton @pitr-ch @eregon This is the additional work I've done on ropes since the last pull request. I did it as a separate pull request against the branch you've already reviewed so you can see the new changes more easily.

The work here was primarily to improve performance over the original branch. The first one was basically just functionally complete. This PR has a larger emphasis on doing things more efficiently with ropes. There's still a lot more we can do, particularly in the area of regular expressions.

Notable highlights:

* Introduced a rope table for StringLiteralNodes (NB: There's a problem with the key being removed as it's a WeakHashMap, but everything still works if the cached entry isn't available and these are all built during parse, so the effect of GC is minimized).
* Symbols are now based ropes.
* Rope construction methods have been converted to nodes.
* Some primitives to help in debugging ropes were added.
* Fixed `Rope#hashCode` and `Rope#equals`.
* Replaced recursive byte[] building with an iterative algorithm.
* Removed a bunch of places where we were constructing ByteLists (and subsequently materializing the rope's byte[]) unnecessarily.

### DIFF
--- a/truffle/src/main/java/org/jruby/truffle/nodes/StringCachingGuards.java
+++ b/truffle/src/main/java/org/jruby/truffle/nodes/StringCachingGuards.java
@@ -13,33 +13,37 @@ package org.jruby.truffle.nodes;
 import com.oracle.truffle.api.object.DynamicObject;
 import org.jruby.truffle.runtime.core.StringOperations;
 import org.jruby.truffle.runtime.layouts.Layouts;
+import org.jruby.truffle.runtime.rope.Rope;
 import org.jruby.util.ByteList;
 
 public abstract class StringCachingGuards {
 
-    public static ByteList privatizeByteList(DynamicObject string) {
+    public static Rope privatizeRope(DynamicObject string) {
         if (RubyGuards.isRubyString(string)) {
-            return StringOperations.getByteListReadOnly(string).dup();
+            // TODO (nirvdrum 25-Jan-16) Should we flatten the rope to avoid caching a potentially deep rope tree?
+            return StringOperations.rope(string);
         } else {
             return null;
         }
     }
 
-    public static boolean byteListsEqual(DynamicObject string, ByteList byteList) {
+    public static boolean ropesEqual(DynamicObject string, Rope rope) {
         if (RubyGuards.isRubyString(string)) {
+            final Rope stringRope = StringOperations.rope(string);
+
             // equal below does not check encoding
-            if (Layouts.STRING.getRope(string).getEncoding() != byteList.getEncoding()) {
+            if (stringRope.getEncoding() != rope.getEncoding()) {
                 return false;
             }
-            // TODO CS 8-Nov-15 this code goes off into the woods - need to break it apart and branch profile it
-            return StringOperations.getByteListReadOnly(string).equal(byteList);
+
+            return stringRope.equals(rope);
         } else {
             return false;
         }
     }
 
-    public static int byteListLength(ByteList byteList) {
-        return byteList.length();
+    public static int ropeLength(Rope rope) {
+        return rope.byteLength();
     }
 
 }
--- a/truffle/src/main/java/org/jruby/truffle/nodes/core/RegexpNodes.java
+++ b/truffle/src/main/java/org/jruby/truffle/nodes/core/RegexpNodes.java
@@ -40,6 +40,7 @@ import org.jruby.truffle.runtime.RubyContext;
 import org.jruby.truffle.runtime.core.StringOperations;
 import org.jruby.truffle.runtime.layouts.Layouts;
 import org.jruby.truffle.runtime.rope.Rope;
+import org.jruby.truffle.runtime.rope.RopeOperations;
 import org.jruby.util.*;
 
 import java.nio.charset.StandardCharsets;
@@ -55,17 +56,17 @@ public abstract class RegexpNodes {
         assert RubyGuards.isRubyRegexp(regexp);
         assert RubyGuards.isRubyString(source);
 
-        final ByteList sourceByteList = StringOperations.getByteListReadOnly(source);
+        final Rope sourceRope = StringOperations.rope(source);
 
-        final ByteList bl = Layouts.REGEXP.getSource(regexp);
-        final Encoding enc = checkEncoding(regexp, StringOperations.getCodeRangeableReadOnly(source), true);
-        final ByteList preprocessed = RegexpSupport.preprocess(context.getRuntime(), bl, enc, new Encoding[] { null }, RegexpSupport.ErrorMode.RAISE);
+        final Rope regexpSourceRope = Layouts.REGEXP.getSource(regexp);
+        final Encoding enc = checkEncoding(regexp, sourceRope, true);
+        final ByteList preprocessed = RegexpSupport.preprocess(context.getRuntime(), regexpSourceRope.getUnsafeByteList(), enc, new Encoding[] { null }, RegexpSupport.ErrorMode.RAISE);
 
-        final Regex r = new Regex(preprocessed.getUnsafeBytes(), preprocessed.getBegin(), preprocessed.getBegin() + preprocessed.getRealSize(), Layouts.REGEXP.getOptions(regexp).toJoniOptions(), checkEncoding(regexp, StringOperations.getCodeRangeableReadOnly(source), true));
-        final Matcher matcher = r.matcher(sourceByteList.unsafeBytes(), sourceByteList.begin(), sourceByteList.begin() + sourceByteList.realSize());
-        int range = sourceByteList.begin() + sourceByteList.realSize();
+        final Regex r = new Regex(preprocessed.getUnsafeBytes(), preprocessed.getBegin(), preprocessed.getBegin() + preprocessed.getRealSize(), Layouts.REGEXP.getOptions(regexp).toJoniOptions(), checkEncoding(regexp, sourceRope, true));
+        final Matcher matcher = r.matcher(sourceRope.getBytes(), sourceRope.begin(), sourceRope.begin() + sourceRope.realSize());
+        int range = sourceRope.begin() + sourceRope.realSize();
 
-        return matchCommon(context, regexp, source, operator, setNamedCaptures, matcher, sourceByteList.begin() + startPos, range);
+        return matchCommon(context, regexp, source, operator, setNamedCaptures, matcher, sourceRope.begin() + startPos, range);
     }
 
     @TruffleBoundary
@@ -197,7 +198,7 @@ public abstract class RegexpNodes {
         }
     }
 
-    public static ByteList shimModifiers(ByteList bytes) {
+    public static Rope shimModifiers(Rope bytes) {
         // Joni doesn't support (?u) etc but we can shim some common cases
 
         String bytesString = bytes.toString();
@@ -223,14 +224,15 @@ public abstract class RegexpNodes {
                     throw new UnsupportedOperationException();
             }
 
-            bytes = ByteList.create(bytesString);
+            // TODO (nirvdrum 25-Jan-16): We probably just want a way to create a Rope from a java.lang.String.
+            bytes = StringOperations.ropeFromByteList(ByteList.create(bytesString));
         }
 
         return bytes;
     }
 
     @TruffleBoundary
-    public static Regex compile(Node currentNode, RubyContext context, ByteList bytes, RegexpOptions options) {
+    public static Regex compile(Node currentNode, RubyContext context, Rope bytes, RegexpOptions options) {
         bytes = shimModifiers(bytes);
 
         try {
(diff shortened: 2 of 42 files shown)


truffle/src/main/java/org/jruby/truffle/nodes/StringCachingGuards.java
```diff
@@ -13,33 +13,37 @@
...
+import org.jruby.truffle.runtime.rope.Rope;
 import org.jruby.util.ByteList;
 
 public abstract class StringCachingGuards {
 
-    public static ByteList privatizeByteList(DynamicObject string) {
+    public static Rope privatizeRope(DynamicObject string) {
         if (RubyGuards.isRubyString(string)) {
-            return StringOperations.getByteListReadOnly(string).dup();
+            // TODO (nirvdrum 25-Jan-16) Should we flatten the rope to avoid caching a potentially deep rope tree?
```
eregon: But then you would lose Rope identity comparison, it's a trade-off.

truffle/src/main/java/org/jruby/truffle/nodes/core/RegexpNodes.java
```diff
@@ -54,17 +56,17 @@ public static Object matchCommon(RubyContext context, DynamicObject regexp, Dyna
...
 
-        final ByteList sourceByteList = StringOperations.getByteListReadOnly(source);
+        final Rope sourceRope = StringOperations.rope(source);
 
-        final ByteList bl = Layouts.REGEXP.getSource(regexp);
-        final Encoding enc = checkEncoding(regexp, StringOperations.getCodeRangeableReadOnly(source), true);
-        final ByteList preprocessed = RegexpSupport.preprocess(context.getRuntime(), bl, enc, new Encoding[] { null }, RegexpSupport.ErrorMode.RAISE);
+        final Rope regexpSourceRope = Layouts.REGEXP.getSource(regexp);
+        final Encoding enc = checkEncoding(regexp, sourceRope, true);
+        final ByteList preprocessed = RegexpSupport.preprocess(context.getRuntime(), regexpSourceRope.getUnsafeByteList(), enc, new Encoding[] { null }, RegexpSupport.ErrorMode.RAISE);
```
eregon: I like the `getReadOnlyByteList()` terminology, it might be more coherent with the helper as well.

truffle/src/main/java/org/jruby/truffle/nodes/core/RegexpNodes.java
```diff
@@ -335,31 +337,42 @@ public static Encoding checkEncoding(DynamicObject regexp, CodeRangeable str, bo
...
-        setSource(regexp, setSource);
-        setOptions(regexp, RegexpOptions.fromEmbeddedOptions(options));
-        setRegex(regexp, compile(currentNode, context, setSource, Layouts.REGEXP.getOptions(regexp)));
+        final RegexpOptions regexpOptions = RegexpOptions.fromEmbeddedOptions(options);
+        final Regex regex = compile(currentNode, context, setSource, regexpOptions);
+
+        // The RegexpNodes.compile operation may modify the encoding of the source rope. This modified copy is stored
+        // in the Regex object as the "user object". Since ropes are immutable, we need to take this updated copy when
+        // constructing the final regexp.
+        setSource(regexp, (Rope) regex.getUserObject());
```
eregon: Do we need to keep track of the Rope explicitly then? Or could we just get it from the Regex?

truffle/src/main/java/org/jruby/truffle/nodes/core/RopeNodes.java
```diff
@@ -0,0 +1,438 @@
...
+        }
+
+        @Specialization(guards = { "!left.isEmpty()", "!right.isEmpty()", "left.byteLength() < SHORT_LEAF_BYTESIZE_THRESHOLD", "right.byteLength() < SHORT_LEAF_BYTESIZE_THRESHOLD" })
+        public Rope concatLeaves(LeafRope left, LeafRope right, Encoding encoding,
+                                 @Cached("createBinaryProfile()") ConditionProfile sameCodeRangeProfile,
+                                 @Cached("createBinaryProfile()") ConditionProfile brokenCodeRangeProfile) {
+            if (makeLeafRopeNode == null) {
+                CompilerDirectives.transferToInterpreter();
+                makeLeafRopeNode = insert(RopeNodesFactory.MakeLeafRopeNodeGen.create(getContext(), getSourceSection(), null, null, null));
+            }
```
eregon: That node could be in `@Cached` as well to make it more terse and avoid having a field for all instances of the node.

truffle/src/main/java/org/jruby/truffle/nodes/rubinius/ByteArrayNodes.java
```diff
@@ -40,8 +42,21 @@ public GetByteNode(RubyContext context, SourceSection sourceSection) {
...
+
+            // Handling out-of-bounds issues like this is non-standard. In Rubinius, it would raise an exception instead.
+            // We're modifying the semantics to address a primary use case for this class: Rubinius's @data array
+            // in the String class. Rubinius Strings are NULL-terminated and their code working with Strings takes
+            // advantage of that fact. So, where they expect to receive a NULL byte, we'd be out-of-bounds and raise
+            // an exception. Simply appending a NULL byte may trigger a full copy of the original byte[], which we
+            // want to avoid. The compromise is bending on the semantics here.
+            if (nullByteIndexProfile.profile(index == byteList.realSize())) {
+                return 0;
+            }
```
eregon: Sounds fair enough.

truffle/src/main/java/org/jruby/truffle/runtime/rope/Rope.java
```diff
@@ -93,4 +106,34 @@ public int realSize() {
...
+        return hashCode;
+    }
+
+    @Override
+    public boolean equals(Object o) {
+        if (this == o) {
+            return true;
+        }
+
+        if (o instanceof Rope) {
```
eregon: Could it be anything else?

truffle/src/main/java/org/jruby/truffle/runtime/rope/Rope.java
```diff
@@ -93,4 +106,34 @@ public int realSize() {
...
+            final Rope other = (Rope) o;
+
+            if ((hashCode != 0) && (other.hashCode != 0) && (hashCode != other.hashCode)) {
+                return false;
+            }
+
+            // TODO (nirvdrum 21-Jan-16): We really should be taking the encoding into account here. We're currenly not because it breaks the symbol table.
```
eregon: How does it break it?

truffle/src/main/java/org/jruby/truffle/runtime/rope/RopeOperations.java
```diff
@@ -144,4 +131,216 @@ public static int strLength(Encoding enc, byte[] bytes, int p, int end) {
...
+     * Performs an iterative depth first search of the Rope tree to calculate its byte[] without needing to populate
+     * the byte[] for each level beneath. Every LeafRope has its byte[] populated by definition. The goal is to determine
+     * which descendant LeafRopes contribute bytes to the top-most Rope's logical byte[] and how many bytes they should
+     * contribute. Then each such LeafRope copies the appropriate range of bytes to a shared byte[].
+     *
+     * Rope trees can be very deep. An iterative algorithm is preferable to recursion because it removes the overhead
+     * of stack frame management. Additionally, a recursive algorithm will eventually overflow the stack if the Rope
+     * tree is too deep.
+     */
+    public static byte[] flattenBytes(Rope rope) {
```
eregon: This seems quite complex. So in essence it's a DFS walk on the rope?
Can't we just call some SubstringRope.copy(buffer, startIndex)? That could recurse a bit but anyway the depth is bounded, isn't it?

eregon:
I read through the diff quickly, looks fine :smile:

truffle/src/main/java/org/jruby/truffle/nodes/StringCachingGuards.java
nirvdrum: Indeed. Although we're still doing a full byte[] comparison if the ropes aren't identity equal. We can eliminate that, but I was just maintaining existing functionality for now. The flattening could also work in conjunction with a lookup table to ensure the same references, as well.

truffle/src/main/java/org/jruby/truffle/nodes/core/RegexpNodes.java
nirvdrum: Ideally I want to pull this out of `Rope` entirely, I think. But I've made a note :-)

truffle/src/main/java/org/jruby/truffle/nodes/core/RegexpNodes.java
nirvdrum: I think by going through the user object we lose any potential performance benefits from the OM. I'd rather not rely on the user object field, but the method in question had side effects and this was the easiest way to get around that. The Regexp code in general needs a much more thorough pass.

truffle/src/main/java/org/jruby/truffle/nodes/core/RopeNodes.java
nirvdrum: Ahh. I keep forgetting about that.Thanks!

truffle/src/main/java/org/jruby/truffle/runtime/rope/Rope.java
nirvdrum: I suppose if someone were to screw up. In normal operation, it shouldn't be.

truffle/src/main/java/org/jruby/truffle/runtime/rope/Rope.java
nirvdrum: We were relying on `ByteList#equals` to not take the encoding into account. We currently expect two Symbols with different encodings but the same bytes to be the same Symbol. This is not what MRI.

As a simple example:

```
:abc.encoding => Encoding:US-ASCII
"abc".to_sym.encoding => Encoding:US-ASCII
"def".to_sym.encoding => Encoding:UTF-8
:def.encoding => Encoding:UTF-8
```

How a Symbol ends up in the table in the first place ends up affecting the encoding for future operations that look it up. When I first attempted to fix this, we ended up with duplicate entries (one for UTF-8, one for US-ASCII) and it broke a lot of caches. That fix needs to be combined with something else that converts UTF-8 to US-ASCII on the `String#to_sym` path. I'm not entirely clear on the semantics of maintaining encodings from Strings to Symbols.

truffle/src/main/java/org/jruby/truffle/runtime/rope/RopeOperations.java
nirvdrum: The recursive solution was a lot slower and can never PE. I think in reality this is too complex for PE anyway, but the option is available. But, currently ropes aren't bounded in depth. I'm still trying to find a good heuristic for limiting depth, but the ones I've tried thus far have mostly resulted in large performance regressions. It may be the case that it was a combination of the slower recursive function and the increased call frequency. But I was generally finding things were a lot faster and cheaper with memory if I let the Rope get arbitrarily long and took a single pass across the tree once the bytes were actually needed.

chrisseaton:
Looks good to me. Can we have a final PR for all changes and a merge commit so it's clear when it went it?

nirvdrum:
Yeah. This is already rebased on `truffle-ropes`. I can just close both PRs and open this one against master.

truffle/src/main/java/org/jruby/truffle/nodes/core/RegexpNodes.java
eregon: Regexp#source does not seem perf critical, and an additional read is nothing. I think the gain comes from not having to maintain a double mapping here (and that API is likely stable and intended for this purpose).

truffle/src/main/java/org/jruby/truffle/runtime/rope/Rope.java
eregon: I like fail fast on error-like values, but it's just a preference.

truffle/src/main/java/org/jruby/truffle/runtime/rope/Rope.java
eregon: Waow, pretty strange results!
Looks like a CR scan for literal symbol if 7-bit => ASCII would work intuitively.
`String` also == strings of different encodings, but not sure it's using this method.

truffle/src/main/java/org/jruby/truffle/runtime/rope/RopeOperations.java
eregon: I see, sounds fine then.

truffle/src/main/java/org/jruby/truffle/runtime/rope/Rope.java
nirvdrum: `String#==` is an odd one because `'abc'.encode('US-ASCII') == 'abc'.encode('UTF-8') => true`. So it can't be a direct encoding comparison anyway.

truffle/src/main/java/org/jruby/truffle/nodes/core/RegexpNodes.java
nirvdrum: I'm certainly open to revisiting. Consolidating all of that code would have largely been out of scope for the current work.

pitr-ch:
:+1:

PR #3687, opened 2016-02-19
[Truffle] Lazily initialised, shared string literals.

chrisseaton:
We want to be able to create a Ruby string with a literal value as easily as possible in our Java code, but we only want to create one rope (our equivalent of the `char[]`) per VM, as these are immutable and sharing them improves caching elsewhere.

@nirvdrum


truffle/src/main/java/org/jruby/truffle/language/literal/NilNode.java
```diff
@@ -32,7 +32,7 @@ public Object execute(VirtualFrame frame) {
 
     @Override
     public Object isDefined(VirtualFrame frame) {
-        return create7BitString("nil", UTF8Encoding.INSTANCE);
+        return stringLiterals().NIL.getString();
```
chrisseaton: This is the motivation for this work - we want to be able to refer to a compilation final immutable rope which we can package up in a tiny, escape analysable string object very quickly.

truffle/src/main/java/org/jruby/truffle/core/string/StringLiteral.java
```diff
@@ -0,0 +1,67 @@
...
+    public Rope getRope() {
+        if (rope == null) {
+            CompilerDirectives.transferToInterpreterAndInvalidate();
+            rope = createRope();
+        }
+
+        return rope;
+    }
+
+    public DynamicObject getString() {
```
nirvdrum: I think this method is a bit confusingly named. The wrapper object is a Java `String`, but this returns a Ruby `String`. Additionally, it's not a simple getter since it'll always create a new Ruby `String`.

nirvdrum:
This looks good. My only real gripe is the naming. We have a `StringLiteralNode` that has nothing to do with this `StringLiterals` class. And there's confusing terminology around Java vs Ruby strings.

truffle/src/main/java/org/jruby/truffle/core/string/StringLiteral.java
```diff
@@ -0,0 +1,67 @@
...
+        return StringOperations.createString(context, getRope());
+    }
+
+    private Rope createRope() {
+        // getRope is fully synchronised and returns identically the same object for the same parameters
+
+        return context.getRopeTable().getRope(
+                literal.getBytes(StandardCharsets.US_ASCII),
+                USASCIIEncoding.INSTANCE,
```
nirvdrum: I think you want this to `UTF8Encoding.INSTANCE`.

chrisseaton: I've constrained to 7-bit ASCII so we can use the advantageous 7-bit rope (see the assert in the constructor).

nirvdrum: This would be for correctness, however. In the `defined?` case, for instance, MRI would return a UTF-8 string. We can have a 7-bit UTF-8 rope (that's the dominant case in normal operation).

nirvdrum: I take that back. In the `defined?` case, MRI actually returns `ASCII-8BIT`, which would be `ASCIIEncoding.INSTANCE`.

chrisseaton: Ok I'll check for more cases as I go. Others include exception message.

truffle/src/main/java/org/jruby/truffle/core/string/StringLiteral.java
chrisseaton: Renamed `createInstance`.

chrisseaton:
I renamed it `CoreStrings`. Do you think that's better?

truffle/src/main/java/org/jruby/truffle/core/string/StringLiteral.java
chrisseaton: Yes it looks like `defined?` strings are all `Encoding::ASCII_8BIT`, whatever the magic comment or internal/external encodings. I'll switch to that for this use case and I'll check as I start to use it for other use cases.

chrisseaton:
Can you take a look again please @nirvdrum

nirvdrum:
Other than my cautionary comment, this looks good to me.

eregon:
Is this worth the extra complexity to initialize lazily?
If it's mostly for `defined?` then that's a fairly small set of strings which I believe should not impact startup significantly (or we need to revisit why it is so expensive to create such strings in the first place).

eregon:
Also, maybe the leaf ropes should be weakly cached like Symbols so this explicit global state is not necessary?
There is something similar (the `fstring` table) for MRI for literals and such.

PR #3671, opened 2016-02-16
[Truffle] CI for gems

pitr-ch:
There is a new `ci` command in j+tr tool.

PR #3691, opened 2016-02-22
[Truffle] Add Rails app integration test

pitr-ch:
cc @jruby/truffle


truffle/src/main/java/org/jruby/truffle/core/module/ModuleNodes.java
```diff
@@ -671,7 +671,7 @@ private Object classEvalSource(DynamicObject module, DynamicObject code, String
 
             CompilerDirectives.transferToInterpreter();
             // TODO (pitr 15-Oct-2015): fix this ugly hack, required for AS
-            final String space = new String(new char[line-1]).replace("\0", "\n");
+            final String space = new String(new char[Integer.max(line - 1, 0)]).replace("\0", "\n");
```
chrisseaton: I think this needs to be `Math.max` on Java 7 - that's why Travis has failed.

chrisseaton:
I guess the commented out code is because this is from a Rails generated template app?

Great work, Petr.

truffle/src/main/java/org/jruby/truffle/core/module/ModuleNodes.java
pitr-ch: Yeah, it has to, I was just looking at Travis.

pitr-ch:
yes, I think all comments in rails directory are from the rails generator.

pitr-ch:
Great work on the compatibility everyone! It worked almost out of the box :)

truffle/src/main/java/org/jruby/truffle/core/format/nodes/PackNode.java
```diff
@@ -42,6 +43,11 @@ public PackNode(RubyContext context) {
         this.context = context;
     }
 
+    @CompilerDirectives.TruffleBoundary
+    public static Encoding getAscii() {
+        return Encoding.load("ASCII");
```
eregon: ASCIIEncoding.INSTANCE maybe?

truffle/src/main/java/org/jruby/truffle/core/klass/ClassNodes.java
```diff
@@ -330,6 +330,8 @@ public SuperClassNode(RubyContext context, SourceSection sourceSection) {
         }
 
         @Specialization
+        @TruffleBoundary
+        // TODO (pitr-ch 22-Feb-2016): should it be on fast path?
```
eregon: If it is, then we should have an additional field to keep the superclass in classes since it can only ever be changed by `Class#initialize` and would obviate the need for the lookup.

.travis.yml
```diff
@@ -25,6 +25,7 @@ env:
     - JAVA_OPTS="-XX:+TieredCompilation -XX:TieredStopAtLevel=1"
     - MALLOC_ARENA_MAX=2
   matrix:
+    - JTR='ci --batch ../lib/ruby/truffle/jruby+truffle/gem_ci/travis.txt'
```
eregon: It's also at the end of the matrix. It seems this first PR commit overlaps with #3671.

truffle/src/main/java/org/jruby/truffle/core/format/nodes/PackNode.java
pitr-ch: thanks

truffle/src/main/java/org/jruby/truffle/core/klass/ClassNodes.java
pitr-ch: done

truffle/src/main/java/org/jruby/truffle/language/backtrace/BacktraceFormatter.java
```diff
@@ -128,6 +128,10 @@ public void printBacktrace(RubyContext context, DynamicObject exception, Backtra
     private String formatInLine(List<Activation> activations, DynamicObject exception) {
         final StringBuilder builder = new StringBuilder();
 
+        if (activations.size() == 0) {
+            throw new UnsupportedOperationException();
```
nirvdrum: Please use the message field to explain why it's unsupported.

truffle/src/main/java/org/jruby/truffle/language/backtrace/BacktraceFormatter.java
```diff
@@ -128,6 +128,10 @@ public void printBacktrace(RubyContext context, DynamicObject exception, Backtra
     private String formatInLine(List<Activation> activations, DynamicObject exception) {
         final StringBuilder builder = new StringBuilder();
 
+        if (activations.size() == 0) {
```
nirvdrum: There are differing opinions on this, but it might be clearer to call `activations.isEmpty()`.

truffle/src/main/java/org/jruby/truffle/core/klass/ClassNodes.java
```diff
@@ -330,8 +333,30 @@ public SuperClassNode(RubyContext context, SourceSection sourceSection) {
...
+        public Object getSuperClass(DynamicObject rubyClass,
+                                    @Cached("rubyClass") DynamicObject cachedRubyCLass,
+                                    @Cached("fastLookUp(cachedRubyCLass)") DynamicObject cachedSuperclass) {
+            // caches only initialized classes, just allocated will go through slow look up
+            return cachedSuperclass;
+        }
+
         @Specialization
-        public Object getSuperClass(DynamicObject rubyClass) {
+        DynamicObject getGetSuperClass(DynamicObject rubyClass) {
```
nirvdrum: Missing `contains`?

truffle/src/main/java/org/jruby/truffle/core/format/nodes/read/ReadBase64StringNode.java
```diff
@@ -162,8 +164,7 @@ public Object read(VirtualFrame frame, byte[] source) {
             }
         }
 
-        final Encoding encoding = Encoding.load("ASCII");
-        final ByteList result = new ByteList(lElem, 0, index, encoding, false);
+        final ByteList result = new ByteList(lElem, 0, index, ASCIIEncoding.INSTANCE, false);
```
nirvdrum: Thanks. This is much cleaner.

nirvdrum:
Looks great!

truffle/src/main/java/org/jruby/truffle/core/klass/ClassNodes.java
pitr-ch: I am not sure what's our usual approach to this. Leave few cache instances using generic specialisation for rest or using only generic specialisation when cache is not big enough? Here it would probably make more sense to add contains as Kevin suggests.

eregon: Agreed. The specialization name could be `getSuperClassUncached` to clarify.

PR #3694, opened 2016-02-23
[Truffle] Remove unused frame parameter in guards.

chrisseaton:
@chumer this modification makes my program fail on this assert in the generated code

```java
    @GeneratedBy(methodName = "writeInteger(Frame, int)", value = WriteFrameSlotNode.class)
    private static final class WriteIntegerNode_ extends BaseNode_ {

        WriteIntegerNode_(WriteFrameSlotNodeGen root) {
            super(root, 2);
        }

        @Override
        public Object execute(Frame frameValue, Object arg0Value) {
            if (arg0Value instanceof Integer) {
                int arg0Value_ = (int) arg0Value;
                assert (root.checkIntegerKind());
                return root.writeInteger(frameValue, arg0Value_);
            }
            return getNext().execute(frameValue, arg0Value);
        }

        static BaseNode_ create(WriteFrameSlotNodeGen root) {
            return new WriteIntegerNode_(root);
        }

    }
```

With the frame parameter in, the generated code is

```java
    @GeneratedBy(methodName = "writeInteger(Frame, int)", value = WriteFrameSlotNode.class)
    private static final class WriteIntegerNode_ extends BaseNode_ {

        WriteIntegerNode_(WriteFrameSlotNodeGen root) {
            super(root, 2);
        }

        @Override
        public Object execute(Frame frameValue, Object arg0Value) {
            if (arg0Value instanceof Integer) {
                int arg0Value_ = (int) arg0Value;
                if ((root.checkIntegerKind(frameValue))) {
                    return root.writeInteger(frameValue, arg0Value_);
                }
            }
            return getNext().execute(frameValue, arg0Value);
        }

        static BaseNode_ create(WriteFrameSlotNodeGen root) {
            return new WriteIntegerNode_(root);
        }

    }
```

Do you know what could be wrong? Why is the generated code so different? The `frame` variable in my code isn't even used.

### DIFF
--- a/truffle/src/main/java/org/jruby/truffle/language/locals/WriteFrameSlotNode.java
+++ b/truffle/src/main/java/org/jruby/truffle/language/locals/WriteFrameSlotNode.java
@@ -29,54 +29,54 @@ public abstract class WriteFrameSlotNode extends Node {
 
     public abstract Object executeWrite(Frame frame, Object value);
 
-    @Specialization(guards = "checkBooleanKind(frame)")
+    @Specialization(guards = "checkBooleanKind()")
     public boolean writeBoolean(Frame frame, boolean value) {
         frame.setBoolean(frameSlot, value);
         return value;
     }
 
-    @Specialization(guards = "checkIntegerKind(frame)")
+    @Specialization(guards = "checkIntegerKind()")
     public int writeInteger(Frame frame, int value) {
         frame.setInt(frameSlot, value);
         return value;
     }
 
-    @Specialization(guards = "checkLongKind(frame)")
+    @Specialization(guards = "checkLongKind()")
     public long writeLong(Frame frame, long value) {
         frame.setLong(frameSlot, value);
         return value;
     }
 
-    @Specialization(guards = "checkDoubleKind(frame)")
+    @Specialization(guards = "checkDoubleKind()")
     public double writeDouble(Frame frame, double value) {
         frame.setDouble(frameSlot, value);
         return value;
     }
 
-    @Specialization(guards = "checkObjectKind(frame)")
+    @Specialization(guards = "checkObjectKind()")
     public Object writeObject(Frame frame, Object value) {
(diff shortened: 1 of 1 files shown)


eregon:
The frame or a live value (an argument) must be passed explicitly here, otherwise the guard is considered pure and "cached" and is only ever called once at specialization instantiation time. The consequence is if the frame slot change the guard won't cache it and the assert will check that the guards are pure (return the same result), but they are not!

eregon:
The [SL node](https://github.com/graalvm/truffle/blob/master/truffle/com.oracle.truffle.sl/src/com/oracle/truffle/sl/nodes/local/SLWriteLocalVariableNode.java) does it very similarly, but it does not have a comment about why the Frame is passed.

eregon:
Maybe using `Frame.isInt(FrameSlot)` would be nicer for this use case?

eregon:
> Maybe using Frame.isInt(FrameSlot) would be nicer for this use case?

Nevermind, that seems to check something different.

chumer:
Yes Benoit is right. If no dynamic value is in the expression the whole expression is considered compilation constant. I considered introducing an explicit syntax for this when I implemented that feature, but I thought the more common case would be that its considered compilation constant.

chrisseaton:
Ok I understand the issue now.

I think it's a problem because removing dead code shouldn't change behaviour - it even compiles with no apparent difference until you run it - and if your tests don't cover it for whatever reason it's going to be trouble.

Assuming arbitrary code that nobody has annotated in anyway won't have side effects just seems wrong in the first place.

I'm not sure what to suggest though. Are the pure, cached guards a big win in some cases?

eregon:
> I'm not sure what to suggest though. Are the pure, cached guards a big win in some cases?

Well, yes, otherwise they would need to be checked at every call, even when called on only `@Cached` values.
It's documented in http://graalvm.github.io/truffle/javadoc/latest/com/oracle/truffle/api/dsl/Specialization.html#guards--

PR #3717, opened 2016-03-08
[Truffle] Add copy of 2.2.2 stdlib

pitr-ch:
JRuby+Truffle does not support 2.3 features yet, so it needs copy of 2.2.2 stdlib.
cc @jruby/truffle, @headius, @enebo


headius:
Ugh...we really don't want to have two copies of stdlib in the repo again if we can avoid it. Perhaps the few features you need to use the 2.3 stdlib would not be too difficult to implement?

chrisseaton:
Yeah we thought you may not be keen - it's a reasonable response.

Ideally we'll implement the 2.3 features quickly, but in the mean time it's not great to be working on other things and having some random failures here and there from the new stdlib.

How long could we have this in? Are you releasing imminently?

pitr-ch:
The copy was updated to contain only what's actually needed by JRuby+Truffle, it has only 3,4MB now.

enebo:
I also find this undesirable but so long as the timeframe does not stretch multiple point releases and there is a plan for removing this I think we can eat the extra space in a single point release.  I partially blame us for not collaborating on this version move a bit more.

pitr-ch:
Ok thanks, it helps a lot. I'll rebase and merge.

kares:
you guys considered how confusing such a release might get for end users - even ones trying out truffle?

pitr-ch:
@kares valid point, I'll move the copy to truffle directory so it's clear what is the purpose of the copy.

PR #3722, opened 2016-03-10
thread-context refactor backtrace retrieval

kares:
historically there's `createBacktrace2(int level, boolean nativeException)`
... whose parameters are no longer used, proposing a `getBacktrace()` instead

getter name chosen since thread-context manages a backtrace state - does not need to create/generate one.

alternatively `getBacktrace(int level)` is added although we do not use the API.


headius:
Are the "final" variable changes something your editor/IDE is doing? I don't mind it for new code but making changes like that in otherwise-unmodified code really messes up the diff.

Otherwise, this seems like a fine change and you should go ahead with it.

kares:
@headius yep, thanks - will try to tidy up the commit to avoid unrelated changes.

kares:
cleaned up: https://github.com/jruby/jruby/commit/f060c79dd9a56b72fb453e4c4c03ffecdaf72c49 and pushed to master

PR #3749, opened 2016-03-22
TCK expects execution of each test in an isolated environment. Creati…

jtulach:
TCK expects execution of each test in an isolated environment. Creating the `PolyglotEngine` instance in each invocation of `prepareVM`. With this change the https://github.com/graalvm/truffle/pull/146 request seems to pass OK.

### DIFF
--- a/truffle/src/test/java/org/jruby/truffle/tck/RubyTckTest.java
+++ b/truffle/src/test/java/org/jruby/truffle/tck/RubyTckTest.java
@@ -14,27 +14,15 @@ import com.oracle.truffle.api.vm.PolyglotEngine;
 import com.oracle.truffle.tck.TruffleTCK;
 
 import org.jruby.truffle.RubyLanguage;
-import org.junit.Test;
 
 import static org.junit.Assert.*;
 
 public class RubyTckTest extends TruffleTCK {
-
-    private static PolyglotEngine engine;
-
-    @Test
-    public void checkVM() {
+    @Override
+    protected PolyglotEngine prepareVM() throws Exception {
         PolyglotEngine engine = PolyglotEngine.newBuilder().build();
         assertNotNull(engine.getLanguages().get(mimeType()));
-    }
-
-    @Override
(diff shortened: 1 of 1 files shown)


chrisseaton:
Unfortunately it's a little more complicated than that.

As you can see the TCK is now failing with what appears to be the kernel's OOM killer. JRuby runs tests concurrently, so it's trying to create 40 or so Ruby contexts at the same time and running out of memory on the Travis instance.

We could run the tests sequentially, but then they're starting to take several minutes to run. I think we also had some other problem when we tried to run test sequentially last time we tried - @nirvdrum can you remember what that was?

I'm not sure what the best solution is. Is there something fundamentally wrong with creating one context and running most of the tests in that one context?

jtulach:
Yes, the execution of `TruffleTCK` now takes longer on my computer about 1min instead of 15s.

Running the tests in isolation is the preferred mode. The tests shouldn't influence each other. However, if it is complicated to change that right now, I can try some workaround in the `TruffleTCK`.

chrisseaton:
I'll just wait for @nirvdrum to wake up and see if he remembers anything more about what happened last time we tried to make them run sequentially.

nirvdrum:
It's been a while, but I don't think we had a problem with them running sequentially -- they were just very slow and by default, run on every maven build. If we switch back to running that way, we really need to figure out whatever maven incantation to disable the tests. I tried a few times and failed each time.

The issues I had run into was trying to make them run in parallel. I naively tried to pass the context around, but the test teardown method nulls out the context, IIRC. So, only the first test would pass. Chris got the current setup working.

chrisseaton:
That was the problem - we can't add 3 mins of test time to everyone else's Maven builds, even when they aren't changing anything in Truffle. What options do we have to workaround @jtulach?

jtulach:
I think I have a workaround: https://github.com/graalvm/truffle/pull/146/commits/fe43728f73054a750a26e4baab25e5921a556b66 - for now it should be enough.

PR #3761, opened 2016-03-27
Adding tests that verify that jruby-1067 works as expected - ie Strin…

jstemen:
It looks like jruby-1067 has already been fixed.  These tests show that capturing by named parameters with StringScanner#scan works.


kares:
// cc @eregon does it look :+1: to you?

jstemen:
Hmm... It looks like the tests are failing because they are using the truffle.mspec config.  I didn't see that documented in the BUILDING.md file.  The "-X+T" flag on line 16 seems to be what is causing the failure.  Can anyone point me towards some documentation about what this flag does?  Thanks

nirvdrum:
"-X+T" enables the Truffle backend for JRuby. The Travis target failing is running the specs on the Truffle backend, so the presence of the flag is appropriate. More than likely, the new specs demonstrate a bug in the Truffle backend that we'll need to deal with. In the meanwhile, I think it's appropriate to just simply tag those specs.

Those tags live in `spec/truffle/tags`. But I think you could also just merge this and we can tag once in master.

jstemen:
@nirvdrum , thanks for the info.  I added the tag in to ignore the test for Truffle.  The appveyor/pr build is failing now, but I'm not sure if that's related to my change.

enebo:
This is very old now and it is a shame if these specs did not make it into ruby/spec but if there is still interest these should get added as a PR to ruby/spec.  If this was clean to merge I might just land it and it would make it upstream but as it stands it should just become a ruby/spec PR.  Very sorry this never got merged.

PR #3774, opened 2016-04-01
Enable AppVeyor tests on 64-bit Java on Windows as well.

perlun:
It was already enabled on the master branch, but not here.

Side note: Should we also add the ant MRI tests on master? I noted that it was only enabled in the jruby-1_7 branch, but not in the appveyor.yml on master.

(@enebo - I happened to see this after your comment in #3767)

### DIFF
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -4,6 +4,7 @@ build: off
 
 environment:
   matrix:
+    - JAVA_HOME: C:\Program Files\Java\jdk1.8.0
     - JAVA_HOME: C:\Program Files (x86)\Java\jdk1.8.0
 
 install:


enebo:
@perlun 64 bit is currently disabled because our current native launcher only can in-process launch the JVM in 32 bit mores.  In 64 JVM it fails and the launcher falls back to launching Java via the command-line.  We have an open bug on windows (I don't recall number offhand) which covers the fact that we improperly quote in the native launcher in the presence of quotes.  This causes many tests to fail which invoke sub-procs of JRuby.  We also have no ability to skip tests based on VM arch type (32 v 64) so I disabled running on 64 bit JVMs until we upgrade mjruby into jruby-launcher project.

Whew...a lot of explanation for why I am not merging this today :)  It will be added in once we update jruby-launcher.

perlun:
> Whew...a lot of explanation for why I am not merging this today :) It will be added in once we update jruby-launcher.

Hehe, ok! Thanks for taking the time to explain the "why's" behind this.

PR #3784, opened 2016-04-07
[Truffle] Discussion: introducing first-class Array strategies

eregon:
@chrisseaton @nirvdrum @pitr-ch I am thinking to use `ArrayStrategy` objects for our strategies on Array to reduces the number of specialization and duplication by 4.
This is only a preview to get some feedback (do not merge).
What do you think?
Is it nice / too complex / not that helpful?

This is maybe slightly overlapping with ArrayMirror, although ArrayMirror is really about helpers for Java arrays, and this is really about the storage strategies for our Ruby Arrays.
In the current version, an ArrayStrategy can build an ArrayMirror easily so there is no conflict of features. There are only 5 ArrayStrategy which are singleton instances, so this also avoids potential allocation in the interpreter.

An alternative solution would be to have one node per "functionality" of Array, like "get at index", "set at index", "grow", "delete/shrink", etc. This already exists to some extent but it's underused because it's quite verbose, there would be many many features like that for Ruby Array and it's not obvious to identify them either while still keeping the performance.

### DIFF
--- a/truffle/src/main/java/org/jruby/truffle/core/array/ArrayAppendOneNode.java
+++ b/truffle/src/main/java/org/jruby/truffle/core/array/ArrayAppendOneNode.java
@@ -36,65 +36,22 @@ public abstract class ArrayAppendOneNode extends RubyNode {
 
     // Append into an empty array
 
-    @Specialization(guards = "isNullArray(array)")
-    public DynamicObject appendOneEmpty(DynamicObject array, int value) {
-        Layouts.ARRAY.setStore(array, new int[]{value});
-        Layouts.ARRAY.setSize(array, 1);
-        return array;
-    }
-
-    @Specialization(guards = "isNullArray(array)")
-    public DynamicObject appendOneEmpty(DynamicObject array, long value) {
-        Layouts.ARRAY.setStore(array, new long[]{value});
-        Layouts.ARRAY.setSize(array, 1);
-        return array;
-    }
-
-    @Specialization(guards = "isNullArray(array)")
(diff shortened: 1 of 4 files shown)


nirvdrum:
Upon first reading, I like this overall. It's hard to trace the specializations given how many of them there are. However, I find pushing more logic into guards and @Cached expressions also makes things hard to reason about. The execution order there is a bit backwards and finding where the static method is a bit annoying (no IDE help without going to the generated classes).

The other part I like about this is it opens the door for more strategies without having to fight with even more specialization explosion.

pitr-ch:
I like it! It clearly defines the strategies and its interface.

chrisseaton:
The reason I didn't originally do it this way was because I thought I could save an object reference - if we had an `int[]` I would know the strategy to use, and it seemed cool that the strategies would just be specialisations with guards. I think PE also did not work the same back then - you needed a `VirtualFrame` parameter for stuff to be inlined.

I think it's right to move beyond that to some specialisation classes.

If you are experimenting, try some more unusual strategies early to catch if you are painting yourself into any corners. For example maybe a `Bitset` specialisation for boolean arrays? Or some kind of spare array?

You've probably already read Laurie Tratt's et al's strategies paper, but if you haven't do that.

eregon:
> The reason I didn't originally do it this way was because I thought I could save an object reference - if we had an int[] I would know the strategy to use, and it seemed cool that the strategies would just be specialisations with guards.

It's still the case, the array layout is unchanged, it's just a way to express code more generically.

eregon:
Thanks for the feedback!
I'll integrate this next week.

PR #3785, opened 2016-04-07
[Truffle] Ropes byte visitor

chrisseaton:


truffle/src/main/java/org/jruby/truffle/stdlib/DigestNodes.java
```diff
@@ -146,9 +148,17 @@ public UpdateNode(RubyContext context, SourceSection sourceSection) {
...
+            RopeOperations.visitBytes(StringOperations.rope(message), new BytesVisitor() {
+
+                @Override
+                public void accept(byte[] bytes, int offset, int length) {
+                    digest.update(bytes, offset, length);
+                }
+
+            });
 
-            DigestLayoutImpl.INSTANCE.getDigest(digestObject).update(rope.getBytes(), rope.begin(), rope.byteLength());
```
chrisseaton: This is the whole point of this PR - don't allocate the full byte[] to output it (or feed it into a digest function in this case).

truffle/src/main/java/org/jruby/truffle/core/rope/ConcatRope.java
```diff
@@ -55,47 +56,6 @@ public byte getByteSlow(int index) {
...
-
-            return ret;
-        }
-
-        byte[] leftBytes;
-        byte[] rightBytes;
-        final int leftLength = left.byteLength();
-
-        if (offset < leftLength) {
-            // The left branch might not be large enough to extract the full byte range we want. In that case,
```
nirvdrum: While it's the only comment here, I think carrying it forward in `RopeOperations` has merit.

truffle/src/main/java/org/jruby/truffle/core/rope/RopeOperations.java
```diff
@@ -208,6 +208,92 @@ public static LeafRope flatten(Rope rope) {
...
+
+    @TruffleBoundary
+    public static byte[] extractRange(Rope rope, int offset, int length) {
+        final ByteList result = new ByteList(length);
+
+        visitBytes(rope, new BytesVisitor() {
+
+            @Override
+            public void accept(byte[] bytes, int offset, int length) {
+                result.append(bytes, offset, length);
```
nirvdrum: `ByteList` has a weird assertion that I think will fail with a small set of `bytes` far enough along in the whole string. It may not matter (and it looks like a bug in `ByteList` to me). But since we run tests with assertions, it might come up. It may be better to just do the byte copying operations on our own here.

nirvdrum:
A couple small comments. Otherwise, it looks good.

PR #3797, opened 2016-04-12
[Truffle] Test to check debugger integration

jtulach:
I've taken the `SLDebugTest` and started its conversion to work with Ruby. The algorithm seems to run OK, the breakpoint should be placed, but the execution doesn't stop at it. Right now I don't know why, but in spite of that I decided the share what I have. At least as an inspiration.


jtulach:
Ready for merge. 

After `stepOut` the current source section is the whole `main` method. I am not sure if that is correct, but I leave it up to you guys.

PR #3800, opened 2016-04-13
[Truffle] Begin Sulong integration

chrisseaton:
Adds a command `jruby-cext-c` to compile C extensions.

Loads Sulong modules and calls the `Init_` method.

Example:

```
$ bin/jruby bin/jruby-cext-c test/truffle/cexts/minimum
$ jt run --sulong -Itest/truffle/cexts/minimum/lib test/truffle/cexts/minimum/bin/minimum
```

At the moment none of the C API is supported, and there is some bug that stops `printf` working from the `Init_` method, so it's hard to see that anything is happening, but it is!

You need Sulong, obviously. Depends on https://github.com/graalvm/sulong/pull/150.

@mrigger @nirvdrum @pitr-ch @eregon


bin/jruby-cext-c
```diff
@@ -0,0 +1,62 @@
...
+# GNU General Public License version 2
+# GNU Lesser General Public License version 2.1
+
+# A compiler for JRuby+Truffle C extensions
+
+require 'yaml'
+
+MX_DIR = ENV['MX_DIR']
+
+if MX_DIR.nil?
```
eregon: `unless MX_DIR`

bin/jruby-cext-c
```diff
@@ -0,0 +1,62 @@
...
+  $stderr.puts 'You need to set MX_DIR to a checkout of the mx repository'
+  exit 1
```
eregon: `abort "message"`

core/src/main/java/org/jruby/util/RubyFiletypeDetector.java
```diff
@@ -43,6 +43,9 @@ public String probeContentType(Path path) throws IOException {
             if (firstLine != null && SHEBANG_REGEXP.matcher(firstLine).matches()) {
                 return RUBY_MIME;
             }
+        } catch (Exception e) {
+            // Reading random files as UTF-8 could cause all sorts of errors
```
eregon: Maybe it's still good to limit that to IOException + transcoding exceptions?

eregon:
What does the final `c` in `bin/jruby-cext-c` means? `compile`? Should it be `cc` so we relate by name to the standard C compiler? Maybe it could also be `jruby-cext-make`, as standard workflow to compile a C-ext is: `cd ext/dir; ruby extconf.rb; make;`.

bin/jruby-cext-c
eregon: (same below)

bin/jruby-cext-c
chrisseaton: Fixed

core/src/main/java/org/jruby/util/RubyFiletypeDetector.java
chrisseaton: Apparently you can throw an `IOException`. But this confuses me - it looks like the code that calls this method doesn't catch any exceptions, so an error in any of your file type detectors causes any further detectors to fail. This makes our detector robust.

chrisseaton:
Yes it means JRuby's C extension compiler. I didn't call it 'cc' as it'll need to be able to compile C++ as well. Do people use other languages as well? `make` might be a better idea. Hopefully this isn't for the long term though, as we should modify the `rake` plugin at some point.

core/src/main/java/org/jruby/util/RubyFiletypeDetector.java
eregon: Right, this is confusing. I guess it is for the case where the file does not exist, the caller might as well be warned.
But we can catch IOException anyway I think.
The current diff would also catch even a NullPointerException silently. This seems a good way to hide bugs.

nirvdrum:
I know of extensions written in Rust used in the wild. I've seen interesting in writing them in Go, which is evidently possible now.

PR #3803, opened 2016-04-14
Truffle cext sketch

chrisseaton:
Depends on https://github.com/graalvm/sulong/pull/157

This is a sketch of how I think we could support the Ruby C API using interop. When we used Truffle C, we had several intrinsic nodes, so Ruby functionality implemented in Truffle C. I guess we want to avoid that? Otherwise we'll have to define an API that Sulong provides to register those intrinsics. If we can do it with just normal interop that's probably better.

We have a header that describes the Ruby C API.

We then have a library that implements that in terms of the interop functions that Sulong can provide. That library is compiled into a `.su` and shipped with JRuby+Truffle, and loaded whenever the first C extension is loaded.

Most of the C API is implemented by `INVOKE`ing methods that are implemented by JRuby+Truffle, in a `Truffle::CExt` module, that is exported by default as `ruby_cext`.

There are a lot of open questions here.

If the return type of interop functions like `READ` are `void*`, to read it do I do `*((int*) READ`, or `(int) READ`? What do I do with variadic arguments to pass them on a second time? `RSTRING_PTR` needs to provide what appears to be a `char*` but actually allows me to intercept read and write operations, and that needs to work with pointer arithmetic. We solved all these with Truffle C, so they're all do-able.

### DIFF
--- a/lib/ruby/truffle/cext/ruby.h
+++ b/lib/ruby/truffle/cext/ruby.h
@@ -26,15 +26,25 @@ extern "C" {
 typedef void *ID;
 typedef void *VALUE;
 
-extern VALUE Qfalse;
-extern VALUE Qtrue;
-extern VALUE Qnil;
+VALUE get_Qfalse();
+VALUE get_Qtrue();
+VALUE get_Qnil();
 
-extern VALUE rb_cObject;
-extern VALUE rb_cArray;
-extern VALUE rb_cHash;
+#define Qfalse get_Qfalse()
+#define Qtrue get_Qtrue()
+#define Qnil get_Qnil()
 
-extern VALUE rb_eRuntimeError;
+VALUE get_rb_cObject();
+VALUE get_rb_cArray();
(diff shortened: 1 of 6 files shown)


lib/ruby/truffle/cext/ruby.h
```diff
@@ -26,15 +26,25 @@ extern "C" {
 typedef void *ID;
 typedef void *VALUE;
 
-extern VALUE Qfalse;
-extern VALUE Qtrue;
-extern VALUE Qnil;
+VALUE get_Qfalse();
+VALUE get_Qtrue();
+VALUE get_Qnil();
```
eregon: Why make them functions? Could the constructor below assign them?

chrisseaton: I think an extern variable escapes. I'm not even sure my `static void *ruby_cext;` below is going to PE.

PR #3805, opened 2016-04-14
[Truffle] Experiment with lazy integer ropes.

chrisseaton:
@nirvdrum what do you think of this?

Saves about 5s on test fast for me, and it makes patterns like this constant fold to nothing:

```ruby
loop do
  start = Time.now
  10_000_000.times do
    "foo " + 14.to_s
  end
  puts Time.now - start
end
```

### DIFF
--- a/truffle/src/main/java/org/jruby/truffle/core/numeric/FixnumNodes.java
+++ b/truffle/src/main/java/org/jruby/truffle/core/numeric/FixnumNodes.java
@@ -27,6 +27,7 @@ import org.jruby.truffle.core.CoreLibrary;
 import org.jruby.truffle.core.CoreMethod;
 import org.jruby.truffle.core.CoreMethodArrayArgumentsNode;
 import org.jruby.truffle.core.Layouts;
+import org.jruby.truffle.core.rope.LazyIntRope;
 import org.jruby.truffle.core.rope.RopeConstants;
 import org.jruby.truffle.language.NotProvided;
 import org.jruby.truffle.language.RubyNode;
@@ -1068,15 +1069,18 @@ public abstract class FixnumNodes {
             super(context, sourceSection);
         }
 
-        @TruffleBoundary
         @Specialization
         public DynamicObject inspect(int n) {
-            return create7BitString(Integer.toString(n), USASCIIEncoding.INSTANCE);
+            return createString(new LazyIntRope(n));
         }
 
         @TruffleBoundary
         @Specialization
         public DynamicObject inspect(long n) {
+            if (n >= Integer.MIN_VALUE && n <= Integer.MAX_VALUE) {
+                return inspect((int) n);
+            }
+
             return create7BitString(Long.toString(n), USASCIIEncoding.INSTANCE);
         }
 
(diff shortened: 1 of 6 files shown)


truffle/src/main/java/org/jruby/truffle/core/rope/LazyIntRope.java
```diff
@@ -0,0 +1,110 @@
...
+        if (newCodeRange != getCodeRange()) {
+            CompilerDirectives.transferToInterpreter();
+            throw new UnsupportedOperationException("Cannot fast-path updating encoding with different code range.");
+        }
+
+        return new LazyIntRope(value, newEncoding, length(value));
+    }
+
+    @Override
+    public byte[] fulfil() {
```
nirvdrum: `fulfil` -> `fulfill`? I think we were standardizing on US spelling.

nirvdrum:
The rope portion looks good to me. Do you have a sense of how often the lazy part pays off? The other container ropes generally see an advantage with `GetByteNode`, which is used for `String#[]` with a single index in 7-bit strings. So, they see a lot of utility without having to populate their byte array. But this looks like it faults on that operation.

nirvdrum:
Also, you may want to add something to `RopeNodes.DebugPrintRopeNode`. That's called by `Truffle::Primitive.debug_print_rope`, which can be handy when you want to see the structure of a rope without a debugger.

truffle/src/main/java/org/jruby/truffle/core/numeric/FixnumNodes.java
```diff
@@ -1068,15 +1069,18 @@ public InspectNode(RubyContext context, SourceSection sourceSection) {
...
         @Specialization
         public DynamicObject inspect(int n) {
-            return create7BitString(Integer.toString(n), USASCIIEncoding.INSTANCE);
+            return createString(new LazyIntRope(n));
         }
 
         @TruffleBoundary
         @Specialization
         public DynamicObject inspect(long n) {
+            if (n >= Integer.MIN_VALUE && n <= Integer.MAX_VALUE) {
```
eregon: CoreLibrary.fitsIntoInteger

truffle/src/main/java/org/jruby/truffle/core/rope/LazyIntRope.java
```diff
@@ -0,0 +1,110 @@
...
+            return 7;
+        }
+
+        if (value < 100000000) {
+            return 8;
+        }
+
+        if (value < 1000000000) {
+            return 9;
+        }
```
eregon: Does this perform better than a loop?

truffle/src/main/java/org/jruby/truffle/core/rope/LazyIntRope.java
chrisseaton: Done.

truffle/src/main/java/org/jruby/truffle/core/numeric/FixnumNodes.java
chrisseaton: Done.

truffle/src/main/java/org/jruby/truffle/core/rope/LazyIntRope.java
chrisseaton: Done as a loop - generated code is the same, so going with the more concise loop.

PR #3807, opened 2016-04-18
[ji] allow for initialize() to be a valid Java method

### LINKED ISSUE #1408: Jruby class implementing java interface with "initialize" method as a part of the interface fails to compile with jrubyc
```
class FooInterceptor
 java_implements 'org.apache.flume.interceptor.Interceptor'

 java_signature "void initialize()"
 def init; end

 ...snip...
end
```

Results in the below error when compiling with jrubyc.

```
error: initialize() in FooInterceptor cannot override initialize() in RubyObject
    public void initialize() {
                ^
  return type void is not compatible with IRubyObject
```

Is there a way around this that isn't published?

kares:
(been deprecated all way back to 1.7) ... so maybe a good time to remove?!

and resolve #1408

### DIFF
--- a/core/src/main/java/org/jruby/RubyBasicObject.java
+++ b/core/src/main/java/org/jruby/RubyBasicObject.java
@@ -2952,11 +2952,6 @@ public class RubyBasicObject implements Cloneable, IRubyObject, Serializable, Co
     // Deprecated methods below this line
 
     @Deprecated
-    public IRubyObject initialize() {
-        return getRuntime().getNil();
-    }
-
-    @Deprecated
     protected RubyBasicObject(Ruby runtime, RubyClass metaClass, boolean useObjectSpace, boolean canBeTainted) {
         this(runtime, metaClass, useObjectSpace);
     }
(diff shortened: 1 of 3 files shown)


enebo:
@kares This one makes me nervous() IRubyObject.initialize() just feels like something I could see a native ext calling.  OTOH it has been throwing Deprecated errors for 2+ years.

kares:
@enebo yeah I wasn't sure about this one either. feel free to postpone (or just remove milestone).

enebo:
@kares I feel too leery on this one but I don't want it open until our next major reelase.  So I am closing.

kares:
re-opening ... potentially to be considered for next major release

ivoanjo:
Any hope for this one? I just bumped into this exact same issue 😱

headius:
After looking at this again...I'm not really too scared of removing this long-deprecated method.

@enebo Worst case we'd have to restore it in 9.2.1, right?

enebo:
@headius this is a significant release so I tend to agree with you.  As you say worst-case revert if we get too many troubles :)

Issue #3475, opened 2015-11-20
proc-to-interface dispatch method collision using Java 8 streams


cshupp1:
The following java code:
  ```
public static void main (String[] args) {
    List<Integer> numbers = Arrays.asList(1, 2, 3, 4, 5, 6, 7, 8);
    List<Integer> twoEvenSquares =
        numbers.stream()
               .filter(n -> {
                        System.out.println("filtering " + n);
                        return n % 2 == 0;
                      })
               .map(n -> {
                        System.out.println("mapping " + n);
                        return n * n;
                      })
               .limit(2)
               .collect(() -> new ArrayList<>(),
                           (c, e) -> c.add(e),
                           (c1, c2) -> c1.addAll(c2));
    System.out.println("Output: " + twoEvenSquares);
  }
```
yields:
```
filtering 1
filtering 2
mapping 2
filtering 3
filtering 4
mapping 4
Output: [4, 16]
```

The ostensibly equivalent JRuby code:
```
numbers = java.util.Arrays.asList(1,2,3,4,5,6,7,8)

filter_lamb = ->(n) {
  puts "filtering #{n}"
  n % 2 == 0
}
map_lamb = ->(n) {
  puts "mapping #{n}"
  n*n
}
l1 = -> { java.util.ArrayList.new }
l2 = -> (c,e) { c.add(e) }
l3 = -> (c1,c2) { c1.addAll(c2)}

two_even_squares =
numbers.stream().filter(filter_lamb).map(map_lamb).limit(2).collect(l1,l2,l3)
```

bombs out with:
```
ArgumentError: wrong number of arguments (1 for 2)
```
in the collect call.

Inspection in irb shows my method:

```
two_even_squares.java_class.java_instance_methods.map(&:to_s).reject do
|e| e !~ /.*collect.*/ end
=> ["public final java.lang.Object
java.util.stream.ReferencePipeline.collect(java.util.stream.Collector)",
"public final java.lang.Object
java.util.stream.ReferencePipeline.collect(java.util.function.Supplier,java.util.function.BiConsumer,java.util.function.Bi
Consumer)"]
```

Getting explicit doesn't help:

```
irb(main):024:0>  two_even_squares.java_send :collect,
[java.util.function.Supplier,java.util.function.BiConsumer,java.util.function.BiConsumer],
l1, l2, l3
ArgumentError: wrong number of arguments (1 for 2)
        from org/jruby/java/proxies/JavaProxy.java:352:in `java_send'
[...]

PR #3809, opened 2016-04-19
bind (abstract) methods when implementing interface

kares:
there are two ways Ruby interface implementations are being generated : 
* explicitly using `Iface.impl { |name, *args| }`
* or implicitly when a block is provided for a (commonly functional) inteface last argument (proc-to-iface)

both of these add `method_missing` to the Ruby class which leads to issues when there's a naming clash in the hierarchy (e.g. `Kernel#test` with `java.util.function.Predicate#test`). these are resolved by binding implemented interface methods on the implementing Ruby class - always executing the desired user provided bits (except where he would `ImplClass.class_eval { def test ... }` the method of course).

on the Java (codegen) side, under Java 8, this further revealed [that static methods were unnecessarily being generated](https://github.com/jruby/jruby/commit/9a1e3526e89005e7c367c9c39727385b857134bb) and that interface's default methods are always being overridden ...

1. for `Iface.impl` the behaviour of implementing all methods (including defaults) was kept.
`impl(:foo, :bar)` already allows for implementing method names to be specified and an additional `impl(false)` to generate just the minimum abstract method(s) required is introduced. new code also warns on `impl(:foo)` when `"foo"` is not actually an interface method.

2. for proc-to-iface conversion overriding everything (including default) makes no sense esp. since its mostly used with functional interfaces. so this was changed and allows for slightly advanced concepts with Java 8 (e.g. `java.util.function.Function#compose`) to work correctly.

all-in-all these changes should be welcoming by JI users - it might break compatibility in very corner cases but its usually for the best to review those.

resolves https://github.com/jruby/jruby/issues/3475


enebo:
@kares can you think of any breakage this will cause?  Your last sentence implies there might be breakage but do you actually know of any?

kares:
@enebo only thing I can think of if users realize that these use `method_missing` and they would re-def it but it does make very little sense to do so. we can manage if there's complains but I doubt there will be.

enebo:
@kares if someone might realize we used method_missing as our implementation and then monkeypatch it to change behavior I am ok with that risk.

@headius what do you think?  Can you think of any risks here?

headius:
Ok, this one seems fine to me and I think the impl is right.

PR #3813, opened 2016-04-20
[Truffle] mx clean and mx build work for JRuby

jtulach:
Majority of Graal/Truffle projects is using `mx` build tool. JRuby is an exception. The goal of this pull request is to provide an `mx` facade that delegates execution to existing `mvn` goals, but behaves like a normal `mx` from outside and thus simplifies including of JRuby in other `mx` based projects.

With the current changes I can run the build and it successfully connects Truffle with JRuby and builds both:
```bash
$ mx clean
$ mx build
```

### DIFF
--- a/mx.jruby/mx_jruby.py
+++ b/mx.jruby/mx_jruby.py
@@ -1,0 +1,108 @@
+import mx
+import os
+from os.path import join, exists
+import subprocess
+import shutil
+
+_suite = mx.suite('jruby')
+
+class MavenProject(mx.Project):
+    def __init__(self, suite, name, deps, workingSets, theLicense, **args):
+        mx.Project.__init__(self, suite, name, "", [], deps, workingSets, _suite.dir, theLicense)
+        self.javaCompliance = "1.7"
+        self.build = hasattr(args, 'build')
+        self.prefix = args['prefix']
+
+    def source_dirs(self):
+        return []
+
+    def output_dir(self):
+        dir = os.path.join(_suite.dir, self.prefix)
+        return dir.rstrip('/')
+
+    def source_gen_dir(self):
+        return None
+
+    def getOutput(self, replaceVar=False):
+        return os.path.join(_suite.dir, "target")
+
+    def getResults(self, replaceVar=False):
+        return mx.Project.getResults(self, replaceVar=replaceVar)
+
+    def getBuildTask(self, args):
+        return MavenBuildTask(self, args, None, None)
+
+    def isJavaProject(self):
+        return True
+
+    def archive_prefix(self):
+        return self.prefix
+
+
+class MavenBuildTask(mx.BuildTask):
+    def __init__(self, project, args, vmbuild, vm):
+        mx.BuildTask.__init__(self, project, args, 1)
+        self.vm = vm
+        self.vmbuild = vmbuild
+
+    def __str__(self):
+        return 'Building Maven for {}'.format(self.subject)
+
+    def needsBuild(self, newestInput):
+        return (True, 'Let us re-build everytime')
+
+    def newestOutput(self):
+        return None
+
+    def build(self):
+        if not self.subject.build:
+            mx.log("...skip build of {}".format(self.subject))
+            return
+        mx.log('...perform build of {}'.format(self.subject))
+
+        rubyDir = _suite.dir
+
+        # HACK: since the maven executable plugin does not configure the
+        # java executable that is used we unfortunately need to append it to the PATH
+        javaHome = os.getenv('JAVA_HOME')
+        if javaHome:
+            os.environ["PATH"] = os.environ["JAVA_HOME"] + '/bin' + os.pathsep + os.environ["PATH"]
+
+        mx.logv('Setting PATH to {}'.format(os.environ["PATH"]))
+        mx.logv('Calling java -version')
+        mx.run(['java', '-version'])
+
+        # Truffle version
+
+        truffle = mx.suite('truffle')
+        truffle_commit = truffle.vc.parent(truffle.dir)
+
+        mx.run_mx(['maven-install'], suite=truffle)
+
+        open(join(rubyDir, 'VERSION'), 'w').write('graal-vm\n')
+
+        # Build jruby-truffle and
+
+        mx.run(['find', '.'], nonZeroIsFatal=False, cwd=rubyDir)
+        mx.run_maven(['--version'], nonZeroIsFatal=False, cwd=rubyDir)
+
+        mx.log('Building without tests')
+
+        mx.run_maven(['-DskipTests', '-X', '-Dtruffle.version=' + truffle_commit], cwd=rubyDir)
+
+        mx.log('Building complete version')
+
+        mx.run_maven(['-Pcomplete', '-X', '-DskipTests', '-Dtruffle.version=' + truffle_commit], cwd=rubyDir)
+        mx.run(['zip', '-d', 'maven/jruby-complete/target/jruby-complete-graal-vm.jar', 'META-INF/jruby.home/lib/*'], cwd=rubyDir)
+        mx.run(['bin/jruby', 'bin/gem', 'install', 'bundler', '-v', '1.10.6'], cwd=rubyDir)
+#        shutil.rmtree(os.path.join(_suite.dir, "lib", "target"), True)
+#        shutil.rmtree(os.path.join(_suite.dir, 'lib', 'lib', 'jni'), True)
+#        shutil.copytree(os.path.join(_suite.dir, 'lib', 'jni'), os.path.join(_suite.dir, 'lib', 'lib', 'jni'))
+#        shutil.rmtree(os.path.join(_suite.dir, 'lib', 'jni'), True)
+
+    def clean(self, forBuild=False):
+        if forBuild:
+            return
+        rubyDir = _suite.dir
+        mx.run_maven(['clean'], nonZeroIsFatal=False, cwd=rubyDir)
+
--- a/mx.jruby/suite.py
+++ b/mx.jruby/suite.py
@@ -1,0 +1,101 @@
+suite = {
+    "mxversion": "5.6.6",
+    "name": "jrubytruffle",
+    "defaultLicense": "EPL",
+
+    "imports": {
+        "suites": [
+            {
+                "name": "truffle",
+                "version": "551e8475af2fc8769bc3ead07c9156fe0ccbe338",
+                "urls": [
+                    {"url": "https://github.com/graalvm/truffle.git",
+                        "kind": "git"},
+                    {"url": "https://curio.ssw.jku.at/nexus/content/repositories/snapshots", "kind": "binary"},
+                ]
+            },
+        ],
+    },
+    "libraries": {
+        "RUBY_COMPLETE": {
+            "path": "maven/jruby-complete/target/jruby-complete-graal-vm.jar",
+            "sha1": "NOCHECK",
+            "optional":"true",
+            "license": "EPL"
+        },
+        "RUBY_TRUFFLE": {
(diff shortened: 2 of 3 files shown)


mx.jruby/mx_jruby.py
```diff
@@ -0,0 +1,58 @@
...
+def build(args):
+    mx.log('Building JRuby')
+
+    rubyDir = _suite.dir
+
+    # HACK: since the maven executable plugin does not configure the
+    # java executable that is used we unfortunately need to append it to the PATH
+    javaHome = os.getenv('JAVA_HOME')
+    if javaHome:
+        os.environ["PATH"] = os.environ["JAVA_HOME"] + '/bin' + os.pathsep + os.environ["PATH"]
```
gilles-duboscq: Looks like something that should be done in `run_maven`

mx.jruby/suite.py
```diff
@@ -0,0 +1,83 @@
...
+      "sourceDirs": ["."],
+      "dependencies": [
+      ],
+      "javaCompliance" : "1.7",
+    },
+
+    # truffle/src/main/java
+    "truffle/src/main/java" : {
+      "subDir" : "",
+      "sourceDirs" : ["."],
```
woess: I think you can use `"truffle"` as project name and `"sourceDirs" : ["src/main/java"]`.

mx.jruby/mx_jruby.py
```diff
@@ -0,0 +1,58 @@
...
+        contents = function(contents)
+        open(filename, 'w').write(contents)
+
+#    # Repository and commit of JRuby we want to build
+#    commit = open(_rubyVersionFile()).read().strip()
+
+    # Truffle version
+
+    truffle = mx.suite('truffle')
+    truffle_commit = subprocess.check_output(
```
gilles-duboscq: This should be done through the `VC` abstractions from mx (i.e., `truffle_commit = truffle.vc.parent(truffle.dir)`)

mx.jruby/mx_jruby.py
```diff
@@ -0,0 +1,58 @@
...
+        ['git', 'log', '-n', '1', '--format=%H'], cwd=truffle.dir).strip()
+
+    mx.run_mx(['build'], suite=truffle)
```
gilles-duboscq: recursive mx calls are in general discouraged: only do this as a last resort

mx.jruby/suite.py
jtulach: Maybe I could completely remove the projects and keep just the distribution? The projects are only useful for complication and for IDE support. But compilation is done by Maven and we don't need anything special to support IDEs - they already understand Maven structure.

gilles-duboscq:
In general, overriding `build` and `clean` is not a good idea: it breaks composing the suite with other suites.

The recommended way of achieving this is to make the thing you want to build a `mx.Dependency`, then you can implement `Dependency.getBuildTask` which returns a `mx.BuildTask`. You'll return an implementation of `BuildTask` that implements `BuildTask.build` and `BuildTask.clean` using the appropriate maven commands.
To get a `mx.Dependency` in the system, you can for example have a project in your `suite.py` with a `class` attribute: when such projects are loaded, this class name is looked up in your suite extension (here `mx_jruby.py`) and used as the class for the object representing this project. This class should be a subclass of `mx.Project`.

An example of this is how HotSpot is build in the jvmci suite: in `suite.py` you can see the [`hotspot`](http://hg.openjdk.java.net/graal/graal-jvmci-8/file/3848113e3e6b/mx.jvmci/suite.py#l243) project, as a `class` attribute set to `HotSpotProject`. This class is found in [`mx_jvmci.py`](http://hg.openjdk.java.net/graal/graal-jvmci-8/file/3848113e3e6b/mx.jvmci/mx_jvmci.py#l852) and in particular it overrides `getBuildTask` to return a [`HotSpotBuildTask`](http://hg.openjdk.java.net/graal/graal-jvmci-8/file/3848113e3e6b/mx.jvmci/mx_jvmci.py#l919) which handles the details of building and cleaning hotspot.

This doesn't really answer the question of the resulting distributions but i guess we'd need some kind of `JARDistribution` that simply copies an existing jar (the one produced by maven).

mx.jruby/suite.py
eregon: But then it's confusing with the imported `Truffle` project I guess.

gilles-duboscq:
Regarding the distribution, i think we could add a `class` attribute to those as well.

eregon:
> The goal of this pull request is to provide an mx facade that delegates execution to existing mvn goals, but behaves like a normal mx from outside and thus simplifies including of JRuby in other mx based projects.

Actually, I would be interested in compilation by `mx` not depending on Maven, because it's much faster and more reliable when configured correctly. But this is of course conflicting with this PR...

jtulach:
Compilation with `mx` is not really conflicting, but certainly out of scope of what I am trying to do. If we want 100% compatibility with the way JRuby is currently built, then I am reluctant to compile with `mx`. Of course, if you guys want to use `mx` as primary build tool, then including in other `mx` projects would be piece of cake and we could close this PR. But I assume wrapper around `mvn` will be more gentle approach.

mx.jruby/suite.py
woess: I don't understand. What `Truffle` project?

mx.jruby/mx_jruby.py
```diff
@@ -0,0 +1,57 @@
...
+    mx.run_mx(['maven-install'], suite=truffle)
+
+    # Build jruby-truffle and
+
+    mx.run_maven(['--version'], nonZeroIsFatal=False, cwd=rubyDir)
+
+    mx.run_maven(['-DskipTests', '-Dtruffle.version=' + truffle_commit], cwd=rubyDir)
+    mx.run_maven(['-Pcomplete', '-DskipTests', '-Dtruffle.version=' + truffle_commit], cwd=rubyDir)
+#    mx.run(['zip', '-d', 'maven/jruby-complete/target/jruby-complete-graal-vm.jar', 'META-INF/jruby.home/lib/*'], cwd=rubyDir)
+    mx.run(['bin/jruby', 'bin/gem', 'install', 'bundler', '-v', '1.10.6'], cwd=rubyDir)
```
eregon: Why installing bundler here, on every build?

mx.jruby/suite.py
eregon: the `truffle` (API) imported suite.

eregon: In any case, the projects do not seem useful at the moment if the whole build is done with Maven.

eregon:
@jtulach Right, let's begin with delegating to Maven and then maybe progressively adopt `mx`, as discussed with @gilles-duboscq.

mx.jruby/suite.py
```diff
@@ -23,14 +23,12 @@
 
     "jruby-truffle" : {
       "subDir" : "",
+      "class" : "MavenProject",
```
jtulach: The `MavenProject` and `MavenBuildTask` approach seems to be working quite well. The next step is to do something with the distribution. @gilles-duboscq, please push me in the right direction.

gilles-duboscq: Ok, for the distribution i'll add support for the `class` attribute in mx for distribution and then you can do the same an create something that looks like a `JARDistribution` that just copies the jar created by maven. I'll have a quick to see if it would help to make an abstract `JARDistribution` that you can extend that has all the "consumption" aspects of `JARDistribution` but none of the "creation" aspects.

jtulach:
The goal of this change is to create an `mx` distribution that will contain the whole content of a `jre/language/ruby/` directory as is currently in the [OTN build](http://www.oracle.com/technetwork/oracle-labs/program-languages/downloads/index.html). It will be responsibility of you guys to maintain the content of this ZIP to contain everything you need.

mx.jruby/suite.py
```diff
@@ -0,0 +1,88 @@
...
+                "urls": [
+                    {"url": "https://github.com/graalvm/truffle.git",
+                     "kind": "git"},
+                    { "url" : "https://curio.ssw.jku.at/nexus/content/repositories/snapshots", "kind" : "binary" },
+                ]
+            },
+    ],
+  },
+  "libraries": {
+    "RUBY_COMPLETE": {
```
eregon: This should be called JRUBY_CORE

jtulach: I've changed the path to jruby-complete JAR at the end.

jtulach:
The current version seems to work OK for my `mx` build purposes. Please consider merging it.

mx.jruby/suite.py
```diff
@@ -0,0 +1,102 @@
...
+    },
+    "jruby-lib-jni" : {
+      "subDir" : "lib/jni",
+      "class" : "MavenProject",
+      "prefix" : "lib/jni/",
+      "dependencies": [
+        "truffle:TRUFFLE_API",
+        "truffle:TRUFFLE_DEBUG",
+    ],
+},
```
eregon: Indentation seems wrong here.

mx.jruby/suite.py
```diff
@@ -0,0 +1,102 @@
...
+    "suites": [
+            {
+                "name": "truffle",
+                "version": "551e8475af2fc8769bc3ead07c9156fe0ccbe338",
+                "urls": [
+                    {"url": "https://github.com/graalvm/truffle.git",
+                     "kind": "git"},
+                    { "url" : "https://curio.ssw.jku.at/nexus/content/repositories/snapshots", "kind" : "binary" },
+                ]
+            },
```
eregon: Can we have consistent indentation, i.e. 2 spaces everywhere?

mx.jruby/mx_jruby.py
```diff
@@ -0,0 +1,113 @@
...
+        self.javaCompliance = "1.7"
+        self.build = hasattr(args, 'build')
+        self.prefix = args['prefix']
+
+    def source_dirs(self):
+        return []
+
+    def output_dir(self):
+        dir = os.path.join(_suite.dir, self.prefix)
+        return dir[0:len(dir) - 1]
```
eregon: This sounds brittle, maybe `.rstrip('/')` instead?

mx.jruby/mx_jruby.py
```diff
@@ -0,0 +1,113 @@
...
+        truffle_commit = truffle.vc.parent(truffle.dir)
+
+        mx.run_mx(['maven-install'], suite=truffle)
+
+        # Ruby version
+        def apply_to_file(filename, function):
+            contents = open(filename).read()
+            contents = function(contents)
+            open(filename, 'w').write(contents)
+        apply_to_file(join(rubyDir, 'VERSION'), lambda version: 'graal-vm\n')
```
eregon: In this case, maybe a simple `open(filename, 'w').write("graalvm\n")` is enough?

mx.jruby/suite.py
jtulach: OK. See next comment...

mx.jruby/suite.py
jtulach: Reformatted using 4 spaces in c39b6da38364e9777d2efd56848867e4680c1a4f

eregon:
Looks good :+1:, please address my comments.
I think some names and descriptions can be improved, but I can do that later.

mx.jruby/mx_jruby.py
jtulach: Good idea: done in 3db914340f97ee9ac7ae7d6f75def6c1d6102e3a

mx.jruby/mx_jruby.py
jtulach: Yes, that seems simpler: 5359719f6e135cb16438558cbd28356fa3cfaf53

eregon:
LGTM!

PR #3814, opened 2016-04-20
Expose two Enumerator-related bugs

nightscape:
JRuby seems to have two bugs related to Enumerators yielding Arrays. They can be reproduced when comparing the following code snippets which correctly return `true` for MRI, but `false` on JRuby.

1. `<<` should be an alias for `.yield`
    ```ruby
    Enumerator.new {|y| y.yield([1])}.to_a == [[1]] # MRI: true, JRuby: true
    Enumerator.new {|y| y << [1]}.to_a ==  [[1]] # MRI: true, JRuby: false ([[[1]]] instead)
    ```

2. `map` on lazy Enumerators should not unwrap Arrays
    ```ruby
    Enumerator.new {|y| y.yield([1])}.map {|e| e}.to_a == [[1]] # MRI: true, JRuby: true
    Enumerator.new {|y| y.yield([1])}.lazy.map {|e| e}.to_a == [[1]] # MRI: true, JRuby: false ([1] instead)
    ```

I added corresponding specs, but I'm not sure if they're in the right place and format.

### DIFF
--- a/spec/ruby/core/enumerator/lazy/map_spec.rb
+++ b/spec/ruby/core/enumerator/lazy/map_spec.rb
@@ -5,4 +5,7 @@ require File.expand_path('../shared/collect', __FILE__)
 
 describe "Enumerator::Lazy#map" do
   it_behaves_like :enumerator_lazy_collect, :map
+  it "doesn't unwrap Arrays" do
+    Enumerator.new {|y| y.yield([1])}.lazy.to_a.should == [[1]]
+  end
 end
--- a/spec/ruby/core/enumerator/lazy/map_spec.rb
+++ b/spec/ruby/core/enumerator/lazy/map_spec.rb
@@ -5,6 +5,7 @@ require File.expand_path('../shared/collect', __FILE__)
 
 describe "Enumerator::Lazy#map" do
   it_behaves_like :enumerator_lazy_collect, :map
+
   it "doesn't unwrap Arrays" do
     Enumerator.new {|y| y.yield([1])}.lazy.to_a.should == [[1]]
   end
--- a/spec/ruby/core/enumerator/first_spec.rb
+++ b/spec/ruby/core/enumerator/first_spec.rb
@@ -1,0 +1,7 @@
+require File.expand_path('../../../spec_helper', __FILE__)
+
+describe "Enumerator#first" do
+  it "returns arrays correctly when calling #first (2376)" do
+    Enumerator.new {|y| y << [42] }.first.should == [42]
+  end
+end
--- a/spec/ruby/core/enumerator/yielder/append_spec.rb
+++ b/spec/ruby/core/enumerator/yielder/append_spec.rb
@@ -10,6 +10,13 @@ describe "Enumerator::Yielder#<<" do
     ary.should == [1]
   end
 
+  it "doesn't double-wrap Arrays" do
+    yields = []
+    y = enumerator_class::Yielder.new {|args| yields << args }
+    y << [1]
+    yields.should == [[1]]
+  end
+
   it "returns self" do
     y = enumerator_class::Yielder.new {|x| x + 1}
     (y << 1).should equal(y)


headius:
These are definitely still true and probably related to our block arg-processing logic.

We can merge this but until we fix it we'd need to tag them for JRuby.

nightscape:
So, [this is the commit](https://github.com/jruby/jruby/commit/cb25efe8052eea9d919cdb20d13a9678614f5307) that breaks `<<` vs `yield`. It doesn't contain a test so it's hard to see what the erroneous behavior without it was.
Does anyone remember?

nightscape:
Ok, so the corresponding PR for the above commit is [this one](https://github.com/jruby/jruby/pull/2458) and the [corresponding issue](https://github.com/jruby/jruby/issues/2376)
The functionality that gets fixed by the commit is the following:
```
Enumerator.new{|y| y << [42]}.first == [42] # Without the commit it seems to be 42
```
I'll add a test for this case as well and hopefully find an implementation that passes all tests.

spec/ruby/core/enumerator/lazy/map_spec.rb
```diff
@@ -5,4 +5,7 @@
 
 describe "Enumerator::Lazy#map" do
   it_behaves_like :enumerator_lazy_collect, :map
```
eregon: There should be an extra blank line here.

nightscape:
I added a spec for the bug that the commit fixes.
Does anyone have an idea how to properly fix this?

headius:
@nightscape We've realized for a while we need a rework of how Enumera{ble,tor} passes values through. The current structure tends to lose the incoming argument layout and not match it properly with parameter binding.

9.1.1 might be a good time for us to look into fixing this.

I'm curious: do you have an app or library failing to work because of the current bugs in Enumera{ble,tor}?

nightscape:
Yes, we have a part of our code where we use Enumerators to efficiently stream data through a calculation pipeline to keep memory consumption low. We were running on MRI without issues but moved over to JRuby (we've ported parts over to Spark in order to speed up things) and suddenly things were failing in strange ways. We're not using lazy enumerators anymore now and convert the Enumerator into a real array in order to workaround this issue.

headius:
@nightscape Were the problems you had only in lazy Enumerator or in regular Enumerator too?

nightscape:
Both. We ran into the two problems that I've mentioned in the first post. They almost "cancel each other out" but then again not really ;)

headius:
Ok, I'll try to prioritize this for the first (hopefully quick) maintenance release of 9.1. Thank you for the test cases...they will help us sort out exactly what to fix.

nightscape:
Cool, thanks a lot! If I can help with anything where one doesn't have to be an expert in JRuby internals let me know!

codeforkjeff:
To chime in, I encountered this bug in a data processing script I was trying to run on JRuby. 

I have an understanding of this problem being caused by splats and arg destructuring in /core/src/main/ruby/jruby/kernel/enumerator.rb (and possibly RubyYielder.java?) but haven't come up with a fix. If anyone has any suggestions for an approach to fixing this, i could try to implement it and submit a PR. But as it stands, it's a bit dizzying to puzzle through.

headius:
Sadly still broken in 9.1.3.0, even with arg fixes. We need a rework of Enumerable/Enumerator to fix this, I suspect.

headius:
This may simply be a bit of bad behavior in the `Yielder#<<` implementation. We do the re-wrapping dance...MRI does not.

```c
/* :nodoc: */
static VALUE
yielder_yield(VALUE obj, VALUE args)
{
    struct yielder *ptr = yielder_ptr(obj);

    return rb_proc_call(ptr->proc, args);
}

/* :nodoc: */
static VALUE
yielder_yield_push(VALUE obj, VALUE args)
{
    yielder_yield(obj, args);
    return obj;
}
```

This patch gets your examples working. I am checking tests now.

```diff
diff --git a/core/src/main/java/org/jruby/RubyYielder.java b/core/src/main/java/org/jruby/RubyYielder.java
index 6494e4d..1376db4 100644
--- a/core/src/main/java/org/jruby/RubyYielder.java
+++ b/core/src/main/java/org/jruby/RubyYielder.java
@@ -96,17 +96,13 @@ public class RubyYielder extends RubyObject {
     }
 
     @JRubyMethod(rest = true)
-    public IRubyObject yield(ThreadContext context, IRubyObject[]args) {
+    public IRubyObject yield(ThreadContext context, IRubyObject[] args) {
         checkInit();
         return proc.call19(context, args, Block.NULL_BLOCK);
     }
 
     @JRubyMethod(name = "<<", rest = true)
-    public IRubyObject op_lshift(ThreadContext context, IRubyObject[]args) {
-        if (args.length == 1 &&
-                args[0] instanceof RubyArray &&
-                ((RubyArray) args[0]).getLength() == 1)
-            args[0] = RubyArray.newArray(context.runtime, args[0]);
+    public IRubyObject op_lshift(ThreadContext context, IRubyObject[] args) {
         yield(context, args);
         return this;
     }
```

headius:
Stuffed my change into a PR at #4106. We'll see how it looks.

headius:
Well I realized after reading @nightscape's comment above, I realized that #4106 basically would just revert #2458, which fixed #2376. It's obvious the fix isn't quite right, but we regress a different way if we remove it.

headius:
Replacing our `Yielder` impl does not appear to fix the issue either. The problem seems to lie in the proc logic.

headius:
Ok, replacing both `Yielder#yield` and `Enumerable#first` makes it behave the way we expect, for all the original cases plus the one from #2376:

```ruby
class Enumerator::Yielder
    def yield(*args)
      @proc.call(*args)
    end
end

module Enumerable
  def first
    val = nil
    catch(:foo) do
      each do |x|
        val = x
        throw :foo
      end
    end
    val
  end
end

p Enumerator.new {|y| y.yield([1])}.to_a
p Enumerator.new {|y| y << [1]}.to_a
p Enumerator.new {|y| y.yield([1])}.map {|e| e}.to_a
p Enumerator.new {|y| y.yield([1])}.lazy.map {|e| e}.to_a
p Enumerator.new {|y| y.yield [1]}.first
```
Output
```
$ jruby blah.rb
[[1]]
[[1]]
[[1]]
[[1]]
[1]
```

This is more demonstration that our Java-based Enumerable has several hard-to-fix flaws left in it. I think we're going to need to move to a Ruby-based impl sooner rather than later.

headius:
This PR kinda became a place to discuss the fix, but I'd like to get these specs in. I'm going to merge, tag, and open a separate bug for actually fixing the issue.

PR #3815, opened 2016-04-20
[Truffle] accept expression and parameter names in SnippetNode#execute

pitr-ch:


truffle/src/main/java/org/jruby/truffle/language/SnippetNode.java
```diff
@@ -27,36 +27,37 @@
...
     }
 
     @ExplodeLoop
-    public Object execute(VirtualFrame frame, Object... arguments) {
+    public Object execute(VirtualFrame frame, String expression, Object... arguments) {
         if (directCallNode == null) {
             CompilerDirectives.transferToInterpreter();
 
+            this.expression = expression;
+            parameters = new String[arguments.length / 2];
```
eregon: assert arguments.length is even first maybe?

truffle/src/main/java/org/jruby/truffle/language/SnippetNode.java
```diff
@@ -67,13 +68,25 @@ public Object execute(VirtualFrame frame, Object... arguments) {
...
             }
         }
 
+        if (arguments.length != parameters.length * 2) {
+            CompilerDirectives.transferToInterpreter();
+            throw new UnsupportedOperationException(
+                    "number of arguments doesn't match number of parameters");
+        }
+
+        assert this.expression.equals(expression);
```
eregon: `==` since they should be String literals (therefore interned)

truffle/src/main/java/org/jruby/truffle/language/SnippetNode.java
```diff
@@ -88,13 +101,8 @@ public Object execute(VirtualFrame frame, Object... arguments) {
...
 
-        if (arguments.length != parameterFrameSlots.length) {
-            CompilerDirectives.transferToInterpreter();
-            throw new UnsupportedOperationException("number of arguments doesn't match number of parameters");
-        }
-
         for (int n = 0; n < parameters.length; n++) {
-            parentFrame.setObject(parameterFrameSlots[n], arguments[n]);
+            parentFrame.setObject(parameterFrameSlots[n], arguments[2 * n + 1]);
         }
```
eregon: Can we extract this to a separate method and only annotate that one with ExplodeLoop?

eregon:
Exactly what I had in mind, perfect!

truffle/src/main/java/org/jruby/truffle/language/SnippetNode.java
pitr-ch: I was thinking about cases where some expression could be shared between similar specialisations with possibly some small changes. It seemed not important in assert.

truffle/src/main/java/org/jruby/truffle/language/SnippetNode.java
pitr-ch: Sure.

truffle/src/main/java/org/jruby/truffle/core/array/ArrayNodes.java
```diff
@@ -3372,17 +3372,20 @@ public DynamicObject sortVeryShortObject(VirtualFrame frame, DynamicObject array
...
-            return snippet.execute(frame, getSize(array), block);
+                @Cached("new()") SnippetNode snippet) {
+
+            return snippet.execute(
+                    frame,
+                    "sorted = dup; Rubinius.privately { sorted.isort_block!(0, right, block) }; sorted",
+                    "right",
+                    getSize(array),
+                    "block",
+                    block);
```
chrisseaton: I would format these pairs onto the same line, as `"key", value`.

truffle/src/main/java/org/jruby/truffle/language/SnippetNode.java
eregon: I think it can help to detect wrong usage (say with some "a" + "b").

PR #3825, opened 2016-04-25
Fix Net::HTTP nil body bug

chrisberkhout:


headius:
Huh, I can't even find this code in current `net/http.rb`.

headius:
It appears based on ruby/ruby@b1a0509b5465ce77f52e0384159237889a8d60ec this functionality was moved to `response.rb` and additional guards were added to ensure empty bodies don't get decompressed. If you can reproduce this, it would be good to know if a Ruby 2.2-compatible runtime passes your case.

chrisberkhout:
@headius The problem is not in the JRuby 9k series, only in JRuby 1.x (including the latest 1.7.25), using the 1.9 stdlib (later stdlib versions do fix it).

I noticed there are multiple versions of the stdlib in the 1.7.25 code. My install used the 1.9 versions, and I'm not sure how the others are used (looked for documentation and command line options, but didn't find any).

My intention with the PR is that the upcoming 1.7.26 release doesn't have this no method on nil issue. I asked on IRC and understood that the PR should merge into the branch `jruby-1_7` (as this one requests) to be in the next 1.x release.

headius:
@chrisberkhout Thank you for clarifying about 9k!

The 1.8 and 2.0 stdlibs in JRuby 1.7.x are used when running with `--1.8` or `--2.0` modes. The 2.0 stdlib probably doesn't have its own copy of `net/http`. We could fix 1.8, but it's a pretty low priority.

Thanks for the PR!

chrisberkhout:
@headius No worries.

I checked again, and the weird response we had is correctly processed by 1.8 or 2.0. So that's it! :)

headius:
@chrisberkhout Thanks for confirming!

chrisberkhout:
@BanzaiMan I'm wondering, why wasn't this merged?

enebo:
@chrisberkhout I think @BanzaiMan thought it had already been merged.

chrisberkhout:
Cool. Thanks @enebo!

PR #3831, opened 2016-04-27
Fixes Fixnum#bitLength when the value is negative.

### LINKED ISSUE #3829: Fixnum#bit_length always returns 64 for negative number
According to [CRuby manual](http://docs.ruby-lang.org/en/2.1.0/Fixnum.html#method-i-bit_length), `Fixnum#bit_length` returns "the bit position of the highest bit which is different to the sign bit". In other words, check 0 bit position for negative numbers.

In [JRuby implementation](https://github.com/jruby/jruby/blob/master/core/src/main/java/org/jruby/RubyFixnum.java#L1195) it returns `64 - Long.numberOfLeadingZeros(value)`.

` Long.numberOfLeadingZeros(value)` returns 0 if `value` is negative, so the value is fixed to 64 in the cases.

### Environment
* jruby 9.0.4.0 (2.2.2) 2015-11-12 b9fb7aa OpenJDK 64-Bit Server VM 25.77-b03 on 1.8.0_77-b03 +jit [linux-amd64] on CentOS 7 x64
* JRuby 9.0.5.0 (x64) on Windows7 x64

### Expected Behavior
```rb
p (-1).bit_length # => 0
```
### Actual Behavior
```rb
p (-1).bit_length # => 64
```

lucasallan:
Fixes #3829

### DIFF
--- a/core/src/main/java/org/jruby/RubyFixnum.java
+++ b/core/src/main/java/org/jruby/RubyFixnum.java
@@ -1192,7 +1192,12 @@ public class RubyFixnum extends RubyInteger implements Constantizable {
 
     @JRubyMethod
     public IRubyObject bit_length(ThreadContext context) {
-        return context.runtime.newFixnum(64 - Long.numberOfLeadingZeros(value));
+        long tmpValue = value;
+        if (value < 0) {
+            tmpValue = Math.abs(value);
+        }
+
+        return context.runtime.newFixnum(64 - Long.numberOfLeadingZeros(tmpValue));
     }
 
     @Override


kares:
Hey Lucas! seems a bit wrong isn't it ? ... maybe some test-cases would help : 
```
2.3.0 :002 > 112.bit_length
 => 7 
2.3.0 :003 > -112.bit_length
 => 7 
```

lucasallan:
@kares 

```ruby
>> 112.bit_length
=> 7
>> -112.bit_length
=> 0
```

It seems to be working for me - I'll send a PR with tests to rubyspecs.

kares:
`-112` should **not** have a bit length of `0` ... if it does at your MRI its platform dependent or a bug.
... if you think of it it would be a pretty weird outcome to expect all `< 0` to have a bit length of **0**

lucasallan:
You're right, it seems pretty weird - I was assuming MRI was right - my bad.

kares:
Great, thanks Lucas.

jkr2255:
Thanks for improvement, but for some values still return different value from MRI.

In MRI implementation [uses bit inversion](https://github.com/ruby/ruby/blob/trunk/numeric.c#L4280), `(-(2**x)).bit_length`becomes `x` in MRI (and `x + 1` for this implementation)

lucasallan:
@jkr2255 It will be fixed by https://github.com/jruby/jruby/pull/3833

PR #3832, opened 2016-04-27
Fix a bug where NameError#receiver raises an ArgumentError after #to_s is called

yuki24:
If the `NameError#to_s` is called before `NameError#receiver` is called, the NameError object will [lose the source of the receiver object](https://github.com/jruby/jruby/blob/b856557/core/src/main/java/org/jruby/RubyNameError.java#L223) and the `NameError#receiver` method will raise an ArgumentError when it shouldn't:

```ruby
error = (1.foo rescue $!)
error.receiver # => 1
error.to_s     # => NameError: ... <= This call replaces the internal `message` object
error.receiver # => ArgumentError: no receiver is available
```

This commit changes the initializer to retrieve the receiver object so that the behavior of the method will be consistent.


headius:
Nice find, thank you!

PR #3833, opened 2016-04-27
Fixnum#bit_length - uses bit inversion when the value is negative

lucasallan:
Making sure the behaviour matches MRI https://github.com/ruby/ruby/blob/trunk/numeric.c#L4280

Thanks @jkr2255


headius:
Nice!

headius:
If you haven't already, can you add a ruby/spec for this?

PR #3836, opened 2016-04-29
[Truffle] Experiment: try to make the truffle module Java 8+

eregon:
[Work in progress, do not merge]
This PR intends to evaluate the effect of having Java compatibility
for JRuby at 1.7 and for JRuby+Truffle at 1.8.

cc @jruby/truffle


eregon:
This also shaves ~1h from the total Travis time due to building truffle only for truffle tests.

core/src/main/java/org/jruby/Ruby.java
```diff
@@ -912,6 +912,13 @@ public JRubyTruffleInterface getTruffleContext() {
     private JRubyTruffleInterface loadTruffle() {
         Main.printTruffleTimeMetric("before-load-context");
 
+        String javaVersion = System.getProperty("java.version");
+        String[] parts = javaVersion.split("\\.");
+        if (Integer.valueOf(parts[1]) < 8) {
```
kares: wasn't Java 9 switching from '1.8' to '9.0' **java.version** format ?

eregon: @kares Yeah I was wondering as well, do you have any link?

eregon: http://openjdk.java.net/jeps/223 seems to be the proposed change.

eregon:
```bash
jjs> java.lang.System.getProperty("java.version")
9-ea
```
Indeed, thanks for the catch!

eregon:
@headius @enebo @kares @mkristian Would you be OK with this?
This would make the `truffle` module (in Maven terms) Java 8+.
One can easily exclude it if they want to build/test with Java 7 (as the CI does).

For distribution, I do not know what is the current process. Is a Java 7 `javac` used? If that's the case then everything but truffle should be compiled that way and then switch to a Java 8 `javac` to compile truffle.

eregon:
I added another commit which enables the `truffle` profile and module only when using a JDK 8 or newer. This allows the user with a default jdk 7 still be able to build core JRuby without any change.
For distribution, we of course want the truffle module to be included, so `-Ptruffle` should be specified to guarantee it's built (or it will error out).

enebo:
Thanks. I would like to keep building core with 7 since then we catch 8
APIs slipping in.

eregon:
@enebo Tavis keeps building with 7 as before for most of the tasks.

Do you mean for the distribution?
It can be done although it's easier with a Java 8 and Travis should already catch missing APIs. Something like this seems to work quite nicely (`package` should be replaced by the appropriate phase of course):
```bash
switch to java7
./mvnw package
switch to java8
./mvnw package -pl truffle
```

enebo:
I am ok with using 8 for dist so long as Travis and I can routinely build
core with 7.  Sounds like you got it covered.

eregon:
@enebo Yep, that should work transparently since the truffle module is only considered if there is a Java 8 :smiley:

eregon:
@headius @enebo I plan to merge this today. Please raise your voice if there is any objection :smiley:

eregon:
(the build is green on Travis BTW, just failing on Windows as it already was)

PR #3839, opened 2016-05-02
[packaging] ensure licensing info is included in artifacts

busbey:
* jars should have them in META-INF
* bin tarball/zip should have them in top level

Right now there's still a gap, namely aggregate NOTICE files for those artifacts that bundle 3rd party ASL licensed dependencies. I've done this before in other maven-based projects, but navigating JRuby's build is taking me a bit of time.

### DIFF
--- a/core/pom.rb
+++ b/core/pom.rb
@@ -251,6 +251,12 @@ project 'JRuby Core' do
       target_path '${project.build.sourceDirectory}'
       filtering 'true'
     end
+
+    resource do
+      directory '${project.basedir}/..'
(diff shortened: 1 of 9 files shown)


mkristian:
we shade some dependencies for jruby-core. jruby-stdlib does include complete jars from 3rd parties and then there are those aggregated artifacts: jruby-complete, jruby-jars (which is a gem), jruby-dist and maybe main jruby artifact. and finally there is lib/jruby.jar which is jruby-core with all its dependencies packed into one jar.

mkristian:
travis looks good to me :) - thanks

PR #3853, opened 2016-05-05
[Truffle] Change sets that work with older mx

jtulach:
It is important for my mx changes, to get revision
929664461bec45d1510c9b1613a8a21d47fa85ce
into JRuby repository. That one is known to work. 

It would already be in the repo, but earlier today I squashed my commits and I should have not done that.


eregon:
Closed in favor of a temporary branch to avoid duplicate commits and complex merges:
https://github.com/jruby/jruby/tree/truffle-head-mx-tmp

PR #3864, opened 2016-05-09
Fixed a bug where NameError#receiver didn't return receiver when it's caused by Struct#[]

yuki24:
When a NameError is raised from `Struct#[]`, the `NameError#receiver` method should return the receiver object rather than raising `ArgumentError: no receiver is available`.

Actual:

```ruby
(Struct.new(:does_exist).new(nil)[:doesnt_exist] rescue $!).receiver
# => ArgumentError: no receiver is available
```

Expected:

```ruby
(Struct.new(:does_exist).new(nil)[:doesnt_exist] rescue $!).receiver
# => #<struct does_exist=nil>
```

### DIFF
--- a/core/src/main/java/org/jruby/RubyStruct.java
+++ b/core/src/main/java/org/jruby/RubyStruct.java
@@ -498,7 +498,7 @@ public class RubyStruct extends RubyObject {
     }
 
     private RaiseException notStructMemberError(String name) {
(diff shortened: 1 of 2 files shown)


headius:
Nice, thanks! I will merge this and wire the test up to our suite, but I'll also modify it to use the same sprintf-style NameError logic that the rest of the system uses rather than a custom error method.

PR #3879, opened 2016-05-12
fix quoted paths

### LINKED ISSUE #3878: Can't run rails if jruby's path contains whitespaces
### Environment

jruby 9.1.1.0-SNAPSHOT (2.3.0) 2016-05-11 9a2310a Java HotSpot(TM) 64-Bit Server VM 25.25-b02 on 1.8.0_25-b18 +jit [mswin32-x86_64]

### Expected Behavior

no exception

### Actual Behavior

rails is trying to load Gem.ruby path which is double quoted if the path contains whitespaces
then withExeSuffix(fname) method returns false even if the suffix is present
and it ends up with paths like this one "C:/jruby test/jruby/bin/jruby.exe".exe

https://github.com/jruby/jruby/blob/master/core/src/main/java/org/jruby/util/ShellLauncher.java#L360

```bash
C:\jruby test\jruby\bin\jruby.exe -S rails s

Errno::ENOENT: No such file or directory - "C:/jruby test/jruby/bin/jruby.exe"
           _exec_internal at org/jruby/RubyKernel.java:1635
                     exec at uri:classloader:/jruby/kernel/kernel.rb:17
  block in exec_app_rails at C:/jruby test/jruby/lib/ruby/gems/shared/gems/railties-4.2.5.2/lib/rails/app_rails_loader.rb:41
                     loop at org/jruby/RubyKernel.java:1290
           exec_app_rails at C:/jruby test/jruby/lib/ruby/gems/shared/gems/railties-4.2.5.2/lib/rails/app_rails_loader.rb:34
                    <top> at C:/jruby test/jruby/lib/ruby/gems/shared/gems/railties-4.2.5.2/lib/rails/cli.rb:5
                  require at org/jruby/RubyKernel.java:944
                   (root) at C:/jruby test/jruby/lib/ruby/stdlib/rubygems/core_ext/kernel_require.rb:1
                    <top> at C:/jruby test/jruby/lib/ruby/stdlib/rubygems/core_ext/kernel_require.rb:69
                     load at org/jruby/RubyKernel.java:962
                    <top> at C:/jruby test/jruby/bin/rails:23
```

ahorek:
fixes #3878

### DIFF
--- a/core/src/main/java/org/jruby/util/ShellLauncher.java
+++ b/core/src/main/java/org/jruby/util/ShellLauncher.java
@@ -389,6 +389,9 @@ public class ShellLauncher {
 
     private static File findPathFile(Ruby runtime, String fname, String[] path, boolean isExec) {
         File pathFile = null;
+        if (Platform.IS_WINDOWS && fname.startsWith("\"") && fname.endsWith("\"")) {
+            fname = fname.substring(1, fname.length() - 1); // remove double quotes if present
+        }
         boolean doPathSearch = filenameIsPathSearchable(fname, isExec);
         if (doPathSearch) {
             for (String fdir: path) {


headius:
Well, you're on a roll here :-)

We have generally recommended that JRuby always be installed in a path without spaces, since there's other Ruby libraries that don't like it. But if we can improve some cases, I'm all for it.

headius:
This seems pretty safe, doesn't it @enebo? Need to do some verification.

I'm a little curious why a path is getting in here that has quotes though. Is this a Windows shell quirk of some kind?

enebo:
Yeah we must be passing through the quotes as $0 perhaps.  I do not see the harm in this though.   On unix you can use " in file name but I highly doubt it is valid on windows.  Even if it is that would be crazy to use.

headius:
I think I remember reading that `"` is one of the explicitly-forbidden characters in filenames/paths.

On UNIX, the shell would be removing the quotes for us when it parses out arguments.

I do feel like there's something else going on here that is the actual root cause.

headius:
@ahorek Maybe you can do a little research to figure out how this quoted string is getting into the runtime?

enebo:
Windows CMD does not strip any quoting or process globs (we even have an open bug since we are not handling this stuff on windows).  Perhaps we wrote something in launcher/.bat assuming windows would handle this for us?

headius:
@enebo That would make some sense. Maybe all quoted paths passed into our Windows launchers are wrong?

ahorek:
@headius 
it's hardcoded in rubygems
https://github.com/jruby/jruby/blob/master/lib/ruby/stdlib/rubygems.rb#L854
https://github.com/rubygems/rubygems/blob/master/History.txt#L2558
according to the history Gem.ruby path is escaped because there were some issues with extensions compilation on windows

headius:
@ahorek Ahh that is interesting. So perhaps this quote-stripping is something we actually do need to do.

We should look at MRI's logic for process-launching on Windows and see if they strip this stuff too.

headius:
I did not see any logic at a glance, but there's a lot of win32-specific process stuff.

PR #3886, opened 2016-05-12
Collect all DNS sections by calling to each_resource instead of each_answer

ereslibre:
Disclaimer: same pull request on mruby impl (https://github.com/ruby/ruby/pull/797), and bug open (https://bugs.ruby-lang.org/issues/12372). Not sure if I should report here a bug for the stdlib. For completeness:

A call to `Message#each_answer` does only include answers, however there are also two more kinds of
responses: Authority and Additional. Iterating through `Message#each_resource` allows us to retrieve
all information from the `Message` object.

This is specially important if, for example, we are trying to retrieve nameservers for a certain domain.

```
require 'resolv'
=> true
dns = Resolv::DNS.new(:nameserver => ['199.19.56.1']) # a0.org.afilias-nst.info
=> #<Resolv::DNS:0x3578436e ...>
puts dns.getresources('jruby.org', Resolv::DNS::Resource::IN::NS).map(&:name)
=> nil
```

`nil` is returned. This, in a terminal fashion is equivalent to:

```
~ > dig ns jruby.org @a0.org.afilias-nst.info
;; AUTHORITY SECTION:
jruby.org.      86400   IN  NS  ns25.domaincontrol.com.
jruby.org.      86400   IN  NS  ns26.domaincontrol.com.
```

There is indeed a response, but in the authority section, that is right now not collected by `Resolv::DNS::getresource(s)`.

### DIFF
--- a/lib/ruby/stdlib/resolv.rb
+++ b/lib/ruby/stdlib/resolv.rb
@@ -37,10 +37,10 @@ end
 # * /etc/nsswitch.conf is not supported.
 
 class Resolv
-
+  
   ##
   # Tests whether we're running on Windows
-
+  
   WINDOWS = /mswin|cygwin|mingw|bccwin/ =~ RUBY_PLATFORM || ::RbConfig::CONFIG['host_os'] =~ /mswin/
 
   ##
@@ -2856,3 +2856,4 @@ class Resolv
   AddressRegex = /(?:#{IPv4::Regex})|(?:#{IPv6::Regex})/
 
 end
+
--- a/lib/ruby/stdlib/resolv.rb
+++ b/lib/ruby/stdlib/resolv.rb
@@ -37,10 +37,10 @@ end
 # * /etc/nsswitch.conf is not supported.
 
 class Resolv
-  
+
   ##
   # Tests whether we're running on Windows
-  
+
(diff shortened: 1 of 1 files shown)


headius:
@ereslibre Wow, thanks for the footwork on this one!

Our stdlib is a lightly-patched duplicate of MRI's, but we generally don't patch stdlib unilaterally, so we'll wait to see if MRI accepts this. When that happens, ping again and we'll be happy to merge this into JRuby.

One minor nit: there's unrelated whitespace changes in here. Can you eliminate those?

ereslibre:
@headius Amazing timing. Of course, I will get rid of those space changes, sorry ! Thanks for your wonderful work !

headius:
@ereslibre I pinged the MRI issue to see if we can get some movement.

ereslibre:
Thank you @headius ❤️

headius:
Doesn't look like MRI's going to get back to us in time for 9.1.3.0. If they don't respond for 9.1.4.0 and we're still happy with this change, we'll just merge it.

ereslibre:
Thank you again @headius.

headius:
MRI finally merged your fix! I've gone with their patch (yours sans whitespace changes) in 		9187c5e.

Issue #3480, opened 2015-11-20
Bundle did_you_mean gem


headius:
This is from https://bugs.ruby-lang.org/issues/11252. See #3479 for full Ruby 2.3 checklist.

The `did_you_mean` gem uses `TracePoint` to intercept `NameError` (I think) and report a more useful message, guessing at what error you might have meant.

JRuby implements `TracePoint`, but most tracing is disabled by default for performance. Exception tracing might be ok, though...we need to confirm. And we may need to modify the "tracing is disabled" warning when the requested trace is 100% ok (if that's the case for all tracing `did_you_mean` uses).

I started the process by adding the beta3 version of the gem to lib/pom.rb, but it didn't seem to be able to find it via our proxies. Perhaps the proxy doesn't support beta gems?

```diff
diff --git a/lib/pom.rb b/lib/pom.rb
index b9a110a..03202f6 100644
--- a/lib/pom.rb
+++ b/lib/pom.rb
@@ -20,7 +20,8 @@ default_gems =
    ImportedGem.new( 'psych', '2.0.15' ),
    ImportedGem.new( 'json', '${json.version}' ),
    ImportedGem.new( 'jar-dependencies', '${jar-dependencies.version}' ),
-   ImportedGem.new( 'racc', '${racc.version}')
+   ImportedGem.new( 'racc', '${racc.version}' ),
+   ImportedGem.new( 'did_you_mean', '1.0.0.beta3' )
   ]
 
 project 'JRuby Lib Setup' do
```

yuki24:
The creator of the `did_you_mean` gem here. I've started working on this to help make this happen. `TracePoint` is actually no longer necessary, so it shouldn't be so hard to make it work on JRuby. Also, I've found a bug in `NameError#receiver` which is breaking some of the `did_you_mean` features. I'm working on a fix now and will send a pull request when it's ready.

enebo:
@yuki24 great.  Thanks for the update!

yuki24:
Ok, now the only change we need to add to JRuby is this change: https://bugs.ruby-lang.org/issues/11777. It's a little weird change as this is a special case just for the `did_you_mean` gem. I'm not actually sure if it makes sense to introduce this change to JRuby as it breaks the consistency of the behavior of the `#local_variables` method.

@headius @enebo WDYT?

headius:
[...]

PR #3890, opened 2016-05-15
Upgrade the version of the did_you_mean gem to 1.0.1

yuki24:
This might be trivial since it's not enable at this moment, but I just wanted to let you know that this version includes a fix for a bug where most of DYM's features didn't work on JRuby 9.1.0.0.

#3480


headius:
Thank you!

PR #3892, opened 2016-05-16
Enabled did_you_mean, install during build, and clear gem paths

### LINKED ISSUE #3480: Bundle did_you_mean gem

headius:
This enables did_you_mean and installs it during the build. Fixes #3480.

@mkristian I could not figure out how to make this work without being a default gem. We do not have any logic to preinstall gems that aren't default, do we?

This also includes the `Gem.clear_paths` hack to address too-early booting of RubyGems in embedded scenarios.

### DIFF
--- a/core/src/main/java/org/jruby/util/cli/Options.java
+++ b/core/src/main/java/org/jruby/util/cli/Options.java
@@ -218,7 +218,7 @@ public class Options {
     public static final Option<String> CLI_BACKUP_EXTENSION = string(CLI, "cli.backup.extension", "Backup extension for in-place ARGV files. Same as -i.");
     public static final Option<ProfilingMode> CLI_PROFILING_MODE = enumeration(CLI, "cli.profiling.mode", ProfilingMode.class, ProfilingMode.OFF, "Enable instrumented profiling modes.");
     public static final Option<Boolean> CLI_RUBYGEMS_ENABLE = bool(CLI, "cli.rubygems.enable", true, "Enable/disable RubyGems.");
-    public static final Option<Boolean> CLI_DID_YOU_MEAN_ENABLE = bool(CLI, "cli.did_you_mean.enable", false, "Enable/disable did_you_mean.");
+    public static final Option<Boolean> CLI_DID_YOU_MEAN_ENABLE = bool(CLI, "cli.did_you_mean.enable", true, "Enable/disable did_you_mean.");
     public static final Option<Boolean> CLI_RUBYOPT_ENABLE = bool(CLI, "cli.rubyopt.enable", true, "Enable/disable RUBYOPT processing at start.");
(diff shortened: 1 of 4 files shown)


mkristian:
@headius what is the advantage of preinstalled gems vs. default gem. a preinstalled gem is gone the moment you set GEM_HOME and GEM_PATH, i.e. you want to install gems with jruby-complete.jar.

headius:
@mkristian Not a great deal, I guess. Being preinstalled means you won't be able to require it if gems are disabled...that's the only real difference. But for a library being activated immediately, being preinstalled also boots much more of RG than we want.

MRI chose not to make this a default gem, and they are actively trying to move away from default gems. I don't know their reasons, though.

headius:
I'm open for a vote on what to do from here... merge this now to master as a default gem (and probably remove the `gem` and `Gem.clear_paths` calls), or wait and do something different?

@enebo @mkristian @kares

mkristian:
+1 merge,

we have a number of default gems and the question preinstalled gem vs. default gem is a more general topic

headius:
Might as well let it bake now.

yuki24:
I'm very honored to see the `did_you_mean` gem become part of JRuby. Thank you for making this happen! Please let me know if there's anything I can help. I'm always happy to make JRuby even better!

headius:
@yuki24 Thank you for working through issues with us!

PR #3896, opened 2016-05-17
Test new recursion

headius:
Dunno if anyone wants to review this (e.g. @enebo or @kares) but it appears to have passed all tests and eliminates some really ugly misfeatures of the original port.

Note that I have not removed the old impl in case an external library uses it.


headius:
I have squashed all commits into one, since they didn't really add anything interesting to history.

kares:
looking good to me ... the new logic seems much easier to follow :+1:

PR #3898, opened 2016-05-17
Disable the AddLocalVarLoadStore pass to fix #3891.

### LINKED ISSUE #3891: Changes to local variables not visible to other threads even when protected by a lock
In JRuby 1.7 (all versions we have tested, including 1.7.25), the following script runs forever. The distilled test-case decrements a variable from two threads, while the main thread waits for all decrements have been applied. The decrements are then checked to ensure that each decrement call has seen a unique value of the counter.

The example is obviously nonsensical, but distilled from a much larger program where the counter was decremented in multiple threads, and then when the count reached zero, the code would continue. Tracking the intermediate values was added in order to facilitate debugging.

```ruby
loop do
  print '.'

  remaining = 200
  queue = Queue.new
  lock = Mutex.new

  callbacks = Array.new(remaining) do |i|
    lambda do
      local_remaining = nil
      lock.lock
      begin
        remaining -= 1
        local_remaining = remaining 
      ensure
        lock.unlock
      end
      queue.push(local_remaining)
    end
  end

  Thread.start { callbacks[0, 100].each(&:call) }
  Thread.start { callbacks[100, 100].each(&:call) }

  counts = callbacks.map { queue.pop }

  if counts.uniq != counts
    puts "Duplicate values registered #{counts}"
    exit 1
  end
end
```

In JRuby 9k (tried 9.0.5.0 and 9.1.0.0-SNAPSHOT (2.3.0) 2016-05-03 9ee8643) the script terminates quickly with duplicates in the returned counts.

```
.Duplicate values registered [199, 199, 198, 198, 197, 196, 195, 194, 193, 192, 191, 190, 189, 188, 187, 186, 186, 185, 185, 184, 184, 183, 183, 182, 182, 181, 180, 180, 179, 179, 178, 177, 176, 175, 174, 173, 172, 171, 170, 169, 168, 167, 166, 165, 164, 163, 162, 161, 160, 159, 158, 157, 156, 155, 154, 153, 152, 151, 150, 149, 148, 147, 146, 145, 144, 143, 142, 141, 140, 139, 138, 138, 137, 136, 135, 134, 133, 132, 132, 131, 131, 130, 130, 129, 129, 128, 127, 126, 125, 124, 123, 122, 121, 120, 119, 118, 118, 117, 117, 116, 116, 115, 114, 113, 113, 112, 111, 110, 109, 108, 107, 106, 105, 104, 103, 102, 101, 101, 100, 100, 99, 99, 98, 98, 97, 97, 96, 95, 95, 94, 94, 93, 92, 91, 90, 90, 89, 89, 88, 88, 87, 87, 86, 86, 85, 84, 83, 82, 81, 80, 79, 78, 77, 76, 75, 74, 74, 73, 73, 72, 71, 70, 69, 68, 67, 67, 66, 66, 65, 65, 64, 64, 63, 62, 61, 60, 59, 59, 58, 57, 56, 55, 54, 54, 53, 52, 51, 51, 50, 50, 49, 49, 48, 48, 47, 46, 45, 44, 43, 42]
```

The output contains a `.` for each attempt, and contains a list of each observed value of the counter. As you can see in the output above, the values 199, 186, etc. are repeated (an no value below 42 is present).

Somehow, the updates to the value of `remaining` is not visible to other threads as expected. Notice that `remaining` is only read (and copied into `local_remaining`) inside the critical section.

Changing the script slightly removes the problem. For example:

 - Initializing `remaining` to `[200]` and changing all references to read and write the first element of the array.
 - Replacing `lock.lock` and `lock.unlock` with `lock.synchronize`.
 - Wrapping `remaining -= 1` in `proc { ... }.call`.
 - Changing `remaining` to an instance variable `@remaining`.

To eliminate any issues with `Queue` and `Mutex` we also reproduced the same issues with `java.util.concurrent.locks.ReentrantLock` and `java.util.concurrent.ArrayBlockingQueue`.

Our conclusion is that updates to local variables are not guaranteed to be visible to other threads, even if updates are protected with critical sections.

headius:
This was a good experiment, but we're not properly ensuring the
heap variables are being loaded live when needed, causing examples
like that in #3891 to fail to propagate changes across threads.
By implementing LocalVariable load/store logic in JIT and turning
off the "Add" pass, we basically revert heap vars to always being
read/written immediately, as in JRuby 1.7.25.

It may be possible to improve the pass so that it localizes the
loads and stores better and ensures we don't miss updates we
should see, but this commit will test whether the "nuclear option"
passes all our suites.

cc @subbuss @enebo

### DIFF
--- a/core/src/main/java/org/jruby/ir/IRManager.java
+++ b/core/src/main/java/org/jruby/ir/IRManager.java
@@ -30,7 +30,7 @@ import static org.jruby.ir.IRFlags.REQUIRES_DYNSCOPE;
 public class IRManager {
     public static final String SAFE_COMPILER_PASSES = "";
     public static final String DEFAULT_BUILD_PASSES = "";
-    public static final String DEFAULT_JIT_PASSES = "LocalOptimizationPass,OptimizeDelegationPass,DeadCodeElimination,AddLocalVarLoadStoreInstructions,OptimizeDynScopesPass,AddCallProtocolInstructions,EnsureTempsAssigned";
+    public static final String DEFAULT_JIT_PASSES = "LocalOptimizationPass,OptimizeDelegationPass,DeadCodeElimination,OptimizeDynScopesPass,AddCallProtocolInstructions,EnsureTempsAssigned";
     public static final String DEFAULT_INLINING_COMPILER_PASSES = "LocalOptimizationPass";
 
     private final CompilerPass deadCodeEliminationPass = new DeadCodeElimination();
--- a/core/src/main/java/org/jruby/ir/IRManager.java
+++ b/core/src/main/java/org/jruby/ir/IRManager.java
@@ -30,7 +30,7 @@ import static org.jruby.ir.IRFlags.REQUIRES_DYNSCOPE;
 public class IRManager {
     public static final String SAFE_COMPILER_PASSES = "";
     public static final String DEFAULT_BUILD_PASSES = "";
-    public static final String DEFAULT_JIT_PASSES = "LocalOptimizationPass,OptimizeDelegationPass,DeadCodeElimination,OptimizeDynScopesPass,AddCallProtocolInstructions,EnsureTempsAssigned";
+    public static final String DEFAULT_JIT_PASSES = "LocalOptimizationPass,DeadCodeElimination,OptimizeDynScopesPass,OptimizeDelegationPass,AddCallProtocolInstructions,EnsureTempsAssigned";
     public static final String DEFAULT_INLINING_COMPILER_PASSES = "LocalOptimizationPass";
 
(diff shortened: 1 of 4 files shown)


headius:
This is now working well, and the fixes here should probably not be squashed because there's other bits and pieces with important history.

I did discover #3906 and related minor issues in the process of getting this passing tests. With #3906 outstanding, there's a chance this PR might cause OptimizeDelegation to never work in the presence of keyword arguments, since we do not keep everything in JVM locals anymore.

headius:
@subbuss @enebo This should now be 100%. Do we want to go with this or try to fix ALVLS?

subbuss:
I think it is okay to merge this and also fix the bug .. and then independently figure out whether to enable the pass or not.

Issue #3901, opened 2016-05-18
1.7.21: getScriptSource -> isSymLink raises NullPointerException when script source is stdin


bg:
Hi,

Ever since JRuby 1.7.21, when you specify stdin explicitly as the filepath for the script source, which we do in order to execute a heredoc that takes command line arguments, it results in a NullPointerException in getScriptSource -> isSymLink as detailed below.

### Background

We're stuck on JRuby 1.7 for the time being on OpenVMS, as there is only Java 6 for this platform. The issue I'm reporting, however, is not VMS-only, but is also reproducible on Linux.

### Environment

The issue is reproducible in all of these JRuby versions:

- jruby 1.7.25 (ruby-1.8.7p376) 2016-04-13 867cb81 on Java HotSpot(TM) 64-Bit Server VM 1.6.0-6 +jit [OpenVMS-ia64]
- jruby 1.7.21 (ruby-1.8.7p376) 2015-07-07 a741a82 on OpenJDK 64-Bit Server VM 1.7.0_101-b00 +jit [linux-amd64]
- jruby 1.7.25 (ruby-1.8.7p376) 2016-04-13 867cb81 on OpenJDK 64-Bit Server VM 1.7.0_101-b00 +jit [linux-amd64]

The issue is not reproducible in these versions:

- jruby 1.7.20 (ruby-1.8.7p376) 2015-05-04 3086e6a on Java HotSpot(TM) 64-Bit Server VM 1.6.0-6 +jit [OpenVMS-ia64]
- jruby 1.7.20 (ruby-1.8.7p376) 2015-05-04 3086e6a on OpenJDK 64-Bit Server VM 1.7.0_101-b00 +jit [linux-amd64]

* Operating system and platform (e.g. `uname -a`)
  - HP rx2600  (1.50GHz/6.0MB) running OpenVMS V8.4 (ia64)
  - Linux dvis2 3.16.0-4-amd64 #1 SMP Debian 3.16.7-ckt20-1+deb8u3 (2016-01-17) x86_64 GNU/Linux

### Expected Behavior

In JRuby 1.7.20 on both OSes, this was the previous behaviour when specifying stdin as the script source so that arguments can also be passed to a heredoc script. I have reproduced this in a simple one-liner to illustrate the issue more concisely:

```
echo "p ARGV" | ruby /dev/stdin a b
["a", "b"]
```

On VMS, the shell syntax is different, but this is the equivalent construct and it gives the same result under 1.7.20:

```
[...]

PR #3902, opened 2016-05-18
Soften this restriction to just readable resources.

headius:
Additional fix for issue exposed by #3901. Devices, like ttys,
pipes, fifos and the like may be neither file nor symlink, but
still readable as script source. The restriction here is
unnecessary.

Ping @mkristian... I softened this test but I'm not sure why it was so restrictive in the first place. Is this change ok?


mkristian:
@headius looks ok to me. do not remember the details, but was something like a directory path as resource.

jellymann:
@headius @mkristian I have tested with a directory, it passes `canRead` and then throws at `inputStream`. See #4100

PR #3909, opened 2016-05-20
ZSuper does not require caller's binding/scope.

headius:
This appears to pass all tests, and I could find no path from ZSuper's interpret method to anything requiring DynamicScope.

### DIFF
--- a/core/src/main/java/org/jruby/ir/IRScope.java
+++ b/core/src/main/java/org/jruby/ir/IRScope.java
@@ -666,8 +666,7 @@ public abstract class IRScope implements ParseResult {
             EnumSet.of(
                     CAN_RECEIVE_BREAKS,
                     HAS_NONLOCAL_RETURNS,CAN_RECEIVE_NONLOCAL_RETURNS,
-                    BINDING_HAS_ESCAPED,
-                    USES_ZSUPER);
+                    BINDING_HAS_ESCAPED);
 
     private void computeNeedsDynamicScopeFlag() {
         for (IRFlags f : NEEDS_DYNAMIC_SCOPE_FLAGS) {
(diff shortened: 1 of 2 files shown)


subbuss:
http://dev.jruby.codehaus.narkive.com/VK4b7dEd/jruby-dev-eliminating-block-as-binding-as-a-potential-deoptimizer and your comment there is how CAN_CAPTURE_CALLERS_BINDING flag was set up in ZSuper. I had set it way back in the day and it just propagated forward through all the code refactoring. 

As for USES_ZSUPER and requiring dynscope, I think ec25a24cc8eec1dde7478f5996b9f648eee2199a is where the link probably got severed. Till that time the dynscope nesting was walked up to determine call args.

subbuss:
That said, f67e618c41c0adba7c4dcb1b1bb30bc1d795edc6 made additional changes and added https://github.com/jruby/jruby/blob/915993886ae6663586d893b7d237f750068aceb5/core/src/main/java/org/jruby/ir/instructions/ArgScopeDepthInstr.java which uses https://github.com/jruby/jruby/blob/915993886ae6663586d893b7d237f750068aceb5/core/src/main/java/org/jruby/ir/runtime/IRRuntimeHelpers.java#L1512-L1520 .... Worth checking that none of those depend on the presence of a dynamic scope.

subbuss:
Additionally, we have a more consistent semantics for define_method and zsuper whereas MRI has more restricted semantics. So, worth also checking if implementing those restricted semantics of MRI changes any of this. /cc @enebo

headius:
@subbuss I think the use of DynamicScope in the old logic was more because we had no easy way to get at the current scope's StaticScope. If we didn't actually need the DynamicScope, we'd still push a dummy scope. With these paths modified in this PR to just use the current StaticScope directly, I think we're ok.

What semantics of MRI are you referring to as "restricted semantics"? define_method obviously has more need for bindings and scopes, but I still don't think zsuper needs anything beyond StaticScope.

subbuss:
Okay, I'll trust your and @enebo 's judgment on this since it has been a while and I don't remember all the details .. but, worth checking if the flags set for ArgScopeDepthInstr are correct.

headius:
@subbuss ArgScopeDepthInstr appears to only use the StaticScope, and sets no flags. I think that's correct...for zsuper we really just need to walk up to the proper "method" scope so we can see what the arg structure was.

headius:
The one failure is known to fail sporadically, and that will hopefully be improved by #3929. Proceeding with merge.

PR #3912, opened 2016-05-22
fix four additional MRI string tests

MSNexploder:
essentially backporting MRI changes


headius:
This looks ok but I'm kicking the travis build to try again.

headius:
The failures look unrelated, and it passed `test:mri:jit`, so we'll roll it in.

headius:
Thank you!

PR #3923, opened 2016-05-25
Use "filesystem" encoding for PATH as in MRI. Fixes #3907.

### LINKED ISSUE #3907: ArgumentError: string contains null byte
### Environment

irb(main):002:0> ENV_JAVA['file.encoding']
=> Cp1250
irb(main):003:0> Encoding.find('filesystem')
=> Windows-1250
irb(main):001:0>  Encoding.default_internal
=> nil
irb(main):002:0>  Encoding.default_external
=> #<Encoding:IBM852>
irb(main):009:0> File.expand_path('~')
=> "C:/Users/Pavel Rosick\xF8"
irb(main):010:0> File.expand_path('~').encoding
=> #<Encoding:Windows-1250>
C:\jruby-9.1.1.0\bin>gem list bundler

*** LOCAL GEMS ***

bundler (1.12.4)

jruby 9.1.1.0 (2.3.0) 2016-05-19 fe84e89 Java HotSpot(TM) 64-Bit Server VM 25.92-b14 on 1.8.0_92-b14 +jit [mswin32-x86_64]

related: #3885, #3849 

### Expected Behavior

it works as expected with -J-Dfile.encoding=utf-8 
```bash
C:\jruby-9.1.1.0\bin\devel>..\jruby -J-Dfile.encoding=utf-8 -S bundle install
Warning: this Gemfile contains multiple primary sources. Using `source` more than once without a block is a security risk, and may result in installing unexpected gems. To resolve this warning, use a block to indicate which gems should come from the secondary source. To upgrade this warning to an error, run `bundle config disable_multisource true`.
Fetching gem metadata from http://rubygems.org/...........
Fetching gem metadata from https://rubygems.org/...........
Fetching version metadata from http://rubygems.org/...
Fetching version metadata from https://rubygems.org/...
Fetching dependency metadata from http://rubygems.org/..
Fetching dependency metadata from https://rubygems.org/..
Resolving dependencies.............................................................
Installing rake 11.1.2
Installing i18n 0.7.0
Using json 1.8.3
Installing minitest 5.9.0
Installing thread_safe 0.3.5
```

### Actual Behavior

ArgumentError and blank lines

```bash
C:\jruby-9.1.1.0\bin\devel>..\jruby -S bundle install
Warning: this Gemfile contains multiple primary sources. Using `source` more than once without a block is a security risk, and may result in installing unexpected gems. To resolve this warning, use a block to indicate which gems should come from the secondary source. To upgrade this warning to an error, run `bundle config disable_multisource true`.
Fetching gem metadata from http://rubygems.org/...........
Fetching gem metadata from https://rubygems.org/...........
Fetching version metadata from http://rubygems.org/...
Fetching version metadata from https://rubygems.org/...
Fetching dependency metadata from http://rubygems.org/..
Fetching dependency metadata from https://rubygems.org/..
Resolving dependencies.............................................................

ArgumentError: string contains null byte

Using json 1.8.3











Using bundler 1.12.4





















An error occurred while installing rake (11.1.2), and Bundler cannot continue.
Make sure that `gem install rake -v '11.1.2'` succeeds before bundling.
```

headius:
See #3907.

### DIFF
--- a/core/src/main/java/org/jruby/util/OSEnvironment.java
+++ b/core/src/main/java/org/jruby/util/OSEnvironment.java
@@ -111,8 +111,9 @@ public class OSEnvironment {
             val = entry.getValue();
             if ( ! (val instanceof String) ) continue; // Java devs can stuff non-string objects into env
 
+            // Ensure PATH is encoded like filesystem
             Encoding valueEncoding = keyEncoding;
-            if ( key.equals("PATH") ) {
+            if ( key.equalsIgnoreCase("PATH") ) {
                 valueEncoding = runtime.getEncodingService().getFileSystemEncoding();
             }
 
--- a/core/src/main/java/org/jruby/util/OSEnvironment.java
+++ b/core/src/main/java/org/jruby/util/OSEnvironment.java
@@ -89,15 +89,15 @@ public class OSEnvironment {
     private static Map<RubyString, RubyString> asMapOfRubyStrings(final Ruby runtime, final Map<?, ?> map) {
         @SuppressWarnings("unchecked")
         final Map<RubyString, RubyString> rubyMap = new HashMap(map.size() + 2);
-        Encoding encoding = runtime.getEncodingService().getLocaleEncoding();
+        Encoding keyEncoding = runtime.getEncodingService().getLocaleEncoding();
 
         // On Windows, map doesn't have corresponding keys for these
         if (Platform.IS_WINDOWS) {
             // these may be null when in a restricted environment (JRUBY-6514)
             String home = SafePropertyAccessor.getProperty("user.home");
             String user = SafePropertyAccessor.getProperty("user.name");
-            putRubyKeyValuePair(runtime, rubyMap, "HOME", home == null ? "/" : home, encoding);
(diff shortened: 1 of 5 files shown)


core/src/main/java/org/jruby/util/OSEnvironment.java
```diff
@@ -111,17 +111,22 @@
             val = entry.getValue();
             if ( ! (val instanceof String) ) continue; // Java devs can stuff non-string objects into env
 
-            putRubyKeyValuePair(runtime, rubyMap, key, (String) val, encoding);
+            Encoding valueEncoding = keyEncoding;
+            if ( key.equals("PATH") ) {
```
ahorek: if ( key.equalsIgnoreCase("PATH") ) {

headius: Mmm, I suppose this would be in line with environment vars being case-insensitive on Windows, wouldn't it.

headius:
The failure was valid; my changes were a bit overreaching in trying to get encodings right. I've reverted the modified behavior for "locale" encoding and everything is green again.

headius:
Merged to master in 221ba4f. Merging to 9.1.2.0 branch now.

PR #3925, opened 2016-05-25
Disable -X as there is too much verbosity in the build

jtulach:
I am removing "-X" parameter from "mvn" invocations. The debug output isn't needed anymore.


eregon:
Merged, thanks!

Issue #3728, opened 2016-03-13
BigDecimal#to_r produces grossly incorrect results


felixvf:
### Environment

```
# jruby -v
jruby 9.0.5.0 (2.2.3) 2016-01-26 7bee00d OpenJDK 64-Bit Server VM 25.72-b15 on 1.8.0_72-b15 +jit [linux-amd64]
```

### Expected Behavior

```
# irb
2.3.0 :001 > require 'bigdecimal'
 => true 
2.3.0 :002 > BigDecimal.new("5").to_f
 => 5.0 
2.3.0 :003 > BigDecimal.new("5").to_r.to_f
 => 5.0 
2.3.0 :004 > BigDecimal.new("5").to_r
 => (5/1) 
2.3.0 :005 > BigDecimal.new("1234.56789").to_r
 => (123456789/100000) 
```

### Actual Behavior

```
# irb
jruby-9.0.5.0 :001 > require 'bigdecimal'
 => true 
jruby-9.0.5.0 :002 > BigDecimal.new("5").to_f
 => 5.0 
jruby-9.0.5.0 :003 > BigDecimal.new("5").to_r.to_f
 => 1.0 
jruby-9.0.5.0 :004 > BigDecimal.new("5").to_r
 => (1/1) 
jruby-9.0.5.0 :005 > BigDecimal.new("1234.56789").to_r
 => (123456789/50) 
```

lucasallan:
I'm not able to reproduce it in `master`. cc @kares @felixvf 
```
# irb
>>  require 'bigdecimal'
=> true
>>  BigDecimal.new("5").to_f
=> 5.0
>> BigDecimal.new("5").to_r.to_f
=> 5.0
>> BigDecimal.new("5").to_r
=> (5/1)
>>  BigDecimal.new("1234.56789").to_r
=> (123456789/100000)
>> JRUBY_VERSION
=> "9.1.0.0-SNAPSHOT"
```

Lan5432:
I also tried reproducing the bug and had the same output as lucasallan, BigDecimal appears to be fine in 9.1

(I can edit with code snippet showing the version, but I'm having some troubles currently)

kares:
thanks, it seems to be a duplicate of #3616 which has been fixed with https://github.com/jruby/jruby/commit/16b13e1cca6d74c5d09bf9ad9a5d6f7686c95b0d

PR #3927, opened 2016-05-25
backport BigDecimal .to_r for #3728

ahorek:
#3728


headius:
You using 1.7 eh? Seems fine to backport this if it's needed. We'll let CI run and then merge it.

ahorek:
I think that this change is isolated and can be safely backported (if 1.7 is still maintained). But it's not a real issue for me, I'm using 9k :)

headius:
Not sure why two of the builds failed so I restarted them to see.

headius:
I'm going to try to kick that test-extended job one more time. I see a "Killed" in the log which usually means a JVM process crashed somewhere.

PR #3929, opened 2016-05-26
waitReadable should block

headius:
This attempts to fix some sporadic readpartial tests and specs in the suite, specifically in this case a failure in MRI's `test_readpartial.rb` in `test_open_pipe`. Our old old would never properly wait for the pipe to become readable and would keep spinning and trying to do nonblocking reads followed by waitReadable that never waited. As a result we could have interrupts happen at unexpected times, leaving IO in a weird state that probably caused sporadic `close` errors like in https://travis-ci.org/jruby/jruby/jobs/132959357

### DIFF
--- a/core/src/main/java/org/jruby/RubyIO.java
+++ b/core/src/main/java/org/jruby/RubyIO.java
@@ -2853,9 +2853,10 @@ public class RubyIO extends RubyObject implements IOEncodable {
                     //                n = arg.len;
                     n = OpenFile.readInternal(context, fptr, fptr.fd(), strByteList.unsafeBytes(), strByteList.begin(), len);
                     if (n < 0) {
+                        Errno e = fptr.errno();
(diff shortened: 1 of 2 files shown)


headius:
Another example, from https://travis-ci.org/jruby/jruby/jobs/131738145

```
     [exec]   1) Error:
     [exec] TestReadPartial#test_with_stdio:
     [exec] Errno::EBADF: Bad file descriptor - No message available
     [exec]     org/jruby/RubyIO.java:1955:in `close'
     [exec]     /home/travis/build/jruby/jruby/test/mri/ruby/test_readpartial.rb:14:in `make_pipe'
     [exec]     /home/travis/build/jruby/jruby/test/mri/ruby/test_readpartial.rb:26:in `pipe'
     [exec]     /home/travis/build/jruby/jruby/test/mri/ruby/test_readpartial.rb:61:in `test_with_stdio'
```

PR #3935, opened 2016-05-27
Fixed shift with negative distance on Bignum

MSNexploder:
Untagged two more tests which apparently just pass without any changes


headius:
Nice!

PR #3936, opened 2016-05-27
[Truffle] Added a new internal method for taking ownership of an array store from another array.

nirvdrum:
This is intended to be a faster version of Array#replace when we know that the source array won't be used anywhere else.

While escape analysis may very well catch some of these, I don't think it would catch all of them (e.g., the mergesort! case). And this will work in the interpreter as well.

@jruby/truffle

### DIFF
--- a/truffle/src/main/java/org/jruby/truffle/core/array/TruffleArrayNodes.java
+++ b/truffle/src/main/java/org/jruby/truffle/core/array/TruffleArrayNodes.java
@@ -22,7 +22,7 @@ import static org.jruby.truffle.core.array.ArrayHelpers.setStoreAndSize;
 @CoreClass("Truffle::Array")
 public class TruffleArrayNodes {
 
-    @CoreMethod(names = "take_ownership_of_store", onSingleton = true, required = 2)
+    @CoreMethod(names = "steal_storage", onSingleton = true, required = 2)
     public abstract static class TakeOwnershipOfStoreNode extends CoreMethodArrayArgumentsNode {
 
         @Specialization(guards = "array == other")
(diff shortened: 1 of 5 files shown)


chrisseaton:
Can you write a quick couple of specs for this in `spec/truffle/specs/truffle/array`.

pitr-ch:
👍

eregon:
Just a nitpick on the name: `store` is also the name of the field for the backing Java array. Maybe call this `storage`?
For the full name I like `steal_storage` and also `become` as suggested by Chris.
`swap` might be an interesting way to model this too, with the added semantics of exchanging the storages.

nirvdrum:
Fair enough. I named it with "store" to make it clear that you're updating that field. This is an internal operation.

truffle/src/main/java/org/jruby/truffle/core/array/TruffleArrayNodes.java
```diff
@@ -0,0 +1,45 @@
...
+
+import static org.jruby.truffle.core.array.ArrayHelpers.getSize;
+import static org.jruby.truffle.core.array.ArrayHelpers.getStore;
+import static org.jruby.truffle.core.array.ArrayHelpers.setStoreAndSize;
+
+@CoreClass("Truffle::Array")
+public class TruffleArrayNodes {
+
+    @CoreMethod(names = "steal_storage", onSingleton = true, required = 2)
+    public abstract static class TakeOwnershipOfStoreNode extends CoreMethodArrayArgumentsNode {
```
eregon: (the name of the node should also be updated)

nirvdrum: Thanks. I obviously missed that.

PR #3937, opened 2016-05-28
Implement Thread#report_on_exception

headius:
This impl is a work in progress and has not been finalized yet.

It works as follows:
- nil = report when thread is finalized if nobody has captured the
  exception (i.e. called #join or #value or abort_on_exception
  logic has fired).
- true = report when the thread terminates, regardless of capture.
- false = never report.
- Default global value is nil.
- New threads inherit the current global setting.
- If a thread name has been set from Ruby, it will be combined
  with JRuby's internal name for the report. If it has not been
  set, we will just report the internal thread name.

There are some open questions for this feature:
- If join/value are interrupted, should they still set the capture
  bit? My impl does not; they must complete.
- Is the VERBOSE-like nil/true/false clear enough or should we use
  symbols like :off, :gc, :on?

See the Ruby feature here: https://bugs.ruby-lang.org/issues/6647

### DIFF
--- a/core/src/main/java/org/jruby/RubyThread.java
+++ b/core/src/main/java/org/jruby/RubyThread.java
@@ -33,6 +33,7 @@
 package org.jruby;
 
 import java.io.IOException;
+import java.io.PrintStream;
 import java.lang.ref.WeakReference;
 import java.nio.channels.Channel;
 import java.nio.channels.SelectableChannel;
@@ -58,6 +59,7 @@ import java.util.concurrent.atomic.AtomicReference;
 import java.util.concurrent.locks.Lock;
 
 import org.jcodings.Encoding;
+import org.jcodings.specific.USASCIIEncoding;
 import org.jruby.anno.JRubyClass;
 import org.jruby.anno.JRubyMethod;
 import org.jruby.common.IRubyWarnings.ID;
@@ -121,10 +123,15 @@ public class RubyThread extends RubyObject implements ExecutionContext {
     /** Whether this thread should try to abort the program on exception */
     private volatile boolean abortOnException;
 
+    /** Whether this thread should report_on_exception when this thread GCs, when it terminates, or never */
+    private volatile IRubyObject reportOnException;
+
+    /** Whether this thread's terminating exception has been captured by any code *after* the thread terminated. */
+    private volatile boolean exceptionCaptured;
+
     /** The final value resulting from the thread's execution */
     private volatile IRubyObject finalResult;
 
-    private volatile IRubyObject threadName;
     private String file; private int line; // Thread.new location (for inspect)
 
     /**
@@ -214,6 +221,7 @@ public class RubyThread extends RubyObject implements ExecutionContext {
         super(runtime, type);
 
         finalResult = errorInfo = runtime.getNil();
+        reportOnException = runtime.getReportOnException();
     }
 
     public RubyThread(Ruby runtime, RubyClass klass, Runnable runnable) {
@@ -575,7 +583,7 @@ public class RubyThread extends RubyObject implements ExecutionContext {
             thread.setDaemon(true);
             this.file = context.getFile();
             this.line = context.getLine();
-            setThreadName(runtime, thread, file, line, true);
+            initThreadName(runtime, thread, file, line);
             threadImpl = new NativeThread(this, thread);
 
             addToCorrectThreadGroup(context);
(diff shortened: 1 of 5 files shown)


headius:
NOTE there's a big possibly-breaking change in here: in order to implement CharSequence in RubyString, I had to change the return value of the `length` method. All internal uses of that method have been fixed, but since it was public there may be Java code in the wild that will break as a result.

headius:
Output from a few interesting scenarios. I use `gets` here to pause for a while to let the thread finish without using `join`.

```
[] ~/projects/jruby $ jruby -e "Thread.report_on_exception = true; def foo; Thread.new { raise }; nil; end; foo; gets"
warning: thread "Ruby-0-Thread-1: -e:1" terminated with exception:
RuntimeError: No current exception
  block in foo at -e:1


[] ~/projects/jruby $ jruby -e "Thread.report_on_exception = true; def foo; Thread.new { raise }.name = 'hello'; nil; end; foo; gets"
warning: thread "hello (Ruby-0-Thread-1: -e:1)" terminated with exception:
RuntimeError: No current exception
  block in foo at -e:1


[] ~/projects/jruby $ jruby -e "Thread.report_on_exception = nil; def foo; Thread.new { raise }; nil; end; foo; gets; puts :ok; loop { JRuby.gc; sleep 1 }"

ok
warning: thread "Ruby-0-Thread-1: -e:1" terminated with exception:
RuntimeError: No current exception
  block in foo at -e:1
^C
[] ~/projects/jruby $ jruby -e "Thread.report_on_exception = nil; def foo; Thread.new { raise }.join rescue nil; end; foo; puts :ok; loop { JRuby.gc; sleep 1 }"
ok
^C
```

kares:
:+1: yay `RubyString implements CharSequence` now if only RubyNumeric extended java.lang.Number :)

... will miss the Ruby thread name showing up in Java's thread-name (e.g. VisualVM) does it need to go?

eregon:
Could we try with the default being `true`?

headius:
@kares It doesn't really need to be gone but I'd need to manage three separate names (ruby thread name, java thread name, thread's combined name) to be able to easily recombine them later. I'm open to suggestions on how the three should fit together.

headius:
@eregon We certainly can...do you just want to see what happens if we run like that for a full CI?

core/src/main/java/org/jruby/RubyThread.java
```diff
@@ -1115,6 +1103,10 @@ public synchronized IRubyObject inspect() {
         return getRuntime().newString(part.toString());
     }
 
+    private boolean notEmpty(CharSequence str) {
+        return str != null && str.toString().length() > 0;
```
kares: `str.length()` should do (no need to convert a `RubyString` to a Java `String` here)

core/src/main/java/org/jruby/RubyThread.java
```diff
@@ -604,44 +612,27 @@ private IRubyObject startThread(ThreadContext context, Runnable runnable) throws
...
+            name
+                    .append(':')
+                    .append(' ')
+                    .append(file)
+                    .append(':')
+                    .append(line + 1);
+        }
+        newName = name.toString();
+
+        thread.setName(newName);
```
kares: should `catch (SecurityException ignore) { } // current thread can not modify` like before (an bellow)

core/src/main/java/org/jruby/internal/runtime/NativeThread.java
```diff
@@ -109,4 +113,53 @@ public String toString() {
...
+
+        if (rubyName == null || rubyName.length() == 0) {
+            if (nativeName.equals("")) {
+                return "(unnamed)";
+            } else {
+                return nativeName;
+            }
+        } else {
+            if (nativeName.equals("")) {
+                return rubyName.toString();
```
kares: would add `'@' + rubyName` as well as bellow (so that we're consistent to have @ prefix for Ruby thread-names)

headius:
FWIW a limited version of this was added to MRI in the related feature linked above. They just went with a simple on/off setup for now, defaulting to off.

eregon:
> @eregon We certainly can...do you just want to see what happens if we run like that for a full CI?

Yeah, that would be interesting :)

headius:
This has been superseded by e408638dbb94d50915271ed6c978f76584e5bb35.

PR #3938, opened 2016-05-28
Flag registry

headius:
tldr: this is a registry for all flags we use against Ruby objects in JRuby. The goal is to maximize utilization of the flags and ensure we do not overlap.

See the first commit and the doco for FlagRegistry.

My intent with this is to begin exploring more specialized subclasses of key Ruby core classes. For example, we might want to have a RubyArrayOne that only holds one value initially in a field rather than an array. If it grows, we would set a "failed" flag that would indicate we need to fall back to using the values array rather than the one field. Rather than putting this as a boolean on the class, increasing object size, we can  use flags. But I wanted a better mechanism to ensure I'm not overlapping other flags higher in the hierarchy, as our use of them broadens.

### DIFF
--- a/core/src/main/java/org/jruby/FlagRegistry.java
+++ b/core/src/main/java/org/jruby/FlagRegistry.java
@@ -1,0 +1,75 @@
+package org.jruby;
+
+import java.util.BitSet;
+import java.util.HashMap;
+import java.util.Map;
+
+/**
+ * This class serves as a registry of all bit flags we use on JRuby objects.
+ *
+ * In order to maximally use our bit flags and prevent overlap between ancestors and dependents,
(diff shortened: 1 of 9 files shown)


headius:
FWIW, despite having flag values in RubyBasicObject up to the tenth bit, we only end up using the following values:

```
class org.jruby.RubyBasicObject={0, 1, 2, 3}
class org.jruby.ext.stringio.StringIO={4, 5}
class org.jruby.RubyHash={4, 5}
class org.jruby.RubyMatchData={4}
class org.jruby.RubyModule={4, 5, 6, 7}
class org.jruby.RubyString={4, 5}
```

PR #3942, opened 2016-05-30
Travis: Do not retry the install step.

eregon:
* In general install failures are either
  network issues which will not solve in the next 5 minutes
  or compilation issues which needs a proper fix.


mkristian:
+1 sounds good :)

Issue #3332, opened 2015-09-16
JRuby 9.0.1.0 tries to access O_NONBLOCK under Windows


sudish:
The stack trace snippet at the end is under Windows 7, using jruby-complete 9.0.1.0 with Java 1.7.0_25.

The cause is this bit of code from https://github.com/jruby/jruby/blob/master/core/src/main/java/org/jruby/RubyFile.java#L148

```Java
        if (    // O_NDELAY not defined in OpenFlags
                //OpenFlags.O_NDELAY.defined() ||
                OpenFlags.O_NONBLOCK.defined()) {
            if (!OpenFlags.O_NONBLOCK.defined()) {
//                #   define O_NONBLOCK O_NDELAY
            }
            /* do not block on open or for data to become available */
            constants.setConstant("NONBLOCK", runtime.newFixnum(OpenFlags.O_NONBLOCK.intValue()));
        } else if (Platform.IS_WINDOWS) {
            // FIXME: Should NONBLOCK exist for Windows fcntl flags?
            constants.setConstant("NONBLOCK", runtime.newFixnum(1));
        }
```

O_NONBLOCK isn't defined under Windows so the platform check should come first.

```Java
        if (Platform.IS_WINDOWS) {
            // FIXME: Should NONBLOCK exist for Windows fcntl flags?
            constants.setConstant("NONBLOCK", runtime.newFixnum(1));
        } else if (OpenFlags.O_NONBLOCK.defined()) {
            /* do not block on open or for data to become available */
            constants.setConstant("NONBLOCK", runtime.newFixnum(OpenFlags.O_NONBLOCK.intValue()));
        }
```

Thanks!

Original stack trace follows.

```
Caused by: java.lang.NoSuchMethodError: jnr.constants.platform.OpenFlags.defined()Z
        at org.jruby.RubyFile.createFileClass(RubyFile.java:148)
        at org.jruby.Ruby.initCore(Ruby.java:1551)
        at org.jruby.Ruby.bootstrap(Ruby.java:1336)
        at org.jruby.Ruby.init(Ruby.java:1227)
        at org.jruby.Ruby.newInstance(Ruby.java:330)
        at org.jruby.embed.internal.AbstractLocalContextProvider.getGlobalRuntime(AbstractLocalContexProvider.java:82)
        at org.jruby.embed.internal.SingletonLocalContextProvider.getRuntime(SingletonLocalContextProider.java:99)
[...]

PR #3958, opened 2016-06-09
Windows check should come first as the O_NONBLOCK does work under win…

SquirrelGrip:
Hi Reviewer,

Sudish reported a bug in JRuby running under Windows 7, Issue: #3332

I am currently running cucumber-jruby on various platforms and only under windows do I get this problem.

The fix is trivial, just check that the system is windows first before making any further checks.

Thanks Sudish for making it easy to fix.

I removed commented out code also, as the logic was if something then if not something do nothing. Kind of stoopid.

### DIFF
--- a/core/src/main/java/org/jruby/RubyFile.java
+++ b/core/src/main/java/org/jruby/RubyFile.java
@@ -144,17 +144,12 @@ public class RubyFile extends RubyIO implements EncodingCapable {
         constants.setConstant("CREAT", runtime.newFixnum(OpenFlags.O_CREAT.intValue()));
         /* error if CREAT and the file exists */
         constants.setConstant("EXCL", runtime.newFixnum(OpenFlags.O_EXCL.intValue()));
-        if (    // O_NDELAY not defined in OpenFlags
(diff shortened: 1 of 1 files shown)


headius:
The commented code was there as a "todo" for getting Windows support improved in recently-ported MRI logic.

I mentioned on #3332 that the error produced should not happen on a properly built JRuby. There must be a stale version of jnr-constants causing that error, because JRuby wouldn't even compile if that method does not exist.

Can you explain the environment where you see this error? Something's wrong there.

headius:
Also, sorry for the delay getting back to you, @SquirrelGrip :-)

SquirrelGrip:
Hi Headius, Sorry for also taking a while to get back. I had a look into what you said and revisited the problem. I am now unable to reproduce when I started again.
I think I may have had an environment issue and suspect that I had an older version of JRuby on the path and with my pom I had the latest version.
Thanks for looking into it.
I will revert if I discover any more information for others.

headius:
I will close this and #3332 until you or @sudish can reproduce. Thanks for getting back to us!

Issue #3919, opened 2016-05-24
Arrays.sort


sri:
Is there a reason that JRuby's RubyArray uses it own hand-rolled sort (in Qsort.java) instead of using Java's builtin `Arrays.sort`?

I'm looking at these files:
- https://github.com/jruby/jruby/blob/master/core/src/main/java/org/jruby/RubyArray.java#L3278
- https://github.com/jruby/jruby/blob/master/core/src/main/java/org/jruby/RubyArray.java#L3313
- https://github.com/jruby/jruby/blob/master/core/src/main/java/org/jruby/util/Qsort.java

The reason I'm asking: I encountered a bug in a piece of code. A class was overriding the <=> method, but due to a typo, it was doing this comparision:

`def <=>(other); self.a <=> other.b; end`

instead of

`def <=>(other); self.a <=> other.a; end`.

Obviously, that is a bug in the application. I didn't notice the bug because `a == b` most of the time. However, recently, there were cases where `a != b` and that's when the bug manifisted. And JRuby's sort got stuck in an infinite loop.

When I converted JRuby to use `Arrays.sort` instead of `Qsort` and tried it with the bad data, it threw an exception: "Comparison method violates its general contract!".

Also, there are other benefits of using `Arrays.sort`:
https://github.com/openjdk-mirror/jdk7u-jdk/blob/master/src/share/classes/java/util/TimSort.java#L29

Is there interest in converting RubyArray to use `Arrays.sort`?

Thanks!
-Sriram

sri:
Ah, I see according to this `4c38d8608ef8a3130ffa6b8ff60ae06c1770a261` -- `Improved sort for JRUBY-2198: Array#sort is slower than MRI`, `Arrays.sort` was pulled out because it was slow. But this is from 2008. Looks like `Arrays.sort` was rewritten in 2009:

https://github.com/openjdk-mirror/jdk7u-jdk/commit/cdf72de6556306d665b02917c03dfd6e957b910f

Is there an easy way to benchmark the current Qsort.java with the current `Arrays.sort`?

headius:
@sri Yes, we introduced our own sort both for performance and to match MRI behaviorally.

It would definitely be worth examining whether the rewritten JDK-provided Arrays.sort is now as fast or faster, since we always like deleting code. Are you interested in working on a test patch and benchmarking it?

sri:
[...]

PR #3961, opened 2016-06-09
Use Java's Arrays.sort

sri:
Java's Arrays.sort more robust and is faster is certain cases.

See https://github.com/jruby/jruby/issues/3919 and
https://github.com/sri/jruby-sort-benchmark.


enebo:
@sri could you re-add the util class and @Deprecate it at the top.  I do not know if extensions use that class or not but it is public.

sri:
@enebo Done!

enebo:
@sri thanks for the PR and the time taken to show Java's sort has improved!

sri:
@enebo Great! Thanks for all the pointers!!

PR #3977, opened 2016-06-24
[Truffle] Make global variables constants if they are rarely changed

eregon:
@jruby/truffle @thedarkone So here is my attempt to make global variables constants if they are not changed too often. The logic is unfortunately more complex than I expected, so I would appreciate a review :smiley:

### DIFF
--- a/core/src/main/java/org/jruby/util/cli/Options.java
+++ b/core/src/main/java/org/jruby/util/cli/Options.java
@@ -248,6 +248,9 @@ public class Options {
     public static final Option<Boolean> TRUFFLE_ROPE_LAZY_SUBSTRINGS = bool(TRUFFLE, "truffle.rope.lazy_substrings", true, "Indicates whether a substring operation on a rope should be performed lazily.");
     public static final Option<Boolean> TRUFFLE_ROPE_PRINT_INTERN_STATS = bool(TRUFFLE, "truffle.rope.print_intern_stats", false, "Print interned rope stats at application exit.");
 
+    public static final Option<Integer> TRUFFLE_GLOBAL_VARIABLE_MAX_INVALIDATIONS = integer(TRUFFLE, "truffle.global_variable.max_invalidations", 10,
+            "Maximum number of times a global variable can be changed to be considered constant.");
+
     public static final Option<Integer> TRUFFLE_DEFAULT_CACHE = integer(TRUFFLE, "truffle.default_cache", 8, "Default size for caches.");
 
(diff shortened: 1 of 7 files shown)


core/src/main/java/org/jruby/util/cli/Options.java
```diff
@@ -258,6 +258,7 @@
     public static final Option<Integer> TRUFFLE_BIND_CACHE = integer(TRUFFLE, "truffle.bind.cache", TRUFFLE_DEFAULT_CACHE.load(), "Cache size of test for being able to bind a method to a module.");
     public static final Option<Integer> TRUFFLE_CONSTANT_CACHE = integer(TRUFFLE, "truffle.constant.cache", TRUFFLE_DEFAULT_CACHE.load(), "Constant cache size.");
     public static final Option<Integer> TRUFFLE_INSTANCE_VARIABLE_CACHE = integer(TRUFFLE, "truffle.instance_variable.cache", TRUFFLE_DEFAULT_CACHE.load(), "Instance variable cache size.");
+    public static final Option<Integer> TRUFFLE_GLOBAL_VARIABLE_CACHE = integer(TRUFFLE, "truffle.global_variable.cache", 10, "Maximum number of times a global variable can be changed to be considered constant.");
```
eregon: JRuby classic uses 100 here (see above), but that seems potentially too many deopts.

truffle/src/main/java/org/jruby/truffle/language/globals/GlobalVariableStorage.java
```diff
@@ -10,10 +10,62 @@
...
+import org.jruby.util.cli.Options;
+
 public class GlobalVariableStorage {
-    public volatile Object value;
+
+    private static final int GLOBAL_VARIABLE_CACHE = Options.TRUFFLE_GLOBAL_VARIABLE_CACHE.load();
+
+    private final CyclicAssumption unchangedAssumption = new CyclicAssumption("global variable unchanged");
+    private int changes = 0;
+    @CompilationFinal private volatile boolean assumeConstant = true;
```
eregon: This really means `@CompilationFinal` for compilation and `volatile` in interpreter.

core/src/main/java/org/jruby/util/cli/Options.java
chrisseaton: This should perhaps be called `_INVALIDATIONS` and `.invalidations` as it isn't the size of a cache, like the other `_CACHE` variable are.

truffle/src/main/java/org/jruby/truffle/language/globals/GlobalVariableStorage.java
chrisseaton: Maybe put that in a comment.

thedarkone:
lgtm.

PR #3979, opened 2016-06-26
[Truffle] Use to_enum snippets to provide enumerator size blocks

bjfish:
@chrisseaton  I've tried a few ways to do this. How does this look?


nirvdrum:
I don't want to get nit-picky about style, but can you please convert your tabs to spaces? Almost the entire JRuby codebase uses spaces.

bjfish:
@nirvdrum Fixed looks like my IDE was misconfigured, thanks for reminding me.

truffle/src/main/java/org/jruby/truffle/builtins/SnippetIfNoBlockNode.java
```diff
@@ -0,0 +1,51 @@
+/*
+ * Copyright (c) 2013, 2016 Oracle and/or its affiliates. All rights reserved. This
```
chrisseaton: I think this is all new code, so start at 2016.

truffle/src/main/java/org/jruby/truffle/builtins/SnippetIfNoBlockNode.java
```diff
@@ -0,0 +1,51 @@
...
+
+import com.oracle.truffle.api.CompilerDirectives;
+import com.oracle.truffle.api.frame.VirtualFrame;
+import com.oracle.truffle.api.object.DynamicObject;
+import com.oracle.truffle.api.profiles.ConditionProfile;
+
+public class SnippetIfNoBlockNode extends RubyNode {
+
+    @Child
+    private RubyNode method;
```
chrisseaton: Keep `@Child` and `private` on the same line.

truffle/src/main/java/org/jruby/truffle/core/rubinius/RubiniusSnippets.java
```diff
@@ -0,0 +1,56 @@
+/*
+ * Copyright (c) 2015, 2016 Oracle and/or its affiliates. All rights reserved. This
```
chrisseaton: Start at 2016.

truffle/src/main/java/org/jruby/truffle/core/rubinius/RubiniusSnippets.java
```diff
@@ -0,0 +1,56 @@
...
+ * FOR ANY DIRECT, INDIRECT, INCIDENTAL, SPECIAL, EXEMPLARY, OR CONSEQUENTIAL
+ * DAMAGES (INCLUDING, BUT NOT LIMITED TO, PROCUREMENT OF SUBSTITUTE GOODS OR
+ * SERVICES; LOSS OF USE, DATA, OR PROFITS; OR BUSINESS INTERRUPTION) HOWEVER
+ * CAUSED AND ON ANY THEORY OF LIABILITY, WHETHER IN CONTRACT, STRICT LIABILITY,
+ * OR TORT (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY OUT OF THE USE
+ * OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
+ */
+package org.jruby.truffle.core.rubinius;
+
+public abstract class RubiniusSnippets {
```
chrisseaton: We're extensively modifying the Rubinius code in place now, so I don't think we should use the name Rubinius in our implementation any more. For example I'm now calling 'Rubinius primitives' just 'primitives'. So this could be `ToEnumSnippets` perhaps.

chrisseaton:
Seems a little complex, but I guess you are trying to solve a complex problem here. The declarative aspect is good.

truffle/src/main/java/org/jruby/truffle/core/rubinius/ToEnumSnippets.java
```diff
@@ -0,0 +1,56 @@
...
+    public static final String ARRAY_EACH_WITH_INDEX_TO_ENUM = "to_enum(:each_with_index) { size }";
+    public static final String ARRAY_MAP_TO_ENUM = "to_enum(:map) { size }";
+    public static final String ARRAY_MAP_BANG_TO_ENUM = "to_enum(:map!) { size }";
+    public static final String ARRAY_REJECT_TO_ENUM = "to_enum(:reject) { size }";
+    public static final String ARRAY_REJECT_BANG_TO_ENUM = "to_enum(:reject!) { size }";
+    public static final String ARRAY_SELECT_TO_ENUM = "to_enum(:select) { size }";
+
+    public static final String HASH_EACH_TO_ENUM = "to_enum(:each) { size }";
+
+    public static final String RANGE_EACH_TO_ENUM = "to_enum { size }";
```
eregon: For consistency, maybe always provide the name parameter?

eregon:
This looks good!
Maybe we can enjoy further the fact that all snippets are very similar, by only giving it the method to compute the "enumerator size". We can then get the method name from the frame.
So, something like:
`@CoreMethod(names = "each_with_index", needsBlock = true, enumeratorSize = "size")`

truffle/src/main/java/org/jruby/truffle/builtins/CoreMethodNodeManager.java
```diff
@@ -263,7 +263,9 @@ private static CallTarget makeGenericMethod(RubyContext context, MethodDetails m
         } else {
             sequence = Translator.sequence(context, sourceSection, Arrays.asList(checkArity, methodNode));
 
-            if (method.returnsEnumeratorIfNoBlock()) {
+            if(!method.snippetIfNoBlock().isEmpty()) {
+                sequence = new SnippetIfNoBlockNode(method.snippetIfNoBlock(), sequence);
+            } else if (method.returnsEnumeratorIfNoBlock()) {
```
eregon: Can you add an `assert` so that we make sure both options are no given?

truffle/src/main/java/org/jruby/truffle/builtins/CoreMethodNodeManager.java
```diff
@@ -263,7 +263,9 @@ private static CallTarget makeGenericMethod(RubyContext context, MethodDetails m
         } else {
             sequence = Translator.sequence(context, sourceSection, Arrays.asList(checkArity, methodNode));
 
-            if (method.returnsEnumeratorIfNoBlock()) {
+            if(!method.snippetIfNoBlock().isEmpty()) {
```
eregon: Syntax nitpick: `if (!method...`

bjfish:
@eregon I like this idea `enumeratorSize = "size"` do you know how I can create Block from a string in Java. I think I've been able to create a proc like this before but not a block yet.

bjfish:
@eregon Please disregard the previous question about creating a block from a string. I think I've figured this out.

bjfish:
I've added a commit that refactors to @eregon 's suggestions

truffle/src/main/java/org/jruby/truffle/builtins/EnumeratorSizeNode.java
```diff
@@ -0,0 +1,49 @@
...
+
+    @Child private RubyNode method;
+    @Child private SnippetNode snippetNode;
+    private final ConditionProfile noBlockProfile = ConditionProfile.createBinaryProfile();
+    private final String snippet;
+
+    public EnumeratorSizeNode(String enumeratorSize, String methodName, RubyNode method) {
+        super(method.getContext(), method.getEncapsulatingSourceSection());
+        this.method = method;
+        this.snippet = "to_enum(:" + methodName + ") { " + enumeratorSize + " }";
```
eregon: This is the easiest way indeed :)
I thought about a CallDispatchNode since this is just a call, but building the block is just more trouble.

eregon:
Let's get this merged!

Issue #3941, opened 2016-05-30
JRuby 9.0.5.0+ OOM errors under CentOS


etehtsea:
This issue also reproduced on production app after ~2 days of running under new jruby version. At some point we started to get OOM errors in our error-catcher and rpm on this node decreased relatively to other nodes which ran under 9.0.4.0. After switching to ubuntu based container this issue disappeared.

Results matrix:
JRuby 9.0.4.0 (Centos): :+1: 
JRuby 9.0.5.0+ (Centos): :-1: 
JRuby 9.0.5.0+ (Ubuntu): :+1: 

Images used below available on docker hub:
- https://hub.docker.com/r/etehtsea/jruby-ubuntu/
- https://hub.docker.com/r/etehtsea/jruby-centos/

P.S. Notice that `bundler` `1.12.0+` has another perf issue (https://github.com/bundler/bundler/issues/4478) so I pinned `bundler` version to `1.11.2`.

**Gemfile** used:

```ruby
source 'https://rubygems.org'; gem 'rack'
```

**JRuby 9.0.4.0 (CentOS)**
```
/tmp  ⚡ \time -p docker run --rm -it -v /tmp/Gemfile:/Gemfile -w / etehtsea/jruby-centos:9.0.4.0 jruby -v -S bundle install --verbose --full-index 
jruby 9.0.4.0 (2.2.2) 2015-11-12 b9fb7aa OpenJDK 64-Bit Server VM 25.91-b14 on 1.8.0_91-b14 +jit [linux-amd64]
Don't run Bundler as root. Bundler can ask for sudo if it is needed, and
installing your bundle as root will break this application for all non-root
users on this machine.
Fetching source index from https://rubygems.org/
Need to query more than 500 gems. Downloading full index instead...
HTTP GET https://rubygems.org/quick/Marshal.4.8/rack-0.1.0.gemspec.rz
<..>
HTTP GET https://rubygems.global.ssl.fastly.net/quick/Marshal.4.8/rack-2.0.0.rc1.gemspec.rz
HTTP 200 OK
Resolving dependencies....
Installing rack 1.6.4
0:  rack (1.6.4) from /opt/jruby/lib/ruby/gems/shared/specifications/rack-1.6.4.gemspec
Using bundler 1.11.2
0:  bundler (1.11.2) from /opt/jruby/lib/ruby/gems/shared/specifications/bundler-1.11.2.gemspec
Bundle complete! 1 Gemfile dependency, 2 gems now installed.
[...]

PR #3988, opened 2016-06-30
Packed arrays

headius:
This branch is an implementation of packed arrays for one and two-element Array instances.

Basics
=====

There are three new classes: RubyArraySpecialized, a superclass for all the specialized instances, and RubyArrayOneObject and RubyArrayTwoObject, one and two-elt versions of RubyArray. Generally these are used when constructing a new array known to be one or two elements. If the size of the array needs to change from the packed size, it "unpacks" into an array and proceeds to use normal array-based RubyArray logic.

Background
=========

This work was prompted in part by explorations into the heap of a very large application in #3941 as detailed here: https://github.com/jruby/jruby/issues/3941#issuecomment-224678533

In addition to the many small arrays used by RubyGems, there are many pieces of code in stdlib and other libraries that create small arrays as transient data carriers. We also use small arrays in many places in the JRuby runtime for passing block arguments, splatted arguments, and so on. These cases all can benefit from the packing.

Benefits
======

The two primary benefits of this work are reduced memory usage for one and two-element arrays, and reduced dereferencing requirements for those arrays (since we don't have to hop through an IRubyObject[] to get them).

Memory-wise, the packed arrays add one and two references to the size of RubyArray, but eliminate the much-larger IRubyObject[] header required for arbitrarily-sized arrays. This translates to 60% savings in the one object case and around 40% savings in the two-object case.

Analysis of a Rails app
=================

An analysis of a blank Rails app in development mode shows me the following, after this patch:

There are 909 live RubyArrayOneObject instances and 565 RubyArrayTwoObject instances, consuming about 102KB of memory. If these were all regular RubyArray with a perfectly-sized IRubyObject[], that would be about doubled. If those RubyArray instances had the usual "base" size of 16 elements, it would be considerably higher.

Of the 3046 remaining RubyArray instances, 1626 have one element and 1045 have two elements. These are opportunities to improve our use of packing, but they will require deeper analysis. There are many ways to produce an Array, and the packing logic I've written here can't handle resizes...so many of these arrays may have started out as a different size. In the future we may look at speculatively allocating arrays we know will *eventually* be packable and adding a way to go from unpacked form to packed form (only the reverse exists now).

Of the 909 RubyArrayOneObject instances, only 2 ended up having to unpack. Of the 565 RubyArrayTwoObject instances, 37 unpacked. This tells me we're doing a good job deciding when to use a packed version that won't just end up unpacking.

There are only 190 three-element RubyArray and 229 four-element RubyArray in the remaining pool. The benefit of packing drops of quickly after one and two-element arrays.

A large app with more "data" may show greater benefit from this patch, especially if it uses many small arrays to represent any data (e.g. a large binary tree).

Doing an allocation profile of that same Rails app, booting up the server and hitting the info page and an error page, I see the following results:

* 41781 RubyArray instances are allocated in total.
* 8149 RubyArrayOneObject instances are allocated.
* 20211 RubyArrayTwoObject instances are allocated.

Risks
=====

There are a few risks with this patch:

* There's additional complexity anywhere we construct a new Array to ensure we're giving it a chance to be packed. I detailed the newBlankArray+store pattern in one of the commits to expand use of packed arrays.
* Methods overridden by the packed versions can potentially be polymorphic now when they would not have been before. However I suspect most Array call sites will be dealing with the same-sized array each time.
* New code is risky code. If a packed array is not initialized correctly, or if it does not unpack when it should, it may cause surprising failures.

Future work
=========

* Explore whether three and four-element packed arrays are worth the additional complexity.
* Continue improving and expanding our use of packed arrays wherever possible.
* Begin exploring packed array forms that can handle all-primitive values, to cut down on object boxing for live primitive values on the heap. This will need some analysis to see how frequently large all-Fixnum or all-Float arrays are used in typical libraries and apps.


core/src/main/java/org/jruby/RubyArray.java
```diff
@@ -189,11 +201,15 @@ public static final RubyArray newArrayLight(final Ruby runtime) {
     }
 
     public static RubyArray newArray(Ruby runtime, IRubyObject obj) {
-        return new RubyArray(runtime, new IRubyObject[] { obj });
+        return new RubyArrayOneObject(runtime, obj);
     }
 
     public static RubyArray newArrayLight(Ruby runtime, IRubyObject obj) {
-        return new RubyArray(runtime, new IRubyObject[] { obj }, false);
+        return new RubyArrayOneObject(runtime, obj);
```
kares: so it might show up in ObjectSpace now ... thus not really "light" (not sure it matters at all)

core/src/main/java/org/jruby/RubyHash.java
```diff
@@ -641,18 +641,26 @@ public boolean matches(final RubyHashEntry entry, final Object obj) {
...
+    }
+
+    public static abstract class Visitor extends VisitorWithState {
+        public void visit(ThreadContext context, RubyHash self, IRubyObject key, IRubyObject value, int index, Object state) {
+            visit(key, value);
+        }
         public abstract void visit(IRubyObject key, IRubyObject value);
     }
 
-    public void visitAll(Visitor visitor) {
```
kares: could we have the previous `visitAll` method (as well) for compatibility e.g. jruby-rack is using it.

kares:
did found the `visitAll` method at the end, so ignore that. 
all :+1: to me esp. like the refactoring of RubyArray's `append` on loops.

PR #3994, opened 2016-07-03
[Truffle] Implement Array#hash using murmurhash

bjfish:
Does this need `@TruffleBoundary`?

### DIFF
--- a/truffle/src/main/java/org/jruby/truffle/core/array/ArrayNodes.java
+++ b/truffle/src/main/java/org/jruby/truffle/core/array/ArrayNodes.java
@@ -920,7 +920,6 @@ public abstract class ArrayNodes {
     public abstract static class HashNode extends ArrayCoreMethodNode {
 
         @Child private ToIntNode toIntNode;
-        @Child private ArrayReadSliceNormalizedNode readSliceNode;
 
         @Specialization(guards = "isNullArray(array)")
         public long hashNull(VirtualFrame frame, DynamicObject array) {
@@ -958,7 +957,7 @@ public abstract class ArrayNodes {
                 CompilerDirectives.transferToInterpreterAndInvalidate();
                 toIntNode = insert(ToIntNode.create());
             }
-            Object result = toIntNode.executeIntOrLong(frame, indexObject);
+            final Object result = toIntNode.executeIntOrLong(frame, indexObject);
             if (result instanceof Integer) {
                 return Long.valueOf((int) result);
             } else {
--- a/truffle/src/main/java/org/jruby/truffle/core/array/ArrayNodes.java
+++ b/truffle/src/main/java/org/jruby/truffle/core/array/ArrayNodes.java
@@ -31,11 +31,7 @@ import com.oracle.truffle.api.profiles.BranchProfile;
 import com.oracle.truffle.api.profiles.ConditionProfile;
 import com.oracle.truffle.api.source.SourceSection;
 import org.jcodings.specific.UTF8Encoding;
-import org.jruby.Ruby;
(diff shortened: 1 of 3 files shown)


truffle/src/main/java/org/jruby/truffle/core/array/ArrayNodes.java
```diff
@@ -911,6 +916,56 @@ protected Object fillFallback(VirtualFrame frame, DynamicObject array, Object[]
 
     }
 
+    @CoreMethod(names = "hash_internal")
+    public abstract static class HashNode extends ArrayCoreMethodNode {
+
+        @Child private ToIntNode toIntNode;
+
+        @Specialization(guards = "isNullArray(array)")
+        public long hashNull(VirtualFrame frame, DynamicObject array) {
```
eregon: When the frame is not needed, it's usually best to avoid it in the arguments for clarity.

truffle/src/main/java/org/jruby/truffle/core/array/ArrayNodes.java
```diff
@@ -911,6 +916,56 @@ protected Object fillFallback(VirtualFrame frame, DynamicObject array, Object[]
...
+            long h = Helpers.hashStart(getContext().getJRubyRuntime(), arraySize);
+            h = Helpers.murmurCombine(h, System.identityHashCode(ArrayNodes.class));
+            h = hashEnd(h);
+            return h;
+        }
+
+        @Specialization(guards = "strategy.matches(array)", limit = "ARRAY_STRATEGIES")
+        public long hash(VirtualFrame frame, DynamicObject array,
+                         @Cached("of(array)") ArrayStrategy strategy,
+                         @Cached("new()") SnippetNode snippetNode) {
```
eregon: This could be a simple `CallDispatchHeadNode` (which can be created with `createMethodCall`).

truffle/src/main/java/org/jruby/truffle/core/array/ArrayNodes.java
```diff
@@ -88,8 +89,12 @@
 import org.jruby.truffle.language.objects.TaintNodeGen;
 import org.jruby.truffle.language.yield.YieldNode;
 import org.jruby.util.Memo;
+
 import java.util.Arrays;
 import java.util.Comparator;
+
+import static org.jruby.runtime.Helpers.hashEnd;
+import static org.jruby.runtime.Helpers.murmurCombine;
```
eregon: Try to avoid `static` imports, unless there are many usages.

eregon:
> Does this need @TruffleBoundary?

Looks like it's fine as it is, there is almost no branching and it's just arithmetic.
`hashStart` takes the JRuby runtime though, it would be best if we can just just choose the right seed in advance, and not branch on `siphashEnabled` at every call.

truffle/src/main/java/org/jruby/truffle/core/array/ArrayNodes.java
```diff
@@ -920,13 +918,14 @@ protected Object fillFallback(VirtualFrame frame, DynamicObject array, Object[]
     public abstract static class HashNode extends ArrayCoreMethodNode {
 
         @Child private ToIntNode toIntNode;
+        @Child private CallDispatchHeadNode toHashNode;
```
eregon: This can also be in `@Cached` and actually replace the SnippetNode.

bjfish: @eregon I've updated this to be `@Cached` now and removed the unused SnippetNode

eregon:
Let's merge this!

PR #3997, opened 2016-07-06
[Truffle] Correcting Array#inspect spec with inspect fixes

bjfish:
This Array#inspect spec was failing due to issues with the current `String#inspect` and lack of `rb_inspect` and `rb_str_escape`. I had to get creative/experimental to resolve all these issues.

Any issues with the following? 
Are there any issue with using Truffle::CExt nodes like this in our ruby core code? 
Are there any issues using the `RubyString` implementations?


truffle/src/main/ruby/core/type.rb
```diff
@@ -419,6 +419,22 @@ def self.compatible_encoding(a, b)
       enc
     end
 
+    # similar to rb_inspect
+    def self.object_inspect(val)
```
eregon: `Rubinius::Type.object_*` are usually for the original version of these methods on Object/Kernel, like `class`, etc.
Maybe just `inspect` here is fine? (so we just remove the `rb_` prefix)

truffle/src/main/ruby/core/type.rb
```diff
@@ -419,6 +419,22 @@ def self.compatible_encoding(a, b)
...
+      str = Rubinius::Type.coerce_to(val.inspect, String, :to_s)
+      result_encoding = Encoding.default_internal || Encoding.default_external
+      unless result_encoding.ascii_compatible?
+        unless str.ascii_only?
+          return Truffle::CExt.to_ruby_string(Truffle::CExt.rb_str_escape(Truffle::CExt.CExtString(str)))
+        end
+        return str
+      end
+      if (str.encoding != result_encoding && !str.ascii_only?)
+        return Truffle::CExt.to_ruby_string(Truffle::CExt.rb_str_escape(Truffle::CExt.CExtString(str)))
```
eregon: Duplicated logic here, can we make it into one nice condition?

eregon:
> Are there any issue with using Truffle::CExt nodes like this in our ruby core code? 

Technically, it's OK since it's just a Java `@CoreMethod`.
But it is the wrong location.
What about making it a primitive, say `string_escape` in StringNodes?

It would also avoid the wrapping/unwrapping which seems wrong (`rb_str_escape` returns a VALUE not a char* which is what CExtString represents and no method should deal with that except the existing conversion methods).
Also CExtNodes is for helpers of the C-ext API for functionality not existing otherwise. `rb_str_escape` is not part of the C-ext API.

> Are there any issues using the RubyString implementations?

It's not ideal because it means it's less easy to optimize and might throw jruby exceptions.
In this case it seems OK though.
In the long term, maybe this could make it into StringSupport so there is no need to build the RubyString.
(And in general a more functional-style method which does not rely on global state is always better as keeping both global states in sync is very error-prone).

bjfish:
@eregon I've refactored per your suggestions, thanks!

truffle/src/main/java/org/jruby/truffle/core/string/StringNodes.java
```diff
@@ -3103,6 +3123,19 @@ protected static boolean hasRawBytes(DynamicObject string) {
 
     }
 
+    @CoreMethod(names = "string_escape")
```
eregon: Make this a primitive, like the node just below.
You can then invoke it with `Truffle.invoke_primitive :string_escape, str`.
This avoids polluting String's namespace.

bjfish: @eregon Made it a primitive in the latest commit

eregon: Why not using `invoke_primitive`? I'll push a commit to do that.

eregon:
cc @jruby/truffle 
If `return` is in almost last position (`return :foo if bla; :bar`), I usually prefer a if/else as `return` needs to throw an exception in interpreter and it also makes the flow look more functional. If it's an early exit, `return` is the good choice.
Also performance-wise, `invoke_primitive` is a nice way to insert a node directly into the tree, without call overhead and ensuring it cannot be monkey-patched. But it's quite verbose and less idiomatic so I would only use it to call semi-hidden stuff like that string_escape.

PR #4005, opened 2016-07-09
[Truffle] Add SystemCallError Layout for errno field

bjfish:
Please review the new layout usage here.

### DIFF
--- a/truffle/src/main/java/org/jruby/truffle/core/exception/ExceptionNodes.java
+++ b/truffle/src/main/java/org/jruby/truffle/core/exception/ExceptionNodes.java
@@ -148,6 +148,7 @@ public abstract class ExceptionNodes {
     @Primitive(name = "exception_errno_error", needsSelf = false)
     public static abstract class ExceptionErrnoErrorPrimitiveNode extends PrimitiveArrayArgumentsNode {
 
+        @TruffleBoundary
         @Specialization
         public DynamicObject exceptionErrnoError(DynamicObject message, int errno, DynamicObject location) {
             final String errorMessage;
--- a/truffle/src/main/java/org/jruby/truffle/core/exception/ExceptionNodes.java
+++ b/truffle/src/main/java/org/jruby/truffle/core/exception/ExceptionNodes.java
@@ -27,6 +27,7 @@ import org.jruby.truffle.builtins.Primitive;
 import org.jruby.truffle.builtins.PrimitiveArrayArgumentsNode;
 import org.jruby.truffle.core.string.StringOperations;
 import org.jruby.truffle.language.NotProvided;
+import org.jruby.truffle.language.RubyGuards;
 import org.jruby.truffle.language.backtrace.Backtrace;
 import org.jruby.truffle.language.objects.AllocateObjectNode;
 import org.jruby.truffle.language.objects.AllocateObjectNodeGen;
@@ -139,7 +140,7 @@ public abstract class ExceptionNodes {
 
         @Specialization
         public DynamicObject allocate(DynamicObject rubyClass) {
-            return allocateObjectNode.allocate(rubyClass, null, null);
+            return allocateObjectNode.allocate(rubyClass, null, null, nil());
         }
 
     }
@@ -148,8 +149,14 @@ public abstract class ExceptionNodes {
     public static abstract class ExceptionErrnoErrorPrimitiveNode extends PrimitiveArrayArgumentsNode {
 
         @Specialization
-        public DynamicObject exceptionErrnoError(DynamicObject message, int errno) {
-            return coreExceptions().errnoError(errno, message.toString(), this);
+        public DynamicObject exceptionErrnoError(DynamicObject message, int errno, DynamicObject location) {
+            final String errorMessage;
+            if(RubyGuards.isRubyString(location)){
+                errorMessage = " @ " + location.toString() + " - " + message.toString();
+            } else {
+                errorMessage = " - " + message.toString();
+            }
+            return coreExceptions().errnoError(errno, errorMessage, this);
         }
 
     }
--- a/truffle/src/main/java/org/jruby/truffle/core/exception/SystemCallErrorLayout.java
+++ b/truffle/src/main/java/org/jruby/truffle/core/exception/SystemCallErrorLayout.java
@@ -26,7 +26,7 @@ public interface SystemCallErrorLayout extends ExceptionLayout {
         DynamicObjectFactory factory,
         @Nullable Object message,
         @Nullable Backtrace backtrace,
-        @Nullable Object errno);
+                  Object errno);
 
     Object getErrno(DynamicObject object);
     void setErrno(DynamicObject object, Object value);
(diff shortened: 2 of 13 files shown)


bjfish:
It looks like MRI uses hidden instance vars for these kinds of internal fields. Is using layouts our best equivalent for this?

cc: @nirvdrum

bjfish:
@jruby/truffle Would it be nice if we supported the same thing as MRI for hidden instance vars? I think (please confirm) in MRI if you omit the `@` symbol `rb_ivar_set("errno", 1)` the instance var becomes hidden from things like `instance_variables` and `instance_variable_get/set`. Or is there a better approach and I should use layouts like this pull request?

I noticed our NoMethodError currently shows `@name` instance var in `instance_variables` when it should be hidden. It is defined as `noMethodError.define("@name", context.getSymbolTable().getSymbol(name), 0);`

chrisseaton:
You should use a layout if you need to be able to access the fields from Java without too much hassle.

You should use a `HiddenKey` field name, rather than a `j.l.String`, to produce the kind of hidden instance variables you're after.

truffle/src/main/java/org/jruby/truffle/language/parser/jruby/BodyTranslator.java
```diff
@@ -1827,6 +1828,12 @@ public RubyNode visitInstAsgnNode(org.jruby.ast.InstAsgnNode node) {
                 ret = new WriteInstanceVariableNode(context, sourceSection, name, self, IntegerCastNodeGen.create(context, sourceSection, rhs));
                 return addNewlineIfNeeded(node, ret);
             }
+        } else if (path.equals(corePath + "exception.rb")) {
+            if (name.equals("@errno")) {
```
chrisseaton: Could we just modify `exception.rb` in place to use a primitive to read the error, rather than having to rewrite here?

bjfish: @chrisseaton I've udpated this to use a primitive now

chrisseaton: This all looks good then.

truffle/src/main/java/org/jruby/truffle/core/exception/ExceptionNodes.java
```diff
@@ -148,8 +149,14 @@ public DynamicObject allocate(DynamicObject rubyClass) {
     public static abstract class ExceptionErrnoErrorPrimitiveNode extends PrimitiveArrayArgumentsNode {
 
         @Specialization
-        public DynamicObject exceptionErrnoError(DynamicObject message, int errno) {
-            return coreExceptions().errnoError(errno, message.toString(), this);
+        public DynamicObject exceptionErrnoError(DynamicObject message, int errno, DynamicObject location) {
```
eregon: This now needs a `@TruffleBoundary` since it uses String concatenation.

truffle/src/main/java/org/jruby/truffle/core/exception/SystemCallErrorNodes.java
```diff
@@ -0,0 +1,42 @@
...
+    public abstract static class ErrnoNode extends CoreMethodArrayArgumentsNode {
+
+        @Specialization
+        public Object errno(DynamicObject self) {
+            return Layouts.SYSTEM_CALL_ERROR.getErrno(self);
+        }
+
+    }
+
+    @Primitive(name = "errno_set")
```
eregon: Usually primitives are prefixed with some kind of namespace. Maybe `exception_set_errno` here?

truffle/src/main/java/org/jruby/truffle/core/exception/SystemCallErrorLayout.java
```diff
@@ -0,0 +1,34 @@
...
+
+    DynamicObjectFactory createSystemCallErrorShape(
+        DynamicObject logicalClass,
+        DynamicObject metaClass);
+
+    DynamicObject createSystemCallError(
+        DynamicObjectFactory factory,
+        @Nullable Object message,
+        @Nullable Backtrace backtrace,
+        @Nullable Object errno);
```
eregon: Can it be `null`? If not omit the `@Nullable`.

truffle/src/main/java/org/jruby/truffle/core/exception/SystemCallErrorNodes.java
```diff
@@ -0,0 +1,42 @@
...
+        public Object setErrno(DynamicObject error, int errno) {
+            Layouts.SYSTEM_CALL_ERROR.setErrno(error, errno);
+            return errno;
+        }
+
+        @Specialization(guards = "isNil(errno)")
+        public Object setErrno(DynamicObject error, DynamicObject errno) {
+            Layouts.SYSTEM_CALL_ERROR.setErrno(error, errno);
+            return errno;
+        }
```
eregon: One of them is enough since it's stored as an object.

eregon:
> It looks like MRI uses hidden instance vars for these kinds of internal fields. Is using layouts our best equivalent for this?

Currently I believe yes. There are many potential other ways to hide them, but that's the only currently implemented. `noMethodError.define("@name", ...` is pretty slow so the layout is better perf-wise.

eregon:
Looks good, let's just wait for CI and then let's merge it.

truffle/src/main/java/org/jruby/truffle/core/exception/SystemCallErrorLayout.java
nirvdrum: Can `errno` ever be anything other than an `int`?

bjfish: @nirvdrum I think it will only be int or null, should `@Nullable Integer` work? I think I was having trouble trying to use `@Nullable Integer`.

https://github.com/ruby/ruby/blob/trunk/error.c#L1486

nirvdrum: I guess it was a follow-up to Benoit's question. I didn't think the `errno` value could be `null`.

eregon: I guess we could make it an `int` and use some negative value to say "not available".

bjfish: @eregon An issue with that is that uses can provide any error number including negative numbers:
```
irb(main):003:0> e = SystemCallError.new(-111)
=> #<SystemCallError: Unknown error: -111>
irb(main):004:0> e.errno
=> -111
```

chrisseaton: `errno` can be `nil`

```
irb(main):002:0> SystemCallError.new('foo').errno
=> nil
```

eregon: @bjfish Good catch.
Yes I think @bjfish meant `nil` instead of `null` up there.

PR #4007, opened 2016-07-11
Remove duplicated line.

snim2:
This PR removes an unnecessary line of code and some trailing whitespace.


eregon:
Thank you for the fix!

Issue #3976, opened 2016-06-22
Superclass mixin implementation called instead of subclass implementation


manigandan-rajasekar:
### Environment

jruby 9.1.2.0 (2.3.0) 2016-05-26 7357c8f Java HotSpot(TM) 64-Bit Server VM 25.91-b14 on 1.8.0_91-b14 +jit [darwin-x86_64]

Darwin 15.5.0 Darwin Kernel Version 15.5.0:

### Expected Behavior

The '<' comparison operation should return true for any inputs (contrived example) for both classes. 
This is the behavior seen in CRuby.

```
StringSubclass Is a<b? true
StringSubclass Is b<c? true
StringSubclass Is b<a? true
NotStringSubclass Is a<b? true
NotStringSubclass Is b<c? true
NotStringSubclass Is b<a? true
```

```
class StringSubclass < String
  include Comparable

  def <=>(input)
    #puts "JRuby Never executes this code."
    return -1;
  end
end

class NotStringSubclass
  include Comparable

  def initialize(str)
    @str = str
  end

  def <=>(input)
    #puts "JRuby Executes this code since there is no superclass."
    return -1;
  end
end

puts "StringSubclass Is a<b? " + (StringSubclass.new("a")<StringSubclass.new("b")).to_s
puts "StringSubclass Is b<c? " + (StringSubclass.new("b")<StringSubclass.new("c")).to_s
puts "StringSubclass Is b<a? " + (StringSubclass.new("b")<StringSubclass.new("a")).to_s

puts "NotStringSubclass Is a<b? " + (NotStringSubclass.new("a")<NotStringSubclass.new("b")).to_s
puts "NotStringSubclass Is b<c? " + (NotStringSubclass.new("b")<NotStringSubclass.new("c")).to_s
puts "NotStringSubclass Is b<a? " + (NotStringSubclass.new("b")<NotStringSubclass.new("a")).to_s

```
### Actual Behavior

Instead, the '<' comparison operation returns true/false based on the String class implementation of the mixin function for StringSubclass. NotStringSubclass still behaves as expected.

```
StringSubclass Is a<b? true
StringSubclass Is b<c? true
StringSubclass Is b<a? false
NotStringSubclass Is a<b? true
NotStringSubclass Is b<c? true
NotStringSubclass Is b<a? true
```

headius:
[...]

PR #4009, opened 2016-07-12
Remove hardcoded base impl

bmulvihill:
This removes the hardcoded base implementation for the op_lt, op_le, op_gt, and op_ge methods

Based on https://github.com/jruby/jruby/issues/3976


headius:
This looks fine. We'll confirm Travis is green, then I'll merge and see if there's any broken tests we can enable.

We'll probably introduce this optimization again in a safe way once I set up a mechanism for knowing whether a core class method has been overwritten.

headius:
As it happened, I started introducing the aforementioned mechanism for other reasons. I will reject this in favor of the optimized approach. Thanks for your help!

PR #4017, opened 2016-07-14
[Truffle] Refactor sprintf parser/builder, update format nodes

bjfish:
I've updated to a new parser to match the spec more closely. Also, did some updates to format nodes.

I still need to continue implementing more formatting here but looking for feedback now.


truffle/src/main/java/org/jruby/truffle/core/format/format/FormatFloatNode.java
```diff
@@ -71,32 +109,42 @@ public FormatFloatNode(RubyContext context, int width, int precision, char forma
...
 
-    @TruffleBoundary
-    @Specialization(guards = "isInfinite(value)")
-    public byte[] formatInfinite(double value) {
-        final String infinityString = String.format(infiniteFormatString, value);
-        final String shortenInfinityString = infinityString.substring(0, infinityString.length() - 5);
-        return shortenInfinityString.getBytes(StandardCharsets.US_ASCII);
-    }
+    private String getInfiniteFormatString(final int width){
+        final StringBuilder inifiniteFormatBuilder = new StringBuilder();
```
eregon: inifinite?

truffle/src/main/java/org/jruby/truffle/core/format/format/FormatFloatNode.java
```diff
@@ -15,42 +15,80 @@
...
 
-        infiniteFormatString = inifiniteFormatBuilder.toString();
+    private static String mapFiniteResult(String input){
+        if(input.contains("NAN")){
+            return input.replaceFirst("NAN", "NaN");
+        }
+        return input;
+    }
 
+    private static String mapInifityResult(String input){
```
eregon: typo

truffle/src/main/java/org/jruby/truffle/core/format/format/FormatIntegerNode.java
```diff
@@ -10,63 +10,200 @@
...
-            },
-            limit = "getLimit()"
-    )
-    public byte[] formatCached(int width,
-                               Object value,
-                               @Cached("width") int cachedWidth,
-                               @Cached("makeFormatString(width)") String cachedFormatString) {
-        return doFormat(value, cachedFormatString);
+    @Specialization
+    public byte[] format(int width, int precision, int value) {
```
eregon: This kind of large method won't compile well, String concatenation, a large switch, etc.
Probably it can be just a boundary since doFormat was already a boundary.

eregon:
So this essentially the Antlr parser with a hand-written one?
Is Antlr not flexible or practical here?
The disadvantage is the whole parsing state is mixed with validation logic, the parser logic is re-implemented, and it's probably more prone to bugs in parsing.
OTOH, it might be interesting performance-wise.

bjfish:
@eregon It was getting difficult to match the MRI behavior with the ANTLR parsing, especially its behaviors for parsing errors. I think it might be possible to get the same behavior with an ANTLR parser/listener but I think it would take a lot more effort to get it correct compared to this new parsing.

truffle/src/main/java/org/jruby/truffle/core/format/format/FormatIntegerNode.java
```diff
@@ -10,63 +10,200 @@
...
+                throw new UnsupportedOperationException();
+        }
+        return formatStart(width, precision, formatted, value < 0);
+    }
+
+    private static byte[] elide(byte[] bytes, char format, int precision) {
+        return bytes;
+    }
+
+
```
chrisseaton: No double blank lines, please.

chrisseaton:
It's a shame we can't get Antlr to do what we want. I did struggle to get the error handling behaviour as good as it was before. I found I was adding lots of parse rules that could then be matched as the required error. Of course, passing specs trumps almost all else, so we can go with this instead.

I don't think I'm going to try writing the Ruby parser in Antlr that I was planning.

truffle/src/main/java/org/jruby/truffle/core/format/format/FormatCharacterNode.java
```diff
@@ -0,0 +1,85 @@
...
+
+    @Child private ToIntegerNode toIntegerNode;
+    @Child private ToStringNode toStringNode;
+
+    public FormatCharacterNode(RubyContext context, boolean hasMinusFlag) {
+        super(context);
+        this.hasMinusFlag = hasMinusFlag;
+    }
+
+//    @TruffleBoundary
```
chrisseaton: We try not to leave commented code like this in. At least it should be explained why it is currently commented.

truffle/src/main/java/org/jruby/truffle/core/format/format/FormatCharacterNode.java
```diff
@@ -0,0 +1,85 @@
...
+        } else {
+            final String resultString = new String((byte[]) toStrResult);
+            final int size = resultString.length();
+            if (size > 1) {
+                throw new RaiseException(getContext().getCoreExceptions().argumentError("%c requires a character", this));
+            }
+            charString = resultString;
+        }
+
+
```
chrisseaton: No double blank lines, please.

truffle/src/main/java/org/jruby/truffle/core/format/format/FormatCharacterNode.java
```diff
@@ -0,0 +1,85 @@
...
+                toIntegerNode = insert(ToIntegerNodeGen.create(getContext(), null));
+            }
+            final int charValue = (int) toIntegerNode.executeToInteger(frame, value);
+            // TODO BJF check char length is > 0
+            charString = Character.toString((char) charValue);
+        } else {
+            final String resultString = new String((byte[]) toStrResult);
+            final int size = resultString.length();
+            if (size > 1) {
+                throw new RaiseException(getContext().getCoreExceptions().argumentError("%c requires a character", this));
```
chrisseaton: I'm trying to put all messages into the implementation of `CoreExceptions`, so that they're not spread over the code base and in the future I can replace them with rope literals.

truffle/src/main/java/org/jruby/truffle/core/format/format/FormatCharacterNode.java
```diff
@@ -0,0 +1,85 @@
...
+            charString = resultString;
+        }
+
+
+        final boolean leftJustified = hasMinusFlag || width < 0;
+        if (width < 0) {
+            width = -width;
+        }
+
+        final String result = String.format("%" + (leftJustified ? "-" : "") + width + "." + width + "s", charString);
```
chrisseaton: This'll need to be boundary.

truffle/src/main/java/org/jruby/truffle/core/format/format/FormatFloatNode.java
chrisseaton: I can't see what 'map' means in this context?

truffle/src/main/java/org/jruby/truffle/core/format/format/FormatIntegerNode.java
chrisseaton: Can't most of this be cached?

truffle/src/main/java/org/jruby/truffle/core/format/format/FormatIntegerNode.java
```diff
@@ -10,63 +10,200 @@
...
+    private byte[] formatStart(int width, int precision, String formatted, boolean isNegative) {
+
```
chrisseaton: No blank line here, please.

truffle/src/main/java/org/jruby/truffle/core/format/printf/PrintfSimpleParser.java
```diff
@@ -0,0 +1,592 @@
...
+        public int getNextI() {
+            return nextI;
+        }
+    }
+
+    public static boolean isDigit(char c) {
+        return c == '0' || c == '1' || c == '2' || c == '3' || c == '4' || c == '5' || c == '6' || c == '7' || c == '8' || c == '9';
+    }
+
+    public static class SprintfConfig {
```
chrisseaton: I'm quite against nested classes. I think they tend to produce source files that are very hard to browse and read. What do other people think?

truffle/src/main/java/org/jruby/truffle/core/format/printf/PrintfSimpleTreeBuilder.java
```diff
@@ -0,0 +1,197 @@
+package org.jruby.truffle.core.format.printf;
```
chrisseaton: Missing copyright information - very important.

chrisseaton:
Are you still iterating on this? I can't quite understand it all so I might check it out and browse it there if you're done. Some of the code needs a bit of an aesthetic tidy up (random extra blank lines and things).

bjfish:
@chrisseaton I'm going to make all the recommended code review updates here to hopefully get it into good shape to merge. I can ping you after this update.

Then, I need to give the `Integer` and `Float` and other nodes more attention to pass more specs, be more readable, and performant as needed.

chrisseaton:
This code was never fast-path was it? We always had the boundary in there? We should add some micro benchmarks (ask me how if you don't know) and then check we are at least the same speed as JRuby.

truffle/src/main/java/org/jruby/truffle/core/format/printf/PrintfSimpleParser.java
eregon: If it's some kind of private implementation details and relatively small I think static nested classes are fine.
But here it's neither so I would move it outside as well.

bjfish:
@eregon @chrisseaton I've added another commit which addresses the code review comments. Thanks for the feedback.

I'm not sure if the FormatIntegerNode specialization was fast path. I can look into adding back caching for this.

truffle/src/main/java/org/jruby/truffle/core/format/format/FormatCharacterNode.java
```diff
@@ -0,0 +1,96 @@
...
+        if (width < 0) {
+            width = -width;
+        }
+
+        return doFormat(charString, width, leftJustified).getBytes(StandardCharsets.US_ASCII);
+    }
+
+    @TruffleBoundary
+    private String doFormat(final String charString, final int width, final boolean leftJustified) {
+        return String.format("%" + (leftJustified ? "-" : "") + width + "." + width + "s", charString);
```
chrisseaton: Can you just cache this generated format string? Doing that is trivial and should be an easy performance win until we implement our own formatting code.

truffle/src/main/java/org/jruby/truffle/core/format/format/FormatIntegerNode.java
```diff
@@ -10,67 +10,203 @@
...
+                    formatted = "..f" + Long.toString(-value, 16);
+                }
+                break;
+            case 'o':
+                formatted = Long.toString(value, 8);
+                if (hasFSharp && value > 0) {
+                    formatted = "0" + formatted;
+                }
+                if (value < 0) {
+                    formatted = "..7" + new String(elide(ConvertBytes.longToOctalBytes(value), format, precision));
```
chrisseaton: What's this `..f` and `..7` for? I've not seen that in any output that I can think of.

bjfish: @chrisseaton It's for hex, octal, and binary negative numbers. E.g.:
```
"%b" % -3 => "..101"
"%o" % -3 => "..75"
"%x" % -3 => "..fd"
```

chrisseaton: Wow I've never seen that before.

truffle/src/main/java/org/jruby/truffle/core/format/format/FormatCharacterNode.java
bjfish: @chrisseaton I've added a commit to cache this format string

chrisseaton:
Is this ready to merge then?

bjfish:
@chrisseaton Yes, I think it's ready to merge.

PR #4025, opened 2016-07-21
[Truffle] Add NameError layout with receiver

bjfish:


truffle/src/main/java/org/jruby/truffle/core/exception/NameErrorLayout.java
```diff
@@ -0,0 +1,38 @@
+/*
+ * Copyright (c) 2015, 2016 Oracle and/or its affiliates. All rights reserved. This
```
nirvdrum: If you didn't copy this code from somewhere else, it should probably just be 2016.

truffle/src/main/java/org/jruby/truffle/core/exception/NameErrorNodes.java
```diff
@@ -0,0 +1,64 @@
+/*
+ * Copyright (c) 2015, 2016 Oracle and/or its affiliates. All rights reserved. This
```
nirvdrum: I think this should just be 2016.

truffle/src/main/java/org/jruby/truffle/core/exception/NameErrorNodes.java
```diff
@@ -0,0 +1,64 @@
...
+    @CoreMethod(names = "receiver")
+    public abstract static class ReceiverNode extends CoreMethodArrayArgumentsNode {
+
+        @Specialization
+        public Object receiver(DynamicObject self) {
+            final Object receiver = Layouts.NAME_ERROR_LAYOUT.getReceiver(self);
+
+            // TODO BJF July 21, 2016 Implement name error in message field
+
+            if(receiver == null){
```
nirvdrum: Minor, but all the code has spaces between `if` and `(`.

bjfish:
@nirvdrum Thanks, I've fixed the copyright and style issue

truffle/src/main/java/org/jruby/truffle/core/exception/NameErrorNodes.java
```diff
@@ -0,0 +1,64 @@
...
+
+            if (receiver == null) {
+                throw new RaiseException(coreExceptions().argumentErrorNoReceiver(this));
+            }
+            return receiver;
+        }
+
+    }
+
+    @Primitive(name = "exception_set_name")
```
eregon: `name_error_set_name` since this only makes sense for name errors.

truffle/src/main/java/org/jruby/truffle/core/exception/NameErrorLayout.java
```diff
@@ -0,0 +1,38 @@
...
+    DynamicObjectFactory createNameErrorShape(
+        DynamicObject logicalClass,
+        DynamicObject metaClass);
+
+    DynamicObject createNameError(
+        DynamicObjectFactory factory,
+        Object message,
+        @Nullable Backtrace backtrace,
+        Object name,
+        @Nullable Object receiver);
```
eregon: Can we have them in the order `receiver`, `name`?
That seems more intuitive to me (`recv.foo()`, `A::B`).

truffle/src/main/java/org/jruby/truffle/core/exception/NameErrorNodes.java
eregon: There is unfortunately no automatic namespacing for primitives, it's a single flat namespace.

truffle/src/main/java/org/jruby/truffle/core/exception/NameErrorLayout.java
bjfish: I could do this update. There may be one small issue is the code above `allocateObjectNode.allocate(rubyClass, nil(), null, nil(), null);` which depends on the ordering of these and it may conflict with other Error layouts.  Should the `allocate` method be called differently depending on the Error class being allocated?

cc: @chrisseaton because I think you made a change to `allocate` recently.

chrisseaton: Yes I just fixed a problem that was introduced when an exception layout was changed and layout sites weren't updated to match. The error was only apparent in compiled code.

I think you can give exception subclasses different layouts so long as you define your own `#allocate` method for each subclass. We aren't doing that for `SystemCallError` at the moment, but we should be.

eregon: Yes, `#allocate` *must* be defined for every class with a custom layout.

chrisseaton: Well, one trick is that if you subclass only adds fields you can pass all the fields to each layout and the super-layout will ignore those it doesn't use. Probably not a good idea to rely on this.

eregon: Indeed, I would even add an assert against this in the DynamicObjectFactory.

bjfish:
@eregon @chrisseaton  I've re-ordered the receiver, name args and added a specialization in the allocate node for NameError, is there any better way to implement the guard there: `public static boolean isNameError(DynamicObject rubyClass)` ?

truffle/src/main/java/org/jruby/truffle/core/exception/ExceptionNodes.java
```diff
@@ -133,11 +133,25 @@ public AllocateNode(RubyContext context, SourceSection sourceSection) {
...
-        @Specialization
+        @Specialization(guards = "!isNameError(rubyClass)")
         public DynamicObject allocate(DynamicObject rubyClass) {
             return allocateObjectNode.allocate(rubyClass, nil(), null, nil());
         }
 
+        @Specialization(guards = "isNameError(rubyClass)")
+        public DynamicObject allocateNameError(DynamicObject rubyClass) {
+            return allocateObjectNode.allocate(rubyClass, nil(), null, null, nil());
+        }
```
eregon: I would define this on NameError, not in Exception so there is no need for guards and it's automatically using the right one.

truffle/src/main/java/org/jruby/truffle/core/exception/ExceptionNodes.java
```diff
@@ -133,9 +133,23 @@ public AllocateNode(RubyContext context, SourceSection sourceSection) {
...
+
+        @Specialization(guards = "isNameError(rubyClass)")
+        public DynamicObject allocateNameError(DynamicObject rubyClass) {
+            return allocateObjectNode.allocate(rubyClass, nil(), null, null, nil());
+        }
+
+        public static boolean isNameError(DynamicObject rubyClass){
+           for(DynamicObject ancestor : Layouts.MODULE.getFields(rubyClass).ancestors()){
+                  if(Layouts.MODULE.getFields(ancestor).getName().equals("NameError")){
```
nirvdrum: Spaces between `for`, `if` and `(` please. I think we universally do them between `)` and `{` as well. Granted, our style is a bit loose.

eregon: Maybe we should have a checkstyle for simple rules.
I have use Eclipse's formatter too but it needs to be convenient for other IDEs as well.

truffle/src/main/java/org/jruby/truffle/core/exception/ExceptionNodes.java
eregon: So you'll need a NameErrorNodes.java, or define it in Ruby, invoking a primitive.
NameErrorNodes makes sense since it's not just one method.

truffle/src/main/java/org/jruby/truffle/core/exception/ExceptionNodes.java
nirvdrum: I know you asked me about this and I don't think I fully understood what you were doing at the time. If you really have the class handle and you're looking for a direct class match (rather an "is a" relationship), you should be able to do a reference comparison to `coreLibrary().getNameErrorClass()`.

truffle/src/main/java/org/jruby/truffle/core/exception/ExceptionNodes.java
bjfish: @eregon I've moved this specialization to `NameErrorNodes`

PR #4027, opened 2016-07-23
[Truffle] Add NoMethodError layout with args

bjfish:


eregon:
Looks good! :+1:

truffle/src/main/ruby/core/exception.rb
```diff
@@ -189,12 +189,11 @@ def initialize(*args)
 end
 
 class NoMethodError < NameError
-  attr_reader :args
 
   def initialize(*arguments)
     super(arguments.shift)
     @name = arguments.shift
```
eregon: Is this going to work?
Maybe at this point it would make sense to move `initialize` of exceptions in Java.

bjfish: @eregon I don't think this `initialize` method works as is. I will update this method.

truffle/src/main/ruby/core/exception.rb
```diff
@@ -183,18 +183,18 @@ class LocalJumpError < StandardError
...
 class NoMethodError < NameError
-  attr_reader :args
 
   def initialize(*arguments)
-    super(arguments.shift)
-    @name = arguments.shift
-    @args = arguments.shift
+    args = arguments.size > 2 ? arguments.pop : nil
+    super(*arguments) # TODO BJF Jul 24, 2016 Need to handle NoMethodError.new(1,2,3,4)
+    Truffle.invoke_primitive :no_method_error_set_args, self, args
```
bjfish: @eregon I've updated these to match more closely the expected behavior. There is still an edge case here that isn't handled yet mentioned in the `TODO`. It looks like there will be a new field in 2.4 and this edge case behavior is also better documented there.

PR #4038, opened 2016-07-29
Fix some parts of Socket::Option

etehtsea:
Specs passed, so, hopefully, it should work.

P.S. Also I wanted to fix `Socket.linger`, but it requires more substantial changes and I put it aside for a while.

Disclaimer:
1 sntxfelfuqlsfdnvzkj6ia

I will be happy to see some feedback how it could be improved.


core/src/main/java/org/jruby/ext/socket/Option.java
```diff
@@ -166,24 +167,34 @@ private String optionValue() {
...
 
-    @JRubyMethod(meta = true)
-    public IRubyObject bool(ThreadContext context, IRubyObject self) {
-        return context.nil;
+    @JRubyMethod(required = 4, meta = true)
+    public static IRubyObject bool(ThreadContext context, IRubyObject self, IRubyObject[] args) {
+        ProtocolFamily family = SocketUtils.protocolFamilyFromArg(args[0]);
+        SocketLevel level = RubyBasicSocket.levelFromArg(args[1]);
+        SocketOption option = RubyBasicSocket.optionFromArg(args[2]);
+        int intData = args[3].isTrue() ? 1 : 0;
```
etehtsea: Should it be checked there that it actually a boolean arg and what to do in case if it's not?

Issue #4051, opened 2016-08-06
Wrong line numbers in Kernel#caller inside set_trace_func (with code to reproduce)


ivoanjo:
I was trying to see if I could get [`pry-nav`](https://github.com/nixme/pry-nav) to work with JRuby. It doesn't have a lot of features, but it's a straightforward use of `set_trace_func` and I wanted to experiment with it a bit.

I hit some issues, and investigating them led me to discover that when running code inside `set_trace_func` and then calling `Kernel#caller` (such as when stepping) leads to line numbers being wrong in the stack trace, which confuses `pry`.

### Environment

Running `jruby 9.1.2.0 (2.3.0) 2016-05-26 7357c8f OpenJDK 64-Bit Server VM 25.91-b14 on 1.8.0_91-8u91-b14-3ubuntu1~15.10.1-b14 [linux-x86_64]` on `Ubuntu 15.10`.

### Expected Behavior

Example code:
```ruby
def sum(arg1, arg2)
  arg1 + arg2
end

def crazy_sums
  sum(1, 2)
end

set_trace_func proc { |type, _, line, *stuff|
  unless ['c-return', 'c-call'].include? type
    puts "Event #{type} @ #{line}, caller: #{caller.first}"
  end
}

crazy_sums
```

Output in MRI 2.3.1:
```
Event line @ 15, caller: test.rb:15:in `<main>'
Event call @ 5, caller: test.rb:5:in `crazy_sums'
Event line @ 6, caller: test.rb:6:in `crazy_sums'
Event call @ 1, caller: test.rb:1:in `sum'
Event line @ 2, caller: test.rb:2:in `sum'
Event return @ 3, caller: test.rb:3:in `sum'
Event return @ 7, caller: test.rb:7:in `crazy_sums'
```
With MRI, `caller.first` matches the line we get in `set_trace_func`.

### Actual Behavior

Output in JRuby:
```
Event line @ 15, caller: test.rb:9:in `<top>'
Event call @ 5, caller: test.rb:15:in `crazy_sums'
Event line @ 6, caller: test.rb:15:in `crazy_sums'
Event call @ 1, caller: test.rb:6:in `sum'
Event line @ 2, caller: test.rb:6:in `sum'
Event return @ 3, caller: test.rb:2:in `sum'
Event return @ 7, caller: test.rb:6:in `crazy_sums'
```
With JRuby, the line numbers don't match the line we get in `set_trace_func`.

[...]

PR #4052, opened 2016-08-06
Partially fix set_trace_func line numbers

ivoanjo:
After reporting issue #4051 I decided to look at the generated IR and noticed that trace instructions were being added without line numbers. This led to the observed issue there stack traces generated inside `set_trace_func` did not have the correct line numbers.

For the `LINE` and `CALL` events the fix is straightforward: just add a line number instruction **before** the trace instruction, as the line number was already known statically by the `IRBuilder`.

Unfortunately, for the `RETURN` and `END` events the parser does not include this information and thus there's no simple way of adding line number instructions before those. I think that particular fix is a bit beyond my capabilities so I added some `FIXME`s so at least this doesn't get forgotten.

Nevertheless, and happily, the partial fix seems to be enough to get [`pry-nav`](https://github.com/ivoanjo/pry-nav) to work!

On a side note, I must admit that

but nevertheless for a first time diving inside JRuby's codebase, this went better than expected :smile: 

Please let me know if there's some way I can help fix this further!

Issue #4051

### DIFF
--- a/core/src/main/java/org/jruby/ir/IRBuilder.java
+++ b/core/src/main/java/org/jruby/ir/IRBuilder.java
@@ -965,6 +965,10 @@ public class IRBuilder {
         addInstr(new ReceiveJRubyExceptionInstr(exc));
 
         if (RubyInstanceConfig.FULL_TRACE_ENABLED) {
+            // FIXME: We also should add a line number instruction so that backtraces
+            // inside the trace function get the correct line. Unfortunately, we don't
+            // have one here and we can't do it dynamically like TraceInstr does.
+            // See https://github.com/jruby/jruby/issues/4051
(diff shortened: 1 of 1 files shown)


ivoanjo:
Update: Along with [`pry-nav`](https://github.com/ivoanjo/pry-nav), I have now gotten [`pry-debugger`](https://github.com/ivoanjo/pry-debugger) to work too.

Woohoo, finally breakpoints and flow control inside pry! :tada:

enebo:
@ivoanjo both of those changes make total sense to me so that is good :)  I am happy to here some libraries are working.  I will try and better understand the end/return line number problems to see if I can totally put this to bed.

ivoanjo:
Awesome, thanks for the quick review!

PR #4054, opened 2016-08-08
RubyEnumerator implements java.util.Iterator

kares:
feels like a good fit and this makes enumerator usable with Java 8 (with some boilerplate) : 
`java.util.Spliterators.spliterator(enum.to_java, enum.size || -1, charactestistics)`

... ideally some more integration would be handy but JRuby would need compiling under Java 8 for easy : `[1, 2, 3, 4, 5].each.to_java.stream` or `Enumerator.new { ... }.to_java.spliterator` implementations.
esp. since the Enumeration's size and the spliterator characteristic are depend on internals. so this would like get revisited after a potential compilation switch in another release.

Ruby enumerator (and thus Ruby array) usability with 8 streaming is confirmed with test-cases.

### DIFF
--- a/core/src/main/java/org/jruby/RubyEnumerator.java
+++ b/core/src/main/java/org/jruby/RubyEnumerator.java
@@ -501,7 +501,8 @@ public class RubyEnumerator extends RubyObject {
 
     @JRubyMethod
     public synchronized IRubyObject next(ThreadContext context) {
-        ensureNexter(context);
+        ensureNexter(context.runtime);
+
         if (!feedValue.isNil()) feedValue = context.nil;
         return nexter.next();
     }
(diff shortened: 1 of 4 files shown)


kares:
> ... ideally some more integration would be handy but JRuby would need compiling under Java 8 for easy : [1, 2, 3, 4, 5].each.to_java.stream or Enumerator.new { ... }.to_java.spliterator implementations.
esp. since the Enumeration's size and the spliterator characteristic are depend on internals. so this would like get revisited after a potential compilation switch in another release.

also possible by hooking up `stream`/`spliterator` on `org.jruby.RubyEnumerator` (in plain .rb for now)

// cc @headius @enebo let me know what do you think about getting this merged for **9.1.3**

headius:
@kares 👍  from me. @enebo?

kares:
thanks for the review, have added the .rb part helpers (under Java 8) `spliterator` and `stream` since due compilation under Java 7 those can not be implemented in `org.jruby.RubyEnumerator` atm

enebo:
@kares looks fine by me

PR #4058, opened 2016-08-09
[Truffle] Add CExt test for ruby-argon2

lxp:
Depends on https://github.com/jruby/jruby-truffle-gem-test-pack/pull/1

### DIFF
--- a/tool/jt.rb
+++ b/tool/jt.rb
@@ -30,6 +30,7 @@ JDEBUG = "-J-agentlib:jdwp=transport=dt_socket,server=y,address=#{JDEBUG_PORT},s
 JDEBUG_TEST = "-Dmaven.surefire.debug=-Xdebug -Xrunjdwp:transport=dt_socket,server=y,suspend=y,address=#{JDEBUG_PORT} -Xnoagent -Djava.compiler=NONE"
 JEXCEPTION = "-Xtruffle.exceptions.print_java=true"
 METRICS_REPS = 10
+CEXTC_CONF_FILE = '.jruby-cext-build.yml'
 
 VERBOSE = ENV.include? 'V'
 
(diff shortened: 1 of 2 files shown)


tool/jt.rb
```diff
@@ -838,17 +847,20 @@ def test_cexts(*args)
     [
         ['oily_png', ['chunky_png-1.3.6', 'oily_png-1.2.0'], ['oily_png']],
         ['psd_native', ['chunky_png-1.3.6', 'oily_png-1.2.0', 'bindata-2.3.1', 'hashie-3.4.4', 'psd-enginedata-1.1.1', 'psd-2.1.2', 'psd_native-1.1.3'], ['oily_png', 'psd_native']],
-        ['nokogiri', [], ['nokogiri']]
-    ].each do |gem_name, dependencies, libs|
+        ['nokogiri', [], ['nokogiri']],
+        ['ruby-argon2', [], ['ruby-argon2'], {'GEM_HOME' => "#{ENV['GEM_HOME']}/bundler/gems/ruby-argon2-afe6d65e37c6", 'TEST_CHECKS' => '1'}]
```
eregon: Should GEM_HOME be set to a specific gem like that?

eregon: I fixed `jt cextc` so this should no longer be needed only
` ['ruby-argon2', [], ['ruby-argon2']],`.
For `TEST_CHECK`, maybe the default can be `1` in the test harness?
Configuration is always easier to keep closer to the source :)

eregon: We need a way to specify the path to the argon2 dir in the test pack though, probably can be an extra argument in this list.

pitr-ch:
jruby/jruby-truffle-gem-test-pack#1 looks good

PR #4061, opened 2016-08-10
Fix Socket.tcp connect_timeout option

etehtsea:
Before commit `Socket.tcp(host, port, connect_timeout: 0.1)`
ignored timeout option. This happened because `connect_internal`
mistakenly responded with `:wait_readable` instead `:wait_writable`.
After fixing it this code started to constantly fail with timeout error.
Checking for `OP_CONNECT` in `IO#wait_writable` fixed this issue.

Added adopted spec from MRI, because it's excluded by other reasons.


eregon:
@etehtsea This looks good, it would only need a guard to say it's for a Ruby >= 2.3.
Why did you close it?

etehtsea:
@eregon https://github.com/ruby/spec/pull/279 fixed

PR #4063, opened 2016-08-11
Explicitly close input stream in CompoundJarURLStreamHandler

camlow325:
This commit ensures that the stream which is opened for a `URL` within
`CompoundJarURLStreamHandler` is explicitly closed before it falls out
of scope.  Previously, the stream would still eventually be closed but
only after Garbage Collection.


enebo:
@kares should this get merged to master?

camlow325:
@enebo It didn't look to me like the `CompoundJarURLStreamHandler` class exists in the current code on master.  Is that right?  Also, I did some testing with fairly recent release off of master, 9.1.2.0, and wasn't able to reproduce the issue with an excessive number open file descriptors from the "stdlib" jar.  Do you think there's anything worth merging to master from this, then?

enebo:
@camlow325 you are correct!  It is not there.  I was just double checking since I saw a PR merged to 1.7.  Better to ask than forget.

PR #4081, opened 2016-08-16
Truffle fix match backref

nirvdrum:
@jruby/truffle 

This fixes $~ to be frame-local (and thread-local), which addresses some issues where values were leaking in places they shouldn't have been. I don't think I've handled every case accurately, but I do think this is more accurate than what we had before and all specs continue to pass.

Since it was a bit involved, I'd appreciate a careful review.

### DIFF
--- a/truffle/src/main/java/org/jruby/truffle/core/regexp/RegexpNodes.java
+++ b/truffle/src/main/java/org/jruby/truffle/core/regexp/RegexpNodes.java
@@ -100,15 +100,15 @@ public abstract class RegexpNodes {
     }
 
     @TruffleBoundary
-    public static Object matchCommon(RubyContext context, Node currentNode, RopeNodes.MakeSubstringNode makeSubstringNode, DynamicObject regexp, DynamicObject string, boolean operator,
+    public static DynamicObject matchCommon(RubyContext context, Node currentNode, RopeNodes.MakeSubstringNode makeSubstringNode, DynamicObject regexp, DynamicObject string,
             boolean setNamedCaptures, int startPos) {
         final Matcher matcher = createMatcher(context, regexp, string);
         int range = StringOperations.rope(string).byteLength();
-        return matchCommon(context, currentNode, makeSubstringNode, regexp, string, operator, setNamedCaptures, matcher, startPos, range);
+        return matchCommon(context, currentNode, makeSubstringNode, regexp, string, setNamedCaptures, matcher, startPos, range);
     }
 
     @TruffleBoundary
-    public static Object matchCommon(RubyContext context, Node currentNode, RopeNodes.MakeSubstringNode makeSubstringNode, DynamicObject regexp, DynamicObject string, boolean operator,
+    public static DynamicObject matchCommon(RubyContext context, Node currentNode, RopeNodes.MakeSubstringNode makeSubstringNode, DynamicObject regexp, DynamicObject string,
             boolean setNamedCaptures, Matcher matcher, int startPos, int range) {
         assert RubyGuards.isRubyRegexp(regexp);
         assert RubyGuards.isRubyString(string);
@@ -191,11 +191,7 @@ public abstract class RegexpNodes {
             }
         }
 
-        if (operator) {
-            return matcher.getBegin();
-        } else {
-            return matchData;
-        }
+        return matchData;
     }
 
     // because the factory is not constant
@@ -392,17 +388,29 @@ public abstract class RegexpNodes {
     public abstract static class MatchOperatorNode extends CoreMethodArrayArgumentsNode {
 
         @Child private RopeNodes.MakeSubstringNode makeSubstringNode;
+        @Child private RegexpSetLastMatchPrimitiveNode setLastMatchNode;
         @Child private CallDispatchHeadNode toSNode;
         @Child private ToStrNode toStrNode;
 
         public MatchOperatorNode(RubyContext context, SourceSection sourceSection) {
             super(context, sourceSection);
             makeSubstringNode = RopeNodesFactory.MakeSubstringNodeGen.create(null, null, null);
+            setLastMatchNode = RegexpPrimitiveNodesFactory.RegexpSetLastMatchPrimitiveNodeFactory.create(null);
         }
 
         @Specialization(guards = "isRubyString(string)")
         public Object match(DynamicObject regexp, DynamicObject string) {
-            return matchCommon(getContext(), this, makeSubstringNode, regexp, string, true, true, 0);
+            final Matcher matcher = createMatcher(getContext(), regexp, string);
+            final int range = StringOperations.rope(string).byteLength();
+
+            final DynamicObject matchData = matchCommon(getContext(), this, makeSubstringNode, regexp, string, true, matcher, 0, range);
+            setLastMatchNode.executeSetLastMatch(matchData);
+
+            if (matchData != nil()) {
+                return matcher.getBegin();
+            }
+
+            return nil();
         }
 
         @Specialization(guards = "isRubySymbol(symbol)")
@@ -456,7 +464,7 @@ public abstract class RegexpNodes {
 
         @Specialization(guards = "isRubyString(string)")
         public Object matchStart(DynamicObject regexp, DynamicObject string, int startPos) {
-            final Object matchResult = matchCommon(getContext(), this, makeSubstringNode, regexp, string, false, false, startPos);
+            final Object matchResult = matchCommon(getContext(), this, makeSubstringNode, regexp, string, false, startPos);
             if (RubyGuards.isRubyMatchData(matchResult) && Layouts.MATCH_DATA.getRegion((DynamicObject) matchResult).numRegs > 0
                 && Layouts.MATCH_DATA.getRegion((DynamicObject) matchResult).beg[0] == startPos) {
                 return matchResult;
@@ -510,7 +518,7 @@ public abstract class RegexpNodes {
 
         @Specialization(guards = "isRubyString(string)")
         public Object searchFrom(DynamicObject regexp, DynamicObject string, int startPos) {
-            return matchCommon(getContext(), this, makeSubstringNode, regexp, string, false, false, startPos);
+            return matchCommon(getContext(), this, makeSubstringNode, regexp, string, false, startPos);
         }
     }
 
--- a/truffle/src/main/java/org/jruby/truffle/core/regexp/RegexpNodes.java
+++ b/truffle/src/main/java/org/jruby/truffle/core/regexp/RegexpNodes.java
@@ -20,12 +20,15 @@ package org.jruby.truffle.core.regexp;
 
 import com.oracle.truffle.api.CompilerDirectives;
 import com.oracle.truffle.api.CompilerDirectives.TruffleBoundary;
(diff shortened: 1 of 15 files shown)


truffle/src/main/java/org/jruby/truffle/core/regexp/RegexpNodes.java
```diff
@@ -426,7 +430,25 @@ public Object matchGeneric(VirtualFrame frame, DynamicObject regexp, DynamicObje
...
+        private Object matchWithStringCopy(DynamicObject regexp, DynamicObject string) {
+            final Matcher matcher = createMatcher(getContext(), regexp, string);
+            final int range = StringOperations.rope(string).byteLength();
+
+            final DynamicObject matchData = matchCommon(getContext(), this, makeSubstringNode, regexp, string, true, matcher, 0, range);
+            setLastMatchNode.executeSetLastMatch(matchData);
+
+            if (matchData != nil()) {
+                return matcher.getBegin();
+            }
```
eregon: I'm not really following, where is the copy here?

truffle/src/main/java/org/jruby/truffle/core/regexp/RegexpNodes.java
```diff
@@ -446,24 +468,83 @@ public int hash(DynamicObject regexp) {
...
+            if (matchData == nil()) {
+                return nil();
+            } else {
+                return getIndexNode.executeGetIndex(frame, matchData, index, NotProvided.INSTANCE);
+            }
+        }
+
+        @TruffleBoundary
+        private Object getMatchData() {
+            Frame frame = getContext().getCallStack().getCallerFrameIgnoringSend().getFrame(FrameInstance.FrameAccess.READ_WRITE, true);
```
eregon: READ_WRITE access is fairly expensive, so maybe we should optimize by trying to access it in READ_ONLY first.
But here it seems the frame is never written to.

truffle/src/main/java/org/jruby/truffle/core/regexp/RegexpNodes.java
```diff
@@ -446,24 +468,83 @@ public int hash(DynamicObject regexp) {
...
+                    return nil();
+                } else {
+                    slot = nextFrame.getFrameDescriptor().findFrameSlot("$~");
+                    frame = nextFrame;
+                }
+            }
+
+            final Object previousMatchData;
+            try {
+                previousMatchData = frame.getObject(slot);
```
eregon: To avoid the exception, you could use `getValue()` here.

truffle/src/main/java/org/jruby/truffle/core/regexp/RegexpNodes.java
```diff
@@ -446,24 +468,83 @@ public int hash(DynamicObject regexp) {
...
+
+                if (previousMatchData instanceof ThreadLocalObject) {
+                    final ThreadLocalObject threadLocalObject = (ThreadLocalObject) previousMatchData;
+
+                    return threadLocalObject.get();
+                } else {
+                    return previousMatchData;
```
eregon: I would return an explicit `nil` here for clarity.

truffle/src/main/java/org/jruby/truffle/core/rubinius/RegexpPrimitiveNodes.java
```diff
@@ -110,48 +114,114 @@ private String formatError(DynamicObject string) {
...
 
-        @Specialization
-        public DynamicObject setBlockLastMatch(DynamicObject regexpClass) {
-            // TODO CS 7-Mar-15 what does this do?
-            return nil();
+        @TruffleBoundary
+        @Specialization(guards = { "isRubyProc(block)", "isSuitableMatchDataType(getContext(), matchData)" })
+        public Object setBlockLastMatch(DynamicObject block, DynamicObject matchData) {
+
+            Frame callerFrame = Layouts.PROC.getDeclarationFrame(block);
```
eregon: Could you rename this to `methodFrame`?

truffle/src/main/java/org/jruby/truffle/core/rubinius/RegexpPrimitiveNodes.java
```diff
@@ -110,48 +114,114 @@ private String formatError(DynamicObject string) {
...
+
+            if (callerFrame == null) {
```
eregon: Is this case possible? A block without a declaration frame sounds strange.

truffle/src/main/java/org/jruby/truffle/core/string/StringNodes.java
```diff
@@ -569,8 +570,20 @@ public Object slice1(
                 DynamicObject string,
                 DynamicObject regexp,
                 Object capture,
-                @Cached("new()") SnippetNode snippetNode) {
-            return snippetNode.execute(frame, "match, str = subpattern(index, 0); Regexp.last_match = match; str", "index", regexp);
+                @Cached("createMethodCallIgnoreVisibility()") CallDispatchHeadNode callNode,
+                @Cached("create()") RegexpPrimitiveNodes.RegexpSetLastMatchPrimitiveNode setLastMatchNode) {
+            final Object matchStrPair = callNode.call(frame, string, "subpattern", regexp, 0);
```
eregon: It looks like this could call the method below to avoid duplication.

truffle/src/main/ruby/core/string.rb
```diff
@@ -597,75 +597,14 @@ def upto(stop, exclusive=false)
...
-
-    ret.taint if tainted
-    ret.untrust if untrusted
-
-    ret
+  def sub(pattern, replacement=undefined, &block)
+    s = dup
+    s.sub!(pattern, replacement, &block)
+    Truffle.invoke_primitive(:regexp_set_last_match, $~)
+    s
```
eregon: Awesome! :+1:

truffle/src/main/java/org/jruby/truffle/core/regexp/RegexpNodes.java
```diff
@@ -436,7 +430,25 @@ public Object matchGeneric(VirtualFrame frame, DynamicObject regexp, DynamicObje
...
             }
 
-            return match(regexp, toStrNode.executeToStr(frame, other));
+            return matchWithStringCopy(regexp, toStrNode.executeToStr(frame, other));
+        }
+
+        // Creating a MatchData will store a copy of the source string. It's tempting to use a rope here, but a bit
+        // inconvenient because we can't work with ropes directly in Ruby (useful for implementing some MatchData
+        // methods). Likewise, we need to taint objects based on the source string's taint state. We mustn't allow the
+        // source string's contents to change, however, so we must ensure that we have a private copy of that string.
```
eregon: What about making a copy when creating the MatchData?

nirvdrum: That was going to get messy because it's all behind a boundary. And we'd end up duplicating in the couple cases above where we call `to_s` or `to_str`.

truffle/src/main/java/org/jruby/truffle/core/regexp/RegexpNodes.java
nirvdrum: It's being made by the caller. Not ideal, but it didn't seem very useful to allocate a string from `to_s` and then immediately `dup` it.

truffle/src/main/java/org/jruby/truffle/core/regexp/RegexpNodes.java
nirvdrum: Thanks. Bad copy & paste.

truffle/src/main/java/org/jruby/truffle/core/regexp/RegexpNodes.java
nirvdrum: I didn't know about that one. Thanks.

truffle/src/main/java/org/jruby/truffle/core/regexp/RegexpNodes.java
nirvdrum: I was trying to not hide the issue if a bad value were stored. I guess I could just add an assertion.

truffle/src/main/java/org/jruby/truffle/core/rubinius/RegexpPrimitiveNodes.java
nirvdrum: What you're saying makes sense, but I did add it for a reason. I need to look through the history in the IDE to see if it was added for a different approach (I churned through several).

truffle/src/main/java/org/jruby/truffle/core/regexp/RegexpNodes.java
eregon: Nothing guarantees `to_s` actually duplicates the original string.
`@source` is only ever used in `#string` (where it's copied) and `MatchData.pre_match_from` and even then it returns a copy since `byteslice` is called.
I think it's safe and OK to not do an explicit copy and keep it simple here.

nirvdrum: In the case of the String specialization, I continue to `dup` it. I guess strictly speaking some non-String object could cache its `to_s` output, but wouldn't that violate Ruby semantics? Even Symbols get newly allocated Strings from `to_s`.

We do need an explicit copy (or, switch to storing taint state and a rope and updating everything using the source string as necessary). The change was made to fix an issue whereby you can mutate the source string from outside the MatchData via the shared reference.

eregon: How would you mutate it? It's only accessible from the regexp.rb source file and never returned directly isn't it?

nirvdrum: That was what I thought originally and why I introduced the bug. But the following fragment shows how that can fall apart:

```
x = "abc"
x =~ /(.*)/
x.upcase!
```

Without making a defensive copy, `$~.string` will be `"ABC"`, rather than `"abc"`, which is what the match was really made against.

eregon: Oh, right, my apologies. It's a mutation the other way round. Please mention this in the comment.
Why not using the `makeSubstringNode(0,length)` in `matchCommon` then?
This seems to express the goal more clearly.

nirvdrum: The comment already says we need to make a defensive copy to prevent modifications. I take it you're looking for more details about that?

Re: the substring.... Unfortunately, it's not that easy because that substring method is really only dealing with things at the rope level and won't carry forward the taint state. We can deal with that within matchCommon, but I don't think it'll wind up much cleaner and passing a frame around gets complicated with the boundary in place.

Really, I want to overhaul all this anyway. I think when we take a pass over the regexp subsystem, a lot of this will get reworked.

eregon: Yeah, nothing like a good Ruby example to clarify it :)

Right, this `matchCommon` seems like it's asking to become a node, but let's do that later.

Issue #4019, opened 2016-07-18
gem list don't show some gems as default


etehtsea:
`jruby 9.1.2.0 (2.3.0) 2016-05-26 7357c8f Java HotSpot(TM) 64-Bit Server VM 25.77-b03 on 1.8.0_77-b03 [darwin-x86_64]`

```
$ gem list |grep default 
jar-dependencies (default: 0.3.2)
minitest (5.9.0, 5.8.4, default: 5.4.1)
net-telnet (default: 0.1.1)
power_assert (default: 0.2.3)
rake (11.2.2, 10.5.0, default: 10.4.2, 10.1.1)
rdoc (4.2.2, default: 4.2.0)
test-unit (default: 3.1.1, 2.5.5)
```

Gems that are missing from this list:
- racc
- psych
- json
- jruby-openssl

```
$ gem uninstall json
ERROR:  While executing gem ... (Gem::InstallError)
    gem "json" cannot be uninstalled because it is a default gem
```

P.S. Why rake is default on JRuby and is not on MRI?

mkristian:
rake is preinstalled with jruby but should not be a default gem.

the rest looks inconsistent as well.

etehtsea:
Okay, issue that some gems aren't shown as `default` is unrelated to JRuby itself.
Leaving it open until rubygems will be updated.

headius:
@etehtsea Help us by letting us know when there's a RubyGems release out. We can pull it into 9.1.3.0 if it's not too drastic.

etehtsea:
@headius for sure

headius:
I don't think this will be ready for 9.1.3.0. Bumping.

etehtsea:
please reopen, because fix didn't get into release.

headius:
Still waiting on a RubyGems release with the patch in question. We can patch it ourselves, but the first time anyone (or anything, like rvm) updates RubyGems, it will wipe out our change.

I have requested a backport of the patch and a 2.6.9 release of RubyGems in rubygems/rubygems#1685.

kares:
this one is resolved by now, right? I am seeing default gems being reported using the current SNAPSHOT

etehtsea:
@kares it seems that currently bundled rubygems is patched, but there is still no upstream rubygems version with this fix.

headius:
I'm going to close this. If there's a RubyGems fix we don't have, or a fix that's not in RG, that needs to happen first. But we will be on latest RG for 9.2.

PR #4083, opened 2016-08-17
Don't install non-default gems to stdlib

etehtsea:
See https://github.com/jruby/jruby/issues/4019#issuecomment-233254017


lib/pom.rb
```diff
@@ -73,8 +73,8 @@ def initialize( name, version, default_spec = true )
   #            :id => 'gem-staging' )
 
   plugin( :clean,
-          :filesets => [ { :directory => '${basedir}/ruby/gems/shared/specifications/default',
-                           :includes => [ '*' ] },
+          :filesets => [ { :directory => '${basedir}/ruby/gems',
```
etehtsea: Is it safe to drop the whole dir?

lib/pom.rb
```diff
@@ -287,7 +302,7 @@ def installer.ensure_required_ruby_version_met; end
 
     resource do
       directory '${gem.home}'
-      includes 'gems/rake-${rake.version}/bin/r*', 'gems/rdoc-${rdoc.version}/bin/r*', 'specifications/default/*.gemspec'
+      includes 'gems/rake-${rake.version}/**', 'gems/jruby-readline-1.1.dev-java/**', 'gems/rdoc-${rdoc.version}/bin/r*', 'specifications/default/*.gemspec', 'specifications/*.gemspec'
```
etehtsea: How `SNAPSHOT` version should be handled here?

.gitignore
```diff
@@ -50,24 +50,18 @@ lib/ruby/gems
...
 lib/ruby/stdlib/openssl*
 lib/ruby/stdlib/org/
 lib/ruby/stdlib/ripper.jar
 lib/ruby/stdlib/rubygems/defaults/jruby_native.rb
 lib/ruby/stdlib/racc*
 lib/ruby/stdlib/rdoc*
-lib/ruby/stdlib/test*
 lib/ruby/stdlib/hoe*
-lib/ruby/stdlib/minitest*
-lib/ruby/stdlib/power_assert*
```
mkristian: don't we want to ignore all those whether they are default gems or auxiliary gems - they both will get installed during the build ?

lib/pom.rb
mkristian: it is safe, but people expect their gems to stayed installed even after a clean build. at least some people do so :)

so I would keep it as it currently is

.gitignore
etehtsea: bundled gems will be installed to `lib/ruby/gems/shared/gems/`

mkristian: ok.

lib/pom.rb
etehtsea: It couldn't be left as it is, because bundled gems will be installed to `ruby/gems/shared/gems` and theirs specifications will be installed to to `ruby/gems/shared/specifications`.

Second problem with this approach is that all bundled gems need to be added [here](https://github.com/jruby/jruby/pull/4083/files#diff-a5692734fb1e233c63e88629e6f7bd12L290) and this is easy to forget to update this list.
And I didn't understand how to handle `SNAPSHOT` versions in this list, because `${jruby-readline.version}` is equal to `1.1.dev-SNAPSHOT`, but gem path should be `jruby-readline-1.1.dev-java`.

Also please see https://github.com/jruby/jruby/issues/4084 so more state makes build process more unpredictable.

headius:
Where do we stand on this? @mkristian does this seem ok? I don't think this is a release-specific thing, so I'm removing the 9.1.3.0 tag.

headius:
Er, I'm adding the non-release milestone. It was not set to a milestone before.

lib/pom.rb
mkristian: if we move this extra clean into 'dist' profile, then a `mvn clean install -Pdist` is doing the right thing.

there is no need to add new gems in this line https://github.com/jruby/jruby/pull/4083/files#diff-a5692734fb1e233c63e88629e6f7bd12L290 - the bin files can be probably omitted here.

mkristian:
@headius if we want `mvn clean` to keep installed gems in place, there is some more adjustments to be done. otherwise the PR makes things easier as it is now. for our releases @enebo uses new clone of the repo. 

from my side this PR is ok but it does change the `mvn clean` semantic as it really removes all installed gems.

lib/pom.rb
etehtsea: If I don't add gems dir will look like:

```
/tmp/0  ⚡ ls jruby-9.1.3.0-SNAPSHOT/lib/ruby/gems/shared/gems/**
jruby-9.1.3.0-SNAPSHOT/lib/ruby/gems/shared/gems/rake-10.4.2:
bin/

jruby-9.1.3.0-SNAPSHOT/lib/ruby/gems/shared/gems/rdoc-4.2.0:
bin/
```

etehtsea:
@mkristian updated and left clean phase intact.

kares:
+1 here ... on the related note - would love a `gem clean`-er as its painful to switch branches on times ;(
e.g. in a profile `mvn clean -Pdefault-gems` ... or is it there already? 
can't seem to find anything similar in **lib/pom.rb**

mkristian:
still need to look at travis as there are a few failing tests.

@kares is `mvn clean` not enough as it clean `/lib/ruby/gems/specifications/default/*` ?

kares:
@mkristian no as I want to be able to undo the stdlib installed files but its a specific requirement thus no blocker here. just mentioning @etehtsea's original (reverted) clean logic might be useful (in corner cases).

mkristian:
@kares I understood that this is an extra feature. but deleting installed gems is probably not the solution, as the installed files in lib/ruby/stdlib remains. the problem here is that after switching the branch the pom.rb has no more info about the files installed there. such a feature needs to go via `git ls-files lib/ruby/stdlib` to find the files which should remain.

kares:
this has diverged quite esp. now with default and pre-installed gems in 9.2 ... so this would need a rework.
`gem list` seems to be doing fine and I am not sure which commit still makes sense, thus closing as is.
if anyone feels parts are worth a resurrect maybe its best to start from master with a clean slate. ..

PR #4086, opened 2016-08-18
Socket branch partial merge

etehtsea:
I propose to merge commits from ruby-2.3+socket branch that don't depend from each other and still apply cleanly.


etehtsea:
There is better way how to merge this in than "get everything that still apply cleanly".

PR #4095, opened 2016-08-20
Add a simple deadlock detector for require locks.

headius:
This detector works only with two threads, using the following
heuristic:

1. If the lock is already acquired, obtain its owner thread.
2. Check if the owner thread is waiting on a lock.
3. If the lock the owner is waiting on is held by the current
   thread, raise an error.
4. If there's no owner and no deadlock, try to lock the thread for
   500ms + rand(100)ms.
5. If the file has been locked, return success.
6. Otherwise, start over at 1.

We may want a more advanced detector that can handle arbitrarily
many threads, but it would be much heavier and require scanning
all threads in the system, including those unrelated to the
current program and those not actually trying to require files.

See #3341.

### DIFF
--- a/core/src/main/java/org/jruby/runtime/load/LoadService.java
+++ b/core/src/main/java/org/jruby/runtime/load/LoadService.java
@@ -36,6 +36,9 @@ package org.jruby.runtime.load;
 import java.io.File;
 import java.io.FileNotFoundException;
 import java.io.IOException;
+import java.lang.management.ManagementFactory;
+import java.lang.management.ThreadInfo;
+import java.lang.management.ThreadMXBean;
 import java.net.MalformedURLException;
 import java.net.URISyntaxException;
 import java.net.URI;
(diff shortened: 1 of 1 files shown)


headius:
I'm opening this PR to discuss the implementation and seek input.

A few known issues:

* Hard dependency on management extensions. These are not available on all platforms and I have no fallbacks (nor graceful handling when those extensions are unavailable).
* Potentially heavy to acquire this info. Does it matter if we only do it for contended threads?
* Only works for two threads with mutually-exclusive locking. There are algorithms for detecting deadlocks between many threads, but I believe they're NP. I have not researched yet.

headius:
I made a modification to mitigate known issue 2 above: it now will attempt to lock, only doing deadlock detection if that attempt fails.

kares:
should safe fail when mx (`java.lang.management`) not available e.g. on Android

headius:
Punting to 9.1.4.0. Too risky to stitch in at the last moment for 9.1.3.0.

headius:
Needs rebase and more thought. Bumping and rebasing.

headius:
De-targeting. This may be useful some day but there's work required and many open questions.

Issue #4070, opened 2016-08-12
Non-ASCII Symbol gives ArgumentError when calling inspect on the symbol


donv:
### Environment

$ ruby -v
jruby 9.1.3.0-SNAPSHOT (2.3.0) 2016-08-12 93bd82f Java HotSpot(TM) 64-Bit Server VM 25.92-b14 on 1.8.0_92-b14 +jit [darwin-x86_64]
$ uname -a
Darwin macbeth-3.local 14.5.0 Darwin Kernel Version 14.5.0: Thu Jun 16 19:58:21 PDT 2016; root:xnu-2782.50.4~1/RELEASE_X86_64 x86_64
### Expected Behavior

$ ruby -v
ruby 2.3.1p112 (2016-04-26 revision 54768) [x86_64-darwin14]
$ irb
irb(main):001:0> :Renè
=> :Renè
### Actual Behavior

$ ruby -v
jruby 9.1.3.0-SNAPSHOT (2.3.0) 2016-08-12 93bd82f Java HotSpot(TM) 64-Bit Server VM 25.92-b14 on 1.8.0_92-b14 +jit [darwin-x86_64]
$ irb
irb(main):002:0> :René
ArgumentError: invalid byte sequence in UTF-8
    from org/jruby/RubySymbol.java:274:in `inspect'
    from org/jruby/RubySymbol.java:259:in`inspect'
    from org/jruby/RubyKernel.java:1295:in `loop'
    from org/jruby/RubyKernel.java:1114:in`catch'
    from org/jruby/RubyKernel.java:1114:in `catch'
    from /Users/uwe/.rubies/jruby-9.1.3.0-snapshot/bin/irb:13:in`<main>'

enebo:
This appears to be specific to irb itself.  If I make a symbol in a file or via -e then :Renè parses fine.

phluid61:
It's an error in `#inspect`

```
$ bin/jruby -v -e 'p :Renè'
jruby 9.1.3.0-SNAPSHOT (2.3.0) 2016-07-07 6600d9f OpenJDK 64-Bit Server VM 25.91-b14 on 1.8.0_91-8u91-b14-3ubuntu1~16.04.1-b14 +jit [linux-x86_64]
ArgumentError: invalid byte sequence in UTF-8
  inspect at org/jruby/RubySymbol.java:274
  inspect at org/jruby/RubySymbol.java:259
        p at org/jruby/RubyKernel.java:476
   <main> at -e:1
```

[...]

Issue #3880, opened 2016-05-12
define_method using symbols string syntax works incorrectly


catfishuang:
Environment

jruby 9.1.0.0 (2.2.3) 2016-01-26 7bee00d Java HotSpot(TM) 64-Bit Server VM 24.79-b02 on 1.7.0_79-b15 +jit [Windows 7-amd64]

Expected Behavior

```ruby
def 牛逼
puts '111'
end

牛逼 # It's OK

@name = "中文"
send :define_method,:"#{@name}" do
puts "中文test"
end

中文 # Error occur but it run well in cruby
```

Actual Behavior

NameError: undefined local variable or method `??' for main:Object

it seems that  the issue has not been resolved....... i checked it just now

headius:
Trivial repro:

```ruby
define_method(:"中文")do
  puts "中文test"
end
中文
```

headius:
Another example that shows we're encoding identifiers and symbols differently:

```
class X
  def 你好; end
end
nihao = X.instance_methods(false)[0]
zhongwen = :"中文"

p nihao # => :??
p nihao.encoding # => #<Encoding:ASCII-8BIT>
p zhongwen # => :中文
p zhongwen.encoding # => #<Encoding:UTF-8>
```

This isn't really a problem with `define_method`, it's a known problem with how we encode method names versus symbols. I know we have other issues open for this...will try to locate.

headius:
Comparative output from MRI:

```
$ ruby23 blah.rb
:你好
#<Encoding:UTF-8>
:中文
#<Encoding:UTF-8>
```

headius:
Related to #1340, #3719, #3697 at least.

enebo:
I corrected the title to be more descriptive of what is broken.  Note that this works in cases with no quote chars in the symbol:

```ruby
# coding: utf-8
define_method(:中文)do
  puts "中文test"
end
中文
```

This is very specific to encoding with the newer-ish quote syntax (2.1?) of symbols specifically.  I also wonder whether the reduced case and the interpolated version originally reported are the same problem but they are both in the parser even if so.  As such, I think the other issues linked are problems but fixing this issue will fix none of the other linked issues (although #3719 might be the same kind of fix).

headius:
@enebo Very strange! The parser is definitely handling the quoted case differently:

```
[] ~/projects/jruby $ ast -e "puts :'中文'"
AST:
RootNode 0
  FCallNode:puts 0
    ArrayNode 0
[...]

Issue #3719, opened 2016-03-09
encoding of utf8 symbols (eg. :λ) 


elia:
The failing Opal build: https://travis-ci.org/opal/opal/jobs/114673459#L372
[The failing specs][specs] have been added [recently][commit], also maybe it's just a matter of setting the correct locale env vars as in https://github.com/jruby/jruby/issues/2204

[specs]: https://github.com/opal/opal/blob/0ac003566b7fc53845d7395b47d84f54cfc6421c/spec/lib/parser/variables_spec.rb#L105-L109
[commit]: https://github.com/opal/opal/commit/5716c3456feabe2280091b8f623f147ff109d789

### Environment

```bash
$ java -Xmx32m -version
java version "1.7.0_76"
Java(TM) SE Runtime Environment (build 1.7.0_76-b13)
Java HotSpot(TM) 64-Bit Server VM (build 24.76-b04, mixed mode)
$ javac -J-Xmx32m -version
javac 1.7.0_76
$ ruby --version
jruby: warning: unknown property jruby.cext.enabled
jruby 9.1.0.0-SNAPSHOT (2.3.0) 2016-03-09 5b08612 Java HotSpot(TM) 64-Bit Server VM 24.76-b04 on 1.7.0_76-b13 +jit [linux-amd64]
$ rvm --version
rvm 1.26.11 (latest) by Wayne E. Seguin <wayneeseguin@gmail.com>, Michal Papis <mpapis@gmail.com> [https://rvm.io/]
$ bundle --version
jruby: warning: unknown property jruby.cext.enabled
Bundler version 1.6.9
$ gem --version
jruby: warning: unknown property jruby.cext.enabled
2.4.8
```

### Expected/Actual Behavior

```rb
  1) Opal::Parser parses unicode local variables
     Failure/Error: parsed("λ = 1").should == [:lasgn, :λ, [:int, 1]]
       expected: [:lasgn, :?, [:int, 1]]
            got: (:lasgn, :λ, (:int, 1)) (using ==)
     # ./spec/lib/parser/variables_spec.rb:106:in `block in (root)'
  2) Opal::Parser parses unicode constants
     Failure/Error: parsed("FOOλ = 1").should == [:cdecl, :FOOλ, [:int, 1]]
       expected: [:cdecl, :FOO?, [:int, 1]]
            got: (:cdecl, :FOOλ, (:int, 1)) (using ==)
     # ./spec/lib/parser/variables_spec.rb:102:in `block in (root)'
```

[...]

Issue #2896, opened 2015-04-30
Symbol#inspect differs from MRI Ruby


k77ch7:
The following symbol_test.rb differs from MRI Ruby. The different result occurs in the case of UTF8 string symbol.

symbol_test.rb
```
# encoding: utf-8

p :"Ãa1".inspect
p :"a1".inspect
p :"1".inspect
```

MRI Ruby 1.9.3, 2.0.0, 2.1.4 and 2.2,1
```
$ ruby ~/development/symbol_test.rb
":Ãa1"
":a1"
":\"1\""
```
JRuby master and 1_7 branch
```
$ jruby ~/development/symbol_test.rb
":\"Ãa1\""
":a1"
":\"1\""
```

My env
```
jruby 9.0.0.0-SNAPSHOT (2.2.2) 2015-04-30 1cade78 Java HotSpot(TM) 64-Bit Server VM 25.5-b02 on 1.8.0_05-b13 +jit [darwin-x86_64]

jruby 1.7.20-SNAPSHOT (1.9.3p551) 2015-04-30 5fd659d on Java HotSpot(TM) 64-Bit Server VM 1.8.0_05-b13 +jit [darwin-x86_64]

```

PR #4096, opened 2016-08-21
Extra tests for symbol encoding Re: #4070 + #3719 and possibly #3880

brocktimus:
Saw this tweet https://twitter.com/headius/status/766771177301774337

So I decided to write a few simple failing tests isolating issues issues discussed in #4070 and #3719. This could also overlap with #3880 so I figured I would mention that too.

Very happy to rename specs since I'm bad at naming them or committing these into ruby/spec if that is appropriate. I just put them here because of similar specs relating to #2896.


brocktimus:
More than happy to close this in lieu of #4099

PR #4100, opened 2016-08-22
Change script file restriction

jellymann:
Fix for issue exposed by #3901.

Extension of #3902 for 9k, with tests and an `isDirectory` check (`canRead` returns true for directories, but then throws when calling `inputStream`)

I'm not sure about testing `CON` on Windows. I have only tested on macOS so far.

These tests (and isDirectory change) would also need to backported to the 1.7 branch

CC @mkristian and @headius


mkristian:
+1 for the commit

@jellymann could you remove all those maven-metadata-local.xml files. I have seen them creeping in before but do not have an idea how this happens.

jellymann:
@mkristian Sorry about that. I saw them get created and meant to make sure not to commit them, but I must have forgotten. They seem to have been created the first time I ran `mvn` after cloning

mkristian:
yes, I just saw them too on jruby-readline after the first run of maven. at
least some clue to start digging :)

mkristian:
@headius for me this is good to merge

Issue #3869, opened 2016-05-10
Unable to use UnboundMethod referencing super with define_singleton_method


saturnflyer:
I'm unable to transplant a method defined in a module to define a method on another object

### Environment

jruby 9.1.0.0 (2.3.0) 2016-05-02 a633c63 Java HotSpot(TM) 64-Bit Server VM 25.77-b03 on 1.8.0_77-b03 +jit [darwin-x86_64]

### Expected Behavior

I expect to be able to define methods on an object using `define_singleton_method` and instance methods from a module

Example code:

https://github.com/saturnflyer/behavioral/blob/00dae0f871ac921b77983f1535276c375f0baaa8/lib/behavioral.rb

Or using this sample in an irb session should work:

```ruby
class P
  def name
    'name'
  end
end
module Mod
  def name
    "mod #{super}"
  end
end
p = P.new
p.define_singleton_method(:name, Mod.instance_method(:name))
p.send :name #=> "mod name"
```

### Actual Behavior

NullPointerException

I'm still trying to track down the exact cause of this and will update this issue if I find the concrete trigger but for now I have failing tests at: https://travis-ci.org/saturnflyer/behavioral/jobs/128500121

I believe it has to do with using `super` inside an UnboundMethod used to define the singleton_class instance method.

If you remove the `super` call from the example code above, it works fine.

Example stack trace from the link to travis-ci above:
```
  1) Error:
Behavioral#test_0003_allows adding multiple behavior modules at once:
Java::JavaLang::NullPointerException: 
    org.jruby.ir.runtime.IRRuntimeHelpers.unresolvedSuper(IRRuntimeHelpers.java:991)
    org.jruby.ir.instructions.UnresolvedSuperInstr.interpret(UnresolvedSuperInstr.java:83)
    org.jruby.ir.interpreter.InterpreterEngine.processCall(InterpreterEngine.java:348)
    org.jruby.ir.interpreter.StartupInterpreterEngine.interpret(StartupInterpreterEngine.java:77)
    org.jruby.ir.interpreter.InterpreterEngine.interpret(InterpreterEngine.java:80)
    org.jruby.internal.runtime.methods.MixedModeIRMethod.INTERPRET_METHOD(MixedModeIRMethod.java:144)
```

headius:
(edited your stack trace for brevity)

Heh, well you definitely shouldn't get a NullPointerException!

[...]

Issue #3708, opened 2016-03-02
Unresolved super with module_function


Exilor:
Using 9.0.5.0 on Windows:

```ruby
module Foo
  module_function

  def load(file)
    super
  end
end

Foo.load("a file")
```
Error:
```
Unhandled Java exception: java.lang.NullPointerException
java.lang.NullPointerException: null
           unresolvedSuper at org/jruby/ir/runtime/IRRuntimeHelpers.java:958
  unresolvedSuperSplatArgs at org/jruby/ir/runtime/IRRuntimeHelpers.java:947
                      load at l2r.rb:645
                      call at org/jruby/internal/runtime/methods/CompiledIRMethod.java:131
                      call at org/jruby/internal/runtime/methods/DynamicMethod.java:197
                      call at org/jruby/internal/runtime/methods/WrapperMethod.java:59
              cacheAndCall at org/jruby/runtime/callsite/CachingCallSite.java:313
                      call at org/jruby/runtime/callsite/CachingCallSite.java:163
                     <top> at l2r.rb:649
                      load at org/jruby/ir/Compiler.java:111
                 runScript at org/jruby/Ruby.java:817
                 runScript at org/jruby/Ruby.java:809
               runNormally at org/jruby/Ruby.java:747
               runFromMain at org/jruby/Ruby.java:569
             doRunFromMain at org/jruby/Main.java:415
               internalRun at org/jruby/Main.java:310
                       run at org/jruby/Main.java:239
                      main at org/jruby/Main.java:201
```
Changing ```module_function``` to ```extend self``` correctly calls ```Kernel.load``` (and raises LoadError, as expected).

headius:
Still happens in JRuby 9.1.3.0 (master).

headius:
I believe I have a fix. Testing locally.

PR #4102, opened 2016-08-22
No more wrapper

headius:
These commits remove all remaining uses of WrapperMethod, which fixes issues like #3708 and #3869.

See the commits for the longer story, but basically WrapperMethod does not pass through the "new" impl class to the contained method, which results in it supering up the wrong hierarchy.

I have a few concerns, which is why I opened a PR for review:

* Methods that were wrapped before will now be their own true instances of the original method.
* Because we now attempt to dup rather than wrap, I had to modify duping of IR-based methods to always use clone. This makes them have the same serial number, so that transplanted methods (e.g. those in #3869) will still be `==` and `eql?`. The wrapper previously made them appear to be two methods with the same serial, so this may be ok...but if we depend on the serial number being unique after a method dup, that will no longer work (I don't know that we do).
* This is a relatively deep change right before 9.1.3.0, but it fixes known bugs with super that have dogged us for a long time.


headius:
@enebo Review please!

enebo:
@headius sorry we went over this yesterday and there are no issues I can readily see.  This seems much nicer fwiw too.

PR #4106, opened 2016-08-23
MRI does same logic for Yielder#yield and #<<. Fixes #3814 cases.

headius:
See #3814. This appears to match MRI and fixes the cases added there.


headius:
Invalid...simply reverts #2458. See #3814 for progress toward a workable solution.

PR #4107, opened 2016-08-24
workaround compilation problems with unsafe-mock.jar

### LINKED ISSUE #4103: JRuby fails in building from source when using openjdk-7-jdk trusty package
### Environment

Provide at least:
- Trying to build jruby `9.1.2.0` from source
- `uname -a`:
  `Linux d64e610fc91e 4.4.16-boot2docker #1 SMP Fri Jul 29 00:13:24 UTC 2016 x86_64 x86_64 x86_64 GNU/Linux`
### Expected Behavior
- JRuby builds from source using version `7u111-2.6.7-0ubuntu0.14.04.3` of the ubuntu trusty openjdk-7-jdk package.
### Actual Behavior
- JRuby fails in building from source using version `7u111-2.6.7-0ubuntu0.14.04.3` of the ubuntu trusty openjdk-7-jdk package.

Reproduction:

``` bash
git clone git@github.com:cloudfoundry/binary-builder.git
cd binary-builder
docker run -it -v $(pwd):/binary-builder cloudfoundry/cflinuxfs2 bash
cd /binary-builder
./bin/binary-builder --name=jruby --version=9.1.2.0_ruby-2.3.0 --sha256=dfd44f532105725a4d4a2378a3b6a06899bae660be21538a5b36b41e4dd045e6
```

The full error log after switching maven to try and build jruby with the `-X` flag enabled is here: https://gist.github.com/RochesterinNYC/fac1f36bb98ded1f5a135f10b2c8741c

Binary builder will try to install openjdk via apt-get, build and compile ant and maven, and then use maven to try to build jruby. It runs the following command to try to build jruby: `mvn -Djruby.default.ruby.version=2.3.0`

Side Note:
It might be worth noting that for our situation, this build process was working when we were on openjdk-7-jdk `7u111-2.6.7-0ubuntu0.14.04.2`. When openjdk-7-jdk `7u111-2.6.7-0ubuntu0.14.04.3` was [released on August 11th](http://changelogs.ubuntu.com/changelogs/pool/main/o/openjdk-7/openjdk-7_7u111-2.6.7-0ubuntu0.14.04.3/changelog), our build process automatically started using this (because we use `apt-get install`) and the build started failing.

mkristian:
fixes #4103

### DIFF
--- a/core/pom.rb
+++ b/core/pom.rb
@@ -190,6 +190,7 @@ project 'JRuby Core' do
                                        '-J-Duser.language=en',
                                        '-J-Dfile.encoding=UTF-8',
                                        '-J-Xbootclasspath/p:${unsafe.jar}',
+                                       '-Xbootclasspath/p:${unsafe.jar}',
                                        '-J-Xmx${jruby.compile.memory}' ] )
     execute_goals( 'compile',
                    :id => 'populators',
@@ -198,6 +199,7 @@ project 'JRuby Core' do
                                        '-J-Duser.language=en',
                                        '-J-Dfile.encoding=UTF-8',
                                        '-J-Xbootclasspath/p:${unsafe.jar}',
+                                       '-Xbootclasspath/p:${unsafe.jar}',
                                        '-J-Xmx${jruby.compile.memory}' ],
                    'includes' => [ 'org/jruby/gen/**/*.java' ] )
     execute_goals( 'compile',
--- a/core/pom.xml
+++ b/core/pom.xml
@@ -491,6 +491,7 @@ DO NOT MODIFIY - GENERATED CODE
                 <compilerArg>-J-Duser.language=en</compilerArg>
                 <compilerArg>-J-Dfile.encoding=UTF-8</compilerArg>
                 <compilerArg>-J-Xbootclasspath/p:${unsafe.jar}</compilerArg>
+                <compilerArg>-Xbootclasspath/p:${unsafe.jar}</compilerArg>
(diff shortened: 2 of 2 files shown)


headius:
Gross. That this actually works is even grosser :-)

I'll run some sanity checks here.

headius:
Oracle Java 7 and 8 both seem fine with the change. I'm checking J9 now.

headius:
As I feared, this patch does not appear to work on J9. I'll try to get the exact error.

headius:
Here's my daily rant about Maven refusing to show the actual error when compilation fails.

`(╯°□°）╯︵ ┻━┻`

Now that that's over...

I captured the javac command line, and it worked just fine in J9 at the command line. So perhaps J9 issue was something else? I'D SURE LIKE TO SEE WHAT THE ERROR WAS, MAVEN.

headius:
Well I'm stumped. I can't get Maven to tell me why J9 javac is failing, and it works ok at the command line. I suppose we can go with this, since the bare -X flag does appear to work on J9 Java 8 javac.

mkristian:
we can also enforce a particular jdk to build, like jdk version bigger then 8, or make the switches depending on the jdk version via profiles for those versions.

headius:
Ok, I've confirmed that J9 Java 8 on Ubuntu can't build JRuby either way, so we'll just merge this.

RochesterinNYC:
Awesome! so once a new release of jruby comes out that includes this change, we'll be able to build using that latest version of the openjdk-7-jdk ubuntu package.

PR #4109, opened 2016-08-24
[Truffle] Polyglot eval io exception

jtulach:
Truffle's `PolyglotEngine.eval` no longer throws `IOException` after https://github.com/graalvm/truffle/commit/5e141bd0895239307edb5a4c64eaccb7d37409ef

This change adjust JRuby code to be compilable with both versions. The old one that throws `IOException` and the new one which doesn't.

### DIFF
--- a/truffle/src/main/java/org/jruby/truffle/JRubyTruffleImpl.java
+++ b/truffle/src/main/java/org/jruby/truffle/JRubyTruffleImpl.java
@@ -55,6 +55,8 @@ public class JRubyTruffleImpl implements JRubyTruffleInterface {
 
         try {
             return engine.eval(loadSource("Truffle::Boot.run_jruby_root", "run_jruby_root")).get();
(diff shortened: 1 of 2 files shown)


jtulach:
Mergeable? Or is there a problem that I am supposed to fix?

eregon:
@jtulach I just did the change yesterday to use latest truffle: https://github.com/jruby/jruby/commit/3f5e7252fc5230257f1aa02d52ba98e1d4be4b5d.
I think that commit is much cleaner and also more correct.
I'll merge it when we update Truffle.
Thank you for the change, but I prefer not merging such a hack.

jtulach:
C'mon it is not a hack: it is a transition state. Anyway I am trying your
latest commits, maybe they'll be sufficient OK.

2016-08-25 11:49 GMT+02:00 Benoit Daloze <notifications@github.com>:

PR #4110, opened 2016-08-24
[Truffle] Add ThreadGroup layout and nodes

bjfish:
- Layout needs review, Atomic/Volatile maybe needed


eregon:
Just wondering, why does this need to be done in Java and not just Ruby?
Probably we need a group field in ThreadLayout but what else?

bjfish:
@eregon I can't think of any issues of doing it in ruby, I'll try it in ruby

bjfish:
@eregon I've updated now to use Ruby

truffle/src/main/java/org/jruby/truffle/core/thread/ThreadNodes.java
```diff
@@ -425,7 +440,8 @@ public DynamicObject allocate(
                     new AtomicReference<>(null),
                     new AtomicReference<>(null),
                     new AtomicBoolean(false),
-                    new AtomicInteger(0));
+                    new AtomicInteger(0),
+                    currentGroup);
```
bjfish: I tried to set the group in `Thread#initialize` method but I don't think `initialize` method was being called. Is allocate okay to set this for now?

eregon: Yes that sounds fine.

truffle/src/main/java/org/jruby/truffle/core/thread/ThreadManager.java
```diff
@@ -69,7 +69,8 @@ public static DynamicObject createRubyThread(RubyContext context) {
                 null,
                 null,
                 new AtomicBoolean(false),
-                0);
+                0,
+                null);
```
eregon: If you initialize this to `nil()` then I think there is no need for `null` handling in other places.

truffle/src/main/ruby/core/thread.rb
```diff
@@ -352,15 +352,38 @@ def freeze
...
+    return nil unless from_tg
+    raise ThreadError, "can't move from the frozen thread group" if from_tg.frozen?
+    raise ThreadError, "can't move from the enclosed thread group" if from_tg.enclosed?
+
+    Truffle.invoke_primitive :thread_set_group, thread, self
+    self
   end
 
+  def list
+    Thread.list.find_all { |th| th.group == self }
```
eregon: I think you could keep like the original implementation with a Ruby-level `@list` array. Or is that problematic?

eregon:
Well done, this looks much cleaner and smaller now :)

truffle/src/main/ruby/core/thread.rb
bjfish: @eregon I made this a calculated field only to match MRI. It might have some advantages: harder to change a calculated field versus a instance var, etc.

eregon: It's much less efficient though.
Nitpick: could you use the common alias of `find_all`, that is `select`?

bjfish: I've updated to select.

truffle/src/main/java/org/jruby/truffle/core/thread/ThreadManager.java
bjfish: I've updated to use `nil` here and remove the `null` handling.

eregon:
Let's get this merged once the CI passes!

truffle/src/main/ruby/core/thread.rb
```diff
@@ -352,15 +352,38 @@ def freeze
...
   end
 
+  def list
+    Thread.list.select { |th| th.group == self }
+  end
+
+  Default = ThreadGroup.new
+
 end
+Truffle.invoke_primitive :thread_set_group, Thread.current, ThreadGroup::Default
```
eregon: This is probably causing the CI failure.
@chrisseaton So should thread_set_group not have `UnsafeGroup.THREADS` since it's a primitive only called internally or is there another solution?

chrisseaton: Yes, no need for `UnsafeGroup.THREADS` on `thread_set_group` since it doesn't really do anything unsafe (IO, syscalls, creating new threads, etc).

eregon:
Build is green, merged! Thanks for going through all my comments!

Issue #4024, opened 2016-07-21
StackOverflowError in exception handling


erikogan:
### Environment

Similar errors in: 
- jruby 9.1.3.0-SNAPSHOT (2.3.0) 2016-07-20 8c97591 Java HotSpot(TM) 64-Bit Server VM 25.60-b23 on 1.8.0_60-b27 +jit [darwin-x86_64]
- jruby 9.1.2.0 (2.3.0) 2016-05-26 7357c8f Java HotSpot(TM) 64-Bit Server VM 25.60-b23 on 1.8.0_60-b27 +jit [darwin-x86_64]

Gems involved: 
- Rails 5.0.0
- RSpec 3.5.1
- aws-sdk 2.3.15-2.4.0 (2.3.14 does not throw the underlying exception)
### Expected Behavior

I expect a `TypeError: no implicit conversion of nil into String` to be raised due to a bug in aws-sdk.
### Actual Behavior

I am seeing a StackOverflowError exception when an RSpec integration test receives a `TypeError: no implicit conversion of nil into String`. 

Attempting to reproduce that error in isolation fails (the correct exception is raised)

Details here: https://github.com/aws/aws-sdk-ruby/issues/1232#issuecomment-234146274

When run with `JRUBY_OPTS=-X-C`, the following (truncated) stack trace occurs:

```
     Failure/Error: Unable to find org.jruby.ir.interpreter.InterpreterEngine.processBookKeepingOp(InterpreterEngine.java to read failed line
     Java::JavaLang::StackOverflowError:
     # org.jruby.ir.interpreter.InterpreterEngine.processBookKeepingOp(InterpreterEngine.java:368)
     # org.jruby.ir.interpreter.InterpreterEngine.interpret(InterpreterEngine.java:196)
     # org.jruby.ir.interpreter.Interpreter.INTERPRET_BLOCK(Interpreter.java:132)
     # org.jruby.runtime.InterpretedIRBlockBody.callDirect(InterpretedIRBlockBody.java:102)
     # org.jruby.runtime.IRBlockBody.call(IRBlockBody.java:64)
     # org.jruby.runtime.Block.call(Block.java:126)
     # org.jruby.RubyProc.call(RubyProc.java:342)
     # org.jruby.internal.runtime.methods.ProcMethod.call(ProcMethod.java:63)
     # org.jruby.internal.runtime.methods.DynamicMethod.call(DynamicMethod.java:193)
[...]

PR #4117, opened 2016-08-26
Show that ArgumentError from encoding error is recursive

nightscape:
In general Java exceptions should not have themselves as cause and JRuby in most cases handles this correctly.
We just ran into a problem where an `ArgumentError` raised due to an encoding problem does have itself as cause which leads to a [problem in an ActiveSupport method](https://github.com/rails/rails/blob/master/activesupport/lib/active_support/rescuable.rb#L124).
The ActiveSupport method is [a little dangerous/questionable by itself](https://github.com/rails/rails/pull/26246), but the recursive cause should be prevented in JRuby nonetheless and there is also a test that checks this for `RuntimeError`s.
We're not sure if this is the most general reproduction, so feel free to ignore this PR if you find a more general test.

### DIFF
--- a/test/mri/ruby/test_exception.rb
+++ b/test/mri/ruby/test_exception.rb
@@ -625,6 +625,18 @@ end.join
     assert_not_same(e, e.cause, "#{msg}: should not be recursive")
   end
 
+  def test_cause_reraised_for_encoding_error
+    msg = "ArgumentError from encoding problems"
+    e = assert_raise(ArgumentError) {
+      begin
+        "hi \255".split
+      rescue => e
+        raise e
+      end
+    }
+    assert_not_same(e, e.cause, "#{msg}: should not be recursive")
+  end
+
   def test_raise_with_cause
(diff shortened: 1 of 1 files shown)


nightscape:
Issue https://github.com/jruby/jruby/issues/4024 might be related to this.

headius:
I think this is simply us being too eager to use the in-flight exception when re-raising, not realizing we're stuffing the same exception into itself again. Looking.

headius:
This only appears to affect re-raising using the `raise ex` form, and I have a fix coming.

```
$ jruby -e 'begin; "hi \255".split; rescue => e; p e.cause; raise e; end rescue p $!.cause'
nil
#<ArgumentError: invalid byte sequence in UTF-8>
```

headius:
I've fixed the issue, but the test would be better going into ruby/spec (spec/ruby/core/exception or spec/ruby/core/kernel/raise_spec.rb).

PR #4120, opened 2016-08-27
[Truffle] Add name to Thread layout

bjfish:


truffle/src/main/java/org/jruby/truffle/core/thread/ThreadLayout.java
```diff
@@ -46,7 +46,8 @@ DynamicObject createThread(
             @Nullable @Volatile Object value,
             AtomicBoolean wakeUp,
             @Volatile int priority,
-            @Nullable DynamicObject threadGroup);
+            @Nullable DynamicObject threadGroup,
```
eregon: BTW, threadGroup no longer needs to be @Nullable, right?

truffle/src/main/ruby/core/thread.rb
```diff
@@ -193,6 +193,21 @@ def priority=(val)
     Kernel.raise ThreadError, "Thread#priority= primitive failed"
   end
 
+  def name
+    Truffle.primitive :thread_get_name
+    Kernel.raise ThreadError, "Thread#priority primitive failed"
```
eregon: Typo

truffle/src/main/java/org/jruby/truffle/core/thread/ThreadLayout.java
bjfish: Yes, I've removed the `@Nullable`.

truffle/src/main/ruby/core/thread.rb
bjfish: Typo fixed.

PR #4123, opened 2016-08-31
Fix IO#syswrite. Fixes #3799

### LINKED ISSUE #3799: Unix sockets raise exception on long writes
### Environment
- `jruby 9.0.5.0 (2.2.3) 2016-01-26 7bee00d Java HotSpot(TM) 64-Bit Server VM 25.72-b15 on 1.8.0_72-b15 +jit [darwin-x86_64]`
- `Darwin veritas.local 15.4.0 Darwin Kernel Version 15.4.0: Fri Feb 26 22:08:05 PST 2016; root:xnu-3248.40.184~3/RELEASE_X86_64 x86_64`
- Rails app running puma: `https://github.com/cdonadeo/puma_big_response_test`
### Expected Behavior

Long strings passed to `UNIXSocket#write` writes the data and returns the number of bytes written.

### Actual Behavior

`#write` writes a subset of the data and raises an exception: `#<SystemCallError: Unknown error - >}`. The amount of data varies but I've seen 8192 show up not-randomly, almost like there was a 8k buffer somewhere.

The git repo linked to above shows the behavior. The exception in question is raise inside `#fast_write`. Changing the code to use `#write` instead of `#syswrite` does not change the behavior.

etehtsea:
As a result of debugging I found out that `SystemCallError` that happens during `syswrite` is actually `"Resource temporarily unavailable"` that corresponds to `EAGAIN`. So first idea of fix was to put `select` somewhere. Related fix for `IO#sysread` gave me hope that I'm moving in right direction. 

As a result I did similar fix as in 857b76f461acb5cbb67d582b864145e617ccaee8

### DIFF
--- a/core/pom.rb
+++ b/core/pom.rb
@@ -46,7 +46,7 @@ project 'JRuby Core' do
 
   # exclude jnr-ffi to avoid problems with shading and relocation of the asm packages
   jar 'com.github.jnr:jnr-netdb:1.1.6', :exclusions => ['com.github.jnr:jnr-ffi']
-  jar 'com.github.jnr:jnr-enxio:0.13', :exclusions => ['com.github.jnr:jnr-ffi']
+  jar 'com.github.jnr:jnr-enxio:0.14-SNAPSHOT', :exclusions => ['com.github.jnr:jnr-ffi']
   jar 'com.github.jnr:jnr-x86asm:1.0.2', :exclusions => ['com.github.jnr:jnr-ffi']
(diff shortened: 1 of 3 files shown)


headius:
@etehtsea I do not see the equivalent logic in MRI's implementation, which makes me reluctant to include this. I think the actual problem may be that jnr-unixsocket is not properly doing a blocking write, or the UNIX socket is getting set non-blocking for some reason.

I think we should try to investigate from the bottom up and figure out why jnr-unixsocket's write is not blocking and handling EAGAIN the same way as the write call MRI makes.

headius:
Thank you!

PR #4126, opened 2016-09-01
Implement SignalException class

### LINKED ISSUE #3954: SignalException#signo (and signm) not defined.
`jruby 9.1.1.0 (2.3.0) 2016-05-19 fe84e89 Java HotSpot(TM) 64-Bit Server VM 25.45-b02 on 1.8.0_45-b14 +jit [darwin-x86_64]`

It appears that SignalException#signo and #signm are not defined:

```
se = SignalException.new(6)
p se.signo # Expect 6
p se.signm # Expect "SIGABRT"
```

JRuby output:

```
NoMethodError: undefined method `signo' for #<SignalException: 6>
NoMethodError: undefined method `signm' for #<SignalException: 6>
```

Courtesy of berger_spec.

etehtsea:
Fixes #3954

### DIFF
--- a/core/src/main/java/org/jruby/RubySignalException.java
+++ b/core/src/main/java/org/jruby/RubySignalException.java
@@ -1,0 +1,128 @@
+/***** BEGIN LICENSE BLOCK *****
+ * Version: EPL 1.0/GPL 2.0/LGPL 2.1
+ *
+ * The contents of this file are subject to the Eclipse Public
+ * License Version 1.0 (the "License"); you may not use this file
+ * except in compliance with the License. You may obtain a copy of
+ * the License at http://www.eclipse.org/legal/epl-v10.html
+ *
+ * Software distributed under the License is distributed on an "AS
+ * IS" basis, WITHOUT WARRANTY OF ANY KIND, either express or
+ * implied. See the License for the specific language governing
+ * rights and limitations under the License.
+ *
+ * Alternatively, the contents of this file may be used under the terms of
+ * either of the GNU General Public License Version 2 or later (the "GPL"),
+ * or the GNU Lesser General Public License Version 2.1 or later (the "LGPL"),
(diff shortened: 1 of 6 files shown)


core/src/main/java/org/jruby/RubySignalException.java
```diff
@@ -0,0 +1,128 @@
...
+        signalExceptionClass.defineAnnotatedMethods(RubySignalException.class);
+
+        return signalExceptionClass;
+    }
+
+    @JRubyMethod(optional = 2, visibility = PRIVATE)
+    public IRubyObject initialize(IRubyObject[] args, Block block) {
+        final Ruby runtime = getRuntime();
+        int argnum = 1;
+        IRubyObject sig = runtime.getCurrentContext().nil;
```
kares: its probably better to receive `ThreadContext context` instead or simply do a `runtime.getNil()` if the context is not needed anywhere else.

etehtsea: done. thanks for review!

enebo: @etehtsea @kares whether context is used or not we should always provide it as first argument to @JRubyMethod annotations so it will be future-proofed.  We want to make sure these method signatures never need to change for native extension authors.

@etehtsea a second tidbit of info is getRuntime().getCurrentContext() is not super quick which lead to us adding the ability to pass ThreadContext to method bindings.  We probably should have made it mandatory in 9k but we didn't :|

etehtsea: done

kares:
should be :green_apple:  to go.

headius:
@etehtsea There seems to be a number of whitespace changes in your commits. Would it be possible to strip those out? Otherwise I'll need to merge manually and omit them myself.

etehtsea:
@headius done

Issue #4127, opened 2016-09-01
OutOfMemoryError with 9.1.3.0


MartinKoerner:
### Environment

Provide at least:
* jruby 9.1.3.0 (2.3.1) 2016-08-29 a2a3b29 Java HotSpot(TM) 64-Bit Server VM 25.20-b23 on 1.8.0_20-b26 +jit [darwin-x86_64]

* JRUBY_OPTS='--debug --dev -Xcompile.invokedynamic=false -J-XX:TieredStopAtLevel=1 -J-XX:+CICompilerCountPerCPU -Xjit.threshold=500 -J-XX:+TieredCompilation -J-XX:+UseG1GC -J-XX:ReservedCodeCacheSize=256m -J-Xss2048k -J-Xms2048m -J-Xmx2048m'

* uname -a: 15.6.0 Darwin Kernel Version 15.6.0: Thu Jun 23 18:25:34 PDT 2016; root:xnu-3248.60.10~1/RELEASE_X86_64 x86_64

* Rails 4.2.

### Expected Behavior

* Command "bin/rails g our/generator" should run without problems
* The generator itself is resource-intensive, but ran always without problems in JRuby 1.7.x, 9.0.4.0, 9.0.5.0 and 9.1.2.0

### Actual Behavior

Generator runs for a while, than he hangs on one file (it is the same file every time, but it is not very special in the generated output) and after some time prints:
<pre>
LoadError: load error: rails/commands/generate -- java.lang.OutOfMemoryError: Java heap space
              require at org/jruby/RubyKernel.java:955
     block in require at /usr/local/var/rbenv/versions/jruby-9.1.3.0/lib/ruby/gems/shared/gems/activesupport-4.2.7.1/lib/active_support/dependencies.rb:274
      load_dependency at /usr/local/var/rbenv/versions/jruby-9.1.3.0/lib/ruby/gems/shared/gems/activesupport-4.2.7.1/lib/active_support/dependencies.rb:240
              require at /usr/local/var/rbenv/versions/jruby-9.1.3.0/lib/ruby/gems/shared/gems/activesupport-4.2.7.1/lib/active_support/dependencies.rb:274
     require_command! at /usr/local/var/rbenv/versions/jruby-9.1.3.0/lib/ruby/gems/shared/gems/railties-4.2.7.1/lib/rails/commands/commands_tasks.rb:123
[...]

PR #4128, opened 2016-09-01
Inspect directly into a RubyString and avoid StringBuilder/char[].

headius:
This should reduce the overhead of inspecting by avoiding the
intermediate StringBuilder/char[] and subsequent conversion to
byte[]. There are a few extra objects created along the way
(java.lang.String.getBytes, RubyString for class name, etc) but
overall memory use should reduce around 3x: a US-ASCII
string would need 2x byte size in the StringBuilder, and the
builder is still alive when we create the eventual byte[] for a
total of 3x max memory use during inspect.

Relates to #4127, but does not constitute a fix for memory issues
there.

### DIFF
--- a/core/src/main/java/org/jruby/RubyBasicObject.java
+++ b/core/src/main/java/org/jruby/RubyBasicObject.java
@@ -69,6 +69,7 @@ import org.jruby.runtime.builtin.Variable;
 import org.jruby.runtime.component.VariableEntry;
 import org.jruby.runtime.marshal.CoreObjectType;
 import org.jruby.util.ArraySupport;
+import org.jruby.util.ConvertBytes;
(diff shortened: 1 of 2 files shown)


headius:
Inspect perf comparison.

Script:
```ruby
obj = top = Object.new
100.times do
  obj2 = Object.new
  obj.instance_variable_set :@obj, obj2
  obj = obj2
end
loop do
  t = Time.now
  1000.times { top.inspect }
  puts Time.now - t
end
```

Before:
```
1.306
0.918
0.95
0.883
0.915
0.956
0.903
0.876
0.899
0.89
```

After:
```
0.339
0.18
0.184
0.205
0.148
0.155
0.195
0.153
0.149
0.198
0.153
```

Issue #4135, opened 2016-09-06
9.1.3.0 and 9.1.4.0 have a -SNAPSHOT dependency


chrisseaton:
Both releases depend on `com.github.jnr:jnr-constants:0.9.3-SNAPSHOT`

https://github.com/jruby/jruby/blob/9.1.3.0/core/pom.rb#L51
https://github.com/jruby/jruby/blob/9.1.4.0/core/pom.rb#L51

headius:
Boo...I though Sonatype was supposed to catch this sort of thing. https://issues.sonatype.org/browse/OSSRH-24828

More motivation for another quick-flip release.

mosabua:
I am not sure how your build works totally but typically you can catch this with using the Maven release plugin (can be a PITA.. I know). 

Alternatively for a more lightweight check you could add the enforcer plugin invocation to your release profile.

http://maven.apache.org/enforcer/enforcer-rules/requireReleaseDeps.html

headius:
Thank you for the suggestions, @mosabua!

Our release is rather complicated (almost a dozen different artifacts, submodules, etc) so the release plugin may not work for us. Adding a lightweight plugin to check it probably will work out better.

@mkristian what do you think?

chrisseaton:
If we just need to solve the problem, a grep of pom.rb for SNAPSHOT would have caught this. Could that go into whatever scripts you have for deploying.

headius:
It appears we do have this plugin in the root poms, but perhaps it's not properly getting enabled for a non-snapshot release:

/pom.rb
```ruby
    unless model.version =~ /-SNAPSHOT/
       #rules[:requireReleaseDeps] = { :message => 'No Snapshots Allowed!' }
    end
    plugin :enforcer, '1.4' do
      execute_goal :enforce, :rules => rules
    end
```

@chrisseaton I certainly wouldn't consider it a solution if we have to add more manual steps...and it's probably not a very good solution if we have to add something to @enebo's scripts when we should be shrinking them.

This really should be automated by either Maven itself or by Sonatype OSSRH. It's never correct for a maven-central jar to be released with snapshot deps.

mosabua:
You really just have to figure out there is no "-SNAPSHOT" version string. The "Maven way" to work around this is to use the release plugin or the enforcer rule. It seems you are already using it but the usage or the requireReleaseDeps rule is deactivated if I see this correctly from the snippet.

headius:
[...]

PR #4137, opened 2016-09-06
Unconditionally enable maven enforcer in release profile.

headius:
This will prevent issues like #4135 from happening in the future.


headius:
Ping @enebo and @mkristian for review.

PR #4144, opened 2016-09-11
Remove the last of NativeException's influence.

headius:
NativeException was replaced by simply rethrowing the actual Java
exception some time ago, and most code did the right thing.
However, a few places still threw NativeException in error along
unlikely call paths (e.g. unrecoverable class lookup errors).
This commit removes the last bits of NativeException use.

1. NativeException itself is stripped down to just be a simple
   subclass of RuntimeException in Ruby with no special backing
   Java object. Since it will no longer be created, code that
   uses it will never see it anyway. The class and its public
   methods remain to allow such code to compile.
2. All remaining places that threw NativeException were modified
   to rethrow the original exception.
3. Tests referencing NativeException were removed, along with any
   test utilities they alone consumed. The only remaining tests
   did so synthetically by constructing NativeException directly.


headius:
I went too far...we do still create NativeException when someone uses it in a rescue, and I broke many things expecting that NativeException would never be in flight. Will have to re-attempt this.

Issue #3992, opened 2016-07-02
[Truffle] Rubinius::Type.check_convert_type


bjfish:
This `Rubinius::Type.check_convert_type` comment says that it corresponds to `rb_check_convert_type` but it appears that it doesn't handle `respond_to_missing?` or `missing_method` methods.

Correcting this should fix the `array/flatten_spec.rb` `flatten performs respond_to? and method_missing-aware checks when coercing elements to array`.

`Rubinius::Type.coerce_to` should probably be reviewed too.

eregon:
This should be handled by the default `respond_to?`, no?
`check_convert_type` uses `vm_object_respond_to`, which uses `Kernel#respond_to?`, so I am not sure what's going wrong here.

eregon:
The issue seems to be there is no equivalent of `rb_check_funcall` at the Ruby level.

bjfish:
@eregon Could there be an issue with `vm_object_respond_to` ? The following example:

```
class Test
  def respond_to?(method_name, include_private = false)
    puts "IN RESPOND TO"
    true
  end
end
test = Test.new
puts "test.respond_to? :to_ary - #{test.respond_to? :to_ary}"
puts "Rubinius::Type.object_respond_to? test, :to_ary, true - #{Rubinius::Type.object_respond_to? test, :to_ary, true}"
```
Outputs:
```
IN RESPOND TO
test.respond_to? :to_ary - true
Rubinius::Type.object_respond_to? test, :to_ary, true - false
```

eregon:
@bjfish Yes, `vm_object_respond_to` is bound to `BasicObject#respond_to?`, so indeed we should probably try calling respond_to? dynamically. I tried a few things but gave up since it seems there are many differences here with MRI and it's quite hard to follow the code.

bjfish:
This was fixed by #4146

PR #4146, opened 2016-09-12
[Truffle] Add check_convert_type and check_funcall methods

bjfish:
This is a work in progress. Please review. 

This is for issue #3992.

I'm currently having difficulty here with 3 examples regressing. Here is a short example:

```ruby
class Test
  def initialize(name, options={})
    @name = name
    @null = options[:null_object]
  end

  def method_missing(sym, *args, &block)
    if @null 
      self 
    else  
      super
    end
  end
  private :method_missing
end

puts [Test.new("foo")].flatten
```
Output:
```
/Users/brandonfish/Documents/jruby-patches/mmtest.rb:11:in `method_missing': undefined method `to_ary' for #<Test:0xcc @name="foo" @null=nil>:Test (NoMethodError)
	from /Users/brandonfish/Documents/jruby-patches/mmtest.rb:11:in `method_missing'
	from /Users/brandonfish/Documents/jruby/truffle/src/main/ruby/core/type.rb:266:in `__send__'
	from /Users/brandonfish/Documents/jruby/truffle/src/main/ruby/core/type.rb:266:in `check_funcall_missing'
	from /Users/brandonfish/Documents/jruby/truffle/src/main/ruby/core/type.rb:250:in `check_funcall_default'
	from /Users/brandonfish/Documents/jruby/truffle/src/main/ruby/core/type.rb:244:in `check_funcall'
	from /Users/brandonfish/Documents/jruby/truffle/src/main/ruby/core/type.rb:233:in `convert_type'
	from /Users/brandonfish/Documents/jruby/truffle/src/main/ruby/core/type.rb:224:in `rb_check_convert_type'
	from /Users/brandonfish/Documents/jruby/truffle/src/main/ruby/core/array.rb:1196:in `block in recursively_flatten'
	from /Users/brandonfish/Documents/jruby/truffle/src/main/ruby/core/thread.rb:102:in `detect_recursion'
	from /Users/brandonfish/Documents/jruby/truffle/src/main/ruby/core/array.rb:1187:in `recursively_flatten'
	from /Users/brandonfish/Documents/jruby/truffle/src/main/ruby/core/array.rb:420:in `flatten'
	from /Users/brandonfish/Documents/jruby-patches/mmtest.rb:17:in `<main>'
```

Expected:
```
#<Test:0x007f863b051220>
```


eregon:
Just saw this today, maybe it can help:
https://github.com/jruby/jruby/blob/master/truffle/src/main/java/org/jruby/truffle/language/dispatch/DoesRespondDispatchHeadNode.java#L26

truffle/src/main/java/org/jruby/truffle/RubyContext.java
```diff
@@ -315,6 +318,14 @@ public AttachmentsManager getAttachmentsManager() {
...
     }
 
+    public boolean isRunning() {
+        return running;
+    }
+
+    public void setRunning(boolean running) {
+        this.running = running;
+    }
+
```
eregon: There is `state` in CoreLibrary for this :wink:

eregon:
Could you explain the semantics as you understand them?
They seem very strange and quite specific to Array#flatten.

truffle/src/main/java/org/jruby/truffle/core/module/ModuleNodes.java
```diff
@@ -398,7 +398,8 @@ private void createAccesor(DynamicObject module, String name) {
             final RubyNode sequence = Translator.sequence(getContext(), sourceSection.getSource(), rubySourceSection, Arrays.asList(checkArity, accessInstanceVariable));
             final RubyRootNode rootNode = new RubyRootNode(getContext(), sourceSection, null, sharedMethodInfo, sequence, false);
             final CallTarget callTarget = Truffle.getRuntime().createCallTarget(rootNode);
-            final InternalMethod method = new InternalMethod(sharedMethodInfo, accessorName, module, visibility, callTarget);
+            final boolean builtIn = !getContext().isRunning();
+            final InternalMethod method = new InternalMethod(sharedMethodInfo, accessorName, module, visibility, builtIn, callTarget);
```
eregon: This logic could be part of addMethod or InternalMethod constructor to avoid duplication.

truffle/src/main/ruby/core/type.rb
```diff
@@ -82,6 +82,15 @@ def self.object_respond_to?(obj, name, include_private = false)
       Truffle.invoke_primitive :vm_object_respond_to, obj, name, include_private
     end
 
+    def self.object_respond_to_no_built_in?(obj, name, include_private = false)
+      Truffle.invoke_primitive :vm_object_respond_to_no_built_in, obj, name, include_private
+    end
+
+    def self.check_funcall_callable(obj, me)
```
eregon: `me` is method entry in MRI, but maybe just call it `meth` for method here?

eregon:
Just a thought: maybe check_funcall can be implemented with sth like `m = obj.method(meth) and m.call(*args)` ?

bjfish:
@eregon `Array#flatten` calls `rb_check_convert_type(ary, T_ARRAY, "Array", "to_ary")` and which calls `convert_type` , I think these methods are used by most implicit type conversions like `NUM2INT` , etc ... so these are not specific to flatten.  I'm still working on understand the semantics for this and I think tracing this could offer better explanation than I could currently: https://github.com/ruby/ruby/blob/4915ce691d147a460e14531f4031d6934c5e8c5c/object.c#L2627

bjfish:
I don't believe `BasicObject` has this method`obj.method(meth)` and I don't think `method` filters "basic" methods.

truffle/src/main/java/org/jruby/truffle/RubyContext.java
bjfish: Thanks, I've updated this to use `state`

truffle/src/main/java/org/jruby/truffle/core/module/ModuleNodes.java
bjfish: I've moved this to the constructor

truffle/src/main/ruby/core/type.rb
bjfish: I've renamed `me` to `meth`

eregon:
> I don't believe BasicObject has this method obj.method(meth) and I don't think method filters "basic" methods.

I see.

> so these are not specific to flatten.

Right, the Rubinius code seemed pretty specific but MRI isn't.

eregon:
Let's look at the C then:

```c
VALUE
rb_check_funcall(VALUE recv, ID mid, int argc, const VALUE *argv)
{
    return rb_check_funcall_default(recv, mid, argc, argv, Qundef);
}

VALUE
rb_check_funcall_default(VALUE recv, ID mid, int argc, const VALUE *argv, VALUE def)
{
    VALUE klass = CLASS_OF(recv);
    const rb_callable_method_entry_t *me;
    rb_thread_t *th = GET_THREAD();
    int respond = check_funcall_respond_to(th, klass, recv, mid);

    if (!respond) // if check_funcall_respond_to returns 0, return Qundef
	return def;

    me = rb_search_method_entry(recv, mid);
    if (!check_funcall_callable(th, me)) {
	return check_funcall_missing(th, klass, recv, mid, argc, argv,
				     respond, def);
    }
    stack_check(th);
    return vm_call0(th, recv, mid, argc, argv, me);
}

static int
check_funcall_respond_to(rb_thread_t *th, VALUE klass, VALUE recv, ID mid)
{
    return vm_respond_to(th, klass, recv, mid, TRUE);
}

static int
vm_respond_to(rb_thread_t *th, VALUE klass, VALUE obj, ID id, int priv)
{
    VALUE defined_class;
    const ID resid = idRespond_to;
// This is looking up respond_to? on the metaclass of the object
    const rb_method_entry_t *const me =
	method_entry_get(klass, resid, &defined_class);

    if (!me) return -1; // if no respond_to?, assume it's OK
    if (METHOD_ENTRY_BASIC(me)) {
// if respond_to? is built-in (maybe we could just check if it is Kernel#respond_to?
// since that is the only built-in respond_to?), return -1 which means "OK keep going"
	return -1;
    }
    else {
// Some arity checking, doesnt seem important for now
	int argc = 1;
	VALUE args[2];
	VALUE result;

	args[0] = ID2SYM(id);
	args[1] = Qtrue;
	if (priv) {
	    argc = rb_method_entry_arity(me);
	    if (argc > 2) {
		rb_raise(rb_eArgError,
			 "respond_to? must accept 1 or 2 arguments (requires %d)",
			 argc);
	    }
	    if (argc != 1) {
		argc = 2;
	    }
	    else if (!NIL_P(ruby_verbose)) {
		VALUE location = rb_method_entry_location(me);
		rb_warn("%"PRIsVALUE"%c""respond_to?(:%"PRIsVALUE") uses"
			" the deprecated method signature, which takes one parameter",
			(FL_TEST(klass, FL_SINGLETON) ? obj : klass),
			(FL_TEST(klass, FL_SINGLETON) ? '.' : '#'),
			QUOTE_ID(id));
		if (!NIL_P(location)) {
		    VALUE path = RARRAY_AREF(location, 0);
		    VALUE line = RARRAY_AREF(location, 1);
		    if (!NIL_P(path)) {
			rb_compile_warn(RSTRING_PTR(path), NUM2INT(line),
					"respond_to? is defined here");
		    }
		}
	    }
	}
// if obj's respond_to? is not Kernel's and is defined, actually call it
// aka !!meth.call(name)
	result = call_method_entry(th, defined_class, obj, resid, me, argc, args);
	return RTEST(result);
    }
}
```

truffle/src/main/ruby/core/type.rb
```diff
@@ -210,6 +219,64 @@ def self.execute_check_convert_type(obj, cls, meth)
...
+      unless object_kind_of?(v, cls)
+        raise TypeError, "Coercion error: obj.#{meth} did NOT return a #{cls} (was #{object_class(v)})"
+      end
+      v
+    end
+
+    def self.convert_type(obj, cls, meth, raise)
+      r = check_funcall(obj, meth)
+      if undefined.equal?(r)
+        if raise
```
eregon: Don't name it `raise` here.

truffle/src/main/java/org/jruby/truffle/core/VMPrimitiveNodes.java
```diff
@@ -263,6 +273,66 @@ public boolean vmObjectRespondTo(VirtualFrame frame, Object object, Object name,
...
+            }
+
+            if(method.isBuiltIn()){
+                return false;
+            }
+
+            return true;
+        }
+
+    }
```
eregon: I think this would be easier to follow if it returned a Ruby Method or nil.
And maybe a separate primitive to check if a Method is built-in or not.

bjfish: I've refactored this to use the primitives you mentioned

truffle/src/main/ruby/core/type.rb
bjfish: I've updated the name of this param

truffle/src/main/ruby/core/type.rb
```diff
@@ -82,6 +82,16 @@ def self.object_respond_to?(obj, name, include_private = false)
       Truffle.invoke_primitive :vm_object_respond_to, obj, name, include_private
     end
 
+    def self.object_respond_to_no_built_in?(obj, name, include_private = false)
+      meth = Truffle.invoke_primitive :vm_method_lookup, obj, name
+      !meth.nil? && !(Truffle.invoke_primitive :vm_method_is_basic, meth)
```
eregon: I would move parenthesis around the call here: `!Truffle.invoke_primitive(:vm_method_is_basic, meth)`

truffle/src/main/ruby/core/type.rb
```diff
@@ -210,6 +220,64 @@ def self.execute_check_convert_type(obj, cls, meth)
...
+      if undefined.equal?(r)
+        if raise_on_error
+          raise TypeError, "can't convert #{obj} into #{cls} with #{meth}"
+        end
+        return nil
+      end
+      r
+    end
+
+    def self.check_funcall(recv, meth, args = undefined)
```
eregon: args should probably default to `[]` or be removed if it's never used.

truffle/src/main/ruby/core/type.rb
```diff
@@ -210,6 +220,64 @@ def self.execute_check_convert_type(obj, cls, meth)
...
+
+    def self.check_funcall_respond_to(obj, meth, priv)
+      # TODO Review BJF vm_respond_to
+      return true unless object_respond_to_no_built_in?(obj, :respond_to?, true)
+      !!obj.__send__(:respond_to?, meth, true)
+    end
+
+    def self.check_funcall_missing(recv, meth, args, respond, default)
+      ret = basic_obj_respond_to_missing(recv, meth, false) #PRIV false
+      return default unless undefined.equal?(ret) || ret
```
eregon: The logic is equivalent I think:
```c
if (!RTEST(ret)) return def;
```
So
```ruby
return default unless basic_obj_respond_to_missing(recv, meth, false) #PRIV false
```

truffle/src/main/ruby/core/type.rb
```diff
@@ -210,6 +220,64 @@ def self.execute_check_convert_type(obj, cls, meth)
...
+      if object_respond_to_no_built_in?(recv, :method_missing, true)
+        begin
+          return recv.__send__(:method_missing, meth)    # , args
+        rescue NoMethodError
+          # TODO BJF usually more is done here
+          raise
```
eregon: Maybe the missing logic here is causing the failure?

truffle/src/main/ruby/core/type.rb
bjfish: I've updated this to your recommendation, thanks

truffle/src/main/ruby/core/type.rb
bjfish: I've update this to `[]` and used them later on to send `*args`

truffle/src/main/ruby/core/type.rb
bjfish: @eregon I think you right, I've added a new commit with more `check_funcall_missing` implemented which resolves the spec failures

truffle/src/main/ruby/core/type.rb
```diff
@@ -210,6 +220,80 @@ def self.execute_check_convert_type(obj, cls, meth)
...
+      ret = respond > 0
+      ret = basic_obj_respond_to_missing(recv, meth, false) #PRIV false
+      return default unless ret
+      respond_to_missing = !undefined.equal?(ret)
+      ret = default
+      if object_respond_to_no_built_in?(recv, :method_missing, true)
+        begin
+          return recv.__send__(:method_missing, meth, *args)
+        rescue NoMethodError
+          # TODO BJF usually more is done here
```
eregon: Is that TODO addressed now?

eregon:
Looks good, congrats for translating this fairly complex code!

truffle/src/main/ruby/core/type.rb
bjfish: I think there's some cases that still aren't handled here yet: https://github.com/ruby/ruby/blob/dba03c38c54835248766f28033b5aeddc5f052ef/vm_eval.c#L369

bjfish:
@eregon Thanks for the help with this.  I forgot to mention, this was the last untagged core array spec, so array is at 100% now, 🎉

Issue #1405, opened 2014-01-15
SecureRandom performance could be better


headius:
Our SecureRandom in 1.7.10 (from securerandom.rb) was implemented using Java integration calling out to java.security.SecureRandom. It constructed a new instance each time.

In 09b169a6cd047c6d4d49104a37cf30a723c2bb36, I moved the meat of the logic (#random_bytes) into Java code, which roughly doubled performance on my machine, bringing it close to MRI.

@bbrowning and @nirvdrum discovered a JVM property that, when passed, sets SecureRandom to use /dev/urandom or /dev/random, which improves performance further: java.security.egd=/dev/random. On my system, this again doubles performance, but the format appears to vary across JVMs.

Finally, I made a modification that constructs a SecureRandom per thread, avoiding synchronization contention, and started to move other methods into Java starting with #hex. This brought about the best performance so far, but I have not committed this change because I'm unsure about saving the SecureRandom per thread and whether that does something "bad" as far as the security of its results.

We need to evaluate all these options and come up with a good balance of security and performance, since it seems that SecureRandom is hit very heavily by frameworks like Rails.

Performance with per-thread SecureRandom and the Java property across JVMs on OS X: https://gist.github.com/headius/8428930

Patch to use a SecureRandom per thread and make #hex native: https://gist.github.com/headius/8429206

headius:
Ok, waking this one up again...

There are still outstanding reports of blocking or slow behavior using SecureRandom, like nats-io/ruby-nats#123 and [this comment](https://github.com/jruby/jruby/issues/3351#issuecomment-190393752) by @cheald.

It seems we're still not using whatever is equivalent to MRI, since as far as I know MRI does not get reports of entropy-starved blocking behaviors.

So...anyone still seeing this problem that can help us explore solutions? @cheald @trinode @digitalextremist @tarcieri @wallyqs

cheald:
I haven't seen it lately since I've added entropy daemons to my standard saltstack on new machines, but it might be possible to replicate it on a fresh VM. I'll run some experiments on AWS VMs.

cheald:
I'm unable to replicate the issue in a fresh VM here. Amazon Linux, no entropy daemon. I'm keeping the entropy pool drained:

````
# cat /dev/random > /dev/null &
# cat /proc/sys/kernel/random/entropy_avail
4
````

Then installed JRuby and benchmarked SecureRandom:
```
[...]

PR #4149, opened 2016-09-12
Backport securerandom

headius:
This PR backports recent SecureRandom entropy fixes from 9k to 1.7. See #1405.

We had a 1.7 user in #jruby today with entropy problems.

### DIFF
--- a/core/src/main/java/org/jruby/util/cli/Options.java
+++ b/core/src/main/java/org/jruby/util/cli/Options.java
@@ -146,7 +146,7 @@ public class Options {
     public static final Option<Boolean> REIFY_VARIABLES = bool(MISCELLANEOUS, "reify.variables", false, "Attempt to expand instance vars into Java fields");
     public static final Option<Boolean> PREFER_IPV4 = bool(MISCELLANEOUS, "net.preferIPv4", true, "Prefer IPv4 network stack");
     public static final Option<Boolean> FCNTL_LOCKING = bool(MISCELLANEOUS, "file.flock.fcntl", true, "Use fcntl rather than flock for File#flock");
-    public static final Option<String> PREFERRED_PRNG = string(MISCELLANEOUS, "preferred.prng", "NativePRNGNonBlocking", "Maintain children static scopes to support scope dumping.");
+    public static final Option<String> PREFERRED_PRNG = string(MISCELLANEOUS, "preferred.prng", "NativePRNGNonBlocking", "Set the preferred JDK-supported random number generator to use.");
 
     public static final Option<Boolean> DEBUG_LOADSERVICE = bool(DEBUG, "debug.loadService", false, "Log require/load file searches.");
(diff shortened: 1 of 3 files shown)


kares:
isn't https://github.com/jruby/jruby/pull/4149/commits/8fb4b9314dff380b6bccd5174bf456d0b6c271d7 "dangerous" to back-port (changing a previously initialized field on `ThreadContext` to null) ?

core/src/main/java/org/jruby/util/cli/Options.java
```diff
@@ -146,6 +146,7 @@
     public static final Option<Boolean> REIFY_VARIABLES = bool(MISCELLANEOUS, "reify.variables", false, "Attempt to expand instance vars into Java fields");
     public static final Option<Boolean> PREFER_IPV4 = bool(MISCELLANEOUS, "net.preferIPv4", true, "Prefer IPv4 network stack");
     public static final Option<Boolean> FCNTL_LOCKING = bool(MISCELLANEOUS, "file.flock.fcntl", true, "Use fcntl rather than flock for File#flock");
+    public static final Option<String> PREFERRED_PRNG = string(MISCELLANEOUS, "preferred.prng", "NativePRNGNonBlocking", "Maintain children static scopes to support scope dumping.");
```
kares: the help message for the option seems like a copy :spaghetti: left over

headius: Indeed...I'll fix that on both master and in this PR.

headius:
> @kares: isn't 8fb4b93 "dangerous" to back-port (changing a previously initialized field on ThreadContext to null) ?

A little dangerous, I agree. However the internals of ThreadContext have never been a blessed API. I know there might be people poking at them, but we've never shown people examples that use TC internals and we have generally discouraged messing with it.

I'm not too concerned, especially since it helps thread startup times (important for Fiber and Enumerator#next as well). @enebo?

headius:
@takanuva This is the backported SecureRandom changes I mentioned. Since you have an environment with entropy problems, perhaps you could test out a HEAD build of JRuby 1.7 for us once we merge?

It would be REALLY nice to know that this actually fixes your issue without installing an entropy-generator. Up to now we've just assumed it does, since it's hard to simulate an environment with entropy-starvation problems.

enebo:
@bbrowning or @mkristian can you see a problem with this change to threadcontext not having securerandom initialized?  I personally think the potential for this breaking stuff could only happen in a deep extension but not very likely.

bbrowning:
@enebo I don't know of anything that uses securerandom directly like this. I agree it would have to be an extension really digging deep to get hit by this. While it is possible that one exists somewhere, a quick GitHub search didn't turn up anything obvious.

kares:
... no (more) worries from me than - thanks everyone for looking into it!

takanuva:
Sure, @headius, just give me a few days and I'll gladly test that for you guys. 😄

PR #4152, opened 2016-09-13
Fix for #3229

### LINKED ISSUE #3229: Socket.unix and friends use missing Addrinfo features
Reported on IRC...

Trying to use `Socket.unix("path")` fails with `IOError: path: Name or service not known`

This method and a few others like it are implemented using Addrinfo to produce the new socket. Since Addrinfo is not fully implemented on JRuby, there's some missing pieces here.

This affects at least JRuby 1.7 and may affect 9k, but I have not confirmed.

etehtsea:
Almost full `Addrinfo` support and many other sockets-related improvements.

fixes #3229


core/src/main/java/org/jruby/ext/socket/RubyIPSocket.java
```diff
@@ -46,51 +48,37 @@
  * @author <a href="mailto:ola.bini@ki.se">Ola Bini</a>
  */
 @JRubyClass(name="IPSocket", parent="BasicSocket")
-public class RubyIPSocket extends RubyBasicSocket {
+public abstract class RubyIPSocket extends RubyBasicSocket {
```
kares: `IPSocket` is still `allocate`-able ... not that it matters much (probably doesn't at all)
```
2.3.1 :029 > IPSocket.new
NoMethodError: undefined method `initialize' for #<IPSocket:0x0000000128ed70>
	from (irb):29:in `new'
	from (irb):29
	from /opt/local/rvm/rubies/ruby-2.3.1/bin/irb:11:in `<main>'
2.3.1 :030 > IPSocket.allocate
 => #<IPSocket:0x000000012574b0> 
```

headius:
This is a big PR! Well done stitching your commits and my commits together!

I must raise the issue of whitespace again, though. You have a number of places here changing nothing but whitespace. In such a big PR I don't expect you to go back and fix that, but please be mindful of it in the future and try to set your IDE to not make spurious formatting changes.

I'll review.

core/src/main/java/org/jruby/ext/socket/RubyIPSocket.java
headius: I presume this means you could extend IPSocket and make your own...not that I know how you'd do that with so much hidden in C.

I agree we should probably undo this change.

headius:
@etehtsea It looks good to me. Hard to tell which code is yours and which is mine now :-)

The IPSocket abstract/NotAllocatable change should be reverted, but that's all I saw.

Maybe @enebo can weigh in from Japan if he's got a moment.

enebo:
EPIC...I did not spend much time on this but landing sooner and getting it pushed through the meat grinder (e.g. we need some bake + expanded testing).

One other tiny note: context.getRuntime() is fine but it is ok to do context.runtime  if you want.  We do this in so many places now I think it is an acceptable pattern to access the field.

etehtsea:
@headius done. Also reverted for `BasicSocket`:

JRuby after fix:

```
>> BasicSocket.allocate
=> #<BasicSocket:0x1f7030a6>
>> BasicSocket.new
NoMethodError: undefined method `initialize' for #<BasicSocket:0x396f6598>
>> IPSocket.allocate
=> #<IPSocket:0x394e1a0f>
>> IPsocket.new
NoMethodError: undefined method `initialize' for #<IPSocket:0x1d29cf23>
```

MRI 2.3.1:

```
[1] pry(main)> BasicSocket.allocate
=> #<BasicSocket:0x007fb4eb5292a0>
[2] pry(main)> BasicSocket.new
NoMethodError: undefined method `initialize' for #<BasicSocket:0x007fb4eb4f0950>
from (pry):2:in `new'
[3] pry(main)> IPSocket.allocate
=> #<IPSocket:0x007fb4eb4bb408>
[4] pry(main)> IPSocket.new
NoMethodError: undefined method `initialize' for #<IPSocket:0x007fb4e9c5f918>
```

headius:
@eregon I think we need to address the rubysl-socket spec question. The specs there are considerably more complete, and appear to be licensed BSD. I think we should pull them into ruby/spec, so we can all start running them as part of that suite.

eregon:
@headius I missed this message, but essentially I think the simplest for now is to integrate the rubysl-socket specs in CI separately, as there is an overlap between both kind of specs and I'm unsure what to do LICENSE-wise.

PR #4153, opened 2016-09-14
Fix for #4147

### LINKED ISSUE #4147: Logger: log rotation inter-process lock failed. on Windows with JRuby 9.x
### Environment

Provide at least:
* jruby 9.1.2.0 (2.3.0) 2016-05-26 7357c8f Java HotSpot(TM) 64-Bit Server VM 25.102-b14 on 1.8.0_102-b14 +jit [mswin32-x86_64]
* Windows 10 + Oracle Java SE JDK 1.8.0_102

Test Codes
```
  require 'logger'
  logger = Logger.new("C:\\log.txt", 5, 1024)
  150.times do |i|
    logger.info i
  end
```

### Expected Behavior

* Create "log.txt", "log.txt.0", "log.txt.1", "log.txt.2", "log.txt.3" files.

### Actual Behavior

* Raise error:
```
  log rotation inter-process lock failed. Permission denied - C:\log.txt or C:\log.txt.0
  log writing failed. closed stream
  log shifting failed. closed stream
  ...
```

We test it in JRuby1.6.8, JRuby1.7.22 and Ruby 2.3.1,
they are correct.

And we find the source codes in jruby-9.1.2.0\lib\ruby\stdlib\logger.rb line 745
```
  if /mswin|mingw/ =~ RUBY_PLATFORM
    def lock_shift_log
      yield
    end
  else
    ...
  end
```
For Ruby, the value of RUBY_PLATFORM likes "x64-mingw32".
But for JRuby, the value of RUBY_PLATFORM is "java".

For test, we change
`  if /mswin|mingw/ =~ RUBY_PLATFORM`
to
`  if /mswin|mingw|java/ =~ RUBY_PLATFORM`
it is correct too.

So, is this a JRuby bug?

smk0621:
When check the OS platform, the RUBY_PLATFORM return 'java' in JRuby.
It is better to use RbConfig::CONFIG['host_os'] .

fixes #4147

### DIFF
--- a/lib/ruby/stdlib/logger.rb
+++ b/lib/ruby/stdlib/logger.rb
@@ -742,7 +742,7 @@ private
       end
     end
 
-    if /mswin|mingw/ =~ RUBY_PLATFORM
+    if /mswin|mingw/ =~ RbConfig::CONFIG['host_os']
       def lock_shift_log
         yield
       end


enebo:
@headius can you commit this upstream? I don't know if they want rbconfig but this is a better check

PR #4161, opened 2016-09-18
[Truffle] Encoding#replicate

bjfish:
Work In Progress.

I've got one known failure here which I think is due to the following from `encoding.rb` not recognizing the encodings added at runtime. I'm still thinking about a fix for this. I'll probably replace these with a primitive for the few places these are used.
```
  EncodingList = Encoding.list.freeze
  EncodingMap = build_encoding_map
```

Please review @nirvdrum

### DIFF
--- a/truffle/src/main/java/org/jruby/truffle/core/encoding/EncodingManager.java
+++ b/truffle/src/main/java/org/jruby/truffle/core/encoding/EncodingManager.java
@@ -29,14 +29,16 @@ import org.jruby.util.encoding.ISO_8859_16;
 import java.nio.charset.Charset;
 import java.nio.charset.StandardCharsets;
 import java.nio.charset.UnsupportedCharsetException;
+import java.util.ArrayList;
 import java.util.HashMap;
+import java.util.List;
 import java.util.Locale;
 import java.util.Map;
 
 public class EncodingManager {
 
-    private final DynamicObject[] ENCODING_LIST_BY_ENCODING_LIST_INDEX = new DynamicObject[EncodingDB.getEncodings().size()];
-    private final DynamicObject[] ENCODING_LIST_BY_ENCODING_INDEX = new DynamicObject[EncodingDB.getEncodings().size()];
+    private final List<DynamicObject> ENCODING_LIST_BY_ENCODING_LIST_INDEX = new ArrayList<DynamicObject>(EncodingDB.getEncodings().size());
+    private final Map<Integer, DynamicObject> ENCODING_LIST_BY_ENCODING_INDEX = new HashMap<Integer,DynamicObject>(EncodingDB.getEncodings().size());
     private final Map<String, DynamicObject> LOOKUP = new HashMap<>();
 
     private final RubyContext context;
@@ -55,19 +57,19 @@ public class EncodingManager {
         return Layouts.ENCODING.createEncoding(context.getCoreLibrary().getEncodingFactory(), encoding, string, dummy);
     }
 
-    public DynamicObject[] getUnsafeEncodingList() {
+    public List<DynamicObject> getUnsafeEncodingList() {
         return ENCODING_LIST_BY_ENCODING_LIST_INDEX;
     }
 
     public DynamicObject getRubyEncoding(Encoding encoding) {
-        DynamicObject rubyEncoding = ENCODING_LIST_BY_ENCODING_INDEX[encoding.getIndex()];
+        DynamicObject rubyEncoding = ENCODING_LIST_BY_ENCODING_INDEX.get(encoding.getIndex());
 
         if (rubyEncoding == null) {
             // Bounded by the number of encodings
             CompilerDirectives.transferToInterpreterAndInvalidate();
 
             rubyEncoding = LOOKUP.get(new String(encoding.getName(), StandardCharsets.UTF_8).toLowerCase(Locale.ENGLISH));
-            ENCODING_LIST_BY_ENCODING_INDEX[encoding.getIndex()] = rubyEncoding;
+            ENCODING_LIST_BY_ENCODING_INDEX.put(encoding.getIndex(),rubyEncoding);
         }
 
         return rubyEncoding;
@@ -80,14 +82,15 @@ public class EncodingManager {
 
     @TruffleBoundary
     public DynamicObject getRubyEncoding(int encodingListIndex) {
-        return ENCODING_LIST_BY_ENCODING_LIST_INDEX[encodingListIndex];
+        return ENCODING_LIST_BY_ENCODING_LIST_INDEX.get(encodingListIndex);
     }
 
     @TruffleBoundary
     public void defineEncoding(EncodingDB.Entry encodingEntry, byte[] name, int p, int end) {
         final DynamicObject rubyEncoding = newRubyEncoding(context, null, name, p, end, encodingEntry.isDummy());
 
-        ENCODING_LIST_BY_ENCODING_LIST_INDEX[encodingEntry.getIndex()] = rubyEncoding;
+        assert ENCODING_LIST_BY_ENCODING_LIST_INDEX.size() == encodingEntry.getIndex();
+        ENCODING_LIST_BY_ENCODING_LIST_INDEX.add(encodingEntry.getIndex(), rubyEncoding);
         LOOKUP.put(Layouts.ENCODING.getName(rubyEncoding).toString().toLowerCase(Locale.ENGLISH), rubyEncoding);
     }
 
--- a/truffle/src/main/java/org/jruby/truffle/core/encoding/EncodingManager.java
+++ b/truffle/src/main/java/org/jruby/truffle/core/encoding/EncodingManager.java
@@ -61,6 +61,7 @@ public class EncodingManager {
         return ENCODING_LIST_BY_ENCODING_LIST_INDEX;
     }
 
+    @TruffleBoundary
     public DynamicObject getRubyEncoding(Encoding encoding) {
         DynamicObject rubyEncoding = ENCODING_LIST_BY_ENCODING_INDEX.get(encoding.getIndex());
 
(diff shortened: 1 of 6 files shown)


eregon:
>  I'll probably replace these with a primitive for the few places these are used.

:+1:

truffle/src/main/java/org/jruby/truffle/core/encoding/EncodingManager.java
```diff
@@ -29,14 +29,16 @@
...
 import java.util.Locale;
 import java.util.Map;
 
 public class EncodingManager {
 
-    private final DynamicObject[] ENCODING_LIST_BY_ENCODING_LIST_INDEX = new DynamicObject[EncodingDB.getEncodings().size()];
-    private final DynamicObject[] ENCODING_LIST_BY_ENCODING_INDEX = new DynamicObject[EncodingDB.getEncodings().size()];
+    private final List<DynamicObject> ENCODING_LIST_BY_ENCODING_LIST_INDEX = new ArrayList<DynamicObject>(EncodingDB.getEncodings().size());
+    private final Map<Integer, DynamicObject> ENCODING_LIST_BY_ENCODING_INDEX = new HashMap<Integer,DynamicObject>(EncodingDB.getEncodings().size());
     private final Map<String, DynamicObject> LOOKUP = new HashMap<>();
```
eregon: These will probably need some synchronization if encodings can be added after startup.
Maybe a `CopyOnWriteArrayList`  and a `ConcurrentHashMap`.

bjfish:
It is pretty difficult to make updates in the `Encoding` area because we have lists/maps of Encodings in the EncodingDB, EncodingService, EncodingManager, and in the Encoding::EncodingList/Map.

Also, I don't think JRuby replicates correctly currently because I don't see the Encoding appear in the Encoding.name_list or Encoding.list.

truffle/src/main/java/org/jruby/truffle/core/encoding/EncodingManager.java
bjfish: I've added the `synchronized` keyword as we discussed.

nirvdrum:
You've probably worked this out by now, but `EncodingDB` is jcodings's way of storing the underlying encoding objects, `EncodingService` is JRuby's way of bridging Ruby and jcodings, `EncodingManager` is our way of bridging Ruby and jcodings, and `Encoding::EncodingList/Map` is our way of bridging Rubinius and jcodings.

nirvdrum:
@eregon would know better than me, but I think `CopyOnWriteArrayList` is better than `synchronized` here. In virtually all expected cases, the encoding list is processed during bootstrap and then effectively becomes read-only. Synchronizing those read methods seems unnecessarily heavy to me.

bjfish:
@nirvdrum, @eregon would know better than me too here, correct me if I am wrong ...

I discussed choosing `synchronized` over `CopyOnWriteArrayList` with @eregon  here and he agreed but we didn't really discuss why.  

I though `CopyOnWriteArrayList\ConcurrentHashMap` here would be too granular because our methods update/read more than one of these data structures at once. So, I thought `synchronized` on every method would work to ensure that our  reads/writes to multiple data structures are consistent. 

I agree that `synchronized` may be more synchronization than we need because we probably don't need to synchronize reads when no writing is occurring which is where we could use a `ReadWriteLock` to only sync reads when writes are happening.  For the moment, I just chose to use the blunt approach `synchronized` because I didn't think read contention here would be a common scenario for us right now but maybe it is?

nirvdrum:
I don't think you really need to worry about consistency all that much in general. Replicating an encoding doesn't affect the indices or name mappings of existing encodings in EncodingDB or EncodingManager. So, if a thread is looking at an older version of one of the already bootstrapped data structures, they should still get valid data. The contention is really around one thread replicating an encoding and another thread trying to use it, which should be handled by user code in my opinion. As long as a single thread's view doesn't become inconsistent, I think we're okay.

truffle/src/main/java/org/jruby/truffle/core/encoding/EncodingManager.java
```diff
@@ -55,44 +57,46 @@ public static DynamicObject newRubyEncoding(RubyContext context, Encoding encodi
         return Layouts.ENCODING.createEncoding(context.getCoreLibrary().getEncodingFactory(), encoding, string, dummy);
     }
 
-    public DynamicObject[] getUnsafeEncodingList() {
+    public synchronized List<DynamicObject> getUnsafeEncodingList() {
```
eregon: This synchronized is useless though :p
Since the full list is returned there could be ConcurrentModificationException when iterating while it's being modified.
So either use a CopyOnWriteArrayList here or replace with a helper that just returns one encoding and not the whole list.

eregon:
@nirvdrum `synchronized` is not so heavy but here I'm of the opinion using concurrent data-structures might be enough and actually easier to follow.
`synchronized` would ensure all three are kept consistent but I would suspect only one of them is used for reading at a time, so inconsistencies should be unnoticeable.

@bjfish I'd like to have a look in my IDE and make this thread-safe after your PR, could you remove the `synchronized` commit? Then let's merge this without mixing with concurrency :)

truffle/src/main/java/org/jruby/truffle/core/encoding/EncodingNodes.java
```diff
@@ -554,6 +560,37 @@ public DynamicObject encodingGetObjectEncodingNil(DynamicObject object) {
...
+            final Entry entry = EncodingDB.getEncodings().get(nameString.getBytes());
+            getContext().getEncodingManager().defineEncoding(entry, nameString.getBytes(), 0, nameString.getBytes().length);
+            final DynamicObject newEncoding = getContext().getEncodingManager().getRubyEncoding(nameString.toLowerCase(Locale.ENGLISH));
+            snippetNode.execute(frame, "Encoding::EncodingMap[enc.name.upcase.to_sym] = [nil, index]", "enc", newEncoding, "index", entry.getIndex());
+            return newEncoding;
+        }
+
+    }
+
+    @Primitive(name = "encoding_get_object_encoding_by_index", needsSelf = false)
```
eregon: Maybe just `encoding_get_encoding_by_index` ?
Since anyway primitives return Ruby-level values.

bjfish:
@eregon I've reverted the synchronized commit at 4222dc81f21be61384bbe635cc90d32ad4b19c8b

truffle/src/main/java/org/jruby/truffle/core/encoding/EncodingNodes.java
bjfish: Fixed at 79e5a8adf1370fd4a78147a5cb7e50ffdde35c1c

eregon:
Let's get this merged as soon as it's green!

truffle/src/main/java/org/jruby/truffle/core/encoding/EncodingManager.java
```diff
@@ -29,14 +29,16 @@
...
+import java.util.List;
 import java.util.Locale;
 import java.util.Map;
 
 public class EncodingManager {
 
-    private final DynamicObject[] ENCODING_LIST_BY_ENCODING_LIST_INDEX = new DynamicObject[EncodingDB.getEncodings().size()];
-    private final DynamicObject[] ENCODING_LIST_BY_ENCODING_INDEX = new DynamicObject[EncodingDB.getEncodings().size()];
+    private final List<DynamicObject> ENCODING_LIST_BY_ENCODING_LIST_INDEX = new ArrayList<DynamicObject>(EncodingDB.getEncodings().size());
+    private final Map<Integer, DynamicObject> ENCODING_LIST_BY_ENCODING_INDEX = new HashMap<Integer,DynamicObject>(EncodingDB.getEncodings().size());
```
eregon: Why did this become a map? That's fairly expensive.

eregon: Because the List API doesn't support adding at an index > size I guess.

nirvdrum: Aliases resolve to their aliased encoding:

```
irb(main):002:0> 'abc'.encode('ASCII-8BIT').encoding.name
=> "ASCII-8BIT"
irb(main):003:0> 'abc'.encode('BINARY').encoding.name
=> "ASCII-8BIT"
```

Since we associate only the jcodings `Encoding` with the rope, we need to look up the associated `DynamicObject`. Prior to this change, there was essentially a 1:1 mapping, since aliases always resolved to their aliased encoding.

`Encoding#replicate`, however, works differently.

```
irb(main):004:0> 'abc'.encode('BINARY').encoding.name
=> "ASCII-8BIT"
irb(main):005:0> 'abc'.encode(Encoding::BINARY.replicate('blah')).encoding.name
=> "blah"
```

I'm pretty sure he made it a map so he could store multiple `DynamicObject`s with the same encoding index.

eregon: Interestingly, this works, but I am not sure why:
```
jt ruby -e 'Encoding::BINARY.replicate("blah"); p e=Encoding.find("blah"); p "".encode(e).encoding; p "".encode(Encoding::BINARY).encoding'
#<Encoding:blah>
#<Encoding:blah>
#<Encoding:ASCII-8BIT>
```

eregon: It creates another Encoding instance, lazily, see `Encoding.replicate`.

bjfish:
@eregon 
> Because the List API doesn't support adding at an index > size I guess.

This is correct.  This could probably be done with an `ArrayList.set` as long as we ensure that the index we are setting is `index < size` which may be done by adding a `null` every time an encoding is defined.

@nirvdrum 
> I'm pretty sure he made it a map so he could store multiple DynamicObjects with the same encoding index.

My understanding was that `EncodingDB.replicate` registers a new Encoding with it's own index. So, I did not intend to  "store multiple DynamicObjects with the same encoding index."

nirvdrum:
@bjfish Thanks for the clarification.

PR #4166, opened 2016-09-21
Truffle: use jt in tests

eregon:
This PR uses `jt` in test/truffle and in jtr if available. This has several advantages:

* Can run using other launchers than bin/jruby (notably with mx, jruby_eclipse, etc).
* Does no longer require to modify the PATH or ENV! (so commands in *.sh can be copy-pasted to CLI and work most of the cases)
* Shows the command line used as jt does by default which helps debugging.

@pitr-ch Please review the jtr modification to use `jt` as the "interpreter".
@nirvdrum @pitr-ch I needed to introduce the equivalent of jtr's `--no-use-fs-core` in `jt`.
I named it `--no-core-load-path`, but maybe we can use a common name, do you have any idea?
Maybe `--fs-core`/`--no-fs-core`?


eregon:
cc @jruby/truffle

chrisseaton:
Does this let us test a standalone executable like SVM or the Ruby binary in a GraalVM build?

eregon:
It can easily, but we'll need to tell `jt` to use that.
Probably using an environment variable (`RUBY_BIN`, `RUBY_LAUNCHER`)?.

eregon:
@chrisseaton Actually RUBY_BIN is already supported by `jt`, so it will just need to be set when needed.

chrisseaton:
Ok. My only worry is ensuring we are running what we think we are when we test. Since the command is printed that's probably fine.

Issue #4167, opened 2016-09-21
Create a different hierarchy of DynamicScope for use in startup interp + remove null checks in the existing hierarchy since they are redundant


subbuss:
The runtime now has a AddMissingInitsPass that the full interp and JIT modes use to add any required initializations of variables. This eliminates the need to do the same in DynamicScopes when lvars are looked up and they can be removed which might potentially lead to some perf. improvements by making DynamicScope code lean.

However, the startup interpreter won't have these IR-added inits in place and will still need the null checks. A good way to do that is add a different hierarchy there, and push those dynamic scopes for IR scopes that are running in startup mode. Maybe even a single 'multi-var-dynamic-scope' is sufficient for starters to evaluate what this yields. If startup impact is bad, the single scope can be specialized for different scenarios.

headius:
Some ideas from discussing with Tom:

* Initial move could be to add a new DynamicScope that supports arbitrarily-many variables (like ManyVars) and which does null checking. Remove null checking from the other scopes, since they'll only be made by Full and JIT.
* Any methods that currently null check would need to keep null-checking, for non-Ruby code that doesn't have guarantees about initialization. Ruby code would call through new non-checked paths. This may be a way to avoid adding a new DynamicScope type at the cost of duplicating all current null-checking methods (i.e. Simple just uses existing null-checked methods, Full and JIT move to using unchecked methods).
* It may be time we bought or built a library that can generate specialized collections from a template. For example, we don't want to have to maintain N DynamicScope classes in source for N widths of frame...better would be to have one template and generate those N classes at runtime as needed. This also allows us to potentially make JIT go directly at fields on those generated objects, something that was infeasible with full null checking.

subbuss:
> Any methods that currently null check would need to keep null-checking, for non-Ruby code that doesn't have guarantees about initialization. Ruby code would call through new non-checked paths. This may be a way to avoid adding a new DynamicScope type at the cost of duplicating all current null-checking methods (i.e. Simple just uses existing null-checked methods, Full and JIT move to using unchecked methods).

My thought as well.

headius:
I've merged my PR that adds unchecked get* paths to all DynamicScope types. It also removes a scope type we no longer used, and fixes a null leak from Thread#name uncovered by the DynamicScope changes.

[...]

PR #4168, opened 2016-09-22
Add paths to *DynamicScope that don't check nulls, offsets.

headius:
Our IR performes liveness checks on variables and should never
emit code that accesses variable indices outside the prescribed
scope size. This patch adds new getters that do not perform these
unnecessary checks, and wires up the JIT to call them.

This patch does not modify Full interpretation to use these
unchecked paths, because it would require either a boolean in
LocalVariaable operand and LoadLocalVariable instruction or a
pass to replace them with "Unchecked" versions.

It would also be better if we could generate these shapes of
DynamicScope on the fly, rather than maintaining a finite set of
mostly-identical subclasses.

See #4167.


headius:
The one JIT failure appears to be in this method; the "object" variable here is coming back as null. Why?

lib/ruby/stdlib/test/unit/assertions.rb:393
```ruby
      def assert_nil(object, message="")
        full_message = build_message(message, <<EOT, object)
<?> expected to be nil.
EOT
        assert_block(full_message) { object.nil? } # this nil? call fails
      end
```

headius:
Ok I think the error is because Thread#name is not initialized to any value. It's returning nulls into Ruby!

Pushing a fix shortly.

headius:
At least the Thread#name fix needs to get into 9.1.6.0.

PR #4170, opened 2016-09-22
Bring BigMath::log behaviour closer to MRI

### LINKED ISSUE #4158: BigMath.log(Rational,...) produces garbage digits within the requested precision
```
for ruby_implementation in jruby-9.1.5.0 ruby-2.3.1; do rvm "$ruby_implementation" do ruby -e 'require "bigdecimal"; puts("#{"%5s" % RUBY_ENGINE}: #{BigMath.log(Rational(1234567890,987654321),50)}")'; done
jruby: 0.223143542201713714247503163075284070668726387597801926772700265233450212669566607742943532376739080993640224513865907256089370906573E0
 ruby: 0.22314354220170971436137296411949880462556361100856391620766259404746040597133837784E0 
```

### Environment

JRuby 9.1.5.0 on OpenJDK 1.8.0_101 on x86-64

### Expected Behavior

`BigMath.log(Rational(1234567890,987654321),50).to_s` should return `"0.22314354220170971436137296411949880462556361100[...]"` (it should be correct in the first 50 digits)

### Actual Behavior

`BigMath.log(Rational(1234567890,987654321),50).to_s` currently returns`"0.22314354220171371424750316307528407066872638759[...]"` (it is incorrect after the first 14 digits)

### Additional Notes

MRI seems to perform the division of the numerator with the denominator with the requested precision (50 decimal digits in the example), and then it performs the logarithm with the requested precision.

JRuby seems to perform the division only with double precision (about 14 decimal digits, which is far lower than the requested precision), and then it performs the logarithm with the requested precision. If this is the case, then JRuby should not call ".to_f", but ".to_d(precision)" on the argument to BigDecimal.log.

shirshendu:
Especially for a Rational, high precision use-case

Fixes #4158

### DIFF
--- a/test/mri/bigdecimal/test_bigdecimal.rb
+++ b/test/mri/bigdecimal/test_bigdecimal.rb
@@ -1497,6 +1497,13 @@ class TestBigDecimal < Test::Unit::TestCase
     assert_in_delta(3, log_3, 0.0000000000_0000000000_0000000000_0000000000_0000000001)
   end
 
+  def test_BigMath_log_with_rational_high_precision_case
+    # From MRI
+    result = BigDecimal('0.22314354220170971436137296411949880462556361100856391620766259404746040597133837784E0')
+    r = Rational(1_234_567_890, 987_654_321)
+    assert_in_delta(result, BigMath.log(r, 50), 0.0000000000_0000000000_0000000000_0000000000_0000000001)
+  end
+
   def test_BigMath_log_with_42
     assert_in_delta(Math.log(42), BigMath.log(42, 20))
     assert_in_delta(Math.log(42), BigMath.log(42.0, 20))
--- a/core/src/main/ruby/jruby/bigdecimal.rb
+++ b/core/src/main/ruby/jruby/bigdecimal.rb
@@ -28,7 +28,7 @@ module BigMath
       return BigDecimal::INFINITY if x.infinite?
       return BigDecimal::NAN if x.nan?
     end
-    x = x.is_a?(Rational) ? x.to_d(Float::DIG) : x.to_d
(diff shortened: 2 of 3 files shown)


shirshendu:
Just for convenience, the command for running the newly added test case:
`bin/jruby test/mri/bigdecimal/test_bigdecimal.rb -n test_BigMath_log_with_rational_high_precision_case`

test/mri/bigdecimal/test_bigdecimal.rb
```diff
@@ -1497,6 +1497,13 @@ def test_BigMath_log_with_high_precision_case
     assert_in_delta(3, log_3, 0.0000000000_0000000000_0000000000_0000000000_0000000001)
   end
 
+  def test_BigMath_log_with_rational_high_precision_case
+    # From MRI
+    result = BigDecimal('0.22314354220170971436137296411949880462556361100856391620766259404746040597133837784E0')
+    r = Rational(1_234_567_890, 987_654_321)
+    assert_in_delta(result, BigMath.log(r, 50), 0.0000000000_0000000000_0000000000_0000000000_0000000001)
+  end
+
```
headius: We don't usually edit the MRI tests since there's no good process for upstreaming those changes. I'll move this to spec/ruby.

headius: We don't usually edit the MRI tests since there's no good process for upstreaming those changes. I'll move this to spec/ruby.

headius:
Merged your fix in 68552a398b229cef15561dd001951642b68eb720 and the spec/ruby change in 84a5f780a07576509e73467321fe775fb9c0ee77. Thanks!

PR #4174, opened 2016-09-23
Remove premade field-based DynamicScope and generate instead.

headius:
Instead of having hand-written DynamicScope subtypes that use
fields instead of an array, this logic generates those subclasses
as needed.

Currently it only overrides the base getValue and setValue methods
and DynamicScope now implements the other previously-abstract
methods by calling those two. This is not optimal for generic
access via those indexed methods, but the expectation is that JIT
can just go after the fields directly. The additional overrides
would likely be useful for the interpreter, though.

See #4167.


headius:
It might also be possible to implement getValue and setValue using Unsafe rather than switches, since Unsafe could provide direct access to the field without any branching logic.

headius:
I did a prototype of directly accessing fields from the specialized scopes. It worked reasonably well, but there's a few snags. It does not boot properly (because I removed ManyVarsDynamicScope and other classes that are actually still needed), but it does generate the expected code.

https://gist.github.com/headius/2be4be6b1b8bc71f714901519ee6b5b0

1. Classloading is complicated. JIT code needs to load into a classloader that can see all scopes that code will try to access. Since we can't have multiple one-shot classloaders as parents, we would need to put all specialized scopes into a common classloader. This in turn keeps them from GCing. That may not be a huge issue, though...a given application will always have a finite, probably-small set of scope sizes.
2. In order to access the fields, all accesses must checkcast. Unknown what impact this has on performance yet. This could be mitigated by knowing the scope types for all accessed variables and doing the cast once. It can also be statically created with the right type if it is the depth zero scope, negating the need for casts.

The comparative bytecode is here: https://gist.github.com/headius/2c0b2459d8e5fb3c69bbaec254295ea8

This should have better performance than any previous mechanism, unless the checkcast is killing us. Just an experiment for now.

headius:
@subbuss @enebo @kares I'm pretty much done with this PR and would like input and review.

The concerns I had/have are:

* Generating too many classes. This is mitigated by a limit of 50 in my most recent commit.
* Cost of construction. This has been mitigated by using MethodHandle instead of reflection and caching that handle in StaticScope, so there's fewer steps from newDynamicScope to having an object in hand.
* Caching classes statically. This is still done, but since there will never be more than 50 within the classloader that booted JRuby, I am not worried.

The biggest improvement in this PR from beginning to end is that there are now ten offset-specialized get/set methods that are generated for all specialized DynScopes, and the JIT will automatically use them. Adding more requires only implementing them in DynamicScope and ManyVarsDynamicScope and adding their names to the lists in DynamicScopeGenerator. These methods allow heap variable accesses to inline down to a single field read or write with no extra overhead.

Inlining-wise at the JVM level, a given jitted code body will generally have only one StaticScope and will only ever see one DynamicScope type, so every method body should be able to inline logic for its own scope.

I have not revisited the direct field access because I was able to expand the use of offset-specialized methods. However, it might be interesting to see indy-based scope var access, which could specialize beyond the base set of offsets. I have no plans to do this right now.

headius:
Oh, this is also partially done on master too, but there's only two hand-written dynamic scope classes now: DynamicScope and ManyVarsDynamicScope.

core/src/main/java/org/jruby/ir/targets/JVMVisitor.java
```diff
@@ -495,6 +433,31 @@ private void jvmStoreLocal(Runnable source, Variable variable) {
...
     }
 
+    private void genSetValue(LocalVariable localvariable) {
+        int depth = localvariable.getScopeDepth();
+        int location = localvariable.getLocation();
+
+        String baseName = p(DynamicScope.class);
+
+        if (depth == 0) {
+
```
enebo: extra newlines in this method

core/src/main/java/org/jruby/runtime/scope/ManyVarsDynamicScope.java
```diff
@@ -43,6 +53,7 @@ private void allocate() {
     }
     
     public DynamicScope cloneScope() {
```
enebo: We can deprecate all cloneScopes since I do not see them used anywhere (even without this PR)

headius: Oh, very nice. Ok.

core/src/main/java/org/jruby/ir/targets/JVMVisitor.java
headius: The newlines existed before the PR too. May be unnecessary with the simpler body, though.

PR #4175, opened 2016-09-23
Fix #4116. Make dirname aware of the file system's path separator

### LINKED ISSUE #4116: File.dirname interprets backslashes as path separators on Mac OS X
### Environment

```
~/> ruby -v
jruby 1.7.21 (1.9.3p551) 2015-07-07 a741a82 on Java HotSpot(TM) 64-Bit Server VM 1.8.0_31-b13 +jit [darwin-x86_64]
~> uname -a
Darwin 15.6.0 Darwin Kernel Version 15.6.0: Thu Jun 23 18:25:34 PDT 2016; root:xnu-3248.60.10~1/RELEASE_X86_64 x86_64
```
### Expected Behavior

```
~> ruby -v
ruby 1.9.3p551 (2014-11-13 revision 48407) [x86_64-darwin15.6.0]
~> irb
irb(main):001:0> path = 'this/path/has/filename\that\contains\backslashes'
=> "this/path/has/filename\\that\\contains\\backslashes"
irb(main):002:0> File.dirname path
=> "this/path/has"
irb(main):003:0>
```

```
~> ruby -v
ruby 2.2.4p230 (2015-12-16 revision 53155) [x86_64-darwin15]
~> irb
irb(main):001:0> path = 'this/path/has/filename\that\contains\backslashes'
=> "this/path/has/filename\\that\\contains\\backslashes"
irb(main):002:0> File.dirname path
=> "this/path/has"
irb(main):003:0>
```
### Actual Behavior

```
~> ruby -v
jruby 1.7.21 (1.9.3p551) 2015-07-07 a741a82 on Java HotSpot(TM) 64-Bit Server VM 1.8.0_31-b13 +jit [darwin-x86_64]
~> irb
irb(main):001:0> path = 'this/path/has/filename\that\contains\backslashes'
=> "this/path/has/filename\\that\\contains\\backslashes"
irb(main):002:0> File.dirname path
=> "this/path/has/filename\\that\\contains"
irb(main):003:0>
```

```
~> ruby -v
jruby 9.1.2.0 (2.3.0) 2016-05-26 7357c8f Java HotSpot(TM) 64-Bit Server VM 25.31-b07 on 1.8.0_31-b13 +jit [darwin-x86_64]
~> irb
irb(main):001:0>  path = 'this/path/has/filename\that\contains\backslashes'
=> "this/path/has/filename\\that\\contains\\backslashes"
irb(main):002:0>  File.dirname path
=> "this/path/has/filename\\that\\contains"
irb(main):003:0>
```
### Additional Information

The code in [RubyFile.java](https://github.com/jruby/jruby/blob/master/core/src/main/java/org/jruby/RubyFile.java) uses hard-coded path separator values, e.g. line 724 `if (endChar == '/' || endChar == '\\') {`, rather than system dependent constants, such as Java `File.separator`.

jsyeo:
I removed the line that normalizes the path to be separated by `/` and instead make it aware of the file system's separator.

### DIFF
--- a/core/src/main/java/org/jruby/RubyFile.java
+++ b/core/src/main/java/org/jruby/RubyFile.java
@@ -651,38 +651,37 @@ public class RubyFile extends RubyIO implements EncodingCapable {
 
     static Pattern PROTOCOL_PATTERN = Pattern.compile(URI_PREFIX_STRING + ".*");
     public static String dirname(ThreadContext context, String jfilename) {
-        String name = jfilename.replace('\\', '/');
         int minPathLength = 1;
         boolean trimmedSlashes = false;
 
-        boolean startsWithDriveLetterOnWindows = startsWithDriveLetterOnWindows(name);
+        boolean startsWithDriveLetterOnWindows = startsWithDriveLetterOnWindows(jfilename);
 
         if (startsWithDriveLetterOnWindows) {
             minPathLength = 3;
         }
 
         // jar like paths
-        if (name.contains(".jar!/")) {
-            int start = name.indexOf("!/") + 1;
-            String path = dirname(context, name.substring(start));
+        if (jfilename.contains(".jar!/")) {
+            int start = jfilename.indexOf("!/") + 1;
+            String path = dirname(context, jfilename.substring(start));
             if (path.equals(".") || path.equals("/")) path = "";
-            return name.substring(0, start) + path;
+            return jfilename.substring(0, start) + path;
         }
         // address all the url like paths first
-        if (PROTOCOL_PATTERN.matcher(name).matches()) {
-            int start = name.indexOf(":/") + 2;
-            String path = dirname(context, name.substring(start));
+        if (PROTOCOL_PATTERN.matcher(jfilename).matches()) {
+            int start = jfilename.indexOf(":/") + 2;
+            String path = dirname(context, jfilename.substring(start));
             if (path.equals(".")) path = "";
-            return name.substring(0, start) + path;
+            return jfilename.substring(0, start) + path;
         }
 
-        while (name.length() > minPathLength && name.charAt(name.length() - 1) == '/') {
+        while (jfilename.length() > minPathLength && jfilename.charAt(jfilename.length() - 1) == '/') {
             trimmedSlashes = true;
-            name = name.substring(0, name.length() - 1);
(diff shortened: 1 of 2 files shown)


enebo:
@jsyeo This change will break windows compatibility but fix non-windows.  On windows both / and \\ will work in dirname (it is how MRI works anyways) whereas on all other platforms it treats \ as literalizing character.  So I think maybe the fix might be to just put an if check on the original gsub to not convert \\ to / if only on windows.

headius:
@enebo I don't think this will break anything on Windows, since the backslashes are not going to be interpreted as escapes (they're literally in the string, not syntactically inside a literal string).

However it may break other things by returning backslashes in the resulting path. I think MRI normalizes everything to / internally, right? Does any core MRI File method return a backslashy path on Windows?

@jsyeo Thank you for the PR! Are you on Windows or a UNIX system?

jsyeo:
> @jsyeo Thank you for the PR! Are you on Windows or a UNIX system?

I'm on an unix system. Hmm, I kinda assumed that it would be the inverse on a windows system. Let me spin up a windows vm and install mri to try out its `File.dirname` behavior.

jellymann:
The MRI implementation does pretty much what it says in the documentation:

>Returns all components of the filename given in _file_name_ except the last one. The filename can be formed using both `File::SEPARATOR` and `File::ALT_SEPARATOR` as the separator when `File::ALT_SEPARATOR` is not `nil`.

`File::SEPARATOR` is always `/`. However, `File::ALT_SEPARATOR` is `nil` on Unix and `\` on Windows. The net result is that on Windows, either separator can be used interchangeably, but on Unix, only `/` is recognised.

I started working on a solution to #4116 for a while now, but I couldn't test on windows since I had some issues with Maven on my VM. Now that they've been fixed I've managed to put together what I believe to be a solid fix for both Unix and Windows environments. You can examine my proposed changes here: #4177.

enebo:
@jellymann thanks for checking on the details of this.  You gave a clearer explanation than me of the issue.  I see there is a new PR to fix this so we will likely have a golden PR now :)

@headius I only pointed this out since I tested it on windows and it returns using either \ and / as a delimeter (which fits @jellymann desciption as to why).  I think we can maybe even make a nice test for this in spec which adds ALT_SEPARATOR so we can test windows behavior on non-windows ci runs too.

Duplcating as there is a new PR now.

PR #4197, opened 2016-10-01
Generify accept methods in AST

jsyeo:
Some of the AST's accept methods were returning `Object` instead of the generic type.


enebo:
@jsyeo If I have compiled with the previous visitor API will this still run with this generic version once compiled?  Since JVM does type erasure I think it will but I am not positive.  Do you know?  I consider AST an internal API more or less for jruby proper but I still like to avoid compat errors if possible.

headius:
I believe it still looks identical (returning Object) if it claims to return T.

jsyeo:
> If I have compiled with the previous visitor API will this still run with this generic version once compiled?

Yes. I have tested it.

PR #4201, opened 2016-10-03
[Truffle] Iterative implementations for getByteSlow

bjfish:


truffle/src/main/java/org/jruby/truffle/core/rope/SubstringRope.java
```diff
@@ -62,7 +62,13 @@ public Rope withEncoding(Encoding newEncoding, CodeRange newCodeRange) {
 
     @Override
     public byte getByteSlow(int index) {
-        return child.getByteSlow(index + offset);
+        int idx = index + offset;
+        Rope nextChild = child;
+        while (nextChild instanceof SubstringRope) {
```
nirvdrum: While there's nothing wrong with this check, when we actually create `SubstringRope` instances we always collapse substrings of substrings (see https://github.com/jruby/jruby/blob/5c97ec03aefe6671cb7a2ad61b88a50bf3319b33/truffle/src/main/java/org/jruby/truffle/core/rope/RopeNodes.java#L119-L124 for details).

And since you only handle that one case here, I think whatever you were looking to accomplish devolves into the code you replaced.

eregon:
So this is just avoiding recursive calls in case of largely homegeneous ropes, right?
Is it to avoid too deep recursions (which could stack overflow) or is it for performance.
If the latter, do you have numbers?

bjfish:
@eregon Yes, I did this to avoid stack overflow error. For example, it fixes stack overflow for:
`a = "a"; 100000.times { a << "b" }; a[0]`

I'm not sure about performance.

bjfish:
Going to decline this for now because: it is not a complete solution, code in this area is expected to change soon, and it is not currently needed for gem installing with a workaround in place.

PR #4207, opened 2016-10-06
Flush destination IO in IO.copy_stream

grddev:
Previously, any buffered IO was ignored when copying to the destination stream.

Untagged a couple of specs that are no longer failing with the additional flush.

Tested this against 9.1.5.0, 9.0.5.0, 1.7.26, and 1.7.19, and buffered IO was ignored in all instances.


headius:
Very nice, thanks!

PR #4214, opened 2016-10-09
[Truffle] Use the new getgroups POSIX call.

chrisseaton:
Depends on https://github.com/jnr/jnr-posix/pull/82.

So far I'm just using this for Truffle, but that shows it works. I'm not sure what priority you want to give using your Java method for this compared to the JNR version in JRuby classic.


chrisseaton:
Travis failures are because the JNR snapshot isn't published.

PR #4216, opened 2016-10-10
UDP socket improvements

etehtsea:


etehtsea:
Ok, there is a regression that need to be fixed.

etehtsea:
Ready for reviewing (if specs will pass).

headius:
Thanks again :-)

PR #4221, opened 2016-10-11
UNIX sockets improvements

etehtsea:


enebo:
@etehtsea I guess landing this caused: https://travis-ci.org/jruby/jruby/jobs/167369337.  Any chance you can look into it?  I would guess we are getting random chars back and we should not be looking at them based on a return status?  Just a guess...

etehtsea:
@enebo I'll try to look into it this week.

PR #4224, opened 2016-10-12
Fix the incorrect opscode mapping in f2i()

himosqui:
- f2i() should be mapped to F2I, instead of F2D. Note that F2I is a valid constant in org.objectweb.asm.Opcodes.
 - The incorrect mapping appears to be a copy-paste editing error from the above f2d() method.
 - Due to the error, when float-to-integer narrowing numeric conversion should occur, float-to-double widening numeric conversion was being done, which can have a negative impact on performance and memory usage.


kares:
najs :scissors: find!

PR #4225, opened 2016-10-13
Don't dup every element in Enumerator#drop (9K)

### LINKED ISSUE #4218: Enumerable#drop calls #dup on every element
Given:

``` ruby
class Stuff
  include Enumerable

  def initialize(stuff)
    @stuff = stuff
  end

  def each(&block)
    @stuff.each(&block)
  end
end

original_stuff = [Object.new, Object.new, Object.new]
enumerable_stuff = Stuff.new(original_stuff)
```
### Expected Behavior

``` ruby
original_stuff[1].equal?(enumerable_stuff.to_a[1]) # => true
original_stuff[1].equal?(enumerable_stuff.drop(1).first) # => true
enumerable_stuff.to_a[1].equal?(enumerable_stuff.drop(1).first) # => true
```

After dropping an element from an `Enumerable` I expect that the remaining elements are _the same_ as in the original `Enumerable`. Calling `#drop` should not modify the elements.
### Actual Behavior

``` ruby
original_stuff[1].equal?(enumerable_stuff.to_a[1]) # => true
original_stuff[1].equal?(enumerable_stuff.drop(1).first) # => false
enumerable_stuff.to_a[1].equal?(enumerable_stuff.drop(1).first) # => false
```

Calling `#drop` on an `Enumerable` calls `#dup` on every non-dropped element. Besides the side-effect illustrated above it also expensive to do all that copying, and it doesn't work with objects whose `#dup` doesn't behave well, or objects that can't be `#dup`'ed.

I discovered this while writing a native extension with a class that was not allocatable. It took me a while to understand why my code failed with errors saying that my class wasn't allocatable – I wasn't trying to create instances.

It looks like this behaviour was introduced in https://github.com/jruby/jruby/commit/06f044124bef7856340b2bb7bdad6bac7d68f771, but it seems like the fix for the problem described in the commit message was applied in the wrong place (it ended up in `Enumerable` instead of `Enumerator`).
### Environment
- jruby 9.1.5.0 (2.3.1) 2016-09-07 036ce39 Java HotSpot(TM) 64-Bit Server VM 25.0-b70 on 1.8.0-b132 +jit [darwin-x86_64]
- jruby 1.7.25 (1.9.3p551) 2016-04-13 867cb81 on Java HotSpot(TM) 64-Bit Server VM 1.8.0-b132 +jit [darwin-x86_64]

iconara:
This fixes #4218 in JRuby 9K.

In 06f044124bef7856340b2bb7bdad6bac7d68f771 (JRUBY-6892) `Enumerable#drop` was changed to call `#dup` on every element that ended up in the result, but that might not work. Not all objects correctly implement `#dup`, or are even allocatable.

`#take` was not changed, and looking at the differences between `#take` and `#drop` the only thing that stood out was that the signature of the #each call was different. Changing from `Signature.NO_ARGUMENTS` to `Signature.ONE_REQUIRED` make the issue that 06f044124bef7856340b2bb7bdad6bac7d68f771 tried to fix go away.

### DIFF
--- a/core/src/main/java/org/jruby/RubyEnumerable.java
+++ b/core/src/main/java/org/jruby/RubyEnumerable.java
@@ -355,16 +355,13 @@ public class RubyEnumerable {
         final RubyArray result = runtime.newArray();
 
         try {
-            each(context, self, new JavaInternalBlockBody(runtime, Signature.NO_ARGUMENTS) {
+            each(context, self, new JavaInternalBlockBody(runtime, Signature.ONE_REQUIRED) {
                 long i = len; // Atomic ?
                 public IRubyObject yield(ThreadContext context, IRubyObject[] args) {
-                    IRubyObject packedArg = packEnumValues(context.runtime, args);
                     synchronized (result) {
                         if (i == 0) {
-                            // While iterating over an RubyEnumerator, "arg"
-                            // gets overwritten by the new value, leading to JRUBY-6892.
-                            // So call .dup() whenever appropriate.
-                            result.append(packedArg.isImmediate() ? packedArg : packedArg.dup());
+                            IRubyObject packedArg = packEnumValues(context.runtime, args);
+                            result.append(packedArg);
                         } else {
                             --i;
                         }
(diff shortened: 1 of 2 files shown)


headius:
Our CI went red for a while, so I pulled locally and ran a couple suites. Things look good. Merged.

PR #4230, opened 2016-10-17
[Truffle] Add bundler workarounds files, jt test bundle command

bjfish:
So far this works for `bundle install` and `bundle exec rake` for the gem mentioned in the changes: `hike`.  I'm still working on getting this working with an example gem that has a native build.

Please let me know if you have any feedback.

### DIFF
--- a/lib/ruby/truffle/truffle/bundler-workarounds.rb
+++ b/lib/ruby/truffle/truffle/bundler-workarounds.rb
@@ -1,0 +1,582 @@
+$quiet_workaround = false
+
+workaround_header = <<-HEREDOC
+==========================================
+WORKING AROUND BUNDLE INSTALL ISSUES
+Instructions:
+  1. Run bundle install command with workarounds. E.g.:
+      GEM_HOME=/Users/brandonfish/Documents/truffle_gem_home ../../jruby/bin/jruby -X+T -rbundler-workarounds -S bundle install
+  2. Try running gem tests, e.g.:
+      GEM_HOME=/Users/brandonfish/Documents/truffle_gem_home ../../jruby/bin/jruby -X+T -rbundler-workarounds -S bundle exec rake
+HEREDOC
+puts workaround_header unless $quiet_workaround
+
+
+
+have_extensions = <<-HEREDOC
+==========================================
+Workaround: Add nil check Gem::BasicSpecification.have_extensions?
+Error:
+  undefined method `empty?' for nil:NilClass
+Called here:
+  lib/ruby/stdlib/rubygems/basic_specification.rb:313
+HEREDOC
+puts have_extensions unless $quiet_workaround
+class Gem::BasicSpecification
+  def have_extensions?; !extensions.nil? && !extensions.empty?; end
+end
+
+
+remote_fetcher = <<-HEREDOC
+==========================================
+Workaround: Use wget in remote fetcher
+HEREDOC
+puts remote_fetcher unless $quiet_workaround
+
+require 'rubygems/remote_fetcher'
+class Gem::RemoteFetcher
+  def download(spec, source_uri, install_dir = Gem.dir)
+    cache_dir =
+        if Dir.pwd == install_dir then # see fetch_command
+          install_dir
+        elsif File.writable? install_dir then
+          File.join install_dir, "cache"
+        else
+          File.join Gem.user_dir, "cache"
+        end
+
+    gem_file_name = File.basename spec.cache_file
+    local_gem_path = File.join cache_dir, gem_file_name
+
+    FileUtils.mkdir_p cache_dir rescue nil unless File.exist? cache_dir
+
+    # Always escape URI's to deal with potential spaces and such
+    # It should also be considered that source_uri may already be
+    # a valid URI with escaped characters. e.g. "{DESede}" is encoded
+    # as "%7BDESede%7D". If this is escaped again the percentage
+    # symbols will be escaped.
+    unless source_uri.is_a?(URI::Generic)
+      begin
+        source_uri = URI.parse(source_uri)
+      rescue
+        source_uri = URI.parse(URI.const_defined?(:DEFAULT_PARSER) ?
+                                   URI::DEFAULT_PARSER.escape(source_uri.to_s) :
+                                   URI.escape(source_uri.to_s))
+      end
+    end
+
+    scheme = source_uri.scheme
+
+    # URI.parse gets confused by MS Windows paths with forward slashes.
+    scheme = nil if scheme =~ /^[a-z]$/i
+
+    # REFACTOR: split this up and dispatch on scheme (eg download_http)
+    # REFACTOR: be sure to clean up fake fetcher when you do this... cleaner
+    case scheme
+      when 'http', 'https', 's3' then
+        unless File.exist? local_gem_path then
+          begin
+            verbose "Downloading gem #{gem_file_name}"
+
+            remote_gem_path = source_uri + "gems/#{gem_file_name}"
+
+            # WORKAROUND STARTS HERE
+            cmd = "wget #{remote_gem_path} -O #{local_gem_path}"
+            `#{cmd}`
+              # self.cache_update_path remote_gem_path, local_gem_path
+            # WORKAROUND ENDS HERE
+          rescue Gem::RemoteFetcher::FetchError
+            raise if spec.original_platform == spec.platform
+
+            alternate_name = "#{spec.original_name}.gem"
+
+            verbose "Failed, downloading gem #{alternate_name}"
+
+            remote_gem_path = source_uri + "gems/#{alternate_name}"
+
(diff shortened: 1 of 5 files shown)


truffle/src/main/java/org/jruby/truffle/core/kernel/KernelNodes.java
```diff
@@ -710,12 +714,21 @@ public Object exec(VirtualFrame frame, Object command, Object[] args) {
 
         @TruffleBoundary
         private String[] buildCommandLine(Object command, Object[] args) {
-            final String[] commandLine = new String[1 + args.length];
-            commandLine[0] = command.toString();
+            final List<String> commandLine = new ArrayList<>(1 + args.length);
+            if (RubyGuards.isRubyArray(command)) {
+                final Object[] store = (Object[]) getStore((DynamicObject) command);
+                commandLine.add(store[0].toString());
```
eregon: Why only the first element is considered here?

tool/jt.rb
```diff
@@ -954,6 +956,42 @@ def test_ecosystem(env={}, *args)
   end
   private :test_ecosystem
 
+  def test_bundle(*args)
+    gems = [{:url => "https://github.com/sstephenson/hike", :commit => "3abf0b3feb47c26911f8cedf2cd409471fd26da1"}]
+    gems.each do |gem|
+      gem_url = gem[:url]
+      name = gem_url.split('/')[-1]
+      require 'tmpdir'
+      temp_dir = Dir.mktmpdir(name)
```
eregon: There is a block form for `Dir.mktmpdir`

lib/ruby/truffle/truffle/bundler-workarounds.rb
```diff
@@ -0,0 +1,516 @@
+$quiet_workaround = false
+
+workaround_header = <<~HEREDOC
```
eregon: Squiggly heredoc is a very new feature, maybe it's better to stick to `<<-` here, since anyway the indentation is not needed.

truffle/src/main/java/org/jruby/truffle/core/kernel/KernelNodes.java
bjfish: I'm not sure how to implement the `argv0` yet which is mentioned in the documentation:
> second argument is used as the argv[0] value, which may show up in process listings.
> exec([cmdname, argv0], arg1, ...)
https://ruby-doc.org/core-2.3.1/Kernel.html#method-i-exec

Only implementing the cmdname was enough to get `bundle` working for now.

tool/jt.rb
bjfish: In the block form, the temp directory is deleted at the end of the block. For testing, I want to frequently disable this by commenting out `# FileUtils.remove_entry temp_dir` to inspect the results. Also, in the future, I probably want to change this to not delete the temp directory whenever the test fails.

truffle/src/main/java/org/jruby/truffle/core/kernel/KernelNodes.java
eregon: I see, please mention the `exec([cmdname, argv0], arg1, ...)` there to clarify the meaning.

bjfish: I've added some comments to `exec` for `argv0`

lib/ruby/truffle/truffle/bundler-workarounds.rb
bjfish: I've update these to `<--`

bjfish:
I also updated backtick input stream reading because I had issues with:
```
deps = `curl https://rubygems.org/api/v1/dependencies?gems=bundler`
puts "#{Marshal.load(deps)}"
```

lib/ruby/truffle/truffle/bundler-workarounds.rb
```diff
@@ -0,0 +1,572 @@
...
+        response = @fetcher.call(remote_path, headers)
+        return if response.is_a?(Net::HTTPNotModified)
+
+        content = response.body
+
+        if response["Content-Encoding"] == "gzip"
+          IO.binwrite("#{local_temp_dir}/gzipped_versions", content)
+
+          content = Zlib::GzipReader.new(StringIO.new(content)).read
+          #content = `gunzip -c #{local_temp_dir}/gzipped_versions`
```
pitr-ch: Does it suppose to be commented?

lib/ruby/truffle/truffle/bundler-workarounds.rb
```diff
@@ -0,0 +1,572 @@
...
+      ==========================================
+      Workaround: Stub OpenSSL::SSL::VERIFY_NONE
+      Error:
+        uninitialized constant OpenSSL::SSL::VERIFY_NONE
+      Called here:
+        bundler/vendor/net/http/persistent.rb:1142
+  HEREDOC
+  puts verify_none unless $quiet_workaround
+  module SSL
+    VERIFY_NONE = 1
```
eregon: Could you give those constants different values? Otherwise it might have unintended side effects.

tool/jt.rb
```diff
@@ -954,6 +956,43 @@ def test_ecosystem(env={}, *args)
...
+            raw_sh(*['git', 'checkout', gem[:commit]])
+          end
+          run({'GEM_HOME' => gem_home}, *["-rbundler-workarounds", "-S", "gem", "install", "bundler", "-v","1.12.5"])
+          run({'GEM_HOME' => gem_home}, *["-rbundler-workarounds", "-S", "bundle", "install"])
+          # Need to workaround ruby_install name `jruby-truffle` in the gems binstubs (command can't be found )
+          # or figure out how to get it on the path to get `bundle exec rake` working
+          #run({'GEM_HOME' => gem_home}, *["-rbundler-workarounds", "-S", "bundle", "exec", "rake"])
+        end
+      ensure
+          FileUtils.remove_entry temp_dir
```
eregon: This is misaligned.

lib/ruby/truffle/truffle/bundler-workarounds.rb
bjfish: Thanks, i've uncommented/commented this

lib/ruby/truffle/truffle/bundler-workarounds.rb
bjfish: I've updated this to `VERIFY_NONE = 0`

tool/jt.rb
bjfish: Fix this alignment

tool/jt.rb
```diff
@@ -954,6 +956,43 @@ def test_ecosystem(env={}, *args)
...
+                     puts "Using temporary GEM_HOME:#{tmp_home}"
+                     tmp_home
+                   end
+        Dir.chdir(gem_dir) do
+          if gem.has_key?(:commit)
+            raw_sh(*['git', 'checkout', gem[:commit]])
+          end
+          run({'GEM_HOME' => gem_home}, *["-rbundler-workarounds", "-S", "gem", "install", "bundler", "-v","1.12.5"])
+          run({'GEM_HOME' => gem_home}, *["-rbundler-workarounds", "-S", "bundle", "install"])
+          # Need to workaround ruby_install name `jruby-truffle` in the gems binstubs (command can't be found )
```
eregon: Isn't it a full path based on bindir + ruby_install_name?
There is `bin/jruby-truffle`.

eregon:
Let's get this merged!
Obviously, we should focus or removing these workarounds progressively.

lib/ruby/truffle/truffle/bundler-workarounds.rb
```diff
@@ -0,0 +1,582 @@
...
+      end
+    end
+  end
+end
+
+resolver_search_for = <<-HEREDOC
+==========================================
+Workaround: Change =~ to == for resolver#search_for
+Error:  type mismatch: String given
+        stdlib/rubygems/resolver.rb:237:in `block in search_for'
```
eregon: How does it work in MRI?
MRI also gives an error for String =~ String.

tool/jt.rb
bjfish: I get the following error which I think is caused by the gem bin file (rake) having the header `#!/usr/bin/env jruby-truffle`

```
rbenv: jruby-truffle: command not found

The `jruby-truffle' command exists in these Ruby versions:
  jruby-local

```

I'm not sure how to fix this yet.

lib/ruby/truffle/truffle/bundler-workarounds.rb
bjfish: I didn't fully figure out how this works but I think it originates from `Gem::Platform` in `rubygems/platform`.

It looks like `Gem::Platform.local` can return a `String` or `Gem::Platform` in some cases. If it returned a `Gem::Platform` then it has it's own `Gem::Platform#=~` implementation to use.

PR #4250, opened 2016-10-26
Prepend must set method's "defined class" to the new bottom.

headius:
See freerange/mocha#274.


headius:
Note that I could not find any code in MRI that does this method-table-walking, so I'm confused how they realign methods to the new hierarchy. This appears to pass specs, but travis is a bit red right now.

headius:
I'm going to merge this. The method-walking already was occurring and this just adds an additional bit of state.

I believe MRI does not need to walk the method table because their method structs do not contain the originating class. Instead, when they search the method table, they produce a separate struct specific to that method table and its host class. Basically, they lazily determine what we store as implClass based on where the method is actually found at runtime.

We may want to adopt this approach in the future. I believe this would help fix the problems we have with method table cloning + super going up the wrong hierarchy. cc @enebo @subbuss.

PR #4252, opened 2016-10-26
Also perform uninitialized check on temp locals.

### LINKED ISSUE #4251: nil initialization missing from optionally-set argument
### Environment

Current JRuby master (pre-9.1.6.0)
### Expected Behavior

This script should work:

``` ruby
def foo(a=(b=1))
  [a,b]
end

foo
```
### Actual Behavior

The jitted code fails to verify because the load of `b` happens before it has been initialized. It should have been set to nil somewhere before that.

This is currently causing CI of `rake spec:compiler` to fail.

```
$ jruby -Xjit.logging -Xjit.logging.verbose -e 'def foo(a=(b=1));[a,b];end;foo'
2016-10-26T18:01:42.934-05:00 [main] ERROR Ruby : failed to compile target script -e: 
org.jruby.compiler.NotCompilableException: failed to compile script -e
    at org.jruby.ir.Compiler.execute(Compiler.java:62)
    at org.jruby.ir.Compiler.execute(Compiler.java:31)
    at org.jruby.ir.IRTranslator.execute(IRTranslator.java:42)
    at org.jruby.Ruby.tryCompile(Ruby.java:823)
    at org.jruby.Ruby.precompileCLI(Ruby.java:788)
    at org.jruby.Ruby.runNormally(Ruby.java:750)
    at org.jruby.Ruby.runNormally(Ruby.java:779)
    at org.jruby.Ruby.runFromMain(Ruby.java:592)
    at org.jruby.Main.doRunFromMain(Main.java:425)
    at org.jruby.Main.internalRun(Main.java:313)
    at org.jruby.Main.run(Main.java:242)
    at org.jruby.Main.main(Main.java:204)
Caused by: java.lang.VerifyError: Bad local variable type
Exception Details:
  Location:
    DashE.RUBY$method$foo$0(Lorg/jruby/runtime/ThreadContext;Lorg/jruby/parser/StaticScope;Lorg/jruby/runtime/builtin/IRubyObject;[Lorg/jruby/runtime/builtin/IRubyObject;Lorg/jruby/runtime/Block;Lorg/jruby/RubyModule;Ljava/lang/String;)Lorg/jruby/runtime/builtin/IRubyObject; @59: aload
  Reason:
    Type top (current frame, locals[9]) is not assignable to reference type
  Current Frame:
    bci: @59
    flags: { }
    locals: { 'org/jruby/runtime/ThreadContext', 'org/jruby/parser/StaticScope', 'org/jruby/runtime/builtin/IRubyObject', '[Lorg/jruby/runtime/builtin/IRubyObject;', 'org/jruby/runtime/Block', 'org/jruby/RubyModule', 'java/lang/String', top, 'org/jruby/runtime/builtin/IRubyObject' }
    stack: { 'org/jruby/runtime/ThreadContext', 'org/jruby/runtime/builtin/IRubyObject' }
  Bytecode:
    0x0000000: 0000 2a2b 2d12 1212 1312 1212 1212 1401
    0x0000010: b800 1a2d 0303 0312 12b8 001e 3a08 2a19
    0x0000020: 08b2 0024 b800 2a9a 0010 002a b800 3f3a
    0x0000030: 092a b800 3f3a 0800 2a19 0819 09b8 0043
    0x0000040: 3a0a 190a b0bf                         
  Stackmap Table:
    append_frame(@55,Top,Object[#87])
    full_frame(@69,{},{Object[#89]})

    at java.lang.Class.getDeclaredFields0(Native Method)
    at java.lang.Class.privateGetDeclaredFields(Class.java:2583)
    at java.lang.Class.getField0(Class.java:2975)
    at java.lang.Class.getField(Class.java:1701)
    at org.jruby.ir.targets.JVMVisitor.defineFromBytecode(JVMVisitor.java:101)
    at org.jruby.ir.Compiler.execute(Compiler.java:54)
    ... 11 more
```

cc @subbuss @enebo

headius:
Fixes #4251.

### DIFF
--- a/core/src/main/java/org/jruby/ir/dataflow/analyses/DefinedVariableNode.java
+++ b/core/src/main/java/org/jruby/ir/dataflow/analyses/DefinedVariableNode.java
@@ -7,6 +7,7 @@ import org.jruby.ir.instructions.Instr;
 import org.jruby.ir.instructions.ResultInstr;
 import org.jruby.ir.operands.LocalVariable;
 import org.jruby.ir.operands.Operand;
+import org.jruby.ir.operands.TemporaryLocalVariable;
 import org.jruby.ir.operands.Variable;
 import org.jruby.ir.operands.WrappedIRClosure;
 import org.jruby.ir.IRClosure;
@@ -116,10 +117,9 @@ public class DefinedVariableNode extends FlowGraphNode<DefinedVariablesProblem, 
         initSolution();
         for (Instr i: basicBlock.getInstrs()) {
             for (Variable v: i.getUsedVariables()) {
-                LocalVariable lv;
                 if (!v.isSelf()) {
                     if (v instanceof LocalVariable) {
-                        lv = (LocalVariable)v;
+                        LocalVariable lv = (LocalVariable) v;
                         // Variables that belong to outer scopes
                         // are considered already defined.
                         if (lv.getScopeDepth() < parentScopeDepth && !tmp.get(problem.getDFVar(v))) {
@@ -127,6 +127,12 @@ public class DefinedVariableNode extends FlowGraphNode<DefinedVariablesProblem, 
                             undefinedVars.add(lv.getScopeDepth() == 0 ? lv : lv.cloneForDepth(0));
                         }
                         tmp.set(problem.getDFVar(lv));
+                    } else if (v instanceof TemporaryLocalVariable) {
+                        TemporaryLocalVariable tlv = (TemporaryLocalVariable) v;
+                        if (!tmp.get(problem.getDFVar(v))) {
+                            undefinedVars.add(tlv);
+                        }
+                        tmp.set(problem.getDFVar(tlv));
                     }
                 }
             }
--- a/spec/compiler/general_spec.rb
+++ b/spec/compiler/general_spec.rb
@@ -326,6 +326,12 @@ modes.each do |mode|
       expect{run("def foo(a, b=(c=1));[a,b,c];end;foo(1,2,3)")}.to raise_error(ArgumentError)
     end
 
+    it "compiles accesses of uninitialized variables" do
+      run("def foo(a); if a; b = 1; end; b.inspect; end; foo(false)") {|result| expect(result).to eq("nil") }
+      run("def foo(a); a ||= (b = 1); b.inspect; end; foo(1)") {|result| expect(result).to eq("nil")}
+      run("def foo(a); a &&= (b = 1); b.inspect; end; foo(nil)") {|result| expect(result).to eq("nil")}
+    end
+
     it "compiles grouped and intra-list rest args" do
       run("def foo(a, (b, *, c), d, *e, f, (g, *h, i), j); [a,b,c,d,e,f,g,h,i,j]; end; foo(1,[2,3,4],5,6,7,8,[9,10,11],12)") do |result|
         expect(result).to eq([1, 2, 4, 5, [6, 7], 8, 9, [10], 11, 12])


headius:
@subbuss I am confused though...I do not see any TemporaryLocalVariable handling before or after your patch, but obviously the issue in #4251 was exactly why we started initializing all temp locals eagerly to nil. What am I missing?

enebo:
@subbuss @headius I think our issue in this specific case is the optarg is missing an init when it is an lvar and then we convert it to a temp.  So this fixes the problem but I think the underlying issue is we convert non-nil-corrected lvars into temp vars.  

Another argument worth asking is should we worry about keeping this assumption that temp vars are always initialized or should we just add this extra logic and cover our asses?

subbuss:
This patch is incomplete. I need to page in this code into my memory. But, looks like I started doing the initialization for all vars (see buildDataflowVars which gets all vars, not just LocalVariables), and at some point decided to test for inits for only LocalVariables.

So, anywhere there is a check for LocalVariable in the code needs an update to not assume lvar. I can follow up if you want, or I can point you to other places to fix.

applyTransferFunction is the only other place it seems .. since tmps don't cross scope boundaries, identifyUndefinedVarsInClosure is correct to look at lvars.

subbuss:
You can reproduce the error with this snippet as well:
```
def foo(a)
  if (a)
    b = 1
  end
  p b
end

foo 0
```
You will find that foo will fail to JIT with the same error. This is because we run the the AddMissingInitsPass at the end to catch all possible scenarios of missing inits (which is why the code should work for tmps and lvars). In this case, the OptimizeDynamicVarsPass converts all the lvars to tmps and the tmps will have a missing init instead of the lvar. So, your patch is on the right track. You just need to fix applyTransferFunction as well.

enebo:
@subbuss I am agreeing it is better to have this work on all variable types now since we have hit this problem more than once; but if we move missing inits higher up the pass chain your snippet will not be broken anymore.

As a point of general guidance should all passes be made to work in any order as a principle do you think?  Or should we just get better about enforcing dependencies?

subbuss:
Both.

1. It is more robust to have a pass work anywhere in the chain, and that is how some should work. I really meant to have AddMissingInitsPass work anywhere ... it is just plain oversight since I have been context switching jruby work across long periods of time and I lose attention. So, some of that can be improved by adding tests and specs.

2. Some passes really can work after the IR is in some pre-determined state (ex: CFG has been built, or call protocol instrs haven't been added, code is in SSA form, etc.). So, there are also dependencies that have to be respected and enforced. In most cases, failed dependencies will fail fast. But, better to enforce them.

headius:
> So, your patch is on the right track. You just need to fix applyTransferFunction as well.

Great! I'll work on that once we're back from Austin.

headius:
I added the fix for `applyTransferFunction` and re-pushed. I'll rebase as well so we can see it with recent greening on master.

subbuss:
You will hate me, but now that I look at the code again, I was wrong. tmp vars don't leak in from outer scopes. Only lvars do! So, applyTransferFunction was correct as is. So, your previous fix was indeed complete. Sorry, but I seem to mix up this "simple" tmp / lvar distinction every once in a while.

headius:
@subbuss Oh, so like the closure var walker later in the file, this should only ever see `LocalVariable`. I'll remove my change and merge, since CI looks pretty good.

subbuss:
Yes. 

Can you please add JIT tests (either as part of the PR or in a followup commit)? The example I added earlier in https://github.com/jruby/jruby/pull/4252#issuecomment-256537591 should be good enough.

headius:
@subbuss This was already failing a couple JIT tests, but I'll add any cases that aren't covered.

Issue #3655, opened 2016-02-09
Missing stdout using `Open3.popen` after `Dir.chdir`


hsbt:
It's related with https://github.com/jruby/jruby/issues/3653 probably. I found another issue.

I invoke following code named inlineruby.rb.

```rb
require 'open3'

RUBY = ENV['RUBY'] || File.join(
  RbConfig::CONFIG['bindir'],
  RbConfig::CONFIG['ruby_install_name'] + RbConfig::CONFIG['EXEEXT'])

def run_ruby(*option_list)
  Open3.popen3(RUBY, *option_list) {|inn, out, err, wait|
    inn.close

    @exit = wait ? wait.value : $?
    @out = out.read
    @err = err.read
  }
  p [@exit, @out, @err]
end


run_ruby '-e', 'puts ["a"].size'

Dir.chdir 'test' # prepare to create test dir.
run_ruby '-e', 'puts ["a"].size'
```

I got following results with JRuby 9.0.5.0

```
% ruby -v inlineruby.rb
jruby 9.0.5.0 (2.2.3) 2016-01-26 7bee00d Java HotSpot(TM) 64-Bit Server VM 25.72-b15 on 1.8.0_72-b15 +jit [darwin-x86_64]
[#<Process::Status: pid 50198 exit 0>, "1\n", ""]
[#<Process::Status: pid 50199 exit 0>, "\n", ""]
```

JRuby 1.7.24 returns same results.

```
jruby 1.7.24 (1.9.3p551) 2016-01-20 bd68d85 on Java HotSpot(TM) 64-Bit Server VM 1.8.0_72-b15 +jit [darwin-x86_64]
[#<Process::Status: pid=50468,exited(0)>, "1\n", ""]
[#<Process::Status: pid=50470,exited(0)>, "1\n", ""]
```

Is it intentional changes?

enebo:
@hsbt our IO subsystem and to some exent our process-oriented code was completely changed since 1.7.x so this is still just breakage from those changes.

hsbt:
I understand it.

but CRuby returns same results as JRuby 1.7.x

```
% ruby -v inlineruby.rb
ruby 2.4.0dev (2016-02-10 trunk 53797) [x86_64-darwin15]
[#<Process::Status: pid 21864 exit 0>, "1\n", ""]
[#<Process::Status: pid 21865 exit 0>, "1\n", ""]
```

I think this `Dir.chdir` behavior is inconsistency.

/cc @headius

[...]

Issue #3653, opened 2016-02-09
`system` method is different behavior from JRuby 1.7.x and CRuby 2.3.x


hsbt:
I'am preparing to release Rake 11. but Rake master has some blocker related JRuby 9.0.5.0.

see https://travis-ci.org/ruby/rake/jobs/107731466

I investigated 3 and 4 in above log. I found strange difference at JRuby 9.0.5.0. Please see re-produce code: https://github.com/hsbt/rake-issue-chdir

JRuby 9.0.5.0 expand ENV variables to environmental variable in `system` shell runner after invoking `Dir.chdir`.

Is this intentional behavior?

headius:
Ok, interesting. I'll have a look.

headius:
Ok, the problem here is that the new logic for process spawning does not support chdir in the same way as MRI.

In MRI, if you specify chdir when calling spawn or system or whatever, it will actually do a native chdir before launching the child process. It does this either after forking or temporarily around the call to spawn(2) (setting it back after launching the process).

MRI also actually does a real native chdir when doing `Dir.chdir`.

We have never been comfortable changing the directory the JVM is running in, because we worry that will at least confuse the JVM and at worst interfere with other threads or applications running in the same process. However, we have never tested real chdir to see if anything bad happens.

A short term fix for this in JRuby would be to fall back on the old JVM process logic when we need to chdir, since that logic uses fork+exec similar to MRI. This might be ok for system, since all it does is capture all child process output. However, it would be too great a change for e.g. `spawn` since the JVM logic can't support the many options `spawn` requires.

headius:
It's worth pointing out that MRI will behave the same as us even without chdir if the command given looks like a command line. For example, if I change
```
sh RUBY, 'check_no_expansion.rb', '$RAKE_TEST_SH', 'someval'
```
to
```
sh RUBY + ' check_no_expansion.rb $RAKE_TEST_SH someval'
```
MRI will also expand the env var. So it's a bit fragile to build commands that have loose env var references and expect them not to expand.

headius:
This is fixed, but it's still an issue for other spawn forms. However, I still feel like it's asking for trouble spawning subprocesses with env vars and expecting them not to expand.

hsbt:
> I still feel like it's asking for trouble spawning subprocesses with env vars and expecting them not to expand.

[...]

Issue #985, opened 2013-09-04
Kernel.spawn doesn't support some options (ex: :chdir)


tagomoris:
Some options like :pgroup works, but others (ex: :chdir, filehandles) doesn't work.

```
$ jruby -e 'pid = spawn("sleep 5", :pgroup => 0, :chdir => "/", STDOUT => :close); p pid; Process.detach(pid)'
file:/Users/tagomoris/.rbenv/versions/jruby-1.7.4/lib/jruby.jar!/jruby/kernel19/process.rb:4 warning: unsupported spawn option: chdir
TypeError: wrong exec option: RubyIO(66, 1)
  _spawn_internal at org/jruby/RubyProcess.java:1067
            spawn at file:/Users/tagomoris/.rbenv/versions/jruby-1.7.4/lib/jruby.jar!/jruby/kernel19/process.rb:3
            spawn at file:/Users/tagomoris/.rbenv/versions/jruby-1.7.4/lib/jruby.jar!/jruby/kernel19/kernel.rb:29
           (root) at -e:1
```

jcode:
any update on this? We're having problems using the schmooze gem :(

doudou:
chdir seems to work fine on 9.1.7.0 with `system` and `spawn` ... but not `exec`

headius:
chdir for system and spawn is achieved by modifying the command line to run as an sh script that chdir's first. Unfortunately a similar approach with popen breaks other things, so we can't support that yet.

headius:
The original example appears to run correctly.

The example provided by @BanzaiMan still fails to execute:

```
$ rm -f bar; ruby -v -e 'exec("date", :chdir => "/", STDOUT => "bar")'; cat bar
jruby 9.3.0.0-SNAPSHOT (2.6.5) 2021-06-15 36b083365a OpenJDK 64-Bit Server VM 11.0.4+11 on 11.0.4+11 +jit [darwin-x86_64]
uri:classloader:/jruby/kernel/kernel.rb:4: warning: unsupported exec option: chdir
Tue Jun 15 16:20:00 CDT 2021
cat: bar: No such file or directory
```

The `Open3.popen2e` example should work correctly but the example is not a free-standing example. There has been work in the past couple years to improve the compatibility of open3 and it now nearly the same code as CRuby.

There are other issues filed for more specific examples not mentioned here, so I will close this one as fixed in 9.3.

PR #4263, opened 2016-11-03
Add tests for Kernel#spawn's chdir option

seanfisk:
First of all, thank you for all of your hard work on JRuby! It is very much appreciated.

It seems like some of the issues surrounding the current implementation of the `chdir` option to `Kernel#spawn` and friends are well-known (#985, #3653, #3655). I am aware that JRuby transitioned from Java's subprocess execution system to MRI's in recent releases, which was the source of some bugs. I am also aware of the concern associated with doing a native `chdir` in the JVM process.

Instead of just reporting an issue, I thought it would be helpful to add some [pending] tests that illustrate how the current behavior differs from MRI/the expected behavior. The two tests with `pend` fail in JRuby, while they pass in MRI 2.3.1 with flying colors (after removing `pend`).

The behavior of these functions when given the `chdir` option varies from MRI in the following ways:
- JRuby *always* executes the command through the shell when given the `chdir` option. This is not appropriate because characters in a shell command have different meaning than those passed directly to `exec` and friends. For example, a shell expands `$var`, etc. while those characters retain literal meaning in `exec`. `Kernel#spawn` has a specific (albeit unnecessarily complex) contract as to when it will execute a command-line through the shell vs. directly, and `chdir` in JRuby overrides that.
- JRuby currently surrounds the value of the `chdir` option with single quotes and inserts it into the command passed to the shell. Without too much effort, it is easy to achieve arbitrary code execution by injecting shell commands into the value of `chdir`. This is a possible security vulnerability, especially when the value of `chdir` is taken from user input.

I'm not sure of the best way to address these issues in the implementation. Per @headius's comments [here](https://github.com/jruby/jruby/issues/3653#issuecomment-182051175), I think it would be worth giving native `chdir` a shot just to see what happens. I'd be happy to assist in implementation and testing with all the time I have available.

What do you think? Again, thank you for JRuby!

### DIFF
--- a/antlib/extra.xml
+++ b/antlib/extra.xml
@@ -129,6 +129,7 @@ build jruby-complete.jar
 <arg value='test/test_instantiating_interfaces.rb'/>
 <arg value='test/test_openssl.rb'/>
 <arg value='test/test_tempfile_cleanup.rb'/>
+<arg value='test/test_spawn_chdir.rb'/>
 <!--arg value='-v'/-->
 </exec>
 </target>
--- a/test/jruby.index
+++ b/test/jruby.index
@@ -63,6 +63,7 @@ jruby/test_rbconfig
 jruby/test_require_once
 jruby/test_respond_to
 jruby/test_socket
+jruby/test_spawn_chdir
 jruby/test_string_java_bytes
 jruby/test_string_printf
 jruby/test_string_to_number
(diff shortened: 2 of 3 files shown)


seanfisk:
The [first Travis build failure](https://travis-ci.org/jruby/jruby/jobs/172835806) seems related to these new tests; not quite sure why the other Travis jobs are succeeding while this is failing, though. [The second Travis failure](https://travis-ci.org/jruby/jruby/jobs/172835808) and the Appveyor failures seem unrelated to the new tests, however.

headius:
This is quite old and we are now running the relevant CRuby and ruby/spec tests for this behavior.

PR #4268, opened 2016-11-07
upgrade jruby-openssl to 0.9.18

kares:
...


enebo:
@kares how critical is this update?  We want 9.1.6.0 out in next couple of days.  I am happy to merge if you tell me the risk is low.

kares:
not much in jossl - xouple of small fixes. BC is updated to latest. should be low risk but I do not mind skiping till 9.1.7.0 users can install latest gem already.

enebo:
@kares thanks.  Let's do that then.  We will try a quicker turnaround for 9.1.7 and 9.2 will be at end of year.

PR #4285, opened 2016-11-14
Catch LinkageError during DynamicScope generation.

snowp:
When multiple threads are generating the same DynamicScope class, both
threads may try to generate the same class at the same time. When the second
thread hits defineClass, it throws a LinkageError because the class has already
been defined.

With this change we catch this exception and call loadClass to get the class that
the previous thread created.


headius:
Maybe we should just lock around the class generation bit?

headius:
I will incorporate your patch but with some double-checked locking around class generation, so we can simply avoid the exception when there's parallel generation going on.

headius:
Great find, thanks!

snowp:
Thanks for the fast response!

PR #4295, opened 2016-11-15
Add File#empty? and File#zero? methods

maclover7:
TODO:

- get tests passing
- add support for IO objects as argument to method (and add tests for this behavior)

### DIFF
--- a/core/src/main/java/org/jruby/RubyFile.java
+++ b/core/src/main/java/org/jruby/RubyFile.java
@@ -1526,6 +1526,20 @@ public class RubyFile extends RubyIO implements EncodingCapable {
                                 && path.charAt(1) == ':'));
     }
 
+    @JRubyMethod(name = {"empty?", "zero?"})
+    public static boolean empty(ThreadContext context, String name) {
+      Ruby runtime = context.runtime;
+      int fileSize = runtime.newFileStat(name).size();
+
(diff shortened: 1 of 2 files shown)


kares:
seems that its not compilable: 
*src/main/java/org/jruby/RubyFile.java:[1532,28] error: no suitable method found for newFileStat(String)*

enebo:
@maclover7 @kares Amazingly someone else had submitted a PR before you and I landed it but perhaps a single stat is better than an exist? + stat() hmm.  exist? on java.io.File is quicker than us performing a stat via jnr-posix but I am guessing more than half the time you expect the file to exist so you are doing two things...

enebo:
The other PR probably works better on a minutes reflection as this mechanism will raise an ENOENT or ENODIR I think if the file does not actually exist in stat.setup().

@kcdragon @maclover7 what would work and if either of you are up for it would be to do something like:

```text
file = JRubyFile.createResource(runtime, filename);
if file == null then 
  false
else 
  file.stat.exist?
end
```
We try really hard to avoid exception when we can as they are expensive.  This above approach will eliminate the original exist() + stat() calling on existing file (which is two stat() calls) to a single one.

enebo:
Open a new PR if either of you want to take that on.   I am closing this one.

Issue #4293, opened 2016-11-14
Ruby 2.4 support


headius:
This bug will track all NEWS items new in Ruby 2.4. This list is based off https://github.com/ruby/ruby/blob/trunk/NEWS.

Only changes relevant to JRuby are listed here. MRI-specific internal changes and features are not included.

NOTE: Pull requests should be done against the `ruby-2.4` branch.

### Language changes

* [x] Multiple assignment in conditional expression is now allowed. [Feature #10617](https://bugs.ruby-lang.org/issues/10617)

* [ ] Refinements is enabled at method by Symbol#to_proc.  [Feature #9451](https://bugs.ruby-lang.org/issues/9451)

* [ ] Refinements is enabled with Kernel#send and BasicObject#\_\_send\_\_. [Feature #11476](https://bugs.ruby-lang.org/issues/11476)

* [x] Rescue modifier now applicable to method arguments. [Feature #12686](https://bugs.ruby-lang.org/issues/12686)

* [ ] Toplevel return is now allowed. [Feature #4840](https://bugs.ruby-lang.org/issues/4840)

### Core classes updates (outstanding ones only)

* Array

  * [x] Array#max and Array#min.  [Feature #12172](https://bugs.ruby-lang.org/issues/12172)

  * [x] Array#sum  [Feature #12217](https://bugs.ruby-lang.org/issues/12217)

  * [x] Array#concat [Feature #12333](https://bugs.ruby-lang.org/issues/12333)

  * [x] Array#pack now takes optional argument `buffer:` to reuse an already-allocated buffer. [Feature #12217](https://bugs.ruby-lang.org/issues/12217)

* Comparable

  * [x] Comparable#clamp.  [Feature #10594](https://bugs.ruby-lang.org/issues/10594)

* Dir

  * [x] Dir.empty?.  [Feature #10121](https://bugs.ruby-lang.org/issues/10121)

* Enumerable

  * [x] Enumerable#sum  [Feature #12217](https://bugs.ruby-lang.org/issues/12217)
  * [x] Enumerable#uniq  [Feature #11090](https://bugs.ruby-lang.org/issues/11090)
  * [x] Enumerable#chunk called without a block now return an Enumerator. [Feature #2172](https://bugs.ruby-lang.org/issues/2172)

* Enumerator::Lazy

  * [x] Enumerator::Lazy#chunk_while  [GH-1186](https://github.com/ruby/ruby/pull/1186)
[...]

PR #4296, opened 2016-11-15
Add File.empty?

kcdragon:
This PR adds the class methods `File.empty?` (and `File.zero?`). I included `File.zero?` because `File.empty?` in the docs links to `File.zero?`. https://ruby-doc.org/core-2.4.0_preview3/File.html#method-c-empty-3F (I'm guessing one is an alias of the other)

See #4293 where feature is originally mentioned.

This PR gets the test `test_empty_p` in "test/mri/ruby/test_file_exhaustive.rb" to pass. I figured that I didn't need to add any other tests since this functionality is tested by the MRI tests. Let me know if I do need to add more tests.

I did run into an issue while testing with just that file. If I ran just this command

```
bin/jruby test/mri/ruby/test_file_exhaustive.rb -n test_empty_p
```

I get an error about missing `assert_file`.

```
Error: test_empty_p(TestFileExhaustive):
  NameError: undefined local variable or method `assert_file' for #<TestFileExhaustive:0x29a5f4e7>
```

I needed to add `require_relative 'envutil'` (which adds `assert_file`) in order to run the tests. I would have included that change in this PR but it looks like those tests are generated based on the tests in Ruby and I wasn't sure if changes to those files would be overridden when new tests are brought in. That file is probably eventually required when the whole suite is run.

This is my first time contributing to JRuby so let me know if there is something missing that I need to include.

### DIFF
--- a/core/src/main/java/org/jruby/RubyFile.java
+++ b/core/src/main/java/org/jruby/RubyFile.java
@@ -784,6 +784,24 @@ public class RubyFile extends RubyIO implements EncodingCapable {
         return result;
     }
 
+    @JRubyMethod(name = {"empty?", "zero?"}, required = 1, meta = true)
+    public static IRubyObject empty(ThreadContext context, IRubyObject self, IRubyObject str) {
+        Ruby runtime = context.runtime;
+        String f = StringSupport.checkEmbeddedNulls(runtime, get_path(context, str)).getUnicodeValue();
+
+        if ( ! new File(f).exists() ) {
+            return runtime.getFalse();
+        }
+
+        int size = runtime.newFileStat(f, false).size().convertToInteger().getIntValue();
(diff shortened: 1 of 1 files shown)


enebo:
@kcdragon Logic was right on.  My only comment for next one is to try and use a full name like filename and not f.  We have some other brace logic which is like 95% followed so that might be another style tweak suggestion in the future. 

Congratulations you are the on JRuby commit board.  The tracks have been greased...onward! :)

headius:
@kcdragon There are some examples of running individual test files from the MRI suite in BUILDING.md. Basically, the test/mri/runner.rb script can be used to set up appropriate paths and such, and there's a feature in there that's used to exclude tests known to fail (for maintaining a high-water mark).

Perhaps we should have a better page in the wiki for this.

Issue #4306, opened 2016-11-16
JRuby 1.7: update to latest jruby-openssl version


cprice404:
### Environment

JRuby version 1.7.26 ships with jruby-stdlib 1.7.26, which bundles version 0.9.16 of the jruby-openssl gem.  We've had a few customers report some compatibility issues between some of their other x509 tools and older versions of bouncycastle/jruby-openssl, so we're trying to upgrade our packages to ship with jruby-openssl 0.9.18.  It seems like the cleanest path for that by far will be to upgrade the version that is included in jruby-stdlib.

### Expected Behavior

Latest JRuby 1.7.x release includes jruby-openssl 0.9.18.

### Actual Behavior

Latest JRuby 1.7.x release includes jruby-openssl 0.9.16.

I'll see if I can work up a PR.

cprice404:
There are some errors that have shown up in CI for the related PR.  After exploration, I believe that the MRI `x509store` tests should be failing on JRuby9k with jruby-openssl `0.9.18`, just like the failure that is showing up in CI for JRuby 1.7 on my PR.  For more info see jruby/jruby-openssl#113.

If I should open another issue for JRuby9k (since this one's title mentions 1.7), let me know and I'll be happy to do so.

cprice404:
Asking about this in IRC, and was told that the MRI tests for JRuby9k are run in CI via `mvn -Prake -Dtask=test:mri`.  I ran that and grepped all of the output for "openssl" and "x509" and didn't see any matches, leading me to believe that the openssl tests may not be getting run in CI for JRuby9k.

kares:
on 0.9.17 for the moment, hopefully 0.9.19 will be released soon-ish ... (before the next/final? 1.7 release)

kares:
and we're updated to **0.9.19** - took quite longer than expected.
hopefully it won't need to be done again (at least for 1.7)

cprice404:
hooray, thanks!

PR #4307, opened 2016-11-16
JRuby 1.7: update jruby openssl to 0.9.18

cprice404:
This should address #4306 .

### DIFF
--- a/lib/pom.rb
+++ b/lib/pom.rb
@@ -10,7 +10,7 @@ end
 
 default_gems =
   [
-   ImportedGem.new( 'jruby-openssl', '0.9.16' ),
+   ImportedGem.new( 'jruby-openssl', '0.9.18' ),
    ImportedGem.new( 'rake', '${rake.version}' ),
    ImportedGem.new( 'rdoc', '${rdoc.version}' ),
    ImportedGem.new( 'json', '1.8.0' ),
--- a/lib/pom.xml
+++ b/lib/pom.xml
@@ -32,7 +32,7 @@ DO NOT MODIFIY - GENERATED CODE
     <dependency>
       <groupId>rubygems</groupId>
       <artifactId>jruby-openssl</artifactId>
-      <version>0.9.16</version>
+      <version>0.9.18</version>
       <type>gem</type>
     </dependency>
     <dependency>


kares:
hey, thanks! unfortunately a (1.7 specific?) regression seems to have showed up on travis-ci : 
```
     [java]   1) Error:
     [java] test_verify(OpenSSL::TestX509Store):
     [java] Java::JavaLang::ClassCastException: org.bouncycastle.jce.provider.X509CertificateObject cannot be cast to org.jruby.ext.openssl.X509Cert
     [java]     org.jruby.ext.openssl.X509StoreContext.initialize(X509StoreContext.java:132)
     [java]     org.jruby.ext.openssl.X509StoreContext.newStoreContext(X509StoreContext.java:102)
     [java]     org.jruby.ext.openssl.X509Store.verify(X509Store.java:225)
```
... so this is a blocker - will likely need to be looked at jruby-openssl.
would have halted the jossl 0.9.18 release but I assumed if the suite passes on master than its fine :(

cprice404:
Drat.  So you think the fix for the failure will be in jruby-openssl, as opposed to in JRuby itself?  I've never worked on that codebase before, is it something you think would be approachable?  Or do you expect to have time to look into it prior to the next 1.7.x release?

kares:
yes likely in jossl. the difference between 9K and 1.7 needs to be understood. it shouldn't be hard to figure.
never worked on the code base either before I ended up somehow maintaining it for the sole benefit of all :)

cprice404:
Hmmm.  Well, I cloned jruby-openssl and ran it's tests against JRuby 1.7.26 based on the README, and don't get any failures.  I guess I'll see if I can repro the failure of the jruby core travis test locally.

cprice404:
I've been intermittently trying to make some progress on this.  No major discoveries yet but it took me a while to figure out a minimal set of steps to try to repro the test failure for debugging, so I thought I'd capture that here... one of the errors can be reproduced with this command:

```
./bin/jruby -I"lib" -X-C -J-ea --1.9 -I test/externals/ruby1.9 -I test/externals/ruby1.9/ruby -r ./test/ruby19_env.rb -r minitest/excludes -I"./lib/ruby/shared" "./lib/ruby/shared/rake/rake_test_loader.rb" "test/externals/ruby1.9/openssl/test_x509store.rb"
```

cprice404:
I left some comments about this on the original issue (#4306), as well as on a new issue I filed under jruby-openssl (jruby/jruby-openssl#113) - but the tl;dr is that I believe these same failures occur with JRuby9k, and that they slipped through the cracks because the CI for JRuby 1.7 runs the affected tests, while the CI for jruby-openssl and JRuby9k do not.

PR #4309, opened 2016-11-17
fix Array#sum and Enumerable#sum

phluid61:
This changes `Array#sum` and `Enumerable#sum` to use Kahan's compensated summation algorithm, so we have the same rounding artifacts as MRI.

It also includes some of the MRI's cascading coercion behaviour in `Array#sum`

Now all tests in `test/mri/ruby/test_array.rb` and `test/mri/ruby/test_enum.rb` pass, except for those depending on 'mathn' which I can't test because of an unrelated issue.

Consider this a regression fix for #4297


phluid61:
I'm not entirely comfortable that RubyArray is now an expert on Fixnum, Bignum, Rational and Float, but the behaviour matches MRI's.

headius:
Looks good! Thank you!

PR #4310, opened 2016-11-17
Disable notifying travis-rubies of passing [skip ci]

BanzaiMan:
`teavis-rubies` now uses 3 Mac jobs to create archives for various OS
releases.

This is a bit wasteful if multiple builds pass in a short period.

Instead, Travis CI is now running a nightly build of jruby-head
(around 22:30 UTC). This will happen regardless of the state of the
`master` branch.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -98,10 +98,6 @@ notifications:
       - "%{repository} (%{branch}:%{commit} by %{author}): %{message} (%{build_url})"
     skip_join: true
 
-  # update jruby-head installed on Travis CI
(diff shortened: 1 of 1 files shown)


headius:
Does the build need to succeed? Is there a sanity check we could add to it, like a quick spec run?

BanzaiMan:
Travis does run a few sanity checks before uploading to S3. https://github.com/travis-ci/travis-rubies/blob/build/build.sh We can add a few more if we want.

PR #4318, opened 2016-11-17
Change decimal length for result of time subtraction

maclover7:
This fixes an inconsistency between MRI and JRuby while trying to get
the Active Support test suite passing on JRuby.

### DIFF
--- a/core/src/main/java/org/jruby/RubyTime.java
+++ b/core/src/main/java/org/jruby/RubyTime.java
@@ -640,11 +640,14 @@ public class RubyTime extends RubyObject {
         }
     }
 
-    private IRubyObject opMinus(RubyTime other) {
+    private IRubyObject opMinus(ThreadContext context, RubyTime other) {
         long timeInMillis = (getTimeInMillis() - other.getTimeInMillis());
         double timeInSeconds = timeInMillis/1000.0 + (getNSec() - other.getNSec())/1000000000.0;
 
-        return RubyFloat.newFloat(getRuntime(), timeInSeconds); // float number of seconds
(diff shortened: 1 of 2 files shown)


maclover7:
cc @headius

kares:
there's a failing regression spec that needs to be considered : 
```
     [exec] Failures:
     [exec] 
     [exec]   1) time minus precision handles BC time correctly
     [exec]      Failure/Error: DEFAULT_FAILURE_NOTIFIER = lambda { |failure, _opts| raise failure }
     [exec] 
     [exec]        expected: -63504993600.0
     [exec]             got: -922337203.6854776
     [exec] 
     [exec]        (compared using ==)
     [exec]      # ./spec/regression/GH-652_time_add_spec.rb:14:in `block in (root)'
```
... maybe it just needs to be run under MRI 2.3 and confirm its a mis-behavior on JRuby's end

kares:
the (failing) regression spec: */spec/regression/GH-652_time_add_spec.rb* runs fine on MRI 2.3.3

kares:
don't think this is worth a hustle since its good "enough" : 
```
irb(main):001:0> time1 = Time.utc(2000, 1, 2, 23, 59, 59, Rational(999999999, 1000))
=> 2000-01-02 23:59:59 UTC
irb(main):002:0> time2 = Time.utc(2000, 1, 2, 0, 0, 0, Rational(1, 1000))
=> 2000-01-02 00:00:00 UTC
irb(main):003:0> time1 - time2
=> 86399.99999999799
```
... test can be adjusted to allow one 10^2 less precision to work, we shall worry if this is a real-world problem

Issue #4302, opened 2016-11-16
Dumping Hash with default_proc that was reset


kirs:
```ruby
# default_proc.rb
hash = Hash.new { |k| "default value" }
hash.default_proc = nil
Marshal.dump(hash)

$ jruby default_proc.rb
TypeError: can't dump hash with default proc
    dump at org/jruby/RubyMarshal.java:102
  <main> at default_proc.rb:3
```

Running on `jruby 9.1.6.0 (2.3.1) 2016-11-09 0150a76 Java HotSpot(TM) 64-Bit Server VM 25.45-b02 on 1.8.0_45-b14 +jit [darwin-x86_64]`

### Expected Behavior

JRuby should be able to dump a hash that was initialized with `default_proc` that was later set to `nil`.

### Actual Behavior

JRuby currently ignores `default_proc` set to `nil` and gives an error that a hash with default_proc cannot be dumped.

@headius @guilleiguaran

guilleiguaran:
I guess this is related to this failure:
https://github.com/jruby/jruby/blob/master/spec/tags/ruby/core/hash/default_proc_tags.txt

headius:
@guilleiguaran Very likely, as well as some failures @enebo was working on in ActiveRecord!

kirs:
@enebo please let me know if you already worked on it. If not, I'll be happy to submit a PR

enebo:
@guilleiguaran OMGZ you rock.  I have been trying to figure out what was supplying a default Hash to marshal in a local unit test in arjdbc and did not realize 2.3 allowed this.  So it is in fact a JRuby bug and not a arjdbc one.  This will save me time since I will now look in the right place!

enebo:
@guilleiguaran wow github is not working great this morning.  Yeah if you can figure this out then go for it.  I would figure out the commit in MRI and work from there.

kirs:
So I cloned the repo, installed dependencies and tried to run test related to the Hash class:

```
$ git checkout master
$ bin/jruby test/mri/runner.rb test/mri/ruby/test_hash.rb
```

But many of them failed:

```
206 tests, 1413 assertions, 12 failures, 10 errors, 1 skips
```

It is supposed to be stable and green atm? Or is there some specific branch I should look at?

enebo:
```text
jruby -r ./test/mri_test_env.rb test/mri/runner.rb  -v test/mri/ruby/test_hash.rb
```
[...]

PR #4321, opened 2016-11-20
Fix marshaling Hash with default_proc set to nil

kirs:
When we set `default_proc` to nil, we also have to update the internal flag.

Fixes https://github.com/jruby/jruby/issues/4302

@enebo


enebo:
@kirs Excellent.  I was worried this involved more than nil so thanks for digging into this!

PR #4326, opened 2016-11-21
[Truffle] Compile c specs with extconf make

bjfish:
It looks like `mkmf` will always look for all c files in the directory. So, I copied the c spec to a temp directory for compilation. This appears to be working. It may run a little slower than the previous method since I had to run `extconf.rb` with another jruby+truffle (or the makefile would contain values from previous extension).

cc: @chrisseaton

### DIFF
--- a/spec/ruby/optional/capi/spec_helper.rb
+++ b/spec/ruby/optional/capi/spec_helper.rb
@@ -3,6 +3,7 @@ $extmk = false
 
 require 'rbconfig'
 require 'fileutils'
+require 'tmpdir'
 
 OBJDIR ||= File.expand_path("../../../ext/#{RUBY_NAME}/#{RUBY_VERSION}", __FILE__)
 FileUtils.makedirs(OBJDIR)
@@ -43,7 +44,7 @@ def compile_extension(name)
     require 'mkmf'
     hdrdir = $hdrdir
   elsif RUBY_NAME == 'jruby+truffle'
-    return compile_extension_jruby_truffle(name)
+    return compile_jruby_truffle_extconf_make(name, path, objdir)
   else
     raise "Don't know how to build C extensions with #{RUBY_NAME}"
   end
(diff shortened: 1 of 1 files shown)


chrisseaton:
Looks good.

You could also just move the files in the repo and commit that.

spec/ruby/optional/capi/spec_helper.rb
```diff
@@ -119,6 +120,33 @@ def compile_extension_jruby_truffle(name)
...
+  temp_dir = Dir.mktmpdir
+  copy =  File.join(temp_dir, file)
+  lib_target = File.join(objdir, "#{ext}.#{RbConfig::CONFIG['DLEXT']}")
+  begin
+    FileUtils.cp File.join(path, "rubyspec.h"), File.join(temp_dir, "rubyspec.h")
+    FileUtils.cp File.join(path, "jruby_truffle.h"), File.join(temp_dir, "jruby_truffle.h")
+    FileUtils.cp source, copy
+    extconf_src = "require 'mkmf'\n" +
+                  "create_makefile('#{ext}', '#{temp_dir}')" # ,
+    File.open(File.join(temp_dir, "extconf.rb"), 'w') {|f| f.write(extconf_src) }
```
eregon: space between `{` and `|`

eregon: or alternatively use `File.write`

spec/ruby/optional/capi/spec_helper.rb
```diff
@@ -119,6 +120,33 @@ def compile_extension_jruby_truffle(name)
...
+    Dir.chdir(temp_dir) do
+      ruby = File.join(RbConfig::CONFIG["bindir"], RbConfig::CONFIG["RUBY_INSTALL_NAME"])
+      system "#{ruby} extconf.rb"
+      system "make"                                                     # run make in temp dir
+      FileUtils.copy_file("#{ext}.su", lib_target)                      # copy to .su file to library dir
+      FileUtils.copy_file("#{ext}.ll", File.join(objdir, "#{ext}.ll"))  # copy to .ll file to library dir
```
eregon: Why `copy_file` but `cp` above?

eregon:
Looks good, please address my comments.
In general, don't bother with `File.join`, it just adds `/` between the arguments, even on Windows, and adds quite a bit of noise.

chrisseaton:
Ha you hate File.join!

eregon:
@chrisseaton Indeed :smile: 
More precisely, I dislike mixing File/FileUtils and other pathname-related modules, and File.join is usually all over the place when it does nothing useful beyond what would be clearer as `"#{bin}/#{ruby}"`.

bjfish:
@eregon I've pushed an update with all your recommended updates, thanks

eregon:
This is a lot easier to read, thanks! :+1:

spec/ruby/optional/capi/spec_helper.rb
```diff
@@ -119,6 +120,32 @@ def compile_extension_jruby_truffle(name)
   File.delete(sulong_config_file) if File.exist?(sulong_config_file)
 end
 
+def compile_jruby_truffle_extconf_make(name, path, objdir)
+  ext       = "#{name}_spec"
+  file      = "#{ext}.c"
+  source    = "#{path}/#{file}"
+  temp_dir = Dir.mktmpdir
```
eregon: This should be right before the `begin`, otherwise it could leave an extra tmpdir if the lines below fail.
One way is moving those lines in the begin.

eregon:
Let's merge this once my final nitpicks are adressed :smiley:

bjfish:
@eregon I just pushed updates for the last two comments

PR #4331, opened 2016-11-23
Skip less specs

kirs:
With https://github.com/jruby/jruby/pull/4321 merged, we no longer have to skip that ruby spec.

@enebo


headius:
Thank you!

PR #4335, opened 2016-11-25
[Truffle] New system for options.

chrisseaton:
We need to disconnect from the JRuby system because:

* We don't want dependencies on the core JRuby code as it causes AOT problems.
* `PolyglotEngine` has a configuration system and we should be going through that.
* Options need to be compilation final.
* Adding options in multiple places is a chore.
* Different options should be able to be set for `PolyglotEngine`s in the same VM.

I'm suggesting the system implemented below, with the extra phase that we will code generate some of the classes from the text file. This would be like the parser - we'd commit the generated code as it rarely changes and is small. The code generator would just be a Ruby template - you'd `jt build options`.

I haven't written the code generator yet. The name `NewOptions` would just become `Options`. I didn't make that change to keep this diff small.

All configuration which used to come from `RubyInstanceConfig` will now be an option. An example of where this is different to before is that `arguments` (`ARGV`) would be an option. At the moment there's no way to set this in `PolyglotEngine` - this new system makes it possible to set it both as a `PolyglotEngine` configuration option and even in a system property.

I'm proposing we use this one system for all options.

For the difference it makes in the key areas, see

* https://github.com/jruby/jruby/compare/truffle-head...truffle-options?expand=1#diff-7251a9f169e965e77978f3841e2c9bfaR118
* https://github.com/jruby/jruby/compare/truffle-head...truffle-options?expand=1#diff-56a6b80386767522713adc1aafcd9d45R39 (all fields we use in `RubyInstanceConfig` would instead become one of these lines)

Opinions and votes for and against please @pitr-ch @bjfish @eregon @nirvdrum.

Don't merge.


chrisseaton:
@jtulach is this what we're supposed to be doing with the config in `PolyglotEngine`? Making that the place where all options can be set?

truffle/src/main/java/org/jruby/truffle/RubyEngine.java
```diff
@@ -35,6 +36,7 @@ public RubyEngine(RubyInstanceConfig instanceConfig) {
 
         engine = PolyglotEngine.newBuilder()
                 .config(RubyLanguage.MIME_TYPE, INSTANCE_CONFIG_KEY, instanceConfig)
+                .config(RubyLanguage.MIME_TYPE, OptionsCatalogue.ARGUMENTS.getName(), instanceConfig.getArgv())
```
eregon: Is this line needed?

truffle/src/main/java/org/jruby/truffle/options/NewOptions.java
```diff
@@ -0,0 +1,25 @@
...
+ * GNU General Public License version 2
+ * GNU Lesser General Public License version 2.1
+ */
+package org.jruby.truffle.options;
+
+// This file would be automatically generated from the list of options in the text file.
+
+public class NewOptions {
+
+    public final String[] ARGUMENTS;
```
chrisseaton: Ah I didn't know about that - good idea.

truffle/src/main/java/org/jruby/truffle/RubyEngine.java
chrisseaton: We're getting rid of the `RubyInstanceConfig` on the line above when I convert all options over. So this is the only way we'll have to communicate arguments across. And it's a good thing because you'll then be able to set Ruby arguments for code that you invoke from other languages.

eregon:
Looks good to me 😃  👍

truffle/src/main/java/org/jruby/truffle/RubyEngine.java
eregon: How about passing them as a HashMap or so instead?
This could become quickly a long list, isn't it?

eregon: Ah, but options can just not be passed and then use the defaults I guess, except from some key things like ARGV coming from non-system property or OptionDescriptor defaults.

chrisseaton: It's already a hash map - one that the polyglot engine maintains - these `config` calls are like `put`. And it's already private to our language.

There are only ten or so fields like this that we need to pass. We don't need to do this for options that `RubyInstanceConfig` doesn't currently set.

pitr-ch:
Looks good 👍 Only one suggestion, could the txt file be something structured? E.g.: a Ruby hash in a file using `eval File.read('options.rb')`, or a Yaml file (, or JSON)?

chrisseaton:
I thought the simple structure was a bonus? I'll write it as Ruby DSL or a Yaml or JSON file if people want. Chime in if you want that changed.

pitr-ch:
But in unstructured txt you have to format manually and then parsed. Hash can be eval-ed (or yaml loaded) and it's done.

nirvdrum:
I agree with Petr. If the idea is the text file isn't going to change that frequently, I'd rather use an established file format, even if it's more verbose.

truffle/src/main/java/org/jruby/truffle/options/OptionsCatalogue.java
```diff
@@ -0,0 +1,31 @@
...
+ *
+ * Eclipse Public License version 1.0
+ * GNU General Public License version 2
+ * GNU Lesser General Public License version 2.1
+ */
+package org.jruby.truffle.options;
+
+// This file would be automatically generated from the list of options in the text file.
+
+public class OptionsCatalogue {
```
nirvdrum: I think we settled on using the improved American spelling.

core/src/main/java/org/jruby/util/cli/Options.java
```diff
@@ -296,7 +296,7 @@
 
     public static final Option<Integer> TRUFFLE_INSTRUMENTATION_SERVER_PORT = integer(TRUFFLE, "truffle.instrumentation_server_port", 0, "Port number to run an HTTP server on that provides instrumentation services");
     public static final Option<Boolean> TRUFFLE_EXCEPTIONS_STORE_JAVA = bool(TRUFFLE, "truffle.exceptions.store_java", false, "Store the Java exception with the Ruby backtrace");
-    public static final Option<Boolean> TRUFFLE_EXCEPTIONS_PRINT_JAVA = bool(TRUFFLE, "truffle.exceptions.print_java", false, "Print Java exceptions at the point of translating them to Ruby exceptions.");
+    //public static final Option<Boolean> TRUFFLE_EXCEPTIONS_PRINT_JAVA = bool(TRUFFLE, "truffle.exceptions.print_java", false, "Print Java exceptions at the point of translating them to Ruby exceptions.");
```
nirvdrum: I take it this is being removed?

truffle/src/main/java/org/jruby/truffle/Options.java
```diff
@@ -33,7 +33,7 @@
 import static org.jruby.util.cli.Options.TRUFFLE_ENCODING_COMPATIBLE_QUERY_CACHE;
 import static org.jruby.util.cli.Options.TRUFFLE_ENCODING_LOADED_CLASSES_CACHE;
 import static org.jruby.util.cli.Options.TRUFFLE_EVAL_CACHE;
-import static org.jruby.util.cli.Options.TRUFFLE_EXCEPTIONS_PRINT_JAVA;
+//import static org.jruby.util.cli.Options.TRUFFLE_EXCEPTIONS_PRINT_JAVA;
```
nirvdrum: I take it this is being removed?

truffle/src/main/java/org/jruby/truffle/Options.java
```diff
@@ -159,7 +159,7 @@
 
     public final int INSTRUMENTATION_SERVER_PORT = TRUFFLE_INSTRUMENTATION_SERVER_PORT.load();
     public final boolean EXCEPTIONS_STORE_JAVA = TRUFFLE_EXCEPTIONS_STORE_JAVA.load();
-    public final boolean EXCEPTIONS_PRINT_JAVA = TRUFFLE_EXCEPTIONS_PRINT_JAVA.load();
+    //public final boolean EXCEPTIONS_PRINT_JAVA = TRUFFLE_EXCEPTIONS_PRINT_JAVA.load();
```
nirvdrum: I take it this is being removed?

eregon:
:+1: for Yaml.
Alternatively the current format could be easily be made valid CSV, but that's a weird format to manually enter

truffle/src/main/java/org/jruby/truffle/options/OptionsBuilder.java
```diff
@@ -0,0 +1,65 @@
...
+
+public class OptionsBuilder {
+
+    private final String LEGACY_PREFIX = "jruby.truffle.";
+
+    private final Map<OptionDescription, Object> options = new HashMap<>();
+
+    public void set(Properties properties) {
+        for (Map.Entry<Object, Object> property : properties.entrySet()) {
+            final String name = (String) property.getKey();
```
nirvdrum: Is there a case you're trying to handle here by not just calling `toString`?

truffle/src/main/java/org/jruby/truffle/options/OptionsBuilder.java
```diff
@@ -0,0 +1,65 @@
...
+        }
+    }
+
+    private void set(String name, Object value) {
+        final OptionDescription description = OptionsCatalogue.fromName(name);
+
+        if (description == null) {
+            //throw new UnsupportedOperationException(name);
+
+            // Don't throw for now - not all the options are transalted across
```
nirvdrum: It might be beneficial to print out a warning (configurable via option!) for any option without a description.

truffle/src/main/java/org/jruby/truffle/options/StringArrayOptionDescription.java
```diff
@@ -0,0 +1,131 @@
...
+        } else if (value instanceof String) {
+            return parseStringArray((String) value);
+        } else {
+            throw new OptionTypeException();
+        }
+    }
+
+    // Allows input such as [foo, "bar", 'baz']. Doesn't support escape sequences.
+
+    private String[] parseStringArray(String string) {
```
nirvdrum: Unless I've misunderstood the intent of this method, I think it could go away if we just used something like JSON or YAML for the config file format.

truffle/src/main/java/org/jruby/truffle/options/OptionTypeException.java
```diff
@@ -0,0 +1,14 @@
...
+ * code is released under a tri EPL/GPL/LGPL license. You can use it,
+ * redistribute it and/or modify it under the terms of the:
+ *
+ * Eclipse Public License version 1.0
+ * GNU General Public License version 2
+ * GNU Lesser General Public License version 2.1
+ */
+package org.jruby.truffle.options;
+
+public class OptionTypeException extends UnsupportedOperationException {
```
nirvdrum: I think it'd be helpful for debugging if this exception type took the option name as its argument. Or perhaps we'd just enforce a message passed with the option name.

truffle/src/main/java/org/jruby/truffle/options/UnknownOptionException.java
```diff
@@ -0,0 +1,14 @@
...
+ * code is released under a tri EPL/GPL/LGPL license. You can use it,
+ * redistribute it and/or modify it under the terms of the:
+ *
+ * Eclipse Public License version 1.0
+ * GNU General Public License version 2
+ * GNU Lesser General Public License version 2.1
+ */
+package org.jruby.truffle.options;
+
+public class UnknownOptionException extends UnsupportedOperationException {
```
nirvdrum: As with `OptionTypeException`, I think it'd help debugging a lot of if the option name were required.

truffle/src/main/java/org/jruby/truffle/options/StringArrayOptionDescription.java
nirvdrum: I guess that wouldn't help the argv case, so maybe a value parser is still necessary. It just seems a bit complex.

eregon: Indeed, what is this for? Do we have a use-case to pass a String[] as a String? A comma-separated list is much nicer to parse and to give on the command line: `someOption=foo,bar,baz`.

truffle/src/main/java/org/jruby/truffle/options/OptionsCatalogue.java
chrisseaton: Yes.

core/src/main/java/org/jruby/util/cli/Options.java
chrisseaton: Yes.

truffle/src/main/java/org/jruby/truffle/options/OptionsCatalogue.java
chrisseaton: Done.

truffle/src/main/java/org/jruby/truffle/Options.java
chrisseaton: Yes.

truffle/src/main/java/org/jruby/truffle/Options.java
chrisseaton: Yes.

truffle/src/main/java/org/jruby/truffle/options/OptionsBuilder.java
chrisseaton: I will do when they're all in this system.

truffle/src/main/java/org/jruby/truffle/options/StringArrayOptionDescription.java
chrisseaton: I switched to a simple comma separated format, with escaping for commas if needed. The reason for the string format is so you can set things like Ruby `ARGV` using system properties. Imagine starting a JS node application and for some reason wanting Ruby code called using interop still have an `ARGV`.

truffle/src/main/java/org/jruby/truffle/options/UnknownOptionException.java
chrisseaton: Done.

chrisseaton:
Can people review the generator.

I actually have a problem with my ERB templates giving me extra blank lines. Does anyone know how to fix that?

Other than that, raise any final objections now before I do all the work to switch the current options over and merge.

truffle/src/main/java/org/jruby/truffle/options/OptionsBuilder.java
nirvdrum: I meant is there a reason to avoid `property.getKey().toString()`?

truffle/src/main/java/org/jruby/truffle/options/OptionsBuilder.java
```diff
@@ -0,0 +1,65 @@
...
+import java.util.Map;
+import java.util.Properties;
+
+public class OptionsBuilder {
+
+    private final String LEGACY_PREFIX = "jruby.truffle.";
+
+    private final Map<OptionDescription, Object> options = new HashMap<>();
+
+    public void set(Properties properties) {
```
nirvdrum: I can't readily tell, but I suspect we're going to want a boundary on this.

truffle/src/main/java/org/jruby/truffle/options/StringArrayOptionDescription.java
```diff
@@ -0,0 +1,100 @@
...
+        } else if (value instanceof String) {
+            return parseStringArray((String) value);
+        } else {
+            throw new OptionTypeException(getName());
+        }
+    }
+
+    // Allows input such as foo,bar,baz. You can escape commas.
+
+    private String[] parseStringArray(String string) {
```
nirvdrum: I think we're going to want a boundary on this.

chrisseaton: I don't think so - you can't set options after the engine is created, so it can't be fast path.

tool/truffle/generate-options.rb
```diff
@@ -0,0 +1,108 @@
...
+    <% end %>
+
+    NewOptions(OptionsBuilder builder) {
+        <% options.each do |o| %>
+        <%= o.constant %> = builder.getOrDefault(OptionsCatalog.<%= o.constant %>);
+        <% end %>
+    }
+
+}
+}).result)
```
pitr-ch:
```ruby
File.write('truffle/src/main/java/org/jruby/truffle/options/NewOptions.java', ERB.new(<<-JAVA).result)
/*
 * Copyright (c) 2016 Oracle and/or its affiliates. All rights reserved. This
 * code is released under a tri EPL/GPL/LGPL license. You can use it,
 * redistribute it and/or modify it under the terms of the:
 *
 * Eclipse Public License version 1.0
 * GNU General Public License version 2
 * GNU Lesser General Public License version 2.1
 */
package org.jruby.truffle.options;

// This file is automatically generated by options.yml with 'jt build options'

import com.oracle.truffle.api.CompilerDirectives.CompilationFinal;

import javax.annotation.Generated;

@Generated("tool/truffle/generate-options.rb")
public class NewOptions {

    <% options.each do |o| %>
    <% if o.type.end_with?('[]') %>@CompilationFinal(dimensions=1) <% end %>public final <%= o.type %> <%= o.constant %>;
    <% end %>

    NewOptions(OptionsBuilder builder) {
        <% options.each do |o| %>
        <%= o.constant %> = builder.getOrDefault(OptionsCatalog.<%= o.constant %>);
        <% end %>
    }

}
JAVA
```
gives highlighting in idea

pitr-ch:
```
[13] pry(main)> ERB.new("<%= :a -%>\n", nil, '-').result
=> "a"
```
should help

truffle/src/main/java/org/jruby/truffle/options/OptionTypeException.java
```diff
@@ -0,0 +1,24 @@
...
+
+    private final String name;
+
+    public OptionTypeException(String name) {
+        this.name = name;
```
eregon: This should `super()` and given a meaningful message.

tool/truffle/generate-options.rb
chrisseaton: Wow! Thanks! There's an example great feature, @eregon.

truffle/src/main/java/org/jruby/truffle/options/OptionTypeException.java
chrisseaton: Done.

tool/truffle/generate-options.rb
```diff
@@ -0,0 +1,108 @@
...
+#
+# Eclipse Public License version 1.0
+# GNU General Public License version 2
+# GNU Lesser General Public License version 2.1
+
+require 'ostruct'
+require 'yaml'
+require 'erb'
+
+options = YAML.load_file('truffle/src/main/java/org/jruby/truffle/options/options.yml')
```
chrisseaton: Done.

eregon:
The full PR looks great, thanks!

PR #4338, opened 2016-11-27
Support building on Java 9

chrisseaton:
This was blocked by a few minor issues but these have been resolved now.

@mkristian can we get a release of `polyglot-ruby` so we don't need to use a `SNAPSHOT`?

Then @headius can we consider merging this to master. We can't build on Java 9 without it. I'm not sure if it has any side effects but it seems to work.

### DIFF
--- a/.mvn/extensions.xml
+++ b/.mvn/extensions.xml
@@ -3,6 +3,6 @@
   <extension>
     <groupId>io.takari.polyglot</groupId>
     <artifactId>polyglot-ruby</artifactId>
-    <version>0.1.15</version>
+    <version>0.1.20-SNAPSHOT</version>
   </extension>
 </extensions>
--- a/core/pom.rb
+++ b/core/pom.rb
@@ -166,8 +166,8 @@ project 'JRuby Core' do
           'compilerArgs' => { 'arg' => '-J-Xmx1G' },
           'showWarnings' => 'true',
           'showDeprecation' => 'true',
-          'source' => [ '${base.java.version}', '1.7' ],
-          'target' => [ '${base.javac.version}', '1.7' ],
+          'source' => '${base.java.version}',
+          'target' => '${base.javac.version}',
           'useIncrementalCompilation' =>  'false' ) do
     execute_goals( 'compile',
                    :id => 'anno',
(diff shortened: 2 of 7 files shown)


enebo:
@chrisseaton I see the activations for Java 8.  This won't change the requirement that we need to be java 8 or higher does it?   I know we build our releases using Java 1.8 so truffle can build but we still generate 1.7 bytecode.

chrisseaton:
You're right - this doesn't produce code that works on Java 7 or 8. Back to the drawing board.

mkristian:
so you guys are saying for jdk8 the compiler needs to switch to 1.8 as well - i.e. target/source version needs to done via properties and the activation needs to overwrite the default: 

https://github.com/jruby/jruby/blob/07f50dd4963b82fa746bdda5a00165daa01721b2/core/pom.rb#L175-L176

beside the compile tasks - how can check/verify that the code is doing the right thing ? some jdk8 test working only with jdk8 or something !

chrisseaton:
>  how can check/verify that the code is doing the right thing

We need to check every combination of (7, 8, 9) x (8, 9) - 6x the jobs we have on Travis at the moment.

mkristian:
maybe pick one or two out of these many to make sure the compile tasks did the right thing. of course a complete test coverage is desirable but looks really huge

chrisseaton:
Oh and I should also say that I'd expect there to be test failures on 9 in general. For example Joda does not work well on Java 9 - we've switched in Truffle to Java 8's date/time API because of this. I opened issues with Joda saying that their tests were failing, but they just changed the tests ha ha.

mkristian:
TDD it is or not :)

mkristian:
@chrisseaton I tried to rebase to master and had too many conflicts on truffle side of things. I release the mavengem library, the polyglot-maven is still pending. and I am working on the target/source flags for the compiler. is this branch suppose to be in line with master? i.e. are there any special patches for this branch beside the maven related ones ?

mkristian:
`[ERROR] javac: option -Xbootclasspath/p: not allowed with target 1.9`

this needs to be addressed first if we want to be able to build with jdk9 and target 1.9 !

@headius just a friendly reminder to the unsafe problem ;)

mkristian:
@chrisseaton the latest patch will produce core classes with 52.0 version and truffle classes with 53.0 when compiled with jdk9. once core can be build without extra bootclasspath then we can switch to the 53.0 everywhere. hope this is on the right track.

chrisseaton:
@mkristian yes this is supposed to go into master. Do you want me to try rebasing myself?

mkristian:
@chrisseaton no need, I will make a new branch and cherry-pick these 2-3 commits.

mkristian:
@enebo @headius I do not know whether to continue or not. currently it does not compile with jdk9 I downloaded yesterday.

the current setup is that if you build with jdk8 it uses target/source 1.8 with jdk9 it uses target.source 1.9.

I cleanup the branch and keep it as is until there is indication that the jruby codebase should compile under jdk9 as well. or maybe the current switch in the profile is already nice to have.

PR #4347, opened 2016-11-30
[Truffle] Hash merge check for splatted elements

bjfish:
This should raise a type error:
```
h = {1 => 2, b: 3}
{"a" => "b", **h}
```
As far as I can tell, this ConcatHashLiteralNode is used for merging hashes with the splat `**` operator. I observed that the pair matching the `pair.getKey() == null` statement in the body translator is what is to be merged into the hash literal. These null key pairs need to check that all their keys are symbols after they are executed. So, I added an array of booleans to track which hashes were being merged in to pass to the concat node.

This is done here in MRI:
`core_hash_merge_kwd`
https://github.com/ruby/ruby/blob/22c0994bc8b34e4b8b80de3259363b0f27c24988/vm.c#L2675

### DIFF
--- a/truffle/src/main/java/org/jruby/truffle/core/hash/EnsureSymbolKeysNode.java
+++ b/truffle/src/main/java/org/jruby/truffle/core/hash/EnsureSymbolKeysNode.java
@@ -1,0 +1,34 @@
+package org.jruby.truffle.core.hash;
+
+import com.oracle.truffle.api.frame.VirtualFrame;
+import com.oracle.truffle.api.object.DynamicObject;
+import com.oracle.truffle.api.profiles.BranchProfile;
+import com.oracle.truffle.api.source.SourceSection;
+import org.jruby.truffle.RubyContext;
+import org.jruby.truffle.language.RubyGuards;
+import org.jruby.truffle.language.RubyNode;
+import org.jruby.truffle.language.control.RaiseException;
+
+public class EnsureSymbolKeysNode extends RubyNode {
+
+    @Child private RubyNode child;
+
+    private final BranchProfile errorProfile = BranchProfile.create();
+
+    public EnsureSymbolKeysNode(RubyContext context, SourceSection sourceSection, RubyNode child) {
+        super(context, sourceSection);
+        this.child = child;
+    }
+
+    @Override
+    public Object execute(VirtualFrame frame) {
+        final Object hash = child.execute(frame);
+        for (KeyValue keyValue : HashOperations.iterableKeyValues((DynamicObject) hash)) {
+            if (!RubyGuards.isRubySymbol(keyValue.getKey())) {
+                errorProfile.enter();
+                throw new RaiseException(coreExceptions().typeErrorWrongArgumentType(keyValue.getKey(), "Symbol", this));
+            }
+        }
+        return hash;
+    }
+}
--- a/truffle/src/main/java/org/jruby/truffle/parser/BodyTranslator.java
+++ b/truffle/src/main/java/org/jruby/truffle/parser/BodyTranslator.java
@@ -44,6 +44,7 @@ import org.jruby.truffle.core.cast.ToProcNodeGen;
 import org.jruby.truffle.core.cast.ToSNode;
 import org.jruby.truffle.core.cast.ToSNodeGen;
 import org.jruby.truffle.core.hash.ConcatHashLiteralNode;
+import org.jruby.truffle.core.hash.EnsureSymbolKeysNode;
 import org.jruby.truffle.core.hash.HashLiteralNode;
 import org.jruby.truffle.core.hash.HashNodesFactory;
 import org.jruby.truffle.core.kernel.KernelNodesFactory;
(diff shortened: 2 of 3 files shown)


chrisseaton:
Looks good.

truffle/src/main/java/org/jruby/truffle/parser/BodyTranslator.java
```diff
@@ -1826,14 +1826,17 @@ public RubyNode visitHashNode(HashParseNode node) {
         final SourceSection fullSourceSection = sourceSection.toSourceSection(source);
 
         final List<RubyNode> hashConcats = new ArrayList<>();
+        final List<Boolean> hashConcatsIsMerge = new ArrayList<>();
 
         final List<RubyNode> keyValues = new ArrayList<>();
 
         for (KeyValuePair<ParseNode, ParseNode> pair: node.getPairs()) {
             if (pair.getKey() == null) {
```
eregon: This case with a null key means that we have a ** splat like in `{a: 1, **{b: 2}, c: 3}`.
It would merit a comment for the least!

truffle/src/main/java/org/jruby/truffle/parser/BodyTranslator.java
```diff
@@ -1826,14 +1826,17 @@ public RubyNode visitHashNode(HashParseNode node) {
...
                 final RubyNode hashLiteralSoFar = HashLiteralNode.create(context, fullSourceSection, keyValues.toArray(new RubyNode[keyValues.size()]));
                 hashConcats.add(hashLiteralSoFar);
+                hashConcatsIsMerge.add(false);
                 hashConcats.add(HashCastNodeGen.create(context, fullSourceSection, pair.getValue().accept(this)));
+                hashConcatsIsMerge.add(true);
```
eregon: Why such a complex approach rather than introducing a node above the HashCastNode here that checks keys are all Symbols like `EnsureSymbolKeysNode` ?
Then that would be a single line change here, and no change in ConcatHashLiteralNode.
Don't be afraid to add new nodes.

truffle/src/main/java/org/jruby/truffle/core/hash/ConcatHashLiteralNode.java
```diff
@@ -54,6 +59,13 @@ private Object buildHash(DynamicObject[] parts) {
         for (int i = 0; i < children.length; i++) {
             try {
                 DynamicObject hash = children[i].executeDynamicObject(frame);
+                if (mergeChildren[i]) {
+                    for (KeyValue keyValue : HashOperations.iterableKeyValues(hash)) {
```
eregon: Just a note: this would not be fine with the `@ExplodeLoop` above I think

truffle/src/main/java/org/jruby/truffle/parser/BodyTranslator.java
bjfish: I'll try refactoring to use a new node. I think it would match the behavior I am looking for.

bjfish: @eregon I've pushed an update to use a wrapper node.

truffle/src/main/java/org/jruby/truffle/core/hash/EnsureSymbolKeysNode.java
```diff
@@ -0,0 +1,31 @@
...
+
+    public EnsureSymbolKeysNode(RubyContext context, SourceSection sourceSection, RubyNode child) {
+        super(context, sourceSection);
+        this.child = child;
+    }
+
+    @Override
+    public Object execute(VirtualFrame frame) {
+        final Object hash = child.execute(frame);
+        assert RubyGuards.isRubyHash(hash);
```
eregon: iterateKeyValues will already assert, so not really necessary.

truffle/src/main/java/org/jruby/truffle/core/hash/EnsureSymbolKeysNode.java
```diff
@@ -0,0 +1,31 @@
...
+        for (KeyValue keyValue : HashOperations.iterableKeyValues((DynamicObject) hash)) {
+            if (!RubyGuards.isRubySymbol(keyValue.getKey())) {
+                throw new RaiseException(coreExceptions().typeErrorWrongArgumentType(keyValue.getKey(), "Symbol", this));
```
eregon: You want a BranchProfile here before the `throw` so that part is not compiled until we actually meet a non-Symbol.
The BranchProfile should be a final field of the node.

truffle/src/main/java/org/jruby/truffle/core/hash/EnsureSymbolKeysNode.java
bjfish: I've removed the assert

truffle/src/main/java/org/jruby/truffle/core/hash/EnsureSymbolKeysNode.java
bjfish: I've added a branch profile

PR #4355, opened 2016-12-02
[Truffle] Fix optional assignment specs

nirvdrum:
This fixes all the remaining optional variable assignment specs. I think what I did makes sense, but I'd appreciate at least another pair of eyes looking over the changes.

@jruby/truffle

### DIFF
--- a/truffle/src/main/java/org/jruby/truffle/parser/BodyTranslator.java
+++ b/truffle/src/main/java/org/jruby/truffle/parser/BodyTranslator.java
@@ -2579,10 +2579,14 @@ public class BodyTranslator extends Translator {
         final RubySourceSection sourceSection = translate(node.getPosition());
         final SourceSection fullSourceSection = sourceSection.toSourceSection(source);
 
-        final ParseNode lhs = node.getFirstNode();
-        final ParseNode rhs = node.getSecondNode();
+        RubyNode lhs = node.getFirstNode().accept(this);
+        RubyNode rhs = node.getSecondNode().accept(this);
 
-        final RubyNode andNode = new AndNode(lhs.accept(this), rhs.accept(this));
+        if (lhs instanceof ReadConstantNode && !(rhs instanceof WriteConstantNode)) {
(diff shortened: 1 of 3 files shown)


truffle/src/main/java/org/jruby/truffle/parser/BodyTranslator.java
```diff
@@ -2570,6 +2570,10 @@ public RubyNode visitNthRefNode(NthRefParseNode node) {
 
     @Override
     public RubyNode visitOpAsgnAndNode(OpAsgnAndParseNode node) {
+        return visitOpAsgnAndNode(node, node.getFirstNode().accept(this), node.getSecondNode().accept(this));
+    }
+
+    private RubyNode visitOpAsgnAndNode(ParseNode node, RubyNode lhs, RubyNode rhs) {
```
eregon: I think it's clearer to give it another name like `translateOpAsgnAndNode`.

truffle/src/main/java/org/jruby/truffle/parser/BodyTranslator.java
```diff
@@ -2591,11 +2592,33 @@ public RubyNode visitOpAsgnAndNode(OpAsgnAndParseNode node) {
...
+            case "||": {
+                final RubyNode defined = new DefinedNode(context, translateSourceSection(source, lhs.getRubySourceSection()), lhs);
+                lhs = new AndNode(defined, lhs);
+
+                return visitOpAsgnOrNode(node, lhs, rhs);
+            }
+
+            default: {
+                final RubySourceSection sourceSection = translate(node.getPosition());
+                final RubyCallNodeParameters callParameters = new RubyCallNodeParameters(context, sourceSection.toSourceSection(source), lhs, node.getOperator(), null, new RubyNode[] { rhs }, false, true, false, false, false);
```
eregon: There is a slightly shorter version if the last 3 boolean are false for RubyCallNodeParameters.

Issue #4361, opened 2016-12-06
Unexpected value for `$?.exitstatus` when process exited successfully


javierhonduco:
### Environment

* jruby 9.1.5.0 (2.3.1) 2016-09-07 036ce39 Java HotSpot(TM) 64-Bit Server VM 24.65-b04 on 1.7.0_67-b01 +jit [darwin-x86_64]
* Darwin 16.1.0 Darwin Kernel Version 16.1.0: Wed Oct 19 20:31:56 PDT 2016; root:xnu-3789.21.4~4/RELEASE_X86_64 x86_64

There's nothing special (gems or whatsoever) running on this machine.

### Expected Behavior
When running this snippet of code:
```ruby
system "exit 0"
puts "ruby_success=#{$?.success?} exit_code=#{$?.exitstatus}"
```
MRI (2.3.1) outputs 
```ruby_success=true exit_code=0```
### Actual Behavior
Unfortunately this is the output I'm getting under the JRuby version mentioned above:
```
ruby_success=false exit_code=127
```

I don't know if this is expected as per JRuby `system`'s implementation, sorry if that's the case!
Thanks,

**Comment by olleolleolle** [comment]
Same on 9.1.6.0.

```shell
$ rvm jruby-9.1.6.0 do ruby -e 'system "exit 0"; puts "ruby_success=#{$?.success?} exit_code=#{$?.exitstatus}"'
ruby_success=false exit_code=127
```

headius:
Looks like we're not doing the right masking of the exit status.

headius:
Ok, so here's the problem. We do the correct masking; however in this case the system call is actually failing. It tries to find an executable `exit` command, and seeing none it bails out with exit code of 127. MRI must be doing some extra logic here to know it needs to run with `sh -c`.

javierhonduco:
Thanks a lot guys!

💚

PR #4364, opened 2016-12-06
Add regression test for shell exit

javierhonduco:
Adds a regression test for 69da0af4f5af2615640f5575ef80ba288fb4eb5f reported in in #4361.


kares:
OSS+TDD at its best :) ... thanks Javier!

Issue #3171, opened 2015-07-23
Hang with threads and Hash[]= on unfrozen keys


Freaky:
The following code reliably hangs JRuby 9k.  Note the hashes are _not_ shared between threads:

```
require 'securerandom'

Array.new(4).map do
  Thread.new do
    100_000.times do |i|
      print "#{Thread.current.inspect}: #{i}\n" if i % 1000 == 0
      key = SecureRandom.random_bytes(4)
      hash = {}
      hash[key] = true
    end
  end
end.map(&:join)
```

The hung threads are ending up here:

```
    at java.util.WeakHashMap.get(WeakHashMap.java:403)
    at org.jruby.Ruby.freezeAndDedupString(Ruby.java:4733)
    at org.jruby.RubyHash.op_asetForString(RubyHash.java:1017)
    at org.jruby.RubyHash.fastASetCheckString(RubyHash.java:969)
    at org.jruby.RubyHash.op_aset(RubyHash.java:1006)
```

Calling key.freeze prevents the problem.

Reproduced in:
- jruby 9.0.0.0 (2.2.2) 2015-07-21 e10ec96 OpenJDK 64-Bit Server VM 25.51-b03 on 1.8.0_51-b16 +jit [FreeBSD-amd64]
- jruby 9.0.0.0 (2.2.2) 2015-07-21 e10ec96 Java HotSpot(TM) 64-Bit Server VM 25.51-b03 on 1.8.0_51-b16 +jit [Windows 8.1-amd64]

nirvdrum:
The problem is `WeakHashMap` isn't thread-safe and an internal iterator can be corrupted, causing `WeakHashMap#get` to go into an infinite loop.  The easiest solution is to synchronize the whole thing, but this will incur a small penalty on all reads.

nirvdrum:
@headius @enebo This is a naive fix, by synchronizing all the operations on that map.  It's possible a `ReadWriteLock` could yield better performance if you're expecting mostly read-only entries.  That lock is heavier weight than a mutex, however, so you'd need to make sure it pays for itself.

enebo:
I think we should review this cache if this is the solution...re-opening

kares:
a `org.jruby.util.collections.WeakValuedMap` should provide a `ConcurrentHashMap`-like cache

headius:
[...]

PR #4366, opened 2016-12-06
Optimize getDefinedMessage by deduping during initialization

ivoanjo:
While benchmarking a barebones rails application I was seeing a lot of
lock contention. When I looked into it, all threads were spending a lot
of time waiting to grab the mutex on the `dedupMap` inside
`Ruby.freezeAndDedupString()`.

This happened because activesupport uses
`defined?(@somevar) && @somevar` a lot while filling the
output template, and thus threads were spending a lot of time in the
`Ruby.getDefinedMessage()` method fighting to dedup the same strings
over and over and over again.

To fix this, instead of deduping when they are accessed, let's dedup the
defined messages when they are initially populated inside the
definedMessages map. This way we get the same effect, but only dedup
once.

This fix has a considerable effect on single-thread performance:

```ruby
require 'benchmark/ips'

class Test
  def initialize
    @foo = 'foo'
  end

  def defined_test
    defined?(@foo) && @foo
  end
end

t = Test.new

Benchmark.ips do |benchmark|
  benchmark.time = 20
  benchmark.warmup = 20

  benchmark.report("defined") { t.defined_test }

  benchmark.compare!
end
```

with fix:

```
             defined   256.296k i/100ms
             defined     11.992M (± 4.4%) i/s -    239.380M in  20.008996s
```

without fix:

```
             defined   224.042k i/100ms
             defined      7.373M (± 2.5%) i/s -    147.420M in  20.009003s
```

...but it has an even more noticeable effect on multi-threaded performance:

```ruby
def defined_loop(count)
  @foo = true
  count.times do
    defined?(@foo) && @foo
  end
end

defined_loop(2**27)

puts "Warmup done"

start = Time.now

t = (0..20).map do
  Thread.new do
    defined_loop(2**25)
  end
end

t.each(&:join)

puts "Done in #{Time.now - start} seconds"
```

In this case I wasn't able to find a decent ruby multithreaded benchmark
harness, so pardon the crappiness, but I ran it fine times and got
(on a linux dual-core skylake i7):

with fix:

```
Done in 12.442 seconds
Done in 12.284 seconds
Done in 12.803 seconds
Done in 13.724 seconds
Done in 13.844 seconds
```

and under visualvm it looks like this:

Thread state graph with fix

without fix:

```
Done in 71.088 seconds
Done in 71.664 seconds
Done in 64.57 seconds
Done in 65.504 seconds
Done in 62.81 seconds
```

and under visualvm it looks like this:

Thread state graph without fix

As expected, this is a major win on multithreaded scenarios.

### DIFF
--- a/core/src/main/java/org/jruby/Ruby.java
+++ b/core/src/main/java/org/jruby/Ruby.java
@@ -1283,8 +1283,8 @@ public final class Ruby implements Constantizable {
 
     private void initDefinedMessages() {
         for (DefinedMessage definedMessage : DefinedMessage.values()) {
-            RubyString str = RubyString.newString(this, ByteList.create(definedMessage.getText()));
-            str.setFrozen(true);
+            RubyString str = freezeAndDedupString(
+                RubyString.newString(this, ByteList.create(definedMessage.getText())));
             definedMessages.put(definedMessage, str);
         }
     }
@@ -4839,7 +4839,7 @@ public final class Ruby implements Constantizable {
     }
 
     public RubyString getDefinedMessage(DefinedMessage definedMessage) {
-        return freezeAndDedupString(definedMessages.get(definedMessage));
+        return definedMessages.get(definedMessage);
     }
 
     public RubyString getThreadStatus(RubyThread.Status status) {


kares:
Excellent Ivo, this is certainly merge material.
Throwing in some discussion about revisiting `dedupMap = Collections.synchronizedMap(new WeakHashMap<RubyString, WeakReference<RubyString>>());` I believe this should be further improved by not using a locking map impl -> possibly replacing the cache with a concurrent map.

... would certainly not want to trade locking for less memory usage with frozen string literals. thoughts?

there seems to be a note on that already around `freezeAndDedupString` : 
```
     * Note that this cache does some sync against the Ruby instance. This
     * could cause contention under heavy concurrent load, so a reexamination
     * of this design might be warranted.
```

ivoanjo:
Fully agreed @kares.

I had that jotted down on my notebook as something to be explored next, and it wouldn't even be the first time I'd write a concurrent hash map with weak keys (you need some extra mojo to remove the weak references after they get collected).

If you think that's the way to go, I'm willing to take a stab at it :)

kares:
@ivoanjo ah right `WeakConcurrentMap` - probably the reason why its not there in the first place.
believe it makes sense (if its not too much work on top of the JDK concurrent map) ... at `org.jruby.util`

the sync was introduced in https://github.com/ivoanjo/jruby/commit/e88911e6813acc5f734e7d16447fe4d07a9e7b91 as a bug work-around, you should look at https://github.com/jruby/jruby/issues/3171 for details

headius:
Wow, this is a great find! Could explain slower-than-expected performance other folks have reported. Thank you! Looking forward to other discoveries :-)

headius:
I pushed an additional improvement that caches the defined? messages inline. This avoids the round-trip through the runtime and the defined messages map entirely.

We still would like to de-bottleneck the dedup cache, of course, since it's used many other places.

headius:
My numbers with the additional inline-caching tweak and invokedynamic enabled:

Before:
After:

We can further improve these once we improve defined? to cache its lookups (like method calls do).

nirvdrum:
Wow. Apologies if that #3171 fix was too blunt. But it seems like an easy performance win to fix then.

ivoanjo:
@headius if I may ask a probably-derp question about your approach :smile: 
It obviously is much better to just have a pointer for the ruby string containing the message, rather than going the roundabout way of getting it from the `EnumMap`.

When I was cooking the patch, I investigated a bit and the reason why there was a `freezeAndDedup` there in the first place, was, according to b3fb6e42f7c076377655fb9d7ef012a924ee59bb, so that "We need to dedup defined message strings so they all have same id", which makes sense to me.

But your approach always seems to create new `FrozenString` instances, so intuitively it would seem to me that this would not hold. But I tried experimenting with it, and it does hold, I still get the same `object_id` every time with your patch. So... why? I'm scratching my head a bit :)

ivoanjo:
@kares I've been trying to see where `freezeAndDedupString` is still being used in order to devise a benchmark that can then be used to test an alternative approach to the current weakhashmap and it seems to me that after the change in this PR, there aren't any performance-critical uses of it anymore, so the added complexity may not be entirely worth it.

Am I missing some use case?

kares:
@ivoanjo thought that the map is a cache for eventually all frozen strings (haven't looked into the details)

ivoanjo:
@kares I don't think frozen strings are deduped by default, e.g.

```ruby
jruby-9.1.5.0 :001 > x = 97.chr.freeze
 => "a" 
jruby-9.1.5.0 :002 > y = 'a'.freeze
 => "a" 
jruby-9.1.5.0 :003 > x.object_id
 => 2002 
jruby-9.1.5.0 :004 > y.object_id
 => 2004 
```

I think they are only deduped when they are read in as literals:

```ruby
jruby-9.1.5.0 :001 > x = 'a'.freeze
 => "a" 
jruby-9.1.5.0 :002 > y = 'a'.freeze
 => "a" 
jruby-9.1.5.0 :003 > x.object_id
 => 2002 
jruby-9.1.5.0 :004 > y.object_id
 => 2002 
```

This is why I'm having a bit of a difficulty in coding a benchmark to test this feature.

PR #4368, opened 2016-12-08
Implemented Hash#compact and Hash#compact!

herwinw:
It passes the TestHash#test_compact method of the MRI test.


headius:
Nice, thanks!

PR #4371, opened 2016-12-09
[Truffle] Attempt at removing core from mx.

chrisseaton:


mx.jruby/mx_jruby.py
```diff
@@ -203,6 +107,10 @@ def extractArguments(cli_args):
                 rubyArgs.append(arg)
                 rubyArgs.extend(args)
                 break
+    if classic:
+        main_class = "org.jruby.Main"
+    else:
+        main_class = "org.jruby.truffle.Main"
```
eregon: classic will not launch anymore if the jar is not on the classpath, so probably this launcher should not support classic anymore.

mx.jruby/suite.py
```diff
@@ -143,21 +222,10 @@ def mavenLib(mavenDep, sha1, sourceSha1, license):
             "license": "EPL-1.0",
         },
 
-        # Depends on jruby-core extracting jni libs in lib/jni
-        "jruby-lib-jni": {
```
eregon: I thought the native *.so needed to be real files on the filesystem, is this not required?

mx.jruby/mx_jruby.py
chrisseaton: Fixed.

PR #4372, opened 2016-12-09
[Truffle] Privatise ByteList

chrisseaton:
We can drop logic such as invalidating the hash, and space such as holding the hash and string value, if we privatise this.

### DIFF
--- a/truffle/src/main/java/org/jruby/truffle/util/ByteList.java
+++ b/truffle/src/main/java/org/jruby/truffle/util/ByteList.java
@@ -49,8 +49,7 @@ import org.jcodings.specific.ASCIIEncoding;
  * of characters. However, its API resembles StringBuffer/StringBuilder more than String
  * because it is a mutable object.
  */
-public class ByteList implements Comparable, CharSequence, Serializable {
-    private static final long serialVersionUID = -1286166947275543731L;
+public class ByteList implements Comparable, CharSequence {
 
     public static final byte[] NULL_ARRAY = new byte[0];
     public static final ByteList EMPTY_BYTELIST = new ByteList(0);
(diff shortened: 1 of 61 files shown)


truffle/src/main/java/org/jruby/truffle/util/ByteList.java
```diff
@@ -866,17 +861,11 @@ public int hashCode() {
...
      * @return an ISO-8859-1 representation of the byte list
      */
     @Override
     public String toString() {
-        String decoded = this.stringValue;
-        if (decoded == null) {
-            this.stringValue = decoded = decode(bytes, begin, realSize, "ISO-8859-1");
-        }
-        return decoded;
+        return decode(bytes, begin, realSize, "ISO-8859-1");
```
eregon: `"ISO-8859-1"` is wrong here (it should just use the ByteList encoding to decode), so we could take this opportunity to fix it as well.

nirvdrum:
If we have our own copy of `ByteList`, we may as well remove all the unused methods.

truffle/src/main/java/org/jruby/truffle/util/ByteListKey.java
```diff
@@ -0,0 +1,39 @@
...
+    }
+
+    @Override
+    public boolean equals(Object o) {
+        if (this == o) return true;
+        if (o == null || getClass() != o.getClass()) return false;
+
+        ByteListKey that = (ByteListKey) o;
+
+        return bytes.equals(that.bytes);
```
nirvdrum: Note that `ByteList#equals` ignores encoding. This has been problematic in the past with things like our symbol table.

PR #4376, opened 2016-12-10
[Truffle] Run irb instead of reading from stdin by default.

chrisseaton:
This continually confuses non-Ruby users who just run `ruby` and expect to see something.

I could be convinced to not commit this.

### DIFF
--- a/truffle/src/main/java/org/jruby/truffle/options/RubyInstanceConfig.java
+++ b/truffle/src/main/java/org/jruby/truffle/options/RubyInstanceConfig.java
@@ -93,8 +93,12 @@ public class RubyInstanceConfig {
     }
 
     public void processArguments(String[] arguments) {
-        new ArgumentProcessor(arguments, this).processArguments();
+        final ArgumentProcessor processor = new ArgumentProcessor(arguments, this);
+        processor.processArguments();
         tryProcessArgumentsWithRubyopts();
+        if (getInlineScript() == null && getScriptFileName() == null && System.console() != null) {
+            setScriptFileName(processor.resolveScript("irb"));
+        }
     }
 
     public void tryProcessArgumentsWithRubyopts() {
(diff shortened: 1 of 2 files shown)


chrisseaton:
This is a really crazy set of failures for such a small change.

eregon:
Maybe `System.console()` is not working as expected or there is some bug?
The laguage specs fail for instance with:
```
Global variable $?
- is read-only
- is thread-localirb(main):001:0> 
```

And it's just doing:
```bash
some/ruby -e "exit 0"
```

truffle/src/main/java/org/jruby/truffle/options/RubyInstanceConfig.java
```diff
@@ -96,8 +96,12 @@ private void initEnvironment() {
     }
 
     public void processArguments(String[] arguments) {
-        new ArgumentProcessor(arguments, this).processArguments();
+        final ArgumentProcessor processor = new ArgumentProcessor(arguments, this);
+        processor.processArguments();
         tryProcessArgumentsWithRubyopts();
+        if (getScriptFileName() == null && System.console() != null) {
```
eregon: You probably need to check for the case with `-e` here.

enebo:
Does it still work as expected if you pipe input to the process?

```shell
cat mydata | jruby -ne '...'
```

chrisseaton:
Yes - that's what the `System.console() != null` check is for. It should be `null` if `stdin` is not connected directly to a terminal.

enebo:
Ok I guess I was not aware one could tell the difference. If so I don't think anyone intentionally types Ruby without a pipe or redirection

chrisseaton:
> I don't think anyone intentionally types Ruby without a pipe or redirection

This is actually in response to user testing with real people doing that and then sitting there staring at it for ages waiting for it to do something. They tend to be non Ruby programmers just wanting to experiment with Ruby like they do Python.

enebo:
Ah I came from awk, Perl, and all the languages which do sit there. I agree that if you can know this case it seems like a good idea to enter a repl. It might be worth also creating an RFE on Ruby lang

PR #4379, opened 2016-12-12
Fix bigdecimal infinity round

kirs:
Expected: MRI can `round(0)` Infinity:
```
ruby 2.3.1p112 (2016-04-26 revision 54768) [x86_64-darwin16]

BigDecimal("Infinity").round(0)
=> #<BigDecimal:7fad7f83d798,'Infinity',9(18)>
```

But JRuby-master raises an error:

```
jruby 9.1.7.0-SNAPSHOT (2.3.1) 2016-12-12 fd9515b Java HotSpot(TM) 64-Bit Server VM 25.45-b02 on 1.8.0_45-b14 +jit [darwin-x86_64]

BigDecimal("Infinity").round(0)
FloatDomainError: Computation results to 'Infinity'
    org/jruby/ext/bigdecimal/RubyBigDecimal.java:1368:in `round'
```

We raised an error when the method was called with no argument, but we checked it in the wrong way so `round(0)` was broken.

@enebo @headius


headius:
Looks ok to me!

PR #4380, opened 2016-12-13
[Truffle] Sprintf check for too many args when $DEBUG is true

bjfish:
Opening PR for feedback

### DIFF
--- a/truffle/src/main/java/org/jruby/truffle/core/format/printf/PrintfSimpleParser.java
+++ b/truffle/src/main/java/org/jruby/truffle/core/format/printf/PrintfSimpleParser.java
@@ -19,10 +19,12 @@ public class PrintfSimpleParser {
 
     private final char[] source;
     private final Object[] arguments;
+    private final boolean isDebug;
 
-    public PrintfSimpleParser(char[] source, Object[] arguments) {
+    public PrintfSimpleParser(char[] source, Object[] arguments, boolean isDebug) {
         this.source = source;
         this.arguments = arguments;
+        this.isDebug = isDebug;
     }
 
     @SuppressWarnings("fallthrough")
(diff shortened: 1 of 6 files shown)


truffle/src/main/java/org/jruby/truffle/core/format/printf/PrintfSimpleParser.java
```diff
@@ -255,6 +264,13 @@ public PrintfSimpleParser(char[] source, Object[] arguments) {
                 }
             }
         }
+        if((argType == ArgType.UNNUMBERED || argType == ArgType.NONE) &&
+           arguments.length > argCount ){
+            if(isDebug){
```
eregon: The spacing around here looks inconsistent with the rest.

eregon:
Looks good overall!

truffle/src/main/java/org/jruby/truffle/core/kernel/KernelNodes.java
```diff
@@ -1849,17 +1864,25 @@ private DynamicObject finishFormat(int formatLength, BytesResult result) {
...
-                        .compile(Layouts.STRING.getRope(format).getBytes(), arguments);
+                        .compile(Layouts.STRING.getRope(format).getBytes(), arguments, isDebug);
             } catch (InvalidFormatException e) {
                 throw new RaiseException(coreExceptions().argumentError(e.getMessage(), this));
             }
         }
 
+        protected boolean isDebug(VirtualFrame frame) {
+            try {
+                return readDebugGlobalNode.executeBoolean(frame);
```
eregon: `BooleanCastNode. executeBoolean(frame, value)`.
BTW, we could remove the frame there, it's not needed.

PR #4381, opened 2016-12-13
Remove unsafe-mock from our build altogether.

### LINKED ISSUE #4103: JRuby fails in building from source when using openjdk-7-jdk trusty package

headius:
This commit replaces the unsafe-mock dependency -- which caused
various build problems living in boot classpath -- with a new
library that just shims the three Unsafe methods added in Java 8:
fullFence, loadFence, and storeFence. This allows us to eliminate
all calls to those methods from JRuby proper as well as the build
tricks.

Fixes #4103.

### DIFF
--- a/core/pom.rb
+++ b/core/pom.rb
@@ -13,9 +13,6 @@ project 'JRuby Core' do
               'tzdata.version' => '2013d',
               'tzdata.scope' => 'provided',
 
-              'unsafe.version' => '8.92.1',
-              'unsafe.jar' => '${settings.localRepository}/com/headius/unsafe-mock/${unsafe.version}/unsafe-mock-${unsafe.version}.jar',
-
               'maven.build.timestamp.format' => 'yyyy-MM-dd',
               'maven.test.skip' => 'true',
               'build.date' => '${maven.build.timestamp}',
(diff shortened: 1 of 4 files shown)


headius:
@enebo Any objections? It does not add a dependency, but it is a new library (of one class) and it does get shaded in. I think it's worth it to eliminate the tweaky bootclasspath flags we've been passing.

Issue #3810, opened 2016-04-19
Maximum stack depth appears to have worsened in 9k


headius:
Tested against JRuby 9.1, but JRuby 9.0.x likely also suffer from this.

I ran some numbers to test stack depth for #3741. We do poorly in JRuby 9.1 and should try to improve this.

All on Java 8u60, with the following code:

```
def foo(a)
  if a > 0
    foo(a - 1)
  else
    puts :success
  end
end
foo(N)
```

Here's interpreter. In 9.1 it is only the simple interpreter.

|  | JRuby 1.7.24 | JRuby 9.1 |
| --- | --- | --- |
| Default flags | 650 | 700 |
| with -Xss2048k | 1350 | 1250 |

And with normal jit settings:

|  | JRuby 1.7.24 | JRuby 9.1 |
| --- | --- | --- |
| Default flags | 1450 | 800 |
| with -Xss2048k | 3400 | 1700 |

This difference in 9.1 is a bit of a worry. The interpreter seems about the same (I'd hope it would be better, even though this is a simple AST). The compiler is significantly worse. A general reason for this may be additional frames around our IR interpreted and compiled paths that could be collapsed. The compiler may suffer from invokedynamic "lambda forms" bloating the stack for early calls.

headius:
With a simpler recursive method, current numbers for 9.1.3.0 and 1.7:

```ruby
def foo(a)
  puts a
  foo(a + 1)
end

foo(1)
```

9.1.3.0 int: 1715-1740
1.7.25 int: 1757
9.1.3.0 jit: 1862
1.7.25 jit: 3498

The reduction in depth in the JIT is likely due to the way we're using call sites. In order to reduce the amount of bytecode in any one method body, the lazy construction and use of a CallSite is done in a synthetic method for every method call. This adds one frame compared to 1.7.

Here's the relevant traces for 9k and 1.7, interpreter and JIT:

9.1.3.0 int stack:
```
	at org.jruby.internal.runtime.methods.InterpretedIRMethod.INTERPRET_METHOD(InterpretedIRMethod.java:157)
	at org.jruby.internal.runtime.methods.InterpretedIRMethod.call(InterpretedIRMethod.java:148)
	at org.jruby.internal.runtime.methods.DynamicMethod.call(DynamicMethod.java:197)
	at org.jruby.runtime.callsite.CachingCallSite.call(CachingCallSite.java:161)
[...]

PR #4382, opened 2016-12-13
Improve stack depth

headius:
These commits add more call paths through a few of our DynamicMethod subclasses for Ruby methods, to reduce the number of stack frames they consume. This improves JIT stack depth for the examples in #3810 by nearly 50%, with a somewhat smaller gain for the interpreter (which has more frames).

### DIFF
--- a/core/src/main/java/org/jruby/internal/runtime/methods/CompiledIRMetaClassBody.java
+++ b/core/src/main/java/org/jruby/internal/runtime/methods/CompiledIRMetaClassBody.java
@@ -7,6 +7,7 @@ import org.jruby.parser.StaticScope;
 import org.jruby.runtime.ArgumentDescriptor;
 import org.jruby.runtime.Block;
 import org.jruby.runtime.DynamicScope;
+import org.jruby.runtime.Helpers;
 import org.jruby.runtime.ThreadContext;
 import org.jruby.runtime.Visibility;
 import org.jruby.runtime.builtin.IRubyObject;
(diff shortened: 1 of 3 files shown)


core/src/main/java/org/jruby/internal/runtime/methods/CompiledIRMetaClassBody.java
```diff
@@ -7,33 +7,67 @@
...
 
 public class CompiledIRMetaClassBody extends CompiledIRMethod {
-    private final boolean pushNewDynScope;
-    private final boolean popDynScope;
+    private final boolean scope;
 
     public CompiledIRMetaClassBody(MethodHandle handle, IRScope scope, RubyModule implementationClass) {
         super(handle, scope, Visibility.PUBLIC, implementationClass, scope.receivesKeywordArgs());
 
-        boolean reuseParentDynScope = scope.getFlags().contains(IRFlags.REUSE_PARENT_DYNSCOPE);
```
enebo: This simplification confuses me.  Did we never reuse dyn scopes?

kares:
+1 would be interested in the actual numbers you observed ... "nearly 50%" compared to which in #3810 :)

headius:
On the simple recursion benchmark, I got anywhere from 1800 to 2000 without the patch and 2500-2600 with it. That's not quite 50% but it's pretty good. The variability may be due to the JVM's background jit swapping in more efficient versions of the methods on the stack earlier or later. Another side effect of JIT.

A version of the bench that runs ten times, reporting the last stack depth, gives the following numbers

JRuby 9.1.6.0: 5671
JRuby master: 10399
JRuby 9.1.6.0 -X-C: 2969
JRuby master -X-C: 2924

I can get these longer-running numbers to be fairly consistent if I disable all background JIT in the JVM with `-Xbatch`. We are definitely beholden to the JIT when it comes to stack depth too.

headius:
Note that these only improve the stack depth of calling Ruby methods; it does not do anything specific to improve calls to native Java code or through Java integration, block/proc dispatches through yield or call, or reflected invocations like send and method.call. There's more work to do.

PR #4385, opened 2016-12-14
[Truffle] undefined is now the same as NotProvided.INSTANCE.

eregon:
@jruby/truffle Please share your opinion.

I think the code is in general better with these changes, `undefined` can't get away and can only be used as a marker for "not provided argument" which is the main purpose.

The only case which might be worth supporting would be to allow it is `case` statements, which could be done more easily with something like:
```ruby
case arg
when String
  ...
when Undefined
  ...
end
```
and then Undefined would have a special `===` that only matches the right instance.
I don't think it's worth it, usually logic is just duplicated for the undefined case, so it might as well be normalized before, with `arg = sensible_default if undefined.equal?(arg)`


truffle/src/main/ruby/core/array.rb
```diff
@@ -359,12 +362,12 @@ def fill_internal(a=undefined, b=undefined, c=undefined)
...
       return self if right <= left           # Nothing to modify
 
-    elsif one and !undefined.equal?(one)
+    elsif one
       left = Rubinius::Type.coerce_to_collection_length one
       left += size if left < 0
       left = 0 if left < 0
 
-      if two and !undefined.equal?(two)
+      if !undefined.equal?(two) and two
```
chrisseaton: Why `and` rather than `&&`?

truffle/src/main/java/org/jruby/truffle/core/CoreLibrary.java
```diff
@@ -278,7 +279,7 @@
     private final GlobalVariables globalVariables;
     private final DynamicObject mainObject;
     private final DynamicObject nilObject;
-    private final DynamicObject rubiniusUndefined;
+    private final Object rubiniusUndefined;
```
nirvdrum: Probably out of scope, but I think we can rename this now.

truffle/src/main/ruby/core/complex.rb
```diff
@@ -53,9 +53,10 @@ class Complex < Numeric
   undef_method :i
 
   def self.convert(real, imag = undefined)
-    if real.equal?(nil) || imag.equal?(nil)
+    if nil.equal?(real) || nil.equal?(imag)
```
nirvdrum: Since you modified the order here, is there any reason to not just use `real.nil?` and `imag.nil?` ?

truffle/src/main/ruby/core/encoding.rb
```diff
@@ -141,7 +141,7 @@ class Converter
     def self.asciicompat_encoding(string_or_encoding)
       encoding = Rubinius::Type.try_convert_to_encoding string_or_encoding
 
-      return if not encoding or undefined.equal? encoding
+      return unless encoding
```
nirvdrum: Doesn't `BooleanCastNode` need to be updated to handle the `undefined` object in order for this to work?

truffle/src/main/ruby/core/io.rb
```diff
@@ -588,13 +588,12 @@ def self.foreach(name, separator=undefined, limit=undefined, options=undefined)
 
     name = Rubinius::Type.coerce_to_path name
 
+    separator = $/ if undefined.equal?(separator)
```
nirvdrum: It probably doesn't matter all that much, but this seems to do extra processing that we didn't have before and doesn't seem to make things all that much clearer.

nirvdrum: Nevermind. I see this is the `case` issue you mentioned in the PR description.

truffle/src/main/java/org/jruby/truffle/core/CoreLibrary.java
eregon: Yes, some cleanup can be done after, since we can get rid of this whole field and guard.

truffle/src/main/ruby/core/complex.rb
eregon: they might be `undefined`, and calling methods on `undefined` does not work, on purpose.

eregon: Even though calling `literal.equal?(some_var)` is less nice, it's actually better for the semantics, otherwise you could invoke a method on a user object (more important for `==` than `equal?` which is rarely redefined). We could play some tricks in core, but I'd rather avoid too much magic unless needed.

truffle/src/main/ruby/core/encoding.rb
eregon: `undefined` used to be "truthy" so I did not dare to accept it in BooleanCastNode.

truffle/src/main/ruby/core/complex.rb
nirvdrum: Okay. This seems like a rule someone (probably me) is bound to forget.

eregon: Yeah, there are a few places in core where it's not like this.
For `undefined` you can't forget as it will trigger an assert if you write it in the other way :)

truffle/src/main/ruby/core/encoding.rb
nirvdrum: Wow. I would have guessed it was "falsey."

truffle/src/main/ruby/core/array.rb
pitr-ch: afaik: Original intention is that `&&` is a logical operator and `and` is a control flow operator. They are not meant to be interchangeable. Following is interchangeable, although the first is much more common.
```ruby
return :nothing if test.nil?
test.nil? and return :nothing
```

eregon: Just my opinion: second one looks like Perl and `and` is needed for assignment in conditionals.

pitr-ch: Convention is to add parenthesis `if (a = true) && b.nil?` to make the assignment stand out more. I like `test or raise something` form since it puts the test first not the unimportant exception as in `raise something unless test`. Anyway this is not important, it's opinionated and gain is just minor readability improvements.

PR #4386, opened 2016-12-14
[Truffle] Improved exec

nirvdrum:


bjfish:
@nirvdrum I think there are some spec failures on Mac. Here is my console output:
https://gist.github.com/bjfish/d70799f5362123df458d3b35671a6618

nirvdrum:
Hmm . . . I'll have to get access to a Mac to see what's going on. Thanks for the heads up.

nirvdrum:
@bjfish Oops. When I added my comments, I accidentally deleted a line of legitimate code. Good thing I didn't use `[skip ci]`.

truffle/src/main/ruby/core/process_mirror.rb
```diff
@@ -267,6 +269,68 @@ def convert_env_value(value)
...
+            env.merge! Hash[add_to_env]
+          end
+
+          @env_array = env.map { |k, v| "#{k}=#{v}" }
+
+          if @options
+            pgroup = options[:pgroup]
+            if pgroup
+              Truffle::POSIX.setpgid(0, pgroup)
+            end
```
eregon: Style note: I am fine with assignment in a `if` like `if pgroup = options[:pgroup]`, but do as you prefer.

truffle/src/main/ruby/core/process_mirror.rb
```diff
@@ -267,6 +269,68 @@ def convert_env_value(value)
           "a+" => ::File::RDWR   | ::File::APPEND | ::File::CREAT
         }
 
+        def spawn_setup
+          require 'fcntl'
+
+          env = options&.delete(:unsetenv_others) ? {} : ENV.to_hash
```
eregon: I think `options` should always be a Hash. Unfortunately it seems the constructor doesn't do that, even though its documentation says it does:
```
         # Assigns @environment, @command, @argv, @redirects, @options. All
         # elements are guaranteed to be non-nil. When no env or options are
         # given, empty hashes are returned.
```
Could you fix the constructor?

truffle/src/main/ruby/core/process_mirror.rb
```diff
@@ -109,6 +110,7 @@ def initialize(env_or_cmd, *args)
 
             @command = command
             @argv = argv
+            @env_array = []
```
eregon: I would not initialize it, so it remains nil until it's properly set and not empty.

truffle/src/main/ruby/core/process_mirror.rb
```diff
@@ -283,10 +347,70 @@ def spawn(options, command, arguments)
...
-        def exec(command, args)
-          Truffle.primitive :vm_exec
-          raise PrimitiveFailure,
-            "Rubinius::Mirror::Process::Execute#exec primitive failed"
+        def exec(command, args, env_array)
+          command.each_byte do |b|
+            if b == 0x00
+              raise ArgumentError.new("string contains null byte")
+            end
+          end
```
eregon: Rubinius::Type.check_null_safe

truffle/src/main/ruby/core/process_mirror.rb
```diff
@@ -283,10 +347,70 @@ def spawn(options, command, arguments)
...
+            split_command, *split_args = command.strip.split(' ')
+
+            if should_search_path?(split_command)
+              resolved_command = resolve_in_path(split_command)
+
+              if resolved_command
+                command, args = '/bin/sh', ['/bin/sh', '-c', command]
+              else
+                raise Errno::ENOENT.new("No such file or directory - #{command}")
+              end
```
eregon: Can't we let the shell do the lookup in $PATH in all cases?

nirvdrum: Do you mean by running `/usr/bin/env`? Otherwise, in one of these cases we're not supposed to run in the shell at all. In the other case, we'd end up with a different error than MRI (and the specs check for that).

truffle/src/main/ruby/core/process_mirror.rb
nirvdrum: Thanks. I didn't know about that one.

truffle/src/main/ruby/core/process_mirror.rb
nirvdrum: I can adapt. I've just been burned too many times by `=` vs `==` in other environments. I think Java gets this right.

truffle/src/main/ruby/core/process_mirror.rb
nirvdrum: Thanks. This was leftover from when I used to bail early if `@options` wasn't defined. Setting that in the constructor fixes everything.

truffle/src/main/ruby/core/process_mirror.rb
nirvdrum: Done.

truffle/src/main/ruby/core/process_mirror.rb
eregon: Somehow I adapted and don't recall having an unintentional `=` in a `if` for ages. I think it's not the worse mistake and relatively easy to catch.

truffle/src/main/ruby/core/process_mirror.rb
eregon: I meant by just giving the command to `/bin/sh`. But indeed if the error must be different than the shell error it's better this way. We need the search for the other case anyway.

Issue #4266, opened 2016-11-07
JRuby unable to find methods unless jar is in boot classpath


jmiettinen:
### Environment

* JRuby 9.1.5.0 or 1.7.23 on OS X (`Darwin jmiettinen.local 14.5.0 Darwin Kernel Version 14.5.0: Sun Sep 25 22:07:15 PDT 2016; root:xnu-2782.50.9~1/RELEASE_X86_64 x86_64`)
* No gems, no `JAVA_OPTS`
* To reproduce, clone [test-case repository](https://github.com/jmiettinen/jruby-test) and run `run.sh` which runs both failing and succeeding case. 
* JRuby-complete-jar is used to run `org.jruby.Main` which then runs a JRuby-script calling a few Java-classes.

### Expected Behavior

When running a simple [script](https://github.com/jmiettinen/jruby-test/blob/master/src/main/ruby/test.rb) with just having `JRuby` in the classpath, not in boot classpath, I would expect the JRuby to correctly resolve Java-methods and call them.

So in that case, the script should print out results of two different builder implementations:
```
SimpleClassImpl/5
SimpleClassImpl/5
BrokenClassImpl/5
BrokenClassImpl/5
```

### Actual Behavior

What actually happens is that calls succeed when the target Java-class is publicly visible, the calls succeed, but JRuby cannot find the target when implementor of the method is package-private.

```
SimpleClassImpl/5
SimpleClassImpl/5
NoMethodError: undefined method `set_value' for #<Java::FiRelex::BrokenClass::Builder:0x6fa0450e>
  <main> at src/main/ruby/test.rb:9
```

mkristian:
to get your test case to the point where it fails I applied follow patch
```diff
diff --git a/src/main/java/fi/relex/SimpleClassImpl.java b/src/main/java/fi/relex/SimpleClassImpl.java
index 4ea0874..db6978c 100644
--- a/src/main/java/fi/relex/SimpleClassImpl.java
+++ b/src/main/java/fi/relex/SimpleClassImpl.java
@@ -1,10 +1,10 @@
 package fi.relex;

-public class SimpleClassImpl extends SimpleClass {
+class SimpleClassImpl extends SimpleClass {

     private final long value;

-    public SimpleClassImpl(long value) {
+    SimpleClassImpl(long value) {
         this.value = value;
     }
```
[...]

PR #4389, opened 2016-12-15
Check permission to AccessibleObject#setAccessible(boolean) a better way

jmiettinen:
In #4266, we noticed a difference what methods JRuby finds in Java-classes between JRuby-jar being in boot classpath and in the normal classpath. This change removes that difference.

The reason for the difference is that all classes in boot classpath are privileged and `AccessController#checkPermission(Permission)` will always return `true` to them, whereas for others the result depends on security policy in effect.

In normal (at least Oracle) JVM installations, the policy doesn't allow `suppressAccessChecks`, _but_ call to `AccessibleObject#setAccessible(boolean)` will still succeed as there's no `SecurityManager` set.

I posit that trying to call `setAccessible` is a better way to see if we might later on succeed in calling `setAccessible` than checking the permission.

### DIFF
--- a/core/src/main/java/org/jruby/javasupport/DummyForJavaUtil.java
+++ b/core/src/main/java/org/jruby/javasupport/DummyForJavaUtil.java
@@ -1,0 +1,11 @@
+package org.jruby.javasupport;
+
+/**
+ * This class exists only for access-checks to set {@linkplain JavaUtil#CAN_SET_ACCESSIBLE}
+ */
+public class DummyForJavaUtil {
+
+    private static final Object PRIVATE = new Object[0];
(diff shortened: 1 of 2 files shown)


headius:
I like the idea, but perhaps we should test it against a class in JRuby, so we're not modifying permissions for a globally-visible class. Perhaps make an Integer look-alike static inner class and use that?

jmiettinen:
Calling `Field.setAccessible(boolean)` only overrides the accessibility checks for that `Field`-object. 

It is, however, better to do checks for code that's under our control. I don't know what would be the best place to insert such class so I dumped it into same package for now.

headius:
This looks fine to me...I'll merge. Thanks!

PR #4390, opened 2016-12-15
Specific arity jit cleanup

headius:
This PR cleans up method jit by only emitting one body regardless of whether there's a specific-arity path or not. Previously, both variable and specific paths were emitted in full. With this PR, the variable path will call the specific path. This reduces bytecode generated for a given method body by up to 50%, including all contained blocks and metaclass bodies.


headius:
This PR depends on #4388.

Issue #4304, opened 2016-11-16
Singleton on non-persistent Java type Java::JavaLang::NullPointerException


tobymurray-nanometrics:
We just upgraded from 9.1.2.0 to 9.1.6.0, and we saw a CI build failure with the exception below. It's only failed once so far, but we like to stay on top of these things. Look familiar at all? Anything we can do to provide more helpful logs/details? I don't have a reproducible case, so I'm just looking for some guidance as to whether you think this is an us problem or a JRuby 9.1.6.0 problem. 

### Environment
JRuby 1.9.6.0
Rails 4.2.7.1

```
openjdk version "1.8.0_111"
OpenJDK Runtime Environment (build 1.8.0_111-8u111-b14-2~bpo8+1-b14)
OpenJDK 64-Bit Server VM (build 25.111-b14, mixed mode)
```
```
Linux 468c25bf6746 4.4.0-47-generic #68-Ubuntu SMP Wed Oct 26 19:39:52 UTC 2016 x86_64 GNU/Linux
```

### Command executed
```
[INFO] Running precompile assets with command java -Xms2g -Xmx2g -jar /root/.m2/repository/org/jruby/jruby-complete/9.1.6.0/jruby-complete-9.1.6.0.jar -S /opt/atlassian/pipelines/agent/build/athena/.gems/bin/rake assets:precompile
```

### Stacktrace 
```
I, [2016-11-16T16:48:29.889000 #387]  INFO -- : Writing /opt/atlassian/pipelines/agent/build/athena/public/assets/pages/threshold_statuses/seismicity_plot-06f1df67a58d8f1dcfeddae2bab4912e0248623dce7f1b2b7927dffb0d1e13ce.js.gz
/opt/atlassian/pipelines/agent/build/athena/.gems/gems/rake-11.3.0/lib/rake/task.rb:197: warning: singleton on non-persistent Java type Java::JavaLang::NullPointerException (http://wiki.jruby.org/Persistence)
rake aborted!
Java::JavaLang::NullPointerException: 
org.jruby.ir.dataflow.DataFlowProblem.compute_MOP_Solution(DataFlowProblem.java:56)
org.jruby.ir.passes.LiveVariableAnalysis.execute(LiveVariableAnalysis.java:77)
org.jruby.ir.passes.CompilerPass.run(CompilerPass.java:111)
org.jruby.ir.passes.CompilerPass.makeSureDependencyHasRunOnce(CompilerPass.java:133)
[...]

PR #4392, opened 2016-12-16
compile back to hard scope when closure compile/build requested

enebo:
tbh I am super bothered I cannot figure out a test case to at least occasionally fail.   I explain the basic issue in #4304 (more specifically https://github.com/jruby/jruby/issues/4304#issuecomment-266764789).   Summarized as two threads both decide to build something and one is a child closure and the other is a parent of that closure.  We get a sequence like:

1. child starts
2. parent starts
3. parent accesses childs CFG
4. child changes CFG
5. parent accesses childs CFG
6. BB is missing

This code will ask hard scope to full build itself for blocks and the two entry points to building on IRScope will built the closures before the parent.  It does not appear to infinite loop so yay?  I could not see anything bust with this change and I think the intent is clear but I am unsure in best particular impl to use.  Part of it is based on children wanting parent info for passes vs parents wanting children info for passes.

@subbuss I think you have best knowledge of these expectations and @headius I would like your thoughts as well.

### DIFF
--- a/core/src/main/java/org/jruby/compiler/FullBuildTask.java
+++ b/core/src/main/java/org/jruby/compiler/FullBuildTask.java
@@ -16,6 +16,8 @@ class FullBuildTask implements Runnable {
 
     public void run() {
         try {
+            method.getIRScope().getTopLevelScope().prepareFullBuild();
+
             method.completeBuild(method.getIRScope().prepareFullBuild());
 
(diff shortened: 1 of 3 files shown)


enebo:
Giving it a shot...

enebo:
#4424 also seems to be this issue.  If so we should resolve#4424 as a duplicate of #4304

PR #4395, opened 2016-12-18
Truffle perf warnings

chrisseaton:
* Fold performance warnings in logging
* When something isn't optimised, warn about it only if it gets compiled, otherwise they probably won't worry that it's slow anyway
* With logging levels we don't need an option to disable performance warnings


truffle/src/main/java/org/jruby/truffle/Log.java
```diff
@@ -44,6 +59,59 @@ public void close() throws SecurityException {
...
+     * Warn about code that works but is not yet optimized as Truffle code normally would be. Only prints the warning
+     * if called from compiled code. Don't call this method from behind a boundary or transfer, as it will never print
+     * the warning because it will never be called from compiled code. Use {@link #performance} instead if you need to
+     * warn in code that is never compiled.
+     */
+    public static void notOptimized(String message) {
+        if (CompilerDirectives.inCompiledCode()) {
+            performance(message);
+        }
+    }
```
eregon: Do we need this one? It seems unused so far.

chrisseaton: We're not using it at the moment. I just added it as another option. Do you think I should remove until we actually use it?

eregon: Yes, it's trivial to add if we need it and I'd rather avoid extra methods we don't currently need.

eregon:
How does one enable/disable these warnings?

truffle/src/main/java/org/jruby/truffle/Log.java
chrisseaton: Done.

chrisseaton:
I added a sample logging configuration file. It's a little fussy, I'm afraid, but it's the standard Java way.

https://github.com/jruby/jruby/pull/4395/files#diff-790cdfd51b2ae9a0f85753a2a80d02d3

The great thing about using the standard API is that you can change a few configurations and have us play nicely with any other Java framework you might be using.

truffle/src/main/java/org/jruby/truffle/Log.java
```diff
@@ -9,39 +9,94 @@
...
 
-        });
+    /**
+     * Warn about something that has lower performance than might be expected. Only prints the warning once.
+     */
+    @TruffleBoundary
+    public static void performanceOnce(String message) {
+        // This isn't double-checked locking, because we aren't publishing an object that is then used
```
eregon: The comment can go as well mow :)

chrisseaton: Fixed.

eregon:
Merged!

Issue #4133, opened 2016-09-03
[Truffle] Unable to unzip large gzip file


bjfish:
### Environment

```
require "zlib"

#path = "/Users/brandonfish/Documents/jruby-patches/gzipped_versions"
path = "./gzipped_versions"

content = IO.binread(path)
result = Zlib::GzipReader.new(StringIO.new(content)).read

puts result
puts path
puts ""
```

[gzipped_versions.gz](https://github.com/jruby/jruby/files/453476/gzipped_versions.gz)

Found while trying to do `bundle install`
### Expected Behavior

The script prints a bunch of text.
### Actual Behavior

Fails with Out of Memory Error

chrisseaton:
I can run this example with a gzipped file of 5 MB (so about the same size) in 90 seconds. The file attached to this issue doesn't seem to ever finish and gradually consumes more memory.

<img width="676" alt="screen shot 2016-10-30 at 14 15 25" src="https://cloud.githubusercontent.com/assets/341094/19836780/631e0970-9eab-11e6-8af7-6ac64e9fd536.png">

A heap dump shows it's all ASCII leaf ropes.

<img width="803" alt="screen shot 2016-10-30 at 14 18 21" src="https://cloud.githubusercontent.com/assets/341094/19836793/bf136932-9eab-11e6-9f97-6b746e9b10da.png">

A majority of the ropes seem to be 8,216 or 25 bytes long. These these some buffer sizes somewhere?

Presumably we're creating pathologically deep ropes here and need a new flattening heuristic. I don't understand why a real gzipped file behaves differently to a random one, but the random one is very slow anyway.

bjfish:
@chrisseaton My current understanding is that append to rope is worst case because every append will add two nodes which consumes a lot of memory. @nirvdrum said he has some ideas on how to fix this.

The shape looks something like this:
```
str = "a"
10.times {
  str << "a"
}
Truffle::Ropes.debug_print_rope(str)
```
```
Legend: 
BN = Bytes Null? (byte[] not yet populated)
BL = Byte Length
CL = Character Length
CR = Code Range
O = Offset (SubstringRope only)
T = Times (RepeatingRope only)
D = Depth
LD = Left Depth (ConcatRope only)
RD = Right Depth (ConcatRope only)
aaaaaaaaaaa (ConcatRope; BN: true; BL: 11; CL: 11; CR: CR_7BIT; D: 11; LD: 10; RD: 1)
[...]

PR #4397, opened 2016-12-18
[Truffle] Enhanced byte array.

nirvdrum:
Full disclosure: I don't have confidence I've accounted for every change in przlib here. However, with the change we still pass the zlib specs we passed prior to this change and I've managed to successfully install a gem with this code in place. So, while there may very well be something broken, we are able to run at least one complicated use case without incident.

The basic idea here is przlib really expects String to function well as a byte buffer. E.g., it'll allocate a 64 KB string and then execute `String#setbyte` on each position. Currently, with ropes, this results in the construction of 64K `ConcatRope` instances and degrades to linked list performance for any `String#getbyte` calls.

There are likely improvements we can make to ropes to better support this use case, but we need zlib working now. I anticipate all of this code being replaced with something native, so I took some liberties in modifying files in place. If we need to, we can also revert the commit or merge later on.

I've also added specs for `Rubinius::ByteArray`. They're not comprehensive and nor is the API -- it's for internal use so I don't think we need to support all the various call forms that `String` does. E.g., I rewrote any form of `String#[Range]` to `Rubinius::ByteArray[index, length]`.

### DIFF
--- a/spec/truffle/specs/truffle/byte_array/index_spec.rb
+++ b/spec/truffle/specs/truffle/byte_array/index_spec.rb
@@ -1,0 +1,40 @@
+# Copyright (c) 2016 Oracle and/or its affiliates. All rights reserved. This
+# code is released under a tri EPL/GPL/LGPL license. You can use it,
+# redistribute it and/or modify it under the terms of the:
+#
+# Eclipse Public License version 1.0
+# GNU General Public License version 2
+# GNU Lesser General Public License version 2.1
+# OTHER DEALINGS IN THE SOFTWARE.
+
+require_relative '../../../../ruby/spec_helper'
+
+describe 'Rubinius::ByteArray#index' do
+
+  before :each do
+    @byte_array = Rubinius::ByteArray.new(16, 0xff)
+  end
+
+  describe 'with value not in string' do
+
+    it 'should return nil' do
+      @byte_array.index(0xbb).should be_nil
+    end
+
+  end
+
+  describe 'with index and length' do
+
+    it 'should return the index corresponding to the first occurrence of the value' do
+
+      @byte_array[5] = 0xcc
+      @byte_array.index(0xcc).should == 5
+
+      @byte_array[2] = 0xcc
+      @byte_array.index(0xcc).should == 2
+
+    end
+
+  end
+
+end
--- a/truffle/src/main/java/org/jruby/truffle/core/rubinius/ByteArrayNodes.java
+++ b/truffle/src/main/java/org/jruby/truffle/core/rubinius/ByteArrayNodes.java
@@ -16,7 +16,10 @@ import com.oracle.truffle.api.object.DynamicObject;
 import com.oracle.truffle.api.object.DynamicObjectFactory;
 import com.oracle.truffle.api.profiles.BranchProfile;
 import com.oracle.truffle.api.profiles.ConditionProfile;
+import com.oracle.truffle.api.source.SourceSection;
+import org.jcodings.specific.ASCIIEncoding;
 import org.jruby.truffle.Layouts;
+import org.jruby.truffle.RubyContext;
 import org.jruby.truffle.builtins.CoreClass;
 import org.jruby.truffle.builtins.CoreMethod;
 import org.jruby.truffle.builtins.CoreMethodArrayArgumentsNode;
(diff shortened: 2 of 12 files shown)


bjfish:
I'm curious if it fixes the case mentioned here: https://github.com/jruby/jruby/issues/4133

truffle/src/main/ruby/core/byte_array.rb
```diff
@@ -0,0 +1,33 @@
...
+# redistribute it and/or modify it under the terms of the:
+#
+# Eclipse Public License version 1.0
+# GNU General Public License version 2
+# GNU Lesser General Public License version 2.1
+# OTHER DEALINGS IN THE SOFTWARE.
+
+module Rubinius
+class ByteArray
+
```
chrisseaton: Do we not usually indent twice?

spec/truffle/specs/truffle/byte_array/index_spec.rb
```diff
@@ -0,0 +1,40 @@
...
+    it 'should return nil' do
+      @byte_array.index(0xbb).should be_nil
+    end
+
+  end
+
+  describe 'with index and length' do
+
+    it 'should return the index corresponding to the first occurrence of the value' do
+
```
eregon: Traditionally in specs, there is no extra vertical space before/after `it`, `describe` or around the example code.
IMHO it hurts readability.

truffle/src/main/java/org/jruby/truffle/core/rubinius/ByteArrayNodes.java
```diff
@@ -116,13 +217,46 @@ public Object getByte(DynamicObject bytes, DynamicObject pattern, int start, int
...
+
+        public AllocateNode(RubyContext context, SourceSection sourceSection) {
+            super(context, sourceSection);
+            allocateObjectNode = AllocateObjectNode.create();
+        }
+
         @Specialization
         public DynamicObject allocate(DynamicObject rubyClass) {
-            throw new RaiseException(coreExceptions().typeErrorAllocatorUndefinedFor(rubyClass, this));
+            return allocateObjectNode.allocate(rubyClass, ByteList.EMPTY_BYTELIST);
```
eregon: This sounds dangerous, the `EMPTY_BYTELIST` might not remain empty very long.

truffle/src/main/ruby/core/byte_array.rb
nirvdrum: None of the nested Rubinius module code is indented twice. I just followed that style. E.g., see the various `Rubinius::FFI` classes.

eregon:
Wild idea: could we just use an Array instead of ByteArray here? It would take more space in memory, but require no new code except changes in pr-zlib.

chrisseaton:
Hmmm yeah. And add a new `byte` specialisation to `Array`?

eregon:
I'd rather not deal with `byte` in Ruby, so I was thinking `int[]`.

nirvdrum:
I could look at that again. I started going down that path, but encountered quite a few places things needed  to change and we already had a specialized version of array for bytes, so I enhanced that one. I was trying to avoid changing method calls in przlib too much as they expect a string-like API.

eregon:
Seems like an OK workaround, but we probably want quickly a native zlib, at least for the most common methods.
Ropes will need to handle some fallback when getting too deep, but I guess that can wait a bit longer.

nirvdrum:
I needed to do this because we can't install gems with zlib as it stands. Maybe increasing the heap will do the trick, but as I mentioned in the PR description it creates very deep ropes and is incredibly slow (assuming you don't run out of memory first).

The mutable rope is extremely limited. I started down that path, but it was quite invasive. This is meant to be a patch over until we have a better story for mutable `byte[]` backed strings.

truffle/src/main/ruby/core/byte_array.rb
eregon: OTOH, `string_mirror.rb` and `process_mirror.rb` indent. I think it's better to indent, as that's the most common style for Ruby code.

truffle/src/main/java/org/jruby/truffle/core/rubinius/ByteArrayNodes.java
eregon: You could define `ByteArray.new` so allocation and initialization would not need to be split. `@Nullable` seems fine too.

nirvdrum:
Improvements made to ropes have drastically reduced the performance gap that this PR was intended to address. While this branch is still faster (14s vs 17s for zlib specs), it's probably not a strong enough improvement to justify its cost.

PR #4399, opened 2016-12-19
RubyHash's compare_by_identify Java method naming is confusing

kares:
`Hash#compare_by_identity` is more of a writer (does call a setter)

... looks esp. weird since there's a java get/setCompareByIdentity


kares:
hopefully this is internal - and no one actually cares to have it the old way. but just in case, here's a PR ...

eregon:
Indeed, this confused me quite a bit while looking at `git log -p` :)

headius:
:+1:

PR #4401, opened 2016-12-19
[Truffle] Lazy translation

chrisseaton:
Add support for lazy translation, as a first step towards lazy parsing and maybe serialisation.

It's quite effective on its own though. For `-e 14` minimum heap goes from 35 to 27 MB (saves 8 MB), total allocation from 177.85 MB ± 2.75 MB to 125.78 MB ± 2.2 MB (saves 52 MB), time from 3.47 to 3.13 s (saves 0.34 s).

We can't turn it on however, as it interferes with Truffle instrumentation. I think that's a Truffle bug and I'll be using this new lazy translation as a test case to show that. It's off by default for that reason. There might be a couple of other bugs that are ours which I'll look at after the Truffle issues is fixed.

Contains some work by Brian Belleville.

### DIFF
--- a/truffle/src/main/java/org/jruby/truffle/parser/LazyTranslationNode.java
+++ b/truffle/src/main/java/org/jruby/truffle/parser/LazyTranslationNode.java
@@ -32,9 +32,11 @@ public class LazyTranslationNode extends RubyNode {
     }
 
     public RubyNode resolve() {
-        final RubyNode resolved = resolver.get();
-        replace(resolved, "lazy translation node resolved");
-        return resolved;
+        return atomic(() -> {
+            final RubyNode resolved = resolver.get();
+            replace(resolved, "lazy translation node resolved");
+            return resolved;
+        });
     }
 
 }
--- a/truffle/src/main/java/org/jruby/truffle/parser/LazyTranslationNode.java
+++ b/truffle/src/main/java/org/jruby/truffle/parser/LazyTranslationNode.java
@@ -19,26 +19,15 @@ import java.util.function.Supplier;
 
 public class LazyTranslationNode extends RubyNode {
 
-    private static final AtomicLong createdCount = new AtomicLong();
-    private static final AtomicLong resolvedCount = new AtomicLong();
-
-    static {
-        Runtime.getRuntime().addShutdownHook(new Thread(() -> {
-            Log.info("lazy translation %d/%d", resolvedCount.get(), createdCount.get());
-        }));
-    }
-
     private final Supplier<RubyNode> resolver;
 
     public LazyTranslationNode(Supplier<RubyNode> resolver) {
         this.resolver = resolver;
-        createdCount.getAndIncrement();
     }
 
     @Override
     public Object execute(VirtualFrame frame) {
         CompilerDirectives.transferToInterpreterAndInvalidate();
-        resolvedCount.getAndIncrement();
         final RubyNode resolved = resolver.get();
         replace(resolved, "lazy translation node resolved");
         return resolved.execute(frame);
--- a/truffle/src/main/java/org/jruby/truffle/parser/LazyTranslationNode.java
+++ b/truffle/src/main/java/org/jruby/truffle/parser/LazyTranslationNode.java
@@ -15,6 +15,7 @@ import org.jruby.truffle.Log;
 import org.jruby.truffle.language.RubyNode;
 
 import java.util.concurrent.atomic.AtomicLong;
+import java.util.function.Supplier;
 
 public class LazyTranslationNode extends RubyNode {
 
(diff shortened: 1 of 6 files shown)


chrisseaton:
Harder to measure Rails because it doesn't have an easy to observe start and end, but it seems to save around 17% for time to be up and serving.

truffle/src/main/java/org/jruby/truffle/parser/MethodTranslator.java
```diff
@@ -290,14 +293,31 @@ public RubyNode doCompileMethodBody(RubySourceSection sourceSection, String meth
     }
 
     public MethodDefinitionNode compileMethodNode(RubySourceSection sourceSection, String methodName, MethodDefParseNode defNode, ParseNode bodyNode, SharedMethodInfo sharedMethodInfo) {
-        final RubyNode body = compileMethodBody(sourceSection, methodName, bodyNode, sharedMethodInfo);
+        final TranslatorState state = getCurrentState();
```
eregon: The state only needs to be captured if we are doing lazy translation I think

eregon:
Nice one!

truffle/src/main/java/org/jruby/truffle/parser/LazyTranslationNode.java
```diff
@@ -0,0 +1,40 @@
...
+
+    @Override
+    public Object execute(VirtualFrame frame) {
+        CompilerDirectives.transferToInterpreterAndInvalidate();
+        return resolve().execute(frame);
+    }
+
+    public RubyNode resolve() {
+        final RubyNode resolved = resolver.get();
+        replace(resolved, "lazy translation node resolved");
```
nirvdrum: Do we need to worry about two threads resolving the same method at the same time? Or maybe method aliases?

truffle/src/main/java/org/jruby/truffle/parser/MethodTranslator.java
```diff
@@ -449,4 +469,33 @@ private static String indentation(String line) {
...
 
+    /*
+     * The following methods allow us to save and restore enough of
+     * the current state of the Translator to allow lazy parsing. When
+     * the lazy parsing is actually performed, the state is restored
+     * to what it would have been if the method had been parsed
+     * eagerly.
+     */
+    public TranslatorState getCurrentState() {
+        return new TranslatorState(getEnvironment().unsafeGetLexicalScope(), getEnvironment().isDynamicConstantLookup(), new ArrayDeque<>(parentSourceSection));
```
nirvdrum: I don't recall the exact issue, but I had to do it in BodyTranslator a while ago (see f9df82cff1b48fb56545637de3bd1e605d19c11a). I think there was a problem with nested visit calls that each modified the `parentSourceSection`, which is just a field on the translator. Perhaps more locally scoped variables could work, but the stack approach made sense to me.

nirvdrum:
This is surprisingly elegant.

truffle/src/main/java/org/jruby/truffle/parser/MethodTranslator.java
chrisseaton: Yeah I'm just leaving the stack alone at the moment as I don't totally understand it. I'll look at it again later.

truffle/src/main/java/org/jruby/truffle/parser/LazyTranslationNode.java
chrisseaton: Fixed.

truffle/src/main/java/org/jruby/truffle/parser/LazyTranslationNode.java
```diff
@@ -0,0 +1,42 @@
...
+        this.resolver = resolver;
+    }
+
+    @Override
+    public Object execute(VirtualFrame frame) {
+        CompilerDirectives.transferToInterpreterAndInvalidate();
+        return resolve().execute(frame);
+    }
+
+    public RubyNode resolve() {
```
pitr-ch: could be private?

pitr-ch:
Nice!

truffle/src/main/java/org/jruby/truffle/parser/LazyTranslationNode.java
chrisseaton: I'm going to argue that Truffle should call it when it wants to instrument the node.

truffle/src/main/java/org/jruby/truffle/parser/MethodTranslator.java
eregon: Removed in c8e4a45e54675b8fc72a9bbab07cccc21e0fde47.

PR #4406, opened 2016-12-21
[Truffle] Refactor source section abstractions

chrisseaton:
Sorry, this is all refactoring so it won't be possible to review as normal. But hopefully this refactoring will make the review for the actual precise parser possible to read.

Removes JRuby's `SourcePosition` and associated interfaces, and just use `TempSourceSection`, which is a start and end line, which is what we actually store in a `RubyBaseNode`. Switch from 0 indexed lines to 1 indexed, as Truffle does.

Inline source position information so that we have less objects. `TempSourceSection` is just a throwaway tuple to pass them around.

The next step is actually making source information character index precise and then making that work with the Truffle API, which may be tricky as I also have to solve the `byte[]` / `String` problems.


chrisseaton:
Oh and another big change - don't carry around the source file. We can always figure that out anyway.

truffle/src/main/java/org/jruby/truffle/parser/RubyWarnings.java
```diff
@@ -188,7 +188,7 @@ public void warning(String message) {
     public void warning(ID id, String fileName, int lineNumber, String message) {
         if (!runtime.warningsEnabled() || !runtime.isVerbose()) return;
 
-        warn(id, fileName, lineNumber, message);
+        warn(id, fileName, lineNumber - 1, message);
```
eregon: i.e. does warn does `+ 1` or so?

eregon:
You might be able to avoid TempSourceSection completely, by writing the 2 fields on the translated nodes from the ParseNode directly, like
```java
return t(node, new NilLiteralNode());
void t(parserNode, truffleNode) {
    truffleNode.startLine = parserNode.startLine;
    truffleNode.endLine = parserNode.endLine;
}
```
That's similar to what I do in Mozart-Graal: https://github.com/eregon/mozart-graal/blob/master/vm/src/org/mozartoz/truffle/translator/Translator.java#L285
Nice to have, but can be done later as it requires changing some constructors taking SourceSection or TempSourceSection.

chrisseaton:
Yes I think eventually I'll have `rubyNode.transferSourceSection(parseNode)`.

truffle/src/main/java/org/jruby/truffle/parser/RubyWarnings.java
chrisseaton: Factoring bug. Fixed.

chrisseaton:
Nothing seems to fail either way so we obviously aren't testing these warnings.

eregon:
There are multiple bugs currently with warnings.
I fixed the most obvious one which prevented any warning to show: 0e27962287fbb21f86e4abd6d6ab6a6d5c092840.
Line numbers were currently wrong, so I think the last commit is right.
To test, you can try something like
`jt ruby -w test.rb`
```ruby
o = Object.new
[1,2].each { |o|  }
```

chrisseaton:
The rename was just on a whim - it's shouldn't be around too long (will be refactored away) so I gave it a name to keep that clear in my mind as I use it. The rename isn't the important bit. Removing all the abstraction was.

PR #4411, opened 2016-12-26
[Truffle] Implementing String#scrub

bjfish:


eregon:
Was the Ruby implementation incorrect? It's 50 versus 300 lines, which seems a lot for this.
Maybe some of it can be simplified using Ruby code or using higher-level methods.

truffle/src/main/java/org/jruby/truffle/core/string/StringNodes.java
```diff
@@ -1449,6 +1459,321 @@ private int prevCharHead(Encoding enc, byte[]bytes, int p, int s, int end) {
...
 
+    @CoreMethod(names = "scrub_internal", optional = 1, needsBlock = true)
+    public abstract static class ScrubNode extends YieldingCoreMethodNode {
+
+        private static final byte[] SCRUB_REPL_UTF8 = new byte[]{(byte)0xEF, (byte)0xBF, (byte)0xBD};
+        private static final byte[] SCRUB_REPL_ASCII = new byte[]{(byte)'?'};
+        private static final byte[] SCRUB_REPL_UTF16BE = new byte[]{(byte)0xFF, (byte)0xFD};
+        private static final byte[] SCRUB_REPL_UTF16LE = new byte[]{(byte)0xFD, (byte)0xFF};
+        private static final byte[] SCRUB_REPL_UTF32BE = new byte[]{(byte)0x00, (byte)0x00, (byte)0xFF, (byte)0xFD};
+        private static final byte[] SCRUB_REPL_UTF32LE = new byte[]{(byte)0xFD, (byte)0xFF, (byte)0x00, (byte)0x00};
```
eregon: This looks like it could be simplified by doing something like `uncode_codepoint.chr(encoding)`.

bjfish:
@eregon Yes, the previous ruby implementation wasn't complete. It relied heavily on converters of which many conversions are missing for different encodings.

This also passes most of the mri test for scrub in `ruby/test_m17n.rb`

I tried writing this in ruby a few times and it was never as clean or complete as I would have liked. So, transliterating this java code section was pretty quick on first try and mostly correct it seems. I think if I tried once more to write it in ruby I could make it more complete but it may end up just as verbose. Not sure if it is worth the effort.

eregon:
@bjfish Sounds more than fair then :)
Of course you can use a mix, which you are already doing.

truffle/src/main/java/org/jruby/truffle/core/string/StringNodes.java
```diff
@@ -1449,6 +1459,321 @@ private int prevCharHead(Encoding enc, byte[]bytes, int p, int s, int end) {
         }
     }
 
+    @CoreMethod(names = "scrub_internal", optional = 1, needsBlock = true)
```
eregon: I prefer primitives over extra methods, since they are better hidden to the user.

bjfish: @eregon Do you have an example on how to pass a block to a primitive? I tried what I thought would work but got some internal exception.

eregon: You can pass it as a normal argument, like `prim(arg1, block)` and in the surrounding method capture it with `&block`.

truffle/src/main/java/org/jruby/truffle/core/string/StringNodes.java
```diff
@@ -1449,6 +1459,315 @@ private int prevCharHead(Encoding enc, byte[]bytes, int p, int s, int end) {
...
+            makeConcatNode = RopeNodesFactory.MakeConcatNodeGen.create(null, null, null);
+        }
+
+        @Specialization
+        public DynamicObject scrubNoBlock(VirtualFrame frame, DynamicObject string, Object repl, NotProvided block) {
+            return scrubDefault(frame, string, repl, nil());
+        }
+
+        @Specialization
+        public DynamicObject scrubDefault(VirtualFrame frame, DynamicObject string, Object repl, DynamicObject block) {
```
eregon: This is a huge and complex method to compile. Unfortunately it needs the frame to `yield()`.
Maybe some of it could be extracted in smaller TruffleBoundary functions, or the high-level logic be extracted to Ruby.

bjfish: @eregon  I think I can start by splitting into two specializations by ascii compatibility. Are there specific things here which should have TruffleBoundary?

bjfish:
I've rebased this to remove the source section constructor.

I did a small test to check the code size below `CodeSize        12138`. @eregon about what code size should I try to reduce this method to?
```
[truffle] opt done         block in <main> ../jruby-patches/scrubloop.rb:4 <opt> <split-461c3709>|ASTSize      12/   78 |Time  2013( 846+1166)ms |DirectCallNodes I    3/D    0 |GraalNodes  1232/ 3213 |CodeSize        11901 |Source ../jruby-patches/scrubloop.rb:4 
[truffle] opt done         #scrub /Users/brandonfish/Documents/jruby/truffle/src/main/ruby/core/string.rb:1202 <opt>|ASTSize      49/   66 |Time   461( 244+216 )ms |DirectCallNodes I    2/D    0 |GraalNodes  1177/ 3140 |CodeSize        11873 |Source /Users/brandonfish/Documents/jruby/truffle/src/main/ruby/core/string.rb:1202 
[truffle] opt done         String#scrub_internal (builtin) <opt> <split-7e3d7dd>       |ASTSize      13/   13 |Time   314( 124+190 )ms |DirectCallNodes I    0/D    0 |GraalNodes  1240/ 3093 |CodeSize        12138 |Source       (core):1 
[truffle] opt done         Kernel#nil? (builtin) <opt> <split-413bef78>                |ASTSize       4/    4 |Time    18(  15+3   )ms |DirectCallNodes I    0/D    0 |GraalNodes    23/    3 |CodeSize           69 |Source       (core):1 
```

truffle/src/main/java/org/jruby/truffle/core/encoding/EncodingNodes.java
```diff
@@ -483,6 +484,16 @@ public DynamicObject getEncoding(int index) {
 
     }
 
+    @Primitive(name = "encoding_minlength", needsSelf = false)
+    public static abstract class EncodingMinLengthPrimitiveNode extends PrimitiveArrayArgumentsNode {
+
+        @Specialization
+        public int minLength(DynamicObject encoding) {
+            return EncodingUtils.encMbminlen(Layouts.ENCODING.getEncoding(encoding));
```
nirvdrum: I'm not sure the utility method adds much here. I guess it looks more like MRI's code, but it seems to be a bit of a useless wrapper.

truffle/src/main/java/org/jruby/truffle/core/string/StringNodes.java
```diff
@@ -1355,6 +1365,307 @@ private int prevCharHead(Encoding enc, byte[]bytes, int p, int s, int end) {
...
+            Encoding enc;
+            Encoding encidx;
+            Rope buf = null;
+            byte[] repBytes;
+            int rep;
+            int replen;
+            boolean tainted = false;
+
+            if (cr == CR_7BIT || cr == CR_VALID)
+                return nil();
```
nirvdrum: Let's use curly braces. I know code we port from JRuby doesn't always have them, but if this is new code we should make sure they're there.

truffle/src/main/java/org/jruby/truffle/core/string/StringNodes.java
```diff
@@ -1355,6 +1365,307 @@ private int prevCharHead(Encoding enc, byte[]bytes, int p, int s, int end) {
...
+                    /*
+                     * p1~p: valid ascii/multibyte chars
+                     * p ~e: invalid bytes + unknown bytes
+                     */
+                        int clen = enc.maxLength();
+                        if(buf == null){
+                            buf = RopeConstants.EMPTY_ASCII_8BIT_ROPE;
+                        }
+                        if (p > p1) {
+                               buf = makeConcatNode.executeMake(buf, RopeOperations.ropeFromByteList(new ByteList(pBytes, p1, p - p1)), enc);
```
nirvdrum: Indentation is a bit off here.

nirvdrum: I don't think you need to create a `ByteList` here. You should be able to get by with `ArrayUtils.extractRange` and `RopeOperations.create`.

truffle/src/main/java/org/jruby/truffle/core/string/StringNodes.java
```diff
@@ -1355,6 +1365,307 @@ private int prevCharHead(Encoding enc, byte[]bytes, int p, int s, int end) {
...
+                        cr = CR_VALID;
+                        p += MBCLEN_CHARFOUND_LEN(ret);
+                    }
+                    else if (MBCLEN_INVALID_P(ret)) {
+                    /*
+                     * p1~p: valid ascii/multibyte chars
+                     * p ~e: invalid bytes + unknown bytes
+                     */
+                        int clen = enc.maxLength();
+                        if(buf == null){
```
nirvdrum: Small thing, but please add a space before the `(`.

truffle/src/main/java/org/jruby/truffle/core/string/StringNodes.java
```diff
@@ -1355,6 +1365,307 @@ private int prevCharHead(Encoding enc, byte[]bytes, int p, int s, int end) {
...
+                            clen--;
+                            for (; clen > 1; clen--) {
+                                ret = enc.length(pBytes, q, q + clen);
+                                if (MBCLEN_NEEDMORE_P(ret)) break;
+                                if (MBCLEN_INVALID_P(ret)) continue;
+                            }
+                        }
+                        if (repBytes != null) {
+                            buf = makeConcatNode.executeMake(buf, RopeOperations.ropeFromByteList(new ByteList(repBytes, rep, replen)), enc);
+                            if (!rep7bit_p){
```
nirvdrum: Small thing, but please add a space before the `{`.

truffle/src/main/java/org/jruby/truffle/core/string/StringNodes.java
```diff
@@ -1355,6 +1365,307 @@ private int prevCharHead(Encoding enc, byte[]bytes, int p, int s, int end) {
...
+                            int q = p;
+                            clen--;
+                            for (; clen > 1; clen--) {
+                                ret = enc.length(pBytes, q, q + clen);
+                                if (MBCLEN_NEEDMORE_P(ret)) break;
+                                if (MBCLEN_INVALID_P(ret)) continue;
+                            }
+                        }
+                        if (repBytes != null) {
+                            buf = makeConcatNode.executeMake(buf, RopeOperations.ropeFromByteList(new ByteList(repBytes, rep, replen)), enc);
```
nirvdrum: Similar deal here with not needing to create a `ByteList`. There's also a bit of a weird combination of slow & fast methods here. You probably should be using `MakeLeafRopeNode` rather than `RopeOperations` to create the leaf, unless you're on a slow path.

truffle/src/main/java/org/jruby/truffle/core/string/StringNodes.java
```diff
@@ -1355,6 +1365,307 @@ private int prevCharHead(Encoding enc, byte[]bytes, int p, int s, int end) {
...
+                            if (!rep7bit_p){
+                                cr = CR_VALID;
+                            }
+                        }
+                        else {
+                            repl = yield(frame, (DynamicObject)block, createString(new ByteList(pBytes, p, clen, enc, true)));
+                            repl = strCompatAndValidNode.call(frame, string, "str_compat_and_valid", repl, rubyEncoding);
+                            tainted |= isTaintedNode.executeIsTainted(repl);
+                            buf = makeConcatNode.executeMake(buf, rope((DynamicObject)repl), enc);
+                            if (rope((DynamicObject) repl).getCodeRange() == CR_VALID){
```
nirvdrum: Small thing, but please add a space before the {.

nirvdrum: `StringGuards.isValidCodeRange` might make this a bit easier to read.

truffle/src/main/java/org/jruby/truffle/core/string/StringNodes.java
```diff
@@ -1355,6 +1365,307 @@ private int prevCharHead(Encoding enc, byte[]bytes, int p, int s, int end) {
...
+                            }
+                        }
+                        if (repBytes != null) {
+                            buf = makeConcatNode.executeMake(buf, RopeOperations.ropeFromByteList(new ByteList(repBytes, rep, replen)), enc);
+                            if (!rep7bit_p){
+                                cr = CR_VALID;
+                            }
+                        }
+                        else {
+                            repl = yield(frame, (DynamicObject)block, createString(new ByteList(pBytes, p, clen, enc, true)));
```
nirvdrum: Do these bytes get modified again? If not, we shouldn't need to make a copy of them. But, I think you can avoid that mess by using `ArrayUtils.extractRange`.

truffle/src/main/java/org/jruby/truffle/core/string/StringNodes.java
```diff
@@ -1355,6 +1365,307 @@ private int prevCharHead(Encoding enc, byte[]bytes, int p, int s, int end) {
...
+                }
+                if (buf == null) {
+                    if (p == e) {
+//                        setCodeRange(cr);
+                        return nil();
+                    }
+                    buf = RopeConstants.EMPTY_ASCII_8BIT_ROPE;
+                }
+                if (p1 < p) {
+                    buf = makeConcatNode.executeMake(buf, RopeOperations.ropeFromByteList(new ByteList(pBytes, p1, p - p1)), enc);
```
nirvdrum: Same idea with not creating a `ByteList`.

truffle/src/main/java/org/jruby/truffle/core/string/StringNodes.java
```diff
@@ -1355,6 +1365,307 @@ private int prevCharHead(Encoding enc, byte[]bytes, int p, int s, int end) {
...
+                }
+                if (p < e) {
+                    if (repBytes != null) {
+                        buf = makeConcatNode.executeMake(buf, RopeOperations.ropeFromByteList(new ByteList(repBytes, rep, replen)), enc);
```
nirvdrum: Same idea with not creating a `ByteList`.

truffle/src/main/java/org/jruby/truffle/core/string/StringNodes.java
```diff
@@ -1355,6 +1365,307 @@ private int prevCharHead(Encoding enc, byte[]bytes, int p, int s, int end) {
...
+                    if (MBCLEN_NEEDMORE_P(ret)) {
+                        break;
+                    }
+                    else if (MBCLEN_CHARFOUND_P(ret)) {
+                        p += MBCLEN_CHARFOUND_LEN(ret);
+                    }
+                    else if (MBCLEN_INVALID_P(ret)) {
+                        int q = p;
+                        int clen = enc.maxLength();
+                        if(buf == null){
```
nirvdrum: Missing spaces.

truffle/src/main/java/org/jruby/truffle/core/string/StringNodes.java
```diff
@@ -1355,6 +1365,307 @@ private int prevCharHead(Encoding enc, byte[]bytes, int p, int s, int end) {
...
+                            buf = RopeConstants.EMPTY_ASCII_8BIT_ROPE;
+                        };
+                        if(p > p1){
```
nirvdrum: Missing spaces and maybe insert a newline above the `if`.

truffle/src/main/java/org/jruby/truffle/core/string/StringNodes.java
```diff
@@ -1355,6 +1365,307 @@ private int prevCharHead(Encoding enc, byte[]bytes, int p, int s, int end) {
...
+                            buf = makeConcatNode.executeMake(buf, RopeOperations.ropeFromByteList(new ByteList(pBytes, p1, p - p1)), enc);
+                        }
+
+
```
nirvdrum: Why so much space here?

truffle/src/main/java/org/jruby/truffle/core/string/StringNodes.java
```diff
@@ -1355,6 +1365,307 @@ private int prevCharHead(Encoding enc, byte[]bytes, int p, int s, int end) {
...
+                        p += MBCLEN_CHARFOUND_LEN(ret);
+                    }
+                    else if (MBCLEN_INVALID_P(ret)) {
+                        int q = p;
+                        int clen = enc.maxLength();
+                        if(buf == null){
+                            buf = RopeConstants.EMPTY_ASCII_8BIT_ROPE;
+                        };
+                        if(p > p1){
+                            buf = makeConcatNode.executeMake(buf, RopeOperations.ropeFromByteList(new ByteList(pBytes, p1, p - p1)), enc);
```
nirvdrum: Same idea with not creating a `ByteList`.

truffle/src/main/java/org/jruby/truffle/core/string/StringNodes.java
```diff
@@ -1355,6 +1365,307 @@ private int prevCharHead(Encoding enc, byte[]bytes, int p, int s, int end) {
...
+                        else {
+                            clen -= mbminlen;
+                            for (; clen > mbminlen; clen-=mbminlen) {
+                                ret = enc.length(pBytes, q, q + clen);
+                                if (MBCLEN_NEEDMORE_P(ret)) break;
+                                if (MBCLEN_INVALID_P(ret)) continue;
+                            }
+                        }
+                        if (repBytes != null) {
+                            buf = makeConcatNode.executeMake(buf, RopeOperations.ropeFromByteList(new ByteList(repBytes, rep, replen)), enc);
```
nirvdrum: Same idea with not creating a `ByteList`.

truffle/src/main/java/org/jruby/truffle/core/string/StringNodes.java
```diff
@@ -1355,6 +1365,307 @@ private int prevCharHead(Encoding enc, byte[]bytes, int p, int s, int end) {
...
+                        }
+                        else {
+                            repl = yield(frame, (DynamicObject)block, createString(new ByteList(pBytes, p, e - p, enc, true)));
```
nirvdrum: Same idea with not creating a `ByteList`.

truffle/src/main/java/org/jruby/truffle/core/string/StringNodes.java
```diff
@@ -1355,6 +1365,307 @@ private int prevCharHead(Encoding enc, byte[]bytes, int p, int s, int end) {
...
+                    }
+                }
+                cr = CR_VALID;
+            }
+
+            final boolean isTaint = isTaintedNode.executeIsTainted(string);
+
+            final DynamicObject resultString = createString(buf);
+
+            if(tainted | isTaint){
```
nirvdrum: Missing spaces.

truffle/src/main/java/org/jruby/truffle/core/string/StringNodes.java
```diff
@@ -1355,6 +1365,307 @@ private int prevCharHead(Encoding enc, byte[]bytes, int p, int s, int end) {
...
+                }
+                if (buf == null) {
+                    if (p == e) {
+//                        setCodeRange(CR_VALID);
+                        return nil();
+                    }
+                    buf = RopeConstants.EMPTY_ASCII_8BIT_ROPE;
+                }
+                if (p1 < p) {
+                    buf = makeConcatNode.executeMake(buf, RopeOperations.ropeFromByteList(new ByteList(pBytes, p1, p - p1)), enc);
```
nirvdrum: Same idea with not creating a `ByteList`.

truffle/src/main/java/org/jruby/truffle/core/string/StringNodes.java
```diff
@@ -1355,6 +1365,307 @@ private int prevCharHead(Encoding enc, byte[]bytes, int p, int s, int end) {
...
+                }
+                if (p < e) {
+                    if (repBytes != null) {
+                        buf = makeConcatNode.executeMake(buf, RopeOperations.ropeFromByteList(new ByteList(repBytes, rep, replen)), enc);
```
nirvdrum: Same idea with not creating a `ByteList`.

truffle/src/main/java/org/jruby/truffle/core/string/StringNodes.java
```diff
@@ -1355,6 +1365,307 @@ private int prevCharHead(Encoding enc, byte[]bytes, int p, int s, int end) {
...
+                    }
+                    else {
+                        repl = yield(frame, (DynamicObject)block, createString(new ByteList(pBytes, p, e - p, enc, true)));
```
nirvdrum: Same idea with not creating a `ByteList`.

truffle/src/main/java/org/jruby/truffle/core/string/StringNodes.java
```diff
@@ -1355,6 +1365,307 @@ private int prevCharHead(Encoding enc, byte[]bytes, int p, int s, int end) {
...
+                        p = StringSupport.searchNonAscii(pBytes, p, e);
+                        if (p == -1) {
+                            p = e;
+                            break;
+                        }
+                    }
+                }
+                if (buf == null) {
+                    if (p == e) {
+//                        setCodeRange(cr);
```
nirvdrum: I'll need to look through this more, but I'd feel better if we have an assert that the code range is the same. Or maybe throw an exception if they're different. If they are different, there's a behavioral difference here and we may need to create a new rope with the correct code range and update the string's rope reference.

truffle/src/main/java/org/jruby/truffle/core/string/StringNodes.java
```diff
@@ -1355,6 +1365,307 @@ private int prevCharHead(Encoding enc, byte[]bytes, int p, int s, int end) {
...
+                            tainted |= isTaintedNode.executeIsTainted(repl);
+                          buf = makeConcatNode.executeMake(buf, rope((DynamicObject)repl), enc);
+                        }
+                        p += clen;
+                        p1 = p;
+                    }
+                }
+                if (buf == null) {
+                    if (p == e) {
+//                        setCodeRange(CR_VALID);
```
nirvdrum: Same concern as the previous commented out line setting the code range.

truffle/src/main/ruby/core/string.rb
```diff
@@ -1195,72 +1199,33 @@ def match(pattern, pos=0)
...
     return self
   end
 
+  def str_compat_and_valid(str, enc)
+    res = StringValue(str)
+    if !res.valid_encoding?
+      raise ArgumentError, "replacement must be valid byte sequence"
+    else
+      min_length = Truffle.invoke_primitive(:encoding_minlength, enc)
+      if res.ascii_only? ? min_length != 1 : enc != res.encoding
```
nirvdrum: If all you're doing is checking if the encoding is ASCII compatible, you can just use `Encoding#ascii_compatible?` and can get rid of the primitive entirely.

truffle/src/main/java/org/jruby/truffle/core/string/StringNodes.java
```diff
@@ -1355,6 +1365,307 @@ private int prevCharHead(Encoding enc, byte[]bytes, int p, int s, int end) {
...
+
+            final boolean isTaint = isTaintedNode.executeIsTainted(string);
+
+            final DynamicObject resultString = createString(buf);
+
+            if(tainted | isTaint){
+                taintNode.executeTaint(resultString);
+            }
+
+//            ((RubyString)buf).setEncodingAndCodeRange(enc, cr);
```
nirvdrum: Similar concern here that the encoding and code range might not be the same.

truffle/src/main/java/org/jruby/truffle/core/string/StringNodes.java
```diff
@@ -1355,6 +1365,307 @@ private int prevCharHead(Encoding enc, byte[]bytes, int p, int s, int end) {
...
+            else {
+	        /* ASCII incompatible */
+                byte[] pBytes = value.unsafeBytes();
+                int p = value.begin();
+                int e = p + value.getRealSize();
+                int p1 = p;
+                int mbminlen = enc.minLength();
+                if (repl != nil()) {
+                    final Rope replRope = rope((DynamicObject) repl);
+                    final ByteList replValue = RopeOperations.getByteListReadOnly(replRope);
```
nirvdrum: I don't think you need to convert to a `ByteList` here. You should be able to just call `Rope#getBytesCopy`. The begin value will always be 0 and the byte array will always be the exact length, so you don't need a "real size" method. If you don't need to modify the array at all, then you can just use `Rope#getBytes`.

truffle/src/main/java/org/jruby/truffle/core/string/StringNodes.java
```diff
@@ -1355,6 +1365,307 @@ private int prevCharHead(Encoding enc, byte[]bytes, int p, int s, int end) {
...
+                boolean rep7bit_p;
+                if (block != nil()) {
+                    repBytes = null;
+                    rep = 0;
+                    replen = 0;
+                    rep7bit_p = false;
+                }
+                else if (repl != nil()) {
+                    final Rope replRope = rope((DynamicObject) repl);
+                    final ByteList replValue = RopeOperations.getByteListReadOnly(replRope);
```
nirvdrum: I don't think you need to convert to a `ByteList` here. You should be able to just call `Rope#getBytesCopy`. The begin value will always be 0 and the byte array will always be the exact length, so you don't need a "real size" method. If you don't need to modify the array at all, then you can just use `Rope#getBytes`.

truffle/src/main/java/org/jruby/truffle/core/encoding/EncodingNodes.java
bjfish: Fixed.

truffle/src/main/java/org/jruby/truffle/core/string/StringNodes.java
bjfish: Fixed.

truffle/src/main/java/org/jruby/truffle/core/string/StringNodes.java
bjfish: This was giving me failures when I tried to switch (bytes copy too). I'll keep investigating. look okay? :
```
                    repBytes = rope.getBytes();
                    rep = 0;
                    replen = repBytes.length;

```

truffle/src/main/java/org/jruby/truffle/core/string/StringNodes.java
bjfish: Fixed.

truffle/src/main/java/org/jruby/truffle/core/string/StringNodes.java
bjfish: Fixed.

truffle/src/main/java/org/jruby/truffle/core/string/StringNodes.java
bjfish: I updated to extract range. I haven't done MakeLeafRopeNode yet since I don't quite understand that yet.

truffle/src/main/java/org/jruby/truffle/core/string/StringNodes.java
bjfish: Fixed.

truffle/src/main/java/org/jruby/truffle/core/string/StringNodes.java
bjfish: Fixed.

truffle/src/main/java/org/jruby/truffle/core/string/StringNodes.java
bjfish: Fixed.

truffle/src/main/java/org/jruby/truffle/core/string/StringNodes.java
bjfish: I had an issue making this update. See previous comment

bjfish:
@nirvdrum I think I Fixed all the style comments you mentioned but let me know if you see any issues. 

Code size down to `CodeSize        10145` with the updates I've made so far

truffle/src/main/java/org/jruby/truffle/core/string/StringNodes.java
```diff
@@ -1355,6 +1365,307 @@ private int prevCharHead(Encoding enc, byte[]bytes, int p, int s, int end) {
...
+            CodeRange cr = rope.getCodeRange();
+            Encoding enc;
+            Encoding encidx;
+            Rope buf = null;
+            byte[] repBytes;
+            int rep;
+            int replen;
+            boolean tainted = false;
+
+            if (cr == CR_7BIT || cr == CR_VALID) {
```
eregon: This should have a ConditionProfile.

bjfish: @eregon Could you give me some more background on `ConditionProfile` ? As I currently understand it, it allows to compile different paths depending on a condition. How would this compare to adding a new specialization for this condition?

I thought the `enc.isAsciiCompatible()` if branch might be a good location for a binary profile. It should reduce the code size in half.

bjfish:
Closing and re-opening to re-build

truffle/src/main/java/org/jruby/truffle/core/string/StringNodes.java
eregon: Adding a specialization for this would have the same effect.
ConditionProfile allows to compile only the branch used.

In the case the CR it 7-bit or valid we only want the fast path to get compiled.
Yes, always use createBinaryProfile(). The counting one is for advanced scenarios.

Indeed, it would be very helpful on `if (enc.isAsciiCompatible())`.

bjfish: @eregon I've added these profiles

truffle/src/main/ruby/core/string.rb
bjfish: @nirvdrum I've updated this to use `Encoding#ascii_compatible?` and removed the encoding primitive

truffle/src/main/java/org/jruby/truffle/core/string/StringNodes.java
bjfish: Fixed.

truffle/src/main/java/org/jruby/truffle/core/string/StringNodes.java
bjfish: Spaces added.

truffle/src/main/ruby/core/string.rb
```diff
@@ -1195,72 +1199,32 @@ def match(pattern, pos=0)
...
+    res = StringValue(str)
+    if !res.valid_encoding?
+      raise ArgumentError, "replacement must be valid byte sequence"
+    else
+      if res.ascii_only? ? !enc.ascii_compatible? : enc != res.encoding
+        raise Encoding::CompatibilityError, "incompatible character encodings"
+      end
+    end
+    res
+  end
```
eregon: This should be private.

truffle/src/main/java/org/jruby/truffle/core/string/StringNodes.java
nirvdrum: I don't see what's wrong here. Just note that if you don't make a bytes copy, you will modify the bytes in an existing rope and that will cause problems if the rope has been shared at all.

truffle/src/main/ruby/core/string.rb
bjfish: Thanks, I've made this private now

truffle/src/main/java/org/jruby/truffle/core/string/StringNodes.java
bjfish: @nirvdrum I tried it again and it appears to work now. I just pushed a new commit

truffle/src/main/java/org/jruby/truffle/core/string/StringNodes.java
```diff
@@ -1355,6 +1365,309 @@ private int prevCharHead(Encoding enc, byte[]bytes, int p, int s, int end) {
...
+                    repBytes = null;
+                    rep = 0;
+                    replen = 0;
+                    rep7bit_p = false;
+                }
+                else if (repl != nil()) {
+                    final Rope replRope = rope((DynamicObject) repl);
+                    repBytes = replRope.getBytesCopy();
+                    rep = 0;
+                    replen = repBytes.length;
```
eregon: There is no longer a need for `rep`, `replen` etc now

bjfish: @eregon I've cleaned up these variables and made some final in the latest commit.

truffle/src/main/java/org/jruby/truffle/core/string/StringNodes.java
```diff
@@ -1355,6 +1365,285 @@ private int prevCharHead(Encoding enc, byte[]bytes, int p, int s, int end) {
...
+            return scrubDefault(frame, string, repl, nil(), validRangeProfile, asciiCompatibleProfile);
+        }
+
+        @Specialization
+        public DynamicObject scrubDefault(VirtualFrame frame, DynamicObject string, Object repl, DynamicObject block,
+                                          @Cached("createBinaryProfile()") ConditionProfile validRangeProfile,
+                                          @Cached("createBinaryProfile()") ConditionProfile asciiCompatibleProfile) {
+            final DynamicObject rubyEncoding = getContext().getEncodingManager().getRubyEncoding(StringOperations.encoding(string));
+            final Rope rope = rope(string);
+            final ByteList value = RopeOperations.getByteListReadOnly(rope);
```
eregon: This should not be needed, just `Rope#getBytes()` should work fine (and then there is no need for `begin()`).

eregon:
I'm starting to see a bit through the logic here, isn't the core of it as simple as:
```ruby
"abc\x82\u3042\x81".chars.map { |c| c.valid_encoding? ? c : 65533.chr(c.encoding) }.join
```
?

truffle/src/main/java/org/jruby/truffle/core/string/StringNodes.java
bjfish: @eregon I've added a new commit to remove the byte list

nirvdrum:
@eregon I'm not sure if `String#chars` really has defined behavior if the string is invalid. Certainly there are specs that rely on this (e.g., `String#length` on an invalid string), but I think they're wrong as well. What does a character even mean in that case?

nirvdrum:
@eregon Also, perhaps this is a case we should handle better, but you'd make a big rope where each node is a single character. In the worst case where there's nothing to scrub, you could take a nice flat leaf and turn it into a linked list.

eregon:
@nirvdrum The code is illustrative, it could of course be tuned to be more efficient :)
In any case scrub is made to walk over a broken String, and it must respect existing valid characters boundaries, isn't it?

nirvdrum:
@eregon The block form of `String#scrub` should yield invalid bytes, yes. But I'm not sure that `String#chars` is guaranteed to match the same output. Maybe it is (where the char would represent the invalid byte array).

eregon:
@bjfish @nirvdrum I moved most of the logic to Ruby code, except the part that cannot be written in Ruby, which remains in the `string_scrub` primitive (for instance the primitive always takes a block, never a replacement String directly).
This is usually a good combination: do the minimum in Java, and have the logic to Ruby where it's much easier to read and does not multiply the complexity.

truffle/src/main/java/org/jruby/truffle/core/string/StringNodes.java
```diff
@@ -23,7 +23,7 @@
  * Copyright (C) 2006 Ola Bini <ola@ologix.com>
  * Copyright (C) 2007 Nick Sieger <nicksieger@gmail.com>
  *
- * Some of the code in this class is transposed from org.jruby.util.ByteList,
+ * Some of the code in this class is transposed from org.jruby.util.ByteList and org.jruby.RubyString,
```
nirvdrum: This should go away. We note that we take from `org.jruby.RubyString` in the copyright notice above this one.

truffle/src/main/ruby/core/string.rb
nirvdrum: I, personally, would prefer if we moved non-standard and non-primitive methods out to `Truffle::<Type>`. We already have `Truffle::String` for other things. But if you two are happy with this, we can leave it.

truffle/src/main/ruby/core/string.rb
```diff
@@ -1195,72 +1199,55 @@ def match(pattern, pos=0)
...
     replace(scrub(replace, &block))
     return self
   end
 
+  def str_compat_and_valid(str, enc)
+    res = StringValue(str)
+    if !res.valid_encoding?
+      raise ArgumentError, "replacement must be valid byte sequence"
+    else
+      if res.ascii_only? ? !enc.ascii_compatible? : enc != res.encoding
```
nirvdrum: Looking at this more, it's not clear why this isn't just `Encoding.compatible?` or `Rubinius::Type.compatible_encoding`. There's a bit of a logical difference here, since `Encoding.compatible?` checks if strings are empty as well. I can't easily tell if the behavioral difference is intended or is a subtle bug.

truffle/src/main/java/org/jruby/truffle/core/string/StringNodes.java
eregon: c19dd704aebacfeafe2a79473e21becc363ba3bc

truffle/src/main/ruby/core/string.rb
eregon: Or the Mirror maybe?
Truffle::String only has 1 Java method it seems currently.

truffle/src/main/java/org/jruby/truffle/core/string/StringNodes.java
```diff
@@ -1355,6 +1360,153 @@ private int prevCharHead(Encoding enc, byte[]bytes, int p, int s, int end) {
...
     }
 
+    @Primitive(name = "string_scrub")
+    @ImportStatic(StringGuards.class)
+    public abstract static class ScrubNode extends PrimitiveArrayArgumentsNode {
+
+        @Child private YieldNode yieldNode = new YieldNode();
+        @Child private RopeNodes.MakeConcatNode makeConcatNode = RopeNodesFactory.MakeConcatNodeGen.create(null, null, null);
+
+        @Specialization(guards = { "!is7Bit(string)", "!isValidCodeRange(string)", "isAsciiCompatible(string)" })
```
nirvdrum: Replace `"!is7Bit(string)", "!isValidCodeRange(string)"` with `isBrokenCodeRange(string)`. We never have ropes with unknown code ranges floating around, so we don't need to worry about that case. When you eliminate the other two, broken code ranges are all that remain. Incidentally, I think it also clarifies the intent more.

truffle/src/main/java/org/jruby/truffle/core/string/StringNodes.java
```diff
@@ -1355,6 +1360,153 @@ private int prevCharHead(Encoding enc, byte[]bytes, int p, int s, int end) {
...
+            if (p < e) {
+                Rope invalid = RopeOperations.create(ArrayUtils.extractRange(pBytes, p, e), enc, CodeRange.CR_BROKEN);
+                DynamicObject repl = (DynamicObject) yield(frame, block, createString(invalid));
+                buf = makeConcatNode.executeMake(buf, rope(repl), enc);
+            }
+
+            return createString(buf);
+        }
+
+        @Specialization(guards = { "!is7Bit(string)", "!isValidCodeRange(string)", "!isAsciiCompatible(string)" })
```
nirvdrum: Replace `"!is7Bit(string)", "!isValidCodeRange(string)"` with `isBrokenCodeRange(string)`. We never have ropes with unknown code ranges floating around, so we don't need to worry about that case. When you eliminate the other two, broken code ranges are all that remain. Incidentally, I think it also clarifies the intent more.

truffle/src/main/java/org/jruby/truffle/core/string/StringNodes.java
```diff
@@ -1355,6 +1360,153 @@ private int prevCharHead(Encoding enc, byte[]bytes, int p, int s, int end) {
...
+                if (MBCLEN_NEEDMORE_P(ret)) {
+                    break;
+                } else if (MBCLEN_CHARFOUND_P(ret)) {
+                    p += MBCLEN_CHARFOUND_LEN(ret);
+                } else if (MBCLEN_INVALID_P(ret)) {
+                    // p1~p: valid ascii/multibyte chars
+                    // p ~e: invalid bytes + unknown bytes
+                    int clen = enc.maxLength();
+                    if (p1 < p) {
+                        buf = makeConcatNode.executeMake(buf, RopeOperations.create(ArrayUtils.extractRange(pBytes, p1, p), enc, CodeRange.CR_VALID), enc);
```
nirvdrum: I'm a bit torn. This code is slow path, but we don't have a boundary. We should probably just use `RopeNodes.MakeLeafRopeNode` node here.

nirvdrum: I think instead of `CR_VALID`, we need to pass `CR_UNKNOWN` here and allow a byte scan to occur. It's possible that the substring of `CR_VALID` string is actually `CR_7BIT`.

It's possible this could be modeled as a `SubstringRope` as well and then we wouldn't need to extract anything.

truffle/src/main/java/org/jruby/truffle/core/string/StringNodes.java
```diff
@@ -1355,6 +1360,153 @@ private int prevCharHead(Encoding enc, byte[]bytes, int p, int s, int end) {
...
+                    p1 = p;
+                    p = StringSupport.searchNonAscii(pBytes, p, e);
+                    if (p == -1) {
+                        p = e;
+                        break;
+                    }
+                }
+            }
+            if (p1 < p) {
+                buf = makeConcatNode.executeMake(buf, RopeOperations.create(ArrayUtils.extractRange(pBytes, p1, p), enc, CodeRange.CR_VALID), enc);
```
nirvdrum: Same comments about making leaves an ensuring the correct code range.

truffle/src/main/java/org/jruby/truffle/core/string/StringNodes.java
```diff
@@ -1355,6 +1360,153 @@ private int prevCharHead(Encoding enc, byte[]bytes, int p, int s, int end) {
...
+                if (MBCLEN_NEEDMORE_P(ret)) {
+                    break;
+                } else if (MBCLEN_CHARFOUND_P(ret)) {
+                    p += MBCLEN_CHARFOUND_LEN(ret);
+                } else if (MBCLEN_INVALID_P(ret)) {
+                    final int q = p;
+                    int clen = enc.maxLength();
+
+                    if (p1 < p) {
+                        buf = makeConcatNode.executeMake(buf, RopeOperations.create(ArrayUtils.extractRange(pBytes, p1, p), enc, CodeRange.CR_VALID), enc);
```
nirvdrum: Same comments about making leaves an ensuring the correct code range.

truffle/src/main/java/org/jruby/truffle/core/string/StringNodes.java
```diff
@@ -1355,6 +1360,153 @@ private int prevCharHead(Encoding enc, byte[]bytes, int p, int s, int end) {
...
+
+                    Rope invalid = RopeOperations.create(ArrayUtils.extractRange(pBytes, p, e), enc, CodeRange.CR_BROKEN);
+                    DynamicObject repl = (DynamicObject) yield(frame, block, createString(invalid));
+                    buf = makeConcatNode.executeMake(buf, rope(repl), enc);
+                    p += clen;
+                    p1 = p;
+                }
+            }
+            if (p1 < p) {
+                buf = makeConcatNode.executeMake(buf, RopeOperations.create(ArrayUtils.extractRange(pBytes, p1, p), enc, CodeRange.CR_VALID), enc);
```
nirvdrum: Same comments about making leaves an ensuring the correct code range.

truffle/src/main/ruby/core/string.rb
nirvdrum: Either way. I think `Truffle::String` came about when we were trying to not modify the Rubinius code, since that made merges annoying. I'd just like to get string.rb down to core methods if we can.

truffle/src/main/ruby/core/string.rb
eregon: It seems to behave quite differently then traditional compatible checks.

truffle/src/main/ruby/core/string.rb
```diff
@@ -1208,16 +1208,26 @@ def scrub(replace = nil, &block)
     end
 
     taint = tainted?
+
+    validate = -> str {
```
nirvdrum: You may have found the one case where we use `{}` for multi-line blocks.

truffle/src/main/ruby/core/string.rb
```diff
@@ -1208,16 +1208,26 @@ def scrub(replace = nil, &block)
...
+      str = StringValue(str)
+      unless str.valid_encoding?
+        raise ArgumentError, "replacement must be valid byte sequence"
+      end
+      # Special encoding check for #scrub
+      if str.ascii_only? ? !encoding.ascii_compatible? : encoding != str.encoding
+        raise Encoding::CompatibilityError, "incompatible character encodings"
+      end
+      taint = true if str.tainted?
```
nirvdrum: I don't really care for modifying the captured value like this. It's too easy to read as setting a local. Maybe add a comment?

truffle/src/main/ruby/core/string.rb
nirvdrum: +1 for getting rid of the method.

truffle/src/main/java/org/jruby/truffle/core/string/StringNodes.java
```diff
@@ -1366,6 +1366,7 @@ private int prevCharHead(Encoding enc, byte[]bytes, int p, int s, int end) {
 
         @Child private YieldNode yieldNode = new YieldNode();
         @Child private RopeNodes.MakeConcatNode makeConcatNode = RopeNodesFactory.MakeConcatNodeGen.create(null, null, null);
+        @Child private RopeNodes.MakeSubstringNode makeSubstringNode = RopeNodesFactory.MakeSubstringNodeGen.create(null, null, null);
```
nirvdrum: You can just use `RopeNodes.MakeSubstringNode.create()`.

spec/ruby/core/string/scrub_spec.rb
```diff
@@ -39,7 +39,7 @@
     "abc\u3042#{x81}".scrub("*").should == "abc\u3042*"
   end
 
-  it "replaces groups of sequences together with a single replacement" do
+  it "replaces and incomplete character at the end with a single replacement" do
```
nirvdrum: "and" -> "an"

eregon:
Rebased and merged to truffle-head.

Issue #4403, opened 2016-12-20
MRI and JRuby disagree about File.exist?("/non_directory_file.name/")


the-michael-toy:
```
% cat test.rb
puts "This file is #{__FILE__}, add a slash to it in #{RUBY_DESCRIPTION}"
puts File.exist?(__FILE__+"/") ? "is INCORRECTLY found" : "is correctly not found"
```

Here's MRI

```
% rvm use 2.3.1
Using /Users/mtoy/.rvm/gems/ruby-2.3.1
% ruby test.rb 
This file is test.rb, add a slash to it in ruby 2.3.1p112 (2016-04-26 revision 54768) [x86_64-darwin16]
is correctly not found
```

And both 9.1.5.0 and 1.7.19 ( the two versions we use around here) ...
```
% rvm use jruby-9.1.5.0
Using /Users/mtoy/.rvm/gems/jruby-9.1.5.0
% ruby test.rb 
This file is test.rb, add a slash to it in jruby 9.1.5.0 (2.3.1) 2016-09-07 036ce39 Java HotSpot(TM) 64-Bit Server VM 25.112-b16 on 1.8.0_112-b16 +jit [darwin-x86_64]
is INCORRECTLY found
```

enebo:
It appears jnr-posix JavaSecureFile will strip off the extra slash erroneously here.

enebo:
Actually two things to point out in this issue:
1. java.io.File actually exhibits the bad behavior too:

```text
jruby -e 'f = java.io.File.new("../snippets/cp2.rb/"); p f.exists?'
true
```

cp2.rb is a regular file.  It is possible we can just secondarily always check isDirectory() here and then throw ENOTDIR.

2. The bad behavior in 1 actually strips out the / when it normalizes the path.  So even if we use the path from this file to call a native stat() we have already stripped the trailing / so we cannot know to ENOTDIR.

The base reported problem is this should not stat but it is coupled with proper return value as well.  is cp2.rb did not exist we should return ENOENT but if it does exist we should return ENOTDIR.

enebo:
Since we have had this problem perhaps for the entire length of JRuby I am not prioritizing this for the current release.

sumitmah:
@enebo shouldn't cruby resolve the path and return true?

enebo:
@sumitmah you can raise it as an issue no cruby's site as a bug and see if they agree or not.  I don't actually understand why they return false.  It could be as simple as they split on / and get '' as name?

[...]

PR #4422, opened 2017-01-06
[#4403] To be consistent with MRI, File.exists? for <path to file> end…

sumitmah:
…ing with / should return false


kares:
Thanks Sumit, there [are a few regressions](https://travis-ci.org/jruby/jruby/builds/189597347) thus this won't do as is. Likely needs work for the corner cases.

kares:
closing since the change had regression and the author did not get back to it

Issue #4353, opened 2016-12-02
Dir.glob returns doubles, if used with recursive directory matchers "**"


MartinKoerner:
### Environment

* Mac OS Sierra 10.12.1
* Directory structure with multiple subdirectories:
```
   tmp
    |\2
       |\abc
       |   \2 - file1.pdf
       |   \4 - file2.pdf
       |\def
       |   \2 - file3.pdf
       |   \4 - file4.pdf
```
### Expected Behavior
```ruby
   # MRI
   # rbenv shell 2.3.1
   # irb
      Dir.glob("tmp").size # = 1
      Dir.glob("tmp/2/abc/2/*.pdf").size # = 1
      Dir.glob("tmp/**/**/**/*.pdf").size # = 4
```

### Actual Behavior
```ruby
   # rbenv shell jruby-9.1.6.0
   # jirb
     Dir.glob("tmp").size # = 1
     Dir.glob("tmp/2/abc/2/*.pdf").size # = 1
     Dir.glob("tmp/**/**/**/*.pdf").size # = 12 <- WRONG!
     Dir.glob("tmp/**/**/**/*.pdf").uniq.size # 4 # doubles are equals according to String
```

sumitmah:
@kares does the following PR fixes the issue?
 https://github.com/jruby/jruby/pull/4428

kares:
@sumitmah well, we can't tell for sure until we do not have a test-case, right? :)

sumitmah:
@kares I can add test case, I was concerned about other scenarios? do we have tests for them?

headius:
The PR seems logical. You can run other test cases by running spec/mspec/bin/mspec ci spec/ruby/core/dir or some tests somewhere in test/mri (probably test/mri/ruby/test_dir.rb). You can also add cases to spec/ruby or test/jruby (rather not add to test/mri) that mimic the above cases.

sumitmah:
Thanks! @headius I'll check the specs.

enebo:
@sumitmah I do not want to land this right before 9.1.7.0 but we will merge afterwards.

sumitmah:
@enebo when is 9.1.7.0  release planned?

sumitmah:
@headius @kares I've added test case

kares:
should be resolved as #4428 got merged.

PR #4428, opened 2017-01-07
[#4353] Ignore consecutive DOUBLE_STAR  in glob_helper.

sumitmah:


kares:
najs! any chance for a cleanup or even better use Dir.tmpdir as a base so we do not have to worry about?

sumitmah:
@kares In the tests you're asking?

kares:
yeah just whatever to cleanup the created dirs e.g. an `ensure FileUtils.rm_r 'testDir_bug4353'`

sumitmah:
@kares done!

kares:
@sumitmah thank you! 
@headius @enebo this seems good to go for 9.1.7 ... need another set of eyes to confirm.

enebo:
@kares I would prefer to wait on this one since it has been wrong since the beginning of JRuby and we can just merge right after 9.1.7.0.

kares:
@enebo np, your call

Issue #4429, opened 2017-01-08
corrupted JIT byte-code for method


kares:
### Environment

Noticed this with the verifier being switched off. Happens in 9.1.5/9.1.6 as well as on master.

* `JAVA_OPTS=-noverify`

Somehow I am failing to reproduce in isolation, even as I try with a empty Rails app (will try more).

Was able to cause a JVM segfault running tests as well as with `rake assets:precompile`.

Problematic piece (from *sass* gem) :

```ruby
def rgba(*args)
  case args.size
    when 2
      color, alpha = args

      assert_type color, :Color, :color
      if calc?(alpha)
        unquoted_string("rgba(#{color.red}, #{color.green}, #{color.blue}, #{alpha})")
      else
        assert_type alpha, :Number, :alpha
        check_alpha_unit alpha, 'rgba'
        color.with(:alpha => alpha.value)
      end
    when 4
      red, green, blue, alpha = args
      if calc?(red) || calc?(green) || calc?(blue) || calc?(alpha)
        unquoted_string("rgba(#{red}, #{green}, #{blue}, #{alpha})")
      else
        rgba(rgb(red, green, blue), alpha)
      end
    else
      raise ArgumentError.new("wrong number of arguments (#{args.size} for 4)")
  end
end
```

### Expected Behavior

* Generated (method) byte-code should not be refected by the verifier.

Thinking JRuby might want to set `-noverify` for some of its **jit** test suite - thus JIT compiler issues would be caught earlier, make sense?

<details>
<summary>Full Log: for the problematic (`Sass::Script::Functions.rgba`) method</summary>
<pre><code>
*** Dumping opt/local/rvm/gems/jruby_minus_9_dot_1_dot_6_dot_0_at_40_enrian/gems/sass_minus_3_dot_4_dot_22/lib/sass/script/functions.RUBY$method$rgba$0 ***
   L0
   L1
    NOP
   L2
    NOP
    ALOAD 0
    ALOAD 1
    ALOAD 3
    LDC 0
    LDC 0
    LDC true
    LDC false
    LDC -1
    ACONST_NULL
    INVOKESTATIC org/jruby/ir/runtime/IRRuntimeHelpers.checkArity (Lorg/jruby/runtime/ThreadContext;Lorg/jruby/parser/StaticScope;[Ljava/lang/Object;IIZZILorg/jruby/runtime/Block$Type;)V
    ALOAD 0
    ALOAD 3
    ICONST_0
    ICONST_0
[...]

PR #4430, opened 2017-01-09
JIT fixnum cases (with a value span <= 32) as a tableswitch

kares:
... instead of falling back to lookupswitch

noticed (while fixing #4429) that only "perfect" consecutive fixnum cases generate a *tableswitch*

javac does some [table/lookup space cost calc to decide](http://hg.openjdk.java.net/jdk8/jdk8/langtools/file/30db5e0aaf83/src/share/classes/com/sun/tools/javac/jvm/Gen.java#l1153), here just a simple size bound (of 32) is implemented

`case 0, 10, 20, 30` will generate a table-switch (filling in 'wholes' e.g. case 1: default)
while `case 0, 10, 50` will generate a lookup-switch just like before

### DIFF
--- a/core/src/main/java/org/jruby/ir/targets/JVMVisitor.java
+++ b/core/src/main/java/org/jruby/ir/targets/JVMVisitor.java
@@ -848,6 +848,11 @@ public class JVMVisitor extends IRVisitor {
 
     }
 
+    // max (case) values span that generates a tableswitch instruction
+    //   case 0, 2, 4, 8, 16, 32 still generates a tableswitch (filling in holes)
+    //   case 0, 10, 100 fallbacks to generating a lookupswitch
+    private static final int MAX_TABLE_SWITCH_SIZE = 32 + 1;
+
     public void BSwitchInstr(BSwitchInstr bswitchinstr) {
         visit(bswitchinstr.getCaseOperand());
         jvmAdapter().dup();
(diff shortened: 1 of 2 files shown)


kares:
... feel free to postpone to next release if you feel like it could use some more work (e.g. a different max)

enebo:
@kares It seems reasonable to me but I wonder how often this happens to warrant the extra code.  I guess JVM does it so they thought it was worth it.

Second nit is you are still not putting your closing } onto same line :)

kares:
yep, sorry about the style changes (although in this case I might have done it on purpose - the `if ()` gets lost in dense lines bellow and above but I shall respect your style taste :). 

since an update some time back my IDEA has never been the same - was glad to get the jruby project working again I need to spent some time setting it up as I need to manually undo changes it does for me.

will do a micro bench test whether its worth the change (should be since its expected to jump directly skiping the lookup part) -> will post back with numbers.
... such casing should be not that uncommon e.g. for `def my(*args);` methods

olleolleolle:
Super-interesting deep-dive into a detail.

(Hi @kares, could you choose another name than `p` for a local variable in the Ruby tests?

My eyes – and some code highlighters – read that as `Kernel#p`.)

kares:
thanks @olleolleolle ... sorry not into style updates really at this point (this code might get rejected).

have done some micro benchmarks previously (as @enebo suggested) and the change seemed only make a little difference, so I am not sure if tableswitch makes sense over lookupswitch if it 'complicates' code. but than maybe I have done it wrong - someone better understanding the difference in terms of JVM (and native code jitting) should decide whether this makes sense or not really ...

headius:
This looks fine. Merging so it can bake for 9.1.9.0.

Issue #4416, opened 2016-12-30
The i18n gem does not work in AOT mode


glebm:
### Environment

```console
$ jruby -v
jruby 9.1.6.0 (2.3.1) 2016-11-09 0150a76 Java HotSpot(TM) 64-Bit Server VM 25.45-b02 on 1.8.0_45-b14 +jit [darwin-x86_64]
$ echo $JRUBY_OPTS                                                                                                                                       
-Xcompile.mode=FORCE
$ uname -a
Darwin Macintosh.local 16.3.0 Darwin Kernel Version 16.3.0: Thu Nov 17 20:23:58 PST 2016; root:xnu-3789.31.2~1/RELEASE_X86_64 x86_64
```

### Expected Behavior

The i18n gem test suite passes in AOT mode:

```console
$ git clone https://github.com/svenfuchs/i18n.git
$ cd i18n
$ bundle
$ JRUBY_OPTS="-Xcompile.mode=FORCE" bundle exec rake
```

### Actual Behavior

```
# Running:

NoMethodError: private method `run' called for #<I18nBackendPluralizationTest:0x1a02ebe0>
     run_one_method at /Users/glebm/.rvm/gems/jruby-9.1.6.0@i18n/gems/minitest-5.10.1/lib/minitest.rb:822
     run_one_method at /Users/glebm/.rvm/gems/jruby-9.1.6.0@i18n/gems/minitest-5.10.1/lib/minitest.rb:323
       block in run at /Users/glebm/.rvm/gems/jruby-9.1.6.0@i18n/gems/minitest-5.10.1/lib/minitest.rb:310
               each at org/jruby/RubyArray.java:1734
       block in run at /Users/glebm/.rvm/gems/jruby-9.1.6.0@i18n/gems/minitest-5.10.1/lib/minitest.rb:309
          on_signal at /Users/glebm/.rvm/gems/jruby-9.1.6.0@i18n/gems/minitest-5.10.1/lib/minitest.rb:349
  with_info_handler at /Users/glebm/.rvm/gems/jruby-9.1.6.0@i18n/gems/minitest-5.10.1/lib/minitest.rb:336
                run at /Users/glebm/.rvm/gems/jruby-9.1.6.0@i18n/gems/minitest-5.10.1/lib/minitest.rb:308
     block in __run at /Users/glebm/.rvm/gems/jruby-9.1.6.0@i18n/gems/minitest-5.10.1/lib/minitest.rb:158
[...]

PR #4431, opened 2017-01-09
Set toplevel module

headius:
This includes a fix for #4416 (setting top-level module when running with -X+C) and enables the `test:jruby:aot` suite to verify the AOT mode.


kares:
excellent, wanted to add `test:jruby:aot` as well :)

PR #4434, opened 2017-01-11
[Truffle] Add check if property is undefined for uncached specialization

bjfish:
I wanted to get code review on this since I am not too familiar with this.

I was getting `instance variable not defined` errors on this line when running Addressable specs: https://github.com/sporkmonger/addressable/blob/4983bf53765c83b5f06b8b9707e6c4cbc2225acb/lib/addressable/uri.rb#L2488

This change resolved these errors to reduce Addressable spec failures from 62 => 9.

If I understand this correctly, it appears to me that the cached node was checking to see if a property was defined but the uncached version was not.

### DIFF
--- a/truffle/src/main/java/org/jruby/truffle/language/objects/ReadObjectFieldNode.java
+++ b/truffle/src/main/java/org/jruby/truffle/language/objects/ReadObjectFieldNode.java
@@ -58,6 +58,9 @@ public abstract class ReadObjectFieldNode extends RubyBaseNode {
     @TruffleBoundary
     @Specialization(contains = { "readObjectFieldCached", "updateShapeAndRead" })
     protected Object readObjectFieldUncached(DynamicObject receiver) {
+        if(getProperty(receiver.getShape(), name) == null){
+            return defaultValue;
+        }
         return receiver.get(name, defaultValue);
     }
 


truffle/src/main/java/org/jruby/truffle/language/objects/ReadObjectFieldNode.java
```diff
@@ -58,6 +58,9 @@ public Object updateShapeAndRead(DynamicObject object) {
     @TruffleBoundary
     @Specialization(contains = { "readObjectFieldCached", "updateShapeAndRead" })
     protected Object readObjectFieldUncached(DynamicObject receiver) {
+        if(getProperty(receiver.getShape(), name) == null){
+            return defaultValue;
+        }
         return receiver.get(name, defaultValue);
```
eregon: This line intended to be the same since http://lafo.ssw.uni-linz.ac.at/javadoc/truffle/latest/com/oracle/truffle/api/object/DynamicObject.html#get-java.lang.Object-java.lang.Object- should return `defaultValue` if the property does not exist.
However, this is a problem as we have a "removed property flag" and so we need to use `PropertyFlags.isDefined(property)`. I'll fix, thanks for the report and analysis!

PR #4440, opened 2017-01-14
Regexp#match?

sumitmah:
* Regexp#match?
    This returns bool and doesn't save backref.

### DIFF
--- a/core/src/main/java/org/jruby/RubyRegexp.java
+++ b/core/src/main/java/org/jruby/RubyRegexp.java
@@ -1107,6 +1107,94 @@ public class RubyRegexp extends RubyObject implements ReOptions, EncodingCapable
         return match19Common(context, str, 0, true, block);
     }
 
+    /**
+     * MRI: rb_reg_match_m_p
+     *
+     * @param context thread context
+     * @param str to be matched
+     */
+    @JRubyMethod(name = "match?")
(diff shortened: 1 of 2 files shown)


core/src/main/java/org/jruby/RubyRegexp.java
```diff
@@ -1107,6 +1107,94 @@ public IRubyObject match_m19(ThreadContext context, IRubyObject str, Block block
...
+            }
+        } else {
+            this.useCount--;
+        }
+
+        if (matchResult < 0) {
+            if (matchResult == -1) {
+                return runtime.getFalse();
+            } else {
+                throw runtime.newRegexpError(exception == null ? "FIXME: missing message" : exception.getMessage());
```
sumitmah: @headius Do we handle all of these error codes in jruby/joni ?https://github.com/ruby/ruby/blob/88d667ee444948b8c3bccafaadc653da0a319146/regerror.c

kares:
Thanks, this is a Ruby 2.4 feature - you should read https://github.com/jruby/jruby/issues/4293 and check the appropriate branch (its not yet to be merged onto master - will be at some point later probably after 9.1.8 is released). 
You'll find there that the `match?` feature is already implemented there. Closing as duplicate.

PR #4441, opened 2017-01-14
Remove TruffleRuby

chrisseaton:
Code still available at https://github.com/graalvm/truffleruby


claudiug:
may I ask why remove rubytruffle? That is a huge PR :)

chrisseaton:
@claudiug this mailing list post explains the reasons http://lists.ruby-lang.org/pipermail/jruby/2017-January/000511.html

TruffleRuby doesn't have any source code dependencies on JRuby anymore and hasn't for a while. We're having to tidy up our code base quite a bit in order to prepare for an ahead-of-time compiled (no JVM required) version of TruffleRuby that we are going release soon, and with both of those things it seemed time to start to develop in a separate repo.

claudiug:
@chrisseaton thx for the quite fast reply.
AOT for ruby? wow

chrisseaton:
It's AOT for the Java program that implements an interpreter for Ruby, not AOT for the Ruby program. It's confusing I know.

It allows us to build a single, statically linked `truffleruby` executable that doesn't need a JVM to run, so much like the `ruby` executable you get with MRI. This reduces startup time by orders of magnitude, memory footprint, and deployment issues.

PR #4447, opened 2017-01-19
Fix GzipReader "position" update for #each and #each_line.

manu-crealytics:


enebo:
@manu-crealytics huge bonus if you add something to test this in ruby/spec project.

manu-crealytics:
I had already done that as part of a different issue, that's why I discovered this bug ;)
Here the PR (already merged): https://github.com/ruby/spec/pull/383

enebo:
@manu-crealytics Thanks a ton then for fixing our issue after that!

PR #4470, opened 2017-02-02
Implement `String#casecmp?` and `Symbol#casecmp?`

kcdragon:
* Adds implementation of `casecmp?` based on downcasing the string and comparing
* Adds unit tests for `casecmp?` from MRI

There is a test failing in each of the tests (the last assertion in each method) I brought over. This is because the MRI implementation (https://github.com/ruby/ruby/commit/ad619e02c465eebd15fa57dc658f8e042decdebb) of `casecmp?` relies on passing the `:fold` option to `downcase` however that option has not been added yet in JRuby. It is referenced in #4293 though.

I decided to keep the failing tests in for completeness since they will need to be fixed anyway for 2.4 support. I figured that this PR would be fine without those specific cases supported but maybe it would be better to wait until `:fold` is implemented.

See #4293

### DIFF
--- a/core/src/main/java/org/jruby/RubyString.java
+++ b/core/src/main/java/org/jruby/RubyString.java
@@ -1596,8 +1596,14 @@ public class RubyString extends RubyObject implements EncodingCapable, MarshalEn
 
     @JRubyMethod(name = "casecmp?")
     public IRubyObject casecmp_p(ThreadContext context, IRubyObject other) {
+        Ruby runtime = context.runtime;
+        RubyString otherStr = other.convertToString();
+
+        Encoding enc = StringSupport.areCompatible(this, otherStr);
+        if (enc == null) return runtime.getNil();
+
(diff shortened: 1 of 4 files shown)


kares:
excellent, any reason why the encoding compatible check is not being implemented (like in MRI) ?

kcdragon:
> excellent, any reason why the encoding compatible check is not being implemented (like in MRI) ?

@kares 

There's no reason. I can add that. Are you referring to something like this being done?

```
Encoding enc = StringSupport.areCompatible(this, otherStr);
if (enc == null) return runtime.getNil();
```
https://github.com/jruby/jruby/blob/master/core/src/main/java/org/jruby/RubyString.java#L1572-L1573

kares:
yy

kcdragon:
@kares I added the encoding compatibility check.

headius:
Looks ok to me!

PR #4471, opened 2017-02-03
Make sure to iterate over all keys when cancelling keys in a SelectorPool

cheister:
When we updated our server from 9.1.2.0 to 9.1.7.0 we started seeing a file descriptor leak which, if left for long enough, would run the machine out of file descriptors.

We tracked down the problem to https://github.com/jruby/jruby/pull/3952 which changed how SelectionKeys were closed.  

It looks like if exceptions aren't caught while calling `cancel` on the SelectionKeys then the loop can exit early and leave some keys un-cancelled.

When we went back to the same key cancelling code that was used in 9.1.2.0 the problem went away.

### DIFF
--- a/core/src/main/java/org/jruby/util/io/SelectorPool.java
+++ b/core/src/main/java/org/jruby/util/io/SelectorPool.java
@@ -85,8 +85,11 @@ public class SelectorPool {
      * @param selector the selector to put back
      */
     public void put(Selector selector) {
-        Iterator<SelectionKey> key_iterator = selector.keys().iterator();
-        while(key_iterator.hasNext()) key_iterator.next().cancel();
+        for (SelectionKey key : selector.keys()) {
+            if (key != null) {
+                key.cancel();
+            }
+        }
 
         try {
             selector.selectNow();


cheister:
cc @mohamedhafez

mohamedhafez:
Hmm, what exceptions are being thrown at https://github.com/jruby/jruby/pull/4471/files#diff-f99fc044de4e2aa68b3bb1f212348b07R92 ? If the cancelling fails for some reason and the `SelectionKey` is still valid and uncancelled, then it will lead to old `SelectionKeys` being returned in `Selectors` returned by `retrieveFromPool`, which is what was causing the problems that lead me to file https://github.com/jruby/jruby/pull/3952 in the first place.

If the `SelectionKey` isn't valid (i.e. calling `isValid` on it returns false), then it won't cause the problems I observed and I suppose it would be fine to ignore them the way we were doing prior to my PR, but perhaps it would be best to handle them?

cheister:
The exceptions are getting swallowed somewhere so I haven't seen what they are. I could try adding logging to capture them if you're interested?

From what you're saying it sounds like I should remove the `if (key.isValid())` check from the loop?

mohamedhafez:
@cheister yeah could you put in logging to see what the exceptions are, and also to check the status of `key.isValid()` inside the catch statement at https://github.com/jruby/jruby/pull/4471/files#diff-f99fc044de4e2aa68b3bb1f212348b07R93

cheister:
The exception being thrown was an NPE because the key was null.  So changing the code to check for null also makes it work.

headius:
Great find, thank you!

Issue #4478, opened 2017-02-06
Block doesn't retain it's binding through instance_eval


sofaking:
```ruby
module Outer
  def foo(&block)
    Inner.new.plain_method {}
  end

  def bar(&block)
    Inner.new(&block)
  end

  class Inner
    def initialize(&block)
      @block = block

      instance_exec(&block) if block_given?
    end

    def plain_method(&block)
      block.binding.receiver.to_s
    end

    def method_missing(method_name, *args)
      @block.binding.receiver.to_s
    end
  end
end

class Extender
  extend Outer

  puts foo {}
  bar do
    puts inexisting_method 'abc'
  end
end
```

### Environment

* jruby 9.1.7.0 (2.3.1) 2017-01-11 68056ae Java HotSpot(TM) 64-Bit Server VM 25.102-b14 on 1.8.0_102-b14 +jit [darwin-x86_64]
* Darwin Antons-MacBook-Pro.local 16.4.0 Darwin Kernel Version 16.4.0: Thu Dec 22 22:53:21 PST 2016; root:xnu-3789.41.3~3/RELEASE_X86_64 x86_64

### Expected Behavior

I expect this code to return 'Extender' 2 times, in the same manner as MRI does.

### Actual Behavior

For some reason, after instance_eval a passed block looses it's binding, and becomes bound to the object it was instance_eval-ed. So the actual result is 
```
Extender
#<Outer::Inner:0x4923ab24>
```

Sorry, probably I could come up with a simpler example. I've just roughly cropped my real application to the smallest possible size.

headius:
Yep, this is a bug. I've long known our `instance_eval` permanently modifies the incoming block, but we had no clear example of failures that would cause. Kudos!

This should be fixable by cloning more of the block. This will impact the performance of instance_eval (more to clone, more to clean up) but we may be able to do it only for blocks that have been reified into procs (since literal blocks are only alive for the duration of the instance_eval call).

headius:
FWIW this is a much better repro than we usually get. We appreciate the extra effort to narrow this down.

headius:
Ahh boo, I pushed #4485 for this but forgot to tag this issue. If it looks ok I'll re-commit that change.

headius:
Ugh, and #4485 was not even for this bug. Nevermind.

headius:
Oops.

headius:
Well I'm not sure if it was 55b3080, but this appears to be fixed on master.

```
[...]

Issue #4475, opened 2017-02-04
Activating refinements in binding.eval doesn't stick outside of the evaluated string.


najamelan:
### Environment

`jruby -v`    : jruby 9.1.7.0 (2.3.1) 2017-01-11 68056ae OpenJDK 64-Bit Server VM 25.121-b13 on 1.8.0_121-8u121-b13-2-b13 +jit [linux-x86_64]
`uname -a` : Linux computer 4.9.0-1-amd64 #1 SMP Debian 4.9.2-2 (2017-01-12) x86_64 GNU/Linux

### Expected Behavior
```ruby
module Refine

	refine ::Array do

		def first= value

			self[ 0 ] = value

		end

	end

end

binding.eval 'using Refine'
binding.eval '[].first = 4'

p = [].first = 'foo'

puts p.inspect
```

### Actual Behavior

In JRuby both uses of #first= will fail, where in MRI both work. See the outcome of [this issue](https://bugs.ruby-lang.org/issues/10818).

```
NoMethodError: undefined method `first=' for []:Array
Did you mean?  first
  <main> at test.rb:18
```

headius:
The second behavior here is likely just an odd side effect of the eval's binding and the surrounding scope being the same. It fails for the same reason that the first behavior fails.

May I ask what purpose this behavior serves? Why would you ever want to do this?

najamelan:
Have a library which exposes a bunch of refinements, that users can add in one go without writing `using` lines for every one of them on top of every single source file that uses them with:
```ruby
Susu.refine binding

# or

Susu:refine binding, [ :Array, :Hash ]
```
For me refinements are pretty useless if I can't do this because of the high quantity of boilerplate. I will probably even further split them up to separate refinements that change the behaviour of existing standard ruby methods and refinements that only add new methods.

```ruby
REFINES =
{
	Array:   'using Susu::Refine::Array'   ,
	Date:    'using Susu::Refine::Date'    ,
	Hash:    'using Susu::Refine::Hash'    ,
	Module:  'using Susu::Refine::Module'  ,
	Numeric: 'using Susu::Refine::Numeric' ,
	String:  'using Susu::Refine::String'  ,
	Time:    'using Susu::Refine::Time'    ,

	Fs:      'using Susu::Fs::Refine'      ,
	Options: 'using Susu::Options::Refine' ,
}



def self.refine context, which = :all

	which.kind_of?( Array ) or which = [ which ]

[...]

PR #4485, opened 2017-02-09
Allow refinements in one eval to be seen in the next.

headius:
This change just carries the original dynamic scope along into
the cloned binding used for an eval. This allows the "cref" data
for refinements to be progressively updated.


headius:
Fix for #4478. Recommit with bug number.

headius:
This fix is not for #4478, it is for #4475.

headius:
Merged separately in 55b3080.

PR #4375, opened 2016-12-09
[ruby 2.4] Implemented Integer#digits

herwinw:
It passes all the following tests of TestInteger of MRI:

* test_digits
* test_digits_for_negative_numbers
* test_digits_for_invalid_base_numbers
* test_digits_for_non_integral_base_numbers
* test_digits_for_non_numeric_base_argument

PR #4490, opened 2017-02-13
[ruby 2.4] Fixes for Integer#digits in PR #4375

whwilder:
I've split the implementation for Integer#digits amongst Fixnum and Bignum, which should fix the issues in #4375 

ref #4293


core/src/main/java/org/jruby/RubyFixnum.java
```diff
@@ -302,6 +302,53 @@ public IRubyObject times(ThreadContext context, Block block) {
         return RubyEnumerator.enumeratorizeWithSize(context, this, "times", timesSizeFn(context.runtime));
     }
 
+    /** rb_fix_digits
+     *
+     */
+    @Override
+    public RubyArray digits(ThreadContext context, IRubyObject base) {
+
+        long self = ((RubyFixnum)this).getLongValue();
```
headius: This is not necessary...`this` is already known to be RubyFixnum.

core/src/main/java/org/jruby/RubyBignum.java
```diff
@@ -236,6 +236,54 @@ public static BigInteger long2big(long arg) {
      *  ================
      */
 
+    /** rb_big_digits
+     *
+     */
+    @Override
+    public RubyArray digits(ThreadContext context, IRubyObject base) {
+        BigInteger self = (this).getValue();
```
headius: You could just access the field here, or at least call it without the parens. I assume this is just a mistake converting the RubyFixnum-only logic.

core/src/main/java/org/jruby/RubyBignum.java
```diff
@@ -236,6 +236,54 @@ public static BigInteger long2big(long arg) {
...
+        if (self.compareTo(new BigInteger("0")) == -1) {
+            throw context.getRuntime().newMathDomainError("out of domain");
```
headius: Throw context.runtime into a local variable at the top and reuse it, in the style of other methods.

whwilder:
Thanks for the help!  Squashed a few other style issues too.

whwilder:
@headius Is there anything preventing this PR from being merged, or is it just a low priority for the moment?

kares:
@whwilder great work - wanted to get `#digits` in as well, thanks :)

PR #4496, opened 2017-02-19
Use ipaddr to compare the sender's address

### LINKED ISSUE #3663: Never-ending getaddress() call when using compressed IPv6 nameservers in /etc/resolv.conf
Hi,

I managed to reproduce the following bug in a CentOS 7.2 box with only IPv6 nameservers and dual stack, jRuby 1.7.20.1 and java-1.7.0-openjdk-1.7.0.95-2.6.4.0.el7_2. The same code works fine on MRI 2.0.0.  Perhaps this has already been fixed in newer versions, as to me the bug looks too obvious to be still alive. Anyway, here we go:

```
# cat /etc/resolv.conf
# generated by /usr/sbin/dhclient-script
search cern.ch.
nameserver 2001:1458:201:1000::5
nameserver 2001:1458:201:1100::5
```

The following program and the resolv.conf above makes the interpreter hang for a long time and return a failure:

``` ruby
require 'resolv'
puts Resolv.getaddress 'web.cern.ch'
```

```
# time java -cp /usr/share/puppetserver/puppet-server-release.jar clojure.main -m puppetlabs.puppetserver.cli.ruby --config /etc/puppetserver/conf.d -- -e "require 'resolv'; puts Resolv.getaddress 'web.cern.ch'"
Resolv::ResolvError: no address for web.cern.ch
  getaddress at /usr/share/puppetserver/puppet-server-release.jar!/META-INF/jruby.home/lib/ruby/1.9/resolv.rb:98
  getaddress at /usr/share/puppetserver/puppet-server-release.jar!/META-INF/jruby.home/lib/ruby/1.9/resolv.rb:48
      (root) at -e:1
      invoke at jruby_puppet_core.clj:232
      invoke at jruby_puppet_core.clj:226
      invoke at subcommand.clj:38
    doInvoke at ruby.clj:7
      invoke at core.clj:624
      invoke at main.clj:315
    doInvoke at main.clj:420

real    2m51.910s
user    0m21.020s
sys 0m0.654s
```

These are the last lines in a debugging session before the first exception is raised. The program blocks in the select() call until it times out and ResolvTimeout is raised.

```
#0:/usr/share/puppetserver/puppet-server-release.jar!/META-INF/jruby.home/lib/ruby/1.9/resolv.rb:676:Resolv::DNS::Requester:-:           if s = @senders[[from,msg.id]]
#0:/usr/share/puppetserver/puppet-server-release.jar!/META-INF/jruby.home/lib/ruby/1.9/resolv.rb:654:Resolv::DNS::Requester:-:           now = Time.now
#0:/usr/share/puppetserver/puppet-server-release.jar!/META-INF/jruby.home/lib/ruby/1.9/resolv.rb:655:Resolv::DNS::Requester:-:           timeout = timelimit - now
#0:/usr/share/puppetserver/puppet-server-release.jar!/META-INF/jruby.home/lib/ruby/1.9/resolv.rb:656:Resolv::DNS::Requester:-:           if timeout <= 0
#0:/usr/share/puppetserver/puppet-server-release.jar!/META-INF/jruby.home/lib/ruby/1.9/resolv.rb:659:Resolv::DNS::Requester:-:           select_result = IO.select(@socks, nil, nil, timeout)
#0:/usr/share/puppetserver/puppet-server-release.jar!/META-INF/jruby.home/lib/ruby/1.9/resolv.rb:660:Resolv::DNS::Requester:-:           if !select_result
#0:/usr/share/puppetserver/puppet-server-release.jar!/META-INF/jruby.home/lib/ruby/1.9/resolv.rb:661:Resolv::DNS::Requester:-:             raise ResolvTimeout
#0:/usr/share/puppetserver/puppet-server-release.jar!/META-INF/jruby.home/lib/ruby/1.9/resolv.rb:661:Resolv::DNS::Requester:^:             raise ResolvTimeout
/usr/share/puppetserver/puppet-server-release.jar!/META-INF/jruby.home/lib/ruby/1.9/resolv.rb:661: `Resolv::ResolvTimeout' (Resolv::ResolvTimeout)
```

I think that this is happening because during the previous iteration [this statement](https://github.com/jruby/jruby/blob/1.7.20.1/lib/ruby/1.9/resolv.rb#L676) was evaluated as false because 

> @senders[[from,msg.id]]

was nil. Why was it nil? Because the search keys didn't match, as 'from' contains the uncompressed flavor of the IPv6 address of the DNS server, whereas 'senders' has the compacted form (presumably coming from /etc/resolv.conf):

```
(rdb:1) p from
["2001:1458:201:1000:0:0:0:5", 53]
(rdb:1) p msg.id
12941
(rdb:1) p @senders
{[["2001:1458:201:1000::5", 53], 12941]=>#<Resolv::DNS::Requester::UnconnectedUDP::Sender:0x4605f6fa @host="2001:1458:201:1000::5", @msg="2\x8D\x01\x00\x00\x01\x00\x00\x00\x00\x00\x00\x03web\x04cern\x02ch\x00\x00\x01\x00\x01", @data=#<Resolv::DNS::Name: web.cern.ch.>, @sock=#<UDPSocket:fd 28>, @port=53>}
```

This situation leads to the outer loop not stopping (see "unexpected DNS message ignored"), therefore the program executes IO.select again but there's nothing to read anymore so ResolvTimeout is raised. This exception is probably caught by the caller further up and at some point [.request](https://github.com/jruby/jruby/blob/1.7.20.1/lib/ruby/1.9/resolv.rb#L650) is called again, leading to another loop. This situation repeats many times after a few minutes the call returns a ResolvError all the way up back to the user.

Handcrafting resolv.conf so all addreses are expanded there makes resolv.rb happy:

```
# generated by /usr/sbin/dhclient-script
search cern.ch.
nameserver 2001:1458:201:1000:0:0:0:5
nameserver 2001:1458:201:1100:0:0:0:5
```

This way the program quickly exits without hanging:

```
# time java -cp /usr/share/puppetserver/puppet-server-release.jar clojure.main -m puppetlabs.puppetserver.cli.ruby --config /etc/puppetserver/conf.d -- -e "require 'resolv'; puts Resolv.getaddress 'web.cern.ch'"
188.184.9.235

real    0m10.945s
user    0m19.857s
sys 0m0.553s
```

Configuring a Resolv::DNS object by hand with a compressed IPv6 address seems to work.

```
# cat /etc/resolv.conf
# generated by /usr/sbin/dhclient-script
search cern.ch.
nameserver 2001:1458:201:1000::5
nameserver 2001:1458:201:1100::5
....
irb(main):008:0* require 'resolv'
=> true
irb(main):009:0> Resolv::DNS.new(:nameserver => '2001:1458:201:1000::5').getaddress "web.cern.ch"
=> #<Resolv::IPv4 188.184.9.235>
irb(main):010:0> Resolv::DNS.new().getaddress "web.cern.ch"
# Hangs...
```

nbarrientos:
When resolving names using compressed IPv6-only DNS servers, it could
happen that when processing the response of a name query, the IP of the
sender is not compressed so using a simple string comparison is not
safe.

This patch encapsulates the from address and the address of the sender in
IPAddr objects so the comparison is safe and just works in all cases.

This fixes #3663.

### DIFF
--- a/lib/ruby/stdlib/resolv.rb
+++ b/lib/ruby/stdlib/resolv.rb
@@ -4,6 +4,7 @@ require 'socket'
 require 'timeout'
 require 'thread'
 require 'io/wait'
+require 'ipaddr'
 
 begin
   require 'securerandom'
@@ -768,13 +769,13 @@ class Resolv
 
         def recv_reply(readable_socks)
           reply, from = readable_socks[0].recvfrom(UDPSize)
-          return reply, [from[3],from[1]]
+          return reply, [IPAddr.new(from[3]),from[1]]
         end
 
         def sender(msg, data, host, port=Port)
           sock = @socks_hash[host.index(':') ? "::" : "0.0.0.0"]
           return nil if !sock
-          service = [host, port]
+          service = [IPAddr.new(host), port]
           id = DNS.allocate_request_id(host, port)
           request = msg.encode
           request[0,2] = [id].pack('n')


headius:
This looks fine to me, and we'll probably merge it, but I have a few concerns.

We normally try to avoid patching the standard library if we can help it. Especially like this, where a new dependency is being introduced. But perhaps MRI has this problem too when running in an IPv6 environment?

I don't have any IPv6 setup right now. Any chance you can try to reproduce with MRI?

nbarrientos:
Hi,

Thanks for looking at the patch.

I could not reproduce the problem with MRI, see [this comment](https://github.com/jruby/jruby/issues/3663#issuecomment-183701789). But anyway:

```shell
# grep nameserver /etc/resolv.conf
nameserver 2001:1458:201:1100::5
nameserver 2001:1458:201:1100::5
# ruby -v
ruby 2.0.0p648 (2015-12-16) [x86_64-linux]
# time ruby -e "require 'resolv'; puts Resolv.getaddress 'web.cern.ch'"
188.184.9.235

real	0m0.083s
user	0m0.061s
sys	0m0.021s
```

Same machine, same environment, but on top of jRuby:

```
# jruby -v
jruby 9.1.7.0 (2.3.1) 2017-01-11 68056ae OpenJDK 64-Bit Server VM 25.121-b13 on 1.8.0_121-b13 +jit [linux-x86_64]
# time jruby -e "require 'resolv'; puts Resolv.getaddress 'web.cern.ch'"
Resolv::ResolvError: no address for web.cern.ch
  getaddress at /root/jruby-9.1.7.0/lib/ruby/stdlib/resolv.rb:100
  getaddress at /root/jruby-9.1.7.0/lib/ruby/stdlib/resolv.rb:50
      <main> at -e:1

real	2m45.023s
user	0m5.508s
sys	0m0.343s
#
```

I'm happy to perform other tests with MRI, just let me know :)

Do you see feasible to backport the patch to 1.7? That'd be awesome!

headius:
Ok, bummer that MRI works ok here but I think we can move forward with this patch for now.

@enebo Any objections to this patch other than what I mentioned above? It would at least get people on their feet while we investigate the root cause.

enebo:
@headius I think if we can open up a new issue to figure out why we are different and close the issue around this PR we can land it.  Then it will work as a workaround until we dig deeper.  I would rather we release sooner than later...and hey it will not be our first modification to stdlib files :|

I would like to close the original report so we can display this is now working in the release in our notes (but we can link to it)

headius:
@enebo Agreed.

nbarrientos:
Thanks. Any chance to see it backported to 1.7?

PR #4498, opened 2017-02-21
Separate varargs and specific-arity names in JIT. Fixes #4482

### LINKED ISSUE #4482: Doing require_relative too often leads to TypeError
### Environment

* ruby -v  `jruby 9.1.8.0-SNAPSHOT (2.3.1) 2017-02-06 3f97c7a Java HotSpot(TM) 64-Bit Server VM 25.111-b14 on 1.8.0_111-b14 +indy +jit [linux-x86_64]`
* uname -a `Linux ok 4.4.0-59-generic #80-Ubuntu SMP Fri Jan 6 17:47:47 UTC 2017 x86_64 x86_64 x86_64 GNU/Linux`

### Setup

Empty file: `a.rb`

Contents of `b.rb`, in the same directory:

```
400.times{
  require_relative 'a'
}
```

### Expected Behavior

* Should print nothing

### Actual Behavior

* Throws following error:

```
TypeError: no implicit conversion of nil into String
          realpath at org/jruby/RubyFile.java:862
  require_relative at uri:classloader:/jruby/kernel/kernel.rb:11
  require_relative at uri:classloader:/jruby/kernel/kernel.rb:-1
     block in b.rb at b.rb:2
             times at org/jruby/RubyFixnum.java:299
            <main> at b.rb:1
```

Works find for a smaller iteration count, for example, 4.

headius:
Originally, varargs and specific-arity paths were emitted in toto
as their own JVM method bodies. With recent changes, the varargs
path now calls the specific-arity path, causing an extra
synthetic entry in the JVM stack trace. This confounded the
backtrace miner, since it had no marker with which to omit the
synthetic version, and so Kernel#caller and friends contained an
additional invalid entry. This in turn caused require_relative to
fail to find the caller's path, resulting in a nil path value
passed to File.realpath, which triggered this bug.

It likely only affected invokedynamic because our indy logic must
be calling no-arg methods via the varargs path (a separate issue
to be fixed).

My modification here explicitly separates the synthetic varargs
name from the specific-arity name by adding a known varargs marker
that can be filtered out when mining the backtrace.

This is a big hacky; the state being managed by the jit is very
ad-hoc and prone to bugs. It will need some additional cleanup
after 9.1.8.0.

See #4482.

### DIFF
--- a/core/src/main/java/org/jruby/compiler/BlockJITTask.java
+++ b/core/src/main/java/org/jruby/compiler/BlockJITTask.java
@@ -76,7 +76,7 @@ class BlockJITTask implements Runnable {
                 JITCompiler.log(body.getImplementationClass(), body.getFile(), body.getLine(), className + "." + methodName, "done jitting");
             }
 
-            String jittedName = context.getJittedName();
+            String jittedName = context.getVariableName();
 
             // blocks only have variable-arity
             body.completeBuild(
--- a/core/src/main/java/org/jruby/compiler/MethodJITTask.java
+++ b/core/src/main/java/org/jruby/compiler/MethodJITTask.java
@@ -112,8 +112,8 @@ class MethodJITTask implements Runnable {
                 JITCompiler.log(method.getImplementationClass(), method.getFile(), method.getLine(), className + '.' + methodName, "done jitting");
             }
 
-            final String jittedName = context.getJittedName();
-            MethodHandle variable = JITCompiler.PUBLIC_LOOKUP.findStatic(sourceClass, jittedName, context.getNativeSignature(-1));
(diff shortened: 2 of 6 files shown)


headius:
The additional "issue" I mentioned is a non-issue for now. Indy will always call through the varargs path the first time a call site is bound, since it only has a single varargs binding path in InvokeSite. Fixing the issue would require having separate bootstrap logic for each arity, and would only save us the varargs array and overhead on the first (re)bind of a given call site.

Since most indy sites will eventually settle into one or more direct variable OR specific paths, this is not necessary. However, we may want to fix it for sites that fail to bind, such as those that have too many targets or that are rebound too many times.

PR #4502, opened 2017-02-23
Implemented optional buffer argument for Array#pack

whwilder:
ref #4293

### DIFF
--- a/core/src/main/java/org/jruby/RubyArray.java
+++ b/core/src/main/java/org/jruby/RubyArray.java
@@ -40,6 +40,7 @@ package org.jruby;
 import org.jcodings.specific.USASCIIEncoding;
 import org.jruby.anno.JRubyClass;
 import org.jruby.anno.JRubyMethod;
+import org.jruby.ast.util.ArgsUtil;
 import org.jruby.common.IRubyWarnings.ID;
 import org.jruby.java.util.ArrayUtils;
 import org.jruby.javasupport.JavaUtil;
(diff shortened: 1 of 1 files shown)


core/src/main/java/org/jruby/RubyArray.java
```diff
@@ -4639,6 +4640,26 @@ public RubyString pack(ThreadContext context, IRubyObject obj) {
         }
     }
 
+    @JRubyMethod(name = "pack")
+    public RubyString pack(ThreadContext context, IRubyObject obj, IRubyObject maybeOpts) {
+        Ruby runtime = context.runtime;
+        IRubyObject opts = ArgsUtil.getOptionsArg(runtime, maybeOpts);
+        RubyString str = RubyString.newEmptyString(runtime);
```
kares: preferably use `RubyString.newString()` the newEmptyString is for cases such as `return ''`

core/src/main/java/org/jruby/RubyArray.java
```diff
@@ -4639,6 +4640,26 @@ public RubyString pack(ThreadContext context, IRubyObject obj) {
...
+        RubyString iFmt = obj.convertToString();
+        try {
+            if (!opts.isNil()){
+                IRubyObject buffer = ((RubyHash) opts).fastARef(runtime.newSymbol("buffer"));
+                if (buffer != null) {
+                    str = (RubyString) buffer;
```
kares: maybe the `RubyString` allocation (above) can be delayed until its sure a buffer is not provided?

core/src/main/java/org/jruby/RubyArray.java
```diff
@@ -4644,19 +4644,21 @@ public RubyString pack(ThreadContext context, IRubyObject obj) {
...
+                str = (RubyString) buffer;
+            } else {
+                str = RubyString.newString(runtime, "");
             }
+            return str.append(pack(context, obj));
 
-            return str.append(Pack.pack(context, context.runtime, this, iFmt));
         } catch (ArrayIndexOutOfBoundsException e) {
             throw concurrentModification(context.runtime, e);
+        } catch (ClassCastException e){
```
kares: why is this introduced here? is it due the `obj.convertToString()` removal? (would keep it than)

whwilder: In my original implementation I hadn't considered what would happen if the buffer were nil or some other class besides String.  MRI Ruby throws a type error, and this seemed to be the most straightforward way of mirroring that behavior.

kares: `buffer.convertToString()` will do that for you - raising a `TypeError` e.g. on `nil`
... `ClassCastException` isn't very bad but theoretically might happen from some other place
if you could update to get rid of it that would be great ... the rest of the code is great merge material!

PR #4503, opened 2017-02-24
Allow reads after calling BasicSocket#close_write.

### LINKED ISSUE #4500: IO#close_write seems to close the read side too for unix sockets
This works in MRI 2.3.3 and 2.4.0, but fails on jruby-9.1.7.0:

```ruby
require "debug_socket"
DebugSocket.start("boom")
require "socket"
sock = UNIXSocket.new("boom")
sock.write("1")
sock.close_write
sock.read
```

jruby failure:

```
IOError: not opened for reading
from org/jruby/RubyIO.java:2992:in `read'
```

Travis failure (has all the jruby version info + platform info):
https://travis-ci.org/square/debug_socket/jobs/204702060

Test case: https://github.com/square/debug_socket/blob/a94f1c857b2626273eba9331dbc43d3d2ad0cf4e/spec/debug_socket_spec.rb#L29

snowp:
`shutdownInternal` was always shutting down the read side, even when we were trying to close the write side of the socket.

Fixes #4500.

### DIFF
--- a/core/src/main/java/org/jruby/ext/socket/RubyBasicSocket.java
+++ b/core/src/main/java/org/jruby/ext/socket/RubyBasicSocket.java
@@ -456,7 +456,8 @@ public class RubyBasicSocket extends RubyIO {
         }
 
         // shutdown half
-        shutdownInternal(context, fptr, 0);
+        int how = closeHalf == OpenFile.READABLE ? 0 : 1;
+        shutdownInternal(context, fptr, how);
         fptr.setMode(fptr.getMode() & ~closeHalf);
 
         return context.nil;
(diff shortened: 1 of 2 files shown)


snowp:
looking at it again im thinking the `fptr.setMode(fptr.getMode() & ~closeHalf);` in closeHalf might be redundant (since `shutdownInternal` nulls out the flag already). ill try it out in the morning

headius:
Ahh I had not seen you made a PR. Will review.

headius:
You're right, that line is redundant if the contract of shutdownInternal is to also update read/write flags. Seems like that's appropriate, since I can't imagine why we'd want it NOT to do so.

I'll merge and remove the redundant line.

nerdrew:
@snowp THANKS!

PR #4504, opened 2017-02-24
Improve some of the debug logging in ConstantLookupSite.

snowp:
Added this while debugging an issue, figured it'd be useful
upstream. Adds the method name to the log statement + adds
logging for searchModuleForConst. Spelling out the method
name is useful to distinguish between lexicalSearchConst
and searchConst, but also makes it easier to map the log
statements with the origin method when you're not familiar
with the code.

Happy to change the format of the logs if there's any objections, 
this is just what I ended up using.

### DIFF
--- a/core/src/main/java/org/jruby/ir/targets/ConstantLookupSite.java
+++ b/core/src/main/java/org/jruby/ir/targets/ConstantLookupSite.java
@@ -97,7 +97,7 @@ public class ConstantLookupSite extends MutableCallSite {
         setTarget(switchPoint.guardWithTest(target, fallback));
 
         if (Options.INVOKEDYNAMIC_LOG_CONSTANTS.load()) {
-            LOG.info(name + "\tretrieved and cached from scope " + staticScope.getIRScope());
+            LOG.info(name + "\tretrieved and cached from scope (searchConst) " + staticScope.getIRScope());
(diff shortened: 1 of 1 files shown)


headius:
Looks great! Richer logging is always welcome.

headius:
Incidentally, was the problem you were debugging something wrong in JRuby?

snowp:
It's an issue that's causing increased GC usage, seemingly introduced after some changes made to the ConstantLookupSite for 9.1.3.0. I'm hoping to file an issue about it by EOW

snowp:
@headius: Was able to track down the problem, so not gonna create an issue, but I'll outline it here: after updating from 9.1.2.0 to 9.1.7.0 we saw a large increase in time spent in ParNew (would hit 30% during peak hours after a few days, depending on the amount of traffic). We tracked down the issue to being introduced either in 1ae29af235bc37c0c210cedd2cfb6de33208e593 or d15e35acee9e8ca81e17b3712a83da3d3b7b71eb. Turns out your change to make the idTest on RubyModule lazy fixed the issue, so we're just going to patch that change in to .7. Would love to find out why creating redundant MethodHandles  was causing increased GC pressure (memory leak?), but its not crucial at this point.

PR #4507, opened 2017-02-24
Actually cache the idTest on the module.

snowp:
This was assigning to the local variable, so the `MethodHandle` would be created ever time this is called.


headius:
D'oh. Wanted to avoid eagerly creating them and went too far. Thanks!

Issue #4510, opened 2017-02-25
Fiddle::Function returns FFI::Pointer instead of Fiddle::Pointer


jellymann:
### Environment

* JRuby version: 9.1.8.0-SNAPSHOT
* Operating System: Mac OS X 10.11.6

### Expected Behavior

code:
```
require 'fiddle'
libc = Fiddle.dlopen('/usr/lib/libc.dylib')
f = Fiddle::Function.new(libc['strcpy'], [Fiddle::TYPE_VOIDP, Fiddle::TYPE_VOIDP], Fiddle::TYPE_VOIDP)
buff = "000"
str = f.call(buff, "123")
```

MRI:
```
str.to_s
#=> "123"
```

### Actual Behavior

JRuby:
```
str.to_s
#=> "#<FFI::Pointer address=0x7000003137a0>"
```

headius:
Ok, I merged your PR, and thanks again for that.

This issue is a bit trickier, since it's trying to let FFI classes masquerade as Fiddle classes. The `call` for the Fiddle::Function gets bound directly to the `call` of the FFI::Function, so unsurprisingly it returns an FFI::Pointer. I think we need to rework this to wrap the FFI::Function rather than binding it directly to `call`, so we can produce an appropriate Fiddle type on the way out.

headius:
No time to work on this for 9.1.8.0 but with your help we'd like to fix it for 9.1.9.0 or 9.2 (whichever comes first).

jellymann:
@headius I've made a PR to fix this into your Fiddle Fixes branch here: headius/jruby#2

headius:
The PR in question was merged, so hopefully this was fixed around the same time.

jellymann:
@headius I believe it was merged into your fork of JRuby, not sure if the changes made it into this repo?

jellymann:
Okay it looks like this commit fixed the issue: https://github.com/jruby/jruby/commit/366f9f1a7efd57e2e624030dd448231b0a6f6650
Not mine but it does the same thing 👍

Issue #3477, opened 2015-11-20
IllegalArgumentException when using Fiddle to 


npomf:
Using the `fiddle` stdlib library from Ruby 2.2.2, the following error is raised:

`Java::JavaLang::IllegalArgumentException: Invalid parameter type: VOID`

This is presumably during the process of mapping a foreign function's signature against the loaded library (in this case, `OpenGL32.dll`).  The function signatures have been written to as closely match the OpenGL 4.5 spec as possible.  An example is provided below:

```ruby
extern 'void glTexImage2D( GLenum, GLint, GLint, GLsizei, GLsizei, GLint, GLenum, GLenum, void* )'
```

This function returns `void` (no value) and takes a number of parameters whose types are aliases of normal C types.  These aliases are defined before any function is loaded via the overwritten `extern()` method of the `Fiddle::Importer` implementation.

Though not listed in the stack trace in the next section, my best estimate is that the error occurs in or around the marked section of code:

```ruby
def self.extern( signature, *opts )

    # **** LIKELY SUSPECT A ****
    sym, returns, args = parse_signature( signature, @type_alias )
    name = sym.gsub( /@.+/, '')
    opt = parse_bind_options( opts )

    # **** LIKELY SUSPECT B ****
    fptr = import_function( sym, returns, args, opt[ :call_type ]) rescue procaddr( name, returns, args )

    # get invocation location
    file, line = caller.first.scan( /^(.+?):(\d+)/ ).map {|(f, l)| [f, l ? l.to_i - 2 : nil] }.flatten

    # register function
    GL_FUNCTIONS_MAP[ name.to_sym ] = fptr
    module_eval( %Q{
        def #{name}( *args, &block )
            GL_FUNCTIONS_MAP[ :#{name} ].call( *args, &block )
        end
    }, file || __FILE__, line || __LINE__ + 3 )
    module_function( name )

    # return pointer
    fptr
end

# Define the ProcAddress loader for Windows and Linux systems, and load the remote procedure into a new `Fiddle::Function`.
def self.procaddr( name, returns, args )
    if @@procaddr.nil?
        func = case @@platform
            when Platform::Windows then 'wglGetProcAddress'
            when Platform::Linux then 'glXGetProcAddress'
            when Platform::Macintosh then raise NotImplementedError, "#{name}() is not implemented on your system"
        end

        # **** LIKELY SUSPECT C ****
[...]

PR #4511, opened 2017-02-26
Fix some issues with Fiddle

### LINKED ISSUE #3462: Uninitialized constant in Fiddle::Pointer.malloc
When allocating memory with Fiddle on JRuby 9.0.3.0, a `NameError` exception is raised because the LibC module is not defined.

```
jruby-9.0.3.0 :001 > require 'fiddle'
 => true 
jruby-9.0.3.0 :002 > Fiddle::Pointer.malloc 16
NameError: uninitialized constant Fiddle::Pointer::LibC
    from org/jruby/RubyModule.java:3181:in `const_missing'
    from /home/user/.rvm/rubies/jruby-9.0.3.0/lib/ruby/stdlib/fiddle/jruby.rb:168:in `malloc'
    from (irb):2:in `<eval>'
    from org/jruby/RubyKernel.java:978:in `eval'
    from org/jruby/RubyKernel.java:1291:in `loop'
    from org/jruby/RubyKernel.java:1098:in `catch'
    from org/jruby/RubyKernel.java:1098:in `catch'
    from /home/user/.rvm/rubies/jruby-9.0.3.0/bin/irb:13:in `<top>'
```

The related code in `stdlib/fiddle/jruby.rb`:

``` ruby
    def self.malloc(size, free = nil)
      self.new(LibC.malloc(size), size, free ? free : LibC::FREE)
    end
```

jellymann:
After having some trouble getting Mittsu working in JRuby, I thought I'd have a look at JRuby's Fiddle implementation. There are quite a number of inconsistencies and straight-up broken bits, so I thought I'd take a crack at some of them. This is not a conclusive fix or anything, but hopefully I can get it working enough for most.

Here are the list of the bigger changes:
1. `TYPE_PTRDIFF_T` was not defined. This has been voiced before (#3477). I had a look, and other types such as `TYPE_SIZE_T`, etc. were also missing. I've added all the missing types, basing their byte sizes and alignment on `TYPE_VOIDP`
2. Make `Fiddle::Pointer` behave more like MRI
  a. `#[]` now gets the correct number of bytes, regardless of null-terminators
  b. Added `Pointer#[]=`
  c. `#to_s` and `#to_str` should behave differently when `len` is omitted. Notably, `#to_s` takes in to account null-terminators, while `to_str` uses the Pointer size
  d. Throw `DLError` instead of FFI errors in some cases
  e. `Pointer.new` accepts anything that can be converted to an Integer as an address, using `Integer()`
  f. Added `Pointer#free` and `Pointer#free=`. `@free` is now stored as an integer.
  g. `Pointer.malloc` does not assign a default `free` function to the resulting pointer
  h. Added `Ponter#==` and `Pointer#<=>`
3. Added some missing items in the `Fiddle` module, notably
  a. `Fiddle.malloc` and `Fiddle.free`, which map directly to their `libc` counterparts
  b. `Fiddle::RUBY_FREE`, which is just the Integer memory address of the `libc` `free` function
  c. `Fiddle::BUILD_RUBY_PLATFORM`, which I've just set to `RUBY_PLATFORM`, which is always `"java"`
  d. `Fiddle::NULL`, a null pointer
  e. `Fiddle::LibC`, an FFI wrapper to `libc`, not exactly required, but it serves to facilitate the implementation of `malloc` and `free` (also resolves #3462)
4. Added a check to `fiddle/helper` in the tests, where the location of `libc` dynamic library could not be found because platform is always `"java"`. Changed it so that it uses `RbConfig` to detect the _actual_ OS.

There are a great deal of things I have not been able to address thus far, most notably:
1. `Pointer.to_ptr(string)` makes a copy of the given string instead of wrapping the string itself. This is evident when you attempt to modify the data through the pointer, only to find the string was not modified. This issue affects the case of shovelling into the string, where in JRuby the Pointer data is then not reflected. I believe in order to rectify this we would have to get the address of the underlying `ByteList` of the string, but i worry that this will be infeasible in Java, especially with the JVM not guaranteeing the memory address of all it's object all the time.
2. `Pointer.to_ptr(file)` does not work. Right not it will use `File#to_i` to get the address, which is obviously incorrect. MRI converts the File object to a `libc` FILE, and I am unsure as to how that could be achieved in JRuby.
3. `Fiddle::Function` returns an `FFI::Pointer` instead of `Fiddle::Pointer`. I've made an issue here (#4510)
4. `Fiddle::Function` does not accept certain values where MRI's fiddle does. For example, it does not appear to accept integers in place of pointers, or even subclasses of `Fiddle::Closure` for that matter. This, and point 3, seem to be that `Fiddle::Function` is just a straight-up wrapper of `FFI::Function`, so it only accepts FFI arguments and returns FFI values.
5. Related to point 1, and probably an easy fix, is that `Pointer.to_ptr(string)` returns a Pointer 1 byte bigger than on MRI. I think that probably has to do with null-terminators or something, although the behaviour doesn't seem to be affected if I "fix" the issue. I reverted the fix since I wanted some more input on the matter.

My personal observation is that it seems JRuby's Fiddle implementation was hastily put together on top of FFI once Ruby adopted Fiddle as the defacto foreign-function interface. MRI's Fiddle is written directly on top of libffi, and in C. Perhaps it is time that JRuby's Fiddle was build directly on top of libffi/jffi, and in Java instead of Ruby.

### DIFF
--- a/lib/ruby/stdlib/ffi/libc.rb
+++ b/lib/ruby/stdlib/ffi/libc.rb
@@ -1,0 +1,216 @@
+module FFI
+  module LibC
+    extend FFI::Library
+
+    typedef :pointer, :FILE
+    typedef :uint32, :in_addr_t
+    typedef :uint16, :in_port_t
+
+    # time.h
+    typedef :long, :clock_t
+
+    ffi_lib [FFI::CURRENT_PROCESS, 'c']
+
+    # The NULL constant
+    NULL = nil
+
+    # errno.h
+    attach_variable :sys_errlist, :pointer
+    attach_variable :sys_nerr, :int
+    attach_variable :errno, :int
+
+    def self.raise_error(error=self.errno)
+      raise(strerror(error))
+    end
+
+    # unistd.h
+    attach_function :brk, [:pointer], :int
+    attach_function :sbrk, [:pointer], :pointer
+    attach_function :getpid, [], :pid_t
+    attach_function :getppid, [], :pid_t
+    attach_function :getuid, [], :uid_t
+    attach_function :geteuid, [], :uid_t
+    attach_function :getgid, [], :gid_t
+    attach_function :getegid, [], :gid_t
+    attach_function :alarm, [:uint], :uint
+
+    # stdlib.h
+    attach_function :calloc, [:size_t, :size_t], :pointer
+    attach_function :malloc, [:size_t], :pointer
+    FREE = attach_function :free, [:pointer], :void
+    attach_function :realloc, [:pointer, :size_t], :pointer
+    attach_function :getenv, [:string], :string
+    attach_function :putenv, [:string], :int
+    attach_function :unsetenv, [:string], :int
+
+    begin
+      attach_function :clearenv, [], :int
+    rescue FFI::NotFoundError
+      # clearenv is not available on OSX
+    end
+
+    # time.h
+    attach_function :clock, [], :clock_t
+    attach_function :time, [:pointer], :time_t
+
+    # sys/time.h
+    attach_function :gettimeofday, [:pointer, :pointer], :int
+    attach_function :settimeofday, [:pointer, :pointer], :int
+
+    # sys/mman.h
+    attach_function :mmap, [:pointer, :size_t, :int, :int, :int, :off_t], :pointer
+    attach_function :munmap, [:pointer, :size_t], :int
+
+    # string.h
+    attach_function :bzero, [:pointer, :size_t], :void
+    attach_function :memset, [:pointer, :int, :size_t], :pointer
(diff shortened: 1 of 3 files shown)


headius:
Thanks for this! Our Fiddle implementation has not had a lot of love since the initial implementation, so any help you can provide here is quite welcome!

We've decided to up the priority on this and get it into 9.1.8.0 rather than waiting for 9.2. If you can help fix the remaining issues after 9.1.8.0 that would be a great help also. I'll take a quick look at them today.

headius:
Ok, so these looked nice on Linux but it appears to have broken something on Windows. I've created #4518 containing your changes and my commit to enable Fiddle tests. We can iterate together on getting things working right based on https://github.com/headius/jruby/tree/fiddle-fixes.

PR #4521, opened 2017-03-03
Use actual impl module to test bindability. Fixes #4477.

### LINKED ISSUE #4477: prepending to the singleton class of a module breaks instance_method
### Environment

jruby 9.1.7.0 (2.3.1) 2017-01-11 68056ae Java HotSpot(TM) 64-Bit Server VM 25.121-b13 on 1.8.0_121-b13 +jit [linux-x86_64]

Ubuntu 16.04 LTS, jruby installed from rbenv with no special flags.

### Expected Behavior

I'm the maintainer of flexmock, and during some of the partial mock operations, flexmock prepends to the singleton class of the object that is being mocked. It also stores the original instance method before mocking it so as to be able to call it as part of the mock process. On both MRI and JRuby, this works fine for plain objects. However, when the object is itself a module, it works on MRI but fails on JRuby

The problem can be reproduced with: 

```
module Base
    def setup_submodel; end
end
module P; end

module M
    extend Base
    class << self
        prepend P
    end
end

m = M.singleton_class.instance_method(:setup_submodel)
m.bind(M).call
```

MRI executes it without any problem, while JRuby fails with 

```
TypeError: bind argument must be an instance of #<Class:0x8646db9>
    bind at org/jruby/RubyUnboundMethod.java:105
  <main> at jruby_prepend_bind_bug.rb:17
```

headius:
This fixes #4477 by not using the origin module/class to test bindable hierarchy but instead using the actual implementation module. I'm not sure why it was doing the latter before, but it meant that any methods captured from a module were not bindable to class hierarchies that included that module (since the module was hidden behind a Prepended or IncludedModule origin).

This seems like the proper logic, but I'm doing it as a PR to check CI before pushing forward for 9.1.8.0.

There may be tests for this we don't pass but I could not find them.

### DIFF
--- a/core/src/main/java/org/jruby/RubyUnboundMethod.java
+++ b/core/src/main/java/org/jruby/RubyUnboundMethod.java
@@ -102,7 +102,7 @@ public class RubyUnboundMethod extends AbstractRubyMethod {
     public RubyMethod bind(ThreadContext context, IRubyObject aReceiver) {
         RubyClass receiverClass = aReceiver.getMetaClass();
         
-        receiverClass.checkValidBindTargetFrom(context, originModule);
+        receiverClass.checkValidBindTargetFrom(context, implementationModule);
         
         return RubyMethod.newMethod(implementationModule, methodName, receiverClass, originName, method, aReceiver);
     }


headius:
Everything passed except for some newer flaky specs for Weakref. Merging.

PR #4534, opened 2017-03-16
Don't yieldSpecific to allow arrays through normal block params.

### LINKED ISSUE #4532: Enumerable#each_entry behaves differently from MRI
`Enumerable#each_entry` is supposed to convert multiple arguments to an array, but the behaviour is different from MRI and jruby 1.7 when multiple arguments are yielded.

`jruby -v`:
  jruby 9.0.5.0 (2.2.3) 2016-01-26 7bee00d Java HotSpot(TM) 64-Bit Server VM 24.76-b04 on 1.7.0_76-b13 +jit [linux-amd64]
  jruby 9.1.5.0 (2.3.1) 2016-09-07 036ce39 Java HotSpot(TM) 64-Bit Server VM 24.76-b04 on 1.7.0_76-b13 +jit [linux-x86_64]

test file:
```rb
class Foo
  include Enumerable

  def each
    yield 1
    yield 1, 2
    yield
  end
end

out = []
Foo.new.each_entry{ |*args| out << args }
puts "each_entry: #{out.inspect}"
```

Expected output:
```
==[1.9.3-p551]==
each_entry: [[1], [[1, 2]], [nil]]

==[2.0.0-p648]==
each_entry: [[1], [[1, 2]], [nil]]

==[2.1.10]==
each_entry: [[1], [[1, 2]], [nil]]

==[2.2.6]==
each_entry: [[1], [[1, 2]], [nil]]

==[2.3.3]==
each_entry: [[1], [[1, 2]], [nil]]

==[2.4.0]==
each_entry: [[1], [[1, 2]], [nil]]

==[jruby-1.7.26]==
each_entry: [[1], [[1, 2]], [nil]]
```

jruby 9.x output:
```
==[jruby-9.0.5.0]==
each_entry: [[1], [1, 2], [nil]]

==[jruby-9.1.7.0]==
each_entry: [[1], [1, 2], [nil]]
```

headius:
Fixes #4532.

### DIFF
--- a/core/src/main/java/org/jruby/RubyEnumerable.java
+++ b/core/src/main/java/org/jruby/RubyEnumerable.java
@@ -1085,7 +1085,7 @@ public class RubyEnumerable {
     public static IRubyObject each_entryCommon(ThreadContext context, final IRubyObject self, final IRubyObject[] args, final Block block) {
         callEach(context.runtime, context, self, args, Signature.OPTIONAL, new BlockCallback() {
             public IRubyObject call(ThreadContext ctx, IRubyObject[] largs, Block blk) {
-                return block.yieldSpecific(ctx, packEnumValues(ctx, largs));
+                return block.yield(ctx, packEnumValues(ctx, largs));
             }
         });
         return self;


enebo:
@headius wow that is literally the only yieldSpecific we have left in RubyEnumerable

toy:
@headius I think some test of behaviour would be great to be sure that tweaking logic doesn't break it.

Issue #3611, opened 2016-01-21
JDK9 + 1.7.x produces io/console warnings


headius:
Not sure why these warnings are coming up.

```
$ jruby -v -e 1
jruby 1.7.25-SNAPSHOT (1.9.3p551) 2016-01-21 592384a on Java HotSpot(TM) 64-Bit Server VM 9-ea+100-2016-01-06-200131.javare.4235.nc +jit [darwin-x86_64]
failed to load native console support: Unsupported platform: x86_64-darwin
io/console not supported; tty will not be manipulated
...
```

We need to make fixes to 1.7 to support JDK9's new platform/version strings, I assume. We already made these changes on master.

perlun:
JRuby 1.7 has reached EOL (#4112). Can we close this?

mprins:
this may be a duplicate of #3700

enebo:
Yeah we will not support Java 9 for 1.7.x and this issue was fixed on 9.x before this issue was even opened.  So I will close this.

PR #4553, opened 2017-03-29
support new JDK version string scheme

iignatev:
jruby should be updated to support new JDK version string scheme -- [JEP 223](http://openjdk.java.net/jeps/223).

before this JEP, java.version property was 1.X.Y_Z, where X is major version number of JDK, now it is '$VNUM(-$PRE)?(\+($BUILD)?(-$OPT)?)?' where VNUM is $MAJOR.$MINOR.$SECURITY.


iignatev:
#3611 should be fixed by this patch.

PR #4563, opened 2017-04-19
(peazy) enumerator backports for 1.7

### LINKED ISSUE #1571: Enumerator#feed

kares:
~ cherry-picked from master, so that feed and some others are available

resolves #1571

### DIFF
--- a/core/src/main/java/org/jruby/RubyArray.java
+++ b/core/src/main/java/org/jruby/RubyArray.java
@@ -1650,7 +1650,7 @@ public class RubyArray extends RubyObject implements List, RandomAccess {
         final int size = RubyNumeric.num2int(arg);
         final Ruby runtime = context.runtime;
         if (size <= 0) throw runtime.newArgumentError("invalid slice size");
-        return block.isGiven() ? eachSlice(context, size, block) : enumeratorize(context.runtime, this, "each_slice", arg);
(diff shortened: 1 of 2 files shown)


enebo:
@kares thanks for backporting/picking back to 1.7.  My only concern with this work is adding 2.x semantics but I can see (I only checked seek) was added in 1.9.x.   So long as compat is cool bombs away!

kares:
@enebo thanks, I did check the docs. `feed`, `peek_values` and `next_values` are there in **1.9.3**

enebo:
@kares yeah I assumed you checked to see if the methods existed in 1.9.  I was partially mentioning semantics of the methods themselves but I stated that perhaps in too general of a way.

headius:
:+1: Looks good.

kares:
ok, semantics isn't 100% MRI compatible like on 9K and since this is simply a back-port it will do the same.
`Enumerator#feed` acts different but it has been decided to be fine for 9K and it hasn't been changing in Ruby 2.x so that is why I thought its better than nothing esp. since there's an old bug report asking for it for 1.7

enebo:
@kares yeah I was just asking for due dilligence on whether there was some 2.0ish behavior making back to 1.9.  So as long as you thought about it I definitely trust your judgement.

kares:
@enebo great thanks. I am grateful as I am never 100% sure either on these. I'll squash and merge than.

PR #4566, opened 2017-04-19
Return Process::Waiter with pid in Process.detach

davishmcclurg:
This adds a `pid` method to threads returned by `Process.detach`,
as described in the open3.rb documentation. The thread is a
`Process::Waiter` to be consistent with standard Ruby.

I haven't worked in Java much before, so please let me know if there's
anything I'm missing.

Before:

```
>> JRUBY_VERSION
=> "9.1.4.0"
>> pid = Process.spawn('sleep 10')
=> 58487
>> thr = Process.detach(_)
=> #<Thread:0x45752059@(irb):3 run>
>> thr.pid
NoMethodError: undefined method `pid' for #<Thread:0x45752059@(irb):3 dead>
```

After:

```
>> JRUBY_VERSION
=> "9.1.9.0-SNAPSHOT"
>> pid = Process.spawn('sleep 10')
=> 59018
>> thr = Process.detach(_)
=> #<Process::Waiter:0x5acf93bb@(irb):3 run>
>> thr.pid
=> 59018
```

### DIFF
--- a/core/src/main/java/org/jruby/RubyProcess.java
+++ b/core/src/main/java/org/jruby/RubyProcess.java
@@ -704,7 +704,7 @@ public class RubyProcess {
     // MRI: rlimit_resource_name2int
     private static int rlimitResourceName2int(String name, int casetype) {
         RLIMIT resource;
-            
+
         OUTER: while (true) {
             switch (Character.toUpperCase(name.charAt(0))) {
                 case 'A':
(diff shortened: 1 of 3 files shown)


headius:
This appears to be the same way MRI implements `Waiter#pid` and the PR looks fine. Merging.

PR #4570, opened 2017-04-21
[jrubyc] respect Ruby visibility rules (do not generate private methods)

### LINKED ISSUE #4469: `private` and `private :foo` have different semantics when compiling to Java
### Environment

<dl>
<dt><kbd>jruby -v</kbd></dt>
<dd><pre>jruby 9.1.7.0 (2.3.1) 2017-01-11 68056ae Java HotSpot(TM) 64-Bit Server VM 25.112-b16 on 1.8.0_112-b16 +jit [darwin-x86_64]</pre></dd>
<dt><kbd>uname -a</kbd></dt>
<dd><pre>Darwin zoroark 16.3.0 Darwin Kernel Version 16.3.0: Thu Nov 17 20:23:58 PST 2016; root:xnu-3789.31.2~1/RELEASE_X86_64 x86_64</pre></dd>
</dl>

### Expected Behavior

```ruby
class Foo
  private
  def foo?
    true
  end
end
```

should be compiled with `jrubyc --java` to the same Java as

```ruby
class Foo
  def foo?
    true
  end
  private :foo?
end
```

(The `foo?` method should not be generated in Java.)

### Actual Behavior

JRuby tries to compile the first example's `foo?` method to Java, which fails because -`?` methods aren't supported in Java.

kares:
`jrubyc --java` generation updated to be a little smarter - trying to respect Ruby's visibility rules.

... `private, protected, public` as well as their `private :method, ...` counterparts now seamlessly 'work'

private Ruby methods (without an explicit `java_signature '...'` which is allowed to force generation) are no longer generated. also protected methods map to Java protected visibility by default.

(kind of) resolves #4469   (still undecided about what to do with with `xxx?` and `yyy!` methods)

### DIFF
--- a/core/src/main/java/org/jruby/RubyClass.java
+++ b/core/src/main/java/org/jruby/RubyClass.java
@@ -197,8 +197,9 @@ public class RubyClass extends RubyModule {
      * constructor (Ruby, RubyClass) on the given class via a static
      * __allocate__ method intermediate.
      *
-     * @param clazz The class from which to grab a standard Ruby __allocate__
-     *            method.
+     * @param clazz The class from which to grab a standard Ruby __allocate__ method.
+     *
+     * @note Used with `jrubyc --java` generated (interoperability) class files.
      */
     public void setRubyStaticAllocator(final Class<?> clazz) {
         try {
--- a/lib/ruby/stdlib/jruby/compiler.rb
+++ b/lib/ruby/stdlib/jruby/compiler.rb
@@ -56,10 +56,6 @@ module JRuby::Compiler
         options[:javac_options] << o
       end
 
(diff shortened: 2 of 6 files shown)


kares:
@headius maybe this is good to go for 9.1.9 - to move some of the `jrubyc --java` users forward (#4469) ?

headius:
The question is whether the old behavior could be considered a feature by some, a feature that we are taking away. I don't feel that way though, so I am on board with including it.

We may want to consider moving jrubyc to a preinstalled gem for 9.2, so it can be updated independently.

kares:
OK ... I'm fine with leaving it for 9.2 as well - let's have @enebo to :8ball: for us

PR #4571, opened 2017-04-21
[ruby 2.4] Implemented Integer#ceil, #floor, and #truncate

whwilder:
Feature #12245 ( ref #4293 )
There was already an implementation of Integer#round.  I noticed it had an implementation in Integer that covered both Fixnum and Bignum.  Would this be a preferable way of doing things, or is my current method of splitting the implementation between Fixnum and Bignum acceptable?

Edit: Hmm, didn't realize it would include all of my previous commits.  I'll close this PR and see if I can fix this.


perlun:
@whwilder did you manage to recreate this including only the relevant parts?

whwilder:
I have.  I just tested opening a PR and it looks like everything is in order.  I'll go ahead with the new PR unless there's some way to fix this one.

PR #4576, opened 2017-04-27
Hash symbols created from ByteList by their encoding. Fixes #1348.

### LINKED ISSUE #1348: Cannot make two symbols with same bytes and different encodings
I noticed that the symbol table in JRuby cannot seem to store two different symbols if the symbols happen to have the same bytes, even if the encodings are different.

Here is a script demonstrating the problem:

``` ruby
sym1 = "ab".force_encoding("UTF-16").to_sym
sym2 = "ab".to_sym
puts sym2.encoding

sym3 = "cd".to_sym
sym4 = "cd".force_encoding("UTF-16").to_sym
puts sym4.encoding
```

Here is the output from my shell demonstrating how MRI gets the encodings for sym2 and sym4 right and JRuby gets them wrong because of the pre-existing symbol in the symbol table:

```
$ ruby -v && ruby test_symbol_table.rb
ruby 2.0.0p0 (2013-02-24) [x64-mingw32]
US-ASCII
UTF-16
$ source use_jruby_179.sh
$ jruby -v && jruby test_symbol_table.rb
jruby 1.7.9 (1.9.3p392) 2013-12-06 87b108a on Java HotSpot(TM) 64-Bit Server VM
1.7.0_45-b18 [Windows 8-amd64]
UTF-16
US-ASCII
```

I think I will work on making a pull request to fix this; advice and objections are welcome.

I have made some progress on issue #1329 (properly setting the encoding of unmarshaled symbols), but before I can truly succeed in fixing that, I think I need to fix this and maybe a few other fundamental things about symbols.

headius:
This change appears to work and so far it's passing tests.

After digging into MRI again, I realized a couple things:

* When looking up a symbol from C, using a char*, MRI has the same issue we do: they can't know what encoding you intended, so they always use US-ASCII. This is similar to our use of java.lang.String as the key in all structural tables (methods, constants, etc). Note however that MRI's structural tables are keyed off an ID, which is a symbol index that can vary depending on encoding for the same bytes.

* The hashing logic to look up a Symbol from a String considers encoding, but only for strings that either have a non-ASCII-compatible encoding or that are not all 7-bit ASCII. Two strings of the same bytes but different ASCII-compatible encodings will not produce two symbols. This reduces the impact on our structural tables since almost all Ruby code uses ASCII-compatible encodings (e.g. not UTF-16), and still strings that are multibyte will almost always have differing bytes. See below:

```
 $ rvm ruby-2.4.0 do ruby -e 'sym1 = "cd".intern; sym2 = "cd".force_encoding("UTF-8").intern; p sym1.object_id, sym2.object_id'
14474580
14474580
```

* Although we have a path to create symbols from java.lang.String, we also have a path to create them from ByteList. This second path is used primarily when turning a Ruby String into a Symbol. My patch fixes this path to consider encoding, which fixes the original issue.

There are some risks with this change.

* The "raw" strings we use in structural tables are not currently always "raw". We are working to fix this now (@enebo's work on properly-encoded identifiers, for example).
* We are not sure how good the coverage is for identifiers in mixed-encoding projects.

My gut feeling is that this *can* be made safe and mostly proper for a few reasons:

* The case of having non-ASCII-compatible identifiers is very, very rare.
* This does fix string interning to produce proper encoded symbols.
* For all multibyte strings there's nearly zero chance the bytes will collide.

We'll see how this does on the rest of the test suite and evaluate how to proceed.


headius:
@enebo That last commit should make this set of changes green, but I wanted to point out that Struct is another place using String/RubyString/RubySymbol/ByteList incorrectly for identifiers. I started to fix it but realized many of the methods I would have to add probably already exist in your branch (e.g. isConstant, defineConstantAt, etc).

headius:
@enebo This branch is indeed green now. But I'm on the fence about including it in 9.1.9.0, even though it is limited to encoding of literal symbols and does not touch (most) identifier logic.

headius:
I will push this and @enebo's bytelist_through_parser branch after testing.

headius:
I have updated this for current master.

kares:
nice work - esp. since all those marshal specs now pass. 
... there seems to be some CI failures after the merge.

core/src/main/java/org/jruby/RubyStruct.java
```diff
@@ -201,7 +201,16 @@ public static RubyClass newInstance(IRubyObject recv, IRubyObject[] args, Block
 
         for (int i = (name == null && !nilName) ? 0 : 1; i < args.length; i++) {
             if (i == args.length - 1 && args[i] instanceof RubyHash) break;
-            member.append(runtime.newSymbol(args[i].asJavaString()));
+
+            IRubyObject arg = args[i];
```
enebo: This is an improvement over what is there but if we can accept any type at this point then RubyString should be checked so we can retrieve bytelist vs javastring and we will maintain encoding in mbc.  I half thought about making an interface for RubySymbol/RubyString like ByteListContent so we don't special check these both as separate branches (although they don't have same method name either).

Unrelated to this PR but I saw this code before and wondered about whether it is possible when we do asJavaString we convert a fixnum to "1" or stuff like that (and don't raise some error).

Issue #3928, opened 2016-05-26
File descriptors from stdlib jars are leaked when ScriptingContainers are terminated


cprice404:
### Environment

JRuby 9.1.1.0, run from a Java program via `ScriptingContainer`, with `setCompileMode` set to `OFF`.
### Expected Behavior

`ScriptingContainer` instances should be able to be created and terminated without leaking memory or other resources, such as file descriptors.
### Actual Behavior

When a `ScriptingContainer` is constructed and executes any Ruby code that results in certain jar-based libraries being loaded from `jruby-stdlib`, there are file descriptors opened pointing to the relevant library jars.  When a `ScriptingContainer` is terminated, these descriptors (which seem to be referenced in a global cache) are not cleaned up.

The cache appears to use weak references, so presumably at some point the memory used by these objects could be reclaimed, but because there is also an open file descriptor associated with each cache entry, it is likely that a program that is recycling `ScriptingContainer` instances will exceed the open file descriptor limit and crash before there is enough memory pressure to cause these objects to be GC'd.

Here is a reproducer that illustrates the issue:

https://github.com/cprice404/jruby9k-benchmarks/tree/651c7ee063d0bf652293fe5c48b5ced3b80e6ade/jar-file-desc-leak

In the reproducer, we just do `require`s on a few libraries like `openssl`, `psych`, `json/ext`.  Here is some `lsof` output that shows file descriptors for jars that definitely seem related to those libraries (bouncycastle, etc.):

```
java    16802 cprice  145r   REG                8,1   673715  1848479 /tmp/jruby-16802/jruby4231261797459246050bcpkix-jdk15on-1.54.jar
java    16802 cprice  146r   REG                8,1  3277268  1848480 /tmp/jruby-16802/jruby2526682390458341590bcprov-jdk15on-1.54.jar
java    16802 cprice  147r   REG                8,1  1029206  1848481 /tmp/jruby-16802/jruby8134376450898266715jopenssl.jar
java    16802 cprice  148r   REG                8,1    29184  1848482 /tmp/jruby-16802/jruby2150106223798361691parser.jar
java    16802 cprice  149r   REG                8,1    36598  1848483 /tmp/jruby-16802/jruby8382658838024689723generator.jar
java    16802 cprice  150r   REG                8,1   673715  1848486 /tmp/jruby-16802/jruby791041166383033485bcpkix-jdk15on-1.54.jar
[...]

PR #4578, opened 2017-04-28
Use shutdown hook to delete temp files instead of File.deleteOnExit

camlow325:
Previously, the JRubyClassLoader would make a .deleteOnExit() call for
the temp directory and each temp file it needed to create.  This results
in the string for each registered path being stored in a hash that
increases each time a new URL is loaded or new ScriptingContainer is
created.

This commit removes the .deleteOnExit() calls and instead registers a
shutdown hook just after the temp directory is initially created.  At
shutdown time, the hook would delete any remaining files and the temp
directory itself.  This should effectively result in the same user
behavior as before but with the benefit of the temp path strings no
longer taking up an increasing amount of memory the longer the process
runs.

### DIFF
--- a/core/src/main/java/org/jruby/util/JRubyClassLoader.java
+++ b/core/src/main/java/org/jruby/util/JRubyClassLoader.java
@@ -117,9 +117,17 @@ public class JRubyClassLoader extends ClassDefiningJRubyClassLoader {
             Runtime.getRuntime().addShutdownHook(new Thread() {
                 public void run() {
                     for (File f : tempDir.listFiles()) {
-                        f.delete();
+                        try {
+                            f.delete();
+                        } catch (Exception ex) {
+                            LOG.debug(ex);
+                        }
                     }
-                    tempDir.delete();
+                    try {
+                        tempDir.delete();
+                    } catch (Exception ex) {
+                        LOG.info("failed to delete temp dir " + tempDir + " : " + ex);
+                    }
                 }
             });
         }
--- a/core/src/main/java/org/jruby/util/JRubyClassLoader.java
+++ b/core/src/main/java/org/jruby/util/JRubyClassLoader.java
@@ -60,7 +60,7 @@ public class JRubyClassLoader extends ClassDefiningJRubyClassLoader {
 
     private Runnable unloader;
 
-    private File tempDir;
+    private static volatile File tempDir;
 
     public JRubyClassLoader(ClassLoader parent) {
         super(parent);
@@ -76,7 +76,6 @@ public class JRubyClassLoader extends ClassDefiningJRubyClassLoader {
             InputStream in = null; OutputStream out = null;
             try {
                 File f = File.createTempFile("jruby", new File(url.getFile()).getName(), getTempDir());
-                f.deleteOnExit();
                 out = new BufferedOutputStream( new FileOutputStream( f ) );
                 in = new BufferedInputStream( url.openStream() );
                 int i = in.read();
@@ -110,15 +109,21 @@ public class JRubyClassLoader extends ClassDefiningJRubyClassLoader {
         super.addURL( url );
     }
 
-    private File getTempDir() {
-        File tempDir = this.tempDir;
+    private static synchronized File getTempDir() {
         if (tempDir != null) return tempDir;
 
         tempDir = new File(systemTmpDir(), tempDirName());
         if (tempDir.mkdirs()) {
-            tempDir.deleteOnExit();
+            Runtime.getRuntime().addShutdownHook(new Thread() {
+                public void run() {
+                    for (File f : tempDir.listFiles()) {
+                        f.delete();
+                    }
+                    tempDir.delete();
+                }
+            });
         }
-        return this.tempDir = tempDir;
+        return tempDir;
     }
 
     private static final String TEMP_DIR_PREFIX = "jruby-";


camlow325:
This issue affects our users who need to have the ability to have `ScriptingContainer` instances be recycled over time -- without restarting the Java process for performance.  We do this in order to protect our users from having custom code extensions that they may have written (which may have some memory leaks underneath) cause the process to crash frequently with OOM errors.  Note that this issue is also related to, but does not solve, https://github.com/jruby/jruby/issues/3928.

kares:
looks good - shutdown-hooks do not need try-catch right (on `tempDir.delete()`)? 
in case someone else writes at the location (e.g. creating a dir) it would fail while previously it did not ...

headius:
I'm thinking we should do both, no? I want to ensure that abnormal termination of JRuby still has the JVM deleteOnExit to fall back on.

mkristian:
it seems to be common approach to delete directory and its content in custom shutdown hook and not using deleteOnExit

camlow325:
> shutdown-hooks do not need try-catch right (on tempDir.delete())?

@kares, looking at the Javadocs for [File.delete](https://docs.oracle.com/javase/7/docs/api/java/io/File.html#delete()), it appeared that the only exception which might be thrown is a `SecurityException`.  I had noticed that in the JDK's implementation of its hook behind the [deleteOnExit method](http://hg.openjdk.java.net/jdk8/jdk8/jdk/file/687fd7c7986d/src/share/classes/java/io/DeleteOnExitHook.java#l80) was also using `File.delete` but wasn't trying to catch `SecurityException` or any other exception around the call.

For the case that the process is unable to delete an entry (e.g., for a non-empty directory), I've seen that `File.delete()` just returns false but doesn't throw an exception.  If you think it would be safer to wrap the custom shutdown hook code in a try/catch that swallows any exceptions found, though, I'd be okay with making that change.  What do you think?

camlow325:
> I'm thinking we should do both, no? I want to ensure that abnormal termination of JRuby still has the JVM deleteOnExit to fall back on.

@headius, thanks for taking a look at this PR.  My intent with this PR was to replace the use of the JVM `deleteOnExit` with a custom shutdown hook because, at least for our use case where JRuby `ScriptingContainer`s can be recycled over time, the calls to `deleteOnExit` basically lead to a memory leak.  For example, we have this sequence:

1) `ScriptingContainer` created.
2) 5 or more strings associated with tempDir jars are passed along to a `deleteOnExit` call.
3) The JVM's DeleteOnExitHook stores each of the new strings in an internal hash.
4) Some time later, the `ScriptingContainer` is terminated.
5) Repeat steps 1 -4.

Over time, then, we end up with a new set of strings being added to the hash each time a new container is created - even though the corresponding tempDir jars should no longer be in use after the corresponding `ScriptingContainer` they had been created for has been terminated.

With the custom shutdown hook approach on this PR, I believe we still get the same basic behavior as what the `deleteOnExit` approach had provided but without the side effect of extra strings for the file paths continuing to stack up over time as containers are recycled.

With the changes in the current PR, I've done some manual testing locally with different kinds of process termination - including hitting CTRL^C to kill a foreground process, sending a SIGTERM to kill a background process, and throwing an exception post-container creation in a way which kills the process.  In each of the cases, I saw all of the tempDir jar files and the tempDir itself be cleaned up in conjunction with the process dying.

Are there other cases you can think of where the `deleteOnExit` approach would cover cleaning up the temporary files / directory where the custom shutdown hook would not?

kares:
> For the case that the process is unable to delete an entry (e.g., for a non-empty directory), I've seen that File.delete() just returns false but doesn't throw an exception. If you think it would be safer to wrap the custom shutdown hook code in a try/catch that swallows any exceptions found, though, I'd be okay with making that change. What do you think?

thanks - have read the java-docs seems that exception handling is unnecessary. although I believe there might have been some implementation issues on IBM JREs in the past (maybe I am wrong). still, the current approach is a bit different - `deleteOnExit` having issues does not print anything, while an issue (uncaught exception) with the proposed changes will print a stack-trace to `System.err`. but that's fine with me, if its accepted we can move it forward to introduce some try-catch handling to log using JRuby's internal loggers.

> Are there other cases you can think of where the deleteOnExit approach would cover cleaning up the temporary files / directory where the custom shutdown hook would not?

you could test-out a (unlikely) case where the directory is not deletable - some files fails to get rm-d. its probably easy to simulate if you just create a new file manually in the dir while its running. although its not that important - the behaviour in this case should be similar to what would happen with `deleteOnExit`.

camlow325:
Thanks, @kares.

> if its accepted we can move it forward to introduce some try-catch handling to log using JRuby's internal loggers.

I'm happy to do that if you'd like.  Would doing something simple like what the `addURL` method in the same class does make sense?  Something like...

```java
for (File f : tempDir.listFiles()) {
    try {
        f.delete();
    } catch (Exception ex) {
        LOG.debug(ex);
    }
}
try {
    tempDir.delete();
} catch (Exception ex) {
    LOG.debug(ex);
}
```

What do you think?

> you could test-out a (unlikely) case where the directory is not deletable - some files fails to get rm-d. its probably easy to simulate if you just create a new file manually in the dir while its running. although its not that important - the behaviour in this case should be similar to what would happen with deleteOnExit.

That makes sense.  I have tried doing that.  For example, I tried putting a file under a subdirectory within the temp jruby directory.  Even though the new code would not be able to delete the subdirectory at shutdown in that case, I don't see any errors written to stdout or stderr - even without the extra try/catch logic.  The `File.delete` call for the directory just returns a silent `false` value without throwing any exceptions.

Anything else you can think of that I should look into in order to decide whether or not this could be merged?  Thanks again, all!

camlow325:
Is there any more information needed to decide whether this is a change that the project would be willing to accept?  Thanks again your time reviewing.

kares:
think its good enough - what do you think of increasing the log level but without the stack-trace : 
```java
for (File f : tempDir.listFiles()) {
    try {
        f.delete();
    } catch (Exception ex) {
        LOG.debug(ex);
    }
}
try {
    tempDir.delete();
} catch (Exception ex) {
    LOG.info("failed to delete temp dir " + tempDir + " : " + ex);
}
```
you know there's something wrong - one is likely to get more details (stack-traces) when running with debug mode ... or is that too noisy?

camlow325:
Thanks, @kares.  I think your suggestion makes sense.  I went ahead and added a commit onto the PR with that change, fc28208.

Issue #4004, opened 2016-07-09
Migrate Ruboto to JRuby 9K


donv:
@headius @enebo 
With the announcement of EOL of JRuby 1.7, I hope we can push Ruboto to use JRuby 9000.

I can do the testing and packaging changes needed in Ruboto, but the JRuby changes needed have proven too much for me.

JRuby master is being tested in travis:  https://travis-ci.org/ruboto/ruboto
You should be able to trigger new builds when there are JRuby changes.

headius:
👍 

I'm trying to read between the thousands of lines in that output. I know of some things that need to be changed, but I'm not sure about everything...especially Java 7isms that have crept into things. How far back are we supporting? I know recent Android have support for much of Java 7, yes?

headius:
A bit of brainstorming.

The biggest hit, if we at least require Android Java 7 support, is that invokedynamic/java.lang.invoke are not available. Invokedynamic we can do without; I think there's only a couple features that always use invokedynamic, and they don't have to. The java.lang.invoke package will be harder.

I think it might be possible to implement method handles using reflection. The invocation and lookup are basically taken care of. The adapters are all simple to implement and can just be generic wrappers. It's a bit of work, but it would be a useful library for others. Come to think of it...maybe someone has started this already?

donv:
Lots of Java 7 features are supported, and I think that is done in the Android tooling so we don't have to worry about that for now.

InvokeDynamic byte code is not available.

The main problem are JVM APIs as you have mentioned, and that they are eager loaded even if they will not be used.

donv:
And I can definitely help sifting through the logs to identify problems.  😄

oneiros:
>  The java.lang.invoke package will be harder.

Do you still think this is even possible? I spent some time looking into this and the package is used all over the place. A quick grep finds imports in ~50 files and then there are the imlicit usages, e.g. invokebinder (imported in 12 files).

If the answer is still yes:

> I think it might be possible to implement method handles using reflection. The invocation and lookup are basically taken care of. The adapters are all simple to implement and can just be generic wrappers.

Could you maybe elaborate a bit more on this?

I would not mind spending some more time on this, but my Java knowledge is a bit rusty and I am only starting to find my way around the jruby code base.

Anyway, any additional information is highly appreciated. I am constantly amazed by the jruby project. Keep up the good work.

And by the way: I guess #2471 and #2982 are kind of duplicates of this and might be closed as such.

headius:
[...]

PR #4589, opened 2017-05-05
Return ObjectIdentityInvalidator if indy disabled.

donv:
See #4004 and ruboto/ruboto/issues/737. 

Everything java.lang.invoke related is not
available on android, so for now revert to the old
behavior here.

This still makes some tests fail, but might suffice
to run on android.

Use jruby's ClassValue.

Instead of java.lang.ClassValue which ist not available on
android. See #4004.

Remove seemingly unused constant.

Could not find a usage for this. And since MethodHandles
is not available on android, I simply removed it for the time
being. See #4004.

Revert to old behavior, when indy disabled.

Just like in 8db25932200fb3ed7160066b184e6db2e29f1921

Do not init JITCompiler if it is not used.

See #4004.

Use jruby's ClassValue.

Instead of java.lang.ClassValue which ist not available on
android. See #4004.

Move idTest to where it is actually needed.

This of course destroys any gains of caching the idTest, but
gets rid of MethodHandle et al in RubyModule which is
important to run on android. See #4004.

Remove unused Method.

Since this will not run on android is not being used
currently anyway, just comment it out for now. See #4004.

Revert "relax seek-channel type requirement from FileChannel to SeekableByteChannel"

This reverts commit dc48f340795c7fa194c202fc057f5eb499c76017.

No SeekableByteChannel on android. See #4004.

Work around missing java.nio libs on android.

See #4004.

Replace ClassValue with internal implementation.

Because java.lang.ClassValue is not available on android.
See #4004.

Add minimal MethodHandle implementation.

java.lang.invoke.Method* classes are not available on
android. See #4004.

Add code generation for android.

See #4004.

Move nullToNil out of the way.

To not have MethodHandles in Ruby.java as this
is not available on android. See #4004.

Guard against missing JITCompiler.

Necessry as JITCompiler will not get instantiated on android.
See #4004.

Remove unused methods getIdTest + newIdTest

Move Android logic to the Ruboto project

Removed extra white space

Remove the maven repository for dexmaker


core/src/main/java/org/jruby/RubyFile.java
```diff
@@ -418,24 +418,21 @@ public IRubyObject ctime(ThreadContext context) {
     public IRubyObject birthtime(ThreadContext context) {
         checkClosed(context);
 
-        FileTime btime = getBirthtimeWithNIO(getPath());
-        if (btime != null) return context.runtime.newTime(btime.toMillis()); // btime comes in nanos
-        return ctime(context);
-    }
-
-    public static final FileTime getBirthtimeWithNIO(String pathString) {
```
headius: Mild concern here about removing a public method, but I don't know how long it has been around.

core/src/main/java/org/jruby/RubyModule.java
```diff
@@ -299,18 +300,6 @@ public void addIncludingHierarchy(IncludedModule hierarchy) {
         }
     }
 
-    public MethodHandle getIdTest() {
```
headius: Same here but it was probably wrong of me to expose it in the first place.

core/src/main/java/org/jruby/ir/targets/Bootstrap.java
```diff
@@ -15,12 +15,7 @@
 import org.jruby.ir.JIT;
 import org.jruby.ir.runtime.IRRuntimeHelpers;
 import org.jruby.parser.StaticScope;
-import org.jruby.runtime.Binding;
-import org.jruby.runtime.Block;
-import org.jruby.runtime.CompiledIRBlockBody;
-import org.jruby.runtime.DynamicScope;
-import org.jruby.runtime.Frame;
-import org.jruby.runtime.ThreadContext;
+import org.jruby.runtime.*;
```
headius: We prefer not to do * except in the most extreme cases. This isn't one of them.

core/src/main/java/org/jruby/ir/targets/Bootstrap.java
```diff
@@ -39,6 +34,7 @@
 import org.objectweb.asm.Opcodes;
 
 import java.lang.invoke.*;
+import java.lang.invoke.CallSite;
```
headius: This * is my fault. We should just expand it.

core/src/main/java/org/jruby/ir/targets/Bootstrap.java
```diff
@@ -657,7 +653,10 @@ public static IRubyObject ivarGet(VariableSite site, IRubyObject self) throws Th
         VariableAccessor accessor = realClass.getVariableAccessorForRead(site.name());
 
         // produce nil if the variable has not been initialize
-        MethodHandle nullToNil = self.getRuntime().getNullToNilHandle();
```
headius: The purpose of getNullToNilHandle was to have a single place to cache the handle and reuse it everywhere. This is going to recreate handles every time. I agree we need a better place than on Ruby, but we still need these one-time handles.

core/src/main/java/org/jruby/ir/targets/Bootstrap.java
```diff
@@ -657,7 +653,10 @@ public static IRubyObject ivarGet(VariableSite site, IRubyObject self) throws Th
...
+        IRubyObject nilObject = self.getRuntime().getNil();
+        MethodHandle nullToNil = lookup().findStatic(Helpers.class, "nullToNil", methodType(IRubyObject.class, IRubyObject.class, IRubyObject.class));
```
headius: Same issue here with excessive handle creation.

core/src/main/java/org/jruby/ir/targets/ConstantLookupSite.java
```diff
@@ -248,7 +248,15 @@ private void bind(Ruby runtime, RubyModule module, IRubyObject constant, MethodH
         MethodHandle fallback = getFallback(module, cachingFallback);
 
         // Test that module is same as before
-        target = guardWithTest(module.getIdTest(), target, fallback);
+        MethodHandles.Lookup lookup = MethodHandles.lookup();
+        MethodHandle testModuleMatch = null;
+        try {
+            testModuleMatch = lookup.findStatic(Bootstrap.class, "testModuleMatch", MethodType.methodType(boolean.class, ThreadContext.class, IRubyObject.class, int.class));
```
headius: Same issue here with excessive module creation. I'm sure we can find a way/place to save these handles without polluting things so much that Android doesn't work.

core/src/main/java/org/jruby/runtime/BlockBody.java
```diff
@@ -50,7 +50,7 @@
 public abstract class BlockBody {
 
     protected final Signature signature;
-    protected volatile MethodHandle testBlockBody;
```
headius: This is not used but might be later. It is also to avoid recreating the same handle over and over again. It could be made Object and casted elsewhere to avoid the hard reference of MethodHandle.

core/src/main/java/org/jruby/util/invoke/MethodHandle.java
```diff
@@ -0,0 +1,71 @@
+package org.jruby.util.invoke;
+
+import java.lang.reflect.Constructor;
+import java.lang.reflect.InvocationTargetException;
+import java.lang.reflect.Method;
+
+public class MethodHandle {
```
headius: This part is pretty nice, and I almost wonder if we should pull it out into its own library, for others to use.

core/src/main/java/org/jruby/util/io/ChannelFD.java
```diff
@@ -131,7 +130,7 @@ private void initChannelTypes() {
         else chRead = null;
         if (ch instanceof WritableByteChannel) chWrite = (WritableByteChannel)ch;
         else chWrite = null;
-        if (ch instanceof SeekableByteChannel) chSeek = (SeekableByteChannel)ch;
```
headius: What's the reason for this? SeekableByteChannel is not a new class, is it? I suppose on current Java the only SeekableByteChannel is FileChannel, but I prefer to use the more generic interface if possible.

donv:
Thanks for the review!  I can fix the import expansions.

Do you prefer I force-push those changes, or add a new commit?

core/src/main/java/org/jruby/util/io/ChannelFD.java
donv: The reason for this is that SeekableByteChannel was added to Android 7.0 API level 24 ref.:

https://developer.android.com/reference/java/nio/channels/SeekableByteChannel.html

FileChannel has been in Android since the initial release.

I agree that using the interface is the better option, and we can do a patch in Ruboto if necessary, although better if we don't have to 😄 .

headius:
@donv I don't care if you do it as a separate commit or not. The original set of commits was long and has already been squashed, so we don't really have much history left in this PR anyway.

core/src/main/java/org/jruby/util/io/ChannelFD.java
headius: Ugh. I'd really rather not use FileChannel everywhere, even though I don't know of any SeekableByteChannel offhand (even jnr-enxio channels don't implement the interface). And changing chSeek to be a FileChannel makes it essentially just a duplicate of chFile.

I hate to make more work here but this is a pretty icky change to have to make. I wish Google's incorporation of Java 7 APIs weren't so arbitrary.

donv:
OK, I fixed the '*' imports.  Which of the other comments need to be done before merging?

core/src/main/java/org/jruby/util/invoke/MethodType.java
```diff
@@ -0,0 +1,53 @@
+package org.jruby.util.invoke;
+
+import org.jruby.javasupport.ext.JavaLangReflect;
```
kares: :scissors:

core/src/main/java/org/jruby/util/io/ChannelFD.java
donv: I'll do a patch in Ruboto, then.

core/src/main/java/org/jruby/util/invoke/MethodType.java
donv: Removed.

core/src/main/java/org/jruby/util/io/ChannelFD.java
donv: OK, reinstated the use of SeekableByteChannel and puts a copy of the ChannelFD source in Ruboto.

donv:
OK, reinstated SeekableByteChannel.  What more needs to be done?  I am on it in a flash! ⚡️

headius:
@donv The places where we removed MethodHandle from e.g. Ruby, RubyModule, etc, should be reinstated, but signatures changed to something Android supports (maybe just Object) and the creation of the handles indirected through e.g. OptoFactory. We need to still be able to cache those handles rather than recreating them every time. nullToNil, for example.

headius:
@donv It occurs to me we could use generics to fix those methods; when called with a MethodHandle factory, they'd return MethodHandle, so type-safety on non-Android would still be restored. I may try to prototype that this evening.

donv:
Cool!  Looking forward to it 😄

headius:
Ok...obviously I didn't get back to it before my trip.

I'll review what we have now.

headius:
@donv Sorry for the long delay but I made my changes. Since there were only two places where we had removed handle caches, I just restored them as all Object signatures. Unless used (by JIT+indy) they shouldn't get called and shouldn't trigger load of Java 7+ indy classes.

Review my changes please! I think this could go into 9.2. @enebo?

headius:
So I realized that the travis build has not been green. It appears that it's failing because the JIT is still running and using invokedynamic for globals and constants, but indy defaults to false which disables the SwitchPoint-based invalidators for constants and globals.

I'll see what I can figure out.

headius:
Problem stems from this change, which provided a non-indy way to create a constant invalidator: https://github.com/jruby/jruby/commit/1f31168b54d841e57ea55bdeb520fb84d067be01

We'll need to modify the JIT to have non-indy global and const logic.

headius:
I have pushed a few more commits that get things further along:

* Constants and globals now have (uncached, slow) non-indy logic in the JIT. This allows them to build right with indy disabled and still use indy when it is turned on. More on this below
* Another small fix to the `nullToNil` cached filter.

We can't really ship this for JRuby proper without providing some non-indy caching for constants and globals, since that would represent a fairly bad regression. At the very least we need a better way to pick and choose features based on profile: for example, on desktop Java 7 we may use a few indy features even if full indy is not turned on, but on Android we must avoid all indy features but may be able to use some other 7 features.

Rake still seems to be having trouble starting up, now due to an NPE in the IR interpreter. @enebo

headius:
Tests appear to be going green now, yay. We'll see how it looks and then see about optimizing these non-indy compiler changes.

headius:
I'm confused by the MRI failures. A handful of tests that attempt to run in a subprocess seem to be timing out, in this and other builds (see also the spec:ruby:fast timeout).

I'll poke at things today and try to figure it out.

donv:
@headius This is looking great!  Seems we are getting close!

headius:
@donv Whatever's causing those two failures seems to be affecting all branches, so we may actually be good to go with the branch as-is.

donv:
Woohoo!

donv:
@headius Good news!  I used the jruby-jars from this branch and upgraded one of our production apps today.  And it is now operational!  Sure, it is a simple app with very few in-house users, but it works well 😄 

I encountered some problems I had to work around, but I'd like to get this PR merged before I report those.

You mentioned we may be ready to merge this PR?  Doable?  Yes?

donv:
@headius Can we merge this issue now?  I have a problem with subclassing a Java class on Android, but it would be good to work off JRuby master with this PR merged.

donv:
Hi @headius !

Master has moved on a bit.  I can try to resolve the conflicts unless you want to do it.

PR #4590, opened 2017-05-05
Clean up structure of io/console and avoid stty on Windows.

### LINKED ISSUE #3989: JRuby + Cygwin + Jenkins causes message related to stty
### Environment

   jruby 1.7.25 (1.9.3p551) 2016-04-13 867cb81 on Java HotSpot(TM) 64-Bit Server VM 1.8.0_31-b13 +jit [Windows 7-amd64]
### Expected Behavior
- Ruby should not print warnings related to tty when running on Windows
### Actual Behavior

When I run JRuby on Jenkins, but invoked from a Cygwin shell, I get the annoying warning message

```
stty: standard input: Inappropriate ioctl for device
```

This does not happen when I run JRuby from a (Cygwin-) command line (interactively), or on Jenkins from a Windows command line.

I remember that there was  a very similar error in jruby 1.7.23, where a stty warning was always shown when executing jruby from a Cygwin shell script, even with an interactive one. This error had been fixed in 1.7.24.

Running the script as a Jenkins job means that there is no console associated to the job. Perhaps the bugfix mentioned, did not pay attention to this possibility.

Thomas Enebo commented to this in the JRuby mailing list in the following way: "The gist of things is when you run via cygwin on Jenkins it is from a daemon and it has no tty associated with it.  When we run stty you get this error.  When you run from a cygwin windows it does have a tty so stty completes without issue."

headius:
This should fix #3989.

I have restructured io/console to be more modular and now force Windows to always use the stubbed version...never the stty version.

Other platforms should work as they did before, but the code is a bit easier to follow now.


headius:
Request review from @kares, @enebo. This is *almost* structurally identical to the old one, but the old logic was hard to follow and had unusual guards (like the `ttymode` defined check) that I did not carry over.

I believe the new logic is cleaned and more appropriate:

1. If Windows, use stubbed always.
2. If Linux or BSD, attempt to use native; if that fails, continue with (3)
3. Attempt to use stty. If that fails, fall back to stubbed.

kares:
:+1: such a cleanup would be good to have on 9K (pretty much all in one file)
although I am not sure how deep its modified compared to the stock version for MRI (due stdlib imports).

headius:
In MRI, io/console is written entirely in C. Part of this reorg is to get files like `linux_console.rb` out of the io load namespace. I realize I have added a new namespace with JRuby-specific files, but I think it is isolated pretty well.

I will get this into 9k as well since they're both from the same sources.

PR #4592, opened 2017-05-07
Clean up structure of io/console and avoid stty on Windows.

### LINKED ISSUE #3989: JRuby + Cygwin + Jenkins causes message related to stty

headius:
Fixes #3989.

* Restructure the different impls of console into their own files.
* Always use stubbed version on Windows.
* Cascade from native to stty to stubbed on other platforms.

cc @kares @enebo @jmiettinen


headius:
@enebo Any concerns?

enebo:
@headius no.  Are you going to get this upstream too?  Or parts of it at least?

headius:
@enebo Upstream where? This is JRuby-specific; in MRI io/console is a C ext. This change will go into both 9k and 1.7.

enebo:
well I meant MRI and forgot it was a cext.

PR #4599, opened 2017-05-10
Fix bootstrapping on AARCH64

etehtsea:
It fails with:

```
[INFO] --- gem-maven-plugin:1.0.10:initialize (default) @ jruby-tests ---
[INFO] ERROR:  Loading command: install (LoadError)
[INFO]  Unsupported platform: unknown-linux
[INFO] ERROR:  While executing gem ... (NoMethodError)
[INFO]     undefined method `invoke_with_build_args' for nil:NilClass
[INFO]
```

### DIFF
--- a/test/pom.rb
+++ b/test/pom.rb
@@ -61,7 +61,7 @@ project 'JRuby Integration Tests' do
       'includeRubygemsInTestResources' => 'false' }
 
     if version =~ /-SNAPSHOT/
-      options[ 'jrubyVersion' ] = '1.7.12'
+      options[ 'jrubyVersion' ] = '9.1.8.0'
     else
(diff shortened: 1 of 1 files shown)


kares:
hmm, built seems to be taking quite longer ... from 3+h to 4+h - probably due memory constrains? 
- https://travis-ci.org/jruby/jruby/builds/239746070
- https://travis-ci.org/jruby/jruby/builds/239843760
... maybe a later 9K will cut down some :

headius:
@kares I'll kick that build and see if it comes out as long the second time. May have been a glitch.

If not, may be caching 9k + deps first time, and won't affect it in the future.

headius:
@kares Second build seemed fine at 3h11m so it seems ok.

PR #4616, opened 2017-05-19
[ruby 2.4] Implemented Integer#ceil, #floor, #truncate

whwilder:
Feature #12245 ( ref #4293 )
Fixed the issue with my previous PR (#4571)
There was already an implementation of Integer#round. I noticed it had an implementation in Integer that covered both Fixnum and Bignum. Would this be a preferable way of doing things, or is my current method of splitting the implementation between Fixnum and Bignum acceptable?


core/src/main/java/org/jruby/RubyBignum.java
```diff
@@ -236,6 +236,63 @@ public static BigInteger long2big(long arg) {
      *  ================
      */
 
+    /** rb_big_ceil
+     *
+     */
+    @Override
+    public IRubyObject ceil(ThreadContext context, IRubyObject args){
```
kares: `args` should be named `arg` or something (everywhere) - otherwise its a bit confusing.

core/src/main/java/org/jruby/RubyBignum.java
```diff
@@ -236,6 +236,63 @@ public static BigInteger long2big(long arg) {
...
+
+    /** rb_big_truncate
+     *
+     */
+    @Override
+    public IRubyObject truncate(ThreadContext context, IRubyObject args){
+        BigInteger self = value;
+        if (self.compareTo(BigInteger.ZERO) == 1){
+            return floor(context, args);
+        } else if (self.compareTo(BigInteger.ZERO) == -1){
```
kares: maybe `compareTo` could be performance only once and saved in a local instead of twice

core/src/main/java/org/jruby/RubyBignum.java
```diff
@@ -236,6 +236,63 @@ public static BigInteger long2big(long arg) {
...
+    /** rb_big_floor
+     *
+     */
+    @Override
+    public IRubyObject floor(ThreadContext context, IRubyObject args){
+        BigInteger ndigits = args.convertToInteger().getBigIntegerValue();
+        BigInteger self = value;
+        if (ndigits.compareTo(BigInteger.ZERO) == 1){
+            return convertToFloat();
+        } else if (ndigits.compareTo(BigInteger.ZERO) == 0){
```
kares: maybe `compareTo` could be performance only once and saved in a local instead of twice

core/src/main/java/org/jruby/RubyBignum.java
```diff
@@ -236,6 +236,63 @@ public static BigInteger long2big(long arg) {
...
+            return newBignum(context.runtime, res);
+        }
+    }
+
+    /** rb_big_floor
+     *
+     */
+    @Override
+    public IRubyObject floor(ThreadContext context, IRubyObject args){
+        BigInteger ndigits = args.convertToInteger().getBigIntegerValue();
```
kares: is the `arg` really going to be a `Bignul` value isn't it an over-kill ?

core/src/main/java/org/jruby/RubyBignum.java
```diff
@@ -236,6 +236,63 @@ public static BigInteger long2big(long arg) {
      *  ================
      */
 
+    /** rb_big_ceil
+     *
+     */
+    @Override
+    public IRubyObject ceil(ThreadContext context, IRubyObject args){
+        BigInteger ndigits = args.convertToInteger().getBigIntegerValue();
```
kares: is the `arg` really going to be a `Bignul` value isn't it an over-kill ?

PR #4627, opened 2017-05-26
README: Mention jruby-9.1.10.0 in rbenv note

olleolleolle:
This PR updates the README so that it refers to the latest release of JRuby when explaining how to install it using `rbenv`.

See #4622 for a PR that does the same thing for last version.


olleolleolle:
(Would it be mergable if I added `[ci skip]`?

kares:
its fine as is, thanks Olle

PR #4636, opened 2017-05-31
Implement Zlib::GzipReader #ungetbyte and #ungetc

### LINKED ISSUE #4631: Zlib::GzipReader does not support #ungetc or #ungetbyte
### Environment

```
jruby 9.1.10.0 (2.3.3) 2017-05-25 b09c48a Java HotSpot(TM) 64-Bit Server VM 25.121-b13 on 1.8.0_121-b13 +jit [darwin-x86_64]
```

### Expected Behavior

`Zlib::GzipReader` supports pushing data back onto the buffer via `#ungetc` and `#ungetbyte`.

```ruby
# frozen_string_literal: true

require "zlib"

io = StringIO.new
writer = Zlib::GzipWriter.new(io)
writer.write "foo bar baz"
writer.close

reader = Zlib::GzipReader.new(StringIO.new(io.string))
reader.getc
reader.ungetc "b"
puts reader.read
```

```console
$ ruby -v
ruby 2.4.1p111 (2017-03-22 revision 58053) [x86_64-darwin16]

$ ruby gzip_reader_ungetc.rb
boo bar baz

$ ruby -r zlib -e "p Zlib::GzipReader.public_method_defined?(:ungetbyte)"
true
```
### Actual Behavior

`Zlib::GzipReader` implements `#ungetc` as a [no-op](https://github.com/jruby/jruby/blob/9.1.10.0/core/src/main/java/org/jruby/ext/zlib/JZlibRubyGzipReader.java#L637-L639) and does not respond to `#ungetbyte`.

```console
$ ruby -v
jruby 9.1.10.0 (2.3.3) 2017-05-25 b09c48a Java HotSpot(TM) 64-Bit Server VM 25.121-b13 on 1.8.0_121-b13 +jit [darwin-x86_64]

$ ruby gzip_reader_ungetc.rb
oo bar baz

$ ruby -r zlib -e "p Zlib::GzipReader.public_method_defined?(:ungetbyte)"
false
```

haines:
I've had a stab at implementing the `unget*` methods for `Zlib::GzipReader`. The implementation is a bit limited in that the maximum number of bytes that can be "ungot" is hardcoded, since I'm using a `PushbackInputStream`. Also, it doesn't attempt to do anything clever with encoding (although I don't see that as a problem because `getc` [doesn't either](https://github.com/jruby/jruby/blob/9.1.10.0/core/src/main/java/org/jruby/ext/zlib/JZlibRubyGzipReader.java#L488-L489)).

Let me know what you think!

Closes #4631

### DIFF
--- a/core/src/main/java/org/jruby/ext/zlib/JZlibRubyGzipReader.java
+++ b/core/src/main/java/org/jruby/ext/zlib/JZlibRubyGzipReader.java
@@ -34,6 +34,7 @@ import org.jruby.RubyClass;
 import org.jruby.RubyEnumerator;
 import org.jruby.RubyException;
 import org.jruby.RubyIO;
+import org.jruby.RubyInteger;
 import org.jruby.RubyNumeric;
 import org.jruby.RubyString;
 import org.jruby.anno.FrameField;
@@ -42,7 +43,6 @@ import org.jruby.anno.JRubyMethod;
 import org.jruby.exceptions.RaiseException;
 import org.jruby.runtime.Block;
 import org.jruby.runtime.Helpers;
-import org.jruby.runtime.JavaSites;
 import org.jruby.runtime.ObjectAllocator;
 import org.jruby.runtime.ThreadContext;
 import org.jruby.runtime.builtin.IRubyObject;
@@ -50,11 +50,12 @@ import org.jruby.util.ByteList;
 import org.jruby.util.IOInputStream;
 import org.jruby.util.StringSupport;
 import org.jruby.util.TypeConverter;
+import org.jruby.util.io.EncodingUtils;
 import org.jruby.util.io.PosixShim;
 
 import java.io.BufferedInputStream;
 import java.io.IOException;
-import java.io.InputStream;
+import java.io.PushbackInputStream;
 import java.util.ArrayList;
 import java.util.List;
 
@@ -133,7 +134,7 @@ public class JZlibRubyGzipReader extends RubyGzipFile {
 
         position = 0;
         line = 0;
-        bufferedStream = new BufferedInputStream(io);
+        bufferedStream = new PushbackInputStream(new BufferedInputStream(io), 512);
 
         return this;
     }
@@ -502,9 +503,9 @@ public class JZlibRubyGzipReader extends RubyGzipFile {
         // encountered the actual EOF during the reading.
         // So, we compensate for that to provide MRI
         // compatible behavior.
-        bufferedStream.mark(16);
-        bufferedStream.read();
-        bufferedStream.reset();
+        byte[] bytes = new byte[16];
+        int read = bufferedStream.read(bytes, 0, bytes.length);
+        bufferedStream.unread(bytes, 0, read);
 
         return bufferedStream.available() == 0;
     }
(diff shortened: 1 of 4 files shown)


haines:
I think the build failure is actually an issue with the implementation of `gets("")` -- MRI skips the leading newline but JRuby does not.

core/src/main/java/org/jruby/ext/zlib/JZlibRubyGzipReader.java
```diff
@@ -29,32 +29,26 @@
...
 import com.jcraft.jzlib.GZIPInputStream;
 import com.jcraft.jzlib.Inflater;
-import org.jruby.Ruby;
-import org.jruby.RubyClass;
-import org.jruby.RubyEnumerator;
-import org.jruby.RubyException;
-import org.jruby.RubyIO;
-import org.jruby.RubyNumeric;
-import org.jruby.RubyString;
+import org.jruby.*;
```
headius: We generally don't collapse imports unless there's >20 or something.

test/jruby/test_zlib.rb
```diff
@@ -517,6 +517,24 @@ def test_error_input
     assert_equal("not in gzip format", e.message)
     assert_equal("foobarzothoge", e.input)
   end
+
+  def test_gzip_reader_ungetc
```
headius: If these don't exist in either MRI's suite or ruby/spec, it would be nice to add them there. We usually just put JRuby-specific stuff in test/jruby.

haines: Makes sense. There are placeholders for these methods in ruby/spec, so I'll add them there.

Should I submit them to ruby/spec first and then pull them in once merged upstream?

headius: You can submit them to our spec/ruby subdir. We merge both ways.

headius: Make it a separate commit for sure though...makes it easier to merge. 😄

haines:
I've added the specs, which caught a bug in my implementation (I forgot to decrement `pos`), and a couple in MRI:

* ungetting at the start of the stream causes `pos` to underflow (I've raised this: https://bugs.ruby-lang.org/issues/13616)
* ungetting `nil` raises a `TypeError: no implicit conversion of nil`. I'm not sure if I should raise this as a bug in MRI or leave the behaviour unspecified. `File` accepts `nil` and silently ignores it, so I think `GzipReader` should behave in the same way for consistency. On the other hand, trying to unget `nil` is a strange thing to do, so raising a `TypeError` is fairly reasonable. What do you think?

I'll take a look at the `#gets` implementation tomorrow (UK time).

haines:
I think this is good to go (assuming you're happy with it!). Turns out `#gets("")` is supposed to skip any number of leading or trailing newlines, so I've added a spec for that behaviour too.

headius:
Looks good now...I'll merge!

eregon:
@headius @haines 
> ungetting `nil` raises a `TypeError: no implicit conversion of nil`. I'm not sure if I should raise this as a bug in MRI or leave the behaviour unspecified. `File` accepts `nil` and silently ignores it, so I think `GzipReader` should behave in the same way for consistency. On the other hand, trying to unget `nil` is a strange thing to do, so raising a `TypeError` is fairly reasonable. What do you think?

Either follow what MRI does, or raise a bug, but please not spec different behavior without a bug report.
I cannot keep these specs in ruby/spec since they don't pass on MRI and there is no matching MRI bug.
I'll write a bug report and put these specs in quarantine in the meantime.

eregon:
https://bugs.ruby-lang.org/issues/13675

PR #4608, opened 2017-05-17
Fix UDPSocket#(send|write) to raise EMSGSIZE

etehtsea:

Issue #4606, opened 2017-05-14
JRuby raises SystemCallError instead of Errno::EMSGSIZE


dwbutler:
### Environment

```
jruby -v
jruby 9.0.5.0 (2.2.3) 2016-01-26 7bee00d Java HotSpot(TM) 64-Bit Server VM 25.74-b02 on 1.8.0_74-b02 +jit [darwin-x86_64]

uname -a
Darwin Davids-MacBook-Pro-6.local 16.4.0 Darwin Kernel Version 16.4.0: Thu Dec 22 22:53:21 PST 2016; root:xnu-3789.41.3~3/RELEASE_X86_64 x86_64
```

### Summary

JRuby raises `SystemCallError` instead of `Errno:: EMSGSIZE` when writing too large of a message to a UDP socket. See https://github.com/dwbutler/logstash-logger/issues/128.

### Setup

    $ nc -l 5228

Ruby script (run from irb):
```
require 'socket'
socket = UDPSocket.new
socket.connect('127.0.0.1', 5228)
socket.write "1"*100_000
```

### Expected Behavior

MRI 2.4.1:

```
Errno::EMSGSIZE: Message too long
	from (irb):4:in `write'
```

### Actual Behavior

JRuby 9.0.5.0:
```
SystemCallError: Unknown error - No message available
	from org/jruby/RubyIO.java:1437:in `write'
	from org/jruby/RubyIO.java:1408:in `write'
```

etehtsea:
I think I can fix this.

kares:
and all is well ... thanks to @etehtsea

omarmarquez:
Is this the same bug?
`$ ruby -v
jruby 9.1.16.0 (2.3.3) 2018-02-21 8f3f95a Java HotSpot(TM) 64-Bit Server VM 25.92-b14 on 1.8.0_92-b14 +jit [linux-x86_64]

$ ruby rcv.rb
SocketError: No message available
  block in rcv.rb at rcv.rb:16
             loop at org/jruby/RubyKernel.java:1316
           <main> at rcv.rb:15

$ cat rcv.rb
#!/usr/bin/env ruby
require 'socket'
require 'ipaddr'
BIND_ADDR = "0.0.0.0"
MULTICAST_ADDR = "239.203.120.1"
PORT = 10000

sock = UDPSocket.new
ip =  IPAddr.new(MULTICAST_ADDR).hton + IPAddr.new(BIND_ADDR).hton
[...]

PR #4638, opened 2017-06-01
Also fix EMSGSIZE on Alpine Linux. Relates to #4606

etehtsea:
See https://github.com/jruby/jruby/pull/4608#issuecomment-305386816


kares:
najs catch for the :mountain_snow:

headius:
@kares There will definitely be a 9.1.11.0 because of the FFI pointer leak, so I re-marked this.

kares:
@headius good, I thought it will be from a separate branch (not master) ... whatever is fine with me though

Issue #1938, opened 2014-09-01
Reincluded modules do not splice changes into hierarchy in the correct order


headius:
Title is a mouthful, but the basic issue is that if a module is included, then modified by including other modules into itself, then included again, the new includes don't get spliced into the target hierarchy correctly.

The following example shows that the order of modules is different from MRI:

Output:

```
[--dev]
~/projects/jruby $ rvm ruby-2.1 do ruby blah.rb
in Z
in A
in Q

[--dev]
~/projects/jruby $ jruby blah.rb
in A
in Z
in Q
```

Script:

``` ruby
module X
end

class Q
  def foo
    puts 'in Q'
  end
end

class Y < Q
  include X
end

module A
  def foo
    puts 'in A'
    super
  end
end

module X
  include A
end

module Z
  def foo
    puts 'in Z'
    super
  end
end

class Y
  include Z
end

class Y
  include X
end

Y.new.foo
```

This is the likely cause of Compass/compass-rails#179.

tamird:
@headius  was this resolved by #2002?

kares:
just checked - this **works in JRuby 1.7.22 but still fails in 9K** (9.0.4.0)

PR #4639, opened 2017-06-01
Fix Module inclusion order

MSNexploder:
This fixes Module inclusion order as seen in #1938


headius:
Failure appears to be directly related to the patch. Can you take a look into it?

MSNexploder:
Will look into it. Looks like the whole Module thing is not as straightforward as I hoped. 😉

MSNexploder:
Pretty sure I got the Module#prepend working again 🎉

headius:
Well it passes tests...so I'll do a quick review and get it in! Thanks!

Issue #4444, opened 2017-01-17
java.lang.NoSuchMethodError: rubyobj.Grape.Util.InheritableValues.callMethod when running with -Xreify.classes=true


ivoanjo:
### Environment

Running `jruby 9.1.7.0 (2.3.1) 2017-01-11 68056ae Java HotSpot(TM) 64-Bit Server VM 25.111-b14 on 1.8.0_111-b14 +jit [linux-x86_64]` on `Linux maruchan 4.9.0-11-generic #12-Ubuntu SMP Mon Dec 12 16:18:23 UTC 2016 x86_64 x86_64 x86_64 GNU/Linux`.

### Expected Behavior

Example testcase:

```ruby
require 'grape'

class Foo < Grape::API
  post { }
end
```

(Attached version includes `Gemfile` and `Gemfile.lock`: [testcase.zip](https://github.com/jruby/jruby/files/711279/testcase.zip))

When running under JRuby normally, this code has no problems.

### Actual Behavior

When running this code with `jruby -Xreify.classes=true test.rb` it results in the following exception:

```
Unhandled Java exception: java.lang.NoSuchMethodError: rubyobj.Grape.Util.InheritableValues.callMethod(Ljava/lang/String;Lorg/jruby/runtime/builtin/IRubyObject;)Lorg/jruby/runtime/builtin/IRubyObject;
java.lang.NoSuchMethodError: rubyobj.Grape.Util.InheritableValues.callMethod(Ljava/lang/String;Lorg/jruby/runtime/builtin/IRubyObject;)Lorg/jruby/runtime/builtin/IRubyObject;
      initialize_copy at rubyobj/Grape/Util/InheritableValues:-1
      initialize_copy at org/jruby/RubyKernel.java:1836
                 call at org/jruby/RubyKernel$INVOKER$s$1$0$initialize_copy.gen:-1
                 call at org/jruby/internal/runtime/methods/JavaMethod.java:738
        instanceSuper at org/jruby/ir/runtime/IRRuntimeHelpers.java:990
            interpret at org/jruby/ir/instructions/InstanceSuperInstr.java:69
          processCall at org/jruby/ir/interpreter/InterpreterEngine.java:356
            interpret at org/jruby/ir/interpreter/StartupInterpreterEngine.java:73
            interpret at org/jruby/ir/interpreter/InterpreterEngine.java:84
     INTERPRET_METHOD at org/jruby/internal/runtime/methods/MixedModeIRMethod.java:179
[...]

PR #4642, opened 2017-06-01
Fix missing callMethod override

ivoanjo:
Although not used by default, the Reificator class at https://github.com/jruby/jruby/blob/9.1.10.0/core/src/main/java/org/jruby/RubyClass.java#L1583 checks which version of callMethod it should use, and needs all three versions: no argument, one argument, and more than one argument.

One of these overrides did not exist, and thus would result on a

```
java.lang.NoSuchMethodError: rubyobj.BrokenReify.callMethod(Ljava/lang/String;Lorg/jruby/runtime/builtin/IRubyObject;)Lorg/jruby/runtime/builtin/IRubyObject;
```

when running an example with -Xreify.classes=true

```ruby
class BrokenReify
  def initialize_copy(other)
    super
  end
end

puts BrokenReify.new.clone
```

With this fix, the above example starts instead giving a java.lang.StackOverflowError due to another bug/bad interaction of the reify.classes option, which I'll report separately.

Issue #4444

### DIFF
--- a/core/src/main/java/org/jruby/RubyBasicObject.java
+++ b/core/src/main/java/org/jruby/RubyBasicObject.java
@@ -376,6 +376,10 @@ public class RubyBasicObject implements Cloneable, IRubyObject, Serializable, Co
         return Helpers.invoke(getRuntime().getCurrentContext(), this, name, args);
     }
 
+    public final IRubyObject callMethod(String name, IRubyObject arg) {
+        return Helpers.invoke(getRuntime().getCurrentContext(), this, name, arg);
+    }
+
     public final IRubyObject callMethod(String name) {
         return Helpers.invoke(getRuntime().getCurrentContext(), this, name);
     }


kares:
:+1: 

was lately thinking about what needs to be done for `-Xreify` to be reliable in 9K like it used to be in 1.7.
maybe a fast suite for starters running `-Xreify.classes` although it might turn out to be a bit much test time.
but I still find reify quite useful with profiling, thread dumping ... any hints you might have @headius ?

kares:
also this could than incorporate a test such as the one mentioned in the desc. 
seems that the RubyClass calls might get eventually deprecated and we should dispatch to what JIT uses.

ivoanjo:
I plan to open a separate issue, but when I was looking at the code today to debug this I ended up discovering that `reify.classes` can mostly be used for two purposes:

1. To have ruby instances that correspond to their class, rather than `RubyObject` and friends. This is invaluable when debugging memory issues as otherwise it's very hard to see what's going on in VisualVM and friends

2. To have a Java class that contains the methods of the ruby class, allowing them to be called by Java code

I believe that just having 1) would be extremely valuable. Double so if it was enabled by default.

On the other hand, 2) seems to be what mostly limits this feature, and actually is what makes the test above enter an infinite recursion: `RubyKernel` calls `initialize_copy` on the object thinking that `RubyBasicObject.initialize_copy` will be executed, when in reality it was overriden in the reified class with a trampoline that calls `super`, and then off we go again.

headius:
This change is good and I'll merge it. We'll continue discussions about fixing/improving reification in other bugs like #4643.

PR #4646, opened 2017-06-05
try harder to avoid long overflow in Time#+ (fixes #1779)

### LINKED ISSUE #1779: Time.now + bignum incorrect
```
irb(main):061:0> Time.now + 7819128400
=> 2262-04-11 16:46:19 -0700

irb(main):062:0> Time.now + 7819128500
=> 1677-09-20 16:20:34 -0752 # time travel to past!
```

MSNexploder:


kares:
... very nice Stefan, could we have a test or spec please (as this is easy to break at some point) ?

MSNexploder:
Added a small test (and reenabled some existing time related regression tests which pass) but i'm not sure if this is the best way to test.

kares:
@MSNexploder thanks, its not bad (you can leave it in) but does not actually test for the reported issue, 
a new test which would test smt like : 
```
t = Time.local(2000, 1, 1) + (300 * 366 * 24 * 60 * 60)
assert_equal 2300, t.year
```
... would probably do, what do you think?

MSNexploder:
@kares That was too easy 😎 Was focusing way to much on the underlying problem.

kares:
cool thanks a lot ... @headius could you double-check this is good to go for @MSNexploder, please?

headius:
Looks good, thank you!

Issue #4643, opened 2017-06-02
Reificator overriding methods breaks RubyKernel dispatch


ivoanjo:
### Environment

* JRuby: `jruby 9.1.10.0 (2.3.3) 2017-05-25 b09c48a Java HotSpot(TM) 64-Bit Server VM 25.131-b11 on 1.8.0_131-b11 [linux-x86_64]`
* Kernel: `Linux maruchan 4.11.0-041100-generic #201705041534 SMP Thu May 4 19:36:05 UTC 2017 x86_64 x86_64 x86_64 GNU/Linux`
* Distro: `Ubuntu 17.04`

### Expected Behavior

When investigating failures while using the `-Xreify.classes` option and after fixing the issue in #4642, it turns out that then I got bit by #3793.

I spent a few minutes looking at why

```ruby
class BrokenReify
  def to_s
    super
  end
end

puts BrokenReify.new.to_s
```

causes an infinite recursion:

```java
java.lang.StackOverflowError
[...]
        at rubyobj.BrokenReify.to_s(Unknown Source)
        at org.jruby.RubyKernel.to_s(RubyKernel.java:1976)
        at org.jruby.RubyKernel$INVOKER$s$0$0$to_s.call(RubyKernel$INVOKER$s$0$0$to_s.gen)
        at org.jruby.internal.runtime.methods.JavaMethod$JavaMethodN.call(JavaMethod.java:737)
        at org.jruby.ir.runtime.IRRuntimeHelpers.instanceSuper(IRRuntimeHelpers.java:983)
        at org.jruby.ir.instructions.InstanceSuperInstr.interpret(InstanceSuperInstr.java:69)
        at org.jruby.ir.interpreter.InterpreterEngine.processCall(InterpreterEngine.java:355)
        at org.jruby.ir.interpreter.StartupInterpreterEngine.interpret(StartupInterpreterEngine.java:73)
        at org.jruby.ir.interpreter.InterpreterEngine.interpret(InterpreterEngine.java:77)
        at org.jruby.internal.runtime.methods.InterpretedIRMethod.INTERPRET_METHOD(InterpretedIRMethod.java:150)
        at org.jruby.internal.runtime.methods.InterpretedIRMethod.call(InterpretedIRMethod.java:137)
        at org.jruby.RubyClass.finvoke(RubyClass.java:557)
        at org.jruby.runtime.Helpers.invoke(Helpers.java:397)
[...]

PR #4648, opened 2017-06-07
New reify classes

headius:
Work to unify variable and structural reification of Ruby classes. See #4643.


core/src/main/java/org/jruby/specialized/RubyObjectSpecializer.java
```diff
@@ -0,0 +1,277 @@
...
+
+    private static ClassDefiningClassLoader CDCL = new OneShotClassLoader(Ruby.getClassLoader());
+
+    public static ObjectAllocator specializeForVariables(RubyClass klass, Set<String> foundVariables) {
+        int size = foundVariables.size();
+        ObjectAllocator allocator = getClassFromSize(size);
+
+        if (allocator != null) return allocator;
+
+        final String clsPath = "org/jruby/specialize/RubyObject" + size;
```
kares: there's **org/jruby/gen** to distinguish generated code, maybe it makes sense to use it as well (org.jruby.gen.specialize)?

headius: I agree this generated code should go into org.jruby.gen. The "org.jruby.specialize" was a typo on my part; it would have been specialized, which I think we will keep to separate the hand-written pieces from the generated stuff.

ivoanjo:
I've just retried with this branch with the following testcase (from #4643)

```ruby
class BrokenReify
  def to_s
    super
  end
end

puts BrokenReify.new.to_s
``` 

And it still causes a stack overflow. I believe the reason why it fails is still similar: it's because unlike other reification options (e.g. reify variables) we define the instance methods from the ruby class in the reified class, which is what breaks code which tries to call `super`.

I'm not sure what the is best choice here:
* One alternative would just to not add the instance methods
* Or they could be added, but using some name mangling (does any other part of JRuby really use them at all?)
* Or maybe the solution would be to hunt down and fix any assumptions that you can call kernel methods on the object instance e.g https://github.com/jruby/jruby/blob/9.1.10.0/core/src/main/java/org/jruby/RubyKernel.java#L1974-L1977 ).

PR #4656, opened 2017-06-08
SecureRandom update to match Ruby 2.3

### LINKED ISSUE #4607: SecureRandom.random_number doesn't accept a range
The problem is here:
https://github.com/jruby/jruby/blob/master/lib/ruby/stdlib/securerandom.rb#L147

At some point, the code to handle ranges was removed.

### Environment

jruby 9.1.8.0 (2.3.1) 2017-03-06 90fc7ab Java HotSpot(TM) 64-Bit Server VM 25.121-b13 on 1.8.0_121-b13 +jit [mswin32-x86_64]

Windows 10 Professional x64

### Expected Behavior

In standard Ruby I can do `puts SecureRandom.random_number(50..55)` and get a random result between 50 and 55.

### Actual Behavior

 If I do this in JRuby, I get an argument error for using a Range in place of a Fixnum.

kares:
securerandom.rb imported from MRI's 2.3 stable line

re-implemented `random_number` (native in C-Ruby now) in Java to handle ranges

resolves #4607

### DIFF
--- a/core/src/main/java/org/jruby/ext/securerandom/RubySecureRandom.java
+++ b/core/src/main/java/org/jruby/ext/securerandom/RubySecureRandom.java
@@ -28,6 +28,11 @@
  ***** END LICENSE BLOCK *****/
 package org.jruby.ext.securerandom;
 
+import org.jruby.RubyBignum;
+import org.jruby.RubyFixnum;
+import org.jruby.RubyFloat;
+import org.jruby.RubyInteger;
+import org.jruby.RubyRange;
 import org.jruby.RubyString;
 import org.jruby.anno.JRubyMethod;
 import org.jruby.anno.JRubyModule;
@@ -35,18 +40,21 @@ import org.jruby.runtime.ThreadContext;
 import org.jruby.runtime.builtin.IRubyObject;
 import org.jruby.util.ConvertBytes;
 
+import java.math.BigInteger;
+import java.security.SecureRandom;
+
 /**
  * securerandom.rb native parts.
  */
 @JRubyModule(name = "SecureRandom")
 public class RubySecureRandom {
 
-    @JRubyMethod(meta = true)
+    @JRubyMethod(meta = true, name = "random_bytes")
     public static IRubyObject random_bytes(ThreadContext context, IRubyObject self) {
         return RubyString.newStringNoCopy(context.runtime, nextBytes(context, 16));
     }
 
-    @JRubyMethod(meta = true)
+    @JRubyMethod(meta = true, name = "random_bytes", alias = "gen_random") // gen_random for 'better' compat (not-used)
     public static IRubyObject random_bytes(ThreadContext context, IRubyObject self, IRubyObject n) {
         return RubyString.newStringNoCopy(context.runtime, nextBytes(context, n));
     }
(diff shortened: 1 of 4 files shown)


headius:
Looks good to me. How big is the diff with MRI stdlib securerandom now?

I'm guessing this is a library they might plan to gemify, so we'll probably have to address that soon. Perhaps this should be registered as an internal extension for now, so we can just `require "securerandom.jar"`?

kares:
yeah, but its confusing the stdlib repo is a few years stale and contains an old *securerandom.rb*
... which JRuby had, the new library .rb had to be dig out from the ruby repo directly, the diff is :
```diff
3,6c3,5
< begin
<   require 'openssl'
< rescue LoadError
< end
---
> 
> require 'jruby'
> org.jruby.ext.securerandom.SecureRandomLibrary.load JRuby.runtime
50,80d48
< module SecureRandom
<   if defined?(OpenSSL::Random) && /mswin|mingw/ !~ RUBY_PLATFORM
<     def self.gen_random(n)
<       @pid = 0 unless defined?(@pid)
<       pid = $$
<       unless @pid == pid
<         now = Process.clock_gettime(Process::CLOCK_REALTIME, :nanosecond)
<         ary = [now, @pid, pid]
<         OpenSSL::Random.random_add(ary.join("").to_s, 0.0)
<         seed = Random.raw_seed(16)
<         if (seed)
<           OpenSSL::Random.random_add(seed, 16)
<         end
<         @pid = pid
<       end
<       return OpenSSL::Random.random_bytes(n)
<     end
<   else
<     def self.gen_random(n)
<       ret = Random.raw_seed(n)
<       unless ret
<         raise NotImplementedError, "No random device"
<       end
<       unless ret.length == n
<         raise NotImplementedError, "Unexpected partial read from random device: only #{ret.length} for #{n} bytes"
<       end
<       ret
<     end
<   end
< end
< 
176d143
< =begin
194,219c161,162
<     if 0 < n
<       if defined? OpenSSL::BN
<         OpenSSL::BN.rand_range(n).to_i
<       else
<         hex = n.to_s(16)
<         hex = '0' + hex if (hex.length & 1) == 1
<         bin = [hex].pack("H*")
<         mask = bin[0].ord
<         mask |= mask >> 1
<         mask |= mask >> 2
<         mask |= mask >> 4
<         begin
<           rnd = random_bytes(bin.length)
<           rnd[0] = (rnd[0].ord & mask).chr
<         end until rnd < bin
<         rnd.unpack("H*")[0].hex
<       end
<     else
<       # assumption: Float::MANT_DIG <= 64
<       if defined? OpenSSL::BN
<         i64 = OpenSSL::BN.rand(64, -1).to_i
<       else
<         i64 = random_bytes(8).unpack("Q")[0]
<       end
<       Math.ldexp(i64 >> (64-Float::MANT_DIG), -Float::MANT_DIG)
<     end
---
>     # implemented natively in JRuby (just like in MRI)
>     raise NotImplementedError("#{__method__}(#{n})")
221d163
< =end
250a193,197
> # NOTE: JRuby's SecureRandom native part implement some of the methods from Random::Formatter
> # - SecureRandom.random_bytes(n=nil)
> # - SecureRandom.hex(n=nil)
> # - SecureRandom.uuid
> # - SecureRandom.random_number(n=0)
```
... is it preferable if removals are put back (JRuby had some modifications previous so I did not care much)?

headius:
I think we should consider MRI repo the canonical source for these libraries until they've officially completed gemifying each library. I just brought it up to keep in mind.

This particular library diverges so much from MRI (at least in impl) that I'm not concerned about maintaining a diff.

Issue #4655, opened 2017-06-08
Time#utc and #localtime can mutate a frozen object


iconara:
### Environment

This is one of the relevant environments, but the issue can be replicated in all environments I've tested:

```shell
$ ruby -v
jruby 9.1.8.0 (2.3.1) 2017-03-06 90fc7ab Java HotSpot(TM) 64-Bit Server VM 25.112-b16 on 1.8.0_112-b16 +jit [darwin-x86_64]
$ uname -a
Darwin quattro.local 16.6.0 Darwin Kernel Version 16.6.0: Fri Apr 14 16:21:16 PDT 2017; root:xnu-3789.60.24~6/RELEASE_X86_64 x86_64
```

### Expected Behavior

When I freeze a `Time` instance I expect to not be able to run `#utc` since that method changes the instance:

```ruby
irb> Time.new.freeze.utc
RuntimeError: can't modify frozen Time
	from (irb):1:in `utc'
	from (irb):1
```

The same goes for `#localtime`:

```ruby
irb>  Time.new.utc.freeze.localtime
RuntimeError: can't modify frozen Time
	from (irb):1:in `localtime'
	from (irb):1
```

### Actual Behavior

In all versions of JRuby I have tested (1.7.19, 1.7.23, 1.7.26, 9.0.5.0, 9.1.8.0) I can modify frozen `Time` instances:

```ruby
irb> t = Time.now
=> 2017-06-08 10:12:04 +0200
irb> t.freeze
=> 2017-06-08 10:12:04 +0200
irb> t.frozen?
=> true
irb> t.utc
=> 2017-06-08 08:12:04 UTC
irb> t
=> 2017-06-08 08:12:04 UTC
irb> t.localtime
=> 2017-06-08 10:12:04 +0200
irb> t
=> 2017-06-08 10:12:04 +0200
```

It looks like problem lies in these two lines:

* https://github.com/jruby/jruby/blob/d7c5b10a964e592bebf38beea8a0d1595cb0978f/core/src/main/java/org/jruby/RubyTime.java#L512
* https://github.com/jruby/jruby/blob/d7c5b10a964e592bebf38beea8a0d1595cb0978f/core/src/main/java/org/jruby/RubyTime.java#L528

I've looked through the rest of `RubyTime.java` and it looks like these are the only place where `dt` or `nsec` are changed except for at initialization time.

[...]

PR #4657, opened 2017-06-08
Make Time#utc and #localtime raise RuntimeError when the object is frozen

iconara:
This is a fix for #4655 that replicates the MRI behaviour of frozen `Time` objects.

It adds three tests to `spec/ruby/core/time` that cover the MRI behaviour of frozen `Time` objects – I've tried to match the style of the existing tests and copied tests for the same behaviour in `Array`. I hope it looks ok.

The actual fix is very simple, in `#utc` and #localtime` I've added a check that will throw a `RuntimeError` when the instance is frozen.

There are two additional public methods that mutate `Time` objects, but they're not Ruby methods, and I'm unsure about whether or not they should throw errors:

* [`RubyTime#updateCal(DateTime)`](https://github.com/iconara/jruby/blob/d02707acaee08a36e5b37888d2a493e059539f74/core/src/main/java/org/jruby/RubyTime.java#L456-L458)
* [`RubyTime#setDateTime(DateTime)`]( https://github.com/iconara/jruby/blob/d02707acaee08a36e5b37888d2a493e059539f74/core/src/main/java/org/jruby/RubyTime.java#L963-L965)

They also seem to be duplicates of each other, an neither has been touched in 10 years.

Should they be changed to throw the same error, or how is frozen-ness handled when objects are changed from Java code?

### DIFF
--- a/core/src/main/java/org/jruby/RubyTime.java
+++ b/core/src/main/java/org/jruby/RubyTime.java
@@ -509,6 +509,9 @@ public class RubyTime extends RubyObject {
 
     @JRubyMethod(name = {"gmtime", "utc"})
     public RubyTime gmtime() {
+        if (isFrozen()) {
+          throw getRuntime().newFrozenError("Time", true);
+        }
         dt = dt.withZone(DateTimeZone.UTC);
         return this;
     }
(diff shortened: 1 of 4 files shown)


iconara:
Even though this makes JRuby consistent with MRI, and that this obviously is a bug, it's been in all versions of JRuby, as far as I can tell, so merging this could potentially break lots of libraries and applications.

headius:
Generally, if we are doing behavior for which MRI errors, I'm not worried about fixing it. The truth is that all Ruby code is MRI code *first* and if they're not hitting the errors they're pretty much guaranteed not to have a problem with this change on JRuby. Of course there's plenty of JRuby-specific code out there but I doubt much of it is freezing and then calling mutative methods on Time objects.

I'll review and merge if @enebo has no concerns.

enebo:
Yeah I agree fully with @headius. If something​ breaks then it is one off logic. Our strength is in good Ruby compatibility over the long run

Issue #4611, opened 2017-05-17
[Q] __dir__ variable in jar file.


hiroyuki-sato:
I would like to ask about `__dir__` variable in jar archive. 

## 1. environment

* JRuby: 9.1.7.0
* OS: macOS 10.12.4
* java: 1.8.0_112

### 2.1 Expected Behavior

As the following test result, I think `__dir__` should return A or B

* A: uri:classloader:/gems/liquid-4.0.0/lib
* B: /path/to/jar_file.jar!/gems/liquid-4.0.0/lib

### 2.2 Actual Behavior

The `__dir__` returns strange path like `/path/cwd:classloader:/gems/liquid-4.0.0/lib`.
`/path/cwd` is the current working directory which I execute `java` command. 

## 3. Question 

* Is this JRuby issue or not?

## 4. Details

This [sample project](https://github.com/hiroyuki-sato/jruby__dir_test) make a single jar file with liquid-4.0.0 gem package. 
The liquid package are locate below.

```
jar tvf ./build/libs/jruby__dir_test-jruby.jar | grep liquid.rb
  3386 Wed May 17 23:56:38 JST 2017 gems/liquid-4.0.0/lib/liquid.rb
```  

When I put the following codes in `gems/liquid-4.0.0/lib/liquid.rb`
The result were below. 

```ruby
p "                 File.dirname(__FILE__): #{File.dirname(__FILE__)}"
p "File.expand_path(File.dirname(__FILE__): #{File.expand_path(File.dirname(__FILE__))}"
p "                                __dir__: #{__dir__}"
```

| program                                 | result                                                             |
|-----------------------------------------|--------------------------------------------------------------------|
| `File.dirname(__FILE__)`                  | uri:classloader:/gems/liquid-4.0.0/lib                             |
| `File.expand_path(File.dirname(__FILE__)` | uri:classloader:/gems/liquid-4.0.0/lib                             |
| `__dir__`                                 |/private/tmp/jruby__dir_test/uri:classloader:/gems/liquid-4.0.0/lib |

(`/private/tmp/jruby__dir_test` is the java execution directory)

And I also test `Dir.exist?` command.

| Dir.exist?                                                  | result| note           |
|-------------------------------------------------------------|-------|----------------|
| uri:classloader:/gems/liquid-4.0.0/lib                      | true  |                |
| /path/to/jar_file.jar!/gems/liquid-4.0.0/lib                     | true  |                |
[...]

PR #4658, opened 2017-06-09
__dir__ won't work with embed paths such as uri:classloader: (#4611)

kares:
... if `File#getAbsolutePath()` is used - which seems like its not necessary
and we can pretty much just do what MRI does `File.dirname(__FILE__)`

##

wonder if caller is actually necessary in `__dir__` // cc @headius @enebo 
- initially it was caller-less https://github.com/jruby/jruby/commit/3efc72f47bace5aad4719362fa370c46e76fe1ee#diff-73ba50d02cf75f651e0b9d581d6bd4d5R1867
- caller has been added at https://github.com/jruby/jruby/commit/5fc06a0480011a5c35339a68116dd81d88b1876d#diff-e2320b4f8ce4644f50eb844d06a5de2eR1930

as adviced in https://github.com/jruby/jruby/issues/4611#issuecomment-307215241

> @kares I think we need the same special-case logic here, or better, make this call hopefully-existing logic that does the right thing already.

... this is relying on the existing uri:classloader: handling logic in `RubyFile.dirname`

verified this works fine for liquid gem's case presented in https://github.com/jruby/jruby/issues/4611

still need to figure out a where to put a test-case to test this properly

### DIFF
--- a/core/src/main/java/org/jruby/RubyKernel.java
+++ b/core/src/main/java/org/jruby/RubyKernel.java
@@ -1781,7 +1781,8 @@ public class RubyKernel {
 
     @JRubyMethod(name = "__dir__", module = true, visibility = PRIVATE, reads = FILENAME)
     public static IRubyObject __dir__(ThreadContext context, IRubyObject recv) {
-        final String __FILE__ = context.getFile();
+        // NOTE: not using __FILE__ = context.getFile() since it won't work with JIT
+        final String __FILE__ = context.gatherCallerBacktrace()[1].getFileName();
         String dir = RubyFile.dirname(context, __FILE__);
         return RubyString.newString(context.runtime, dir);
     }
--- a/core/src/main/java/org/jruby/RubyKernel.java
+++ b/core/src/main/java/org/jruby/RubyKernel.java
@@ -1781,7 +1781,8 @@ public class RubyKernel {
 
     @JRubyMethod(name = "__dir__", module = true, visibility = PRIVATE, reads = FILENAME)
     public static IRubyObject __dir__(ThreadContext context, IRubyObject recv) {
-        String dir = RubyFile.dirname(context, new File(context.gatherCallerBacktrace()[1].getFileName()).getAbsolutePath());
+        final String __FILE__ = context.getFile();
+        String dir = RubyFile.dirname(context, __FILE__);
         return RubyString.newString(context.runtime, dir);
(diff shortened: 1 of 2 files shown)


core/src/main/java/org/jruby/RubyKernel.java
```diff
@@ -1781,7 +1781,8 @@ public static IRubyObject __method__(ThreadContext context, IRubyObject recv) {
 
     @JRubyMethod(name = "__dir__", module = true, visibility = PRIVATE, reads = FILENAME)
     public static IRubyObject __dir__(ThreadContext context, IRubyObject recv) {
-        String dir = RubyFile.dirname(context, new File(context.gatherCallerBacktrace()[1].getFileName()).getAbsolutePath());
+        final String __FILE__ = context.getFile();
+        String dir = RubyFile.dirname(context, __FILE__);
```
mkristian: looks good.

headius:
The caller logic needs to be there because jitted methods don't set context.getFile to anything. This is obviously unfortunate since it requires generating a stack trace and then mining the filename out of it.

A future improvement would be to mark "file" as a frame field needed for code that calls anything like "__dir__". It would force a frame and force setting file so that downstream calls could pick it up.

Another would be to implement __dir__ as a pseudo-keyword, not making a call if we know it's the builtin one, so it can access the filename of the current method directly. __dir__ would essentially compile down to `dir_method_builtin ? File.dirname(__FILE__) : call_dir_method`.

PR looks good to me!

headius:
Oops, I spoke too soon...I didn't realize you were removing the caller logic.

So caller logic doesn't work because our compiler doesn't embed the full uri:classloader stuff as the name, or something?

core/src/main/java/org/jruby/RubyKernel.java
headius: Caller logic is needed here for methods that have jitted (which won't update context.getFile()). The rest is ok but the filename needs to be acquired using caller logic.

kares:
ah, I see ... that's unfortunate. I'll put the `caller` gathering back than. thanks.

kares:
`caller` logic is back and I have also resurrected `test_jar_complete.rb` on [master](https://github.com/jruby/jruby/commit/08163184b7eb6f824d2bedd1522e7bdaa29606b7) so this can be tested.

PR #4673, opened 2017-06-16
Handle nil buffer in Zlib::GzipReader#readpartial

haines:
In MRI it's ok to explicitly pass a nil buffer to `#readpartial`. However, JRuby currently throws a `TypeError`.

This adds a nil-check to bring JRuby in line with MRI's behaviour.


kares:
good catch, thanks.

Issue #4665, opened 2017-06-13
Provide a maven profile to generate all flex parsers


headius:
There are a couple flex parsers already in place and one coming with #4635. We need a build path for generating all of them.

PR #4635, opened 2017-05-30
[WIP] Implement RubyDateParser in Java

muga:
This PR introduces `RubyDateParser` in Java to improve the performance for strptime. We discussed about this implementation on https://github.com/jruby/jruby/issues/4591.

PR #4674, opened 2017-06-17
Fix StrptimeLexer.flex to add a comment about how to generate the lexer 

muga:
This PR comes from #4635 and fixes StrptimeLexer.flex to add a comment for how to generate the lexer. 

Ideally we should not put those lexer java programs as @enebo and @headius discussed it on #4635 and will fix that on #4665.


core/src/main/java/org/jruby/lexer/StrptimeLexer.java
```diff
@@ -10,7 +17,7 @@
 /**
  * This class is a scanner generated by 
  * <a href="http://www.jflex.de/">JFlex</a> 1.4
- * on 5/6/17 4:40 AM from the specification file
+ * on 6/16/17 3:03 PM from the specification file
  * <tt>core/src/main/java/org/jruby/lexer/StrptimeLexer.flex</tt>
  */
 public class StrptimeLexer {
```
muga: I re-generated .java file. If we don't need the update, please let me know. I will remove it.

muga:
@kares Thank you for reviewing and commenting. Yea, expactly, the latest JFlex is 1.6.1. As you know, the current `StrftimeLexer.flex` and `StrptimeLexer.flex` work with JFlex 1.4 only (Not work with 1.4.1,.. 1.6.1). It's better that the current .flex files are ideally refactored and fixed for JFlex 1.6.1 and then, use jflex-maven-plugin to automatically generate .java files. I think that it would be worked on #4665.

muga:
I'm ignoring the CI failures because it seems that they're not related to this fix.

headius:
Thank you for your continued improvements 👍

PR #4675, opened 2017-06-18
Cache RubyDateParser object in ThreadContext to reduce the cost of object creations

muga:
To improve performance of Java strptime implemented on https://github.com/jruby/jruby/pull/4635, this PR caches a `RubyDateParser` object in `ThreadContext` to reduce the cost of `RubyDateParser` object creations. The cache is basically similar approach to `RubyDateFormatter` in `ThreadContext`. 

The following is my micro benchmark based on https://github.com/macournoyer/tinyrb. The performance by this PR is 1.5 times faster maximumly. 

```
1,000 times Date._strptime('2001-02-03', '%Y-%m-%d')
mri 2.4.1        0.000000   0.000000   0.070000 (  0.092492)
jruby 9.1.9.0    0.000000   0.000000   8.220000 (  3.861129)
jruby (master)   0.000000   0.000000   7.130000 (  3.671143)

10,000 times Date._strptime('2001-02-03', '%Y-%m-%d')
mri 2.4.1        0.000000   0.000000   0.100000 (  0.126605)
jruby 9.1.9.0    0.000000   0.000000  11.570000 (  5.183052)
jruby (master)   0.000000   0.000000   8.280000 (  3.905535)

100,000 times Date._strptime('2001-02-03', '%Y-%m-%d')
mri 2.4.1        0.000000   0.000000   0.200000 (  0.219390)
jruby 9.1.9.0    0.000000   0.000000  17.000000 ( 10.131313)
jruby (master)   0.000000   0.000000  10.620000 (  5.003686)

1,000,000 times Date._strptime('2001-02-03', '%Y-%m-%d')
mri 2.4.1        0.000000   0.000000   1.240000 (  1.361164)
jruby 9.1.9.0    0.000000   0.000000  57.490000 ( 51.720962)
jruby (master)   0.000000   0.000000  17.010000 ( 10.431316)

10,000,000 times Date._strptime('2001-02-03', '%Y-%m-%d')
mri 2.4.1        0.000000   0.000000  12.750000 ( 15.315754)
jruby 9.1.9.0    0.000000   0.000000 525.980000 (586.105106)
jruby (master)   0.000000   0.000000  66.050000 ( 65.739327)
```
The raw test data: https://gist.github.com/muga/b83262ed03009d2f765f8fd0c5319e54

### DIFF
--- a/core/src/main/java/org/jruby/runtime/ThreadContext.java
+++ b/core/src/main/java/org/jruby/runtime/ThreadContext.java
@@ -63,6 +63,7 @@ import org.jruby.runtime.profile.ProfileCollection;
 import org.jruby.runtime.scope.ManyVarsDynamicScope;
 import org.jruby.util.RecursiveComparator;
 import org.jruby.util.RubyDateFormatter;
+import org.jruby.util.RubyDateParser;
 import org.jruby.util.cli.Options;
 import org.jruby.util.log.Logger;
 import org.jruby.util.log.LoggerFactory;
@@ -107,6 +108,7 @@ public final class ThreadContext {
     private ThreadFiber rootFiber; // hard anchor for root threads' fibers
     // Cache format string because it is expensive to create on demand
     private RubyDateFormatter dateFormatter;
+    private RubyDateParser dateParser;
 
     private Frame[] frameStack = new Frame[INITIAL_FRAMES_SIZE];
     private int frameIndex = -1;
@@ -358,6 +360,12 @@ public final class ThreadContext {
         if (dateFormatter == null)
             dateFormatter = new RubyDateFormatter(this);
         return dateFormatter;
+    }
+
+    public RubyDateParser getRubyDateParser() {
+        if (dateParser == null)
+            dateParser = new RubyDateParser();
+        return dateParser;
     }
 
     public void setThread(RubyThread thread) {
--- a/lib/ruby/stdlib/date/format.rb
+++ b/lib/ruby/stdlib/date/format.rb
@@ -376,8 +376,7 @@ class Date
   private_class_method :_strptime_i
 
   def self._strptime(str, fmt='%F')
-    parser = org.jruby.util.RubyDateParser.new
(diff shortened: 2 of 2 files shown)


kares:
first we should understand why its still min 5x slower than MRI. hane nothing against this but it might be optimization done in the wrong end ...

muga:
@kares Thank you for your comment. I'm closing this PR. and will reopen it if needed.

enebo:
I have been looking at this a bit this afternoon and caching the parser 3x the current results on HEAD.  I believe the cost is object creation and initialization overhead for somethign which runs short.  Making stateless lexer/parser (e.g. single result values with inputs passed as params) or other patterns could reduce this allocation cost but at the moment it seems pretty big.

With the current state we are now down to about 6x slower so things are getting a little closer.   The caching makes us only like 40% slower but I think we can get this cost down other ways first before thinking about it.

 A lexer which used ByteList+Joni would be decent.  Another one interesting is not creating a linked list of patterns.  A smaller one is eliminating bag and populating the hash directly.  A tiny one is not re-retrieving all symbols out or a hash but making them fields somewhere.

muga:
@enebo Thank you for your comment. Make sense. I will take a profile later.

enebo:
@muga @kares I am curious on number or potential issues with f2e7b82a3f49dedec71073919d7315f35a68b737 ... I got pretty big speed jump but I was using a very trivial format string and perhaps the actual matching logic will show up being slower than MRI?

kares:
@enebo looking good, maybe I would not count on most apps being reasonable much, guess we'll see :wink:

enebo:
@muga the memory size of this cache probably scales into 10k-100k without a lot of memory.  A linked list of strings for format values seems like it should take very little memory?  Although we may end up adding a cache size in the future if anyone reports this as a leak...

muga:
@enebo Ah I see. The linked lists might not consume a lot of memory because the number of elements in one linked list is about 10 or 20 usually. Your change should be OK.

PR #4676, opened 2017-06-18
Implement key conversions with symbol by Java within Date._strptime

muga:
To improve performance of Java strptime implemented on #4635, this PR fixes `Date._strptime` to implement key conversions with symbol in Java. Because It reduces the count of objects passing between JRuby and Java. 

a RubyDateParser object in ThreadContext to reduce the cost of RubyDateParser object creations. The cache is basically similar approach to RubyDateFormatter in ThreadContext. And this PR assumes #4675.

The following is my micro benchmark based on https://github.com/macournoyer/tinyrb. The performance by this PR is 2.5 times faster maximumly.

```
1,000 times Date._strptime('2001-02-03', '%Y-%m-%d')
mri 2.4.1        0.000000   0.000000   0.070000 (  0.086790)
jruby 9.1.9.0    0.000000   0.000000   8.020000 (  3.789310)
jruby (master)   0.000000   0.000000   7.290000 (  4.043696)

10,000 times Date._strptime('2001-02-03', '%Y-%m-%d')
mri 2.4.1        0.000000   0.000000   0.090000 (  0.142577)
jruby 9.1.9.0    0.000000   0.000000  12.450000 (  7.671689)
jruby (master)   0.000000   0.000000   8.180000 (  4.401070)

100,000 times Date._strptime('2001-02-03', '%Y-%m-%d')
mri 2.4.1        0.000000   0.000000   0.200000 (  0.209632)
jruby 9.1.9.0    0.000000   0.000000  17.420000 ( 10.452343)
jruby (master)   0.000000   0.000000   9.010000 (  4.304226)

1,000,000 times Date._strptime('2001-02-03', '%Y-%m-%d')
mri 2.4.1        0.000000   0.000000   1.260000 (  1.362743)
jruby 9.1.9.0    0.000000   0.000000  60.650000 ( 59.166195)
jruby (master)   0.000000   0.000000  12.170000 (  7.638632)

10,000,000 times Date._strptime('2001-02-03', '%Y-%m-%d')
mri 2.4.1        0.000000   0.000000  11.970000 ( 12.952664)
jruby 9.1.9.0    0.000000   0.000000 482.880000 (516.068042)
jruby (master)   0.000000   0.000000  36.650000 ( 35.192336)
```
The raw test data: https://gist.github.com/muga/e566299fc9afdaa0511a8bb7804d5a7f

### DIFF
--- a/lib/ruby/stdlib/date/format.rb
+++ b/lib/ruby/stdlib/date/format.rb
@@ -376,8 +376,7 @@ class Date
   private_class_method :_strptime_i
 
   def self._strptime(str, fmt='%F')
-    parser = org.jruby.util.RubyDateParser.new
-    map = parser.parse(JRuby.runtime.current_context, fmt, str)
+    map = JRuby.runtime.current_context.getRubyDateParser.parse(JRuby.runtime.current_context, fmt, str)
     return map.nil? ? nil : map.to_hash.inject({}){|hash,(k,v)| hash[k.to_sym] = v; hash}
   end
 
--- a/lib/ruby/stdlib/date/format.rb
+++ b/lib/ruby/stdlib/date/format.rb
@@ -376,8 +376,7 @@ class Date
   private_class_method :_strptime_i
 
   def self._strptime(str, fmt='%F')
-    map = JRuby.runtime.current_context.getRubyDateParser.parse(JRuby.runtime.current_context, fmt, str)
-    return map.nil? ? nil : map.to_hash.inject({}){|hash,(k,v)| hash[k.to_sym] = v; hash}
+    return JRuby.runtime.current_context.getRubyDateParser.parseAndCreateHash(JRuby.runtime.current_context, fmt, str)
   end
 
   def self.s3e(e, y, m, d, bc=false)
--- a/core/src/main/java/org/jruby/runtime/ThreadContext.java
+++ b/core/src/main/java/org/jruby/runtime/ThreadContext.java
@@ -63,6 +63,7 @@ import org.jruby.runtime.profile.ProfileCollection;
 import org.jruby.runtime.scope.ManyVarsDynamicScope;
 import org.jruby.util.RecursiveComparator;
 import org.jruby.util.RubyDateFormatter;
+import org.jruby.util.RubyDateParser;
 import org.jruby.util.cli.Options;
 import org.jruby.util.log.Logger;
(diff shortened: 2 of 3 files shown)


lib/ruby/stdlib/date/format.rb
```diff
@@ -376,9 +376,7 @@ def self._strptime_i(str, fmt, e) # :nodoc:
   private_class_method :_strptime_i
 
   def self._strptime(str, fmt='%F')
-    parser = org.jruby.util.RubyDateParser.new
-    map = parser.parse(JRuby.runtime.current_context, fmt, str)
-    return map.nil? ? nil : map.to_hash.inject({}){|hash,(k,v)| hash[k.to_sym] = v; hash}
+    return JRuby.runtime.current_context.getRubyDateParser.parseAndCreateHash(JRuby.runtime.current_context, fmt, str)
```
kares: quite an ugly design to get current_context to pass it current_context. maybe this whole method should be/delegate to a native ext impl.

enebo: @muga I think you could make parser.parse just return the symboled hash right?  Then @kares comment about currentContext (and having to know there is a cache) can be encapsulated by the parse.

As a broader comment.  A Bag is made then a Map is made and then a Hash with Symbols is made.  I have not viewed how much a Bag modifies its data but I wondered if you could just drop the bag and the map and just use a Hash with symbols directly as you parse?  I can see some post processing happening at end of bag and when map is made but I just wondered about whether this could just be one thing or not.

enebo: @muga @kares oh I misread this a bit...I agree with @kares that if we had this as an Ext the ```def self._strptime``` would just be an @JRubyMethod() in Java.

enebo: @muga @kares I am implementing _strptime as native ext since I think I can do this a faster (in fact I think I am already done).  This will invalidate this PR but I will point out which commit it is in.

enebo:
We can consider caching part of this in the other PR or in a new one but I implemented the other bits and pushed the result in 4323f93.  You can read that commit for the improvement.  @muga this duplicated your strategy but just differently using a native ext as @kares suggested.

muga:
@kares @enebo Thank you for suggesting and fixing. I totally agree with your thoughts and 4323f93.

PR #4684, opened 2017-06-23
Rename readmsg to recvmsg

ioquatix:
* Fix inconsistency of naming w.r.t. MRI.

Uh, I didn't even look at the code originally, but it seems like it's not even implemented. At least we can fix the name.. but can we fix the implementation too?

### DIFF
--- a/core/src/main/java/org/jruby/ext/socket/RubyBasicSocket.java
+++ b/core/src/main/java/org/jruby/ext/socket/RubyBasicSocket.java
@@ -473,13 +473,13 @@ public class RubyBasicSocket extends RubyIO {
     }
 
     @JRubyMethod(rest = true, notImplemented = true)
-    public IRubyObject readmsg(ThreadContext context, IRubyObject[] args) {
-        throw context.runtime.newNotImplementedError("readmsg is not implemented");
+    public IRubyObject recvmsg(ThreadContext context, IRubyObject[] args) {
+        throw context.runtime.newNotImplementedError("recvmsg is not implemented");
     }
 
     @JRubyMethod(rest = true, notImplemented = true)
-    public IRubyObject readmsg_nonblock(ThreadContext context, IRubyObject[] args) {
-        throw context.runtime.newNotImplementedError("readmsg_nonblock is not implemented");
+    public IRubyObject recvmsg_nonblock(ThreadContext context, IRubyObject[] args) {
+        throw context.runtime.newNotImplementedError("recvmsg_nonblock is not implemented");
     }
 
     protected ByteList doRead(ThreadContext context, final ByteBuffer buffer) {


ioquatix:
This bit of functionality is critical for RubyDNS which.. uses UDP sockets and requires them to work consistently, unlike UDPSocket#recvmsg which is a bit of an oddball.

headius:
I have to admit I'm not familiar with the "recv" set of socket functions. What we have implemented today is largely just using "read" and simulating other parameters to "recv" and "recvfrom".

The main challenge is mimicking these lower-level socket APIs on the JVM, where we have only high-level APIs that hide access to the actual file descriptors or the lower-level operations. We've managed to fake it well enough for the standard socket types, but when you get to more unusual sockets we still struggle a bit.

So, to clarify a few things.

* recv/recvfrom/recvmsg are used mostly with bare Ruby "Socket" instances or other connection-free sockets like UDP and UNIX, is that generally true? In other words, is it common to see someone using these functions with a TCPSocket?
* It's unclear to me what the difference is between a read and a recv, except perhaps that the recv works on connectionless sockets?

Supporting the general case of recv on all types of sockets may not be doable with Java's APIs, but a specific goal of getting enough of Socket#recvmsg to work for DNS should be possible.

ioquatix:
> I have to admit I'm not familiar with the "recv" set of socket functions. What we have implemented today is largely just using "read" and simulating other parameters to "recv" and "recvfrom".

`read` is a more generic operation on a file descriptor than `recv` or `recvfrom` which are specific to sockets. Generally speaking you can use read on a socket but you get less information, e.g. `recvfrom` will also give you the address the packet came from so you can reply.

> The main challenge is mimicking these lower-level socket APIs on the JVM, where we have only high-level APIs that hide access to the actual file descriptors or the lower-level operations. We've managed to fake it well enough for the standard socket types, but when you get to more unusual sockets we still struggle a bit.

Java has `DatagramSocket` and `DatagramPacket` which should be sufficient to implement this API.

> recv/recvfrom/recvmsg are used mostly with bare Ruby "Socket" instances or other connection-free sockets like UDP and UNIX, is that generally true? In other words, is it common to see someone using these functions with a TCPSocket?

You will normally use `sendto` on UDP socket in order to specify the destination. Similarly, you would normally use `recvfrom` to know where the UDP data was received from.

UDP sockets can be connected.

> It's unclear to me what the difference is between a read and a recv, except perhaps that the recv works on connectionless sockets?

https://stackoverflow.com/questions/1790750/what-is-the-difference-between-read-and-recv-and-between-send-and-write

> Supporting the general case of recv on all types of sockets may not be doable with Java's APIs, but a specific goal of getting enough of Socket#recvmsg to work for DNS should be possible.

That would be great.

ioquatix:
Bump.

ioquatix:
So, I assume the API you need to use in Java is `DatagramSocket`? Which probably means that it's even more likely you'll need to re-implement `Socket` to multiplex between `ServerSocket`, `ClientSocket`, and `DatagramSocket`, or depend on some other API.

https://github.com/jruby/jruby/issues/4748

ioquatix:
Bump :)

headius:
Sorry, this slipped through for too long. Marking for 9.2.1.

ioquatix:
That would be awesome.

BanzaiMan:
Could this be updated to the current code base? Thank you!

Issue #4686, opened 2017-06-26
Cannot rescue from LocalJumpError after wrongly using break in proc


ivoanjo:
### Environment

* JRuby: `jruby 9.1.12.0 (2.3.3) 2017-06-15 33c6439 Java HotSpot(TM) 64-Bit Server VM 25.131-b11 on 1.8.0_131-b11 [linux-x86_64]`
* Kernel: `Linux maruchan 4.11.0-041100-generic #201705041534 SMP Thu May 4 19:36:05 UTC 2017 x86_64 x86_64 x86_64 GNU/Linux`
* Distro: `Ubuntu 17.04`

### Expected Behavior

The following test case causes a `LocalJumpError` exception to be raised, and then properly handles it.

```ruby
block_with_break = proc { break :foo }

begin
  block_with_break.call
rescue Exception => e
  puts "Handled #{e} (#{e.class}) successfully!"
end

puts 'done'
```

On MRI this works as expected:

```
$ ruby -v
ruby 2.3.3p222 (2016-11-21 revision 56859) [x86_64-linux]
$ ruby test.rb 
Handled break from proc-closure (LocalJumpError) successfully!
done
```

### Actual Behavior

Under JRuby, the code does not seem to be able to catch the `LocalJumpError`:

```
$ jruby -v
jruby 9.1.12.0 (2.3.3) 2017-06-15 33c6439 Java HotSpot(TM) 64-Bit Server VM 25.131-b11 on 1.8.0_131-b11 +jit [linux-x86_64]
$ jruby test.rb 
LocalJumpError: unexpected break
(vm exits)
```

The test case is pretty strange; the real use case where I've bumped into this issue is an app that took a block and created a future from it with `concurrent-ruby`. On a rare code path it would call an erroneous `break`, and because `concurrent-ruby` could not `rescue` and mark the background task as failed, the task would appear like it would never finish.

headius:
A similar or nearly-identical case was found in ruby/spec and reported by @eregon in #4577 but we had not fixed it because of the lack of any real-world use case. Because the erroneous break should be an error in any case, it had not been a real concern to create the Ruby exception a bit lazily.

However, your case shows that there might be third-party code that doesn't behave, a code path that wasn't tested, or a piece of code that worked ok in one context but does not in another. In these cases, not propagating LJE as soon as possible makes it difficult to perform appropriate cleanup.

[...]

Issue #4577, opened 2017-04-28
LocalJumpError: unexpected break in spec/ruby/language/break_spec.rb


eregon:
jruby 9.1.9.0-SNAPSHOT (2.3.3) 2017-04-28 038f790 OpenJDK 64-Bit Server VM 25.121-b14 on 1.8.0_121-b14 +jit [linux-x86_64]

Which can be reproduced with:
```
 bin/jruby spec/mspec/bin/mspec run -Gfails -Gcritical -Gslow spec/ruby/language/break_spec.rb
```

This specific spec is currently excluded in jruby.2.3.mspec so that's why it got unnoticed.
There are only 3 `ruby_exe` calls in that spec so once this is fixed I would advise to include it on CI.

eregon:
@headius @enebo Could you have a look at this?

headius:
Would be nice to get this fixed and the spec running.

headius:
So this is another case where we are propagating the break out like we should, but we don't produce a rescuable LJE until it's too late. So there's some missing logic at the closure boundary (a proc in this case) to know that the break has no target and the LJE needs to start propagating earlier.

eregon:
The full backtrace is
```
1)
An exception occurred during: loading /home/eregon/code/jruby/spec/ruby/language/break_spec.rb
The break statement in a captured block when the invocation of the scope creating the block is still active raises a LocalJumpError when invoking the block from the scope creating the block ERROR
LocalJumpError: unexpected break
org/jruby/RubyKernel.java:979:in `load'
/home/eregon/code/jruby/spec/mspec/lib/mspec/runner/mspec.rb:1:in `block in (root)'
org/jruby/RubyBasicObject.java:1691:in `instance_eval'
/home/eregon/code/jruby/spec/mspec/lib/mspec/runner/mspec.rb:81:in `files'
/home/eregon/code/jruby/spec/mspec/lib/mspec/runner/mspec.rb:98:in `block in protect'
org/jruby/RubyArray.java:1734:in `each'
/home/eregon/code/jruby/spec/mspec/lib/mspec/runner/mspec.rb:81:in `files'
/home/eregon/code/jruby/spec/mspec/lib/mspec/runner/mspec.rb:72:in `each_file'
/home/eregon/code/jruby/spec/mspec/lib/mspec/runner/mspec.rb:77:in `files'
/home/eregon/code/jruby/spec/mspec/lib/mspec/runner/mspec.rb:48:in `process'
[...]

PR #4694, opened 2017-06-28
Fix break not turning into LocalJumpError soon enough

headius:
For #4686 and #4577.

The strategy here is as it was in 1.7: mark literal blocks as "escaped" when the call they were originally passed to returns. Break operations can then immediately trigger a LocalJumpError and be rescued in all the various ways reported in #4686 and #4577.

This does not fix super calls because the additional work is nontrivial (pushing literal-closure param through a lot of code in interpreter and JIT) and the exposure surface is pretty small for super PLUS literal closure PLUS break PLUS escaping of that block.

### DIFF
--- a/core/src/main/java/org/jruby/ir/interpreter/InterpreterEngine.java
+++ b/core/src/main/java/org/jruby/ir/interpreter/InterpreterEngine.java
@@ -419,7 +419,6 @@ public class InterpreterEngine {
     protected static IRubyObject processReturnOp(ThreadContext context, Block block, Instr instr, Operation operation,
                                                  DynamicScope currDynScope, Object[] temp, IRubyObject self,
                                                  StaticScope currScope) {
-        Block.Type blockType = block == null ? null : block.type;
         switch(operation) {
             // --------- Return flavored instructions --------
             case RETURN: {
@@ -433,12 +432,12 @@ public class InterpreterEngine {
                 // This assumes that scopes with break instr. have a frame / dynamic scope
                 // pushed so that we can get to its static scope. For-loops now always have
                 // a dyn-scope pushed onto stack which makes this work in all scenarios.
-                return IRRuntimeHelpers.initiateBreak(context, currDynScope, rv, block, blockType);
+                return IRRuntimeHelpers.initiateBreak(context, currDynScope, rv, block);
             }
             case NONLOCAL_RETURN: {
                 NonlocalReturnInstr ri = (NonlocalReturnInstr)instr;
                 IRubyObject rv = (IRubyObject)retrieveOp(ri.getReturnValue(), context, self, currDynScope, currScope, temp);
-                return IRRuntimeHelpers.initiateNonLocalReturn(context, currDynScope, blockType, rv);
+                return IRRuntimeHelpers.initiateNonLocalReturn(context, currDynScope, block, rv);
             }
             case RETURN_OR_RETHROW_SAVED_EXC: {
                 IRubyObject retVal = (IRubyObject) retrieveOp(((ReturnBase) instr).getReturnValue(), context, self, currDynScope, currScope, temp);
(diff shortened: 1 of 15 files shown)


core/src/main/java/org/jruby/ir/interpreter/InterpreterEngine.java
```diff
@@ -433,7 +433,7 @@ protected static IRubyObject processReturnOp(ThreadContext context, Block block,
                 // This assumes that scopes with break instr. have a frame / dynamic scope
                 // pushed so that we can get to its static scope. For-loops now always have
                 // a dyn-scope pushed onto stack which makes this work in all scenarios.
-                return IRRuntimeHelpers.initiateBreak(context, currDynScope, rv, blockType);
+                return IRRuntimeHelpers.initiateBreak(context, currDynScope, rv, block, blockType);
```
enebo: Looking at this code I wonder if we shouldn't just pass block.  I see also we pass blockType for nonlocal returns but at top of this method we endlessly do block == null check and we may as well push that into initiateBreak and handleNonlocalReturn.

core/src/main/java/org/jruby/ir/runtime/IRRuntimeHelpers.java
```diff
@@ -159,15 +159,21 @@ private static IRScopeType ensureScopeIsClosure(ThreadContext context, DynamicSc
...
+    public static IRubyObject initiateBreak(ThreadContext context, DynamicScope dynScope, IRubyObject breakValue, Block block, Block.Type blockType) throws RuntimeException {
         // Wrap the return value in an exception object and push it through the break exception
         // paths so that ensures are run, frames/scopes are popped from runtime stacks, etc.
         if (inLambda(blockType)) throw new IRWrappedLambdaReturnValue(breakValue);
 
         IRScopeType scopeType = ensureScopeIsClosure(context, dynScope);
 
+        DynamicScope parentScope = dynScope.getParentScope();
+
+        if (block.isEscaped()) {
```
enebo: Based on my previous comment I now wonder if block can be null here ever?  Perhaps this is only called in cases where block is guaranteed non-null?

core/src/main/java/org/jruby/ir/targets/IRBytecodeAdapter6.java
```diff
@@ -589,9 +589,9 @@ public void invokeOtherOneFloat(String file, int line, String name, double flote
         if (!MethodIndex.hasFastFloatOps(name)) {
             pushFloat(flote);
             if (callType == CallType.NORMAL) {
-                invokeOther(file, line, name, 1, false, false);
+                invokeOther(file, line, name, 1, false, false, false);
```
enebo: Not a review comment but just a lament that Java has no keyword arguments :P

headius: Yeah I don't stacking booleans like this. I might combine the two into an enum representing what exactly was passed for block argument.

core/src/main/java/org/jruby/ir/interpreter/InterpreterEngine.java
headius: Turns out blockType was only used by the two cases that always have a non-null block (as we discussed on IRC, non-local return and break *should* only be emitted for block bodies) so I just removed that null check and made each method look at block.type directly. Addresses this and other comment.

core/src/main/java/org/jruby/ir/targets/IRBytecodeAdapter6.java
enebo: Hmmm I wonder how many people static final booleans for this sort of issue?  I don't think I have ever seen that done but it is probably a pattern?

headius:
I've made all suggested tweaks and I'm happy with it.

Note that I had to re-add a null check in IRRuntimeHelpers.initiateNonLocalReturn because it appears that that instruction is also used to return from a singleton class body. I saw an error in the JRuby suite, running without that null check, that looked like this:

```
[exec] Error: test_sclass_return(TestClass): Java::JavaLang::NullPointerException:
[exec] org.jruby.ir.runtime.IRRuntimeHelpers.initiateNonLocalReturn(IRRuntimeHelpers.java:99)
```

We may want to see if this really needs to be a non-local return (@enebo).

I will open a separate issue for the known-unfixed super behavior.

PR #4698, opened 2017-06-29
Added URI::Generic#to_java to convert to java.net.URI

jkutner:
This cleans up direct use of `java.net.URI` usage in Ruby code. For example, when working with a Java `http_client` like this:

```ruby
uri = java.net.URI.new(rest_url(item))
http_client = ctx.get(HttpClient.java_class)
http_client.get(uri)
```

We can now use a Ruby `URI` like this:

```ruby
uri = URI(rest_url(item))
http_client = ctx.get(HttpClient.java_class)
http_client.get(uri.to_java)
```

Using a Ruby `URI` allows for better reuse of the object throughout the code.

If there is a way to make the `uri.to_java` call implicit, I'll be happy to work on that too.


kares:
:+1: patch however it has at least 1 problematic part : 
`spec/ruby` is shared among Ruby impls so it should be moved elsewhere (e.g. spec/java_integration)

the other would be that stdlib files now contain JRuby specific ext - which is kind of a first but than since its  an ext of a .rb library there's no other easy way to not pollute stdlib with java ext pieces (which usually live elsewhere)

jkutner:
@kares i don't like mucking with the stdlib this way either. Maybe the `java.rb` can become a "a pattern" so that it's not a one-off case. Otherwise, maybe there we could add special case logic to the base implementation of `to_java` (which I think is in [`KernelJavaAddons.java`](https://github.com/jruby/jruby/blob/master/core/src/main/java/org/jruby/java/addons/KernelJavaAddons.java#L12-L19)) to check for a URI object.

But if everyone is ok with the `java.rb` I'll move the test to `spec/java_integration`.

enebo:
we need a mechanism for adding our own definitions to the stdlib without adding them to the stdlib file itself.  Unfortunately I almost feel like this would be an exercise in AOP.   I can think our overrides being further up load path which would then somehow require the stdlib file.  This would allow cutpoint defs before or defs after the real stdlib file is required.   Without a special require in our required file we could maybe use require_relative?  Bleh.  

If we had a generic mechanism for this then it would be really easy to patch stdlib in cases where that code makes no sense going upstream to MRI.

headius:
We're going to need to cross the bridge of stdlib-sharing sooner or later, now that 2.5 is gemifying everything. People will start depending on newer versions of stdlib gems and pulling whatever's released, so they'll lose JRuby-specific bits unless we come up with an answer.

The typical way to do this, I think, is to put the impl-specific bits in separate gems, so that readline would depend on readline-mri when installing on mri and readline-jruby when installing on JRuby. But this is a complex issue, since it means readline itself would need a -java release for different dependencies on JRuby.

The other option would be for the readline gem to contain both what JRuby needs and what MRI needs. That will work for some libraries that are pure Ruby, but for libraries with exts we'll still need two separate gems.

PR #4702, opened 2017-07-02
Install HAVEGED for improving entropy source

BanzaiMan:


eregon:
Thanks for the PR! Unfortunately it does not seem to solve the problem.
Log from the travis-ci/pr job: https://travis-ci.org/jruby/jruby/jobs/249340946

BanzaiMan:
Well, these jobs are running on containers, I believe, so maybe HAVEGED is not enough to fill the entropy.

kares:
seems that low entropy haven't been an issue for a while now - good job TravisCI team :)

PR #4703, opened 2017-07-02
Move to Trusty

BanzaiMan:


eregon:
Interestingly on Trusty, both spec:ruby:fast run finished and did not get stuck on the OpenSSL random_bytes specs. (Instead they fail due to actual missing features, as expected.
This might be a way to fix #4696 then.

kares:
this is looking good - run times are certainly better but specs started failing as I understand @eregon this is good - as otherwise we get hangs? have not seen any of those failures anywhere except here ...

@BanzaiMan what about the new failure here: https://travis-ci.org/jruby/jruby/jobs/249341412 JAVA_HOME?

kares:
@eregon ignore that - did not see there's already another ruby/spec import (have tagged the previous one)

eregon:
@kares Tags added in aa19a4f4a98be0a126f01d80efc519eeaa02c444.

BanzaiMan:
Oh, right. I think Oracle JDK 7 is not available on Trusty. :-/

kares:
oh well, that's a blocker :sob:

kares:
since 9.2 decided to drop support for Java 7 I think this is fine to merge ...

headius:
Agreed.

kares:
on master https://github.com/jruby/jruby/commit/7ffb8fd4f69e1b3421465f54b7d9958a5748ab86 with https://github.com/jruby/jruby/commit/18c278e9383ff5593776cec50d554848eb33dd8e CI has gotten greener, yay!

Issue #4688, opened 2017-06-26
RubyEnumerator.SizeFn does not receive enough context


headius:
In order to implement enumerators with sizes, we have a SizeFn interface we implement for various types of collection and enumeration. However this interface currently only receives `IRubyObject[]` which means the runtime, context, self, and other contextual values must be captured (usually as an anonymous inner class) or acquired (e.g. `runtime.getCurrentContext()`).

This interface should be expanded to accept at least context and self, and all uses should be audited to see if we can make the impls static instances rather than creating them anew each time.

anukin:
@headius I have raised a PR here https://github.com/jruby/jruby/pull/4714. Can you take a look ?

headius:
@anukin I have reworked your original PR and will finish the remaining requirements of this issue as part of #6321. Thank you for your help, even though it got left behind.

anukin:
Hey thanks :)

edipofederle:
@headius could you close this issue?

headius:
@edipofederle Yeah I guess this should have gone along with #6321. Thanks for the audit!

PR #4714, opened 2017-07-12
[4688] add context and self to SizeFn

anukin:
@headius Can u take a look at this and suggest some changes if necessary? It's a solution for https://github.com/jruby/jruby/issues/4688


core/src/main/java/org/jruby/RubyArray.java
```diff
@@ -1226,7 +1226,7 @@ protected SizeFn enumLengthFn() {
         final RubyArray self = this;
         return new SizeFn() {
             @Override
-            public IRubyObject size(IRubyObject[] args) {
+            public IRubyObject size(IRubyObject[] args, ThreadContext context, IRubyObject self1) {
```
kares: think it would be better to follow the 'general' order convention: `(context, self, args)`

core/src/main/java/org/jruby/RubyArray.java
```diff
@@ -3793,7 +3793,7 @@ private SizeFn combinationSize(final ThreadContext context) {
...
```
kares: `self1` maybe just `self` ...

core/src/main/java/org/jruby/RubyEnumerable.java
```diff
@@ -32,22 +32,10 @@
...
-import org.jruby.runtime.Arity;
-import org.jruby.runtime.Block;
-import org.jruby.runtime.BlockBody;
-import org.jruby.runtime.BlockCallback;
-import org.jruby.runtime.CallBlock;
-import org.jruby.runtime.CallBlock19;
-import org.jruby.runtime.Helpers;
-import org.jruby.runtime.JavaInternalBlockBody;
-import org.jruby.runtime.JavaSites;
+import org.jruby.runtime.*;
```
kares: :scissors: I hate when IDEA does that ... already tried forcing it not to do so twice :disappointed:

core/src/main/java/org/jruby/RubyArray.java
anukin: Thank you ! I will follow the same

core/src/main/java/org/jruby/RubyEnumerable.java
anukin: I will remove this

core/src/main/java/org/jruby/RubyArray.java
anukin: Sure

anukin:
@kares I will implement the changes as per your review. Thank you very much.

anukin:
@kares I am done with the changes. Now can u take a look ? Ty

core/src/main/java/org/jruby/RubyEnumerator.java
```diff
@@ -473,8 +473,8 @@ private SizeFn enumSizeFn(final ThreadContext context) {
         final RubyEnumerator self = this;
         return new SizeFn() {
             @Override
-            public IRubyObject size(IRubyObject[] args) {
-                return self.size(context);
+            public IRubyObject size(ThreadContext context, IRubyObject self, IRubyObject[] args) {
+                return this.size(context, self, args);
```
kares: will recurse infinitely and thus `fail` ... I see why you had `self1` now - makes sense to have a different name here
if I am reading this right? could have used a different name e.g. `recv` if it conflicts (doesn't matter much though) ...

anukin: Yeah that's why I had to use `self1`. `recv` makes it much better i will go ahead with that then :)

anukin:
@kares This is done as per review. Do take a look now.

anukin:
@kares I have renamed all instances of `self1` to `recv`. Should be good now :)

enebo:
@anukin we typically use recv for JRubyMethod which are for modules but the intent is very clear here and it is a good analogue to self in this case.  self1 bugged me and no doubt was why @kares originally complained.

core/src/main/java/org/jruby/RubyEnumerator.java
```diff
@@ -473,8 +473,8 @@ private SizeFn enumSizeFn(final ThreadContext context) {
         final RubyEnumerator self = this;
         return new SizeFn() {
             @Override
-            public IRubyObject size(IRubyObject[] args) {
-                return self.size(context);
+            public IRubyObject size(ThreadContext context, IRubyObject recv, IRubyObject[] args) {
+                return this.size(context, recv, args);
```
kares: this is still doing smt else as before, isn't it? (calling this's method instead of `RubyEnumerator#size`)

anukin:
@kares Sorry I missed that one. It's taken care of now. Sorry for wasting your time.
@enebo self1 was bugging me as well. But i had seen some context1 in the codebase. So i was thinking self1 might be the best alternative. Seems there is a better one. Ty for clarifying that :)

This PR should be good now.

core/src/main/java/org/jruby/RubyEnumerable.java
```diff
@@ -1179,15 +1177,15 @@ public IRubyObject call(ThreadContext ctx, IRubyObject[] largs, Block blk) {
...
+            public IRubyObject size(ThreadContext context1, IRubyObject self, IRubyObject[] args) {
                 Ruby runtime = context.runtime;
                 assert args != null && args.length > 0 && args[0] instanceof RubyNumeric; // #each_cons ensures arg[0] is numeric
                 long consSize = ((RubyNumeric) args[0]).getLongValue();
                 if (consSize <= 0) {
                     throw runtime.newArgumentError("invalid size");
                 }
 
-                IRubyObject size = enumSizeFn(context, self).size(args);
+                IRubyObject size = enumSizeFn(context, self).size(context1, self, args);
```
kares: :scissors: this is causing the StackOverflow failure on travis-ci

kares:
thanks, the StackOverflow regression is still there: https://travis-ci.org/jruby/jruby/jobs/255126453 (view "Raw Log" and scroll to the end). but we can fix that for you later once we're actually looking at the code from IDE.

enebo:
@anukin Sorry I think I probably messed something up when I tried to update a merge conflict.  Can you update this PR?

headius:
Part of this – the addition of ThreadContext to the SizeFn interface – was done in #5958. The remaining changes from this PR have been reworked and remerged with master in #6321. Those changes alone are not sufficient to resolve #4688 but at least bring the "self" object through so that the remaining work can happen (as described in #6321).

Original changeset retains @anukin attribution. 👍

PR #4718, opened 2017-07-17
[ji] (auto) convert java numbers

kares:
Java Integration does not fully handle Java's `Number` type system (primitives as well as java.math classes).

mostly Java numeric values are converted back and forth as needed, however there's still a fay to explicitly instanciate Java types (`java.lang.Integer.new(10)` etc).

improvements presented here try to fix those Java value proxies to seamlessly work with Ruby's conversion rules (e.g. `to_int`). there has been requests on these already, added test demonstrate what's now possible.

### DIFF
--- a/core/src/main/java/org/jruby/RubyRational.java
+++ b/core/src/main/java/org/jruby/RubyRational.java
@@ -808,10 +808,8 @@ public class RubyRational extends RubyNumeric {
         return f_negate(context, this);
     }
 
-    private IRubyObject op_roundCommonPre(ThreadContext context, IRubyObject n) {
-        checkInteger(context, n);
-        Ruby runtime = context.runtime;
-        return f_expt(context, RubyFixnum.newFixnum(runtime, 10), n);
+    private static IRubyObject op_roundCommonPre(ThreadContext context, IRubyObject n) {
+        return f_expt(context, RubyFixnum.newFixnum(context.runtime, 10), RubyInteger.toInteger(context, n));
     }
 
     private IRubyObject op_roundCommonPost(ThreadContext context, IRubyObject s, IRubyObject n, IRubyObject b) {
@@ -1043,8 +1041,8 @@ public class RubyRational extends RubyNumeric {
     @JRubyMethod(name = "marshal_load")
     public IRubyObject marshal_load(ThreadContext context, IRubyObject arg) {
         RubyArray load = arg.convertToArray();
-        num = load.size() > 0 ? load.eltInternal(0) : context.runtime.getNil();
-        den = load.size() > 1 ? load.eltInternal(1) : context.runtime.getNil();
+        num = load.size() > 0 ? load.eltInternal(0) : context.nil;
+        den = load.size() > 1 ? load.eltInternal(1) : context.nil;
 
         if (f_zero_p(context, den)) {
             throw context.runtime.newZeroDivisionError();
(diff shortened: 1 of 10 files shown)


kares:
@headius am happy to move off and merge this to **ruby-2.4** (expect some conflicts with your recent work) 
continued the experiments against master since wanted a fully :white_check_mark: from test and CI seems to still fail for 2.4

core/src/main/java/org/jruby/util/Pack.java
```diff
@@ -83,23 +83,32 @@
...
+    // MRI: rb_to_float 1.9
+    private static RubyFloat toFloat(Ruby runtime, IRubyObject obj) {
+        if (obj instanceof RubyNumeric) {
+            return ((RubyNumeric) obj).convertToFloat();
+        }
+        if (obj instanceof RubyString || obj.isNil()) {
+            throw runtime.newTypeError(obj, "Float");
+        }
+        return (RubyFloat) TypeConverter.convertToType(obj, runtime.getFloat(), "to_f", true);
+    }
```
headius: The old TypeConverter.toFloat appears to only have been used by Pack. Perhaps we can just replace it with the newer impl rather than having two around? I see in MRI it is used in pack.c, but also in random.c. rb_to_float itself is in object.c.

core/src/main/java/org/jruby/RubyRational.java
```diff
@@ -590,6 +578,12 @@ public IRubyObject op_div(ThreadContext context, IRubyObject other) {
     /** nurat_fdiv
      * 
      */
+//    @Override
+//    @JRubyMethod(name = "fdiv")
+//    public IRubyObject fdiv(ThreadContext context, IRubyObject other) {
+//        return f_div(context, f_to_f(context, this), other);
+//    }
+
```
headius: I fixed some of these names on the 2.4 branch, so there may be some minor conflicts. No need to keep this commented code in any case.

core/src/main/java/org/jruby/javasupport/ext/JavaLang.java
```diff
@@ -340,6 +344,26 @@ public static IRubyObject integer_p(final ThreadContext context, final IRubyObje
...
+        @JRubyMethod(name = "coerce")
+        public static IRubyObject coerce(final ThreadContext context, final IRubyObject self, final IRubyObject type) {
+            java.lang.Number val = (java.lang.Number) self.toJava(java.lang.Number.class);
+
+            // NOTE: a basic stub that always coverts Java numbers to Ruby ones (for simplicity)
+            // gist being this is not expected to be used heavily, if so should get special care
+            final IRubyObject value;
+            if (val instanceof java.math.BigDecimal) {
+                final RubyClass klass = context.runtime.getClass("BigDecimal");
+                if (klass == null) { // user should require 'bigdecimal'
```
headius: Perhaps we should autoload? I was just thinking yesterday it's odd that BigDecimal is the last numeric type not always loaded. Maybe we should point that out to ruby-core as well?

core/src/main/java/org/jruby/util/Pack.java
kares: OK thanks, will look into it ...

core/src/main/java/org/jruby/javasupport/ext/JavaLang.java
kares: guess its not that common to use - like in Java, `BigInteger` is probably used more than BigDecimal.
... but you're right if all others are core maybe this shouldn't be a lib.
was also thinking of `BigDecimal` auto-converting Java<->Ruby, however more tricky the library isn't loaded.

headius: It seems like many more people would use BigDecimal than Complex to me...I'll ping MRI folks about it.

kares:
should now be :green_heart: (based on feedback) ... will pbly wait for the ruby-2.4 merge to resolve potential conflicts

headius:
The 2.4 merge will happen tomorrow (Friday 4 Aug).

kares:
merged onto master manually: https://github.com/jruby/jruby/compare/cd080915bc...2a5b495ce4

PR #4726, opened 2017-07-29
upgrade jruby-openssl to 0.9.21

kares:
this updates the bundler jruby-openssl gem


kares:
changing target release as jossl has now been updated for 9.1.13.0 to **0.9.21**

PR #4753, opened 2017-08-27
MINOR: Make CachingCallSite more JIT friendly

original-brownbear:
Some improvements I found when profiling the red black benchmark ( + some trivialities along the way :)):

* split `org.jruby.RubyModule#searchWithCache(java.lang.String, boolean)` into hot and cold path methods to decrease the size of its caller `org.jruby.runtime.callsite.CachingCallSite#cacheAndCall(org.jruby.runtime.builtin.IRubyObject, org.jruby.RubyClass, org.jruby.runtime.Block, org.jruby.runtime.ThreadContext, org.jruby.runtime.builtin.IRubyObject, org.jruby.runtime.builtin.IRubyObject, org.jruby.runtime.builtin.IRubyObject, org.jruby.runtime.builtin.IRubyObject)`
  * Also removed redundant getter use in it
* Reduced the indirection in `org.jruby.runtime.callsite.CacheEntry#typeOk` and removed its `static` version (the `static` version really made no sense as it required having an instance to feed into it anyhow + the instance version compiles a little smaller)
* Removed the redundant setter `updateCache(entry);` to save a byte
* Made `boolean cacheAndGetBuiltin(RubyClass selfType, String methodName)` `static` since it contains no reference to `this`

Trivial:

* Added missing `@override`, where missing to improve readability
* Made a few methods around the `CachingCallingSite` `final` to improve readability (I make JIT converge to the final state a little faster I guess :D)
* Made a few methods `private` that were needlessly `protected`

Results:

* Redblack benchmark becomes `2.2%` faster for me, measured over 1k iterations
* Codecache size for the red black benchmark converges to `6.5M` instead of `6.8M`


core/src/main/java/org/jruby/runtime/callsite/CacheEntry.java
```diff
@@ -15,11 +15,7 @@ public CacheEntry(DynamicMethod method, int token) {
     }
 
     public final boolean typeOk(RubyClass incomingType) {
-        return typeOk(this, incomingType);
-    }
-
-    public static final boolean typeOk(CacheEntry entry, RubyClass incomingType) {
```
kares: you need to careful when removing methods on places such as these.
might be called from JIT-ed byte-code (so you're not seeing it show up as used)

original-brownbear: @kares yea, I learnt that today in some other spot already :) Questions (if you have a sec):

1. Can I do anything besides running tests + full-text search to make sure I don't kill such a method accidentally?
2. Am I correct to understand that the `@JIT` annotation should be put on methods which could be called dynamically like you described?

Thanks!

kares: >    Can I do anything besides running tests + full-text search to make sure I don't kill such a method accidentally?

usually just grep for the method-name string -> `"typeOf"` should be safe to find whether its used :)

>    Am I correct to understand that the @JIT annotation should be put on methods which could be called dynamically like you described?

well yeah - have run into that as well ... lately.
but what I dislike is that now some places might have a dependency on the IR package (`org.jruby.ir.JIT`) -> should maybe get relocated somewhere else I guess ... not sure
haven't raised that concern yet -> you should chat with @enebo or @headius on IRC if it was meant to be somehow IR specific or not.

original-brownbear: @kares thanks!

PR #4769, opened 2017-08-31
Fix Broken Compilation

original-brownbear:
`master` doesn't compile right now:
* JDK 7 needs `remove` implemented since it doesn't have the `default` here
* `java.lang.IndexOutOfBoundsException` needs a `String` input to the constructor


original-brownbear:
@headius FYI :) seems this is coming from 7236f72463ee454c72c4f201a695cd6a7a0db9ae

headius:
Strange...it worked ok on the branch I was using. Thanks for picking up the
pieces!

headius:
Ahh I know what it was: I run Java 8 most of the time.

Issue #4693, opened 2017-06-28
Dir.glob returns UTF-8 string with Windows-31J encoding


jakago:
C:/blah/α.rb
```
# coding: utf-8

# 'α'.bytes => [206, 177]
path = 'C:/blah/α.rb'
p path.encoding
p path.bytes

Dir.glob(path) do |file|
  p file.encoding
  p file.bytes
end
```

### Environment

jruby 9.1.12.0 (2.3.3) 2017-06-15 33c6439 Java HotSpot(TM) Client VM 24.65-b04 on 1.7.0_65-b19 +jit [mswin32-x86]

Windows 7 Ultimate Service Pack 1 32-bit

### Expected Behavior

```
C:\blah>ruby α.rb
#<Encoding:UTF-8>
[67, 58, 47, 98, 108, 97, 104, 47, 206, 177, 46, 114, 98]
#<Encoding:UTF-8>
[67, 58, 47, 98, 108, 97, 104, 47, 206, 177, 46, 114, 98]
```

### Actual Behavior

```
C:\blah>jruby α.rb
#<Encoding:UTF-8>
[67, 58, 47, 98, 108, 97, 104, 47, 206, 177, 46, 114, 98]
#<Encoding:Windows-31J>
[67, 58, 47, 98, 108, 97, 104, 47, 206, 177, 46, 114, 98]
```

jakago:
quick fix
```
C:\blah>jruby -Eutf-8 α.rb
#<Encoding:UTF-8>
[67, 58, 47, 98, 108, 97, 104, 47, 206, 177, 46, 114, 98]
#<Encoding:UTF-8>
[67, 58, 47, 98, 108, 97, 104, 47, 206, 177, 46, 114, 98]
```

I checked the source code for Dir.glob.
core/src/main/java/org/jruby/RubyDir.java:229:
```
Encoding enc = runtime.getDefaultExternalEncoding();
```
Encoding.default_external for Japanese Windows 7 is Windows-31J (aka cp932 or ms932), and this code uses it for multi-byte file name built by Java which is encoded with UTF-8.

ahorek:
Ruby uses encoding from input patterns. I tried to fix it, but this case still doesn't work:

Dir.glob(['/tmp'.force_encoding('utf-8'), '/tmp'.force_encoding('windows-1250')])
=> [utf-8, windows-1250]

headius:
Reproduced on Unix by forcing Windows-31J as external encoding:

[...]

PR #4773, opened 2017-09-02
WIP: fix glob encoding

ahorek:
#4693


headius:
Fixed in a similar way by 49d1eb3f86ea27f7d16c504812b778721ee86997. Thank you for the help!

PR #4775, opened 2017-09-04
MINOR: Dry up RubyRange#fixnumEach

original-brownbear:
Just a random spot that can be dried up I found during profiling :)
The method has exactly the same code as the removed block (apart from two `final`s).


kares:
maybe it can start as package visible - since its hard to know what is official 'public' API and what not

original-brownbear:
@kares made it package private as requested :)

PR #4776, opened 2017-09-04
Fix DEPENDENCY constant missing getter on LiveVariableAnalysis

original-brownbear:
`LiveVariableAnalysis` didn't have a `getDependencies` method => the `DEPENDENCIES` constant was never used and `OptimizeDynScopesPass` not returned as a dependency:

* Added the `getDependencies` method
* Made `DEPENDENCIES` `private` here and in the parent to make this a little less confusing (other children of `CompilerPass` use the same approach => `private` rules out any inheritance weirdness imo :))

Also:
* Removed dead field `listeners` from `CompilerPass`
* Made `NO_DEPENDENCIES` in `CompilerPass` private since none of the child compiler passes use it (if they don't override `getDependencies` they get no dependencies => no need to use the constant in children)

### DIFF
--- a/core/src/main/java/org/jruby/ir/passes/CompilerPass.java
+++ b/core/src/main/java/org/jruby/ir/passes/CompilerPass.java
@@ -27,9 +27,7 @@ public abstract class CompilerPass {
 
     static final Logger LOG = LoggerFactory.getLogger(CompilerPass.class);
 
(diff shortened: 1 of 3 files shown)


kares:
excellent, I recall reading these, in the past, and being slightly confused as well :+1:

PR #4777, opened 2017-09-05
 Generate smaller bytecode when using Ruby type on implemented Java Interface + Benchmark

original-brownbear:
It seems when compiling Ruby implementations of Java interfaces, we were generating bytecode that runs any non-primitive type through `convertJavaToUsableRubyObject`.
This seems unnecessary to me if we know that we're dealing with an `IRubyObject` type of argument => we can save that call and the loading of the `Ruby` runtime argument, right?

I included the benchmark I used to verify that this comes with a positive performance impact (for me it showed about a `5%` speed-up).

### DIFF
--- a/bench/src/main/java/org/jruby/benchmark/JavaInterfaceBenchmark.java
+++ b/bench/src/main/java/org/jruby/benchmark/JavaInterfaceBenchmark.java
@@ -1,0 +1,90 @@
+package org.jruby.benchmark;
+
+import java.util.concurrent.ThreadLocalRandom;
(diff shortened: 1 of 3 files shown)


kares:
totally makes sense - thanks for digging throught these bits, there's probably a lot of :gem: to be found :)

headius:
👍 looks good. There's a lot of gems hidden in interface impl and concrete subclass logic:

* Use indy to optimize method lookups, dispatch, and type conversion
* Move more logic to utility methods

This is a good start :-)

PR #4780, opened 2017-09-06
PERFORMANCE: Smaller bytecode for interface impl returns

original-brownbear:
Same as #4777 but for the return type.
There is no point in running the typecast if the return type is an `IRubyObject` :)

That said ... small disclaimer:

I'm a little unsure I didn't make a mistake here, functionally this seems fine in my testing (and is fairly straightforward imo) but the performance gain is much bigger than that from #4777.
The `org.jruby.benchmark.JavaInterfaceBenchmark#benchHalfRubyVersion` benchmark goes from `0.036ops/ns` to `0.045ops/ns` for me, which seems like an unexpectedly huge increase.

### DIFF
--- a/core/src/main/java/org/jruby/java/codegen/RealClassGenerator.java
+++ b/core/src/main/java/org/jruby/java/codegen/RealClassGenerator.java
@@ -817,9 +817,12 @@ public abstract class RealClassGenerator {
                     }
                 }
             } else {
-                mv.ldc(Type.getType(returnType));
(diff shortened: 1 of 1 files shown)


headius:
This seems fine to me. Do you have a lot of code where you're returning IRubyObject types from an interface impl into Java?

original-brownbear:
@headius we have a few spots (especially in upcoming Logstash changes), same as with #4777 I see real world improvements in them from this one :)

headius:
Ok cool, thanks for confirming that. Good patch.

PR #4786, opened 2017-09-09
Faster MixedModeIRMethod#call

original-brownbear:
One performance improvement and one potential bugfix here:

Performance improvement:

* `DynamicMethodBox` adds a needless level of indirection. Moved the fields directly into `MixedModeIRMethod`
* Stopped making a copy of `actualMethod` for `null` check and invocation. I think this is fine, there is no place in the code that could possibly set a `null` here as far as I can see, right?

Bugfix:

* `volatile int callCount` is not safe for use as a counter here I think. It is incremented in a synchronized block (`box.callCount++`) in the jit method, but can also be set without synchronization from `setCallCount(int callCount)`.
  * Fixed by synchronizing all write access to `callCount`

---------------

With indy on, this gets me from `0.055 ops/ns` (`master`) to `0.061ops/ns` (this branch) for the JMH `org.jruby.benchmark.JavaInterfaceBenchmark#benchHalfRubyVersion` :)

### DIFF
--- a/core/src/main/java/org/jruby/internal/runtime/methods/MixedModeIRMethod.java
+++ b/core/src/main/java/org/jruby/internal/runtime/methods/MixedModeIRMethod.java
@@ -1,11 +1,13 @@
 package org.jruby.internal.runtime.methods;
 
+import java.io.ByteArrayOutputStream;
 import org.jruby.MetaClass;
 import org.jruby.RubyClass;
 import org.jruby.RubyModule;
 import org.jruby.compiler.Compilable;
 import org.jruby.internal.runtime.AbstractIRMethod;
-import org.jruby.ir.*;
+import org.jruby.ir.IRMethod;
+import org.jruby.ir.IRScope;
 import org.jruby.ir.interpreter.InterpreterContext;
 import org.jruby.ir.persistence.IRDumper;
 import org.jruby.ir.runtime.IRRuntimeHelpers;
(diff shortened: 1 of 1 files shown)


core/src/main/java/org/jruby/internal/runtime/methods/MixedModeIRMethod.java
```diff
@@ -85,12 +81,10 @@ public InterpreterContext ensureInstrsReady() {
...
         if (IRRuntimeHelpers.isDebug()) doDebug();
 
-        DynamicMethodBox box = this.box;
-        if (box.callCount >= 0) tryJit(context, box);
-        DynamicMethod jittedMethod = box.actualMethod;
+        if (callCount >= 0) tryJit(context);
 
-        if (jittedMethod != null) {
-            return jittedMethod.call(context, self, clazz, name, args, block);
+        if (actualMethod != null) {
```
kares: used to be one volatile field read, now its 2 right? 
not sure it matters - but I think Java 8 still doesn't re-arrange volatile reads

original-brownbear: @kares you're right, even though this isn't really contended it's still better to copy. Sec adding the copy back :)

original-brownbear: Fixed :)

kares:
looking good, always wondered whether we actually need to care about volatile semantics around call-count
they're mostly to be reached from a single-thread and if they aren't exact well it does not matter much ...

original-brownbear:
@kares yea I think `volatile` on the call count doesn't get us much except keeping the count exact.
Also, I wonder we couldn't just use `AtomicBoolean`'s `compareAndSet` to guard against compiling multiple times. Might improve startup times to not have a hard synchronization in `tryJit` (though it could lead to other concurrency issues that I just don't see?).

PR #4790, opened 2017-09-17
adds windows build

mkristian:
@enebo this is a way to attach an extra artifact. see the maven/jruby-windows/pom.rb to adjust it to pick the 'real' file or even trigger the build of the real file if this is possible. any questions or if you want to include it somewhere differently then let's discuss it on this PR


headius:
I have added checksum generation to the release artifacts in 3b17b3b29132d05d7d48ad521ddeed73500613a1 and modified the post_process_artifacts target to also generate sha-512 for the installer. The new installer setup will need to also generate these checksums.

enebo:
@mkristian I picked this up and started to work on it.   Besides figuring out how to redo building the installers from maven (which will at some level just be new contents in pom.rb files) I have some questions:

1. if I run what was the first pass I error on:

```text
[ERROR] Failed to execute goal org.apache.maven.plugins:maven-install-plugin:2.4:install (default-install) on project jruby-windows: Failed to install artifact org.jruby:jruby-windows:jar:sources:9.2.10.0-SNAPSHOT: /home/enebo/work/jruby/maven/jruby-windows/src/empty.jar (No such file or directory) -> [Help 1]
```

and I see:

```text
  # we have no sources and attach an empty jar later in the build to
  # satisfy oss.sonatype.org upload
  plugin( :source, 'skipSource' =>  'true' )
```

So what should happen here to make the empty jar?

My other immediate question is since we have 2 installers (32 and 64 bits).  We will need to have 2 jruby-windows subdirs right?  I am thinking jruby-windows and jruby-windows-32.

PR #4803, opened 2017-10-03
Support Pattern.compile(/regex/)

madrob:
Would like to add tests, but I'm not familiar enough with ruby to know where they go.

### DIFF
--- a/core/src/main/java/org/jruby/javasupport/ext/JavaUtilRegex.java
+++ b/core/src/main/java/org/jruby/javasupport/ext/JavaUtilRegex.java
@@ -89,6 +89,11 @@ public abstract class JavaUtilRegex {
             return context.runtime.newBoolean(i);
         }
 
+        @JRubyMethod(name = "compile", required = 1)
+        public static java.util.regex.Pattern compile(final ThreadContext context, final IRubyObject self, IRubyObject regex) {
+            return java.util.regex.Pattern.compile(regex.asJavaString());
+        }
+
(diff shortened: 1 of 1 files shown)


kares:
thanks for your effort, 
but I am really not sure about this change - seems like a one off requirement which isn't thought through ...

we're trying to make Java objects more Ruby-like here (posted the link to these on IRC as a hint of what can you expect from JRuby's extension on java.util.regexp).

... the right place for this would be having a `/ruby/.to_java` to return a `java.util.Regexp` when requested
but since the Java version is less powerful and does not support lot of the features its still questionable.

kares:
besides that I am not sure whether it actually works - escaping and potentially flags are not handled?
sorry but this simply needs more work to make it worth looking at.

madrob:
Implementing this as a `/ruby/.to_java` makes sense, I'll rework it in that direction. What is here now is definitely a WIP, I thought the lack of tests would indicate that it's not ready for commit but should have labelled it explicitly anyway.

The use case for this is when calling java methods that expect a Pattern argument. I understand and agree that `Pattern` is less powerful than `RubyRegexp` but having the ability to intelligently downgrade would be useful. Otherwise, users do end up just using `to_s` with all the rough edges that it may present.

enebo:
@madrob you should definitely figure out what behavior (and how to detect it) that should happen when you try and make a pattern which is not possible in Java Regexp.  It must be able to explicitly know that it is an invalid regexp or I think it will be confusing.

kares:
as agreed this needs a different direction -> seems there isn't much activity so will likely get closed, for now

PR #4806, opened 2017-10-05
Add Ruby part of TracePoint impl

### LINKED ISSUE #4800: TracePoint: Support .trace
This issue records the fact that [`TracePoint.trace`](http://ruby-doc.org/core-2.4.2/TracePoint.html) is not supported. It does a combined instantiation and enabling of a _TracePoint_ instance.

[The JRuby  extension TracePoint.java **seems not to implement** `.trace` and `.stat`](https://github.com/jruby/jruby/blob/master/core/src/main/java/org/jruby/ext/tracepoint/TracePoint.java)

**Workaround**: JRuby's TracePoint **has** `.new`. and `#enable`. Use `TracePoint.new` and then `#enable` the resulting `TracePoint` instance.

### Environment

```
$ jruby -v
jruby 9.1.13.0 (2.3.3) 2017-09-06 8e1c115 Java HotSpot(TM) 64-Bit Server VM 25.111-b14 on 1.8.0_111-b14 +jit [darwin-x86_64]
$ uname -a
Darwin gungnir.lan 16.7.0 Darwin Kernel Version 16.7.0: Thu Jun 15 17:36:27 PDT 2017; root:xnu-3789.70.16~2/RELEASE_X86_64 x86_64
```

### Expected Behavior

* I expect that `TracePoint.trace` instantiates and enables a _TracePoint_.

Example script to run:

```ruby
TracePoint.trace(:call) { puts 'Hi' }
```

Example execution:

```
irb(main):001:0> TracePoint.trace(:call) { puts 'Hi' }
Hi
Hi
Hi
Hi
Hi
Hi
Hi
=> #<TracePoint:enabled>
Hi
Hi
Hi
Hi
Hi
Hi
Hi
Hi
Hi
Hi
Hi
Hi
Hi
Hi
Hi
Hi
Hi
Hi
Hi
Hi
Hi
Hi
Hi
```

### Actual Behavior

A `NoMethodError` is raised.

```
irb(main):002:0> TracePoint.trace(:call) { puts 'Hi' }
NoMethodError: undefined method `trace' for TracePoint:Class
Did you mean?  trace_var
               trap
	from org/jruby/RubyBasicObject.java:1657:in `method_missing'
	from (irb):2:in `<eval>'
	from org/jruby/RubyKernel.java:994:in `eval'
	from org/jruby/RubyKernel.java:1114:in `catch'
	from org/jruby/RubyKernel.java:1114:in `catch'
	from /Users/olle/.rubies/jruby-9.1.13.0/bin/irb:13:in `<main>'
```

olleolleolle:
This PR adds a Ruby implementation of the class method `.trace` to TracePoint.

The file is named after C impl file in MRI.

Fixes #4800

Questions for reviewer:

- Do we want to have a full implementation of each of TracePoint methods?
- Are there tests that I can re-enable, now?

### DIFF
--- a/core/src/main/ruby/jruby/vm_trace.rb
+++ b/core/src/main/ruby/jruby/vm_trace.rb
@@ -1,0 +1,13 @@
+# frozen_string_literal: true
+#
+# .rb part for JRuby's native TracePoint impl
+
+class TracePoint
+
+  def self.trace(*events, &blk)  # :nodoc:
+    tp = new(*events, &blk)
+    tp.enable
+    tp
+  end
+
+end


headius:
Another question: does MRI trace the exit from the `trace` function? Since I suggested you do this in Ruby, that might be hard to avoid if not. Didn't even think of it until now.

olleolleolle:
@headius To answer your question with a "it seems so":

```
--- ruby/jruby ‹feature/4800-tracepoint-trace* ?› » JRUBY_OPTS=--debug irb -r ./vm_trace.rb
Ignoring byebug-9.0.6 because its extensions are not built.  Try: gem pristine byebug --version 9.0.6
irb(main):001:0> TracePoint.trace(:call) { |tp| puts tp.inspect }
#<TracePoint:call>
#<TracePoint:call>
#<TracePoint:call>
#<TracePoint:call>
#<TracePoint:call>
#<TracePoint:call>
#<TracePoint:call>
=> #<TracePoint:enabled>
#<TracePoint:call>
#<TracePoint:call>
#<TracePoint:call>
#<TracePoint:call>
#<TracePoint:call>
#<TracePoint:call>
#<TracePoint:call>
#<TracePoint:call>
#<TracePoint:call>
#<TracePoint:call>
#<TracePoint:call>
#<TracePoint:call>
#<TracePoint:call>
#<TracePoint:call>
#<TracePoint:call>
#<TracePoint:call>
#<TracePoint:call>
#<TracePoint:call>
#<TracePoint:call>
#<TracePoint:call>
#<TracePoint:call>
#<TracePoint:call>
#<TracePoint:call>
irb(main):002:0>
#<TracePoint:call>
#<TracePoint:call>
```

olleolleolle:
@headius Should we jettison this idea of using Ruby for a TracePoint patched implementation?

headius:
So here's the thing. It appears that MRI *does* trace the return from `TracePoint.trace` but it does so as a `:c_return` rather than a `:return`.

```
[] ~/projects/jruby $ ruby -e 'def bar; end; TracePoint.trace(:return) {|x| puts x.method_id }; puts :foo; bar'
foo
bar

[] ~/projects/jruby $ ruby -e 'def bar; end; TracePoint.trace(:c_return) {|x| puts x.method_id }; puts :foo; bar'
trace
to_s
foowrite

write
puts
puts
```

In JRuby, it's reversed, since we would implement it in Ruby instead of "C".

I'm not sure how important this is in practice. If there's logic that's attempting to hook only C methods, it will not see the return from `trace`. If there's logic that's attempting to hook only Ruby methods, it will see an unexpected return from `trace`.

Big deal? No big deal?

headius:
@olleolleolle I think we should just go with it and see what happens. Worst case we do a minor bit of porting and put it in Java.

As for tests, there *might* be something in spec/ruby but more likely there's a test excluded under test/mri/ruby/test_tracepoint.rb. The excludes are in test/mri/excludes with file path matching test class.

headius:
It might also be worth digging up any issue where this was added (in MRI) or opening a new one and clarifying with them how "spec" the `:c_return` vs `:return` is for cases like this.

headius:
Meh, I'm just going to use the Ruby version you provided to implement a Java version. I'm glad you got a chance to see how this fits together in JRuby, but we've already spent too long debating whether this minor difference is worth using Ruby for :-)

enebo:
I said this on irc last week but I think semantically this should return :return if it is implemented in Ruby.  Think of Rubinius/RubyTruffle or even MRI of the future where they start writing stuff in Ruby implwise (for inlining/JIT stuff).  All this stuff will just change what they are implemented in.  Any tests which return this as an error seems like an impl-specific test to me.

headius:
@enebo Yeah I never intended to tweak it so the Ruby version would do :c_return instead of :return. It's a minor difference, but it's also a five-minute patch to just write this in Java and moot the point.

PR #4815, opened 2017-10-12
Fix spec of Thread#status for an externally killed thread that sleeps

### LINKED ISSUE #4705: Thread.current.status is wrong for killed thread
### Environment

Provide at least:
* JRuby version: jruby 9.1.12.0 (2.3.3) 2017-06-15 33c6439 Java HotSpot(TM) 64-Bit Server VM 25.112-b16 on 1.8.0_112-b16 [darwin-x86_64]
* Operating system and platform: Darwin de01pc11303.global.jhcn.net 16.6.0 Darwin Kernel Version 16.6.0: Fri Apr 14 16:21:16 PDT 2017; root:xnu-3789.60.24~6/RELEASE_X86_64 x86_64

### Expected Behavior

A sleeping thread that gets killed should have status 'aborting' and not 'run'.

### Actual Behavior

The following code outputs 'run' on jruby, whereas on mri it outputs 'aborting'.

```ruby
t = Thread.new {
  begin
    sleep
  ensure
    puts Thread.current.status # should output 'aborting' when killed
  end
}
t.kill
```

Example run on jruby:
```ruby
irb(main):020:0> t = Thread.new { begin; sleep; ensure; puts Thread.current.status; end }
=> #<Thread:0x79ad8b2f@(irb):20 sleep>
irb(main):021:0> t.kill
run
=> #<Thread:0x79ad8b2f@(irb):20 dead>
```

Example run on mri ruby:
```ruby
irb(main):013:0>  t = Thread.new { begin; sleep; ensure; puts Thread.current.status; end }
=> #<Thread:0x007faefc8bad78@(irb):13 run>
irb(main):014:0> t.kill
aborting
=> #<Thread:0x007faefc8bad78@(irb):13 sleep>
```

This code is a simplified version of ActiveRecord transaction code that fails.
See:
  * https://github.com/rails/rails/blob/master/activerecord/lib/active_record/connection_adapters/abstract/transaction.rb#L237
  * Failing test case: https://github.com/rails/rails/blob/master/activerecord/test/cases/transactions_test.rb#L587

iaddict:
Fixes #4705

### DIFF
--- a/core/src/main/java/org/jruby/RubyThread.java
+++ b/core/src/main/java/org/jruby/RubyThread.java
@@ -1511,7 +1511,9 @@ public class RubyThread extends RubyObject implements ExecutionContext {
     }
 
     public void exitSleep() {
-        status.set(Status.RUN);
+        if (status.get() != Status.ABORTING) {
+            status.set(Status.RUN);
+        }
     }
 
     @JRubyMethod(name = {"kill", "exit", "terminate"})
--- a/spec/tags/ruby/core/thread/status_tags.txt
+++ b/spec/tags/ruby/core/thread/status_tags.txt
@@ -1,2 +1,1 @@
 unstable:Thread#status reports aborting on a killed thread
-fails:Thread#status reports aborting on an externally killed thread that sleeps


headius:
There could be a race here, if a thread marks itself as aborting and then some other thread tries to exit sleep at the same time, but ideally exitSleep should only be called by that thread when it transitions. The uses I reviewed seem to fit that pattern.

Worst case, some bad actor might cause an aborting thread to mark itself as runnable again. it would be no worse than now but less likely to happen.

eregon:
FWIW The way we do in in TruffleRuby is to save/restore the Thread status around sleep/blocking tasks.
And since only the current thread modifies its status and we poll for interrupts after restoring the thread state, I believe this is safe and race-free.

PR #4817, opened 2017-10-15
Fix an infinite loop when attempting to read from a directory

philr:
When attempting to open a directory with `File.open` and then read from it, JRuby 9.1.13.0 (on platforms other than Windows) will enter an infinite loop and hang:

```ruby
> Timeout::timeout(10) { File.open('.') {|f| f.read } }
Timeout::Error: execution expired
        from org/jruby/RubyIO.java:3009:in `read'
        from org/jruby/RubyIO.java:2987:in `read'
        from (irb):1:in `block in evaluate'
        from org/jruby/RubyIO.java:1156:in `open'
        from (irb):1:in `block in evaluate'
        from org/jruby/ext/timeout/Timeout.java:117:in `timeout'
        from (irb):1:in `<eval>'
```

When attempting to do the same with MRI (tested with versions 1.9.3 through to 2.4.2) an `Errno::EISDIR` exception is raised:

```ruby
> File.open('.') {|f| f.read }
Errno::EISDIR: Is a directory @ io_fread - .
        from (irb):1:in `read'
        from (irb):1:in `block in irb_binding'
        from (irb):1:in `open'
        from (irb):1
```

The hang occurs because `OpenFile.fillbuf()` and `OpenFile.ioBufread()` loop to retry on errors. This is desirable for temporary errors like `EAGAIN`, but not for permanent errors like `EISDIR`.

This pull request modifies `OpenFile.waitReadable()` to abort retries when a permanent error is encountered. It also changes `Helpers.errnoFromException()` to map the `"Is a directory"` `IOException` message to `EISDIR`.

This change only applies to non-Windows platforms. On Windows, `File.open` raises an exception instead.

### DIFF
--- a/core/src/main/java/org/jruby/runtime/Helpers.java
+++ b/core/src/main/java/org/jruby/runtime/Helpers.java
@@ -213,6 +213,8 @@ public class Helpers {
                 case "Message too large": // Alpine Linux
                 case "Message too long":
                     return Errno.EMSGSIZE;
+                case "Is a directory":
+                    return Errno.EISDIR;
             }
         }
         return null;
--- a/core/src/main/java/org/jruby/util/io/OpenFile.java
+++ b/core/src/main/java/org/jruby/util/io/OpenFile.java
@@ -1407,6 +1407,12 @@ public class OpenFile implements Finalizable {
                 return false;
             }
 
+            if (posix.errno != null && posix.errno != Errno.EAGAIN
+                    && posix.errno != Errno.EWOULDBLOCK && posix.errno != Errno.EINTR) {
+                // Encountered a permanent error. Don't read again.
+                return false;
+            }
+
             if (fd.chSelect != null) {
                 unlock();
                 try {
--- a/test/jruby/test_io.rb
+++ b/test/jruby/test_io.rb
@@ -379,6 +379,21 @@ class TestIO < Test::Unit::TestCase
     assert(a)
   end
 
+  unless WINDOWS
+    # On Windows an error is raised when opening a directory instead of when reading.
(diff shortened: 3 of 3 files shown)


kares:
have tried running this against **jruby-9.1** but this time there seems to be more failures from `test:jruby`
... than before (last jruby-9.1 build here) : https://travis-ci.org/kares/jruby/jobs/291393645

philr:
@kares I'm seeing the same test failures from `test:jruby` in the build of the merge of this change to your jruby-9.1 branch (https://travis-ci.org/kares/jruby/jobs/291393645) as in the build of the parent commit (https://travis-ci.org/jruby/jruby/jobs/289840028). The following tests fail in both builds:

- test_ipv6_loopback?
- test_directory_query
- test_executable?_query
- test_executable_real?_query
- test_file_exist_query
- test_file_query
- test_readable_query

I've just merged this onto jruby-9.1 myself (philr/jruby@5d25f7bbdbd7aa4381c76aad3a8789e883d0e96f). Running `test:jruby` in my dev environment, I only see the test_ipv6_loopback? failure. This is the same outcome as the latest jruby-9.1 build (https://travis-ci.org/jruby/jruby/jobs/293860475).

kares:
oh really? have assumed they're builds from the same base, although I did not look carefully ...
hard to work with CI failures :sob:  - but since jruby-9.1 only has the one let's have it for **9.1.14.0**

kares:
done at https://github.com/jruby/jruby/commit/963e2a15549dcaf23aee2fb6a53c43b6ff3e0a81 - thanks for the fix

PR #4818, opened 2017-10-15
Raise Errno::EISDIR when opening a directory for reading on Windows

philr:
On Windows, MRI 2.0.0 and later raises `Errno::EISDIR` when attempting to open a directory for reading:

```ruby
> File.open('C:\\Windows', 'r')
Errno::EISDIR: Is a directory @ rb_sysopen - C:\Windows
        from (irb):1:in `initialize'
        from (irb):1:in `open'
        from (irb):1
```

However, JRuby 9.1.13.0 raises `Errno::EACCES` instead:

```ruby
> File.open('C:\\Windows', 'r')
Errno::EACCES: Permission denied - C:\Windows
        from org/jruby/RubyFile.java:366:in `initialize'
        from org/jruby/RubyIO.java:1154:in `open'
        from (irb):1:in `<eval>'
        from org/jruby/RubyKernel.java:994:in `eval'
        from org/jruby/RubyKernel.java:1114:in `catch'
        from org/jruby/RubyKernel.java:1114:in `catch'
```

This pull request modifies `RegularFileResource.openChannel()` to throw `ResourceException.FileIsDirectory` in all cases when opening a directory instead of just for writes.


core/src/main/java/org/jruby/util/RegularFileResource.java
```diff
@@ -218,7 +218,7 @@ public Channel openChannel(ModeFlags flags, int perm) throws ResourceException {
             return channel;
         }
 
-        if (file.isDirectory() && flags.isWritable()) {
+        if (file.isDirectory()) {
```
enebo: @philr Should this happen on all platforms and not just windows?  Sorry this PR got lost in the cracks.

philr: @enebo On platforms other than Windows (with both MRI and JRuby), `File.open(directory)` succeeds and an `Errno::EISDIR` exception is raised when calling `File#read`.

On Windows, `File.open(directory)` raises an exception. MRI raises `Errno::EISDIR`, but JRuby raises `Errno::EACCES`.

To replicate the behaviour of MRI, it is just this code that handles Windows in the `openChannel()` method that needs to be changed.

PR #4821, opened 2017-10-18
switch ShellLauncher to use getAbsolutePath

zendes:
I ran into some issues with using JRuby on a system that uses busybox as it's userland, and without this patch things won't work.

For a bit of background: busybox's goal is to be a tiny userland (commands like ls and mv) for space-constrained devices. One of the ways it does this is by being a multi-call binary. You symlink it with names such as ls and cp, and it uses $0 to determine how it's called and thus what it should do. JRuby resolves the symlink before it execs, so busybox loses $0 and thus doesn't know what the user originally called. This patch fixes that.

### DIFF
--- a/core/src/main/java/org/jruby/util/ShellLauncher.java
+++ b/core/src/main/java/org/jruby/util/ShellLauncher.java
@@ -1200,11 +1200,7 @@ public class ShellLauncher {
             } else {
                 verifyExecutable();
                 execArgs = args;
-                try {
-                    execArgs[0] = executableFile.getCanonicalPath();
-                } catch (IOException ioe) {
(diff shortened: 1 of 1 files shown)


zendes:
looks like the CI build failed in a bunch of places that are unrelated to my change, and it was also failing to build before my change. what do you need from me to look at this?

mullermp:
+1

I don't use busybox, but we do this same exact fix for a symlinking problem at my company.

kares:
sounds like, this makes sense - maybe even to get into **9.1.14.0**

headius:
I'll merge this to 9.1.14 and master. Thanks for the help @zendes!

Issue #1615, opened 2014-04-09
BigDecimal::limit has not effect


nmk:
Using [BigDecimal::limit](http://www.ruby-doc.org/stdlib-1.9.3/libdoc/bigdecimal/rdoc/BigDecimal.html#method-c-limit) has not effect on following computations with BigDecimal instances. This works in MRI 1.9.3 and later.

Compare the following two sessions:

MRI (1.9.3-p545):

```
~ $ irb -r"bigdecimal" -r "bigdecimal/util"
1.9.3-p545 :001 > BigDecimal.limit(3)
 => 0
1.9.3-p545 :002 > 10.to_d / 3.to_d
 => #<BigDecimal:7fd034801d38,'0.333E1',18(36)>
1.9.3-p545 :003 > (10.to_d / 3.to_d).to_f
 => 3.33
1.9.3-p545 :004 > (10.to_d / 3.to_d).to_s("F")
 => "3.33"  
```

Latest JRuby (1.7.11):

```
~ $ irb -r"bigdecimal" -r"bigdecimal/util"
jruby-1.7.11 :001 > BigDecimal.limit(3)
 => 0
jruby-1.7.11 :002 > 10.to_d / 3.to_d
 => #<BigDecimal:136c2147,'0.33333333333333333333333333333333333333333333333333333333333333333333333333333333333333333333333333333333333333333333333333333333333333333333333333333333333333333333333333333333333333333333333333333333E1',200(204)>
jruby-1.7.11 :003 > (10.to_d / 3.to_d).to_f
 => 3.3333333333333335
jruby-1.7.11 :004 > (10.to_d / 3.to_d).to_s("F")
 => "3.3333333333333333333333333333333333333333333333333333333333333333333333333333333333333333333333333333333333333333333333333333333333333333333333333333333333333333333333333333333333333333333333333333333"
```

dymaxionuk:
This bug has been open for 2 years. Surprised nobody else has felt the pain! 

Since #limit is not working for division, I have to post process all my output data, and traverse every field in my data structures and manually round.

This is cumbersome.   Can someone help fix this?  Thanks

kares:
@dymaxionuk you could have tried fixing it than or posting a bounty. it shouldn't be very hard and there's even some code in the comments to start with ...

MrBerg:
I made a pull request for this: https://github.com/jruby/jruby/pull/4830

headius:
[...]

PR #4830, opened 2017-10-27
Make BigDecimal::limit work for subtraction and division. Fix issue #1615

MrBerg:


core/src/main/java/org/jruby/ext/bigdecimal/RubyBigDecimal.java
```diff
@@ -1040,7 +1035,9 @@ public IRubyObject op_quo(ThreadContext context, IRubyObject other) {
         int pow = len / 4;
         int precision = (pow + 1) * 4 * 2;
 
-        return divWithScale(context, val, precision);
+        IRubyObject result = divWithScale(context, val, precision);
+        if (result.getClass() == RubyBigDecimal.class) return ((RubyBigDecimal) result).setResult();
```
kares: why not an `instanceof` check?

MrBerg: Paranoia that there might be some hidden RubyBigDecimal subclass with strange behaviour. 

Nah, it can totally be changed to `instanceof`.

kares: thx, if you get a chance please adjust - JRuby is quite in control of its sub-classes. 
no need for paranoia :wink:

MrBerg: Sure, I pushed a fix.

headius:
👍 from me too. I'll merge to master and 9.1.

headius:
Actually looks like there are other changes to BigDecimal on master that would need to be merged to jruby-9.1 first, like the divWithScale method. I'll try to sort it out.

headius:
Merged into 9.1 as dc7b498598080f28dea5118f12c10e09139140c3 and 28eb942fc951df228bf5b5fe4a157bf5db154515. The missing commit from master to make it merge nicely was f38605477cbf2ae8914bd5a087367adcf5465985.

Issue #4459, opened 2017-01-25
ArgumentError when using 2-argument for of IO#each


rovf:
jruby 1.7.26 (1.9.3p551) 2016-08-26 69763b8 on Java HotSpot(TM) 64-Bit Server VM 1.7.0_79-b15 +jit [Windows 7-amd64]

f=File.open('some_file','r')
f.each(nil, 1024) {}

I get: 

ArgumentError: wrong number of arguments (2 for 1)
	from (irb):16:in `evaluate'
	from org/jruby/RubyKernel.java:1079:in `eval'
	from org/jruby/RubyKernel.java:1479:in `loop'

The usage of f.each should be correct. From http://ruby-doc.org/core-1.9.3/IO.html#method-i-each :

------------------
each(sep,limit) {|line| block } → ios
Executes the block for every line in ios, where lines are separated by sep.
------------------

aayush2610:
Hi, I am new to jruby and would like to contribute. Please let me know the details of this issue.

kares:
has been working since around the (~2017) JRuby 9.1 days

PR #4833, opened 2017-10-30
Added specs for IO#each(sep,limit)

MrBerg:
As part of my unsuccessful effort of trying to replicate #4459 in master, I added (well, mostly copied...) some specs.


spec/ruby/core/io/shared/each.rb
```diff
@@ -115,6 +115,45 @@
...
+        end
+      end
+      describe "when passed an empty String as a separator" do
+        it "yields each paragraph" do
+          @io.send(@method, "", 1024) { |s| ScratchPad << s }
+          ScratchPad.recorded.should == IOSpecs.paragraphs
+        end
+      end
+    end
+  end
```
eregon: This look good.
Could you ensure there is one blank line between each `describe` block, like in the rest of the file?

MrBerg: Sure. Is https://github.com/jruby/jruby/pull/4833/commits/6dfd1b4cb8da15d36af63ea57532bea070edd3c4 what you had in mind?

eregon: Yes, exactly :)

PR #4853, opened 2017-11-12
Fix typo in a method name comment

olleolleolle:
This PR fixes a typo in a comment in the Ripper code-base: a reference to an MRI function which had a typo.


olleolleolle:
Hej hej @enebo - I added `[ci skip]` to this trivial change.

olleolleolle:
I found this while reading Ripper code to figure out where the error message from this test failure came from:

https://github.com/lsegal/yard/pull/1135#issuecomment-343730649

Issue #4844, opened 2017-11-09
Potential bug in internal/runtime/ThreadService?


nglorioso:
https://github.com/jruby/jruby/blob/8ea1605977039491ffa343b5b9687973f34d3d0a/core/src/main/java/org/jruby/internal/runtime/ThreadService.java#L292

Here, and a bit below, a Map's (a synchronized WeakHashMap) size is checked as being >= 0 to twiddle the state of the `polling` field.

I'm pretty sure WeakHashMap would never return a size < 0. I haven't fully grasped the rest of the file (should this only be set iff there's more than 1 active thread?), but certainly `if (map.size() >= 0)` is always going to be true.

headius:
Yeah seems like this should be `> 0`.

Luckily, it turns out we don't use the `polling` field for anything right now. I remember at one point wanting to add this feature, where like MRI we could turn off certain thread events (like kill, raise across threads) when there's only a main thread. Ultimately checking this boolean turns out to be as expensive as polling anyway, so it never went anywhere.

I'll leave the flag, for API and info purposes, but I'll fix it to be `> 0`.

headius:
Yeah, this flag was all over the place. I also found the following logic, which seems to disable polling when there's one (or zero?) threads, so this isn't even internally consistent.

I'm removing the field, deprecating everything involving it, and modifying `getPolling` to use `> 1` since that's what this was intended to indicate.

headius:
Also worth noting that the access to the map is costly (slower thread init/teardown times), non-atomic (so it's only useful for an immediately-stale point-in-time snapshot) and the boolean is also non-atomically updated, so threads can step on each other.

Just a half-baked feature that never landed.

nglorioso:
Thanks for the fix!

PR #4860, opened 2017-11-21
Fix issues discovered using Error Prone

nglorioso:
Hello!

As part of the investigation that led to #4844, I integrated Error Prone (http://errorprone.info) into the JRuby compiler. (See configuration in https://github.com/nglorioso/jruby/commit/ae3a5cb605b2b8b31ba0fae66332086ce01730f9).

This found a number of bugs (ranging in severity/nagginess). I'm happy to revert or change any of these, but all of them represent real defects:

RubyEnumerable - each_with_index19 is trivially infinitely-recursive. I assumed that each_with_index is appropriate
StructLayout/SymbolProc - `hashCode()` ends up RSHIFT'ing an int value by 32 bits, which always ends up resulting in 0. I changed to 16 bits, but if you needed a stable hash code, I can change to XOR 0
RipperContext/FullInterpreterContext - Array.toString() is unhelpful for error messages. Manually unrolled them with Arrays.toString.
RealClassGenerator - an exception was created and thrown on the floor in an exceptional condition. I threw it instead.
Pack - ulMask was an int constant left shifted by 56, but since it was an int, it shifts out to 0 instead of 0xfe000000.
SelectExecutor - Here, I don't have enough context to make the right changes, but the errorKeyList.contains() calls will never return true, since the file descriptor objects can't be contained in `errorKeyList`

### DIFF
--- a/core/src/main/java/org/jruby/ext/ffi/StructLayout.java
+++ b/core/src/main/java/org/jruby/ext/ffi/StructLayout.java
@@ -658,7 +658,10 @@ public final class StructLayout extends Type {
 
         @Override
         public int hashCode() {
-            return 53 * 5 + (int) (this.offset ^ (this.offset >>> 32)) ^ type.hashCode();
+            int result = super.hashCode();
+            result = 31 * result + type.hashCode();
+            result = 31 * result + offset;
+            return result;
         }
 
         /**
--- a/core/src/main/java/org/jruby/ext/ripper/RipperLexer.java
+++ b/core/src/main/java/org/jruby/ext/ripper/RipperLexer.java
@@ -2349,7 +2349,7 @@ public class RipperLexer extends LexingCommon {
         int length = precise_mbclen();
 
         if (length <= 0) {
-            compile_error("invalid multibyte char (" + getEncoding().getName() + ")");
+            compile_error("invalid multibyte char (" + getEncoding() + ")");
             return false;
         }
 
--- a/core/src/main/java/org/jruby/ir/interpreter/FullInterpreterContext.java
+++ b/core/src/main/java/org/jruby/ir/interpreter/FullInterpreterContext.java
@@ -188,6 +188,6 @@ public class FullInterpreterContext extends InterpreterContext {
             if (ipc <= rescueIPCs[i]) return rescueIPCs[i + 1];
         }
 
-        throw new RuntimeException("BUG: no RPC found for " + getFileName() + ":" + getName() + ":" + ipc + getInstructions());
+        throw new RuntimeException("BUG: no RPC found for " + getFileName() + ":" + getName() + ":" + ipc);
     }
 }
--- a/core/src/main/java/org/jruby/ir/operands/SymbolProc.java
+++ b/core/src/main/java/org/jruby/ir/operands/SymbolProc.java
@@ -34,7 +34,10 @@ public class SymbolProc extends ImmutableLiteral {
 
     @Override
     public int hashCode() {
-        return 47 * 7 + (int) (this.name.hashCode() ^ (this.encoding.hashCode() >>> 32));
+        int result = super.hashCode();
+        result = 31 * result + name.hashCode();
+        result = 31 * result + encoding.hashCode();
+        return result;
     }
 
     @Override
--- a/core/src/main/java/org/jruby/RubyEnumerable.java
+++ b/core/src/main/java/org/jruby/RubyEnumerable.java
@@ -1177,7 +1177,7 @@ public class RubyEnumerable {
 
     @Deprecated @SuppressWarnings("deprecation")
     public static IRubyObject each_with_index19(ThreadContext context, IRubyObject self, IRubyObject[] args, Block block) {
-        return each_with_index19(context, self, args, block);
+        return each_with_index(context, self, args, block);
     }
 
     @JRubyMethod(required = 1)
--- a/core/src/main/java/org/jruby/util/Pack.java
+++ b/core/src/main/java/org/jruby/util/Pack.java
@@ -1320,7 +1320,7 @@ public class Pack {
                     }
 
                     long ul = 0;
-                    long ulmask = (0xfe << 56) & 0xffffffff;
+                    long ulmask = (0xfeL << 56) & 0xffffffff;
                     RubyBignum big128 = RubyBignum.newBignum(runtime, 128);
                     int pos = encode.position();
 
--- a/core/src/main/java/org/jruby/util/io/ChannelStream.java
+++ b/core/src/main/java/org/jruby/util/io/ChannelStream.java
@@ -561,7 +561,7 @@ public class ChannelStream implements Stream, Finalizable {
 
         dst.ensure(Math.min(len, bufferedInputBytesRemaining()));
 
-        if (hasUngotChars() && hasUngotChars()) {
+        if (hasUngotChars()) {
             for(int i = 0; i < ungotChars.length(); i++){
                 byte ungotc = (byte) ungotChars.get(i);
                 ++bytesCopied;


kares:
these are looking really good ... @headius let's have these for 9.1.15 shall we?
have tried find bugs + IDEA's analysis previously but the amount was just too much to find such real gems.

enebo:
The ripper suggestion is wrong even though it did point out there was a bug there.  These all look good except that one.  That one should be something like:

```java
StringBuilder buf = new StringBuilder();
buf.append((char) end);
```

@nglorioso change to that or omit that change otherwise I think it all looks good to me.  Although I surprising bad at bit math stuff so the StructLayout change would be nice if someone else said that is ok.

core/src/main/java/org/jruby/ext/ffi/StructLayout.java
```diff
@@ -658,7 +658,7 @@ public boolean equals(Object obj) {
 
         @Override
         public int hashCode() {
-            return 53 * 5 + (int) (this.offset ^ (this.offset >>> 32)) ^ type.hashCode();
+            return 53 * 5 + (int) (this.offset ^ (this.offset >>> 16)) ^ type.hashCode();
```
headius: This change fixes the `offset` int getting shifted to oblivion. However, I'd prefer to replace this with a hashCode generated by IntelliJ as follows:

```diff
diff --git a/core/src/main/java/org/jruby/ext/ffi/StructLayout.java b/core/src/main/java/org/jruby/ext/ffi/StructLayout.java
index f7e269d777..4d5bfdf0a8 100644
--- a/core/src/main/java/org/jruby/ext/ffi/StructLayout.java
+++ b/core/src/main/java/org/jruby/ext/ffi/StructLayout.java
@@ -658,7 +658,10 @@ public final class StructLayout extends Type {
 
         @Override
         public int hashCode() {
-            return 53 * 5 + (int) (this.offset ^ (this.offset >>> 32)) ^ type.hashCode();
+            int result = super.hashCode();
+            result = 31 * result + type.hashCode();
+            result = 31 * result + offset;
+            return result;
         }
 
         /**
```

core/src/main/java/org/jruby/ext/ripper/StringTerm.java
```diff
@@ -146,7 +146,7 @@ public int parseString(RipperLexer lexer, LexerSource src) throws IOException {
     }
 
     private String parseRegexpFlags(RipperLexer lexer, LexerSource src) throws IOException {
-        StringBuilder buf = new StringBuilder(end);
+        StringBuilder buf = new StringBuilder();
```
headius: I don't understand this change.

core/src/main/java/org/jruby/ir/interpreter/FullInterpreterContext.java
```diff
@@ -188,6 +189,6 @@ public int determineRPC(int ipc) {
             if (ipc <= rescueIPCs[i]) return rescueIPCs[i + 1];
         }
 
-        throw new RuntimeException("BUG: no RPC found for " + getFileName() + ":" + getName() + ":" + ipc + getInstructions());
+        throw new RuntimeException("BUG: no RPC found for " + getFileName() + ":" + getName() + ":" + ipc + Arrays.toString(getInstructions()));
```
headius: This is wrong both ways. The array will just be a mangled `Instr]@0x1234` sort of toString, and the Arrays version will include ALL of the instructions in the error message.

core/src/main/java/org/jruby/ir/operands/SymbolProc.java
```diff
@@ -34,7 +34,7 @@ public Object createCacheObject(ThreadContext context) {
 
     @Override
     public int hashCode() {
-        return 47 * 7 + (int) (this.name.hashCode() ^ (this.encoding.hashCode() >>> 32));
+        return 47 * 7 + (int) (this.name.hashCode() ^ (this.encoding.hashCode() >>> 16));
```
headius: Same reason for fix and argument for generated hashCode as for StructLayout.

core/src/main/java/org/jruby/java/codegen/RealClassGenerator.java
```diff
@@ -144,7 +144,7 @@ public static Class createRealImplClass(Class superClass, Class[] interfaces, Ru
 
         Class newClass = defineRealImplClass(ruby, name, superClass, superTypeNames, simpleToAll);
         if (!newClass.isAssignableFrom(interfaces[0])) {
-            new RuntimeException(newClass.getInterfaces()[0].getClassLoader() + " " + interfaces[0].getClassLoader());
+            throw new RuntimeException(newClass.getInterfaces()[0].getClassLoader() + " " + interfaces[0].getClassLoader());
```
headius: This should never happen anyway, but it's a good change.

core/src/main/java/org/jruby/util/io/SelectExecutor.java
```diff
@@ -209,6 +209,7 @@ IRubyObject selectInternal(ThreadContext context) throws IOException {
                 RubyIO io = TypeConverter.ioGetIO(runtime, obj);
                 RubyIO write_io = io.GetWriteIO();
                 fptr = io.getOpenFileChecked();
+                /* BUG: errorKeyList (SelectionKey) can't contain fptr.fd() (ChannelFD) */
```
headius: This should be excluded. I suppose it objects to a comment starting with "BUG".

core/src/main/java/org/jruby/java/codegen/RealClassGenerator.java
nglorioso: Looks like there a number of failures on the Travis build that are triggered by this change: https://travis-ci.org/jruby/jruby/builds/305072247

Haven't dug into this, but I can revert this change if there are too many failing tests to deal with.

core/src/main/java/org/jruby/util/io/SelectExecutor.java
nglorioso: Yeah, sorry, should have cleared up during the PR comment.

There exists a bug here, and I wanted to mark it. I'm happy to remove the comment from this PR. Should I file a distinct issue to look into this?

core/src/main/java/org/jruby/ir/interpreter/FullInterpreterContext.java
nglorioso: Hrm. Perhaps do you want me to replace `getInstructions()` with `rescueIPCs` and Arrays.toString that?

core/src/main/java/org/jruby/ext/ripper/StringTerm.java
enebo: I made a general comment in the PR about this.  The result of that string ends up appended to another StringBuilder (which indicates perhaps this could get cleaned up) but 'end' should definitely be in the string.  It was likely me assuming character was an acceptable constructor and because this was ported C code we saved the char as an int and confusion ensued.

core/src/main/java/org/jruby/ir/interpreter/FullInterpreterContext.java
enebo: @nglorioso just remove getInstructions() altogether.  Originally we only saw this while developing our new runtime so we probably never cared about anything other than ipc since we could just run the program again and look at the instructions on a second run.  If we corrected this to print out all instrs I think users would complain about a gazillion lines of output.

enebo: As an added note we should consider possibly changing this to be an hserr-like output in addition to notifying the user we hit an internal bug.  We could then dump CFG, instrs, ipc, etc into a file so they could paste that into an issue (sorry for adding this to an incidental PR review but sometimes inspiration strikes!).  I am not asking that here though :)

core/src/main/java/org/jruby/ext/ripper/StringTerm.java
enebo: @nglorioso Actually just remove this.  The existing code should be passing in the StringBuilder from the previous builder in.  I will refactor this a tiny bit and that StringBuilder will go away.

enebo: Addressed in 58cd1331cb7bdc478b60d29d554f4206fd36e981 on jruby-9.1 and cpd over to master.

core/src/main/java/org/jruby/util/io/SelectExecutor.java
nglorioso: FIled #4867 for this.

core/src/main/java/org/jruby/java/codegen/RealClassGenerator.java
nglorioso: Still waiting on this. Can revert the change to not shake up the CI, or leave it in :)

headius: Revert this change and file a bug about this useless line.

core/src/main/java/org/jruby/ext/ripper/RipperLexer.java
```diff
@@ -2349,7 +2350,7 @@ public boolean tokenAddMBC(int first_byte, ByteList buffer) {
         int length = precise_mbclen();
 
         if (length <= 0) {
-            compile_error("invalid multibyte char (" + getEncoding().getName() + ")");
+            compile_error("invalid multibyte char (" + Arrays.toString(getEncoding().getName()) + ")");
```
headius: I didn't notice this before, but this should just be `getEncoding()`. The default toString for Encoding is to return the name of the encoding.

core/src/main/java/org/jruby/util/io/SelectExecutor.java
headius: Thank you. Remove this from the PR and we'll discuss it separately.

core/src/main/java/org/jruby/java/codegen/RealClassGenerator.java
nglorioso: #4880 filed

nglorioso:
Thanks! Extraneous things removed, should be good to merge (assuming no other CI failures)

headius:
Thanks for your patience and help!

nglorioso:
Thank you! Glad to help improve things :)

PR #4881, opened 2017-12-01
Don't unconditionally emit debug level info from RACC

haus:
In
https://github.com/jruby/jruby/commit/3fe1cc2353e8b051a7541d3cb63d5210e99d8539,
jruby's version of racc/parser.rb was brought in line with MRI
racc/parser.rb. This brought along a change to the yyparse method which
changed the last argument from false to true. This causes the java racc
parser implementation to unconditionally emit very noisy debug output
whenever yyparse is called, where previously it would not emit such
output. Passing true is safe for the pure ruby parser, as it ignores the
argument entirely and does nothing with it. It is also safe for the
native c parser, as that parser only emits debug info if it is compiled
with DEBUG set. Tenderlove's version of racc disabled debug by default
in
https://github.com/tenderlove/racc/commit/c5359ab04b5ad858d59664fbe288e9cc7fb1c27f,
but it looks like MRI never picked up that change.
For a time, true was passed and debugging info was not emitted (between
9.1.9.0 and 9.1.12.0). This was because in the merge to bring
racc/parser.rb in line with MRI, the loading of the java cparse library
no longer happened, so the pure ruby version of the library was used
which ignores the debug parameter entirely. This changed in
https://github.com/jruby/jruby/commit/baeaad12f32842a5b714c463558feeb2f7b34c0b
when the cparse-java.jar was being loaded again for jruby.

### DIFF
--- a/lib/ruby/stdlib/racc/parser.rb
+++ b/lib/ruby/stdlib/racc/parser.rb
@@ -322,7 +322,7 @@ module Racc
     # RECEIVER#METHOD_ID is a method to get next token.
     # It must 'yield' the token, which format is [TOKEN-SYMBOL, VALUE].
     def yyparse(recv, mid)
-      __send__(Racc_YY_Parse_Method, recv, mid, _racc_setup(), true)
+      __send__(Racc_YY_Parse_Method, recv, mid, _racc_setup(), false)
     end
 
     def _racc_yyparse_rb(recv, mid, arg, c_debug)


haus:
As an example, Puppet leverages yyparse at runtime. With jruby 9.1.14.0 from homebrew, simply running `jruby -rpuppet` dumps approximately 18000 lines of yyparse debug output to the console. 9.1.12.0 and earlier do not have this behavior.

headius:
Good find! Perhaps you should file an issue with MRI to get them on the same page? They are gemifying libraries for 2.5, but I don't believe racc is one of them (so it won't get updated unless they update it manually).

haus:
Ah thanks. I'll open a ticket on their end as well.

headius:
@haus Link it here when you do. Thanks again!

haus:
@headius i opened https://bugs.ruby-lang.org/issues/14158 for the general case of update racc with upstream changes. if you think i should be more specific and ask for this particular change to be brought over, i can update the ticket.

Issue #4879, opened 2017-12-01
Performance issue with ruby meta programming


GaneshSPatil:
### Environment

* JRuby version (`jruby -v`): `jruby 1.7.26 (1.9.3p551) 2016-08-26 69763b8 on Java HotSpot(TM) 64-Bit Server VM 1.8.0_101-b13 +jit [darwin-x86_64]`
* Operating system and platform (e.g. `uname -a`):  `Darwin **.local 17.2.0 Darwin Kernel Version 17.2.0: Fri Sep 29 18:27:05 PDT 2017; root:xnu-4570.20.62~3/RELEASE_X86_64 x86_64`

### Description
Extensively Using [ruby `extend` functionality](https://ruby-doc.org/core-2.4.2/Object.html#method-i-extend) blocks threads which in turns causes server performance issues.
`extend` method calls [jruby `addSubclass` function](https://github.com/jruby/jruby/blob/master/core/src/main/java/org/jruby/RubyClass.java#L1125-L1131) which is a bottleneck.

### Background
Many libraries that provide JSON serialization make use of `extend`. Under heavy load, this causes performance degradation.

### Code Snippet to Reproduce Issue:
```
module Bar
  def hello
    p "hello from bar!"
  end
end

class Foo
  def hello
    p "hello!"
  end
end

threads = 1000.times.collect do
  Thread.new do
    10000.times do |i|
      foo = Foo.new
      foo.extend(Bar)
    end
  end
end

threads.each {|t| t.join }
```

Run the above code snippet and capture thread dump. You will see a lot of threads waiting to enter synchronized [this](https://github.com/jruby/jruby/blob/master/core/src/main/java/org/jruby/RubyClass.java#L1125-L1131) block.

[here](https://gist.github.com/GaneshSPatil/b645d0af258a024d24b57ffa4e92bcf8) is an example thread dump.

ketan:
For perspective, this is locking up a significant number of jetty web threads because of `object.extend` being invoked by a ruby gem.

kares:
... is this yet another report against (the no longer supported) JRuby 1.7.x or did you guys repro on 9K ?

kares:
got it, that [`addSubclass`](https://github.com/jruby/jruby/blob/6fe0e5a1a93b006e/core/src/main/java/org/jruby/RubyClass.java#L1125-L1131) piece should not have changed since 1.7 so it should be an issue on 9K as well
[...]

PR #4886, opened 2017-12-07
Improvements to subclass management and method cache invalidation

headius:
In #4879, we got a report of heavy thread contention on the `addSubclasses` method of RubyClass. After investigating the locking involved (at @kares's recommendation), it seems as though much of this locking is excessively restrictive, using a runtime-global lock to prevent multiple threads from modifying any class's list of subclasses at the same time.

The subclasses list is also used heavily by method invalidation, and exploring that end of things showed me many places where we could improve performance.

* The global synchronization may be unnecessary most of these places, so long as the collections being modified are themselves thread-safe.
  I have removed most of this but we need much more testing, review, analysis to confirm it's safe.
* The subclass aggregation logic creates many intermediate lists, which can be reduced to a single list.
  I do not believe this code is used in any hot paths (it's mainly there for ObjectSpace) but I modified the logic to only create a single list for all subclasses and descendants in a given subhierarchy.
* The invalidation logic must aggregate invalidators from all subclasses.
  It does this by creating a list of invalidators and adding every invalidator from that class on down, recursively. However, this list is usually all GenerationAndSwitchPointInvalidator, and the SwitchPoints are later reprocessed into another array for bulk invalidation. These collection could be cached, thrown out when any descendant is added or removed by calling up-hierarchy. This would reduce the cost of invalidation, but have a higher in-memory cost. It would also provide no benefits (but no detriments) running on a system that creates many classes.
  This work has yet to be done, and I have not investigated the actual cost of the invalidator gathering.

### DIFF
--- a/core/src/main/java/org/jruby/RubyClass.java
+++ b/core/src/main/java/org/jruby/RubyClass.java
@@ -1097,19 +1097,27 @@ public class RubyClass extends RubyModule {
         return subclasses(false);
     }
 
-    public synchronized Collection<RubyClass> subclasses(boolean includeDescendants) {
+    public Collection<RubyClass> subclasses(boolean includeDescendants) {
         Set<RubyClass> subclasses = this.subclasses;
         if (subclasses != null) {
-            Collection<RubyClass> mine = new ArrayList<>(subclasses);
-            if (includeDescendants) {
-                for (RubyClass klass: subclasses) {
-                    mine.addAll(klass.subclasses(includeDescendants));
-                }
-            }
+            Collection<RubyClass> mine = new ArrayList<>();
+            subclassesInner(mine, includeDescendants);
 
             return mine;
         }
         return Collections.EMPTY_LIST;
+    }
+
+    private void subclassesInner(Collection<RubyClass> mine, boolean includeDescendants) {
+        Set<RubyClass> subclasses = this.subclasses;
+        if (subclasses != null) {
+            mine.addAll(subclasses);
+            if (includeDescendants) {
+                for (RubyClass klass: subclasses) {
+                    klass.subclassesInner(mine, includeDescendants);
+                }
+            }
+        }
     }
 
     /**
@@ -1121,12 +1129,19 @@ public class RubyClass extends RubyModule {
      *
      * @param subclass The subclass to add
      */
-    public synchronized void addSubclass(RubyClass subclass) {
-        synchronized (runtime.getHierarchyLock()) {
-            Set<RubyClass> subclasses = this.subclasses;
-            if (subclasses == null) this.subclasses = subclasses = new WeakHashSet<>(4);
-            subclasses.add(subclass);
+    public void addSubclass(RubyClass subclass) {
+        Set<RubyClass> subclasses = this.subclasses;
+        if (subclasses == null) {
+            // check again
+            synchronized (this) {
+                subclasses = this.subclasses;
+                if (subclasses == null) {
+                    this.subclasses = subclasses = Collections.synchronizedSet(new WeakHashSet<RubyClass>(4));
+                }
+            }
         }
+
+        subclasses.add(subclass);
     }
 
     /**
@@ -1134,13 +1149,11 @@ public class RubyClass extends RubyModule {
      *
      * @param subclass The subclass to remove
      */
-    public synchronized void removeSubclass(RubyClass subclass) {
-        synchronized (runtime.getHierarchyLock()) {
-            Set<RubyClass> subclasses = this.subclasses;
-            if (subclasses == null) return;
+    public void removeSubclass(RubyClass subclass) {
+        Set<RubyClass> subclasses = this.subclasses;
+        if (subclasses == null) return;
 
-            subclasses.remove(subclass);
-        }
+        subclasses.remove(subclass);
     }
 
     /**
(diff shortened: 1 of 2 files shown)


headius:
@kares @mkristian @lopex @enebo This is a work in progress, but I'll definitely need help reviewing and testing these changes. Anything you can suggest or do to help will be appreciated :-)

ketan:
Perhaps I'm wrong — but looking at this PR, a lot of it would likely simplify if `this.subclasses` were initialized to an empty concurrent set.

ketan:
> Perhaps I'm wrong — but looking at this PR, a lot of it would likely simplify if this.subclasses were initialized to an empty concurrent set.

Ignore me. I think I understand why it can't be.

headius:
I'm not sure what you were thinking, but I guess it would be no better than initializing to null. And the reason we leave such fields blank is so there's no overhead for classes that never have subclasses. It's worth some thought though; elsewhere in the codebase we separate read and write usage so reads can be mostly noops when there's nothing to do.

kares:
LGTM

GaneshSPatil:
Hi,

We managed to build our app with this PR, the current bottleneck seems to have been resolved, but we have discovered another similar bottleneck with another synchronized method, which is blocking about 278 threads and grinding the server to a halt as we speak.

We have attached the thread dumps at http://fastthread.io/my-thread-report.jsp?p=c2hhcmVkLzIwMTcvMTIvOC8tLWZpbGVfMi0tNS00My0zODs7LS1maWxlXzItLTUtNDUtMTE= if you'd like to take a look.

```
qtp951880373-287636 - priority:5 - threadId:0x00007fc7a4150800 - nativeId:0x5a88 - state:RUNNABLE
stackTrace:
java.lang.Thread.State: RUNNABLE
at java.lang.invoke.MethodHandleNatives.setCallSiteTargetNormal(Native Method)
at java.lang.invoke.CallSite.setTargetNormal(CallSite.java:258)
at java.lang.invoke.MutableCallSite.setTarget(MutableCallSite.java:154)
at java.lang.invoke.SwitchPoint.invalidateAll(SwitchPoint.java:224)
at org.jruby.runtime.opto.SwitchPointInvalidator.invalidate(SwitchPointInvalidator.java:42)
- locked <0x00000006c9c95ef8> (a org.jruby.runtime.opto.ConstantInvalidator)
at org.jruby.runtime.opto.ConstantInvalidator.invalidate(ConstantInvalidator.java:20)
at org.jruby.RubyModule.invalidateConstantCache(RubyModule.java:1554)
at org.jruby.RubyModule.invalidateConstantCacheForModuleInclusion(RubyModule.java:1358)
at org.jruby.RubyModule.includeModule(RubyModule.java:889)
- locked <0x00000006996b50a8> (a org.jruby.MetaClass)
at org.jruby.RubyModule.extend_object(RubyModule.java:2581)
at org.jruby.RubyModule$INVOKER$i$1$0$extend_object.call(RubyModule$INVOKER$i$1$0$extend_object.gen)
at org.jruby.RubyClass.finvoke(RubyClass.java:908)
at org.jruby.runtime.Helpers.invoke(Helpers.java:421)
at org.jruby.RubyBasicObject.callMethod(RubyBasicObject.java:363)
at org.jruby.RubyBasicObject.extend(RubyBasicObject.java:2634)
at org.jruby.RubyKernel.extend(RubyKernel.java:1977)
at org.jruby.RubyKernel$INVOKER$s$0$0$extend.call(RubyKernel$INVOKER$s$0$0$extend.gen)
at org.jruby.internal.runtime.methods.JavaMethod$JavaMethodN.call(JavaMethod.java:721)
at org.jruby.internal.runtime.methods.DynamicMethod.call(DynamicMethod.java:200)
at org.jruby.runtime.callsite.CachingCallSite.call(CachingCallSite.java:153)
at var.go.server.go_minus_server.go_minus_server_minus_17_dot_12_dot_0.work.jetty_minus_0_dot_0_dot_0_dot_0_minus_8153_minus_cruise_dot_war_minus__go_minus_any_minus_.webapp.WEB_minus_INF.rails_dot_new.vendor.bundle.jruby.$2_dot_3_dot_0.gems.representable_minus_2_dot_1_dot_7.lib.representable.invokeOther0:extend(/var/go/server/go-server/go-server-17.12.0/work/jetty-0.0.0.0-8153-cruise.war-_go-any-/webapp/WEB-INF/rails.new/vendor/bundle/jruby/2.3.0/gems/representable-2.1.7/lib/representable.rb:105)
```

kares:
@GaneshSPatil that seems to be the `extend` explained in your original report at #4879 
constant cache invalidation comes with a cost, however these changes still might make sense for others ...

p.s. me walking in your shoes ... would definitely be looking into refactoring/patching that `extend` piece

enebo:
LGTM but since 9.1.15.0 is out lets put this in today to give it a month to break something.

headius:
@GaneshSPatil This additional case appears to be due to the cost of walking all constants neededing to be invalidated and invalidating them individually. There will be one for every name, but the code (invalidateConstantCacheForModuleInclusion) here does a couple things wrong:

* It walks all constants in all supermodules/superclasses of the incoming module, which will almost certainly invalidate some constants more than once.
* It invalidates each name individually, which comes with locking and JVM JIT impact each time.
* Because it walks all the way up to Kernel, it will invalidate all constants in the system every time, regardless of whether they're redefined or not.

I have a fix for the first two items that should reduce how much locking and JIT impact this has.

The third item is a bit trickier. We will need to experiment a bit to figure out which constants really need to be invalidated.

headius:
@GaneshSPatil Please retest.

GaneshSPatil:
Hello @headius,
Apology for the delayed response.

We tested the changes with the same performance setup we have. We don't any blocked threads as of now, but the performance has degraded going from `1.7.26` to `9.1.15`.

One of our APIs which does JSON serialization used to take 12s to 15s with `JRuby v1.7.26`,  with the latest `JRuby v9.1.15` (including this PR patch) takes around 45s to a minute to respond. 
The timings are captured on an idle server. On a heavily loaded server, we expect the APIs to take more time to respond.

We'll investigate more into the performance degradation and will provide you an update.

GaneshSPatil:
@headius -- We didn't see any blocked threads around `addSubClass` method with these changes. So we think this PR is good to merge which resolves the original issue #4879 .

>  We'll investigate more into the performance degradation and will provide you an update.

We have upgraded rails from `4.x.x` to `5.x.x`, and JRuby from `1.7.26` to `9.1.15`. We'll investigate what exactly has caused a performance hit and provide you an update.

headius:
@GaneshSPatil Ok thank you. You may get some good info about the performance issue by using `--sample` flag to JRuby. This uses a JVM-provided sampling profiler which outputs results on exit. These results may show if there's an obvious bottleneck.

enebo:
whoops we planned on landing this much earlier but we should have some time left before 9.1.16 so it is in!

Issue #2824, opened 2015-04-13
fail to check CSRF at Rails4.2


yousuketto:
I use jruby 1.7.19.
I failed to check CSRF at Rails4.2 by raised ```TypeError [can't convert nil into Integer]```.
https://github.com/rails/rails/blob/v4.2.1/actionpack/lib/action_controller/metal/request_forgery_protection.rb#L321

#### Simple reproduction code
```ruby
# example.rb
ths = []
s1 = "XXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXX"
s2 = "YYYYYYYYYYYYYYYYYYYYYYYYYYYYYYYY"
10_000.times do
  t = Thread.new do
    number_ary = []
    begin
      s1.bytes.zip(s2.bytes).map { |(c1,c2)| number_ary << [c1 , c2]; c1 ^ c2 }
    rescue Exception
      print "#{$!.inspect}\n#{number_ary.inspect}\n"
    end
  end
  ths << t
end

ths.each(&:join)
```
*result*
```
C:\workspace\sandbox\ruby-byte>jruby example.rb
#<TypeError: can't convert nil into Integer>
[[88, 89], [88, nil]]
#<TypeError: can't convert nil into Integer>
[[88, 89], [88, 89], [88, 89], [88, 89], [88, 89], [88, 89], [88, 89], [88, 89], [88, 89], [88, 89], [88, 89], [88, 89], [88, 89], [88, 89], [88, 89], [88, 89], [88, 89], [88, 89], [88, 89], [88, 89], [88, 89], [88, nil]]
```

yousuketto:
Sorry, I was not looking. 
Maybe, this is the same isuee as #2577.

Issue #2577, opened 2015-02-07
Enumerator out of memory error


iconara:
I've found that it's possible to stress JRuby into crashing with an out of memory error with the following code.

The code zips the bytes of two strings together, their lengths don't matter at all, a single character is sufficient. It then checks whether any of the bytes are `nil`, which should be impossible, but happens. Originally I didn't check for `nil`, but the first indication that there was a problem was that I got errors that I did things with `nil` where there couldn't be any `nil`. When I put a `begin…rescue` around it to see what was `nil` I got an out of memory error instead.

```ruby
s1 = 'a'
s2 = 'b'

100000.times do
  b1 = s1.each_byte
  b2 = s2.each_byte
  bytes = b1.zip(b2).flatten
  if bytes.any? { |b| b.nil? }
    puts('this can never happen')
  end
end
```

prints the following in JRuby 1.7.18, 1.7.19 and HEAD (probably all other versions too):

```
this can never happen
this can never happen
this can never happen
this can never happen
this can never happen
Error: Your application used more memory than the safety cap of 500M.
Specify -J-Xmx####m to increase it (#### = cap size in MB).
Specify -w for full OutOfMemoryError stack trace
```

the exact number of "this can never happen" differ.

The full stack trace of the `OutOfMemoryError` is:

```
java.lang.OutOfMemoryError: unable to create new native thread
	at java.lang.Thread.start0(Native Method)
	at java.lang.Thread.start(Thread.java:713)
	at java.util.concurrent.ThreadPoolExecutor.addWorker(ThreadPoolExecutor.java:950)
	at java.util.concurrent.ThreadPoolExecutor.execute(ThreadPoolExecutor.java:1368)
	at java.util.concurrent.AbstractExecutorService.submit(AbstractExecutorService.java:112)
	at org.jruby.RubyEnumerator$ThreadedNexter.ensureStarted(RubyEnumerator.java:700)
	at org.jruby.RubyEnumerator$ThreadedNexter.next(RubyEnumerator.java:654)
	at org.jruby.RubyEnumerator.next(RubyEnumerator.java:461)
	at org.jruby.RubyEnumerator$INVOKER$i$0$0$next.call(RubyEnumerator$INVOKER$i$0$0$next.gen)
	at org.jruby.RubyClass.finvoke(RubyClass.java:616)
	at org.jruby.runtime.Helpers.invoke(Helpers.java:593)
[...]

PR #4887, opened 2017-12-07
ThreadedNexter shutdown can in some situations interrupt a thread executing a different task

jmalves:
It can happen that the `ThreadedNexter#shutdown` from _RubyEnumerator_ sends interrupts to a thread that is executing a different task.

`ThreadedNexter#shutdown` can get the thread value before the _Nexter_ thread is finished but send the interrupts after the _Nexter_ thread is already done with its work. In case that the _Nexter_ thread is already assigned another task those interrupts can cause problems.

From what I could see it seems it affects all JRuby versions with `ThreadedNexter`.

To reproduce the issue:

```ruby
threads = []

200.times do
  threads << Thread.new do
    enumerable1 = (1..1000)
    enumerable2 = (1..1000)
    100.times do
      enumerable1.zip(enumerable2).map { |(a, b)| a^b }
    end
  end
end

threads.each(&:join)
```

Related to #2577 and #2824

### DIFF
--- a/core/src/main/java/org/jruby/RubyEnumerator.java
+++ b/core/src/main/java/org/jruby/RubyEnumerator.java
@@ -745,20 +745,28 @@ public class RubyEnumerator extends RubyObject implements java.util.Iterator<Obj
 
             // mark for death
             die = true;
+            if (dissociateNexterThread(true)) doneObject = null;
+        }
 
-            Thread myThread = thread;
-            if (myThread != null) {
-                if (DEBUG) System.out.println("clearing for shutdown");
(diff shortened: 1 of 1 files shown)


headius:
Very nice find! There have been lingering intermittent interrupted threads reported occasionally...this could have been the culprit.

I'll have a look.

headius:
It looks good, but I will merge manually to the 9.1 branch so it can be in 9.1.16.

headius:
Merged to 9.1 in 3011d3d89f.

headius:
Thank you!

PR #4900, opened 2017-12-16
Improve compatibility with MRI's Ripper for JRuby 9.1

grddev:
Same as #4898, but targeted on jruby-9.1 branch instead.

The main difference is that it skips https://github.com/jruby/jruby/pull/4898/commits/1763e4e3c1a5d121cec042663d60bc5309c6d379 that aligned the main parser and Ripper parser, as it seems the misalignment doesn't exist on this branch.


enebo:
@grddev coolio...I will look at regexp PR now!

PR #4902, opened 2017-12-17
Add support for validating Regexp in Ripper

grddev:
This represents a stab at implementing the remaining Ripper compatibility as mentioned in #4898.

This uses the fact that Regexp tokenization is handled by a single StringTerm, and thus all tSTRING_CONTENT fragments are easily collectable until the tREGEXP_END comes with the options that we need for validation.

The validation itself is a copied/simplified version of what is performed by the main parser, as large parts the validation depended on the AST structure, which we do not have here.

Technically, this doesn't perform the validation at the same point in time as the main parser, as it performs the validation when encountering the tREGEXP_END token rather than when processing the regexp rule.

I speculate that the difference doesn't really matter given that the only thing we could do with the tREGEXP_END token is to apply the regexp rule.

In order to reduce copy-paste between the main parser and Ripper, I opted to shift some Regexp-related code into the Lexer. In theory, the code doesn't belong in the lexer, but putting it in the Lexer has some benefits. First, it is a component that is shared in a reasonable way between the two parsers. Second, it is essentially required by the proposed implementation, as the new validation takes place effectively inside the Lexer.

Unsurprisingly, it turns out that the coverage for Ripper parsing of Regexp isn't very extensive, and I haven't had time to put the code through any additional tests.

### DIFF
--- a/core/src/main/java/org/jruby/ext/ripper/StringTerm.java
+++ b/core/src/main/java/org/jruby/ext/ripper/StringTerm.java
@@ -29,7 +29,7 @@ package org.jruby.ext.ripper;
 
 import java.io.IOException;
 import java.util.ArrayList;
-import java.util.Collection;
+import java.util.List;
 import org.jcodings.Encoding;
 import org.jruby.Ruby;
 import org.jruby.lexer.LexerSource;
@@ -52,7 +52,8 @@ public class StringTerm extends StrTerm {
     // How many strings are nested in the current string term
     private int nest;
 
-    private Collection<ByteList> regexpFragments;
+    private List<ByteList> regexpFragments;
+    private boolean regexpDynamic;
 
     public StringTerm(int flags, int begin, int end) {
         this.flags = flags;
@@ -89,17 +90,14 @@ public class StringTerm extends StrTerm {
     private void validateRegexp(RipperLexer lexer) throws IOException {
         Ruby runtime = lexer.getRuntime();
         RegexpOptions options = lexer.parseRegexpFlags();
-        ByteList last = null;
         for (ByteList fragment : regexpFragments) {
-            last = fragment;
-            if (fragment != null) {
-                lexer.checkRegexpFragment(runtime, fragment, options);
-            }
+            lexer.checkRegexpFragment(runtime, fragment, options);
         }
-        if (last != null && regexpFragments.size() == 1) {
-            lexer.checkRegexpSyntax(runtime, last, options);
+        if (!regexpDynamic && regexpFragments.size() == 1) {
+            lexer.checkRegexpSyntax(runtime, regexpFragments.get(0), options);
         }
         regexpFragments.clear();
+        regexpDynamic = false;
     }
 
     @Override
@@ -137,7 +135,7 @@ public class StringTerm extends StrTerm {
 
             if (token != 0) {
                 if ((flags & STR_FUNC_REGEXP) != 0) {
-                    regexpFragments.add(null);
+                    regexpDynamic = true;
                 }
                 return token;
             } else {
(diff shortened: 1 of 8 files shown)


grddev:
Forgot to mention, but this does nothing to address local variables as mentioned in #4898, but some quick testing seemed to indicate that MRI didn't handle this either (as the below output has a vcall rather than a var_ref towards the end.

```
% RBENV_VERSION=2.4.2 ruby -rripper -e 'p Ripper.sexp("/\$(?<dollars>\d+)\.(?<cents>\d+)/ =~ \"$3.67\"\ndollars")'
[:program, [[:binary, [:regexp_literal, [[:@tstring_content, "$(?<dollars>d+).(?<cents>d+)", [1, 1]]], [:@regexp_end, "/", [1, 29]]], :=~, [:string_literal, [:string_content, [:@tstring_content, "$3.67", [1, 35]]]]], [:vcall, [:@ident, "dollars", [2, 0]]]]]
```

core/src/main/java/org/jruby/ext/ripper/StringTerm.java
```diff
@@ -69,11 +78,30 @@ private int endFound(RipperLexer lexer) throws IOException {
...
+        Ruby runtime = lexer.getRuntime();
+        RegexpOptions options = lexer.parseRegexpFlags();
+        ByteList last = null;
+        for (ByteList fragment : regexpFragments) {
+            last = fragment;
+            if (fragment != null) {
+                lexer.checkRegexpFragment(runtime, fragment, options);
+            }
+        }
+        if (last != null && regexpFragments.size() == 1) {
```
enebo: Is this only possible for /#{ffofofofo}/ and /@{gfogfogogog}/?  If so can you add a comment here as to what scenario this is for?  Without printing out the lex stream I would almost think this was not possible.  A comment will help later on.

core/src/main/java/org/jruby/lexer/LexingCommon.java
```diff
@@ -1011,4 +1016,120 @@ protected void onMagicComment(String name, ByteList value) {
...
+        } else if (getEncoding() == USASCII_ENCODING) {
+            if (!is7BitASCII(value)) {
+                value.setEncoding(USASCII_ENCODING); // This will raise later
+            } else {
+                value.setEncoding(ASCII8BIT_ENCODING);
+            }
+        }
+    }
+
+    private boolean is7BitASCII(ByteList value) {
```
enebo: Note for myself.  We should make sure CR is calculated as we build up string so we do not need to rescan all bytelists after creation.

core/src/main/java/org/jruby/parser/ParserSupport.java
```diff
@@ -1453,13 +1453,8 @@ public Node arg_append(Node node1, Node node2) {
 
     // MRI: reg_fragment_check
     public void regexpFragmentCheck(RegexpNode end, ByteList value) {
```
enebo: Since 9.2 will be considered a major version you can remove this from ParserSupport if you want (or not) and just make callers call lexer directly.  Optional.

core/src/main/java/org/jruby/ext/ripper/StringTerm.java
grddev: Rewrote the implementation with an additional `regexpDynamic` variable instead to make the logic simpler to follow instead on commenting on the cryptic logic.

core/src/main/java/org/jruby/parser/ParserSupport.java
grddev: Done

enebo:
@grddev sorry our last merge broke this.  I did not notice you had addressed the comments so I spaced out landing this.  Can you update it?

grddev:
Rebased the changes on top of master

enebo:
@grddev thanks. sorry I lost this one in the weeds :)

Issue #4876, opened 2017-11-29
Ruby 2.5 Support


headius:
This issue will track Ruby 2.5 support for JRuby. This will be released in JRuby 9.3.0.0 or JRuby 10.

As with previous releases, we will base our checklist on MRI's NEWS file: https://github.com/ruby/ruby/blob/trunk/NEWS
This bug will track all NEWS items new in Ruby 2.5. This list is based off https://github.com/ruby/ruby/blob/trunk/NEWS.

Only changes relevant to JRuby are listed here. MRI-specific internal changes and features are not included.

NOTE: Pull requests should be done against the `ruby-2.5` branch.

### Language changes

- [x] Top-level constant look-up is removed.  [Feature #11547](https://bugs.ruby-lang.org/issues/11547)
- [x] rescue/else/ensure are allowed inside do/end blocks.  [Feature #12906](https://bugs.ruby-lang.org/issues/12906)
- [ ] refinements take place in string interpolations.  [Feature #13812](https://bugs.ruby-lang.org/issues/13812)

#### Array

- [x]  `Array#append`  [Feature #12746](https://bugs.ruby-lang.org/issues/12746)
- [x]  `Array#prepend`  [Feature #12746](https://bugs.ruby-lang.org/issues/12746)

#### Data

- [x] Now deprecated [Feature #3072](https://bugs.ruby-lang.org/issues/3072)

#### Dir

- [x] `Dir.glob` provides new optional keyword argument, `:base`. [Feature #13056](https://bugs.ruby-lang.org/issues/13056)
- [x] `Dir.children`  [Feature #11302](https://bugs.ruby-lang.org/issues/11302)
- [x] `Dir.each_child`  [Feature #11302](https://bugs.ruby-lang.org/issues/11302)

#### File

- [ ]  :newline option to `File.open` implies text mode now.  [Bug #13350](https://bugs.ruby-lang.org/issues/13350)
- [x]  `File#path` now raises an IOError for files opened with `File::Constants::TMPFILE` option. [Feature #13568](https://bugs.ruby-lang.org/issues/13568)
- [x]  Add `File.lutime`  [Feature #4052](https://bugs.ruby-lang.org/issues/4052)

#### Hash

- [x] `Hash#transform_keys`  [Feature #13583](https://bugs.ruby-lang.org/issues/13583)
- [x] `Hash#transform_keys!`  [Feature #13583](https://bugs.ruby-lang.org/issues/13583)
[...]

PR #4904, opened 2017-12-18
Implement Integer.sqrt

nomadium:
Hi folks,

I was taking a look at open jruby issues and how to contribute with adding Ruby 2.5 support, so I prepared this small pull request.

I believe this is not a priority right now since Ruby 2.5 is not released yet, but I decided to contribute anyway, so I'm eager to receive feedback and check if I'm going in the right direction with this.

Please consider this as a WIP, since I could not see the results of the introduced tests in Travis although is working fine for me when I manually tested it in jirb.

Thanks,

### DIFF
--- a/core/src/main/java/org/jruby/RubyInteger.java
+++ b/core/src/main/java/org/jruby/RubyInteger.java
@@ -122,6 +122,19 @@ public abstract class RubyInteger extends RubyNumeric {
         return context.runtime.newBoolean(signum() > 0);
     }
 
+    /*  =============
+     *  Class Methods
+     *  =============
+     */
+
+    /** rb_int_s_isqrt
+     *
+     */
+    @JRubyMethod(meta = true)
+    public static IRubyObject sqrt(ThreadContext context, IRubyObject self, IRubyObject arg) {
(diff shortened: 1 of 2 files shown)


enebo:
@nomadium I believe the code is alright.  The tests themselves we actually get upstream from MRI.  So while I merged those tests in they will at some point get wiped out when we resync with MRI.

The solutions to this problem is either:
a) resubmit them to ruby-lang MRI project
b) rewrite them as specs for ruby/spec (which is getting a lot of contributions lately -- probably most preferred).
c) tell us you did enough hard work and we can try and push these tests upstream.

If these were gotten from upstream then I guess that is d) and we will get them next time we sync with their test suite :)

nomadium:
@enebo Oh, good to know about the tests. Don't worry about it since I just copied them from MRI 2.5 test suite, so you are right about option d).

Thanks for reviewing and accepting the contribution so fast. :)

enebo:
@nomadium yeah thanks for the contribution.  Keep going! :)

nomadium:
Add link to the Ruby 2.5 support tracking issue: https://github.com/jruby/jruby/issues/4876

PR #4905, opened 2017-12-19
Split frame fields to reduce pre/post logic

headius:
One of the primary perf hits in JRuby comes from methods that need a method frame to store data for downstream calls. An example: because there's one `[]` method in the system that needs to write the "backref" frame field, any method that calls `[]` on any type needs to have "backref" storage available.

Unfortunately, there's problems with how we manage this now:

* The need for any frame field force a full frame, with multiple writes in pre and post logic to set up and tear down that frame. If only one field is used, this is wasteful.
* There's no way to tell, at optimization time, that only one frame field is needed.

This PR starts the process of splitting frame push/pop logic to be more fine-grained, allowing for initializing only the parts of the frame we need.

At the moment, this PR does the following:

* Gathers more detailed information on what frame reads and writes might be associated with a given method name.
* Pushes this information through into the scope flags via the current flag calculation mechanism.
* Uses the reads and writes to know whether this scope needs a full frame or just a "backref frame" that only manipulates one field.

In experiments, for a method that contains a call to `[]` but otherwise does nothing, 30 million calls (in JIT) improve from 0.33s down to 0.08s or so, around 4x faster by removing unnecessary frame overhead.


subbuss:
So, the basic premise here is that the only source of use of some of the frame fields comes from the ruby stdlib, and for the JRuby impls in Java, those can be annotated, and optimized around.

Looks like this PR is asserting that this claim is true for the backref field.

Are there any frame fields where that premise doesn't hold?

subbuss:
i.e. can non-stdlib ruby methods directly (vs indirectly via use of stdlib methods) affect use of any of these frame fields?

subbuss:
> @subbuss Non-stdlib methods could only affect this if they were also defined in Java and annotated to access these fields.
I see, so default in the absence of annotations is requiring the full frame?

That default question comes from: can a pure ruby method have code that can affect use of any frame field? I suppose that only applies to the backref field since that is the only one you opt right now.

headius:
@subbuss It is not possible for pure Ruby code to access the caller's frame unless it is passed a binding or closure (which can be turned into a binding).

headius:
To clarify...a method can certainly have its own calls to `[]` or `binding` or whatever, but that's different than needing access to the *caller's* frame, like several core methods do. It's not possible to access the caller's frame from Ruby unless it was handed to you.

subbuss:
Ah, right. I forgot that detail .. been a while. :) Didn't look at the code, but the idea lgtm.

PR #4914, opened 2017-12-28
Make Kernel#pp available by default

nomadium:
Hi folks,

This is another feature targeting Ruby 2.5 [1]: Kernel#pp by default (feature #14123).

~~Note: the tests are copied from MRI.~~


olleolleolle:
@nomadium "Note: the tests are copied from MRI." Which tests are you referring to?

nomadium:
@olleolleolle It was a typo, sorry about that. I copy&pasted the body of another PR and forgot to remove the bit about the tests.

I scanned MRI source code to verify if they were testing this new feature, but I couldn't find any tests, although my search was not exhaustive. So, that's the reason why I didn't include test. If they are needed, just let me know.

PR #4917, opened 2017-12-28
Allow "-" for directory name in `TestThread`

yui-knk:
When I run this test class on Travis CI,
some tests fail because on Travis CI
directrory name includes my account name,
so inspected result becomes:

```
<"#<Thread:0x7cf50001@foo@/home/travis/build/yui-knk/jruby/test/jruby/test_thread.rb:368 run>">
```

which does not match `</\#\<Thread\:0x\h+@foo(@[\w\/\._]+\:\d+)?\srun\>/>`.


kares:
Thanks Yui.

PR #4918, opened 2017-12-30
Check meta-class of method objects in `equals`

yui-knk:
In MRI, `method_eq` checks CLASS_OF method objects

```
(CLASS_OF(method) != CLASS_OF(other))
```

By this change, `TestMethod#test_eq` will be passed.


kares:
nice find - makes sense.

PR #4919, opened 2017-12-30
Set MetaClass of cloned object

yui-knk:
In MRI, `method_clone` uses CLASS_OF method objects

```
clone = TypedData_Make_Struct(CLASS_OF(self), struct METHOD, &method_data_type, data);
```

By this change, `TestMethod#test_clone` will be passed.


kares:
Makes sense. Thanks Yui.

PR #4923, opened 2017-12-31
Include `#test_split_invalid_argument` to test targets

yui-knk:
Currently `#test_split_invalid_argument` does not fail.


kares:
:+1:

PR #4924, opened 2017-12-31
Fix test undef

yui-knk:
Fix a warning message
In MRI, this warning is
```
  -e:1: warning: undefining `object_id' may cause serious problems
```

Remove `__id__` from warning targets
`__id__` is not listed on warning targets in MRI.


kares:
makes sense - nice work, the cross-reference made it clear (wasn't sure where exactly to double check)

PR #4931, opened 2018-01-02
Refactoring `module_function`

yui-knk:
Remove `MODULE_FUNCTION` from `Visibility` and
add `isModuleFunction` to `Frame`.
Implementing `module_function` as `Visibility`
makes the visibility of `foo` to be `MODULE_FUNCTION`:

```
Module.new do
  module_function
  define_method(:foo) {:foo}
end
```

`module_function` is not a type of visibility
but states of frames. So change to manage it
on `Frame`.


yui-knk:
I introduced `FrameVisibility`, by this we can manage states of Frame without adding a new field to Frame.

yui-knk:
I want to revert last commit, because last commit breaks many test cases...

enebo:
@yui-knk do you know why they failed?  Removal of that enum value?

enebo:
@yui-knk I am unsure from reading comments what is needed to move forward on this PR?

PR #4934, opened 2018-01-03
Implement Array#prepend and Array#append aliases

ChrisBr:
to support Ruby 2.5, see https://bugs.ruby-lang.org/issues/12746.
#4876

It seems like tests already existed in 
```spec/ruby/core/array/append_spec.rb``` and ```spec/ruby/core/array/prepend_spec.rb```

PR #4935, opened 2018-01-04
Implement Array#prepend and Array#append aliases

ChrisBr:
to support Ruby 2.5, see https://bugs.ruby-lang.org/issues/12746.
#4876 

It seems like tests already existed in
spec/ruby/core/array/append_spec.rb and spec/ruby/core/array/prepend_spec.rb

Supersedes https://github.com/jruby/jruby/pull/4934


enebo:
@ChrisBr thanks for digging in!

PR #4936, opened 2018-01-04
Implement Dir.glob base option

ChrisBr:
https://bugs.ruby-lang.org/issues/13056
Tests c&p from https://github.com/ruby/ruby/commit/52419a6e1032c4fe8fe6f6f6e2858dcd8fa32ba6.
#4876

### DIFF
--- a/core/src/main/java/org/jruby/RubyDir.java
+++ b/core/src/main/java/org/jruby/RubyDir.java
@@ -62,6 +62,7 @@ import org.jruby.util.FileResource;
 import org.jruby.util.JRubyFile;
 import org.jruby.util.ByteList;
 import org.jruby.util.StringSupport;
+import org.jruby.util.TypeConverter;
 
 /**
  * .The Ruby built-in class Dir.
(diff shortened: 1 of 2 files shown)


core/src/main/java/org/jruby/RubyDir.java
```diff
@@ -206,17 +207,37 @@ private static ByteList globArgumentAsByteList(ThreadContext context, IRubyObjec
...
         List<ByteList> dirs;
         IRubyObject tmp = args[0].checkArrayType();
-        if (tmp.isNil()) {
-            dirs = Dir.push_glob(runtime, runtime.getCurrentDirectory(), globArgumentAsByteList(context, args[0]), flags);
+
+        if (!base.isEmpty() && !(new File(base).exists())){
+            dirs = new ArrayList<ByteList>();
+        } else if (tmp.isNil()) {
+            dir = base.isEmpty() ? runtime.getCurrentDirectory() : base;
+            dirs = Dir.push_glob(runtime, dir, globArgumentAsByteList(context, args[0]), flags);
```
ChrisBr: Wondering what the difference between ``runtime.getCurrentDirectory()`` and ``getCWD(runtime)`` or is it possible to simplify the code and only use either or?

enebo: @ChrisBr getCWD will force canonicalization whereas runtime.getCurrentDirectory is whatever it was set to (which might have already been canonicalized).  Even if we happen to always canon. the stored value getCWD always tries to do it again (unless it is an URI).

Dir.glob is a dirty beast and I do not remember what contract it has for the incoming directory but that would be the proper guidance on which to use.

PR #4937, opened 2018-01-05
Fix "Complex#finite? returns false for NaN FAILED"

yui-knk:
The behavior of `Complex#finite?` for NaN was fixed on
Ruby 2.5.

```
$ ruby -ve "p Complex(Float::NAN, Float::NAN).finite?"
ruby 2.5.0p0 (2017-12-25 revision 61468) [x86_64-darwin14]
false

$ ruby -ve "p Complex(Float::NAN, Float::NAN).finite?"
ruby 2.4.1p111 (2017-03-22 revision 58053) [x86_64-darwin14]
true
```

Ref: https://github.com/ruby/ruby/blob/v2_5_0/complex.c#L248
     https://github.com/ruby/ruby/blob/v2_5_0/numeric.c#L1731

### DIFF
--- a/core/src/main/java/org/jruby/RubyComplex.java
+++ b/core/src/main/java/org/jruby/RubyComplex.java
@@ -995,7 +995,7 @@ public class RubyComplex extends RubyNumeric {
         }
 
         if (magnitude instanceof RubyFloat) {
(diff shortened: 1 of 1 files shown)


yui-knk:
This is duplicated with https://github.com/jruby/jruby/commit/825c75574e7abaca6e6f91b0ef35758491023f80 . So I close this PR.

PR #4940, opened 2018-01-05
Implement KeyError#receiver and KeyError#key

nomadium:
Hi,

This is another feature targeting Ruby 2.5 [1]: KeyError#receiver and KeyError#key (feature #12063).

Note: the tests are copied from MRI.

To implement this I followed the approach used with NameError class, since it has a similar behaviour. If this is not the right approach, just let me know.

### DIFF
--- a/core/src/main/java/org/jruby/RubyKeyError.java
+++ b/core/src/main/java/org/jruby/RubyKeyError.java
@@ -1,0 +1,81 @@
+/***** BEGIN LICENSE BLOCK *****
+ * Version: EPL 2.0/GPL 2.0/LGPL 2.1
+ *
+ * The contents of this file are subject to the Eclipse Public
+ * License Version 2.0 (the "License"); you may not use this file
+ * except in compliance with the License. You may obtain a copy of
+ * the License at http://www.eclipse.org/legal/epl-v10.html
+ *
+ * Software distributed under the License is distributed on an "AS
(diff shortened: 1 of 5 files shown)


core/src/main/java/org/jruby/RubyKeyError.java
```diff
@@ -0,0 +1,81 @@
...
+        this(runtime, exceptionClass, exceptionClass.getName());
+    }
+
+    public RubyKeyError(Ruby runtime, RubyClass exceptionClass, String message) {
+        this(runtime, exceptionClass, message, null, null);
+    }
+
+    public RubyKeyError(Ruby runtime, RubyClass exceptionClass, String message, IRubyObject recv, IRubyObject key) {
+        super(runtime, exceptionClass, message);
+        this.receiver = recv == null ? runtime.getNil() : recv;
```
enebo: Can you remove the null checks and above just pass in nil in the constructor above?  I think we can accept the contract for this internal type can guarantee non-null values.  In Array#fetch (and this is good to know) and all @JRubyMethods no passed in arguments will ever be null.  We eliminate a lot of bulletproofing this way.

nomadium: Good observation, thanks. I believe I addressed your recommendation in the new commits I just pushed. Please let me know if there is something else to improve.

PR #4944, opened 2018-01-06
Set original Module to definedClass of methods

yui-knk:
Set original Module to definedClass of methods when
new methods are defined to PrependedModule.

Before this commit, definedClass of method a
is class c but definedClass of method b
is PrependedModule:

```
M = Module.new
c = Class.new {
  def a; end
  prepend M
  def b; end
}
```

This commit will set definedClass of method b
to class c and fix `TestMethod#test_alias_owner`.

### DIFF
--- a/core/src/main/java/org/jruby/PrependedModule.java
+++ b/core/src/main/java/org/jruby/PrependedModule.java
@@ -54,6 +54,12 @@ public class PrependedModule extends IncludedModule {
     }
 
     @Override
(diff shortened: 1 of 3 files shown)


headius:
I will review for inclusion in 9.1.16.

headius:
Manually merged to 9.1 branch in ce434ecd0c. It will merge to master when @enebo gets to that later this week.

headius:
Thank you for your help!

PR #4945, opened 2018-01-06
Fix warnings

yui-knk:
This commit will fix these warnings:

```
Java HotSpot(TM) 64-Bit Server VM warning: Setting CompressedClassSpaceSize has no effect when compressed class pointers are not used
```


kares:
hmm, not sure this makes sense - do you have the issue locally?
... otherwise (if its only CI) we shall adjust travis-ci accordingly, instead.

yui-knk:
https://github.com/jruby/jruby/commit/e55ba7979438d7583a995ef37b7bd9c074f35478 removed `-XX:-UseCompressedOops`, so now warnings disappeared. I will close this PR :)

PR #4946, opened 2018-01-06
Fix NullPointerException when option is not :base

ChrisBr:
now it throws an ArgumentError like it does in MRI.
Introduced in #4936.


enebo:
@ChrisBr Now that you fixed this I suspect you are missing another Hash type check if arity == 3.  Trying passing in a Fixnum as arg three...

PR #4951, opened 2018-01-08
Fix `Hash#reject`

yui-knk:
In MRI, `Hash#reject`

* creates new hash
* sets keys & values if block returns false

This commit will fix `#test_reject`.

Ref: https://github.com/ruby/ruby/blob/v2_4_0/hash.c#L1328


enebo:
@yui-knk if you realize some of these are also fixes for Ruby 2.3 support I encourage you to submit PRs in those cases against jruby-9.1 branch.  We will merge that branch back to master so both releases get the same fixes.

yui-knk:
I see, thank you for your advice!

PR #4953, opened 2018-01-08
Implement Dir.each_child

nomadium:
Hi folks,

This is another feature targeting Ruby 2.5 [1]: Dir.each_child (feature #11302).

Note: The test is copied from MRI.

### DIFF
--- a/core/src/main/java/org/jruby/RubyDir.java
+++ b/core/src/main/java/org/jruby/RubyDir.java
@@ -392,6 +392,37 @@ public class RubyDir extends RubyObject {
     }
 
     /**
+     * Executes the block once for each entry in the directory except
+     * for "." and "..", passing the filename of each entry as parameter
+     * to the block.
+     */
+    @JRubyMethod(name = "each_child", meta = true)
+    public static IRubyObject each_child(ThreadContext context, IRubyObject recv, IRubyObject arg, Block block) {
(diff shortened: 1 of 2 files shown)


core/src/main/java/org/jruby/RubyDir.java
```diff
@@ -420,6 +451,22 @@ public static IRubyObject foreach19(ThreadContext context, IRubyObject recv, IRu
         return foreachCommon(context, recv, context.runtime, pathString, encoding, block);
     }
 
+    private static IRubyObject eachChildCommon(ThreadContext context, IRubyObject recv, Ruby runtime, RubyString _path, RubyEncoding encoding, Block block) {
+        if (block.isGiven()) {
+            RubyClass dirClass = runtime.getDir();
+            RubyDir dir = (RubyDir) dirClass.newInstance(context, new IRubyObject[]{_path}, block);
+
+            dir.each_child(context, block);
```
enebo: Does this path (block passed in) in MRI really ignore encoding param?

nomadium: Good observation. I'm not really sure about MRI on this, but ``Dir::foreach`` JRuby method implementation is currently doing something similar if I'm not mistaken: https://github.com/jruby/jruby/blob/ruby-2.5/core/src/main/java/org/jruby/RubyDir.java#L481

I'll get back to this later with my findings.

PR #4954, opened 2018-01-08
Alias Set#=== to #include?

ChrisBr:
tests cherry picked from https://github.com/ruby/ruby/commit/2aee703e7a29bcfc2095e40870f5a3c45cd96dd2.
Feature https://bugs.ruby-lang.org/issues/13801.


kares:
:+1: najs job, one minor detail related to adding into *spec/ruby/...*` they're merged/picked back-and-forth to the canonical **ruby/spec** repo

try to introduce changes in there in separate commits so they can be picked easily ...

ChrisBr:
@kares splitted into three different commits now!

PR #4955, opened 2018-01-08
Revert to using weakref.rb, since it's based on a proper weak map.

headius:
Opening as PR since this is a nontrivial change to put into 9.1, even though it aligns us better with MRI. Primary reason for this change is to get weakref specs to be more reliable...perhaps bringing our impl closer to theirs will do this.

This library used to use _id2ref (an internal API) for its
implementation. This was not only very VM-specific (most VMs
can't reconstitute an object reference given only its ID) but also
buggy (an evacuated reference might get overlaid by a new object
due to MRI's conservative GC and use of address for IDs.

The new implementation of the library uses ObjectSpace::WeakMap,
which is considered an internal API but which exposes an opaque
weak identity map that can be implemented on other VMs more
more easily. Given this and our need to continue descending from
Delegate, moving back to the Ruby version may make sense.


headius:
@enebo @kares In the interest of getting CI green I'm going forward with this on 9.1 branch. Let me know if you have any concerns.

enebo:
for posterity I liked the notion of leaning on this library to only have one thing we need to maintain involving weak references :)

PR #4957, opened 2018-01-08
Add Set#reset

ChrisBr:
https://bugs.ruby-lang.org/issues/6589
Tests c&p from https://github.com/ruby/ruby/commit/8c90432af72a59d8934b2c94a1bc847449e1f393.
Supports #4876

### DIFF
--- a/core/src/main/java/org/jruby/ext/set/RubySet.java
+++ b/core/src/main/java/org/jruby/ext/set/RubySet.java
@@ -875,6 +875,12 @@ public class RubySet extends RubyObject implements Set {
         return context.runtime.getFalse();
     }
 
+    @JRubyMethod(name = "reset")
+    public IRubyObject reset(ThreadContext context) {
+        this.hash.rehash();
+        return this;
+    }
+
     @JRubyMethod(name = "eql?")
     public IRubyObject op_eql(ThreadContext context, IRubyObject other) {
         if ( other instanceof RubySet ) {
--- a/test/mri/test_set.rb
+++ b/test/mri/test_set.rb
@@ -734,6 +734,19 @@ class TC_Set < Test::Unit::TestCase
     assert_equal(3, set.size)
     assert_equal(array.uniq.sort, set.sort)
   end
+
+  def test_reset
+   [Set, Class.new(Set)].each { |klass|
+     a = [1, 2]
+     b = [1]
+     set = klass.new([a, b])
+
+     b << 2
+     set.reset
+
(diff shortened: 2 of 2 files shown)


ChrisBr:
Please note that the test will fail because of this issue
https://github.com/jruby/jruby/issues/4958

core/src/main/java/org/jruby/ext/set/RubySet.java
```diff
@@ -875,6 +876,18 @@ public IRubyObject op_equal(ThreadContext context, IRubyObject other) {
         return context.runtime.getFalse();
     }
 
+    @JRubyMethod(name = "reset")
+    public IRubyObject reset(ThreadContext context) {
+        IRubyObject tmp = TypeConverter.checkHashType(context.runtime, this.hash);
+        if (tmp.isNil()) {
+            modifyCheck(context.runtime);
+        } else {
+            System.out.println("rehash");
```
kares: :scissors:

core/src/main/java/org/jruby/ext/set/RubySet.java
```diff
@@ -875,6 +876,18 @@ public IRubyObject op_equal(ThreadContext context, IRubyObject other) {
         return context.runtime.getFalse();
     }
 
+    @JRubyMethod(name = "reset")
+    public IRubyObject reset(ThreadContext context) {
+        IRubyObject tmp = TypeConverter.checkHashType(context.runtime, this.hash);
```
kares: `this.hash` is 'internal' -> can not be changed (wout hacks) on .rb end ... + its already of type `RubyHash`

core/src/main/java/org/jruby/ext/set/RubySet.java
ChrisBr: 😢

core/src/main/java/org/jruby/ext/set/RubySet.java
ChrisBr: Not sure if I understand correct :/ What should I use instead ``this.hash``.
The type check is adapted from the MRI implementation.

kares: > The type check is adapted from the MRI implementation.

figured, since this as a native part, specific to JRuby, just logically simplify what makes sense.
since you have a `RubyHash hash` you do not need any type-checking whether its a `Hash` ...

kares:
Thanks Chris, well done!

PR #4961, opened 2018-01-09
Fix Hash#rehash method for duplicate keys

### LINKED ISSUE #4958: Hash#rehash does not properly rehash
### Environment
* jruby 9.3.0.0-SNAPSHOT (2.5.0) 2018-01-08 4b5cea6 OpenJDK 64-Bit Server VM 25.151-b12 on 1.8.0_151-b12 +jit [linux-x86_64]

### Expected Behavior
When rehashing a hash with equal objects in it, they don't get removed.
```
a = [1,2]
b = [1]
h = {}
h[a] = true
h[b] = true
b << 2
puts h.rehash # { [1,2]: true, [1,2]: true } 
```

I already looked into the issue and it seems that the problem is that Hash#rehash does not handle collisions properly. When two objects have the same hash / bucket number, they get inserted even if they are the same (see example).

### Actual Behavior
Hash#rehash should remove the duplicate entries.

ChrisBr:
Duplicate keys where just ignored before and inserted again.
Fix #4958.

### DIFF
--- a/core/src/main/java/org/jruby/RubyHash.java
+++ b/core/src/main/java/org/jruby/RubyHash.java
@@ -933,8 +933,17 @@ public class RubyHash extends RubyObject implements Map {
                 RubyHashEntry next = entry.next;
                 entry.hash = hashValue(entry.key); // update the hash value
                 int i = bucketIndex(entry.hash, newTable.length);
-                entry.next = newTable[i];
-                newTable[i] = entry;
+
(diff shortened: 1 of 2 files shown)


kares:
@ChrisBr how's the separate commits with changes under **spec/ruby**, do you need a :poodle:  bot around :) ?

ChrisBr:
@kares splitted

headius:
👍

Issue #4943, opened 2018-01-05
Open3.open2e returns `Errno::ENOENT` in jRuby but runs fine on Ruby


abelsromero:
### Environment
JRuby v 9.1.15.0
Run from IDE, no additional configurations.
Windows 7
Java 1.8.0_152

### Expected Behavior

* A script that runs a external command ([kindlegen](https://www.amazon.com/gp/feature.html?docId=1000765211)) using Open3.open2e should work on normal Ruby and jRuby.

### Actual Behavior

* The script works fine on Ruby 2.3.3, but fails when run through jRuby 9.1.15.0 showing `(Errno::ENOENT)`. I assume the error is related to not being able to find some find. is that so?
* Here is is a example project I created to reproduce the issue: https://github.com/abelsromero/jruby-open3-issue.
0. Clone the project and create a KINDLEGEN environment var pointing to the kindlegen file in the `files` directory.
1. Open cmd with Ruby available in PATH and go to the "files" directory.
2. Run "ruby Open3Ruby.rb" --> Script should complete, generating a file called "spine.mobi" in the same directory. If you see the script code, several "puts" show how the actual files involved exist and are available.
3. Open the project in IntelliJ. Run the class "RubyRunner". --> Execution will fail with message
```
C:\home\asalgadr\temp\jruby-open3-issue\.\files\kindlegen.exe
true
C:/home/asalgadr/temp/jruby-open3-issue/files/spine-kf8.epub
true
Exception in thread "main" org.jruby.exceptions.RaiseException: (Errno::ENOENT) C:\\home\\\asalgadr\\temp\\jruby-open3-issue\\.\\files\\kindlegen.exe -dont_append_source -c0 -o spine.mobi C:/home/asalgadr/temp/jruby-open3-issue/files/spine-kf8.epub
	at org.jruby.RubyProcess.spawn(org/jruby/RubyProcess.java:1569)
	at org.jruby.RubyKernel.spawn(org/jruby/RubyKernel.java:1511)
	at RUBY.popen_run(uri:classloader:/META-INF/jruby.home/lib/ruby/stdlib/jruby/open3_windows.rb:204)
	at RUBY.popen2e(uri:classloader:/META-INF/jruby.home/lib/ruby/stdlib/jruby/open3_windows.rb:199)
	at RUBY.<main>(<script>:26)
```
The first four lines are the puts that check that files involved are available.

abelsromero:
I updated the test project to support MacOS.
[...]

PR #4964, opened 2018-01-11
Improve Open3 on Windows

headius:
The Open3 library needs some love to be more functional on Windows.

We have never gotten back to implementing `spawn` properly on Windows, which is heavily used by Open3 for all its popen forms. To get around this, Windows will fall back on an older version of popen3 that only uses JVM processes. Unfortunately, that's the only method we patch, so [others like open2e](https://github.com/jruby/jruby/issues/4943) still try to use the broken spawn logic.

It should not be difficult to patch the other popen forms, since they're mostly just subsets of popen3.


headius:
We may want to incorporate someone else's library into JRuby for Windows support, like https://github.com/enkessler/childprocess or https://github.com/djberg96/win32-process. Both use FFI to do processes "the right way" on Windows, and that would save us reimplementing things.

headius:
My mistake...childprocess just wraps ProcessBuilder as best it can, and win32-process only provides additional methods related to Windows.

JasonLunn:
Since this was added to the list of 9.2 Projects, is this still a blocker for the release of 9.1.17.0?

enebo:
I moved this to .18 but I am unsure if it is resolved for .17?

headius:
This has largely been superseded by #6649 which greatly improved the functionality of open3 on Windows, combined with other fixes for subprocess launching on non-native-process environments (like Windows).

PR #4975, opened 2018-01-14
Feature/string casecmp

ChrisBr:
to return nil for incompatible types instead of raising exception.
Support for Ruby 2.5

See feature https://bugs.ruby-lang.org/issues/13312.


lopex:
@ChrisBr if checkStringType doesnt return nil you can safely cast it to RubyString, otherwise you're doing the conversion twice here by calling convertToString.

ChrisBr:
@lopex you're right, fixed

ChrisBr:
Please note that the casecmp? test is excluded from the test suite because of missing case-fold logic

PR #4980, opened 2018-01-16
Remove not used methods

yui-knk:


kares:
`RubyInstanceConfig` is sensitive API, it might need to stay and go `@Deprecated` instead, right @enebo ?

enebo:
@kares @yui-knk yeah this one I don't think we should remove or annotate as a deprecation.  In this case, an extension author might want to know if ruby is from a command line.  We don't want them to be calling __FILE__ for this.  These decisions are a little difficult YAGNI OTOH it is reasonable information in a consistent location someone might actually use?

The larger issue is we do not have an explicit native extension API (beyond ScriptingContainer which arguably is more of an embedding API) so I could see extension authors expecting this class to remain unchanged.

yui-knk:
@enebo 
Sorry, I didn't explain it clearly enough.
`RubyInstanceConfig` has `public boolean hasInlineScript()` and `public boolean isInlineScript()`. Both of them return `hasInlineScript `, so I think we can omit (or deprecate) one of them.

enebo:
@yui-knk oh ok yeah I agree then one could be deprecated.  My logic would be to keep hasInlineScript() since it implies it was invoked as inline script but it may loads lots of other Ruby.  Anyone else have an opinion?

PR #4981, opened 2018-01-16
Do not leak DNS Request IDs

fxposter:
While https://github.com/jruby/jruby/commit/d1a760e066343e7a4956bbdd7be7975b1eda04cd fixed handling of compressed IPv6 addresses,
it also broke the "freeing" part of it.

Currently every DNS request leaks single request id:
```
require 'resolv'

Resolv::DNS::RequestID.values.map(&:length)
# => []

Resolv::DNS.new.getaddress('example.com')

Resolv::DNS::RequestID.values.map(&:length)
# => [1]
```

Given the fact that request ids are chosen from `0x0000..0xffff` range - the app can issue 65535 requests and then will be blocked forever trying to allocate another one.

This commit makes request id caching work by using same data for allocation and freeing.

### DIFF
--- a/lib/ruby/stdlib/resolv.rb
+++ b/lib/ruby/stdlib/resolv.rb
@@ -772,7 +772,7 @@ class Resolv
           sock = @socks_hash[host.index(':') ? "::" : "0.0.0.0"]
           return nil if !sock
           service = [IPAddr.new(host), port]
-          id = DNS.allocate_request_id(host, port)
+          id = DNS.allocate_request_id(service[0], service[1])
           request = msg.encode
           request[0,2] = [id].pack('n')
           return @senders[[service, id]] =


fxposter:
We can cover the problematic behavior in spect, but it requires either having a working DNS server to run the tests (ie: move the example above into a spec) or adding a spec for internal classes/methods (not sure if it would be useful, though)

enebo:
@fxposter have you tried this against MRI?  This code is theirs.   If it affects them then we should get this moved upstream.

fxposter:
This is not strictly true. The commit mentioned above diverges from the MRI behavior which don't have any `IPAddr.new` calls in this place: https://github.com/ruby/ruby/blob/trunk/lib/resolv.rb#L768.

enebo:
@fxposter oh sorry I should have checked.  See: https://github.com/jruby/ruby/commit/44fe41bb37b13d8d9e93d3a3484b72a06f254a6a#diff-5eaee83b511bb5c304d686f8f47dc6c2L773

That commit shows we maintain doing that in both sender and recv_reply.  Note this is 2.3.x support but I should have realized you probably wanted this PR against jruby-9.1 branch and not master since master is for 2.4.x support atm (well no doubt you will want it in all the places but we have not put out jruby 9.2 yet).

Could I ask two favors?
 1. Submit this against jruby-9.1 branch (the source of resolv.rb has changed a tiny bit) and make sure things are cool still.  I will merge this onto master so 2.4 support works as well as update our stdlib patching which happens outside of jruby itself.
 2. See if you can observe a leak on recv_reply since we do the same thing there.

The great thing about this work is fixing up part of our MRI stdlib changeset.  One less difference is one less headache at some point in the future :)

kares:
this has been stale for a while but seems that this makes sense to go into 9.2 ... marking for now

PR #4992, opened 2018-01-20
Check an encoding of string

yui-knk:
MRI checks an encoding of string is broken
before conveting string to symbol.
This commit will fix `#test_symbol_encoding`.

Ref: https://github.com/ruby/ruby/blob/v2_5_0/symbol.c#L432


enebo:
A recent java 9 change has made seeing results of this difficult but it looks reasonable.

PR #4993, opened 2018-01-20
Ruby 2.5: Add String#delete_prefix and String#delete_prefix!

ChrisBr:
See feature https://bugs.ruby-lang.org/issues/12694

https://github.com/jruby/jruby/issues/4876

### DIFF
--- a/core/src/main/java/org/jruby/RubyString.java
+++ b/core/src/main/java/org/jruby/RubyString.java
@@ -4024,6 +4024,32 @@ public class RubyString extends RubyObject implements EncodingCapable, MarshalEn
       }
     }
 
+    @JRubyMethod(name = "delete_prefix")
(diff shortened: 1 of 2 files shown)


ChrisBr:
Please note that the last test in test_delete_prefix! currently fails:

```
    s = S("ax")
    o = Struct.new(:s).new(s)
    def o.to_str
      s.freeze
      "a"
    end
    assert_raise_with_message(RuntimeError, /frozen/) {s.delete_prefix!(o)}
```

I'm wondering what I do wrong here as I do a frozen check in delete_prefix!
cc @enebo @headius

ChrisBr:
@enebo @headius fixed the failing test. The problem was that the modify_check() was called before the freeze call was executed. I moved the modifyCheck() now one line down which fixed it.

PR #4994, opened 2018-01-21
Fix `Symbol#inspect` of UTF_16/UTF_32

yui-knk:
Stop to append byte before inspect string.
For example, when an encoding of `symbolBytes` is UTF_16LE, "a" is
`[0x61, 0x00]`. If we append `":"` (0x3A) to `symbolBytes` before
inspect it, bytes are `[0x3A, 0x61, 0x00]` with UTF_16LE encoding.
This is not what we want to get. This commit chnages the order of
inspecting and appending to avoid this.

Ref: https://github.com/ruby/ruby/blob/v2_5_0/string.c#L10402

### DIFF
--- a/core/src/main/java/org/jruby/RubySymbol.java
+++ b/core/src/main/java/org/jruby/RubySymbol.java
@@ -256,27 +256,22 @@ public class RubySymbol extends RubyObject implements MarshalEncoding, EncodingC
     }
 
     final RubyString inspect(final Ruby runtime) {
-        ByteList result = new ByteList(symbolBytes.getRealSize() + 1);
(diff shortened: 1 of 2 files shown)


enebo:
@yui-knk You seem to even fix a second bug in here where we seem to be adding :" at the front of a symbol but add no closing ".  I might change this code now that you have fixed this because we potentially make 3 instances of RubyString depending on the symbol being inspected.  I think we can reduce this to just one.

yui-knk:
> ability to determine CR_7BIT is much simpler if we make a string first

I agree :)

PR #4997, opened 2018-01-22
Update openjdk on travis to 8

yui-knk:
23783ceb3bf started to use `java.util.function`, but this
package was introduced from 1.8. Update jdk to keep build passed.


yui-knk:
I see. Thank you for your quick review :)

PR #4999, opened 2018-01-22
IO#write accepts multiple arguments

nomadium:
Hi folks,

This is another feature targeting Ruby 2.5 [1] on IO#write method (feature #9329).

Note: The tests are copied from MRI.


enebo:
@nomadium fun...first reasonably lengthed use of literate chain of closures!

PR #5001, opened 2018-01-23
Implement `Rational#coerce` with `Complex` object

yui-knk:
In MRI

* if the argument is complex with zero imaginary part,
  converts it to ratinal
* if the argument is complex with non-zero imaginary part,
  converts self to complex

Ref: https://github.com/ruby/ruby/blob/v2_3_0/rational.c#L1171

### DIFF
--- a/core/src/main/java/org/jruby/RubyRational.java
+++ b/core/src/main/java/org/jruby/RubyRational.java
@@ -712,6 +712,13 @@ public class RubyRational extends RubyNumeric {
             return runtime.newArray(other, f_to_f(context, this));
         } else if (other instanceof RubyRational) {
             return runtime.newArray(other, this);
+        } else if (other instanceof RubyComplex) {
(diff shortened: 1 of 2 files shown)


kares:
:1st_place_medal: :+1:  ... but maybe its better to target master (9.2) with these - not sure? (unless of course there's bugs)
have been doing cleanup/(internal) refactoring around Numerics lately and some pieces might not merge :green_book: 

on the other hand I am happy to handle these merges we'll need to sync up about the plan for 9.1 as it didn't merge over in a while and it won't go clean with latest de-dup commits on both sides // @headius @enebo

enebo:
@kares I think this is fine.  He is reference 2.3.0 behavior in the MRI link and it removes an issue in MRI test excludes.

enebo:
@kares merging is a problem but I suspect this one will not be very difficult to know where it fits?

PR #5006, opened 2018-01-24
Fix test coerce2

yui-knk:


enebo:
@yui-knk just waiting for test:mri to finish and this looks good.  Weirded out that '.div' is no longer there but I see it is not in Ruby docs for 2.2?  I don't know the history of that merthod though.

yui-knk:
Maybe https://github.com/ruby/ruby/commit/d82ed7e2c6e0f873ccacafb28c36a293383c3c15 this commit removed `Rational#div` and this commit was included Ruby 1.9.2, so I think Ruby docs for 2.2 should not contain that method.
I can not read why this method was removed by the commit message....

enebo:
@yui-knk yeah it is strange to see Ruby API shrink :)

PR #5010, opened 2018-01-25
Fix `convertCommon`

yui-knk:
1. Remove f_to_r call
  MRI does not call `to_r` when `a1` is not a float nor a string.
  Ref: https://github.com/ruby/ruby/blob/v2_3_0/rational.c#L2423-L2428

2. Call `#to_r` when `a1` is an integer
  Ref: https://github.com/ruby/ruby/blob/v2_3_0/rational.c#L2445-L2446

### DIFF
--- a/core/src/main/java/org/jruby/RubyNumeric.java
+++ b/core/src/main/java/org/jruby/RubyNumeric.java
@@ -1184,7 +1184,7 @@ public class RubyNumeric extends RubyObject {
         return sites(context).op_equals.call(context, other, other, this);
     }
 
-    /** num_numerator
+    /** numeric_numerator
      *
      */
     @JRubyMethod(name = "numerator")
@@ -1193,7 +1193,7 @@ public class RubyNumeric extends RubyObject {
         return sites(context).numerator.call(context, rational, rational);
     }
 
-    /** num_denominator
+    /** numeric_denominator
      *
      */
     @JRubyMethod(name = "denominator")
--- a/core/src/main/java/org/jruby/RubyNumeric.java
+++ b/core/src/main/java/org/jruby/RubyNumeric.java
@@ -61,6 +61,7 @@ import static org.jruby.util.Numeric.f_abs;
 import static org.jruby.util.Numeric.f_arg;
 import static org.jruby.util.Numeric.f_mul;
 import static org.jruby.util.Numeric.f_negative_p;
+import static org.jruby.util.Numeric.f_to_r;
 
 /**
  * Base class for all numerical types in ruby.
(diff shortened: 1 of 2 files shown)


enebo:
@yui-knk you have a few regressions in spec:ruby:fast involving this.  They are sometimes a bit pedantic on Ruby dispatch but at least one error looks like it may be recursing a lot.

yui-knk:
@enebo I fixed regressions. The rest of failures are encoding related test cases, which also fail on the head of jruby-9.1:

* TestTranscode#test_windows_1255 
* Integer#chr without argument when Encoding.default_internal is not nil and self is greater than 255 raises RangeError if self is invalid as a codepoint in the default internal encoding
* Integer#chr with an encoding argument raises RangeError if self is invalid as a codepoint in the specified encoding FAILED

enebo:
@yui-knk jcodings was updated yesterday and is the fallout.  New updated data for encoding support.  The regressions need to be examined still.

kares:
not sure about these changes - I think we might have clashed on blindly following MRI
that `f_to_r` with the "Fix spec:ruby:fast" msg doesn not tell much and is not part of the PR's desc

\+ there's a Rational regression (test failing) on master, @yui-knk not sure but might be relate to these?
or maybe do you recall if that `f_to_r` is necessary (a legit fix)? 
(was actually doing the opposite on master - trying to avoid the dynamic call-site when not needed - for a rational instance having to need to go through `to_r` just seems like smt I would try to avoid :smiling_imp: )

enebo:
@kares the line below basically explains that change: "MRI does not call to_r when a1 is not a float nor a string."  @yui-knk seems to be following MRI logic exactly and this fixed an MRI test (unsurprisingly).  

As for master, I have not checked but I half wonder if perhaps 2.4 changed behavior (and we should not have merged this fix to master)?  His PR actually removes call to_r in that particular case for 2.3.x, but original  f_to_r code always performs the dyncall so maybe 2.4.x+ requires always dyncalling now?

kares:
sorry, the first commit is nice and clear but the [following isn't](https://github.com/jruby/jruby/pull/5010/commits/436527f39fb23b755b5e004898266a32d2bb6bd6) ... that is what I was hoping to understand why
anyway I will check it later whether its related to the failing test on master

yui-knk:
The second commit is needed to fix

* "Numeric#denominator works with Numeric subclasses" in spec/ruby/core/numeric/denominator_spec.rb
* "Numeric#numerator works with Numeric subclasses" in spec/ruby/core/numeric/numerator_spec.rb

These specs expect "#to_r" is called when `#denominator`/`#numerator` is called for Numeric subclass.

By the way https://github.com/jruby/jruby/pull/5022 will fix failing test on master.

PR #5011, opened 2018-01-25
Remove TestArray#test_zip exclusion test case

nomadium:
Hi folks,

I noticed the assertion in the excluded test ```test_zip``` at https://github.com/jruby/jruby/blob/master/test/mri/ruby/test_array.rb#L2223 was raising a ```NoMethodError``` instead of the expected ```TypeError```.

With this proposed change, if correct, the test should pass.


olleolleolle:
@nomadium These changes you make are great! 

[ruby/ruby cross-reference link to test_array.rb: `test_zip` method](https://github.com/ruby/ruby/blob/978be08b9db29cf190ef483320ed8c6d3df00091/test/ruby/test_array.rb#L2330)

[Where the TypeError for enumerable is raised](https://github.com/ruby/ruby/blob/trunk/enum.c#L2687)

enebo:
@olleolleolle thanks for finding that respond_to? reference.  I only paused in merging this one before because I thought, "wow..a respond_to? per element".

PR #5013, opened 2018-01-26
fix issue where expr fails on freebsd.

arjanvandervelde:
These scripts produce errors when run on FreeBSD (in the context of logstash). Adding the double dash prevents interpretation of the left operand as an option to expr itself. See also https://discuss.elastic.co/t/cannot-install-multiline-filter-on-freebsd-11/106929?source_topic_id=117201.

In particular, the culprit in the logstash issue was line https://github.com/jruby/jruby/blob/4cdf882cf1a17f71620886250e3519d4ab1ef91b/bin/jruby.bash#L247 expanding to "expr -C : .[.*]".

Thanks.

### DIFF
--- a/bin/jruby.bash
+++ b/bin/jruby.bash
@@ -244,7 +244,7 @@ do
      # Match -Xa.b.c=d to translate to -Da.b.c=d as a java option
      -X*)
         val=${1:2}
-        if expr "$val" : '.*[.]' > /dev/null; then
+        if expr -- "$val" : '.*[.]' > /dev/null; then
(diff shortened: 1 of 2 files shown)


arjanvandervelde:
too bad that timed out

enebo:
@arjanvandervelde our 9.1.x is on the branch jruby-9.1 but that is far from obvious so I will just cherry-pick this over there.

headius:
This fix appears to cause issues on Alpine Linux due to some incompatibility with the `expr` command. See #5091.

PR #5014, opened 2018-01-26
Remove TestArray#test_uniq_bang_with_freeze exclusion test case

nomadium:
Hi folks,

I noticed one assertion in the excluded test ```TestArray#test_uniq_bang_with_freeze``` was failing because ```uniq_bang19``` method in ```RubyArray``` class was not checking if the RubyArray instance can be modified after evaluating the passed block.

With this proposed change, if correct, the test should pass.

MRI does something similar.

They do a check at the beginning of the method [1] and other after checking if a block was given [2].

1. https://github.com/ruby/ruby/blob/trunk/array.c#L4447
2. https://github.com/ruby/ruby/blob/trunk/array.c#L4459


enebo:
@nomadium This test is also on 9.1.x so I am cherry-picking this back.  When possible when you notice these check jruby-9.1 branch first so it is a little easier for us (we regularly merge jruby-9.1 against master).

nomadium:
@enebo Cool, will do on the next ocassion.

PR #5016, opened 2018-01-27
Exclude `#test_windows_1255` from our tests

yui-knk:
We use jcodings 1.0.27 which maps
windows-1255 0xCA to UTF-8 U+05BA.
But this had been supported since CRuby 2.4.

Ref: https://github.com/ruby/ruby/commit/b24e093296c213318f2b5efa070d8294205d4ea3


lopex:
It must be this update then: https://github.com/jruby/jcodings/commit/5f77b82f7e933070a400708ab1a9749d268bda13.
@headius: https://github.com/ruby/ruby/blob/v2_4_3/test/ruby/test_transcode.rb#L366 doesnt seem to be expecting any errors here.

enebo:
@lopex @headius what is the appropriate thing to do here?  Looks like they fixed an issue which presumably means this test was incorrect for 2.3.x?

PR #5021, opened 2018-01-29
Do modifyCheck for each pair in Array#sort!

ChrisBr:
as calling comparison could freeze the Array and should trigger a RuntimeError.


olleolleolle:
Cross-referencing links:


The test was:

```ruby
  def test_sort_bang_with_freeze
    ary = []
    o1 = Object.new
    o1.singleton_class.class_eval {
      define_method(:<=>) {|v|
        ary.freeze
        1
      }
    }
    o2 = o1.clone
    ary << o1 << o2
    orig = ary.dup
    assert_raise(FrozenError, "frozen during comparison") {ary.sort!}
    assert_equal(orig, ary, "must not be modified once frozen")
  end
```

https://github.com/ruby/ruby/blob/978be08b9db29cf190ef483320ed8c6d3df00091/test/ruby/test_array.rb#L1521-L1535

This Git blame view shows the 2 months ago, there was a change so that the `FrozenError` now had become a `RuntimeError` subclass.
https://github.com/ruby/ruby/blame/978be08b9db29cf190ef483320ed8c6d3df00091/test/ruby/test_array.rb#L1521-L1535

ChrisBr:
@olleolleolle @kares should I change it in this PR? Because this one is targeting master but I guess the change is for 2.5, right?

olleolleolle:
@ChrisBr My additional comment about the evolution of Ruby was only informative.

My take: Keep them separate, take no further action for this PR.

kares:
fixed an existing spec, seems no new ones broke. so all should be good as is - @enebo okay to merge?

PR #5027, opened 2018-02-02
Andre Kullmann make the profiling service configurable via java prope…

areman:
Currently the profiling service can only configured as program argument. With this pull request it will be possible to configure it as java property, too. This feature is importent to start the profiling if you don't run the jruby executable. e.g. you use jruby in a war archive or as script engine in a standard java application.


kares:
good, but those new TODOs of yours aren't needed. the runtime is essential, setter makes no sense - since its expected to stay the same...

areman:
Dear @kares, 
yes the setter isn't needed. In my opinion it's not good design pattern to publish an interface which describes the methods to implement but implicit a special constructor is required. I'm not sure if the runtime performance will be  a issue. Currently the constructor is called over reflection.

kares:
@areman thanks for the clarification and for the design pattern hints. 
not sure if you'd like to get this PR merged or start a discussion with those TODOs. 
if you'd like this in, please remove all of the redundant noise, otherwise should address those TODOs.

areman:
Dear @kares,
yes I'll would like to see the request merged. 
I've removed the TODO hints from the "productive" code.

PR #5037, opened 2018-02-08
align C ported memsearch code - need to handle array[length] properly

kares:
... otherwise code doing utf8 str scan might fail with array-out-of-index

resolves GH-2036

someone with C knowledge please review, seems to me that MRI handles the case without special care : 
at `rb_memsearch_qs_utf8_hash(byte[] xBytes, final int x)` while `x == xBytes.length`

### DIFF
--- a/core/src/main/java/org/jruby/util/StringSupport.java
+++ b/core/src/main/java/org/jruby/util/StringSupport.java
@@ -2297,9 +2297,15 @@ public final class StringSupport {
         return -1;
     }
 
-    private static int rb_memsearch_qs_utf8_hash(byte[] xBytes, int x) {
-        int mix = 8353;
-        int h = xBytes[x] & 0xFF;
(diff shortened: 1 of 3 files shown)


core/src/main/java/org/jruby/util/StringSupport.java
```diff
@@ -2297,9 +2297,15 @@ private static int rb_memsearch_qs(byte[] xsBytes, int xs, int m, byte[] ysBytes
...
+        final int mix = 8353;
+        int h;
+        if (x != xBytes.length) {
+            h = xBytes[x] & 0xFF;
+        }
+        else {
+            h = '\0'; // (C) ary end - due y+m at rb_memsearch_qs_utf8
+        }
         if (h < 0xC0) {
             return h + 256;
```
headius: This will always return 256 at string end...is that correct?

kares: yep - should return the very same as the (original) C code when reaching \0 boundary

PR #5040, opened 2018-02-13
Update lib/ruby/stdlib/csv.rb to 2.5

esparta:
Semi-automatic update trying to preserve the history on [MRI ](https://github.com/ruby/ruby).

I checked the spec and theoretically should not be affected nor need to update with upstream.

### DIFF
--- a/lib/ruby/stdlib/csv.rb
+++ b/lib/ruby/stdlib/csv.rb
@@ -1291,7 +1291,7 @@ class CSV
       begin
         yield csv
       ensure
-        csv.close
+        csv.closed? || csv.close
       end
     else
       csv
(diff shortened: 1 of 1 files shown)


kares:
believe this has been done at: https://github.com/jruby/jruby/commit/ef5e53afb383f0497c0d6052a3e29d9ebf7224b2#diff-12ef7a6694696218e75a02514e323f85

ask around at IRC if you're into a similar effort - the stdlib git history isn't relevant that much

headius:
@esparta I am interested in better ways to manage our copy of stdlib.

As it stands right now, we pull from a forked copy of CRuby at https://github.com/jruby/ruby. The patches we apply on top of stdlib are rolled forward each time we update, but we baseline from the actual CRuby repo every time.

A submodule might be an option, or a synchronized subtree like we do for spec/ruby.

There's a bit of massaging to get everything from CRuby into place for JRuby (e.g. we don't have an ext dir so those need to get condensed into lib/ruby/stdlib).

We may not actually want MRI's commit messages in our history either, and having the build basically just pull a known version of CRuby, apply a patch, and copy things into place seems like the cleanest option to me right now.

Issue #5044, opened 2018-02-15
lazy map adds extra array layer to single-element array output


Taywee:
### Environment

jruby 9.1.15.0 (2.3.3) 2017-12-07 929fde8 OpenJDK 64-Bit Server VM 25.151-b12 on 1.8.0_151-b12 +jit [linux-x86_64]

Linux nyarlathotep 4.14.11-gentoo-r2 #1 SMP Thu Jan 4 17:46:16 MST 2018 x86_64 Intel(R) Core(TM) i5-3570K CPU @ 3.40GHz GenuineIntel GNU/Linux

### Expected Behavior

The expected behavior should be as shown in standard Ruby:

```ruby
[15] pry(main)> [1, 2, 3].map{|a| [a + 1]}.to_a
=> [[2], [3], [4]]
[16] pry(main)> [1, 2, 3].lazy.map{|a| [a + 1]}.to_a
=> [[2], [3], [4]]
[17] pry(main)> [[1], [2], [3]].lazy.map(&:itself).to_a
=> [[1], [2], [3]]

```

### Actual Behavior

Run from vanilla jruby's IRB, it looks like this:

```ruby
irb(main):005:0> [1, 2, 3].map{|a| [a + 1]}.to_a
=> [[2], [3], [4]]
irb(main):006:0> [1, 2, 3].lazy.map{|a| [a + 1]}.to_a
=> [[[2]], [[3]], [[4]]]
irb(main):007:0> [[1], [2], [3]].lazy.map(&:itself).to_a
=> [[[1]], [[2]], [[3]]]
```

Taywee:
I found another bug that might be related.  After a take, map erroneously strips arrays, and has some weird behaviour, including an incorrect ArgumentError:

```ruby
[66] pry(main)> [[1],[2],[3]].lazy.take(2).map { |elem| p elem }.force
1
2
=> [1, 2]
[67] pry(main)> [[],[2],[3]].lazy.take(2).map { |elem| p elem }.force
nil
2
=> [nil, 2]
[68] pry(main)> [[1, 2],[2],[3]].lazy.take(2).map { |elem| p elem }.force
1
2
=> [1, 2]
[69] pry(main)> [[1, 3],[2],[3]].lazy.take(2).map { |elem| p elem }.force
1
2
=> [1, 2]
[70] pry(main)> [[1, 3],[2],[3]].lazy.take(2).map { |elem| elem.itself }.force
=> [1, 2]
[...]

PR #5045, opened 2018-02-15
Revert "Fixed yielded arrays of 1 element bug when using Enumerator::…

glucero:
This fixes: https://github.com/jruby/jruby/issues/5044

This reverts commit cb25efe8052eea9d919cdb20d13a9678614f5307.

The code reverted in this pull request comes from https://github.com/jruby/jruby/issues/2376 and the related issue https://github.com/jruby/jruby/pull/2458

The test case in 2376 still works with the code removed. 

```bash
jruby:master > bin/jruby -e "puts Enumerator.new{|y| y << [42]}.first.inspect"
[42]

```

### DIFF
--- a/core/src/main/java/org/jruby/RubyYielder.java
+++ b/core/src/main/java/org/jruby/RubyYielder.java
@@ -104,10 +104,6 @@ public class RubyYielder extends RubyObject {
 
     @JRubyMethod(name = "<<", rest = true)
     public IRubyObject op_lshift(ThreadContext context, IRubyObject[]args) {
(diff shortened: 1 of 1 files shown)


enebo:
@glucero how about two test cases in #3814?

glucero:
Please disregard - I mistakenly had the previous version built on my system when I tested the older test case. This revert causes a regression.

Closing.

enebo:
@glucero we appreciate you looking into this.  Block array dereferencing logic is complicated...to say the least.

PR #5050, opened 2018-02-16
refactor annotation-binder generated populators

kares:
accounts for a slight speed-up - the populator won't execute much 'redundant' code

this isn't the main issue with boot (compared to invokers) but still deserves a hand-craft

### DIFF
--- a/core/src/main/java/org/jruby/Ruby.java
+++ b/core/src/main/java/org/jruby/Ruby.java
@@ -3027,6 +3027,15 @@ public final class Ruby implements Constantizable {
         javaToRuby.put(methodName, rubyName);
     }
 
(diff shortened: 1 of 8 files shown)


headius:
This all seems fine, especially if it's slightly faster.

Note the fix I made for legacy exts that have "compat" set to 1.8.

FWIW I introduced the "packed" paths to try to shrink the amount of code we generate for populators and to reduce the amount of object churn booting up core classes. However I realize now that packing them doesn't reduce total strings since we still have to parse the names out, and may make things worse since they aren't coming from constant pool anymore.

kares:
thanks for the review.
exactly not re-using the string constants, which are already there anyway, seemed counter-productive.
will resolve/rebase with that 1_8 compat fix and merge.

PR #5051, opened 2018-02-16
Remove mathn.rb from stdlib

nomadium:
Hi folks,

This is another change targeting Ruby 2.5 [1]: Remove mathn from stdlib (feature #10169).


enebo:
@nomadium Your previous changes to test/mri were ok because they reflected copies from MRI's test in their repo.  When we resync those tests it ends up being zero sum since your submissions and those tests are identical.

For ruby/spec (e.g. ruby/spec/ruby) they did not actually remove these files right?  If anything they probably changed inclusive range from ..."2.5" to .."2.5".  I am pretty sure those should not be touched.  Perhaps we do need to update them to latest from ruby/spec though?

headius:
I updaged stdlib from 2.5 yesterday, which included removing mathn.rb. Thanks for your help!

PR #5055, opened 2018-02-19
[ruby-2.5] Implement String#delete_suffix

ChrisBr:
See feature https://bugs.ruby-lang.org/issues/13665 and https://github.com/ruby/ruby/commit/510957df33be322875e7d863b9af6934f2574779


headius:
I assume the tests just come from MRI? I can update the tests on ruby-2.5 to match Ruby 2.5.0's tests...I had not up to this point because it will mean hundreds of errors for things we haven't implemented yet, but I suppose that's a good way to know what's left too :-)

There's no harm in pulling in this test now if it's just from MRI, since we'll run their tests soon.

PR #5056, opened 2018-02-21
Time#at receives 3rd argument which specifies the unit of 2nd argument

ChrisBr:
https://bugs.ruby-lang.org/issues/13919
 #4876

### DIFF
--- a/core/src/main/java/org/jruby/RubyTime.java
+++ b/core/src/main/java/org/jruby/RubyTime.java
@@ -1184,6 +1184,12 @@ public class RubyTime extends RubyObject {
 
     @JRubyMethod(meta = true)
     public static IRubyObject at(ThreadContext context, IRubyObject recv, IRubyObject arg1, IRubyObject arg2) {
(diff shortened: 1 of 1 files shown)


core/src/main/java/org/jruby/RubyTime.java
```diff
@@ -1201,16 +1207,45 @@ public static IRubyObject at(ThreadContext context, IRubyObject recv, IRubyObjec
             millisecs = RubyNumeric.num2long(arg1) * 1000;
         }
 
+        String unit;
+        if (arg3 instanceof RubySymbol) {
+            unit = ((RubySymbol) arg3).toString();
+            throw context.runtime.newArgumentError("unexpected unit " + arg3);
```
kares: will always be raising whenever a unit Symbol arg is passed in
seems like the logic to validate supported units is missing here?

ChrisBr: @kares you were right, fixed it now!

ChrisBr:
@kares thanks for the review and sorry for the delay. Fixed now!

PR #5061, opened 2018-02-23
Add keyword_init option to Struct.new

nomadium:
Hi folks,

This is another feature targeting Ruby 2.5 [1]: Add keyword_init option to Struct.new (feature #11925).

### DIFF
--- a/core/src/main/java/org/jruby/RubyStruct.java
+++ b/core/src/main/java/org/jruby/RubyStruct.java
@@ -32,8 +32,11 @@
  ***** END LICENSE BLOCK *****/
 package org.jruby;
 
+import java.util.stream.Collectors;
+import java.util.stream.Stream;
 import org.jruby.anno.JRubyClass;
 import org.jruby.anno.JRubyMethod;
+import org.jruby.ast.util.ArgsUtil;
 import org.jruby.common.IRubyWarnings.ID;
 import org.jruby.exceptions.RaiseException;
 import org.jruby.internal.runtime.methods.DynamicMethod;
@@ -174,6 +177,7 @@ public class RubyStruct extends RubyObject {
     public static RubyClass newInstance(IRubyObject recv, IRubyObject[] args, Block block) {
         String name = null;
         boolean nilName = false;
+        boolean keywordInit = false;
         Ruby runtime = recv.getRuntime();
 
         if (args.length > 0) {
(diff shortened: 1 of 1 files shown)


nomadium:
I gave a shot to the implementation of this new feature and although all existing tests (and the new ones exercising this change) in ```test/mri/ruby/test_struct.rb``` are passing, I'm probably missing something here.

I decided to test if inheritance work as expected and I wrote the following test to verify but I found out some assertions are failing:

```ruby
  def test_struct_new_with_keyword_init_and_inheritance
    klass = @Struct.new(:a, keyword_init: true)
    klass2 = Class.new(klass)
    assert_raise(ArgumentError) { klass2.new(1) } # broken, nothing is raised: => #<struct a=1>
    assert_nothing_raised { klass2.new(a: 5) }    # nothing is raised but is wrong as well: => #<struct a={:a=>5}>
    o = klass2.new(a: 5)                          # wrong => #<struct a={:a=>5}>
    assert_equal(5, o.a)                          # broken: o.a => {:a=>5}
  end
```

Where should I look to get this working?

BTW, do you see worthwhile to contribute that extra test to MRI or ruby/spec?

headius:
Thank you for the PR!

Regarding your test:

* Contribute it to ruby/spec, ideally. It will filter back to JRuby whenever we update. ruby/spec is structured better and should always be the first place you look to contribute tests.
* If you want to try to fix it, look at RubyStruct.createStructClass and newInstance and compare with CRuby's logic. It's possible they changed the class structure surrounding Struct and we missed it.
* You can also defer and file a separate bug for the failing assertions, so we can get the PR in. We'll collab on fixing it in-place then.

nomadium:
@headius Thanks for the pointer, I just saw what's the problem with the inheritance in this case. I'll update the PR soon.

nomadium:
@headius The issue I pointed out in my previous comment is now fixed. I'll try to submit my extra test to ruby/spec later.

core/src/main/java/org/jruby/RubyStruct.java
```diff
@@ -327,11 +348,45 @@ private void checkSize(int length) {
...
+        if (args.size() != values.length) {
+            RubyArray members = __member__();
+            String unknownKeywords = Arrays.stream(args.keys().toJavaArray())
+                .filter(o -> !members.contains(o))
+                .map(IRubyObject::asJavaString)
+                .collect(Collectors.joining(", "));
+            throw getRuntime().newArgumentError("unknown keywords: " + unknownKeywords);
+        }
+    }
+
```
nomadium: I realised I was reimplementing functionality already available in JRuby with these checks. I'll update this PR soon to address that.

Issue #4781, opened 2017-09-06
Improve two-way integration of Java and Ruby exceptions


headius:
For #4699, we removed the automatic initCause call for Java exceptions raised from Ruby. There were a few issues with that logic:

* Java does not automatically set cause for you, so there seemed to be a bit of a mismatch.
* The cause can only be set once, so we must check if it has been set already. Attempting to set it again raises an illegal argument exception from Java.
* The cause field is not volatile so you cannot check and set it safely across threads. It must be set under lock, and even then it is not guaranteed to be visible.
* Cause was getting set on unrelated exceptions just because they got re-raised with $! in flight.

In light of these problems, I think we should not reinstate automatic cause-setting for Java exceptions, but we do need a way to provide a cause. There's a few ways this might be done...I try to describe some below.

RubyException could hold a reference to its "container" RaiseException, so it can be re-raised without making a new one.

When providing the Java cause, this exception could be used, either through a new JRuby-specific method:
```ruby
ex.throwable
# or
ex.to_java(Throwable)
```
...or via some magic trickery during value conversion in JI, when we see a RubyException is being passed for a Throwable.

We have also discussed in the past building a parallel exception hierarchy for RaiseException. In this structure, all exception types that currently exist would still work, but for each Ruby exception there would be an equivalent Java exception of the same name. The two would be inseparable, like the proxy objects we use for JI. This would allow the Ruby exception to behave more naturally as a Java exception, and allow Java code to rescue Ruby exceptions by name (rather than the RaiseException dance we have today).

All this combined would improve our integration between Ruby and Java exceptions.

headius:
I've started work on the parallel exception hierarchy on the parallel_exceptions branch.

headius:
I have abandoned the old branch and have a new (working) branch in exceptions2.

The logic here works like this:

* Each RubyException knows how to construct its own RaiseException type, and holds a reference to it once created lazily.
* Throw a RubyException using the form `throw ex.getRaiseException()`.

At the moment this does not introduce any new Throwable types to the system. I need to work on cleaning up initialization, especially wrt backtraces.

headius:
I have landed the parallel exception hierarchy. Will look into making it solve this issue.

headius:
Ok, without any additional modification, we can now do the following to attach a Ruby exception as the cause for a Java exception:

```
$ jruby -e 'begin; raise "hello"; rescue; raise java.lang.RuntimeException.new($!.to_java); end'
[...]

PR #5063, opened 2018-02-27
Generify IRubyObject.toJava to make it more pleasant to use.

headius:
This change allow callers to skip the cast when calling toJava
on a Ruby object. It was initially done to assist work on #4781
by allowing a simple "throw rubyException.toJava(Throwable.class).

I tested binary compatibility in two ways:

* A full recompile with all generification in place plus testing
  standard JRuby commands. This confirms that classes recompiled
  against the API but without other use of generics still function
  properly.
* A partial recompile with only the generified classes rebuilt.
  This tests that classes not compiled against the API still
  function properly.

Of course any external JRuby extensions (readline, openssl) will
have been tested by running those standard commands (e.g. irb,
gem install).

### DIFF
--- a/core/src/main/java/org/jruby/RubyBoolean.java
+++ b/core/src/main/java/org/jruby/RubyBoolean.java
@@ -162,17 +162,12 @@ public class RubyBoolean extends RubyObject implements Constantizable {
             return RubyString.newStringShared(f.getRuntime(), FALSE_BYTES);
         }
 
-        /*
-        Because Boolean objects can't cast to boolean via Class.cast, and you can't return
-        a primitive from a generified method, this impl remains ungenerified.
-         */
         @Override
-        public Object toJava(Class target) {
-            if (target.isAssignableFrom(Boolean.class) | target.equals(boolean.class)) {
-                return Boolean.FALSE;
-            } else {
-                return super.toJava(target);
+        public <T> T toJava(Class<T> target) {
+            if (target.isAssignableFrom(Boolean.class) || target == boolean.class) {
+                return (T) Boolean.FALSE;
             }
+            return super.toJava(target);
         }
     }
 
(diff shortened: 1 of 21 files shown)


headius:
Bit of a snag...it appears that Class.cast does not work when the desired target is a primitive. This is noted in the toJava for RubyBoolean, since that logic wants to return a boolean but you can't return or cast to boolean from a generified method.

This may not have a generic fix since there does not appear to be any way to programmatically cast primitives.

core/src/main/java/org/jruby/RubyBoolean.java
```diff
@@ -161,6 +161,19 @@ public static IRubyObject false_xor(IRubyObject f, IRubyObject oth) {
...
             return RubyString.newStringShared(f.getRuntime(), FALSE_BYTES);
         }
+
+        /*
+        Because Boolean objects can't cast to boolean via Class.cast, and you can't return
+        a primitive from a generified method, this impl remains ungenerified.
+         */
+        @Override
+        public Object toJava(Class target) {
+            if (target.isAssignableFrom(Boolean.class) | target.equals(boolean.class)) {
```
kares: that one `|` is pbly problematic - also you can skip equals for terminal types and primitives: 
`target == Boolean.class || target == boolean.class`

core/src/main/java/org/jruby/RubyBoolean.java
```diff
@@ -192,6 +205,19 @@ public static IRubyObject true_xor(IRubyObject t, IRubyObject oth) {
...
             return RubyString.newStringShared(t.getRuntime(), TRUE_BYTES);
         }
+
+        /*
+        Because Boolean objects can't cast to boolean via Class.cast, and you can't return
+        a primitive from a generified method, this impl remains ungenerified.
+         */
+        @Override
+        public Object toJava(Class target) {
+            if (target.isAssignableFrom(Boolean.class) | target.equals(boolean.class)) {
```
kares: same here `||` got lost as `|`

headius: Ok it turns out isAssignableFrom is required here because the contract of Ruby Boolean is that it will coerce to a Boolean object for java.lang.Boolean or any supertype up to java.lang.Object.

I'll make the other changes along with my fixes.

headius:
I have additional fixes and no longer believe the casting issue is a big problem.

After inspecting all our toJava implementations, only a handful can toJava into primitive types. Those cases will be fixed to just do manual casting to the container type.

The generic toJava in JavaProxy could be a problem, but it does not appear to break even when specifying primitive target types for a boxed numeric, probably because it doesn't really coerce to a different type:

```
$ jruby -e 'p 1.to_java(:int).to_java(:char)'
#<Java::JavaLang::Integer:0x30a3107a>
```

This may constitute a bug and get fixed at some point, so I'll make sure the blind casting here is at least smart about primitives.

PR #5065, opened 2018-02-27
Parallel exception hierarchy for Java

headius:
This is to address problems related to #4781.

No Ruby exception types have a real exception type in Java, which means in order to capture any of them you must catch all of them (`catch (RaiseException)`) and inspect or rethrow. This is inefficient and ugly.

This PR introduces a parallel Java exception hierarchy that maps 1:1 to the base Ruby exception hierarchy.

* All RubyException objects now know how to construct their appropriate Throwable.
* All base exception types will have their own java.lang.Throwable (subclass of RaiseException).

In addition, this makes it possible for JRuby ext authors to create the same parallel Throwable+RubyException within their libraries to improve exception handling from Java (think `catch (ParserError)` in the Psych extension).

### DIFF
--- a/core/src/main/java/org/jruby/AbstractRubyException.java
+++ b/core/src/main/java/org/jruby/AbstractRubyException.java
@@ -52,8 +52,10 @@ import java.io.PrintStream;
 import static org.jruby.runtime.Visibility.PRIVATE;
 
 /**
+ * An abstract representation of a Ruby exception. All Ruby exception types descend from here.
  *
- * @author  jpetersen
+ * Logic here handles aggregating a cause, connecting up a Java throwable, managing the backtrace, and other Ruby
+ * behavior surrounding exceptions.
  */
 public abstract class AbstractRubyException extends RubyObject {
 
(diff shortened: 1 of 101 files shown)


headius:
More work coming, but mostly just filling out the hierarchy.

headius:
I've implemented this for all types of exceptions standard to Ruby, plus our few additions, but I have *not* done all the errno exceptions. Should I?

I have a minor concern about cluttering up org.jruby with more classes, but I'm not sure where else the new Ruby*Exception types should live. Should we move them somewhere?

All the new Throwables live under org.jruby.exceptions.

This is just about ready, pending questions above and fixing any regressions.

kares:
:100: very nice (and backwards compatible)

minor thing: I know we use get-er style naming for conversions but still
`public RaiseException getRaiseException()` maybe its better (might be subjective) as : 
`public RaiseException toRaiseException()` (or `asRaiseException`) so the 'conversion' intent is clear?

headius:
> minor thing: I know we use get-er style naming for conversions but still
public RaiseException getRaiseException() maybe its better (might be subjective) as :
public RaiseException toRaiseException() (or asRaiseException) so the 'conversion' intent is clear?

Yeah I have also had concerns about some of these names being rather long. Maybe `throwable` or `toThrowable`? It would still return RaiseException.

I believe this is green other than things also failing on master.

PR #5071, opened 2018-03-01
Better use of ThreadContext constants

headius:
There are two changes here:

* A global search and replace for the `context.runtime.getTrue()` for true, false, and nil and replacing them with the `context.tru` pattern.
* ~~Retyping the `tru`, `fals`, and `nil` fields on ThreadContext to be the most specific type.~~

~~The latter change is binary incompatible if any third-party libraries have started to use these fields directly. However these fields were introduced fairly recently.~~

The second change has been removed because of additional internal compatibility problems. There are many places in code generation, and likely in any AOT code in the wild, that try to access the `nil` field using `IRubyObject`.

### DIFF
--- a/core/src/main/java/org/jruby/AbstractRubyMethod.java
+++ b/core/src/main/java/org/jruby/AbstractRubyMethod.java
@@ -145,10 +145,10 @@ public abstract class AbstractRubyMethod extends RubyObject implements DataType 
     }
 
     protected IRubyObject super_method(ThreadContext context, IRubyObject receiver, RubyModule superClass) {
-        if (superClass == null) return context.runtime.getNil();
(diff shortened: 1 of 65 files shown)


headius:
Correction: the `tru` and `fals` fields have existed [for about a year](https://github.com/jruby/jruby/commit/b0f3e366ca8d3279ac43c8ea8401edca5cf730ef#diff-ea5ad4da5d5e74f7e2c19855028b12dc), so roughly 9.1.7.0ish. The `nil` field has been around [substantially longer](https://github.com/jruby/jruby/commit/e373c01f64dc8cd4f29517e6d56e8a4db9a90177), since 1.6 times.

Issue #4506, opened 2017-02-24
Segfault in jnr-ffi finalizer


headius:
This is copied from #4312 where @shirosaki discovered jnr-ffi crashes in the finalizer for its generated assembly code stubs. Repro and output follows.

```java
import java.lang.reflect.Method;
import java.net.URL;
import java.net.URLClassLoader;

public class TestReload {
    public static void main(String[] args) {
        while (true) {
            Thread t = new Thread() {
                public void run() {
                    try (URLClassLoader loader = new URLClassLoader(new URL[] { 
                                new URL("jar:file:./jruby-complete.jar!/")
                            })) {
                        // load class
                        Class<?> container = loader.loadClass("org.jruby.embed.ScriptingContainer");
                        Method runScriptlet = container.getMethod("runScriptlet", String.class);
                        Method setClassLoader = container.getMethod("setClassLoader", ClassLoader.class);
                        Method terminate = container.getMethod("terminate");
                        Object c = container.newInstance();
                        setClassLoader.invoke(c, loader);
                        runScriptlet.invoke(c, "puts \"Hello World!\"");
                        terminate.invoke(c);
                    } catch (Exception e) {
                        e.printStackTrace();
                    }
                    // unload class
                    System.gc();
                    System.runFinalization();
                    System.gc();
                }
            };
            t.start();
            try {
                t.join();
            } catch (Exception e) {
		e.printStackTrace();
            }
        }
    }
}
```

```sh
$ java -XX:+TraceClassUnloading TestReload
(snip)
[thread 9492 also had an error]
#
# A fatal error has been detected by the Java Runtime Environment:
#
#  EXCEPTION_ACCESS_VIOLATION (0xc0000005) at pc=0x000000006e4c24b0, pid=11608, tid=0x00000000000031e4
#
# JRE version: Java(TM) SE Runtime Environment (8.0_111-b14) (build 1.8.0_111-b14)
# Java VM: Java HotSpot(TM) 64-Bit Server VM (25.111-b14 mixed mode windows-amd64 compressed oops)
# Problematic frame:
# C  0x000000006e4c24b0
#
# Failed to write core dump. Minidumps are not enabled by default on client versions of Windows
#
# An error report file with more information is saved as:
(snip)
```

hs_err_pid11608.log contains:
```
#
# A fatal error has been detected by the Java Runtime Environment:
#
[...]

PR #5072, opened 2018-03-01
update jffi and release version for JRuby 1.7

DJRickyB:
prevents SEGFAULTs described in https://github.com/jruby/jruby/issues/4506

I understand the intent is not to update 1.7 anymore, but this change is being tested at our organization currently and could save us a lot of grief


headius:
Sorry for the delay in looking at this. Unfortunately we do not have plans or resources to do a 1.7.28 release. I'll ping @enebo here and we can discuss a path forward. Is there no way for you to update to a JRuby 9.x version?

headius:
Oops, I marked this for 9.1.17 without realizing 9.1 has already been bumped up to jffi 1.2.16.

DJRickyB:
Actually we were still getting segfaults even after this patch.  I see the initial issue has some recent activity.  We are going to be pushing our devs to convert legacy projects to 9k for sure

headius:
@DJRickyB Yes, in #4506 there's discussion about how to fix this better in jnr-ffi, but so far no patch from the contributor. It would likely require a rebuild of the native bits of jffi, which means there's a bit of a long tail in getting it released (we support some obscure platforms that would need rebuilds). But there's a solution out there.

Can you describe your application and how you're running it? There are some workarounds that may help.

DJRickyB:
we have multiple rails apps co-existing on tomcat using (now) a patched version of JRuby 1.7 with the change in this PR.  Each is in a separate warbler-generated war running via jruby-rack.

  Sporadic segfaults have been traced back to JFFI and match #4506

PR #5074, opened 2018-03-01
Fix file descriptor leak in resolv.rb

quixoten:
There's an edge case in resolv.rb that can lead to file descriptor leaks on a highly saturated system (logstash in our case).

file-descriptor-leak

(the cliff at the end is from a restart of logstash)

We're using the logstash-filter-dns plugin which [uses timeout](https://github.com/logstash-plugins/logstash-filter-dns/blob/v3.0.7/lib/logstash/filters/dns.rb#L250-L252).

When a timeout fires while [waiting](https://github.com/jruby/jruby/blob/9.1.13.0/lib/ruby/stdlib/resolv.rb#L659-L668) for a [random](https://github.com/jruby/jruby/blob/9.1.13.0/lib/ruby/stdlib/resolv.rb#L817) [port](https://github.com/jruby/jruby/blob/9.1.13.0/lib/ruby/stdlib/resolv.rb#L764), the new object is never returned to the caller. Instead `requester` is [set to nil](https://github.com/jruby/jruby/blob/9.1.13.0/lib/ruby/stdlib/resolv.rb#L524). There's no handle to any sockets that were created in the initializer, so they can't be closed, and one ore more file descriptors is leaked.

### DIFF
--- a/lib/ruby/stdlib/resolv.rb
+++ b/lib/ruby/stdlib/resolv.rb
@@ -522,7 +522,6 @@ class Resolv
       requester = make_udp_requester
       senders = {}
       begin
-        requester.lazy_initialize
         @config.resolv(name) {|candidate, tout, nameserver, port|
           msg = Message.new
           msg.rd = 1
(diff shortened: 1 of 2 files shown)


headius:
Great find! We just ship CRuby's resolv.rb...so it seems likely they'd be affected too. I'll have a look at your patches.

headius:
Ok yeah, I am pretty sure CRuby would be affected by this as well. We need to submit this to them as a patch with backports to 2.3.

quixoten:
I'll work on getting it submitted to CRuby.

quixoten:
https://bugs.ruby-lang.org/issues/14571

quixoten:
The patch was accepted into CRuby. Should this be closed now?

headius:
We'll merge it in, thanks!

headius:
It doesn't appear they have committed this to their 2.5 branch yet, so I'll merge to our fork for now.

headius:
Once there's a backport for 2.3 we can add this to JRuby 9.1.17.

quixoten:
Looks like this made it into 2.3. Would it be possible to add this to a 9.1.17 release?

PR #5076, opened 2018-03-02
redo date.rb (most) parts in native

kares:
still using JODA but all of the JI stuff in date.rb is gone behind ext classes, motivation : 

- for a while its 'hard' to create `Date` instances in AR-JDBC
- some near era conversion had a nasty bug in all 9K which constantly pops up in AR-JDBC
- some isolated AR benchmarks showed its slower than MRI
- there's going to be some nice JI bonuses - `toJava` working as with `RubyTime` etc.

some parts were already in native (formatting) so the base setup was already there.

NOTE: in MRI pretty much all of the date.rb is in native - so there's still space for improvement.
e.g. `_parse` would make sense to end up in .java maybe someone will attempt it afterwards.

p.s. left behind one minor suite failure (argument validation `hour: 24+1.to_r/2`)
while its simple the bigger picture (of MRI argument handling) feels a bit weird. 
some parts handle the overflow some not, smt Float do not work while Rational is ok - clunky.
would eventually circle back as I feel like that might end up as a MRI report/feature request. 

needs rebase (waiting for tests) but would like to get this in so I can spent time testing + adding JI stuff.

### DIFF
--- a/core/src/main/java/org/jruby/Ruby.java
+++ b/core/src/main/java/org/jruby/Ruby.java
@@ -5251,6 +5251,8 @@ public final class Ruby implements Constantizable {
     // I know of use very few of them.  Even if there are many the size of these lists are modest.
     private final Map<String, List<StrptimeToken>> strptimeFormatCache = new ConcurrentHashMap<>();
 
+    transient RubyString tzVar;
+
     @Deprecated
     private void setNetworkStack() {
         deprecatedNetworkStackProperty();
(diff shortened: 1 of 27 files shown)


mkristian:
❤️ 
might ping you guys on irc on day to talk about repackage an/or third-party jars and its risks for jruby-9.2

kares:
already tested with AR-JDBC - all seems well (and working backwards compatible even on master).
likely going to merge this soon, so I can finish up some JI stuff that will require a set of second eyes.

PR #5083, opened 2018-03-09
Fix time interval test

ChrisBr:
and use RubyTime.convertTimeInterval for parsing to raise correct ArgumentError and TypeError exceptions


kares:
this introduced slightly more failures than it fixed, see CI's log.
e.g. with JI on places were a Numeric is expected (such as `sleep`) Java proxy numbers work as well.
... sample failure for reference https://travis-ci.org/jruby/jruby/jobs/351484156#L1555

ChrisBr:
@kares thanks for the hint.

Unfortunately, the ``Java::JavaLang::Long`` gets passed in as a ``org.jruby.RubyString``. So there is no way to distinguish if the RubyString is from the Long or if it was really a String. So seems there is no (easy) way to fix it or do I miss something?

kares:
oh really? that ain't good than and the "regression" is legit, I will look into a way of pushing this forward ...

ChrisBr:
@kares tried again, seems I did sth wrong the last time :cry: It get's of course passed in as a ``ConcreteJavaProxy``. Anyway, see my updated patch which passes now both tests.

kares:
Thanks Chris, poked around and found out the time-interval method is used for elsewhere.
So, I made sure it works from all the places: https://github.com/jruby/jruby/commit/d19e85043d7ba52caefd7a5232a0fd82da632e7b when a Number proxy is being passed in ...

isn't nice but at least works, for now. the whole piece is a new feature in 9.2, and I recall specifically `sleep(java_num)` was a reported issue that motivated implementing support for Java numbers.

The commit from here is on master as: https://github.com/jruby/jruby/commit/fd9c772864b7670881fd2e561b6a8317d2ed8ef6

PR #5093, opened 2018-03-19
upgrade joda-time library to latest 2.9.9

kares:
... not sure if there was a previous reason to stay so far behind

most changes are bugfixes and more importantly matching up TZ data rules
maybe the time-zone data should get regenerated as well ... https://github.com/jruby/joda-timezones ?


kares:
@headius any feedback here, please?

enebo:
@kares seems reasonable to me.  We want a long bake in case and date/time weirdness falls out of this too.  Sooner we land the better

PR #5094, opened 2018-03-19
Reverse the order in which we print out the backtrace. #4876

headius:
This implements the reverse backtrace introduced in Ruby 2.5. Notable missing bits:

* This does not check for TTY to decide whether to reverse; it always reverses for the printing-out cases and always does not reverse for internal logging.
* This provides no way to disable the reverse trace and go back to a forward trace.

See #4876.


headius:
Low-priority item that can perhaps be done along with compat work in 9.3.

headius:
Dupe of #5510 and closed for the same reasons.

PR #5108, opened 2018-03-25
Thread.report_on_exception default value is now true

nomadium:
Hi folks,

This is another change targeting Ruby 2.5 [1]: Thread.report_on_exception default value is now true (feature 14143).

2. https://bugs.ruby-lang.org/issues/14143

### DIFF
--- a/core/src/main/java/org/jruby/Ruby.java
+++ b/core/src/main/java/org/jruby/Ruby.java
@@ -1388,7 +1388,7 @@ public final class Ruby implements Constantizable {
         trueObject = new RubyBoolean.True(this);
         trueObject.setFrozen(true);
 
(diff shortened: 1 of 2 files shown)


kares:
are you sure about the warning message format? seems they agreed on smt different on the tracker ...

nomadium:
@kares That's the message that I observed when tested MRI and also I read it in their source code. Please see below:

- https://github.com/ruby/ruby/blob/v2_5_0/thread.c#L654
- https://github.com/ruby/ruby/commit/5a3c024df0a83d1f64f10f028f1082e67f88c013#diff-4fb69dc1bf667cfbc2b05dc5fd51e674L651

PR #5113, opened 2018-03-26
Numeric#clone and #dup no longer raises TypeError

nomadium:
Hi folks,

This is another change targeting Ruby 2.5 [1]: Numeric#clone and #dup no longer raises TypeError (bug #13237).

It returns the receiver instead as well as Integer and Float.

2. https://bugs.ruby-lang.org/issues/13237


headius:
Thanks!

PR #5137, opened 2018-04-12
Use ConcurrentWeakHashMap for RubyClass.subclasses

headius:
The naively-synchronized WeakHashSet does not guarantee safe
iteration in the presence of modification, which triggers a
ConcurrentModificationException as seen in #5115.

By using a ConcurrentWeakHashMap (designed by Doug Lea et al, see
http://www.java2s.com/Code/Java/Collections-Data-Structure/Ahashtablewithemweakkeysemfullconcurrencyofretrievalsandadjustableexpectedconcurrencyforupdates.htm)
we can iterate and modify concurrently without errors.

However this implementation is based the design of
ConcurrentHashMap, which means the in-memory structure will likely
be considerably larger than a simple WeakHashMap. I believe it is
also based off the pre-Java 8 ConcurrentHashMap, which means any
load factor above one thread greatly increases in-memory size, and
a load factor of one limits the concurrent use of the collection.

### DIFF
--- a/core/src/main/java/org/jruby/RubyClass.java
+++ b/core/src/main/java/org/jruby/RubyClass.java
@@ -91,6 +91,7 @@ import org.jruby.util.OneShotClassLoader;
 import org.jruby.util.ClassDefiningClassLoader;
 import org.jruby.util.CodegenUtils;
 import org.jruby.util.JavaNameMangler;
+import org.jruby.util.collections.ConcurrentWeakHashMap;
 import org.jruby.util.collections.WeakHashSet;
 import org.jruby.util.log.Logger;
 import org.jruby.util.log.LoggerFactory;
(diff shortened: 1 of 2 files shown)


headius:
It's hardy scientific, but using VisualVM to get a heap dump of the following script:

```ruby
a = Object
10000.times { a = Class.new(a) }
puts 'ok'
sleep
```

Shows a bit over a 2MB size difference, or about 200 bytes per class. This may increase with more subclasses in each collection, but it should be a much smaller increment than the base hit of the collection.

Note that I'm forcing a load factor of 1, so we are getting no concurrency out of this collection...just thread-safe traversal+modification.

Before:
image

After:
image

headius:
This is a pretty non-invasive patch, and the cost of 200 bytes per class doesn't seem *too* bad, so I vote that we go with it.

PR #5121, opened 2018-03-27
Require real path

ChrisBr:
expand load paths to real paths to get rid of duplicate loading from
symbolic-linked directories. [Feature #10222]
ruby/ruby@b6d3927

This fixes

test/mri/ruby/test_require.rb#test_symlink_load_path

Deletes obsolete spec/regression/GH-1940_symlink_load_path_should_not_expand_in_FILE_spec.rb

#5109 #5117

PR #5149, opened 2018-04-23
Fix TestLoad#test_symlinked_jar

ChrisBr:
Introduced in https://github.com/jruby/jruby/pull/5121.
Ruby 2.5 loads now the real path of symlinked libraries.
We will now use the file extension of the symlink
to decide what ResourceLibrary to load.

### DIFF
--- a/core/src/main/java/org/jruby/runtime/load/LibrarySearcher.java
+++ b/core/src/main/java/org/jruby/runtime/load/LibrarySearcher.java
@@ -170,12 +170,11 @@ class LibrarySearcher {
             if (resource.absolutePath() != resource.canonicalPath()) {
                 FileResource expandedResource = JRubyFile.createResourceAsFile(runtime, resource.canonicalPath());
                 if (expandedResource.exists()){
-                    resource = expandedResource;
-                    String scriptName = resolveScriptName(resource, resource.canonicalPath());
-                    String loadName = resolveLoadName(resource, searchName + suffix);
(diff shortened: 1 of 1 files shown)


headius:
We'll let this one finish just to check that test:jruby goes green. It should be green if not for the introduced failure. Thank you for following through on this!

ChrisBr:
@headius the only difference is now that we pass the resource and not the expanded resource to ```ResourceLibrary.create(searchName, scriptName, resource)``` which decides which ResourceLibrary gets created based on the file extension.

However, this will now _always_ use the file extension of the symlink (the extension of the expanded symlink gets always ignored). Is this what we want?

headius:
@ChrisBr That seems to be the MRI behavior, if the tests are to be trusted. As long as this fixes the regression and doesn't introduce anything else, we'll go with it.

I'm sure the truth is somewhere in the MRI load/require code, but that's a really tangled mess.

ChrisBr:
After talking with @headius, I double checked the behavior in MRI and it seems to be undefined:

When I create a symlink ``sym.so`` to ``nokogiri.so`` I get this exception

```
/tmp/ruby-test/sym.so: undefined symbol: Init_sym - /tmp/ruby-test/sym.so
```

The same happens also without file extension.

headius:
I will modify the JRuby test, since it doesn't appear that MRI traverses symlinks for "extensions" like our .jar, which is the only failing test that regressed.

PR #5151, opened 2018-04-24
Bytelist love merge

enebo:
So the branch is a bit misnamed at this point but I will summarize this a tiny bit (have talked quite a lot to @headius about this already).

all identifiers are made as RubySymbols with exception of annotations (currently assumed as 7bit ASCII -- this can be changed later).  RubySymbols have bytelist + 8859_1 identifier string.  Parser and IR now use Symbol.  rest of runtime continues to use the j.l.String as identifier/intern string.  New vocabulary:
  - RubySymbol - name of the identifier
  - String - id of the identifier

RubySymbol.idString() is how you get an id for what you are looking up.

Error reporting has new static str(), names(), and ids() methods for making properly encoded Error strings.  This is not 100% completed but many many errors have been converted.


core/src/main/java/org/jruby/Ruby.java
```diff
@@ -1097,11 +1103,10 @@ public RubyModule defineModuleUnder(String name, RubyModule parent) {
         if (moduleObj != null ) {
             if (moduleObj.isModule()) return (RubyModule)moduleObj;
 
-            if (parentIsObject) {
-                throw newTypeError(moduleObj.getMetaClass().getName() + " is not a module");
-            } else {
-                throw newTypeError(parent.getName() + "::" + moduleObj.getMetaClass().getName() + " is not a module");
-            }
+            RubyString typeName = parentIsObject ?
+                    types(this, moduleObj.getMetaClass()) : types(this, parent, moduleObj.getMetaClass());
```
kares: maybe a `throw`is lost here - unless its intentional

core/src/main/java/org/jruby/RubyBasicObject.java
```diff
@@ -802,19 +805,25 @@ public RubyString convertToString() {
...
-        ByteList bytes = new ByteList(2 + cname.length() + 3 + hex.length() + 1);
-        bytes.append('#').append('<');
-        bytes.append(cname.getBytes(RubyEncoding.UTF8));
-        bytes.append(':').append('0').append('x');
-        bytes.append(hex.getBytes());
-        bytes.append('>');
-        RubyString str = RubyString.newString(getRuntime(), bytes, UTF8Encoding.INSTANCE);
-        str.setTaint(isTaint());
-        return str;
+        RubyString bytes = runtime.newString("#<");
```
kares: just a q: so the direct `ByteList` `concat` wasn't good here, does it get problematic for any reason? 
isn't the new way going to be doing a lot of "unnecessary" Java `String`-> `byte[]` conversion?

core/src/main/java/org/jruby/RubySymbol.java
```diff
@@ -150,15 +153,57 @@ public final String asJavaString() {
...
-    final ByteList getBytes() {
+    public final ByteList getBytes() {
         return symbolBytes;
     }
 
+    /**
+     * Make an instance variable out of this symbol (e.g. :foo will generate :foo=).
+     * @return the new symbol
+     */
+    public RubySymbol asAccessor() {
```
kares: did you mean `asWriter` (not having `asReader` seems as just a coincidence since its the same symbol) ?

core/src/main/java/org/jruby/RubySymbol.java
```diff
@@ -150,15 +153,57 @@ public final String asJavaString() {
...
+    /**
+     * Return an id string (e.g. raw ISO-8859_1 charset String) for use with our method tables etc.
+     */
+    public String idString() {
+        return symbol;
+    }
+
     @Override
     public final String toString() {
-        return symbol;
```
kares: this could use an explanation on why not simply `return symbol;` ...for mere mortals :)

core/src/main/java/org/jruby/util/RubyStringBuilder.java
```diff
@@ -0,0 +1,248 @@
...
+            newResult.cat('"');
+            result = newResult;
+        }
+        return result;
+    }
+    
+    public static String str(Ruby runtime, IRubyObject value, String message) {
+        RubyString buf = (RubyString) value.asString().dup();
+
+        buf.cat19(runtime.newString(message));
```
kares: do we still need to be using 19 legacy named methods? maybe `cat` + `cat19` are worth some 'unification' ...

core/src/main/java/org/jruby/Ruby.java
enebo: throw is immediately below that line.  It was two throws and now is one line since both throws just were assembling a proper typeName.

core/src/main/java/org/jruby/RubyBasicObject.java
enebo: That is a good question I don't recall offhand.  It is possible there was no issue but I learned that these methods are not quite a sensible as you may think.  I will try it out quickly.

core/src/main/java/org/jruby/RubySymbol.java
enebo: Ah nice.  You're right I should have called this asWriter().  I will change that.

core/src/main/java/org/jruby/RubySymbol.java
enebo: Ah yeah I will add an explanatory comment.  Largely toString is doing it's best to make a j.l.String which displays properly but is not perfect since toString() should not be used internally for anything but debugging.  Going full RubyString + decode seemed unnecesary here.  Perhaps I should also add asJavaString() or using str() is more appropriate if you want to work with a string version of a RubySymbol?

core/src/main/java/org/jruby/util/RubyStringBuilder.java
enebo: Unfortunately, they are very different and both should always exist.  I would agree with you that we should refactor with better names as cat19 means nothing but I suspect cat19 is still used by multiple native extensions which use m17n strings in any way (although we can definitely come up with a new name and deprecate).  Main difference is one could care less about what it is cat'ing and the other realizes it needs encoding negotiation.

kares: ah yeah I recall them being impossible to unify now ... thanks

core/src/main/java/org/jruby/RubySymbol.java
kares: overriding `asJavaString` explicitly (from IRubyObject) sounds good ... for the full decoding cycle.
unless its already doing so

enebo: I also just found a few consumers of toString and changed them to idString or asJavaString depending on whether their context was used in an identifier or non-identifier way (for RubySymbol).  I will likely be trying to audit more general toString() usage as we seem to be a bit inconsistent.

core/src/main/java/org/jruby/RubyBasicObject.java
enebo: So I looked at this and I think I can reduce the cost substantially and use a bytelist but there is a specific reason why it works.  First when we have m17n data getting combined into a string we cannot use bytelist since bytelist has no idea how to combine two different encoded things together.  In this very particular case we only have 7bit ascii strings and something else.  So I think I can just mark the new string as same encoding as the something else and it will work....almost anyways.  If the encoding used is not isAsciicompat we will generate garbage.  OTOH my current code was broken in the same way so we are no worse off.

enebo: Addendum since I am thinking about this.  I should try the last broken part here.  I suspect the types() method in RubyStringBuilder is actually what the non-ascii name will display like.   So I need to try that.

core/src/main/java/org/jruby/RubySymbol.java
enebo: @kares oh I should point out asJavaString cannot be decoded version of the method since there are still over 100 users using it to lookup methods.  We have this pattern:

```java
    public static IRubyObject get_inner_class(final ThreadContext context,
        final RubyModule self, final IRubyObject name) { // const_missing delegate
        final String constName = name.asJavaString();
```

All our internal tables for identifiers are keyed off of 8859_1.   I think we may need to make a slightly more restrictive but helpful method to replace this pattern and ween off of asJavaString (idString on RubySymbol is more correct but only on RubySymbol -- probably need something on RubyString as well).  This though need not be resolved immediately.  Just letting you know why asJavaString cannot decode today.

As an aside it is a little weird IRubyObject has asJavaString since this means I could pass a fixnum in here and it would end up as "5" or something.  From Ruby side of things I think people would be surprised that would work even if there was a "5" method.

Issue #4755, opened 2017-08-27
Implement get_dns_server_list


ahorek:
### Environment

jruby 9.2.0.0-SNAPSHOT (2.4.1) 2017-08-20 c22987a Java HotSpot(TM) 64-Bit Server VM 25.131-b11 on 1.8.0_131-b11 +jit [mswin32-x86_64]

#4687

### Expected Behavior

get_dns_server_list is not implemented on jruby
https://github.com/jruby/jruby/blob/d3d918af43fc56b57640cc5359e9184b97aae690/lib/ruby/stdlib/win32/resolv.rb#L75

### Actual Behavior

https://github.com/ruby/ruby/blob/202bbda2bf5f25343e286099140fb9282880ecba/ext/win32/resolv/resolv.c

`
>jgem install benchmark-ips
ERROR:  While executing gem ... (NameError)
    undefined local variable or method `get_dns_server_list' for Win32::Resolv:Module
`

headius:
Indeed it is not! Perhaps you can help?

ahorek:
https://github.com/jruby/jruby/pull/5158 it's a workaround and it doesn't fix the original issue https://bugs.ruby-lang.org/issues/12604
I wasn't able to find simmilar API in Java. Can I use the C extension from iphlpapi.h in JRuby using FFI / JNI?

headius:
@ahorek If it's possible to do it through Ruby FFI that would probably be the quickest option. Go for it, and feel free to ping gitter or IRC with questions!

headius:
Marking for 9.2 since we at least seem to have a workaround.

ahorek:
thanks, I'll look into it!

PR #5158, opened 2018-05-05
fix resolv on windows

ahorek:
bug #4755

### DIFF
--- a/core/src/main/java/org/jruby/Ruby.java
+++ b/core/src/main/java/org/jruby/Ruby.java
@@ -1768,6 +1768,10 @@ public final class Ruby implements Constantizable {
         });
 
         addLazyBuiltin("cgi/escape.jar", "cgi/escape", "org.jruby.ext.cgi.escape.CGIEscape");
+
+        if(Platform.IS_WINDOWS) {
+          addLazyBuiltin("win32/resolv.jar", "win32/resolv", "org.jruby.ext.win32.resolv.Win32Resolv");
+        }
     }
 
     private void initRubyKernel() {
--- a/core/src/main/java/org/jruby/ext/win32/resolv/Win32Resolv.java
+++ b/core/src/main/java/org/jruby/ext/win32/resolv/Win32Resolv.java
@@ -1,0 +1,30 @@
+package org.jruby.ext.win32.resolv;
+
+import java.util.List;
+import org.jruby.Ruby;
+import org.jruby.RubyArray;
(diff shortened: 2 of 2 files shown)


kares:
`sun.net.dns.ResolverConfiguration` is OpenJDK specific right? and it will get hidden away under Java 9+
definitely better than nothing I am just asking if you have any input or ideas on doing this in a different way.

ahorek:
sun.net.dns.ResolverConfiguration is marked as depricated, but it still works
jruby 9.2.0.0-SNAPSHOT (2.5.0) 2018-05-10 00c19bb Java HotSpot(TM) 64-Bit Server VM 10.0.1+10 on 10.0.1+10 +jit [mswin32-x86_64]
btw. with this patch the behaviour is the same as 9.1.17.0

I think the only way to do this right and fix https://bugs.ruby-lang.org/issues/12604 is to reimplement this
https://github.com/ruby/ruby/blob/202bbda2bf5f25343e286099140fb9282880ecba/ext/win32/resolv/resolv.c using JNI.

What do you think? Should I try it or is there a better way?

headius:
@ahorek Is this fix perhaps "good enough" to merge now in case we can't do the "right" fix before the 9.2 release (within a week from now)?

ahorek:
my only concern is that sun.net.dns.ResolverConfiguration will become deprecated (I tested it on Oracle's Java 8 and Java 10 and it works fine for now)
it should be fixed before 9.2 release, otherwise bundler won't work on Windows.

headius:
> my only concern is that sun.net.dns.ResolverConfiguration will become deprecated

A valid concern. There are builds of 11, can you try to test one of them too? That would get us working into the next LTS version of OpenJDK.

> it should be fixed before 9.2 release, otherwise bundler won't work on Windows.

Yikes! 😨 cc @enebo

ahorek:
I've never used jnr.ffi before and I'm a little bit stucked :). Maybe someone could help me with it? https://github.com/jruby/jruby/pull/5178

ahorek:
https://github.com/jruby/jruby/commit/daeb255a3398b416eac43bf9dc3b0c445f4f0781 fixed the issue, thanks @enebo

enebo:
@ahorek thanks for the first stab too.  I agree that jnr/ffi and friends are super confusing in this case.  Using the previous call for number of bytes to next call is weird.  In Ruby FFI it was a bit easier to figure out.  Plus Ruby...

PR #5161, opened 2018-05-10
[ji] lazy loading of Java (proxy) class extensions

kares:
- seems to work fine to the limited testing that I did locally
- the gist is - Java extensions load on demand as Java classes are (first) used
- theoretically we could also add a flag to disable the laziness and behave as before

also ... we could safely add more extensions now and do not have to worry much about bloating load times

did not run any numbers yet, but hopefully we do not do much JI internally on `gem list`, `rake` etc

### DIFF
--- a/core/src/main/java/org/jruby/javasupport/ext/JavaExtensions.java
+++ b/core/src/main/java/org/jruby/javasupport/ext/JavaExtensions.java
@@ -1,0 +1,43 @@
+package org.jruby.javasupport.ext;
+
+import org.jruby.Ruby;
+import org.jruby.RubyModule;
+
+import java.util.Map;
+import java.util.WeakHashMap;
+import java.util.function.Consumer;
+
+/**
+ * Lazy Java class extensions initialization.
+ *
+ * @note Internal API
+ * @author kares
+ */
+public class JavaExtensions {
+
+    private static final boolean IMMEDIATE = false;
+
+    private static final JavaExtensions INSTANCE = new JavaExtensions();
+
+    public static JavaExtensions getInstance() { return INSTANCE; }
+
+    static void put(final Ruby runtime, Class javaClass, Consumer<RubyModule> proxyClass) {
+        if (IMMEDIATE) {
+            proxyClass.accept( org.jruby.javasupport.Java.getProxyClass(runtime, javaClass) );
+            return;
+        }
+        Object previous = INSTANCE.definitions.put(javaClass, proxyClass);
+        assert previous == null;
+    }
+
+    private JavaExtensions() { /* hidden */ }
+
+    final Map<Class, Consumer<RubyModule>> definitions = new WeakHashMap<>();
+
+    public void define(final Class javaClass, final RubyModule proxyClass) {
+        definitions.getOrDefault(javaClass, NOOP).accept(proxyClass);
+    }
+
+    private static final Consumer<RubyModule> NOOP = (noop) -> { /* no extensions */ };
+
+}
--- a/core/src/main/java/org/jruby/javasupport/ext/JavaExtensions.java
+++ b/core/src/main/java/org/jruby/javasupport/ext/JavaExtensions.java
@@ -17,25 +17,20 @@ public class JavaExtensions {
 
     private static final boolean IMMEDIATE = false;
 
-    private static final JavaExtensions INSTANCE = new JavaExtensions();
-
-    public static JavaExtensions getInstance() { return INSTANCE; }
+    private JavaExtensions() { /* hidden */ }
 
     static void put(final Ruby runtime, Class javaClass, Consumer<RubyModule> proxyClass) {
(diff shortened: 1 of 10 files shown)


headius:
No prob, will review. Have you managed to measure anything yet?

kares:
quick test of `time bin/jruby -S gem list`

```
# BEFORE:
 
real	0m1.823s
user	0m2.945s
sys	0m0.157s

real	0m1.778s
user	0m2.951s
sys	0m0.162s

real	0m1.769s
user	0m2.884s
sys	0m0.171s

AVG 1.79s


# AFTER 

real	0m1.739s
user	0m2.780s
sys	0m0.165s

real	0m1.736s
user	0m2.847s
sys	0m0.144s

real	0m1.712s
user	0m2.764s
sys	0m0.186s

AVG 1.73s
```
... gets us around 5% on my machine

core/src/main/java/org/jruby/javasupport/ext/JavaLang.java
```diff
@@ -59,32 +59,28 @@
...
+            proxyClass.defineAlias("to_str", "to_s");
+        });
     }
 
     @JRubyModule(name = "Java::JavaLang::Iterable", include = "Enumerable")
     public static class Iterable {
 
-        static RubyModule define(final Ruby runtime) {
-            final RubyModule Iterable = Java.getProxyClass(runtime, java.lang.Iterable.class);
+        static RubyModule define(final Ruby runtime, final RubyModule Iterable) {
```
headius: Could we use a different name for these? I find it extremely confusing when I look at the code and I see "Iterable" meaning two different things.

headius: Note that at this point in the code, Iterable now means three different things: the inner class, the variable, and (to me at a glance) the actual Java Iterable class. It's quite confusing.

headius: Same comment applies to all other such names below. I understand that it might look a bit nicer to some, but this flies against typical Java local variable naming convention.

It's somewhat confusing that the extension classes have the same names as the Java class they extend, but that seems like less of an issue to me than the local variables having yet again the same name.

core/src/main/java/org/jruby/javasupport/ext/JavaExtensions.java
```diff
@@ -0,0 +1,43 @@
...
+ * Lazy Java class extensions initialization.
+ *
+ * @note Internal API
+ * @author kares
+ */
+public class JavaExtensions {
+
+    private static final boolean IMMEDIATE = false;
+
+    private static final JavaExtensions INSTANCE = new JavaExtensions();
```
headius: This should not be static; a better place would be for it to live on the javasupport Java instance attached to the runtime.

It is weak on the value, but it also anchors a class as its key. The classes we currently extend are system-level classes, but this may be used in the future for classes loaded as part of JRuby or within extensions (e.g. jruby-openssl) that load into their own classloaders, and this reference will prevent them from being collected.

It also is not compatible with multiple JRuby instances in the same JVM, since only the first one will register (and eventually apply) an extension. The others will race to insert the extension or remove it when applying that extension.

headius: Oh somehow my review used an older version that was still static. I see you have fixed that, thanks!

core/src/main/java/org/jruby/javasupport/ext/JavaLang.java
kares: okay, will take a look into that

kares:
> The static reference needs to be fixed, and the naming is problematic, but otherwise this seems ok. 

:heavy_check_mark: 

> I worry a bit about races but the logic that sets up the proxy class should be safe currently

exactly - its plugged in right around proxy-class initialization which is already race condition tuned (some tests are in place)

core/src/main/java/org/jruby/javasupport/ext/JavaExtensions.java
```diff
@@ -0,0 +1,37 @@
...
+    static void put(final Ruby runtime, Class javaClass, Consumer<RubyModule> proxyClass) {
+        if (!LAZY) {
+            proxyClass.accept( org.jruby.javasupport.Java.getProxyClass(runtime, javaClass) );
+            return;
+        }
+        Object previous = runtime.getJavaExtensionDefinitions().put(javaClass, proxyClass);
+        assert previous == null;
+    }
+
+    public static void define(final Class javaClass, final RubyModule proxyClass) {
```
enebo: Seems like runtime could be a parameter here to match put().

enebo: It also seems to line up more with Iterable stuff below which also have runtime.

PR #5175, opened 2018-05-16
Always search full hierarchy for highest definition of a cvar.

### LINKED ISSUE #1554: Class variable in subclass is not overtaken by class variable in superclass
I was experimenting with class variables and I noticed that JRuby's class variables behave differently than MRI's.  If a class variable is defined in a superclass after one of its subclasses already has a class variable of the same name, MRI detects the situation and merges the two variables into one.  JRuby does not detect it, and it lets the class variable have a different value in the subclass and parent class.

Here is a simple script to demonstrate the difference:

``` ruby
class A
end

class B < A
  @@cv = :cvb  
  def cvb
    @@cv
  end
end

class A
  @@cv = :cva
end

p B.new.cvb
```

Here is shell output showing how JRuby prints `:cvb`:

```
$ jruby -v test.rb
jruby 1.7.11 (1.9.3p392) 2014-02-24 86339bb on Java HotSpot(TM) 64-Bit Server VM 1.7.0_45-b18 [Windows 8-amd64]
:cvb
```

Here is shell output showing how MRI prints a warning and then prints `:cva`:

```
$ ruby -v test.rb
ruby 2.0.0p0 (2013-02-24) [x64-mingw32]
test.rb:7: warning: class variable @@cv of B is overtaken by A
:cva
```

(Sorry if this is a duplicate; I searched through the names of all open issues but not closed ones.)

headius:
This makes the feature basically useless for anything requiring
performance, since it always has to search the entire hierarchy
for a late-assigned value further up.

Additional work using our weak subclass hierarchy might be able
to make reads slightly cheaper (giving up at first successfuly
read) by walking down the hierarchy on every write and killing
other vars, but the concurrency aspects of that make my head spin.

Fixes #1554.

### DIFF
--- a/core/src/main/java/org/jruby/RubyModule.java
+++ b/core/src/main/java/org/jruby/RubyModule.java
@@ -3713,13 +3713,14 @@ public class RubyModule extends RubyObject {
      */
     public IRubyObject setClassVar(String name, IRubyObject value) {
         RubyModule module = this;
+        RubyModule highest = this;
         do {
             if (module.hasClassVariable(name)) {
-                return module.storeClassVariable(name, value);
+                highest = module;
             }
         } while ((module = module.getSuperClass()) != null);
 
-        return storeClassVariable(name, value);
+        return highest.storeClassVariable(name, value);
     }
 
     @Deprecated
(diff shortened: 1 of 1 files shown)


kares:
... seems to fail (regress) some MRI tests

PR #5180, opened 2018-05-21
[ji] RubyString implements CharSequence

kares:
found this :paintbrush: at https://github.com/jruby/jruby/pull/3937/files#diff-2f8f0aefd8a8637512ce6c017da265b9

knew I saw a draft of this somewhere but did not find it until I checked some old PRs :)

... there's an opportunity here to have `RubyString implements CharSequence`
this is expected to fit nicely esp. will avoid conversions on logging libraries e.g. log4j2 
(so one doesn't need to do an if check: `logger.debug('foo') if logger.isDebugEnabled`)

but its a breaking change in 2 ways : 
- `RubyString#length` needs to be changed (to return int) -> not expected to be used much
- seeing a `java.lang.CharSequence` by JI will no longer conver to a java string

both seem acceptable for a major release such as 9.2 (we sure can not do this in a minor one)

### DIFF
--- a/core/src/main/java/org/jruby/RubyString.java
+++ b/core/src/main/java/org/jruby/RubyString.java
@@ -6169,10 +6169,14 @@ public class RubyString extends RubyObject implements CharSequence, EncodingCapa
 
     @Override
     public <T> T toJava(Class<T> target) {
-        if (target.isAssignableFrom(String.class)) {
+        // converting on Comparable.class due target.isAssignableFrom(String.class) compatibility (< 9.2)
+        if (target == String.class || target == Comparable.class || target == Object.class) {
             return target.cast(decodeString());
         }
-        if (target.isAssignableFrom(ByteList.class)) {
+        if (target == CharSequence.class) { // explicitly here
+            return (T) this; // used to convert to java.lang.String (< 9.2)
+        }
+        if (target == ByteList.class) {
             return target.cast(value);
         }
         if (target == Character.class || target == Character.TYPE) {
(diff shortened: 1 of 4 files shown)


kares:
other way to avoid conversion (when seeing `CharSequence` type) would be to use ByteList
... which implements it as well but than it would need some shared marking upon the field leaving string.
and it isn't encoding to a Java string properly at this point (maybe with 2.0), right?

kares:
posting some micro-benchmarking numbers : 
```
Benchmark                                        Mode  Cnt       Score      Error   Units
StringLoopBenchmark.benchLongDecodeStringLoop   thrpt    5   10745.123 ±  194.012  ops/ms
StringLoopBenchmark.benchLongJavaStringLoop     thrpt    5  123630.568 ± 2992.694  ops/ms
StringLoopBenchmark.benchLongRubyStringLoop     thrpt    5   20841.347 ± 1095.569  ops/ms
StringLoopBenchmark.benchShortDecodeStringLoop  thrpt    5   15207.295 ±  933.209  ops/ms
StringLoopBenchmark.benchShortJavaStringLoop    thrpt    5  140604.846 ± 2555.282  ops/ms
StringLoopBenchmark.benchShortRubyStringLoop    thrpt    5  103548.193 ± 3440.039  ops/ms
```
... instead of testing 'real-world' the case tries to do `String#contentEquals(CharSequence)`
decoding every time clearly has its price, on the other hand for short strings using the CharSequence interface without converting toString seems to be only about 40% worse than passing a native Java string.

Issue #4741, opened 2017-08-16
core_ext/class.rb:subclasses conflicts with ActiveRecord


atinm:
### Environment

* jruby 9.1.12.0 (2.3.3) 2017-06-15 33c6439

### Expected Behavior
After `require 'jruby/core_ext'` `ActiveRecord::Base.subclasses` should continue to return an array of the ActiveRecord::Base subclasses but instead returns the Java subclasses of ActiveRecord::Base that do not have the same methods available on them anymore.

* Example in irb:
```
    $ irb
    jruby-9.1.12.0 :001 > require 'active_record'
     => true 
    jruby-9.1.12.0 :002 > class Foo < ActiveRecord::Base
    jruby-9.1.12.0 :003?>   def bar
    jruby-9.1.12.0 :004?>     end
    jruby-9.1.12.0 :005?>   end
     => :bar 
    jruby-9.1.12.0 :006 > ActiveRecord::Base.subclasses
     => [Foo (call 'Foo.connection' to establish a connection)] 
    jruby-9.1.12.0 :007 > ActiveRecord::Base.subclasses.first.table_name
     => "foos" 
    jruby-9.1.12.0 :008 > require 'jruby/core_ext'
     => true 
    jruby-9.1.12.0 :009 > ActiveRecord::Base.subclasses
     => [#<ActiveRecord::AttributeMethods::GeneratedAttributeMethods:0x4efc25fc>] 
    jruby-9.1.12.0 :010 > ActiveRecord::Base.subclasses.first.table_name
    NoMethodError: undefined method `table_name' for #
    <ActiveRecord::AttributeMethods::GeneratedAttributeMethods:0x396e6d9>
            from (irb):10:in `<eval>'
            from org/jruby/RubyKernel.java:1000:in 'eval'
            from org/jruby/RubyKernel.java:1298:in 'loop'
            from org/jruby/RubyKernel.java:1120:in 'catch'
            from org/jruby/RubyKernel.java:1120:in 'catch'
            from /home/u/.rvm/rubies/jruby-9.1.12.0/bin/irb:13:in '<main>'
```
### Analysis
* JRuby is defining `subclasses` at:
  `https://github.com/jruby/jruby/blob/f7746d0be94f5d9f7cd27af610f1bc5dd2622794/lib/ruby/stdlib/jruby/core_ext/class.rb#L31`

* ActiveRecord is also defining `subclasses` at:
[...]

PR #5181, opened 2018-05-21
deprecate jruby/core_ext Class#subclasses

kares:
has been discussed at: https://github.com/jruby/jruby/issues/4741 ... non clashing replacement: `JRuby.subclasses_of(klass, all: false)`


kares:
@headius yeah I wasn't sure ... but than wasn't sure about `JRuby.subclasses(klass, all: false)`
... but I'm fine with it if it sounds better for you - my EN isn't native so I might not pick the best :)

PR #5183, opened 2018-05-22
[ji] support Java time APIs with Time/Date/DateTime

kares:
e.g. `Time.now.to_java(java.time.LocalDateTime)` and friends with some consumable Java APIs

... backwards compatibility maintained -> `time.to_java` still gives a `java.util.Date` (by default)

### DIFF
--- a/core/src/main/java/org/jruby/RubyTime.java
+++ b/core/src/main/java/org/jruby/RubyTime.java
@@ -381,37 +381,52 @@ public class RubyTime extends RubyObject {
         return timeClass;
     }
 
-    /** Set the 4 to 9 decimal places of the time.
+    /**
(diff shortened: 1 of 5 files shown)


core/src/main/java/org/jruby/RubyTime.java
```diff
@@ -905,21 +897,20 @@ public long microseconds() {
...
     }
 
-
     /**
      * Return the nano-seconds of this time.
      * @return nano seconds (only)
      */
-    public long getNanos() {
-        return (getTimeInMillis() % 1000) * 1_000_000 + getNSec();
+    public int getNanos() {
```
headius: Why these signature changes? They worry me like the RubyString.length change but in this case there's no alternative you can call with the old signature.

PR #5198, opened 2018-05-26
Consider encoding when deduplicating strings.

### LINKED ISSUE #5190: JRuby 9.2.0.0 string deduplication via String#-@ does not work the same as CRuby
### Environment

```
jruby 9.2.0.0 (2.5.0) 2018-05-24 81156a8 OpenJDK 64-Bit Server VM 25.144-b01 on 1.8.0_144-b01 +jit [OpenBSD-x86_64]
```

Also verified on:

```
jruby 9.2.0.0 (2.5.0) 2018-05-24 81156a8 Java HotSpot(TM) 64-Bit Server VM 25.131-b11 on 1.8.0_131-b11 +jit [mswin32-x86_64]
```

Unlikely to be operating system or java version specific.

### Expected Behavior

String deduplication via `String#-@` does not work the same as CRuby 2.5 for non-literal strings:

```
$ ruby -e "a = String.new('a'); p ((-a).object_id == (-a).object_id)"
ruby 2.5.1p57 (2018-03-29 revision 63029) [x86_64-openbsd]
true
```

### Actual Behavior

```
$ jruby -ve "a = String.new('a'); p ((-a).object_id == (-a).object_id)"
jruby 9.2.0.0 (2.5.0) 2018-05-24 81156a8 OpenJDK 64-Bit Server VM 25.144-b01 on 1.8.0_144-b01 +jit [OpenBSD-x86_64]
false
```

headius:
Fixes #5190.

CRuby's fstring cache, used for frozen string deduplication, uses
slightly different equalitye logic than the default equality for
strings. Specifically, if two strings have the same 7bit ascii
bytes, but two different ascii-compatible strings, the strings are
still considered to be equal. But for the fstring cache, you can
register the same 7-bit string with different ascii-compatible
encodings and they both live in the cache.

In JRuby, we use a standard JDK collection, ConcurrentHashMap,
that always uses the standard equals() method that works like
normal String equality as described above. We are forced to use
a wrapper, both for storage and for lookup. This patch introduces
that wrapper, and also introduces a thread-local caching mechanism
to reduce the cost of looking up deduplicated strings in the
cache.

The additional overhead of the cache is:

* The wrapper object and indirecting through it.
* Constructing a wrapper object (only when the previous lookup
  added a new wrapper or this is the first lookup).
* Accessing a previously cached wrapper via a thread-local
  (inverse of the above conditions)

In the typical case, where the requested string has already been
deduplicated, the system should eventually get to a point where
there's no new entries being added, and the cached wrapper is used
every time. There may be more overhead at startup to create the
wrappers. There may be few calls to lookup a string that do not
trigger a new entry, since most language constructs (FrozenString
in IR for example) save the result, making the cache perhaps
unnecessary.

### DIFF
--- a/core/src/main/java/org/jruby/Ruby.java
+++ b/core/src/main/java/org/jruby/Ruby.java
@@ -4716,6 +4716,20 @@ public final class Ruby implements Constantizable {
      * could cause contention under heavy concurrent load, so a reexamination
      * of this design might be warranted.
      *
+     * Because RubyString.equals does not consider encoding, and MRI's logic for deduplication does need to consider
+     * encoding, we use a wrapper object as the key. These wrappers need to be used on all get operations, so if we
+     * don't need to insert anything we reuse that wrapper the next time.
+     *
+     * The logic here reads like this:
+     *
+     * 1. If the string is not a natural String object, just freeze and return it.
+     * 2. Use the wrapper from the thread-local cache or create and set a new one.
+     * 3. Use the wrapper to look up the deduplicated string.
+     * 4. If there's a dedup in the cache, clear the wrapper for next time and return the dedup.
+     * 5. Remove the wrapper from the threadlocal to avoid reusing it, since we'll insert it.
+     * 6. Atomically set the new entry or repair the GCed entry that already exists.
+     * 7. Return the newly-deduplicated string.
+     *
      * @param string the string to freeze-dup if an equivalent does not already exist
      * @return the freeze-duped version of the string
      */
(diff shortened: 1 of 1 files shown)


kares:
LGTM

kares:
> But for the fstring cache, you can register the same 7-bit string with different ascii-compatible encodings and they both live in the cache.

just a random thought - how many single byte ascii-compatible encodings are there to manage? (a few)
so what about separate stores for encodings ... guess it wouldn't be as performant due double lookup?

PR #5205, opened 2018-06-01
review (cleanup) boot -> standard JRuby extension loading

kares:
... while looking into `jruby` loading performance and making JI native, managed to review the boot process.
there was some legacy and obvious things that we do not need to parse or can afford to lazy load instead.

the most interesting part (some cleanup noise through the commits) is an extension loading support :
`JRuby::Util.load_ext(...)` or the more official `JRuby.load_ext` (ideally after `require 'jruby'`)

avoiding `JRuby.runtime`, and JI proxy loading the bootstrap Java class, JRuby is able to boot **5+%** faster.

note, that we already delivered a ~ 5% improvement in 9.2.0 (over 9.1) this is measured compared to 9.2.0

also this is very useful internally since self-reflecting (JI loading `org.jruby` pieces) leads to a weird internal state (there were already some work-arounds on that from e.g. `subclasses` returning weird junk due `UNDEF`).

NOTE: things to consider

- do we force users extension authors to do a `require 'jruby'`
  now, its auto-loaded from kernel which was likely not intentional - should only auto `require 'java'`
- in which case we shall think about NOT auto `require 'jruby'` 
  (most exts tend to have an explicit `require 'jruby'` or use the new `load_ext` mechanism)
- need to propagate `load_ext` changes to gems we tend to boot on every run (due RGs) : 
  [~~jruby-openssl~~](https://github.com/jruby/jruby-openssl/pull/167), [~~jruby-readline~~](https://github.com/jruby/jruby-readline/pull/3), [psych](https://github.com/ruby/psych/pull/365), [jar-dependencies](https://github.com/mkristian/jar-dependencies/pull/67)
  ... and get them to propagate releases (so this PR delivers the promised loading speed improvement)


core/src/main/java/org/jruby/Ruby.java
```diff
@@ -1717,7 +1717,6 @@ private void initBuiltins() {
         addLazyBuiltin("java.rb", "java", "org.jruby.javasupport.Java");
         addLazyBuiltin("jruby.rb", "jruby", "org.jruby.ext.jruby.JRubyLibrary");
         addLazyBuiltin("jruby/util.rb", "jruby/util", "org.jruby.ext.jruby.JRubyUtilLibrary");
-        addLazyBuiltin("jruby/type.rb", "jruby/type", "org.jruby.ext.jruby.JRubyTypeLibrary");
```
kares: `org.jruby.ext.jruby.JRubyTypeLibrary` class was gone for quite some time - as explained in the commit msg

core/src/main/java/org/jruby/RubySignal.java
```diff
@@ -42,13 +42,13 @@
...
 
-    private final static SignalFacade getSignalFacade() {
+    private final static SignalFacade SIGNAL_FACADE = initSignalFacade();
+
+    private final static SignalFacade initSignalFacade() {
         try {
-            Class realFacadeClass = Class.forName("org.jruby.util.SunSignalFacade");
-            return (SignalFacade)realFacadeClass.newInstance();
-        } catch(Throwable e) {
+            return org.jruby.util.SunSignalFacade.class.newInstance();
```
kares: somehow ended up hitting this `Class.forName` spot (at least) twice while 'naively' profiling.
likely a false alarm, but I still did review the class and its signal.rb pieces ...

kares:
// ping @headius and @enebo for a review - so we can start setting up `JRuby.load_ext` in default gems

kares:
OK, cleaned up the code-style.

> requiring both 'require "java"' or 'require "jruby"' is semantically required even though they both incidentally worked.

at this point than we could remove `require 'jruby'` from booting up on all JRuby boots? 
its unfortunate ... esp. since most of it is covered by the internal `JRuby::Util` helpers
the downside is mostly parts (user exts) doing `JRuby.runtime` without the require ...

PR #5214, opened 2018-06-11
Clone proc blocks as well so they have their own frame etc.

### LINKED ISSUE #2212: Kernel#__method__ within a proc returns incorrect value after method definition
When defining more than one method within a class using a proc that returns `__method__` all methods will return the symbol representing the last defined method:

``` ruby
DEFINITION = proc { __method__ }

class C
  define_method :one, DEFINITION
  define_method :two, DEFINITION
end

o = C.new

puts o.one  #=> :two (in MRI this returns :one)
puts o.two  #=> :two
```

Casting `DEFINITION` into a code block during the method's definition seems to be working correctly though:

``` ruby
class C
  define_method :one, &DEFINITION
  define_method :two, &DEFINITION
end

o = C.new

puts o.one  #=> :one
puts o.two  #=> :two
```

A lengthier description of the problem can be found at [this post](http://stackoverflow.com/questions/26966547/is-jrubys-implementation-of-kernel-method-broken) in Stack Overflow.

headius:
Fixes #2212.

### DIFF
--- a/core/src/main/java/org/jruby/RubyModule.java
+++ b/core/src/main/java/org/jruby/RubyModule.java
@@ -2141,13 +2141,7 @@ public class RubyModule extends RubyObject {
             }
         }
 
-        block = block.cloneBlockAndFrame();
-        RubyProc proc = runtime.newProc(Block.Type.LAMBDA, block);
-
-        // a normal block passed to define_method changes to do arity checking; make it a lambda
-        proc.getBlock().type = Block.Type.LAMBDA;
-
-        newMethod = createProcMethod(name.idString(), visibility, proc);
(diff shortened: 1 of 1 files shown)


headius:
This also needs specs. The example case from #2122 also fails to super properly without this patch:

```ruby
class A
  def a
    p :a
  end
end

PROC = proc {p :proc; super}

class B < A
  define_method(:a, PROC)
  define_method(:b, PROC)
end

B.new.a
```

Which produces the following error (without my fix):

```
$ jruby blah2.rb 
:proc
NoMethodError: super: no superclass method `b' for #<B:0x27f723>
  block in blah2.rb at blah2.rb:7
             <main> at blah2.rb:14
```

Issue #5182, opened 2018-05-22
Java types should simply toString on inspect?


kares:
Java types `inspect` isn't as useful as for Ruby objects.
When using Ruby tools it usually just shows the type but no real information, e.g.

```
  2) Date"to_java java 8 types coerces to java.time.Instant
     Failure/Error: DEFAULT_FAILURE_NOTIFIER = lambda { |failure, _opts| raise failure }

       expected: #<Java::JavaTime::Instant:0x25ce9dc4>
            got: #<Java::JavaTime::Instant:0x17f62e33>

       (compared using ==)

       Diff:
       @@ -1,2 +1,2 @@
       -#<Java::JavaTime::Instant:0x25ce9dc4>
       +#<Java::JavaTime::Instant:0x17f62e33>
     # ./spec/java_integration/types/coercion_spec.rb:946:in `block in (root)'

```

its likely a breaking change -> so only to be done in a major version such as **9.2**

headius:
At one point inspect also showed fields, by reflectively digging them out. Once we removed that, the inspect output is now largely useless other than showing what type the object is.

Perhaps it should be something like `#<Java::SomeJavaPackage::SomeClass:0xcafebabe "toString output">` I think objections in the past were that the toString for a Java object could be very large, and inspect is often used for logging errors etc.

kares:
> At one point inspect also showed fields, by reflectively  digging them out. Once we removed that, the inspect output is now  largely useless other than showing what type the object is.

that would make sense for me, but only doing that if there isn't a toString override (which is tricky).
should mean common types such as `ArrayList` or Java `Date` will return a simple representation (like in Ruby). large or not, there's giant inspect in Rails e.g. when you inspect a `connection` or smt holding it.

headius:
A simple way to do this is to just define java.lang.Object.inspect to call to_s, but that will do it for all Java objects, not just the ones that have overridden to_s.

kares:
have an idea to play some 'ugly' tricks around and hook this up into the JavaProxy class, as I want it to play well with existing Java array inspect-ing while also providing a reflective default, we'll see how it goes ...

headius:
Ok I'll let you play with it then.

kares:
[...]

PR #5219, opened 2018-06-15
[ji] make inspect on Java proxies work

kares:
... `inspect` will now call `toString` on a Java proxy (unless its `java.lang.Object#toString` in which case we do the same "bare" hashy-inspect)

the motivation here is https://github.com/jruby/jruby/issues/5182 ... have seen "useless" hashy-inspect on Java proxied types too many times.

the approach here is the following: proxies use a call-site to determine whether `toString` is overriden.
the side effect is that I reverted an old 'feature' where Java (instance) methods, such as *toString*, only show-up at the lowest level in the Ruby class hierarchy - as explained on https://github.com/jruby/jruby/issues/5182#issuecomment-397203127.

there's likely a "cleaner" way -> having `inspect` alias-ed whenever a type we process has `toString`, maybe we shall eventually refactor to that but there's a much more interesting feature that I am pursuing :

current default `inspect` (for any Java class missing a custom `toString`) does only hashy-inspect without any details. would like to be able to reflectively inspect such hierarchies similar to how Ruby shows all instance variables - while I got it working its not currently usable as there's inconsistencies that I need to think through (more on that elsewhere).

in terms of compatibility there's some "breaking" change related to `java.util.Map` proxies, which seem to have been historically approached differently than the others - the internally sync state with a `Hash`. 
1. this remains the case however [due](https://github.com/jruby/jruby/issues/5182#issuecomment-397203127) noted above - there was an internal "bug" where colliding methods would prefer Hash-like ones instead of the explicit Java override (e.g. `clear` had the same effect but did not return void as is prescribed by `java.util.Map` but instead `self` from `RubyHash`).
2. the `inspect` for map proxy needed to be changed, otherwise it would be weird to have a special exception for the `Map` hierarchy from `inspect` doing `toString` on Java proxies (methods for restoring previous behavior provided).

### DIFF
--- a/core/src/main/java/org/jruby/java/proxies/ArrayJavaProxy.java
+++ b/core/src/main/java/org/jruby/java/proxies/ArrayJavaProxy.java
@@ -491,7 +491,12 @@ public final class ArrayJavaProxy extends JavaProxy {
 
     @JRubyMethod
     public RubyString inspect(ThreadContext context) {
-        return RubyString.newString(context.runtime, arrayToString());
+        return inspect(context.runtime, null); // -> toStringImpl
+    }
+
+    @Override
+    CharSequence toStringImpl(final Ruby runtime, final IRubyObject opts) {
+        return arrayToString();
     }
 
     @Override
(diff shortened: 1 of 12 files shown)


kares:
ping @headius @enebo thoughts, concerns about shipping at least what we do have atm, around here?

kares:
any thoughts, would love to ship a better JI support in tools such as rspec?

kares:
> if a Java type defines its own inspect, that should override the default one from JavaProxy, yes?

yeah, that is always the case with these proxy methods but just to be sure I have added some specs

PR #5225, opened 2018-06-21
update jruby-readline to new release

kares:
... now using `JRuby::Util.load_ext` mechanism for #5205

+ updated JLine to 2.14.6 
   contains a couple of interesting features/fixes - wonder why it lagged behind
   e.g. OSv support, HP-UX working, avoided (class) memory leaks, printing "Display all 104 possibilities? (y or n)" on a new line, ...

UPDATE: there seems to be one visible regression under Linux - printing space after irb completion

believe we can fix that and also make jruby-readline usable as a standalone gem
... should we eventually also push to RubyGems? (would be great to test out new release under Windows)


headius:
The updated jruby-readline appears to be the cause of #5387 so I think we'll need to revert this for 9.2.1 and figure out what went wrong.

Issue #5229, opened 2018-06-23
Mutating result of Module#to_s changes Module#to_s on JRuby 9.2


ivoanjo:
Hello there!

I've found this one while adding JRuby 9.2 to `concurrent-ruby`'s travis CI config, see [ħere](https://travis-ci.org/ruby-concurrency/concurrent-ruby/jobs/395831495) and  https://github.com/ruby-concurrency/concurrent-ruby/pull/733 .

### Environment

* JRuby: `jruby 9.2.0.0 (2.5.0) 2018-05-24 81156a8 OpenJDK 64-Bit Server VM 25.71-b00 on 1.8.0-internal-jenkins_2017_07_04_16_04-b00 [linux-x86_64]`
* Kernel: `Linux maruhime 4.13.0-45-generic #50-Ubuntu SMP Wed May 30 08:23:18 UTC 2018 x86_64 x86_64 x86_64 GNU/Linux`
* Distro: `Ubuntu 17.10`

### Expected Behavior

On MRI when getting the `to_s` of a class or module, the resulting string is a copy and so is safe to be mutated at will, and changing it will have no impact on the `to_s` of the class/module.

Testcase:

```ruby
puts RUBY_DESCRIPTION

class A
end

A.to_s.downcase!

puts A.to_s
```

MRI output:

```
ruby 2.5.1p57 (2018-03-29 revision 63029) [x86_64-linux]
A
```

### Actual Behavior

On JRuby 9.2, mutating the result of `to_s` wrongly changes the actual `to_s` of the class/module.

JRuby output:

```
jruby 9.2.0.0 (2.5.0) 2018-05-24 81156a8 OpenJDK 64-Bit Server VM 25.71-b00 on 1.8.0-internal-jenkins_2017_07_04_16_04-b00 [linux-x86_64]
a 
```

headius:
Yeah probably an easy fix; just clone it. Though I'd argue it should be returning an immutable string instead.

headius:
s/clone/dup

PR #5231, opened 2018-06-25
dont mutate module names

ahorek:
resolves https://github.com/jruby/jruby/issues/5229


kares:
well done, thanks Pavel

PR #5241, opened 2018-07-07
[fix] make RubyFixnum without (internal) cache working

kares:
JRuby was broken with **-Xfixnum.cache=false**, the static helpers such
as `RubyFixnum.zero` did assume the cache to be always there ...

... when running `mvn -Ptest` it didn't even boot properly :
```
java.lang.ArrayIndexOutOfBoundsException: 0
	at org.jruby.RubyFixnum.zero(RubyFixnum.java:220)
	at org.jruby.RubyComplex.createComplexClass(RubyComplex.java:111)
	at org.jruby.Ruby.initCore(Ruby.java:1480)
	at org.jruby.Ruby.bootstrap(Ruby.java:1318)
	at org.jruby.Ruby.init(Ruby.java:1217)
	at org.jruby.Ruby.newInstance(Ruby.java:362)
	at org.jruby.Ruby.newInstance(Ruby.java:350)

```


kares:
isn't really tested only run the suite manually - fix is trivial although might get forgotten at some point

kares:
back-ported to 9.1 https://github.com/jruby/jruby/commit/520e2a7d83c0df286c76a63a3441a696a6ff40bd

PR #5250, opened 2018-07-15
fix invalid date error message

ahorek:
master NoMethodError (undefined method `gsub!' for 4:Integer)
->
patch TypeError: no implicit conversion of Integer into String

https://github.com/ruby/spec/pull/610

### DIFF
--- a/lib/ruby/stdlib/date/format.rb
+++ b/lib/ruby/stdlib/date/format.rb
@@ -577,7 +577,11 @@ class Date
   def self._parse(str, comp=true)
     # Newer MRI version (written in C converts non-strings to strings
     # and also has other checks like all ascii.
-    str = str.to_str if !str.kind_of?(::String) && str.respond_to?(:to_str)
+    if str.kind_of?(::String) || str.respond_to?(:to_str)
+      str = str.to_str if str.respond_to?(:to_str)
+    else
+      raise TypeError, "no implicit conversion of #{str.class.name} into String"
+    end
     str = str.dup
 
     e = Format::Bag.new


kares:
shouldnt this go into parse instead of the low level _parse? did not check MRI ...

ahorek:
I think we should always call .to_str because for instance ActiveSupport::SafeBuffer doesn't work right with regex see https://github.com/rails/rails/issues/1555

ahorek:
right, I will fix it and add more specs

ahorek:
```
Date._parse ActiveSupport::SafeBuffer.new '2011-01-01'
```
doesn't work without the explicit conversion into a String.

is it safe to convert it? MRI uses StringValue()
https://github.com/ruby/ruby/blob/576b245ffae913aec11ae321fe7ddc9c2b688f67/ext/date/date_core.c#L4297

lib/ruby/stdlib/date.rb
```diff
@@ -482,6 +482,9 @@ def self.strptime(str='-4712-01-01', fmt='%F', sg=ITALY)
   #
   # +sg+ specifies the Day of Calendar Reform.
   def self.parse(str='-4712-01-01', comp=true, sg=ITALY)
+    unless str.kind_of?(::String) || str.respond_to?(:to_str)
```
kares: could we just move this to `_parse` ? 
and than it wouldn't need to exist on 3 places + MRI seems to validate there as well:
```
2.5.1 :003 > Date._parse 111
        3: from /opt/local/rvm/rubies/ruby-2.5.1/bin/irb:11:in `<main>'
        2: from (irb):3
        1: from (irb):3:in `_parse'
TypeError (no implicit conversion of Integer into String)
```
... and since this would eventually be moved to native like they did it will be simpler for anyone ending up doing it.

lib/ruby/stdlib/date/format.rb
```diff
@@ -577,7 +577,11 @@ def self._parse_ddd(str, e) # :nodoc:
   def self._parse(str, comp=true)
     # Newer MRI version (written in C converts non-strings to strings
     # and also has other checks like all ascii.
-    str = str.to_str if !str.kind_of?(::String) && str.respond_to?(:to_str)
+    if str.kind_of?(::String) || str.respond_to?(:to_str)
+      str = str.to_str if str.respond_to?(:to_str)
```
ahorek: the second check is for this case which seems very unlikely to me. Do you think it's worth to write a test for it?
```
class Sub < String
  undef_method :to_str
end
Date._parse Sub.new('20160505')
```

ahorek:
_parse is used a lot internally on places where we know that the data type is already ok, but the real impact is probably not measurable

lib/ruby/stdlib/date/format.rb
kares: not really - there wasn't a second check.

kares:
> _parse is used a lot internally on places where we know that the data type is already ok,

ok.
still, makes me wonder `str.to_str` should raise proper `TypeError`, maybe _parse is worth a C read
... feels slightly "wrong" that we have to mess around String checks - `to_str` should do the work for us

ahorek:
but #to_str isn't defined on Integer

other methods like #_iso8601 fails on "=~" with the same error (TypeError) which is fine, but #_parse calls #gsub! first (NoMethodError) and it should be also TypeError

kares:
pushed to master as https://github.com/jruby/jruby/commit/a6dad4b29ac656ce38212c4488f76492bc51d41c

ahorek:
thanks @kares 
but after a6dad4b this case doesn't work anymore

```
require 'active_support'
require 'date'

str = ActiveSupport::SafeBuffer.new('2011-05-07')
```

Date.parse str
```
mri
=> #<Date: 2011-05-07 ((2455689j,0s,0n),+0s,2299161j)>
```

```
jruby
        7: from org/jruby/RubyKernel.java:1178:in `catch'
        6: from org/jruby/RubyKernel.java:1178:in `catch'
        5: from org/jruby/RubyKernel.java:1412:in `loop'
        4: from org/jruby/RubyKernel.java:1040:in `eval'
        3: from (irb):10:in `evaluate'
        2: from /jruby/lib/ruby/stdlib/date.rb:486:in `parse'
        1: from /jruby/lib/ruby/stdlib/date.rb:444:in `new_by_frags'
[1mArgumentError ([4minvalid date[0;1m)[m
```

Date._parse str
```
mri
=> {:year=>2011, :mon=>5, :mday=>7}
```
```
jruby
=> {:mon=>0}
```

I'm not sure if it should work or not, see https://github.com/rails/rails/issues/1555 . According to the rails devs we should never use regex directly on a SafeBuffer... but MRI handles this case, could you recheck?

kares:
guess, they're pushing it to the limits with their `SafeBuffer` ... anyway seems like $1 globals are to blame.
have actually wondered if we can avoid those or really need to fix it to work for sub-strings properly ...

ahorek:
There's no need, I don't think it could cause a real problem and even if so, it's easily fixable by calling .to_str explicitly. Thanks for your comments.

Issue #5252, opened 2018-07-16
can't use Date.iso8601 with a string subclass


ahorek:
### Environment

jruby 9.2.1.0-SNAPSHOT (2.5.0) 2018-07-15 8833646 Java HotSpot(TM) 64-Bit Server VM 10.0.1+10 on 10.0.1+10 +jit [mswin32-x86_64]

### Expected Behavior
```
require 'date'
class StringSubclass < String; end
Date.iso8601(StringSubclass.new("-4712-01-01"))
```
 #<Date: -4712-01-01 ((0j,0s,0n),+0s,2299161j)>

### Actual Behavior
never ends

spec https://github.com/ruby/spec/pull/612

ahorek:
the actual problem is in the type converter, it's broken since
https://github.com/jruby/jruby/commit/8c3a6691125db340d9833aeb1500490b2feddd25#diff-876b1ff02a1d966941bc84c184dbcc68

MRI checks subclasses
https://github.com/ruby/ruby/blob/38e05ff3e194268fd2f38ac7c9530298e464f07b/object.c#L3134

kares:
nice find, thanks @ahorek ... somehow naively assumed the internal ClassIndex is retained for sub-class

kares:
this likely means https://github.com/jruby/jruby/pull/5250 is a no-op (no changes required if this is fixed), right?

ahorek:
btw: only string class inheritance is relevant, we don't have to check other cases. I don't know if there's a better way how to do this without an if/else statement. Please advice.

https://github.com/jruby/jruby/pull/5250 is still relevant, the original problem was a wrong exception class, later I found this.

kares:
okay, if its only `String` than we're simply do one more if check as you tried - will try to tweak your commit

PR #5253, opened 2018-07-16
prevent an infinite loop, check string subclasses

ahorek:
fixes https://github.com/jruby/jruby/issues/5252


kares:
thanks, but this seems as an ugly hack. will look into the change made as you explained in #5252

kares:
ended up doing ~ the same: https://github.com/jruby/jruby/commit/5470f056fa635b50945bc0345e7779b27ba2e517
... the scary piece here was that we relied on user types reporting `CLASS` index

PR #5262, opened 2018-07-26
Proc source location gives wrong result

jmalves:
### Environment
* Ruby version: jruby 9.2.0.0 (2.5.0) 2018-05-24 81156a8 Java HotSpot(TM) 64-Bit Server VM 25.152-b16 on 1.8.0_152-b16 +jit [darwin-x86_64]
* OS/platform: macOS 10.13.5

### Expected Behavior

On ruby 2.5.0 and jruby-9.1.17.0 the following script:

```ruby
expected = __LINE__ + 1
proc = Proc.new do


end

p "#{expected} == #{proc.source_location[1]}"
```

Prints: `2 == 2`

### Actual Behavior

On jruby-9.2.0.0 it prints `2 == 5`

### Investigation

I found this bug by noticing that `rspec` with line filters (e.g. `rspec test_file.rb:20`) was not working.

I traced this back to https://github.com/jruby/jruby/commit/ca254718c8cab05f97ebb527f3f370f375695104

I am doubtful that the fix I proposed in this PR is adequate however it was quite fun to look around this. 

Tested with fast specs, mri's _test_lambda_ and _test_proc_. No spec failures were added but also none were fixed :(

### DIFF
--- a/core/src/main/java/org/jruby/parser/RubyParser.java
+++ b/core/src/main/java/org/jruby/parser/RubyParser.java
@@ -4469,9 +4469,8 @@ states[443] = new ParserState() {
 };
 states[444] = new ParserState() {
   @Override public Object execute(ParserSupport support, RubyLexer lexer, Object yyVal, Object[] yyVals, int yyTop) {
-                    /* FIXME: probably need to correct location here*/
-                    yyVal = new IterNode(lexer.getPosition(), ((ArgsNode)yyVals[-1+yyTop]), ((Node)yyVals[0+yyTop]), support.getCurrentScope());
-                     support.popCurrentScope();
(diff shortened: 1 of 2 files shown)


ahorek:
there're some failures
```
 1) Error:
     [exec] TestSyntax#test_brace_block_after_blockcall_colon_no_arg:
     [exec] Java::JavaLang::NullPointerException: 
     [exec]     org.jruby.parser.RubyParser$408.execute(RubyParser.java:4472)
     [exec]     org.jruby.parser.RubyParser.yyparse(RubyParser.java:1695)
...
```

jmalves:
The idea behind the PR was to use $-1 which I guessed it should be either `{` or `do` but it seems it does not work. Your solution of adding another block makes sense and is also clearer since you do not have to know what $-1 would be.

Thx for taking a look!

enebo:
@jmalves yeah I apologize for not noticing the '-'.   I think this is a difference of jay perhaps and a more modern LALR-yacc like bison.  I believe your PR would work if it wasn't jay.

Issue #5075, opened 2018-03-02
File.utime failing with JRuby 9.1.16.0


bastien:
### Environment

* JRuby 9.1.16.0 (2.3.3) 2018-02-21
* Java version "1.8.0_161"
* Darwin Kernel Version 16.7.0: Thu Jan 11 22:59:40 PST 2018; root:xnu-3789.73.8~1/RELEASE_X86_64 x86_64

### Expected Behavior

With jruby-9.1.15.0
```ruby
irb > File.utime(Time.at(1420099200), Time.at(1420099200), "foo.txt")
=> 1
```

### Actual Behavior

With jruby-9.1.16.0
```ruby
irb > File.utime(Time.at(1420099200), Time.at(1420099200), "foo.txt")
NotImplementedError: utimensat unsupported or native support failed to load; see http://wiki.jruby.org/Native-Libraries
	from org/jruby/RubyFile.java:1162:in `utime'
	from (irb):2:in `<eval>'
	from org/jruby/RubyKernel.java:995:in `eval'
	from org/jruby/RubyKernel.java:1316:in `loop'
	from org/jruby/RubyKernel.java:1138:in `catch'
	from org/jruby/RubyKernel.java:1138:in `catch'
	from /Users/bvaucher/.rbenv/versions/jruby-9.1.16.0/bin/irb:13:in `<main>'
```

headius:
Hmm...seems to work ok on master. I'll poke around.

headius:
Ok, I've confirmed this is also working on JRuby 9.1 HEAD (9.1.17.0) so I'm goin gto need more information here.

```
[] ~/projects/jruby $ jruby -e 'p File.utime(Time.at(1420099200), Time.at(1420099200), "foo.txt")'
1

[] ~/projects/jruby $ ls -la foo.txt 
-rw-r--r--  1 headius  staff  0 Jan  1  2015 foo.txt
```

Can you try running with `-Xnative.verbose` and provide that output?

Also can you confirm what version of OS X you are running on?

bastien:
```
jruby -Xnative.verbose -e 'p File.utime(Time.at(1420099200), Time.at(1420099200), "foo.txt")'
Successfully loaded native POSIX impl.
NotImplementedError: utimensat unsupported or native support failed to load; see http://wiki.jruby.org/Native-Libraries
   utime at org/jruby/RubyFile.java:1162
  <main> at -e:1

ls -la foo.txt
[...]

PR #5263, opened 2018-07-27
Fix parameters to utimes() fallback call (fixes #5261)

### LINKED ISSUE #5261: File.utime failing with Errno::EINVAL on JRuby 9.1.17.0
### Environment
* JRuby version: 9.1.17.0
* OS/platform: `4.4.20-0.1.fm.237.49.326.metal1.x86_64 #1 SMP Wed Sep 21 11:35:01 UTC 2016 x86_64 x86_64 x86_64 GNU/Linux`
* Running inside a Docker container

### Expected Behavior
This should succeed:
```ruby
now = Time.now
File.utime(now, now, "foo.txt")
```

### Actual Behavior
This used to work in 9.1.15.0, but now that we've upgraded to 9.1.17.0 we are getting the following error:
```
Errno::EINVAL: Invalid argument - Invalid argument
org/jruby/RubyFile.java:1177:in `utime' 
```

### Investigation
I tried to troubleshoot this a bit, and I believe I know what's happening. I believe this is related to the patch that was added in #5075. I'm pretty sure what's happening is that the [call to utimensat(...) here](https://github.com/jruby/jruby/blob/9.1.17.0/core/src/main/java/org/jruby/RubyFile.java#L1166) is failing with a `NotImplementedError` as described in #5075, and then it is falling back to calling `utimes` in the [catch block here](https://github.com/jruby/jruby/blob/9.1.17.0/core/src/main/java/org/jruby/RubyFile.java#L1170). However, that line is passing in the same exact `atimespec` and `mtimespec` values from the first call and those values are basically a two-element array containing `[seconds, nanoseconds]`, but `utimes` is expecting `[seconds, microseconds]`. Thus, it fails with an `Errno::EINVAL: Invalid argument - Invalid argument` whenever the nanoseconds value is >= 1,000,000 (which would be an invalid value for microseconds). I'm guessing whoever tested the patch from #5075 must be running on a platform where `utimes` is more tolerant of invalid microsecond values.

To validate this, I ran the above code in a loop 100,000 times and it actually succeeded 134 times (0.134%). You'd expect the nanoseconds value to be < 1,000,000 about 0.1% of the time. I also ran against 9.1.16.0 and got the same exact `NotImplementedError: utimensat unsupported or native support failed to load` error referenced in #5075.

It seems like a pretty simple fix to just divide the nanoseconds value by 1,000 inside the catch block before the call to `utimes`. I will follow up later with a pull request to fix, but that will take me some more time to build and test. I wanted to submit the issue now in case anyone else is having the same problem, and to get any feedback on if my investigation and proposed fix makes sense.

I should also mention that despite the title, I'm assuming this is still an issue with JRuby 9.2.0.0 since the relevant code hasn't changed (we just haven't had a chance to upgrade to 9.2.0.0 yet)

danielford:
This code was rewritten in #5075 to call `utimensat()` which takes nanosecond precision, but then falls back to calling `utimes()` which takes microsecond precision. Passing in microsecond values greater than 1,000,000 caused `utimes()` to fail with `Errno::EINVAL`. See #5261 for more details.

This fixes the issue by converting to microseconds before calling `utimes()`.

### DIFF
--- a/core/src/main/java/org/jruby/RubyFile.java
+++ b/core/src/main/java/org/jruby/RubyFile.java
@@ -1263,6 +1263,9 @@ public class RubyFile extends RubyIO implements EncodingCapable {
             } catch (NotImplementedError re) {
                 // fall back on utimes
                 result = runtime.getPosix().utimes(fileToTouch.getAbsolutePath(), atimespec, mtimespec);
+                long[] atimeval = convertTimespecToTimeval(atimespec);
+                long[] mtimeval = convertTimespecToTimeval(mtimespec);
+                result = runtime.getPosix().utimes(fileToTouch.getAbsolutePath(), atimeval, mtimeval);
             }
 
             if (result == -1) {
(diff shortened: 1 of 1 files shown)


headius:
Looks good, thank you!

PR #5274, opened 2018-08-03
Update specs

eregon:


eregon:
Over 500 of the new specs from rubysl-socket are failing, see https://api.travis-ci.org/v3/job/411913097/log.txt

eregon:
Also, it seems `TCPServer.new('::1', 0)` throws `SocketError` when there is no IPv6, instead of `Errno::EADDRNOTAVAIL` like MRI does:
https://api.travis-ci.org/v3/job/411922756/log.txt

eregon:
The specs pass now, so I will merge this PR.

PR #5276, opened 2018-08-07
Fixes underlying logic of slice_when / chunk_while to handle 1 element collections

### LINKED ISSUE #5275: `slice_when` and `chunk_when` don't work with an 1-element enumerable
### Environment

jruby -v: `jruby 9.2.0.0 (2.5.0) 2018-05-24 81156a8 Java HotSpot(TM) 64-Bit Server VM 25.102-b14 on 1.8.0_102-b14 +jit [darwin-x86_64]`

uname -a: `Darwin ..... 17.7.0 Darwin Kernel Version 17.7.0: Thu Jun 21 22:53:14 PDT 2018; root:xnu-4570.71.2~1/RELEASE_X86_64 x86_64`

### Expected Behavior

I believe this is related to (and an acknowledged caveat of) #4410; I have an Enumerator of an uncertain size, but sometimes contains only a single element. In MRI ruby, when I call `slice_when` (or `chunk_when`), I get:

```ruby
def a
  return self.enum_for(:a) unless block_given?
  yield 1
end

a.slice_when { true }.to_a # => [[1]]
```

### Actual Behavior

In JRuby 9.2.0.0, the single element in not returned:

```ruby
def a
  return self.enum_for(:a) unless block_given?
  yield 1
end

a.slice_when { true }.to_a # => [[]]
```

mejackreed:
Enumerables with 1 element

Fixes #5275. Instead of checking for size 1 at the beginning, this change sets a
flag that never gets modified if `each_cons(2)` does not get called.
This happens in the case of 1 element Enumerables. The previous
behavior was fine for Enumerables that respond to `size` but fails
otherwise.

### DIFF
--- a/core/src/main/ruby/jruby/kernel/enumerable.rb
+++ b/core/src/main/ruby/jruby/kernel/enumerable.rb
@@ -80,30 +80,29 @@ module Enumerable
   end
 
   def __slicey_chunky(invert, enum, block)
-    if respond_to?(:size) && size == 1
-      each {|x| enum.yield [x]}
-    else
-      ary = nil
-      last_after = nil
-      each_cons(2) do |before, after|
-        last_after = after
-        match = block.call before, after
+    ary = nil
+    last_after = nil
+    element_present = false
+    each_cons(2) do |before, after|
+      element_present = true
+      last_after = after
+      match = block.call before, after
 
-        ary ||= []
-        if invert ? !match : match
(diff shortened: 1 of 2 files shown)


enebo:
@mejackreed sorry we did not merge this a lot sooner.

mejackreed:
No problem. Thanks for accepting the PR!

PR #5296, opened 2018-08-29
Fix some failed tests for Enumerable#all?, #any?, #none? and #one?

nomadium:
Hi folks,

I noticed that some tests related to a feature that I introduced months ago were tagged as failing so I gave a shot to fix them.

Those tests should be passing now.

As usual, any feedback is more than welcome.


kares:
along with https://github.com/jruby/jruby/commit/5572b29364b7b5856aa4282e114a1258886b5356 should be good to go. any reason it wasn't merged @enebo ?

PR #5298, opened 2018-08-30
Fix remaining failing tests with Enumerable#all? when given a pattern argument

nomadium:
Hi folks,

This change should fix more failing tests with Enumerable#all? when given a pattern argument.

The approach is similar to what is implemented for Enumerable#any? [1]

1. https://github.com/jruby/jruby/blob/master/core/src/main/java/org/jruby/RubyArray.java#L4435-L4440


kares:
we shall (alias) `Nomadium = Enumerable` for you hard work!

PR #5303, opened 2018-08-31
Fix some failed Enumerable#find_index specs

nomadium:
Hi folks,

This change should fix the remaining failing specs for Enumerable#find_index.

This change removes the assumption that #each would yield passing always a single object. It could yield without passing parameters, passing one or more objects or yield an explicit array (empty or not).


kares:
nice find!

PR #5306, opened 2018-09-02
Fix Hash#initialize spec tagged as failed

nomadium:
Hi folks,

This change should fix a failed spec for Hash#initialize.

Hash#initialize can be used to reset the default value.

Cheers,


kares:
nice!

PR #5308, opened 2018-09-10
Include output of `inspect` in ConcurrencyError

JasonLunn:
Adding `inspect` to the exception message when a `ConcurrencyError` is raised to make identifying the offending structure straightforward.

### DIFF
--- a/core/src/main/java/org/jruby/RubyHash.java
+++ b/core/src/main/java/org/jruby/RubyHash.java
@@ -2600,7 +2600,9 @@ public class RubyHash extends RubyObject implements Map {
 
     private final RaiseException concurrentModification() {
         return getRuntime().newConcurrencyError(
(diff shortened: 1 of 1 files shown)


kares:
not sure this accomplishes that much ... the Hash might be in an inconsistent state.
if so let's just do a simple ':' instead of '. inspect:'

JasonLunn:
Bump

kares:
looking at this again I am not sure about the balance of usability <-> determinism.
if a concurrency error happens -> doing `inspect` another concurrency error is fairly likely to happen as we iterate over the hash ... yet again.
than I am not sure what happens but we might end up blowing the stack. it's likely that at some point we get an inspected result but the printed content might no longer be accurate/useful?

is this too far fetched thinking on my end for a simple change? // cc @enebo @headius

PR #5317, opened 2018-09-19
Allocate method on NilClass should raise TypeError

nomadium:
I noticed an small difference between MRI and JRuby specs. According to them, to try to allocate space for a new instance of NilClass should raise a TypeError instead of returning nil.


enebo:
OPTIMISM at it's best...so sure of this one I did not let travis complete...thanks for the PR!

PR #5316, opened 2018-09-19
Update jnr-constants for fixes after the updated generation logic

headius:
Thanks to @ahorek for help getting these regenerated.

PR #5321, opened 2018-09-21
ProcessBuilder#environment

ahorek:
this is a question.

Before this was merged https://github.com/jruby/jruby/pull/5316
shelling on windows was broken

```
irb(main):001:0> `dir`
ArgumentError: wrong number of arguments (1 for 0)
                          ` at uri:classloader:/jruby/kernel/jruby/process_manager.rb:28
                          ` at uri:classloader:/jruby/kernel/jruby/process_manager.rb:50
```

now it works fine, but according to the javadoc ProcessBuilder#environment don't accept an argument
https://docs.oracle.com/javase/7/docs/api/java/lang/ProcessBuilder.html#environment()

am I missing something?

```
java version "10.0.2" 2018-07-17
Java(TM) SE Runtime Environment 18.3 (build 10.0.2+13)
Java HotSpot(TM) 64-Bit Server VM 18.3 (build 10.0.2+13, mixed mode)
```

### DIFF
--- a/core/src/main/ruby/jruby/kernel/jruby/process_manager.rb
+++ b/core/src/main/ruby/jruby/kernel/jruby/process_manager.rb
@@ -25,7 +25,7 @@ module JRuby
       pb = ProcessBuilder.new(config.exec_args)
       pb.redirect_input(Redirect::INHERIT)
       pb.redirect_error(Redirect::INHERIT)
-      pb.environment(ShellLauncher.get_current_env(JRuby.runtime))
+      pb.environment
       cwd = JRuby.runtime.current_directory
       cwd = cwd.start_with?('uri:classloader:/') ? ENV_JAVA['user.dir'] : cwd
       pb.directory(JFile.new(cwd))


enebo:
This is not Java 10 but 9.2.0.0 could use ``:

https://gist.github.com/enebo/b3cad2c1479c02b38a1c5b073c3720af

I admit I am confused about pb.environment(env), but I wonder what the difference is here?  As I see it working in 9.2.0.0 on win10 and Java 8.

kares:
the change makes sense, although seems weird no windows users complained so far.
that line has been in since: https://github.com/jruby/jruby/blob/master/core/src/main/ruby/jruby/kernel.rb#L16
... as seen from the commit, used to only load on Java 7 which is **no longer** the case on 9.1/9.2

explainable if @enebo sees a warning in $VERBOSE mode: https://github.com/jruby/jruby/blob/master/core/src/main/ruby/jruby/kernel.rb#L16 ... why would the block fail on 10, unable to load ProcessBuilder?

ahorek:
it was revealed by a different bug that was already fixed. Right now this code isn't loaded so shell works fine, 
Maybe the code isn't necessary anymore and it should be removed instead?

kares:
> it was revealed by a different bug that was already fixed.

ah, makes sense than - yeah Win should not run this code - was meant for GAE (not sure it still matters)
... still it might make sense for limited (constrained) environments without ProcessBuilder functionality

PR #5322, opened 2018-09-24
Use IndyBinder instead of generated invokers.

headius:
This is an experimental PR to try (again) to use only invokedynamic method handles rather than our generated classes.

In my quick tests, this binder does not appear to have any difference in performance from running with the normal generated invokers.

The logic works like this:

* Like AnnotationBinder, a class is generated for each Java-based Ruby class with code to wire up all methods. However IndyBinder generates bytecode, so it can statically reference  the handles pointing at each Java-based method on the core class.
* Instead of generating an "INVOKER" class for each method binding, IndyBinder constructs HandleMethod with deferred Callable<MethodHandle> for each slot, which when executed will actually instantiate the related method handle.

There are a couple reasons why this change makes sense:

* The invokers constitute a few thousand classes, perhaps a thousand of which get loaded on every startup. The indy version will load no classes and only generate the handles for method arities actually called.
* Those thousands of classes must be in the jar file. Using the indy binding reduces the size of the main JRuby jar by over 2MB (based on current master, 2.07MB).

I am looking to get this running in Travis and get others familiar with the project to try it out and see how their startup time is affected.  I'm also interested in the runtime performance aspects, but mostly  because I have not audited the indy-based binding logic to see if it is properly taking advantage of the available handles.

### DIFF
--- a/core/pom.rb
+++ b/core/pom.rb
@@ -142,19 +142,6 @@ project 'JRuby Core' do
                      :id => 'add-populators',
                      'sources' => [ '${anno.sources}' ] )
     end
-
-    plugin 'org.codehaus.mojo:exec-maven-plugin' do
-      execute_goals( 'exec',
-                     :id => 'invoker-generator',
-                     'arguments' => [ '-Djruby.bytecode.version=${base.java.version}',
-                                      '-classpath',
-                                      xml( '<classpath/>' ),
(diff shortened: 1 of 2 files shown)


headius:
@kares Work in progress, and my early experiments indicated that it was slower than the generated binders. That does not seem to be the case on Java 8+ anymore (for me, on MacOS).

PR #5323, opened 2018-09-24
Fix tagged as failed spec for Enumerable#slice_when

nomadium:
In order to iterate correctly over Enumerable elements, #each_entry method should be used, because that method will convert multiple values from yield to an array, if required.

This is a fix similar to https://github.com/jruby/jruby/pull/5302.


nomadium:
This is the test that should be passing now: https://github.com/jruby/jruby/blob/master/spec/ruby/core/enumerable/slice_when_spec.rb#L46-L53

PR #5326, opened 2018-09-25
Try running MRI core tests in parallel (four-way).

headius:
Testing the improvement and stability of "-j  4" parallel execution of the MRI tests.

A large number of these tests spawn and wait for subprocesses to boot; if those can run in parallel it may speed up the overall run time substantially. However I got a few failures locally on OS X and I'm unsure if they are supposed to pass. I will re-run locally without parallelism to see if MacOS has decayed.


headius:
We are already running `-j 2` on master through other changes. 👍

PR #5328, opened 2018-09-26
Fix ci deploy

headius:
Numeric changes here to improve the overall time of our CI builds.

* Using -j for MRI suites to parallelize tests.
* Moved MRI test_syntax to a separate run, since it appears to be the slowest test file.
* Eliminated pre-build maven install; it does not appear to be necessary anymore.
* Move deploy of snapshot to a post-test build stage so it does not fire for any of the test builds.
* Reduce git clone to depth 1.

Some changes were necessary but unrelated to the work above:

* Removed allow_failures group of tests because of difficulties getting them to  match with the new "jobs" matrix.

With these changes in place plus @enebo's tweaks to use --dev, the most recent build on this branch took 22 minutes to complete, versus 45 minute for a build before any of this work began. The most recent master build, with @enebo's changes, took 25 minutes.

There's still room to improve.

* More tests could be moved to the "syntax" MRI suite, since that is still shorter than the long version.
* Running "int" and "fullint" for MRI may be excessive. JIT runs off the same IR as "fullint", even if it does not test the interpreter, and "int" will progress stdlib methods to "fullint" as a result of lazy IR optimization.
* Obviously any improvement to startup time would continue to reduce the longest test runs. Many tiny  scripts are launched in test_syntax, for example.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -30,11 +30,13 @@ os:
   - linux
 
 env:
-#  global:
-#    - JAVA_OPTS="-XX:+TieredCompilation -XX:TieredStopAtLevel=1 -Xms48M -Xmx640M -XX:InitialCodeCacheSize=40M -XX:ReservedCodeCacheSize=120M -Djava.security.egd=file:/dev/./urandom"
-#    - MALLOC_ARENA_MAX=2
+  global:
+    - JAVA_OPTS="-XX:+TieredCompilation -XX:TieredStopAtLevel=1 -Xms48M -Xmx640M -XX:InitialCodeCacheSize=40M -XX:ReservedCodeCacheSize=120M -Djava.security.egd=file:/dev/./urandom"
+    - MALLOC_ARENA_MAX=2
 
 jobs:
+  fast_finish: true
+
   include:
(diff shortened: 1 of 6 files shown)


headius:
Additional changes: reordered test suites so the longer ones come at the top; they will start earlier and run while the shorter suites complete. This does mean fewer short suites will run in parallel but given that the MRI suites are nearly as long as the entire CI run, I think the new orientation is better.

I reverted a last-minute bump from j2 to j3 for the remaining core MRI tests. It appeared to slow things down.

headius:
We could also just remove the deploy stage, which would cut another couple minutes off the build. We do not link to the sonatype snapshot repos from anywhere, and have never officially used these builds as our snapshots. The CI builds run periodically on Cloudbees serve as our snapshots at present.

@mkristian I don't think anything in our builds is using the snapshot deploys either, right?

mkristian:
@headius `install` is all the build needs !

kares:
LGTM

headius:
spec:regression remains unstable but I will examine that separately. This PR looks ok.

PR #5335, opened 2018-09-29
Fix mvnw jar file path for Cygwin

fenrir-naru:
mvnw failed under Cygwin with Windows JDK with the following error message.
Error: Could not find or load main class org.apache.maven.wrapper.MavenWrapperMain
This is because maven-wrapper.jar could not be found, and this PR fixes it.


fenrir-naru:
@mkristian Thank you for your comment. I checked, and this PR seems to be equivalent to takari/maven-wrapper#83 . I hope this PR will be closed when the upstream PR is merged.

PR #5337, opened 2018-09-30
index error instead of argument error

ahorek:


ahorek:
there's a test for this behavior, but it still fails for a different reason

test_aset_error
https://github.com/jruby/jruby/blob/e4aee7dd33e1edb4388c584aab87c3e51b42794f/test/mri/excludes/TestArray.rb#L1
https://github.com/jruby/jruby/blob/4e8bb2666a7257f0f5986800f96bb88efdd6acbd/test/mri/ruby/test_array.rb#L2113

should I duplicate it and test this separately in a jruby's test specs?

kares:
@ahorek that is very helpful to know, thanks.
fine with leaving it as is or you could try adding a spec/ruby for it if it makes sense.
think this can get (eventually) merged as is.

PR #5338, opened 2018-09-30
fix shuffle_random

ahorek:


core/src/main/java/org/jruby/RubyRandom.java
```diff
@@ -714,6 +718,11 @@ public static long randomLongLimited(ThreadContext context, IRubyObject obj, lon
         RandomType rnd = tryGetRandomType(context, obj);
 
         if (rnd == null) {
+            DynamicMethod method = obj.getMetaClass().searchMethod("rand");
```
ahorek: basically I need a rb_funcallv_public here, but I didn't find a reference for it in jruby, Probably this could be simplified? What do you think @kares ?
https://github.com/ruby/ruby/commit/74ba0cfc622925c2527b3d281802a3d1ab86363d#diff-a962031992b7b5934db41bf925b3487eL1021

kares: yeah, MRI isn't modelled 100% ... 
think if you do some form of bare "invoke" it should check (public) visibility - seems like that was the point.
maybe try changing `Helpers.invoke` bellow to a call doing the check: `Helpers.invokeFrom` ...

ahorek:
Ok, I've introduced invokePublic (mri uses it only in a few places)

it's much cleaner now. ping @kares?

headius:
@kares I went ahead with this because I want to change the invokePublic to use a call site and the rest looks fine.

headius:
@kares Nevermind that last bit...I see invokePublic has somewhat different behavior than a non-functional call site. Leaving it in place as is 👍

PR #5342, opened 2018-10-02
return early if one of the arrays is empty

ahorek:
```
require 'benchmark/ips'
Benchmark.ips do |x|
  x.report('&', "[] & Array.new(200)")
end
```

before
```
  &    177.163k (± 5.2%) i/s -    886.145k in   5.015640s
```
after
```
  &      2.941M (± 5.4%) i/s -     14.643M in   4.996883s
```


kares:
najs!

PR #5343, opened 2018-10-03
"==" can be redefined

ahorek:
```
     ARY = Array.new(100) { |i| i }
      class Integer
        def == other
          ARY.replace([])
          self.equal?(other)
        end
      end
      ARY.count(42) == 0

=> true
```

unfortunatelly isMethodBuiltin is very expensive
```
require 'benchmark/ips'
a = Array.new(200, 6)

Benchmark.ips do |x|
  x.report('include') { a.include?(5) }
end
```

master
```
 1.044M (± 7.3%) i/s -      5.211M in   5.023214s
```
this patch
```
 462.594k (± 3.2%) i/s -      2.319M in   5.019374s
```
mri
```
 681.590k (± 9.3%) i/s -      3.389M in   5.020591s
```

### DIFF
--- a/core/src/main/java/org/jruby/RubyObject.java
+++ b/core/src/main/java/org/jruby/RubyObject.java
@@ -387,10 +387,12 @@ public class RubyObject extends RubyBasicObject {
     }
 
     private static boolean fastNumEqualInternal(final ThreadContext context, final IRubyObject a, final IRubyObject b) {
-        if (a instanceof RubyFixnum) {
-            if (b instanceof RubyFixnum) return ((RubyFixnum) a).fastEqual((RubyFixnum) b);
-        } else if (a instanceof RubyFloat) {
-            if (b instanceof RubyFloat) return ((RubyFloat) a).fastEqual((RubyFloat) b);
+        if (a instanceof RubyFixnum && b instanceof RubyFixnum &&
+            !context.runtime.getFixnum().isRefinement() && context.runtime.getFixnum().isMethodBuiltin("==")) {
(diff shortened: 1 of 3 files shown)


kares:
hey, is this a real use-case you run into or just trying to fix the test?

maybe this can get redesigned - assuming objects of same type a call-site might help.
in which case, ideally, there should be a single isBuiltin check against the (cached) site ...

ahorek:
I just want to increase the test coverage. This fix is a corner case,  but I think there're other places where redefinition should be considered and is_buildin method shouldn't be that slow. I'll try something with indy.

This shouldn't be merged until the performance problem is solved even if the behaviour is correct.

ahorek:
I'm not sure if I'm using it correctly. Now it's better, but still 2 times slower than before. Any idea how to improve it?

kares:
looks reasonably good, do you happen to know which part is slowest, maybe the `isRefinement` check?

enebo:
@ahorek another idea if it is isBuiltin taking the time would be to save builtin EQ method somewhere similar to what we do in context.sites and then compare the method in callsite against that saved version.  That would simplify this to a single equality check; but also would require you do save it at some point in time during bootstrapping of Ruby.

enebo:
in 2022 There was some very similar code landed for this method (it is missing the !isRefinement part).  Closing just due to age and sorry we did not come to consensus on this.

ahorek:
you're right, it has already been fixed by https://github.com/jruby/jruby/commit/505dea7dc141d9d95b899bbb95d737441b16c2b9 thanks for noticing!

PR #5345, opened 2018-10-03
bump asm version + remove asm-analysis

kares:
... rest of asm jars seems to be used, for now

@enebo do you need asm-analysis.jar around?


enebo:
@kares I don't think so.  @headius was this a left over from 1.7.x days?  I am not away of any direct static analysis we do.

PR #5346, opened 2018-10-03
Use right -sourcepath when jruby is invoked via jdb

nomadium:
When --jdb flag is passed to jruby, it's using an incorrect value for -sourcepath flag. This makes the debugger unable to show the source code of the class being debugged.

Before:

```
shirley:jruby miguel$ ./bin/jruby --jdb -e '[0][1..50] = 2'
Initializing jdb ...
> stop in org.jruby.RubyRange.begLen0
Deferring breakpoint org.jruby.RubyRange.begLen0.
It will be set after the class is loaded.
> run
run org.jruby.Main -e "[0][1..50] = 2"
Set uncaught java.lang.Throwable
Set deferred uncaught java.lang.Throwable
>
VM Started: Set deferred breakpoint org.jruby.RubyRange.begLen0

Breakpoint hit: "thread=main", org.jruby.RubyRange.begLen0(), line=188 bci=0

main[1] list
Source file not found: RubyRange.java
main[1] exit
shirley:jruby miguel$ 
```

After:

```
shirley:jruby miguel$ ./bin/jruby --jdb -e '[0][1..50] = 2'
Initializing jdb ...
> stop in org.jruby.RubyRange.begLen0
Deferring breakpoint org.jruby.RubyRange.begLen0.
It will be set after the class is loaded.
> run
run org.jruby.Main -e "[0][1..50] = 2"
Set uncaught java.lang.Throwable
Set deferred uncaught java.lang.Throwable
> 
VM Started: Set deferred breakpoint org.jruby.RubyRange.begLen0

Breakpoint hit: "thread=main", org.jruby.RubyRange.begLen0(), line=188 bci=0
188            long beg = RubyNumeric.num2long(this.begin);

main[1] list
184            return new long[]{beg, len};
185        }
186    
187        final long begLen0(long len) {
188 =>         long beg = RubyNumeric.num2long(this.begin);
189    
190            if (beg < 0) {
191                beg += len;
192                if (beg < 0) {
193                    throw getRuntime().newRangeError(beg + ".." + (isExclusive ? "." : "") + end + " out of range");
main[1] exit
shirley:jruby miguel$ 
```

### DIFF
--- a/bin/jruby.bash
+++ b/bin/jruby.bash
@@ -281,7 +281,8 @@ do
             JAVACMD="$JAVA_HOME/bin/jdb"
           fi
         fi 
-        java_args=("${java_args[@]}" "-sourcepath" "$JRUBY_HOME/lib/ruby/1.9:.")
+        JDB_SOURCEPATH="${JRUBY_HOME}/core/src/main/java:${JRUBY_HOME}/lib/ruby/stdlib:."
+        java_args=("${java_args[@]}" "-sourcepath" "$JDB_SOURCEPATH")
         JRUBY_OPTS=("${JRUBY_OPTS[@]}" "-X+C") ;;
      --client)
         JAVA_VM=-client ;;
--- a/bin/jruby.sh
+++ b/bin/jruby.sh
@@ -226,7 +226,8 @@ do
         else
           JAVACMD="$JAVA_HOME/bin/jdb"
         fi 
-        java_args="${java_args} -sourcepath $JRUBY_HOME/lib/ruby/1.9:."
+        JDB_SOURCEPATH="${JRUBY_HOME}/core/src/main/java:${JRUBY_HOME}/lib/ruby/stdlib:."
+        java_args="${java_args} -sourcepath ${JDB_SOURCEPATH}"
         JRUBY_OPTS="${JRUBY_OPTS} -X+C" ;;
      --client)
         JAVA_VM=-client ;;


nomadium:
btw, I removed the path ```$JRUBY_HOME/lib/ruby/1.9``` for the source path because it doesn't exist and I was also unsure if that value was ever useful when used with jdb but I could be wrong, of course.

kares:
hard-coding paths that make sense during development shouldn't be the resolution here.
pbly you want to `$JRUBY_HOME/lib/ruby/stdlib:.` here so that its usable with a dist build?

nomadium:
I took a look at a recent dist build, e.g.:

```
shirley:jruby miguel$ unzip -l jruby-dist-9.2.0.0-src.zip | grep RubyHash.java
    90078  05-24-2018 15:12   jruby-9.2.0.0/core/src/main/java/org/jruby/RubyHash.java
     8639  05-24-2018 15:12   jruby-9.2.0.0/core/src/test/java/org/jruby/test/TestRubyHash.java
shirley:jruby miguel$ 

shirley:jruby miguel$ unzip -l jruby-dist-9.2.0.0-src.zip | grep lib/ruby/stdlib | grep \.java$ | wc -l
       0
shirley:jruby miguel$ 
```

@kares OK, I'll remove the path including the generated-sources dir since it doesn't make sense for a dist build as you mentioned. However, I wonder if it makes sense to include the ruby stdlib path in the source path of the java debugger, since there are not .java files in that path.

To clarify, which option is preferable?

```JDB_SOURCEPATH="${JRUBY_HOME}/core/src/main/java:${JRUBY_HOME}/lib/ruby/stdlib:."```

Or just:

```JDB_SOURCEPATH="${JRUBY_HOME}/core/src/main/java:."```

headius:
@nomadium Probably not currently, since even when compiled (via JIT) the file paths will not be rooted in the stdlib dir (they are canonicalized to system-level paths).

I approve, will merge.

headius:
@nomadium I should have been more clear...it's fine for stdlib to be in there even if it doesn't work.

I believe we want the same change in the native JRuby launcher, https://github.com/jruby/jruby-launcher.

PR #5347, opened 2018-10-04
optimalize nextPowOfTwo

ahorek:
numberOfLeadingZeros is vectorized and more effective than the original version

It's also possible to use highestOneBit, but I didn't because of this https://bugs.openjdk.java.net/browse/JDK-8199843

```ruby
require 'benchmark/ips'
require 'set'
Benchmark.ips do |x|
  x.report('set') { 100.times { Set[:x, :a, :h, :h, :h, :g, :h, :h, :h, :i] } }
end
```

patch
```
                      8.632k (± 1.9%) i/s -     43.316k in   5.020055s
```
master
```
                      8.384k (±10.6%) i/s -     41.477k in   5.035840s
```


kares:
heh, interesting ... maybe a branch could be eliminated as well?

ahorek:
I took a safe path, the branch could be eliminated only if we can ensure that the input will always be positive and not 0 or 1. Let me confirm that.

ahorek:
```
         0: ldc           #3                  // int -2147483648
         2: iload_0
         3: iconst_1
         4: isub
         5: invokestatic  #4                  // Method java/lang/Integer.numberOfLeadingZeros:(I)I
         8: iushr
         9: iconst_1
        10: ishl
        11: ireturn
```

```
        8.787k (±14.2%) i/s -     43.036k in   5.046647s
```

PR #5348, opened 2018-10-05
revert volatile mod from values[] RubyArray store?

kares:
maybe I am missing smt but from reviewing parts doesn't bring much value

from the commit desc that introduced it ... seems to be due nil fills after assignment, makes no sense?

`RubyArray` isn't thread-safe to start with and this might help improper concurrent usage

... but only `values` has been volatile and not `begin` or `length` fields - so detection only works with internal array resizing and even than the other thread's view might not be correct, due cached `begin` field


headius:
I approve. RubyArray may deserve some proper volatility checks in the future but this partial change was not really sufficient to make sure the new state of the array is properly visible. It also adds volatile write overhead to every (already not thread-safe) update of the RubyArray values store.

PR #5353, opened 2018-10-10
Improve Gem.ruby shim and test it

MrBerg:
Since https://github.com/jruby/jruby/commit/4b749c3542ae36a8c3ca069bcbfce00871d612b8#diff-f0c4ea89199a6d5d1ede2bfd90cd89b8, jruby has differed from cruby in that ENV['RUBY'] overrides the value from (cruby) Gem.ruby.

We recently updated to jruby-9.2.0.0 in our test environments while testing our Java 10 compatibility. For some unknown reason, the environmental variable RUBY was set for all users in our server, which led to some confusion when using Gem.ruby in our application to check whether we were actually running JRuby, since it pointed to either the old Java launcher version or a non-existing ruby binary. I couldn't find ENV['RUBY'] being used outside of testing, so overriding (cruby) Gem.ruby with it no matter if it's a sensical value or not doesn't seem smart. I couldn't figure out some good way of overriding it for testing only, but I assume a decent compromise would be to only override if the value actually points to an existing file. Another option would be to drop the use of ENV['RUBY'] in the Java side, since all cruby libraries that use it seem to have some kind of EnvUtil (e.g. https://github.com/jruby/jruby/blob/master/test/mri/lib/envutil.rb#L17) already.

### DIFF
--- a/core/src/main/java/org/jruby/ext/jruby/JRubyUtilLibrary.java
+++ b/core/src/main/java/org/jruby/ext/jruby/JRubyUtilLibrary.java
@@ -29,6 +29,7 @@
 
 package org.jruby.ext.jruby;
 
+import java.io.File;
 import java.io.IOException;
 import java.net.URL;
 import java.util.Enumeration;
@@ -141,7 +142,7 @@ public class JRubyUtilLibrary implements Library {
     public static IRubyObject classpath_launcher(ThreadContext context, IRubyObject recv) {
         final Ruby runtime = context.runtime;
         String launcher = runtime.getInstanceConfig().getEnvironment().get("RUBY");
-        if ( launcher == null ) launcher = ClasspathLauncher.jrubyCommand(runtime);
+        if ( launcher == null || !(new File(launcher.split(" -cp :")[0]).exists()) ) launcher = ClasspathLauncher.jrubyCommand(runtime);
         return runtime.newString(launcher);
     }
 
--- a/core/src/test/java/org/jruby/ext/JRubyUtilLibraryTest.java
+++ b/core/src/test/java/org/jruby/ext/JRubyUtilLibraryTest.java
@@ -1,0 +1,79 @@
+package org.jruby.ext;
+
+import java.io.File;
+import java.lang.reflect.Field;
+import java.util.Collections;
+import java.util.HashMap;
+import java.util.Map;
+
+import junit.framework.TestCase;
+import org.jruby.Ruby;
+import org.jruby.RubyInstanceConfig;
+import org.jruby.test.MockRubyObject;
+import org.jruby.util.ClasspathLauncher;
+
+import static org.jruby.ext.jruby.JRubyUtilLibrary.classpath_launcher;
+
+public class JRubyUtilLibraryTest extends TestCase {
+
+    /* Ugly hack to modify environmental variables for these tests only
+       source: https://stackoverflow.com/a/496849
+     */
+    public static void setEnv(Map<String, String> newenv) throws Exception {
+        Class[] classes = Collections.class.getDeclaredClasses();
+        Map<String, String> env = System.getenv();
+        for(Class cl : classes) {
+            if("java.util.Collections$UnmodifiableMap".equals(cl.getName())) {
+                Field field = cl.getDeclaredField("m");
+                field.setAccessible(true);
+                Object obj = field.get(env);
+                Map<String, String> map = (Map<String, String>) obj;
+                map.clear();
+                map.putAll(newenv);
+            }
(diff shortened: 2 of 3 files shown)


kares:
seems a bit weird, not sure if this makes much sense except for hopefully a very specific case of yours ...
there's an explicit check at `Gem.ruby` to support packed launching - its clear with `if jarred_path?`

https://github.com/jruby/jruby/commit/4b749c3542ae36a8c3ca069bcbfce00871d612b8#diff-f0c4ea89199a6d5d1ede2bfd90cd89b8 is nothing new, just moving pieces for "cleaner" boot (so we avoid JI reflecting JRuby's internals)

MrBerg:
I'm not gonna deny that this is mostly a user error, probably due to some weird setup with several ruby version managers at some point, but to be fair there is no documentation (apart from actual code) about the RUBY environmental variable anywhere in the Ruby ecosystem that I could find. I'm still unsure what the use of the RUBY value is and why it is for jarred paths only, and why it was introduced in that commit, because it's not used in either https://github.com/jruby/jruby/blob/4b749c3542ae36a8c3ca069bcbfce00871d612b8/core/src/main/java/org/jruby/util/ClasspathLauncher.java or https://github.com/rubygems/rubygems/blob/master/lib/rubygems.rb#L901

kares:
> I'm not gonna deny that this is mostly a user error, probably due to some weird setup with several ruby version managers at some point, but to be fair there is no documentation (apart from actual code) about the RUBY environmental variable anywhere in the Ruby ecosystem that I could find.

believe its meant to be an impl detail atm - changing 'ruby' was the most direct way of supporting spawn-ing for embedded (jarred) cases. which isn't your case you have a full jruby install but there's other ways to run jruby so it supports those.

> I'm still unsure what the use of the RUBY value is and why it is for jarred paths only, and why it was introduced in that commit, because it's not used in either https://github.com/jruby/jruby/blob/4b749c3542ae36a8c3ca069bcbfce00871d612b8/core/src/main/java/org/jruby/util/ClasspathLauncher.java or https://github.com/rubygems/rubygems/blob/master/lib/rubygems.rb#L901

`Gem.ruby` its merged upon *rubygems.rb* since JRuby has some of its RGs files laid on top.

realized the thing you're "complaining" about from the mentioned commit in 9.2 is likely this line: https://github.com/jruby/jruby/commit/4b749c3542ae36a8c3ca069bcbfce00871d612b8#diff-f0c4ea89199a6d5d1ede2bfd90cd89b8R198 ... which I managed to ignore in my previous commit (sorry about that), its definitely a change and not just refactoring ... there's likely a reason I did that, will elaborate.

kares:
okay, so JRuby's using ENV['RUBY'] internally ... its being set for the [runtime as it initializes](https://github.com/jruby/jruby/blob/master/core/src/main/java/org/jruby/RubyInstanceConfig.java#L670-L676)
... so the added check is there simply to re-use the (already) set `ENV['RUBY']` thus the added check is just confusing and relying on the fact that `-cp` switch will be first with the `java -cp ...` command. 

while I am not against the check, am also not sure about adding it since it might often point to a non-existing location (such as simply `java -cp ... org.jruby.Main -S`) this could introduce more issues in the end.

kares:
not sure, in theory someone might be tuning `export RUBY=...` its not documented but still.
seems that the check existed since its [introduction](https://github.com/jruby/jruby/commit/5c45c39163edc37267199db023323704a4410641) ... we could try forcing ENV['RUBY'] to always set but than someone else might complain that they might want to use it and its surprising that we change it.

kares:
> Removing that check would also solve our problem. Is there any use case for JRuby using the System.getenv() value for ENV['RUBY'] (https://github.com/jruby/jruby/blob/master/core/src/main/java/org/jruby/RubyInstanceConfig.java#L143)?

circling back here - yes there is, since MRI also allows for 'RUBY' to be propagated from system env.

> Another option would be to drop the use of ENV['RUBY'] in the Java side, since all cruby libraries that use it seem to have some kind of EnvUtil

that would break process launching for embedded cases. it could be done without setting `ENV['RUBY']` but would be quite some effort + with libraries such as rake relying on `ENV['RUBY']` already, current way makes more sense.


back to the original issue this PR was trying to solve:

> the environmental variable RUBY was set for all users in our server, which led to some confusion when using Gem.ruby in our application to check whether we were actually running JRuby

this would have broke MRI as well (e.g. rake's process launching). you're mostly using an API for smt else that its meant for. recommend you check JRuby using `defined? JRUBY_VERSION` and remove `RUBY` from system env to avoid future issues.

PR #5357, opened 2018-10-11
WIP: respect frozen_string_literal

ahorek:
reference
https://bugs.ruby-lang.org/issues/15118

see also
https://github.com/jruby/jruby/blob/ae777f0039d1d7f698bb171109a4ae3ac8582b1e/core/src/main/java/org/jruby/ir/IRBuilder.java#L1083


headius:
👍

headius:
Ignore the test:jruby:aot failure; it is a regression that snuck in with timeout fixes.

headius:
Oh oops...it just occurred to me that "WIP" might mean you have more coming? I kinda got lost in merging today.

ahorek:
I was just waiting for CI to confirm it won't break anyhing else. Thanks for merging so quickly!

PR #5359, opened 2018-10-11
1.7.28 version - update jruby-openssl version for IBM cipher fix

jkuchta:
Create a 1.7.28 JRuby version with updated jruby-openssl for IBM JDK SSL cipher support (fixed in jruby-openssl 0.9.21).


kares:
seems it can't just stay dead already :) ... is it problematic for you to simply upgrade jossl after install?
... also if you specify in *Gemfile* it will use that

jkuchta:
I know ;-)

Unfortunately, we have a legacy application that gets warbled so even when including a new jruby-openssl version in the Gemfile it still uses the version that is included in jruby-jars. Long-term I want to move to JRuby 9.2.x.x but the upgrade breaks a lot of legacy code and will require more time to complete.

jkuchta:
@kares agreed. We are currently on 1.7.27 - which has jruby-openssl 0.9.18 and the first version of jruby-openssl that seems to support the IBM ciphers is 0.9.21. In the Gemfile I've included openssl 0.10.1 which works when running the project (rails) directly but does not work once warbled.

headius:
If you have time to explain what breaks upgrading to 9x that would be great. I'm not sure we have the resources or inclination to try to push a 1.7.28 but most of that falls on @enebo.

jkuchta:
@headius sure, the main issue is Warbler fails with a stack overflow on our user model. I suspect this is something related to having to upgrade Devise from 3.x to 4.x. Separate from that there are a handful of other gem updates required that would need a bit of refactoring (these aren't nearly as bad as the Warbler issue).

As far as the 1.7.28 - is there anything else needed here besides changing the 1.7.28-SNAPSHOT versions to 1.7.28? The change itself is fairly small but adding those IBM ciphers would be a big help to give more time for a smoother transition to 9.2.x.x.

kares:
its been EOL for a while, thus maybe if you really need to stay on 1.7 using IBM Java maybe we shall insteadclook into getting the upgraded jossl working when Warbled ... worst case scenario is to repackage the archive bits.

jkuchta:
@kares as far as repackaging, are you refering to the jruby-stdlib and just replacing the jopenssl.jar?

kares:
that would be an option yes - having your own stdlib.jar around that you would replace. 
other one is to understand why its not picking up the declared .jar which is likely a servlet container issue, maybe also solvable by changing RGs version used. if you need support around, feel free to get in touch.

jkuchta:
@kares thanks - is there a way I can print what jruby-openssl version is being used?

kares:
any objections against closing this one? as there are no planned JRuby 1.7.x releases.

PR #5363, opened 2018-10-12
Improvements for modules on Java 9+

headius:
This PR will track a few small commits to improve how we start up on Java 9+, with a specific eye for running well on Java 11.

### DIFF
--- a/bin/jruby.bash
+++ b/bin/jruby.bash
@@ -362,12 +362,24 @@ if $cygwin; then
 
 fi
 
+# Allow overriding default JSA file location
+if [ "$JRUBY_JSA" == "" ]; then
+  JRUBY_JSA=${JRUBY_HOME}/lib/jruby.jsa
+fi
+
 # Determine whether to use -classpath or --module-path
 classmod_flag="-classpath"
-bootclasspath_flags="-Xbootclasspath/a:"
 if [ -d $JAVA_HOME/jmods ]; then
+  # Use module path instead of classpath
   classmod_flag="--module-path"
-  bootclasspath_flags="-Xverify:none --module-path "
+
+  # Switch to non-boot path since we can't use bootclasspath on 9+
+  NO_BOOTCLASSPATH=1
+
+  # If we have a jruby.jsa file, enable AppCDS
+  if [ -f $JRUBY_JSA ]; then
+    JAVA_OPTS="$JAVA_OPTS -XX:+UnlockDiagnosticVMOptions -XX:SharedArchiveFile=$JRUBY_JSA"
+  fi
 fi
 
 # Run JRuby!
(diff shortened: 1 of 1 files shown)


headius:
@kares @enebo I'm terrible at [ba]sh...double check my work here. Any concerns, like about -Xverify:none?

headius:
Ok some thoughts on verification, since I'm reluctant to switch it completely off like this...

First, exposure. There's a number of pages that recommend never turning it off, but as we know lots of folks turn it off in development to speed up commands. This change in the `jruby` script, however, would also happen in a production setting. You'd have to explicitly turn it back on with VERIFY_JRUBY=1.

I have not researched the likelihood or prior art of exploiting Hotspot with verification off.

The main reason this needs to be done is because the bootclasspath flag and the Java module system do not work together, due to boot classes being able to circumvent some aspects of the module system's security guarantees (I believe this is the concern).

So if we can't turn verification off and can't use bootclasspath, we have two options I see:

* Use AppCDS to eliminate most of the cost of verification. However this is still a work in progress, both in OpenJDK and in the tool I'm working on for JRuby users: https://github.com/jruby/jruby-startup
* Always verify.

This last option will slow down startup for most users because they have been running with JRuby in bootclasspath (and therefore unverified) previously.

For `jruby -S gem list` with just a couple dozen gems:

No flags: 3.0-3.2s
With --dev: 2.18s
Verify all: 3.2-3.6s
Verify all --dev: 2.4s

The JIT results are notable noisy, but overall the verification appears to add from 0.2 to 0.4s to startup.

With AppCDS, no verification, --dev: 1.7s
With AppCDS, full verification, --dev: 1.7-1.8s

So turning verification on may be an acceptable step backwards to help move toward AppCDS use?

Note this all comes into play only on Java 9+. Java 8 will continue to work as it does today.

headius:
Note that AppCDS for a very *large* app may only help the first few seconds of startup; code loaded after that will be verified anew. This does not really differ much from today, though, since with bootclasspath we limited the unverified body of code to just what JRuby loaded at boot.

headius:
Ok after some discussion, we're going with the following changes.

When running on Java 9+ (detected by JAVA_HOME/jmods dir), we will not use the unverified bootclasspath logic in the bash script at all and also make the following tweaks:

* Switch `-classpath` to `--module-path`.
*  If JRUBY_HOME/lib/jruby.jsa exists (or another path specifiable with JRUBY_JSA env) add appropriate AppCDS flags to use it.

This means all Java 9+ will now be fully verified, but in module path. Use AppCDS to eliminate the startup hit of that verification.

headius:
Gist comparing Java 8 startup versus Java 11 with and without AppCDS: https://gist.github.com/headius/6228000dcad848c137554496940a9c31

PR #5364, opened 2018-10-12
improve gcd performance

ahorek:
this patch uses a variant of https://en.wikipedia.org/wiki/Binary_GCD_algorithm

https://github.com/ruby/ruby/pull/1596

master
```
         Time#subsec      5.377M (±17.9%) i/s -     24.868M in   4.991341s
              Time#-     17.696M (± 9.1%) i/s -     86.842M in   4.982315s
          Time#round      5.162M (±10.1%) i/s -     25.510M in   4.997497s
           Time#to_f     15.985M (± 7.8%) i/s -     78.822M in   4.990285s
           Time#to_r      5.143M (± 8.4%) i/s -     25.479M in   4.993259s
          Rational#+      8.195M (± 4.6%) i/s -     40.802M in   4.995908s
          Rational#-      7.797M (± 5.2%) i/s -     38.806M in   4.999148s
          Rational#*     10.606M (± 4.5%) i/s -     52.840M in   4.994995s
```

patch
```
         Time#subsec      6.169M (±18.9%) i/s -     27.650M in   4.983793s
              Time#-     18.941M (± 9.2%) i/s -     92.775M in   4.980521s
          Time#round      5.118M (±10.9%) i/s -     25.194M in   4.993703s
           Time#to_f     17.562M (± 8.2%) i/s -     86.509M in   4.988423s
           Time#to_r      6.094M (± 7.6%) i/s -     30.234M in   4.993605s
          Rational#+      8.519M (± 4.4%) i/s -     42.450M in   4.994747s
          Rational#-      9.363M (± 4.1%) i/s -     46.672M in   4.995156s
          Rational#*     12.275M (± 4.9%) i/s -     61.094M in   4.994644s
```

### DIFF
--- a/core/src/main/java/org/jruby/util/Numeric.java
+++ b/core/src/main/java/org/jruby/util/Numeric.java
@@ -510,19 +510,33 @@ public class Numeric {
      *
      */
     public static long i_gcd(long x, long y) {
-        if (x < 0) x = -x;
-        if (y < 0) y = -y;
+        long shift, uz, vz;
+        x = Math.abs(x);
+        y = Math.abs(y);
+        if (x == 0) {
+            return y;
+        }
+        if (y == 0 || x == y) {
+            return x;
+        }
+        uz = Long.numberOfTrailingZeros(x);
+        vz = Long.numberOfTrailingZeros(y);
+        shift = Math.min(uz, vz);
 
-        if (x == 0) return y;
-        if (y == 0) return x;
+        x >>= uz;
+        y >>= vz;
 
-        while (x > 0) {
-            long t = x;
-            x = y % x;
-            y = t;
+        while (x != y) {
+            if (x > y) {
+                x -= y;
+                x >>= Long.numberOfTrailingZeros(x);
+            } else {
+                y -= x;
+                y >>= Long.numberOfTrailingZeros(y);
+            }
         }
 
-        return y;
+        return x << shift;
     }
 
     /** f_gcd
--- a/core/src/main/java/org/jruby/util/Numeric.java
+++ b/core/src/main/java/org/jruby/util/Numeric.java
@@ -513,7 +513,7 @@ public class Numeric {
         long shift, uz, vz;
         if (x == Long.MIN_VALUE) {
             if (y == Long.MIN_VALUE)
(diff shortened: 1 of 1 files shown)


core/src/main/java/org/jruby/util/Numeric.java
```diff
@@ -511,6 +511,15 @@ public static boolean f_odd_p(ThreadContext context, IRubyObject i) {
      */
     public static long i_gcd(long x, long y) {
         long shift, uz, vz;
+        if (x == Long.MIN_VALUE) {
+            if (y == Long.MIN_VALUE)
+                throw new IllegalArgumentException("gcd() is too big");
```
kares: but if x == y here couldn't we simply `return x;` ? there wasn't any kind of explicit Java exception being thrown before

ahorek: by definition the greatest common divisor (gcd) of two or more integers, which are not all zero, is the largest **positive integer** that divides each of the integers.  (https://en.wikipedia.org/wiki/Greatest_common_divisor)

Long.MIN_VALUE => -922337203685477580**8**
Long.MAX_VALUE => 922337203685477580**7**

if x and y are Long.MIN_VALUE, we can't represent it as a positive number because
Math.abs(Long.MIN_VALUE) => -9223372036854775808


gcd(Long.MIN_VALUE, Long.MIN_VALUE)
IMO it should raise an exception because we can't represent a correct result

if we use x than the result will be a negative number which is unexpected

bignum math
-9223372036854775808.gcd(-9223372036854775808)
=>9223372036854775808
the result is positive for the same parameters
-9223372036854775808.i_gcd(-9223372036854775808)
=> -9223372036854775808

given the fact, that the previous algorithm worked this way (and have the same limitation), also Math.abs doesn't raise
Math.abs(Long.MIN_VALUE) => -9223372036854775808
I'll return to the original behaviour

note that this is an edge case and probably never happens in the real code

kares: yy, thanks - have realized that afterwards. not sure what really should happen, maybe an explicit exception is for the best. but its great that for now it will behave as before.

PR #5367, opened 2018-10-13
enable Integer#sqrt

ahorek:
```ruby
require 'benchmark/ips'

def sqrt
  1000.times do |num|
    Integer.sqrt(num)
  end
end

Benchmark.ips do |x|
  x.report('sqrt') { sqrt }
end
```

master
```
 sqrt      9.153k (±20.4%) i/s -     42.000k in   5.027226s
```
patch
```
 sqrt     16.244k (±14.2%) i/s -     79.422k in   5.063626s
```


kares:
najs, esp. since there seem to be no test broken despite the FIXME :)

PR #5369, opened 2018-10-15
Make subtract and addition operators work with DateTime objects

massive:
We recently discovered a few nasty bugs that affect DateTime calculations when the precision of time is high enough. 

As a practical example, when subtracting values produced by `DateTime.current` (via Active Support), you'll get weird results. e.g.:

```irb
[4] pry(main)> DateTime.current - DateTime.current
17841599/43200000
[5] pry(main)> DateTime.current - DateTime.current
-609/1000
```

I traced the problem down to `RubyDate#op_plus_numeric`, where value multiplication produce incorrect results seemingly due to floating point inaccuracy. I was able to "fix" the problem using round, but obviously that is not really elegant solution (maybe it should use BigDecimal under the hood or something?).

While I was working with this bug, I also discovered another bug, which turned out to be related: namely when adding sub-millisecond precision value to `DateTime` object, JRuby interpreter threw an error.

Both of the bugs are reproducible using provided specs.

### DIFF
--- a/core/src/main/java/org/jruby/ext/date/RubyDate.java
+++ b/core/src/main/java/org/jruby/ext/date/RubyDate.java
@@ -1162,15 +1162,15 @@ public class RubyDate extends RubyObject {
         //   sub_millis -= 1
         //   ms += 1
         // end
-        RubyNumeric val = (RubyNumeric) RubyFixnum.newFixnum(runtime, DAY_MS).op_mul(context, n);
+        RubyNumeric val = (RubyNumeric) RubyFixnum.newFixnum(runtime, DAY_MS).op_mul(context, n).convertToFloat().round(context, RubyFixnum.newFixnum(context.runtime, 10));
 
         RubyArray res = (RubyArray) val.divmod(context, RubyFixnum.one(context.runtime));
         long ms = ((RubyInteger) res.eltInternal(0)).getLongValue();
         RubyNumeric sub = (RubyNumeric) res.eltInternal(1);
         //if ( sub.isZero() ) sub = RubyFixnum.zero(runtime); // avoid Rational(0, 1)
         RubyNumeric sub_millis = (RubyNumeric) subMillis(context.runtime).op_plus(context, sub);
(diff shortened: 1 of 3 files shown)


massive:
@kares Thanks for your input. I agree, the rounding part is a subpar solution.

However, while using longs fixes the problem with addition, it's not enough to fix the problems with subtraction. For instance, within the provided spec, the result for `val` in `op_plus_numeric` will be `-864.0000000000001`. The imprecision of the value will flow through the whole calculation causing odd effects (e.g. `(dt - (dt - 0.0001)).to_f => 0.0001` but (dt - (dt - 0.00001)).to_f => -0.9999899884258123`.

I'm not quite yet there to understand how and where to best deal with the problem though.

kares:
yeah having floats around is always problematic, think I have some ideas on how to proceed.
we should be able to somehow detect the "over-flowing" sub-traction case you mentioned.

kares:
also as a work-around you should be able to run fine with JRuby 9.1 .. 9.2.1 release should be close thought

kares:
for the record, only the add part is to be considered a regression, the sub-tract doesn't work in 9.1 either:
```
DateTime#- is able to subtract sub-millisecond precision values FAILED
Expected (7599824371187713/759982437118771200000)
 to equal (1/100000)
```
... we do the calculation "right" in 9.2 ~ the same as in previous JRubies.

can not use the rounding as is since (as CI revealed) that introduces more problems, will try to port the `Date#-`calculation logic to semi-match MRI, its tricky since we're using JODA while MRI has raw structs.

kares:
pushed a modified commit (with the long fix since its a regression) into master: https://github.com/jruby/jruby/commit/bf2a83c0e7e99564126c333dc206255f5a355016 ... so this doesn't hold up for long. tagged the failing subtract specs, might push a fix for that later - have one but it doesn't scale beyond 5 decimal places.

kares:
going to close this one, the provided specs are now passing ... there's still more work to do.
guess we'll track forward as users report, unless some of use decides to pro-actively work on it :)

PR #5371, opened 2018-10-15
Raise error on negative usec param to Time.new.

### LINKED ISSUE #5370: Time can be initialized with negative nanoseconds
### Environment

```
jruby 9.2.0.0 (2.5.0) 2018-05-24 81156a8 OpenJDK 64-Bit Server VM 25.181-b13 on 1.8.0_181-b13 [linux-x86_64]
Linux 4.18.12-arch1-1-ARCH #1 SMP PREEMPT Thu Oct 4 01:01:27 UTC 2018 x86_64 GNU/Linux
```

### Expected Behavior

Initializing `Time` with a negative value for seconds should raise an error.

On MRI it does:

```sh
irb(main):001:0> Time.new(2018, 1, 1, 10, 30, -0.1)
        4: from /home/tils/.rubies/ruby-2.5.1/bin/irb:11:in `<main>'
        3: from (irb):1
        2: from (irb):1:in `new'
        1: from (irb):1:in `initialize'
ArgumentError (argument out of range)
```

On JRuby it does not, when only given fraction of a second:

```sh
irb(main):003:0> Time.new(2018, 1, 1, 10, 30, -0.1)
=> 2018-01-01 10:29:59 +0100
```

When the number is small enough, it will have negative `nsec`:

```sh
irb(main):011:0> Time.new(2018, 1, 1, 10, 30, -0.0009).nsec
=> -900000
```

which becomes a problem in very specific serializations, e.g. with Sequel and postgres-jdbc it produces this invalid timestamp representation:

```sh
[2] pry(main)> DB.literal(Time.new(2018, 1, 1, 10, 1, -0.0009))
=> "'2018-01-01 10:01:00.-00900+0000'"
```

### Actual Behavior

Behave like MRI and always raise an ArgumentError when `sec < 0`.

headius:
Fixes #5370.

I'll add specs momentarily.

### DIFF
--- a/core/src/main/java/org/jruby/RubyTime.java
+++ b/core/src/main/java/org/jruby/RubyTime.java
@@ -93,6 +93,7 @@ public class RubyTime extends RubyObject {
 
     private static final BigDecimal ONE_MILLION_BD = BigDecimal.valueOf(1000000);
     private static final BigDecimal ONE_BILLION_BD = BigDecimal.valueOf(1000000000);
+    public static final int TIME_SCALE = 1000000000;
 
     private DateTime dt;
     private long nsec;
@@ -1715,7 +1716,6 @@ public class RubyTime extends RubyObject {
         int i_args1 = argToInt(context, args, 1 + 2, 0);
         int i_args2 = argToInt(context, args, 2 + 2, 0);
         int i_args3 = argToInt(context, args, 3 + 2, 0);
-        int i_args4 = argToInt(context, args, 4 + 2, 0);
         //int i_args5 = argToInt(context, args, 5 + 2, 0);
 
         // Validate the times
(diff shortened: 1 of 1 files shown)


spec/ruby/core/time/new_spec.rb
```diff
@@ -96,4 +96,12 @@
     Time.new(2000, 1, 1, 0, 0, 0, 86400 - 1).utc_offset.should == (86400 - 1)
     lambda { Time.new(2000, 1, 1, 0, 0, 0, 86400) }.should raise_error(ArgumentError)
   end
+
+  it "raises ArgumentError if the seconds argument is negative" do
+    lambda { Time.new(2000, 1, 1, 0, 0, -1) }.should raise_error(ArgumentError)
+  end
+
+  it "raises ArgumentError if the utf_offset argument is greater than or equal to 10e9" do
```
til: *utc_offset

headius:
@til I think the -1 spec is just as good as the -0.1 spec, since it's just a simple test for negative seconds. The fact that it's a fractional amount does not appear to change the logic.

Issue #5287, opened 2018-08-21
sysread hangs on EOF on mac


andrewdotn:
### Environment

Provide at least:
* JRuby version (`jruby -v`) and command line (flags, JRUBY_OPTS, etc)
`jruby 9.2.0.0 (2.5.0) 2018-05-24 81156a8 Java HotSpot(TM) 64-Bit Server VM 25.161-b12 on 1.8.0_161-b12 +jit [darwin-x86_64]`
* Operating system and platform (e.g. `uname -a`)
`Darwin localhost 17.7.0 Darwin Kernel Version 17.7.0: Thu Jun 21 22:53:14 PDT 2018; root:xnu-4570.71.2~1/RELEASE_X86_64 x86_64`

### Expected Behavior

With this script, I expect to see EOFError raised:

```
f = File.open('/etc/passwd', 'r')
loop do
  x = f.sysread(1024)
  puts "read #{x.length}"
end
```

This is what MRI ruby does, as well as JRuby 9.1.15.0 and earlier.

### Actual Behavior

The script hangs at EOF with JRuby 9.2.0.0 on the mac. The Java debugger shows it’s inside `RubyThread.select()` calling `Selector.select()`. Linux seems to be unaffected.

`git bisect` traces it to https://github.com/jruby/jruby/commit/dad9ab05ae1d885bfa2cfa7e5659a1252392b117, which was cherry-picked as https://github.com/jruby/jruby/commit/3493a91ce04527900b5f82c3ba105a30c03e9078, so 9.1.16.0 and 9.1.17.0 are affected as well.

headius:
Thank you for your help on this!

PR #5373, opened 2018-10-17
[#5287] Use epoll selector for file IO on macosx

alexis779:
This PR addresses https://github.com/jruby/jruby/issues/5287.

Without it, unit test gets stuck on the select call. See [Thread dump](https://gist.github.com/alexis779/b2eae94b3a90f9758e58499dbfeebf2f).

It needs https://github.com/jnr/jnr-enxio/pull/29 feature to avoid `KQSelector` and fallback on `EpollSelector`. We pass `isFile=true` in `NativeDeviceChannel` constructor in the case of a file. This avoids getting stuck on macosx when calling kevent system call at the end of the file.

kqueue does not support EOF. There is no easy way to skip select call to avoid blocking for ever.

We still want to use KQSelector for sockets. Just not vnodes.

### DIFF
--- a/BUILDING.md
+++ b/BUILDING.md
@@ -157,7 +157,7 @@ If you are familiar with Java debuggers, you can attach one to a JRuby process u
 The exact flag may vary with debugger and platform:
 
 ```
-bin/jruby -T-J-agentlib:jdwp=transport=dt_socket,server=y,suspend=y,address=5005 <rest of arguments>
+JRUBY_OPTS="-J-agentlib:jdwp=transport=dt_socket,server=y,suspend=y,address=localhost:5005" bin/jruby <rest of arguments>
 ```
 #### JRuby internal unit tests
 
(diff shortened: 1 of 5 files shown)


kares:
there seems to be an (unrelated) compilation failure with CI ... maybe you should rebase of recent master

alexis779:
```
[ERROR] /home/travis/build/jruby/jruby/core/src/main/java/org/jruby/util/RegularFileResource.java:[208,19] error: no suitable constructor found for NativeDeviceChannel(int,boolean)
```

`NativeDeviceChannel` is part of a different lib. We need to merge in https://github.com/jnr/jnr-enxio/pull/29 in a new jnr-enxio release, version 0.19 ?

headius:
I'm on it.

headius:
This may fail initially if the jnr-enxio 0.19 release has not propagated, but it's merged!

Issue #5333, opened 2018-09-28
no result with glob pattern "\\"


ahorek:
### Environment
jruby 9.2.1.0-SNAPSHOT (2.5.0) 2018-09-28 8e6904f Java HotSpot(TM) 64-Bit Server VM 25.181-b13 on 1.8.0_181-b13 +jit [linux-x86_64]

after this change https://github.com/rails/rails/pull/33860
jruby build started to fail https://travis-ci.org/rails/rails/jobs/434684112

### Expected Behavior
mri
```
Dir["/mnt/c/moje/graal/rails/rails/actionpack/test/fixtures/../\\../test/abstract_unit.rb*"]
=> ["/mnt/c/moje/graal/rails/rails/actionpack/test/fixtures/../../test/abstract_unit.rb"]
```

### Actual Behavior
jruby
```
Dir["/mnt/c/moje/graal/rails/rails/actionpack/test/fixtures/../\\../test/abstract_unit.rb*"]
=> []
```

headius:
Confirmed locally on master. Note that it works if the `\\` are replaced with `//`:

```
irb(main):005:0> Dir["/Users/headius/projects/jruby/lib/../../bench2018/*.rb*"]
=> ["/Users/headius/projects/jruby/lib/../../bench2018/run_all.rb", "/Users/headius/projects/jruby/lib/../../bench2018/bench_mandelbrot.rb", "/Users/headius/projects/jruby/lib/../../bench2018/bench_regexp_match.rb", "/Users/headius/projects/jruby/lib/../../bench2018/args.rb", "/Users/headius/projects/jruby/lib/../../bench2018/bench_weird_sort.rb", "/Users/headius/projects/jruby/lib/../../bench2018/bench_chaining.rb", "/Users/headius/projects/jruby/lib/../../bench2018/bench_aref.rb", "/Users/headius/projects/jruby/lib/../../bench2018/bench_fields_hashes.rb", "/Users/headius/projects/jruby/lib/../../bench2018/bench_strptime.rb", "/Users/headius/projects/jruby/lib/../../bench2018/bench_red_black.rb", "/Users/headius/projects/jruby/lib/../../bench2018/bench_kwargs.rb", "/Users/headius/projects/jruby/lib/../../bench2018/bench_hash.rb"]
irb(main):006:0> Dir["/Users/headius/projects/jruby/lib/../\\../bench2018/*.rb*"]
=> []
[...]

PR #5378, opened 2018-10-19
Always use full logic for processing glob segments.

headius:
This is a workaround for issues found in #5333. I discovered while
researching a fix for that bug that our `glob_helper` appears to
be missing "else" logic for the segment processing that handles
magic characters and unescaping. Put simply, if there's no magic
in the segment it leaves it entirely unprocessed, allowing its
contents to just get rolled into the next segment loop (which may
or may not also have magic). However this leaves escapes in place.

This "fix" basically just omits the magic check for each segment
and allows full processing.

### DIFF
--- a/core/src/main/java/org/jruby/util/Dir.java
+++ b/core/src/main/java/org/jruby/util/Dir.java
@@ -579,9 +579,15 @@ public class Dir {
 
                 continue;
             case '\\':
-                if (escape && i == end) return false;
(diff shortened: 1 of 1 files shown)


headius:
This did not pass but it seems like the only failures are for jarred Ruby scripts/gems. I'll see if I can suss out why those are now failing.

headius:
It looks like limiting my change to only segments that contain `\\` chars will work ok.

Issue #5233, opened 2018-06-28
jruby (ext) loading left-overs to consider


kares:
... left-overs from https://github.com/jruby/jruby/pull/5205

- do we force users extension authors to do a `require 'jruby'` ... loosely needed atm
  (most exts tend to have an explicit `require 'jruby'` or use the new `load_ext` mechanism)
- JRuby auto-loads the `java` library (since ~ 1.6/1.7) think about NOT auto loading `require 'jruby'` 
  now, its auto-loaded from kernel which was likely not intentional (should only auto `require 'java'`)?

- document extension loading (wiki): `require 'jruby'; JRuby.load_ext('foo.bar.MyExtClass')`

- need to propagate `load_ext` changes to gems we tend to boot often (on every run) ~ due RGs : 
  [~~jruby-openssl~~](https://github.com/jruby/jruby-openssl/pull/167), [~~jruby-readline~~](https://github.com/jruby/jruby-readline/pull/3), [psych](https://github.com/ruby/psych/pull/365), [jar-dependencies](https://github.com/mkristian/jar-dependencies/pull/67)
  ... and get them to propagate releases (so this PR delivers the promised loading speed improvement)

enebo:
@kares will load_ext work on 9.1.x?  It is not unsurmountable if it won't but this would make it much easier for extensions to update. 9.1.x is probably living until beginning of 2019 at least.

kares:
@enebo we could back-port if it makes sense ...
but than I am not sure we need that - all of these gems handle `load_ext` backwards compatibly.
your assumption would be that it would help adoption - migrating gems to an official way of ext loading? 

>  9.1.x is probably living until beginning of 2019 at least.

that "long" for 9.1 - thought you guys do not plan on supporting multiple 2.x versions in parallel.

enebo:
@kares I guess I did not realize load_ext worked on 9.1.  If so then no reason to backport.  Right now most exts I have written do this: 'Java::oj.OjLibrary.new.load(JRuby.runtime, false)'.

At this point I think we have reverted to the idea of supporting 2 versions only.  The problem with only one 9.x is that we cannot just say 9.2.0.0 is out use it!  We know it takes several point releases to iron over the work made (2.3->2.5 was larger than we actually planned since we skipped 2.4 too).   I am hoping 9.2.x matures faster than 9.1.x but there has to be some overlap so people don't feel dropped and angry about it.

kares:
[...]

Issue #4528, opened 2017-03-09
Not every class has a superclass?


ioquatix:
It seems like sometimes  `ObjectSpace.each_object(Class).collect(&:superclass)` may fail in JRuby but works fine in MRI.

Example of failure here: https://travis-ci.org/rubyworks/facets/jobs/209292853

```
    undefined method `superclass' for #<Class:Java::OrgJrubyExtPsych>
    lib/core/facets/class/subclasses.rb:20
    18       list = []
    19       ObjectSpace.each_object(Class) do |c|
 => 20         list.unshift c if c.superclass == self
    21       end
    22       list.uniq
```

headius:
I believe this is a small bug in each_object: it is returning modules as well as classes for `each_object(Class)`.

ioquatix:
Ah, awesome, so it would be a good fix. Perhaps even a candidate for a spec?

headius:
Actually it looks even narrower. I'm only seeing a small handful of these unusual modules. They shouldn't be coming up at all.

```
$ jruby -e 'ObjectSpace.each_object(Class) {|c| p c.class.ancestors unless c.respond_to? :superclass}'
[Module, Object, Kernel, BasicObject]
[Module, Object, Kernel, BasicObject]
[Module, Object, Kernel, BasicObject]
[Module, Object, Kernel, BasicObject]
[Module, Object, Kernel, BasicObject]
[Module, Object, Kernel, BasicObject]
[Module, Object, Kernel, BasicObject]
[Module, Object, Kernel, BasicObject]
[Module, Object, Kernel, BasicObject]
[Module, Object, Kernel, BasicObject]
[Module, Object, Kernel, BasicObject]
```

Only one of these modules has a name: `Java::JavaPackage`. There's definitely a spec in here somewhere, but we're doing something pretty extraordinary.

headius:
Ahh, they are indeed the pseudo-modules we use to represent the Java package structure:

```
$ jruby -e 'ObjectSpace.each_object(Class) {|c| p JRuby.reference(c) unless c.respond_to? :superclass}'
#<Class:Java::OrgJrubyPlatform>
#<Class:Java::Javax>
#<Class:Java::JavaxManagement>
#<Class:Java::OrgJrubyUtil>
#<Class:Java::OrgJrubyRuntime>
#<Class:Java::JavaLang>
#<Class:Java::OrgJruby>
#<Class:Java::JavaUtil>
#<Class:Java::Org>
[...]

PR #5381, opened 2018-10-23
only require 'java', users should explicitly require 'jruby'

kares:
seems to makes sense as using JRuby internals make the runtime "dirty"
e.g. `ObjectSpace` will end up finding invalid class pieces (e.g. for UNDEF)

still, its kind of a small "break" in term-of backwards compatibility
although most scripts floating around do the explicit require 'jruby'
... before scripting with internals such as `JRuby.runtime`

left-over from #5205 as noted at #5233


kares:
marking for 9.2 ... its not a hard requirement, more of a discussion - we could post-pone for 9.23 
would be nice to have "less" loading by default
... and as noted above the internals are changing some class internals atm: see https://github.com/jruby/jruby/issues/4528

headius:
@enebo @kares Where do we stand on this? I'm not opposed, but I have no idea whether these features are in use elsewhere.

enebo:
I believe we should constrain require 'jruby' to only the places we know it will be used.  Putting it near the top of JI makes no sense to me.  So I think we should apply this PR but I also wonder why it is here at all?  Do we use #reference or something somewhere?  If so the second part of this PR would be to put the include wherever we do that.

kares:
think it has been there historically to have `JRuby.runtime` available - mostly due ext loading.
but we know have a better way of loading exts using `JRuby::Util.load_ext(name)` which avoids initializing JI for JRuby's internals ... 
on places JRuby still needs `JRuby.runtime` an explicit `require 'jruby'` should do. 
the only down side this has is that it might break some extensions 
but I believe most of them should `require 'jruby'` as they use `JRuby.runtime`

enebo:
9.3.0.0 seems like a good time...significant # and if this does break some extensions we have some time to get them to add a single line to fix it.

kares:
okay - will update PR as this implicit `require 'jruby'` now lives in Java

kares:
PR is now ready ... the `require 'jruby'` removal is pretty much just [two lines](https://github.com/jruby/jruby/pull/5381/files#diff-a40568b9da02963919222c8b1eb53774L1735)
needed to adjust tests and one place we assumed the library to be loaded, also avoided the requirement (to `require 'jruby'`) from `io/nonblock` (would cause self-reflection on JRuby's Java classes so its all for the best)

headius:
I'm approving this but it does feel a little weird now to have `JRuby` defined by default without any methods in it. Perhaps it was a bad idea originally to use the namespace directly for all these methods that would be better in their own namespaces (compilation utilities etc).

in any case, I'm fine going forward with this since most people don't even know about these utilities, and they're very much JRuby-specific anyway.

PR #5384, opened 2018-10-26
Prophile

enebo:
This is beginnings of a callsite-based profiler.  It is enabled with --ir.profiler.  You can try to enable it but it stlll is a work in progress.  The main changes to notice:
 
 1. All instrs which are calls logically extend from CallBase (including removing BacktickInstr to just be an ordinary call).
 1. CallBase is pushed around call/invoke methods within JVMVisitor and BytecodeAdapters.  This was possible from previous change and makes it easier for us to stuff more info into callbase later on.  I needed it for hasLiteralClosure() info deep in bytecode adapter
  1. ProfiledCallSite is what both interps and JIT uses.  Invokedynamic is done differently and is not hooked up yet.

Additional high level design goals.  Be as incremental as possible.  Original profiling work has a global profiler but this is just simple counter on special callsite type.  We can later have callsites coordinate with a global profiler.  Only normal calls are supported but in same spirit we can easily add FCALLS.  Not all method targets are allowed.   Some of these will be able to be incrementally supported later....

An experiment on the experiment is that native implementations can be replaced by pure-ruby versions so we can still inline in common core APIs which accept blocks.


core/src/main/java/org/jruby/RubyInstanceConfig.java
```diff
@@ -1852,6 +1852,9 @@ public boolean shouldPrecompileAll() {
     public static boolean IR_COMPILER_DEBUG = Options.IR_COMPILER_DEBUG.load();
     public static boolean IR_WRITING = Options.IR_WRITING.load();
     public static boolean IR_READING = Options.IR_READING.load();
+    public static boolean IR_INLINER = Options.IR_INLINER.load();
+    public static int IR_INLINER_THRESHOLD = Options.IR_INLINER_THRESHOLD.load();
+    public static boolean IR_INLINER_VERBOSE = Options.IR_INLINER_VERBOSE.load();
```
headius: You could just move these closer to where you use them...I think we want to get away from the pattern of static fields on RubyInstanceConfig for every little thing.

core/src/main/java/org/jruby/ir/IRManager.java
```diff
@@ -299,4 +311,34 @@ protected void optimizeIfSimpleScope(IRScope scope) {
...
     }
+
+    // FIXME: needs info for specialized method selection.
+    // FIXME: should allow non-classpath loading for easier debugging and hacking.
+    public IRMethod loadInternalMethod(ThreadContext context, IRubyObject self, String method) {
+        try {
+            RubyModule type = self.getMetaClass();
+            String fileName = "classpath:/jruby/ruby_implementations/" + type + "/" + method + ".rb";
+            FileResource file = JRubyFile.createResourceAsFile(context.runtime, fileName);
+            InputStream stream = file.openInputStream();
```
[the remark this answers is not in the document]
enebo: @headius will do post merge.  This can only happen via Integer#times and still needs cleanup.  but I will fix that fairly soon.

core/src/main/java/org/jruby/RubyInstanceConfig.java
enebo: @headius ok I will change this quick...although not entirely sure where I will put them maybe in IRManager

core/src/main/java/org/jruby/ir/IRManager.java
```diff
@@ -299,4 +311,34 @@ protected void optimizeIfSimpleScope(IRScope scope) {
...
+            } else {
+                containingScope = new IRClassBody(this, script, scopeNode.getCPath().getName(), 0, scopeNode.getScope());
+            }
+            IRMethod newMethod = new IRMethod(this, containingScope, defNode, context.runtime.newSymbol(method), true, 0, defNode.getScope(), false);
+
+            newMethod.prepareForCompilation();
+
+            return newMethod;
+        } catch (IOException e) {
+            e.printStackTrace();
```
kares: :smiling_imp:

kares: ah, there's already a regarding comment from @headius

core/src/main/java/org/jruby/ir/instructions/YieldInstr.java
```diff
@@ -17,15 +17,17 @@
...
     public final boolean unwrapArray;
+    private long callSiteId;
 
     public YieldInstr(Variable result, Operand block, Operand arg, boolean unwrapArray) {
         super(Operation.YIELD, result, block, arg == null ? UndefinedValue.UNDEFINED : arg);
 
         assert result != null: "YieldInstr result is null";
 
         this.unwrapArray = unwrapArray;
+        this.callSiteId = CallBase.callSiteCounter++;
```
kares: it doesn't matter if some end up with the same id (under concurrent execution), right?

core/src/main/java/org/jruby/ir/representations/BasicBlock.java
```diff
@@ -115,24 +119,43 @@ public boolean isEmpty() {
...
-    // merged (this is only for inlining and inlining does not work on master currently).
-    public BasicBlock splitAtInstruction(Instr splitPoint, Label newLabel, boolean includeSplitPointInstr) {
+    /**
+     * What site object contains this callsiteId or die trying.
+     * @param callsiteId to be found
+     * @return the Site instance (CallBase or YieldInstr)
+     */
+    public Site siteOf(long callsiteId) {
+        for (Instr instr: instrs) {
+            if (instr instanceof Site && ((Site) instr).getCallSiteId() == callsiteId) return (Site) instr;
```
kares: hmm, seems (above) might matter - better of using an AtomicLong instead of long++

core/src/main/java/org/jruby/ir/transformations/inlining/CFGInliner.java
```diff
@@ -15,12 +16,14 @@
 import java.util.List;
 
 public class CFGInliner {
-    private static final boolean debug = false;
+    private static final boolean debug = true;
```
kares: let's flip this back to `false` before :ship:

core/src/main/ruby/jruby/ruby_implementations/Integer/times.rb
```diff
@@ -0,0 +1,9 @@
+class Integer
```
kares: maybe at some point the *ruby_implementations* name should make it clear that its only used when inlining?

enebo: yeah we can change this name at any point.   I am not sure if this will eventually only be for inlining though.  Another thought about the ruby implementations is they can be specialized implementations if we know exactly how they will be used.  In that case, this could be a much more general location?

core/src/main/java/org/jruby/ir/transformations/inlining/CFGInliner.java
enebo: Thanks for pointing this out.  I still am happy to see this output but I should use Log and hide behind the pref.  Just forgot to change this over. I will do that before merging.

core/src/main/java/org/jruby/ir/instructions/YieldInstr.java
enebo: For today, this is not an issue because the conflict needs to be within the same scope we are deciding to inline on.  So it is more a tuple of {IRScope, callsiteId} as identity of the callsite we are interested in.  We cannot parallel build the same scope and we also have a limitation on multiple inlinings happening.  When we do support multiple inlinings we should revisit this (this scenario I suspect would be rare as well as both scopes would need to inline together then inline again -- 3 inlines where 2 methods were built at same moment).  I did not add this originally so this is largely just a continuation of what we do in CallBase.  You do make me realize when we get to recursive inlining we need to be aware we will have potentially n identical calls.  Even with atomic updates we need to cope with uniqueness.

PR #5385, opened 2018-10-27
Update specs

eregon:


eregon:
Specs are green now, I will merge.
Note that I added quite a few tags, mostly related to tainting in pack/unpack:
https://github.com/jruby/jruby/pull/5385/files#diff-eae6077a29027b59086379dea9555107

Issue #5387, opened 2018-10-30
Updated jruby-readline alters terminal settings, breaking interactive prompts


headius:
Something that jirb boots on a Linux or OS X system appears to be altering the tty settings.

```
[] ~/projects/jruby $ stty
speed 9600 baud;
lflags: echoe echoke echoctl pendin
iflags: iutf8
oflags: -oxtabs
cflags: cs8 -parenb

[] ~/projects/jruby $ jirb
irb(main):001:0> exit

[] ~/projects/jruby $ stty
speed 9600 baud;
lflags: -icanon echoe echoke echoctl
iflags: -icrnl -ixon iutf8
oflags: -oxtabs
cflags: cs8 -parenb
dsusp   lnext   
<undef> <undef>
```

Note that after running jirb, several tty settings change including -icrnl, which breaks most interactive prompts.

headius:
Booting jirb with `--noreadline` appears to be ok. It looks like recent updates to jruby-readline may be to blame. @kares perhaps you can try to sort out which of your changes over the summer caused this?

headius:
I have reverted the update to jruby-readline and we can use this issue to track the process of fixing it, hopefully for 9.2.2.

kares:
oh well, the jruby-readline update mostly updated the underlying readline library.
which seems to have been stucked with an old version, seemed to bring in some bug fixes and actually fixed some issues for me locally (linux). we shall look into whether a more recent update fixes the tty changes.

kares:
turns out jline 2.x isn't supported anymore - jruby-readline 1.3.0 was using latest version (2.14.6)

kares:
are you sure previous version doesn't alter settings such as `-icrnl` since the [code looks the same](https://github.com/jline/jline2/commit/07cc3b8a10c9504a0d8b2661f03c6ad19116a486).
... has been in since jline 2.8 (jruby-readline 1.2 uses 2.11 https://github.com/jruby/jruby-readline/commit/177d726b6b591db5c78adc01b06230df4dc1668a)

enebo:
@kares I just tried with master and I do not see stty being changed after using irb.  .set() appears to call stty but perhaps they have some finalization or later stty which unsets that?  Or maybe they added some explicit finalization we are not doing?

kares:
> Or maybe they added some explicit finalization we are not doing?

thanks - did not try myself. that might do it - will try their terminal tear-down logic.

headius:
[...]

PR #5388, opened 2018-10-30
Revert updated jruby-readline

headius:
This reverts commit e1d02a082998bed5b5167f0950dc3ef4281ad3a2, reversing
changes made to b43b9d86088e842ccb447e3068df425c3ff32590.

This reverts #5225.

The updated jruby-readline appears to be the cause of #5387 so I think we'll need to revert this for 9.2.1 and figure out what went wrong.


headius:
#5225 should be retargeted to "Invalid" if we decide to revert this update for 9.2.1.

PR #5393, opened 2018-10-31
Fix Hash#shift when at capacity

### LINKED ISSUE #5392: Hash#shift fails at capacity boundaries
The `Hash#shift` method appears to break at capacity boundaries, such as the default size of 8:

```ruby
h = {}
8.times {|j| h[j] = j}
p h
h.shift
```

This produces an ArrayIndexOutOfBoundsException:

```
Unhandled Java exception: java.lang.ArrayIndexOutOfBoundsException: 16
java.lang.ArrayIndexOutOfBoundsException: 16
                shift at org/jruby/RubyHash.java:1921
                 call at org/jruby/RubyHash$INVOKER$i$0$0$shift.gen:-1
         cacheAndCall at org/jruby/runtime/callsite/CachingCallSite.java:323
                 call at org/jruby/runtime/callsite/CachingCallSite.java:139
   invokeOther5:shift at -e:1
               <main> at -e:1
                 load at org/jruby/ir/Compiler.java:94
            runScript at org/jruby/Ruby.java:852
          runNormally at org/jruby/Ruby.java:771
          runNormally at org/jruby/Ruby.java:789
          runFromMain at org/jruby/Ruby.java:601
        doRunFromMain at org/jruby/Main.java:415
          internalRun at org/jruby/Main.java:307
                  run at org/jruby/Main.java:234
                 main at org/jruby/Main.java:206
```

The cause appears to be lack of a guard for the `end` value being just beyond the end of the current `entries` store.

Fix coming.

headius:
This fixes #5392.

@ChrisBr Please review this fix! I was not sure if this is the best way to proceed. Other methods that mutate the Hash usually call checkResize, which avoids the issue. Since we're shifting it should not need to grow, so I added the additional length check instead.

### DIFF
--- a/core/src/main/java/org/jruby/RubyHash.java
+++ b/core/src/main/java/org/jruby/RubyHash.java
@@ -1916,6 +1916,10 @@ public class RubyHash extends RubyObject implements Map {
         modify();
         IRubyObject key, value;
 
+        int start = this.start;
+        int end = this.end;
+        IRubyObject[] entries = this.entries;
+
         key = entries[start * NUMBER_OF_ENTRIES];
         value = entries[(start * NUMBER_OF_ENTRIES) + 1];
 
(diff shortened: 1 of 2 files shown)


ChrisBr:
Yeah, that makes sense! The patch LGTM, thanks for taking care @headius :+1:

Issue #287, opened 2012-09-05
Class does not inherit singleton_class's class methods


kachick:
I have faced a different behavior with MRI/YARV.
Is this expected?
## set up

``` ruby
class Foo
  class << self
    class << self
      def foo; end
    end
  end
end

class Bar < Foo; end

Bar.singleton_class.foo
```
## MRI/YARV

ruby 1.9.3p194 (2012-04-20 revision 35410) [i686-linux]

``` ruby
Bar.singleton_class.foo #=> nil
```
## JRuby

jruby 1.7.0.preview2 (1.9.3p203) 2012-09-03 b37a0bd on Java HotSpot(TM) Server VM 1.7.0_07-b10 [linux-i386]

``` ruby
Bar.singleton_class.foo #=> NoMethodError: undefined method `foo' for #<Class:Bar>
```

rjnienaber:
This still seems to be an issue in JRuby 1.7.20 and JRuby HEAD:

```
----------
ruby 1.9.3p551 (2014-11-13 revision 48407) [x86_64-darwin14.0.0]
nil
----------
jruby 1.7.20 (1.9.3p551) 2015-05-04 3086e6a on Java HotSpot(TM) 64-Bit Server VM 1.8.0_05-b13 +jit [darwin-x86_64]
NoMethodError: undefined method `foo' for #<Class:Bar>
  (root) at /tmp/test.rb:11
==========
ruby 2.2.2p95 (2015-04-13 revision 50295) [x86_64-darwin14]
nil
----------
jruby 9.0.0.0-SNAPSHOT (2.2.2) 2015-05-05 c0d9bc3 Java HotSpot(TM) 64-Bit Server VM 25.5-b02 on 1.8.0_05-b13 +jit [darwin-x86_64]
NoMethodError: undefined method `foo' for #<Class:Bar>
  method_missing at org/jruby/RubyBasicObject.java:1606
           <top> at /tmp/test.rb:11
----------
```

rcrews:
Just ran the above test case with 9.1.16.0. Can this be closed?
```
15:48:59 ~ 🐾  irb
jruby-9.1.16.0 :001 > class Foo
jruby-9.1.16.0 :002?>     class << self
jruby-9.1.16.0 :003?>         class << self
jruby-9.1.16.0 :004?>             def foo; end
jruby-9.1.16.0 :005?>           end
jruby-9.1.16.0 :006?>       end
jruby-9.1.16.0 :007?>   end
[...]

PR #5394, opened 2018-10-31
[fix] singleton meta class

kares:
JRuby seem to never have `Foo.singleton_class.singleton_class` working as expected
... mostly following super-class hierarchy with singleton classes

this gets the class hierarchy working, not sure what is left to be done for module's singleton
left the "super meta class" logic as it was for now

fixes GH-287 ... an old timer :)

### DIFF
--- a/core/src/main/java/org/jruby/MetaClass.java
+++ b/core/src/main/java/org/jruby/MetaClass.java
@@ -33,8 +33,9 @@ package org.jruby;
 import org.jruby.runtime.builtin.IRubyObject;
 
 public final class MetaClass extends RubyClass {
-    /** rb_class_boot (for MetaClasses) (in makeMetaClass(RubyClass))
-     * 
+
+    /**
+     * rb_class_boot for meta classes ({@link #makeMetaClass(RubyClass)})
      */
     public MetaClass(Ruby runtime, RubyClass superClass, IRubyObject attached) {
         super(runtime, superClass, false);
@@ -43,12 +44,34 @@ public final class MetaClass extends RubyClass {
     }
 
     @Override
+    public final IRubyObject allocate(){
+        throw runtime.newTypeError("can't create instance of virtual class");
+    }
+
+    @Override
     public boolean isSingleton() {
         return true;
     }
 
-    public final IRubyObject allocate(){
-        throw runtime.newTypeError("can't create instance of virtual class");
+    /**
+     * rb_make_metaclass
+     * @param superClass
+     * @return meta-class
+     */
+    @Override
+    public RubyClass makeMetaClass(RubyClass superClass) {
+        MetaClass klass = new MetaClass(runtime, superClass, this); // rb_class_boot
+        setMetaClass(klass);
+
+        klass.setMetaClass(klass);
+        klass.setSuperClass(getSuperClass().getRealClass().getMetaClass());
+
+        return klass;
+    }
+
+    @Override
+    RubyClass initMetaClass(RubyBasicObject target) {
+        return attached == target ? this : super.initMetaClass(target);
     }
 
     public IRubyObject getAttached() {
@@ -59,5 +82,6 @@ public final class MetaClass extends RubyClass {
         this.attached = attached;
     }
 
-    private IRubyObject attached = null;
+    private IRubyObject attached;
+
 }
--- a/core/src/main/java/org/jruby/MetaClass.java
+++ b/core/src/main/java/org/jruby/MetaClass.java
@@ -34,10 +34,15 @@ import org.jruby.runtime.builtin.IRubyObject;
 
 public final class MetaClass extends RubyClass {
 
+    @Deprecated
+    public MetaClass(Ruby runtime, RubyClass superClass, IRubyObject attached) {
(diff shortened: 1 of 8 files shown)


core/src/main/java/org/jruby/MetaClass.java
```diff
@@ -33,31 +33,75 @@
...
+
+        return getSuperClass().getRealClass().getMetaClass(); // NOTE: is this correct?
+    }
+
+    @Override
+    RubyClass toSingletonClass(RubyBasicObject target) {
+        return attached == target ? this : super.toSingletonClass(target);
+    }
+
+    public RubyBasicObject getAttached() {
```
kares: ah, so I initially put in a `RubyClass` type there since the **attached** wasn't clear to be (almost) always a class.
not a must, but this is pretty much internal API, right? ext should pretty much see `RubyClass` as all there is ...

kares: its pretty much assumed to be of a RubyBasicObject

enebo: @kares I was wondering about Module#initialize_copy.  It seems that can be RubyModule and not RubyClass.  Does that work here? like:

 ```ruby
module Foo; end; p Foo.send(:initialize_copy, Foo).singleton_class
```

As a broader question tracing through all this logic I half wonder whether rubymodule or rubyclass if we can narrow this more than basicobject (I see undef is basicobject but we could make that module/class maybe?).

Comment (kares) [reply core/src/main/java/org/jruby/MetaClass.java]: yeah, that "sinleton-class-clone-and-attache" code I wasn't sure ... 
didn't touch it - was hoping to get back to it, seems to be working : 
```
kares@clevo:~/workspace/oss/jruby$ bin/jruby -e "module Foo; end; p Foo.send(:initialize_copy, Foo).singleton_class"
#<Class:Foo>
```
```
kares@clevo:~/workspace/oss/jruby$ bin/jruby -e "module Foo; end; p Foo.send(:initialize_copy, Foo).singleton_class.singleton_class"
#<Class:#<Class:Foo>>
```
... super-class of such also seems fine (working as MRI) : 
```
kares@clevo:~/workspace/oss/jruby$ bin/jruby -e "module Foo; end; p Foo.send(:initialize_copy, Foo).singleton_class.singleton_class.superclass"
#<Class:Module>
```

at this point I am guessing only class singleton_class-es were really broken - which is what has been handled, thus hopefully this is the right direction.

Comment (kares) [reply core/src/main/java/org/jruby/MetaClass.java]: > half wonder whether rubymodule or rubyclass if we can narrow this more than basicobject (I see undef is basicobject but we could make that module/class maybe?)

yeah - except `UNDEF` that it should always be a `RubyModule` (or `RubyClass/MetaClass`)
did not try not using `UNDEF` ... it might work with passing `null` around :see_no_evil: 

and if its to be a "public" API - maybe smt to go along in `RubyClass`' `isSingleton()` to retrieve the "attached" class (not sure about naming) ... but that is mostly just throwing in random ideas :)

Comment (enebo) [reply core/src/main/java/org/jruby/MetaClass.java]: @kares yeah UNDEF I don't know how that is used in this context either.  Maybe null works.  At some point we probably want to more formally restrict what attached means and what it can be.  RubyBasicObject pretty only eliminates a few things I don't think it could ever be already.

Comment (kares) [reply core/src/main/java/org/jruby/MetaClass.java]: actually, it might be fine as is ... was concentrating too much on class's singleton_class
in case of a object ... attaches the bare (BasicObject) object for `object.singleton_class`

that `null` case (instead of UNDEF) still seems legit, as there's some (old) TODO notes down the `getAttached()` path of dealing with NPEs ... which UNDEF will throw as it gets used.

Comment (enebo) [reply core/src/main/java/org/jruby/MetaClass.java]: @kares I thought it might just attach for object singletons but I did not see which code path did that so I was confused.  I think perhaps even thinking that the name 'attached' probably does not best reflect what the field is for.  So another thing to ponder...a better name.

Comment (kares) [reply core/src/main/java/org/jruby/MetaClass.java]: open to suggestions here - already biased as got used to the attached name :)
anyway, should be good to go as is, we could iterate some more over later ...
okay for 9.2.1 or should we up hold till release?

Comment (enebo) [reply core/src/main/java/org/jruby/MetaClass.java]: Yeah my comment was not directed to this PR really.  I just realized attached is probably the most minimal description possible.  'singletonFor'?  Anyways I think we can land this.  @headius does this worry you at all?

File: test/jruby/test_class.rb [hunk 2]
```diff
@@ -337,4 +337,60 @@ def test_const_set
...
+
+    methods = S2.singleton_class.methods
+    #assert_include methods, :foo
+    #assert_include methods, :bar
+
+    methods = Class.new(S1).singleton_class.methods
+    #assert_include methods, :foo
+    #assert_include methods, :bar
+  end
+
```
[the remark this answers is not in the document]
headius: There's a couple other tests that pass from MRI, but unsure how many have any relation to this.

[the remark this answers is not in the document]
enebo: @headius untag/unexclude if you know what they are since it appears you figured that out

Issue #3656, opened 2016-02-10
Autoload is not overwritten by subsequent explicit require


headius:
I'm trying to fix `TestAutoload#test_require_implemented_in_ruby_is_called` from the MRI suite. The patch I started with is below.

``` diff
diff --git a/core/src/main/java/org/jruby/runtime/load/LoadService.java b/core/src/main/java/org/jruby/runtime/load/LoadService.java
index a76e457..4b58e6e 100644
--- a/core/src/main/java/org/jruby/runtime/load/LoadService.java
+++ b/core/src/main/java/org/jruby/runtime/load/LoadService.java
@@ -396,7 +396,7 @@ public class LoadService {
     }

     public boolean autoloadRequire(String requireName) {
-        return requireCommon(requireName, false) != RequireState.CIRCULAR;
+        return runtime.getTopSelf().callMethod(runtime.getCurrentContext(), "require", runtime.newString(requireName)).isTrue();
     }

     private enum RequireState {
```

This matches MRI's behavior and appears to fix the test, except for a little snag...another test starts to fail with a StackOverflowError:

```
[1/1] TestRequire#test_require_with_loaded_features_pop = 3.09 s
  1) Failure:
TestRequire#test_require_with_loaded_features_pop [/Users/headius/projects/jruby/test/mri/ruby/test_require.rb:680]:
[ruby-core:50645].

1. [1/2] Assertion for "stdout"
   | <[":ok"]> expected but was
   | <["nil", ":ok"]>.

2. [2/2] Assertion for "stderr"
   | <[]> expected but was
   | <["Exception in thread \"Ruby-0-Thread-1: -:1\" java.lang.StackOverflowError",
   |  "\tat java.lang.ClassLoader.findLoadedClass(ClassLoader.java:1033)",
   |  "\tat java.lang.ClassLoader.loadClass(ClassLoader.java:406)",
   |  "\tat java.lang.ClassLoader.loadClass(ClassLoader.java:357)",
   |  "\tat Users.headius.projects.jruby.lib.ruby.stdlib.rubygems.core_ext.kernel_require.RUBY$method$require$0(/Users/headius/projects/jruby/lib/ruby/stdlib/rubygems/core_ext/kernel_require.rb:52)",
   |  "\tat org.jruby.internal.runtime.methods.CompiledIRMethod.invokeExact(CompiledIRMethod.java:240)",
[...]

PR #5404, opened 2018-11-01
Make autoload dynamically dispatch to require. Fixes #5403.

### LINKED ISSUE #5403: `autoload` should delegate to `require`

Thanks!!

/cc @enebo

headius:
We'll see how this fares in CI.


headius:
Ping #5403.

tenderlove:
@headius I don't know if this helps, but it looks like the assertion is testing that there are no warnings and the warnings are coming from RubyGems (this is from the test failure output):

```
     [exec] TestRubyOptions#test_verbose [/home/travis/build/jruby/jruby/test/mri/ruby/test_rubyoptions.rb:100]:
     [exec] <[]> expected but was
     [exec] <["/home/travis/build/jruby/jruby/lib/ruby/stdlib/rubygems/version.rb:152: warning: loading in progress, circular require considered harmful - /home/travis/build/jruby/jruby/lib/ruby/stdlib/rubygems/version.rborg/jruby/RubyKernel.java:976:in `require'"
     [exec]  "/home/travis/build/jruby/jruby/lib/ruby/stdlib/rubygems/version.rb:152:in `<main>'"
     [exec]  "org/jruby/RubyKernel.java:976:in `require'"
     [exec]  "/home/travis/build/jruby/jruby/lib/ruby/stdlib/rubygems/specification.rb:1:in `<main>'"
     [exec]  "org/jruby/RubyKernel.java:976:in `require'"
     [exec]  "/home/travis/build/jruby/jruby/lib/ruby/stdlib/rubygems/specification.rb:10:in `<module:<main>>'"
     [exec]  "/home/travis/build/jruby/jruby/lib/ruby/stdlib/rubygems.rb:1:in `<main>'"
     [exec]  "org/jruby/RubyKernel.java:976:in `require'"
     [exec]  "/home/travis/build/jruby/jruby/lib/ruby/stdlib/rubygems.rb:1376:in `<main>'"
     [exec]  "org/jruby/RubyKernel.java:1000:in `load'"
     [exec]  "/home/travis/build/jruby/jruby/lib/ruby/stdlib/rubygems.rb:117:in `<main>'"     [exec] TestRubyOptions#test_verbose [/home/travis/build/jruby/jruby/test/mri/ruby/test_rubyoptions.rb:100]:
     [exec] <[]> expected but was
     [exec] <["/home/travis/build/jruby/jruby/lib/ruby/stdlib/rubygems/version.rb:152: warning: loading in progress, circular require considered harmful - /home/travis/build/jruby/jruby/lib/ruby/stdlib/rubygems/version.rborg/jruby/RubyKernel.java:976:in `require'"
     [exec]  "/home/travis/build/jruby/jruby/lib/ruby/stdlib/rubygems/version.rb:152:in `<main>'"
     [exec]  "org/jruby/RubyKernel.java:976:in `require'"
     [exec]  "/home/travis/build/jruby/jruby/lib/ruby/stdlib/rubygems/specification.rb:1:in `<main>'"
     [exec]  "org/jruby/RubyKernel.java:976:in `require'"
     [exec]  "/home/travis/build/jruby/jruby/lib/ruby/stdlib/rubygems/specification.rb:10:in `<module:<main>>'"
     [exec]  "/home/travis/build/jruby/jruby/lib/ruby/stdlib/rubygems.rb:1:in `<main>'"
     [exec]  "org/jruby/RubyKernel.java:976:in `require'"
     [exec]  "/home/travis/build/jruby/jruby/lib/ruby/stdlib/rubygems.rb:1376:in `<main>'"
     [exec]  "org/jruby/RubyKernel.java:1000:in `load'"
     [exec]  "/home/travis/build/jruby/jruby/lib/ruby/stdlib/rubygems.rb:117:in `<main>'"
```

I noticed this particular test comes from MRI.  [Here is the JRuby version](https://github.com/jruby/jruby/blob/7b14404fb3ded10b4d07820aff5a4a809f0ba6f1/test/mri/ruby/test_rubyoptions.rb#L99-L109), and [here is the MRI version](https://github.com/ruby/ruby/blob/797973bdf8b1c014e80414a77281bfa7271f99a9/test/ruby/test_rubyoptions.rb#L117-L131).  These look mostly the same, except for the MJIT thing.

However, they both call `assert_in_out_err`.  In upstream (MRI), that method will [disable RubyGems](https://github.com/ruby/ruby/blob/797973bdf8b1c014e80414a77281bfa7271f99a9/test/lib/test/unit/assertions.rb#L587), where the implementation in JRuby [does not](https://github.com/jruby/jruby/blob/7b14404fb3ded10b4d07820aff5a4a809f0ba6f1/test/mri/lib/test/unit/assertions.rb#L596-L598).

It could be that disabling RubyGems will fix this test (addressing the warning in RubyGems may have nothing to do with JRuby).

headius:
@tenderlove Thank you! I just circled back to this today, so I'll look into your theory.

headius:
It looks like that disabling of RubyGems didn't happen until this summer, but the tests that we're failing are certainly from long before that.

https://github.com/ruby/ruby/commit/2ac41230c4763e01687b2e5fd001219ab577236e

headius:
Ok, so the problem here is the same thing as #3656, where an explicit require of an autoloaded file does not remove that autoload from firing again. Here's a trivial reproduction:

atest/a.rb
```ruby
class A
  autoload :B, 'atest/b'
end

require 'atest/b'
```

atest/b.rb
```ruby
class A::B
end
```

The normal require starts the process of loading atest/b.rb. That file then attempts to (re)open A::B, which triggers the autoload detection. Autoload does not check if there's a require of that file already in progress before proceeding to dynamically invoke `require`, so it is detected and reported as a circular require.

I am working through CRuby logic for this. I believe what's missing is the `check_autoload_required`.

headius:
Yes, that's basically the issue. This all feeds back into us not having the same search and caching capabilities for load paths, expanded paths, and loaded features that MRI does.

I ported over the "loaded_feature_path" method and added a loop over all currently-loading scripts, as is done in the rb_feature_p function. CRuby checks for an existing feature in the following ways, if I'm following right:

* Check the actual loaded feature index for the requested feature. This maintains a mapping of short feature name to actual loaded path. We have this index but I'm not sure how well it works.
* Use the expanded load path and list of files currently being loaded to see if any combination matches. If load path entry + feature == currently-loading file, we consider the feature provided. I assume that require attempts to lock this path *before* assuming any existing lock is its own.

So basically, we're missing most of the search logic that MRI does during rb_feature_p because we don't walk through the whole load path (with pre-expanded paths) and look for the long name.

See also #2794.

@enebo It's time we bit the bullet and matched their searches, at least, which will likely necessitate incorporating the same caching.

headius:
Rebased atop master with the `--disable-gems` tweak. I believe based on @kares comments that there should be only one failure after this, and it may be something we can skip fixing if it's not critical (dispatching to require when the file is already being loaded on this thread).

enebo:
@headius yeah if we need this for rails 6 then matching MRI search logic will definitely be the fastest path to a stable release.

PR #5405, opened 2018-11-02
WIP: autoclose streams

ahorek:
outermost streams have to be closed, otherwise it'll cause a resource leak


ahorek:
it looks like it'll also close `$stderr` which is global and could be used later

PR #5407, opened 2018-11-03
normalized Float::MIN

ahorek:
https://ruby-doc.org/core-2.5.0/Float.html

The smallest positive normalized number in a double-precision floating point.

Usually defaults to 2.2250738585072014e-308.

If the platform supports denormalized numbers, there are numbers between zero and Float::MIN. 0.0.next_float returns the smallest positive floating point number including denormalized numbers.

```
    public static final double MIN_VALUE = 0x0.0000000000001P-1022; // 4.9e-324
    public static final double MIN_NORMAL = 0x1.0p-1022; // 2.2250738585072014E-308
```


kares:
oh nice, makes sense!

PR #5410, opened 2018-11-03
WIP: remove RubyRegexp#getClass

ahorek:
dead code?


kares:
no response from authors ... so let's kill it, shall we :smiling_imp:

headius:
I think these are null checks. It's a pattern I don't agree with, but `getClass` gets intrinsified to some fast check that's less expensive than doing `== null` I think. Obviously hard to know that without comments, though 🤔 

@ahorek @kares I think we should restore these with comments or just make them old-fashioned null checks.

kares:
in a way they were is just weird, most of them will fail with a `NPE` further down the path.
for those wout an explicit use (maybe one), maybe why not just have an `assert bytes != null` ?

ahorek:
@headius could you add a test case to confirm these checks are necessary? assert != null sounds better...

headius:
I went ahead with the asserts in 84be473f19.

@ahorek It may be possible to prove these never receive null but I figured we should put the checks back for now anyway. Obviously these have not been firing for normal JRuby users or we'd have seen the NPE bug reports. Asserts add to bytecode size but will not execute normally, so this seems ok. And these do need to be be non-null, so having the asserts is appropriate.

Issue #5406, opened 2018-11-02
Hash corrupts some data when working along linear/open threshold


headius:
This is a call for @ChrisBr to help us audit some of the RubyHash work in 9.2.1 😁 

In 6b2fa457a05d2cfa8e250b3c00dec7e227ec5e73, @enebo added a null check for a delete in open addressing mode. However it seems like the null check should not be necessary.

In order to get to this line in the code, there would have to be a bin that's not EMPTY and not DELETED. In such a case, there should be a non-null entry at that point in the entries array. However it comes up as null, as guarded by Tom's fix.

My theory for this involves transitioning back and forth across the size threshold for linear search versus open addressing. For example, the following scenario:

* Hash `myhash` is filled up to the threshold for linear search.
* An additional entry `A` transitions it to open addressing. The new entry gets a proper  value in `bins` but the other entries remain `EMPTY`. This seems wrong.
* A different entry `B` gets removed. Its bin gets marked as DELETED. We are now back in linear search size.
* The `A` entry gets deleted. Now that we are in linear addressing, the key and value are nulled but the bin is still set to a value that's not EMPTY or DELETED.
* Other adds transition us back to open addressing again.
* Now we are in open addressing with an entry in `bins` that is not EMPTY or DELETED but points at null values.

I'm not clear from reviewing the code how the transition across this boundary should be updating bins. The above scenario at least seems like a gap.

headius:
@enebo points out that simply removing an entry shouldn't transition us back to linear addressing, since it is a factor of the size of the `entries` array. But perhaps a rehash could cause us to shrink back down? And then we have a bunch of weird bin entries?

headius:
Hmm, perhaps it's not possible to transition back from open addressing to linear?

There are still questions for me about how the transition from linear to open works, since at that point we have a `bins` filled with `EMPTY`.

enebo:
I think at least a single case of this seems to be that sprockets memory cache (a hash) is getting/deleting concurrently across puma threads.  I see no synchronization.

ChrisBr:
> @enebo points out that simply removing an entry shouldn't transition us back to linear addressing, since it is a factor of the size of the entries array. But perhaps a rehash could cause us to shrink back down? And then we have a bunch of weird bin entries?

This is correct, your initial theory should not happen.

IIRC (did not check the code) we do not compact the hash when removing elements. However, rehash does compact the hash.

[...]

PR #5412, opened 2018-11-05
Temporarily revert open-addressing Hash for concurrency reasons.

headius:
While verifying JRuby 9.2.1, @enebo discovered that certain
concurrent uses of a simple Rails app were triggering intermittent
errors, and forcing the JIT to run very quickly produced NPEs
under concurrent load. In 6b2fa457 @enebo was able to fix this
by adding a null check, but we could not explain how a null could
appear at that point in the code unless the hash had been
corrupted, perhaps across threads.

Additional testing with a number of contrived scripts seems to
show that the new open-addressing hash (plus its linear mode) is
more fragile under concurrency than the old linked buckets
implementation. This is in a way, by design: the new
implementation reduces indirection and allocation by packing
keys and values into a contiguous area of memory, opening up
potential for more races against the same data.

In order to have a stable 9.2.1 release, we are temporarily
rolling back this change until we can get a better grasp of the
changes needed to make it as least as concurrency-friendly as the
old hash implementation, if not properly concurrency-safe (which
may or may not be easier in this new implementation).

### DIFF
--- a/core/src/main/java/org/jruby/RubyHash.java
+++ b/core/src/main/java/org/jruby/RubyHash.java
@@ -77,93 +77,35 @@ import static org.jruby.RubyEnumerator.enumeratorizeWithSize;
 import static org.jruby.runtime.Visibility.PRIVATE;
 import static org.jruby.RubyEnumerator.SizeFn;
 
-/* The original package implemented classic bucket-based hash tables
-   with entries doubly linked for an access by their insertion order.
-   To decrease pointer chasing and as a consequence to improve a data
-   locality the current implementation is based on storing entries in
-   an array and using hash tables with open addressing.  The current
-   entries are more compact in comparison with the original ones and
-   this also improves the data locality.
-   The hash table has two arrays called *bins* and *entries*.
-     bins:
-    -------
-   |       |                  entries array:
-   |-------|            --------------------------------
-   | index |           |      | entry:  |        |      |
-   |-------|           |      |         |        |      |
-   | ...   |           | ...  | hash    |  ...   | ...  |
-   |-------|           |      | key     |        |      |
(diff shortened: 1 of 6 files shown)


headius:
@ChrisBr We are going to delay the release of your open-addressing Hash until 9.2.2, so we can investigate how to make it a bit more robust under concurrent mutation. This is not an indictment of your work...the fact that the 9.2 Hash impl seems to survive more unsafe scenarios may be accidental, or simply due to the fact that it starts using a heavy bucket system immediately.

See discussions under #5406 about @enebo's discovery, but ignore my original assertions because they were based on a flawed understanding of the new Hash impl.

Here are a few examples of scripts that appear to work in 9.2 and fail in 9.2.1:

Leads to NPE in `shift` due to pulling a null key out of `entries`.
```ruby
h = {}
Thread.abort_on_exception = true
t1 = Thread.new { i = 0; loop { h[i+=1] = 1 } }
t2 = Thread.new { loop { h.shift } }
sleep
```

Leads to ArrayIndexOutOfBounds quickly:
```ruby
h = {}
Thread.abort_on_exception = true
loop {
  x = false
  t1 = Thread.new {|t| Thread.pass until x; i = 0; 10_000_000.times { h[i+=1] = Thread.current } }
  t2 = Thread.new { Thread.pass until x; i = 0.1; 10_000_000.times { h[i+=1] = Thread.current } }
  x = true
  t1.join
  t2.join
}
```

I have not included a third script that used `delete` instead of `shift` since it appeared to run without error on both Hash impls. There are likely other cases but these are key mutations (shrinking and growing) that appear to be more fragile with open addressing.

ChrisBr:
@headius hm ok! Any idea how we can make it (more) thread safe?

PR #5419, opened 2018-11-06
Reinstate open-addressing hash to work on concurrency improvements.

headius:
See #5412.

This is a work in progress as we improve @ChrisBr's open-addressing Hash impl for better concurrency.

### DIFF
--- a/core/src/main/java/org/jruby/RubyArray.java
+++ b/core/src/main/java/org/jruby/RubyArray.java
@@ -2044,7 +2044,7 @@ public class RubyArray<T extends IRubyObject> extends RubyObject implements List
 
         boolean useSmallHash = realLength <= 10;
 
-        RubyHash hash = useSmallHash ? RubyHash.newSmallHash(runtime) : RubyHash.newHash(runtime);
+        RubyHash hash = RubyHash.newHash(runtime);
 
         for (int i = 0; i < realLength; i++) {
             IRubyObject elt = eltInternal(i).checkArrayType();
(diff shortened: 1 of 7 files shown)


anup1710:
@headius - Are we planing to include in coming releases ?

ChrisBr:
That would be great, I'm still open for helping out 👍

headius:
@ChrisBr @anup1710 Unsure if we'll get to it for 9.2.7, or even 9.2.8, since priority right now is fixing some long-neglected areas (refinements, newer load/require logic) but this is still definitely in plan.

I'll see about cleaning up what I've done on the branch. It constitutes more of an experiment. The general idea is trying to reduce the number of fields we have to update when adding to the Hash. My approach seems to be working, generally...I split the two implementations (linear search and open addressed) as different classes, so switching between them is switching atomically between two object instances. However, this means an extra hope, and the two impls themselves don't have any more thread-safety than they did before (except perhaps for a bit safer handling of the int values, since I pack them into longs.

We have never managed to isolate the concurrency problems that @enebo found while trying to run rails with this Hash impl, but we suspect the problem is that the state changes happening for open addressing happen to be slightly "less threadsafe" than the already-not-threadsafe Hash on master. If we can mitigate that by simplifying, atomicizing, or otherwise synchronizing those state changes, we can land it.

This may mean we have to go full-on thread-safe Hash, ideally just using atomic updates (i.e. no locks).

We still very much want to land this!

Jrakesh:
@headius - Hope we are moving on track to make this live in 9.3.x.x
Am i being optimistic or far-fetched ?

ChrisBr:
Yeah, happy to continue working on this 👍 @headius any plans how to move forward?

headius:
This version went a bit too far in restructuring so I've rebooted this as #5967.

PR #5422, opened 2018-11-06
[WIP] Backport test runner commit

### LINKED ISSUE #5403: `autoload` should delegate to `require`

tenderlove:
This is just to see if the backport fixes #5403 and #5404

### DIFF
--- a/core/src/main/java/org/jruby/runtime/load/LoadService.java
+++ b/core/src/main/java/org/jruby/runtime/load/LoadService.java
@@ -406,7 +406,7 @@ public class LoadService {
     }
 
     public boolean autoloadRequire(String requireName) {
-        return smartLoadInternal(requireName, false) != RequireState.CIRCULAR;
+        return runtime.getTopSelf().callMethod(runtime.getCurrentContext(), "require", runtime.newString(requireName)).isTrue();
     }
 
     private enum RequireState {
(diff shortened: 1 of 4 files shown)


tenderlove:
Looks like the patch fixed the one build.  I'm not sure why the other one is failing, it seems to have a 0 exit status.

kares:
there seems to be this one spec failure left, that wasn't excluded previously:
```
     [exec] 1)
     [exec] An exception occurred during: Mock.verify_count
     [exec] Module#autoload does not call #require a second time and does not warn if already loading the same feature with #require FAILED
     [exec] Mock 'main' expected to receive require(:any_args) exactly 0 times
     [exec] but received it 1 times
     [exec] org/jruby/RubyArray.java:1789:in `each'
     [exec] org/jruby/RubyHash.java:1419:in `each'
     [exec] org/jruby/RubyBasicObject.java:2594:in `instance_eval'
     [exec] org/jruby/RubyEnumerable.java:1738:in `all?'
     [exec] org/jruby/RubyArray.java:1789:in `each'
     [exec] /home/travis/build/jruby/jruby/spec/ruby/core/module/autoload_spec.rb:16:in `<main>'
```

headius:
I actually merged the `--disable-gems` portion to master by hand. It kinda sorta fixes the autoload/require failures, but I'm not convinced that the logic is actually doing what it's supposed to. We should not do the require if the file is already being loaded on this same thread.

This might be good enough for now, though. The pre-check for autoload just keeps the require from being called, which avoids the circular require errors. Unfortunately on JRuby it also misses logic necessary for divining the correct path to avoid requiring in the first place, so this will have to be teased apart.

I will rebase the other PR atop the `--disable-gems` tweak and see how it looks. If we can get everything passing, we'll reprioritize #2794 and #3656 to rework how load and require are managed internally (maybe for 9.3, early next year, or a 9.2 update if we need work for Rails 6).

Issue #5420, opened 2018-11-06
jruby-dist-9.2.1.0-bin.tar.gz files is missing gem lib/exe folders


jeremyevans:
### Environment

`jruby 9.2.1.0 (2.5.0) 2018-11-06 7b14404 OpenJDK 64-Bit Server VM 25.172-b11 on 1.8.0_172-b11 +jit [OpenBSD-x86_64]`

### Expected Behavior

All gem lib/exe directories should be populated.  `jruby -S rake` should work.

### Actual Behavior

`tar ztf jruby-dist-9.2.1.0-bin.tar.gz | fgrep jruby-9.2.1.0/lib/ruby/gems/shared/gems` output:

```
jruby-9.2.1.0/lib/ruby/gems/shared/gems/net-telnet-0.1.1/LICENSE.txt
jruby-9.2.1.0/lib/ruby/gems/shared/gems/net-telnet-0.1.1/net-telnet.gemspec
jruby-9.2.1.0/lib/ruby/gems/shared/gems/net-telnet-0.1.1/Gemfile
jruby-9.2.1.0/lib/ruby/gems/shared/gems/net-telnet-0.1.1/Rakefile
jruby-9.2.1.0/lib/ruby/gems/shared/gems/net-telnet-0.1.1/README.md
jruby-9.2.1.0/lib/ruby/gems/shared/gems/net-telnet-0.1.1/.travis.yml
jruby-9.2.1.0/lib/ruby/gems/shared/gems/rake-12.3.0/Gemfile
jruby-9.2.1.0/lib/ruby/gems/shared/gems/rake-12.3.0/History.rdoc
jruby-9.2.1.0/lib/ruby/gems/shared/gems/rake-12.3.0/rake.gemspec
jruby-9.2.1.0/lib/ruby/gems/shared/gems/rake-12.3.0/Rakefile
jruby-9.2.1.0/lib/ruby/gems/shared/gems/rake-12.3.0/MIT-LICENSE
jruby-9.2.1.0/lib/ruby/gems/shared/gems/rake-12.3.0/CONTRIBUTING.rdoc
jruby-9.2.1.0/lib/ruby/gems/shared/gems/rake-12.3.0/README.rdoc
jruby-9.2.1.0/lib/ruby/gems/shared/gems/power_assert-1.1.1/BSDL
jruby-9.2.1.0/lib/ruby/gems/shared/gems/power_assert-1.1.1/Gemfile
[...]

PR #5424, opened 2018-11-07
An attempt at fixing #5420.

enebo:
More or less all default and installed gems will completely include themselves
into stdlib artifact which in turn will make them available when making the
jruby-dist package.

What makes me uncertain that this is right is the previous code went through
some pains to not include everything.  I am guessing there is something we don't
want and I am now including it?

Second part of this was unneeded but I removed the ImportedGem class and just
had both gem types use a nested array of arrays leveraging block destructuring.

### DIFF
--- a/lib/pom.rb
+++ b/lib/pom.rb
@@ -171,7 +171,7 @@ project 'JRuby Lib Setup' do
 
         log "copy gem content to #{stdlib_dir}"
         # assume default require_path
-        require_base = File.join( gems, "#{gemname}*", 'lib' )
+        require_base = File.join( gems, "#{gem_name}*", 'lib' )
         require_files = File.join( require_base, '*' )
 
         # copy in new ones and mark writable for future updates (e.g. minitest)
@@ -193,7 +193,7 @@ project 'JRuby Lib Setup' do
         end
 
         # copy bin files if the gem has any
-        bin = File.join( gems, "#{gemname}", 'bin' )
+        bin = File.join( gems, "#{gem_name}", 'bin' )
         if File.exists? bin
           Dir[ File.join( bin, '*' ) ].each do |f|
             log "copy to bin: #{File.basename( f )}"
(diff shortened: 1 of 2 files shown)


enebo:
I should add I just wrote this and looked at the resulting file jruby-stdlib-9.2.1.0.jar.  I definitely see important and needed files in this jar...but I do not know what else I have included.

lib/pom.rb
```diff
@@ -223,7 +214,7 @@ def installer.ensure_required_ruby_version_met; end
           specname = File.basename( specfile )
           log "copy to specifications/default: #{specname}"
 
-          spec = Gem::Package.new( Dir[ File.join( cache, "#{g.name}-#{version}*.gem" ) ].first ).spec
+          spec = Gem::Package.new( Dir[ File.join( cache, "#{gem_name}*.gem" ) ].first ).spec
```
kares: you might end up including more if cache is not empty e.g. `bin/jruby -S gem` install-ed a newer version

headius:
Link #5420.

lib/pom.rb
```diff
@@ -102,7 +92,7 @@ def initialize( name, version, default_spec = true )
     end
   end
 
-  default_gemnames = default_gems.collect { |g| g.name }
+  default_gemnames = default_gems.collect { |name, _| name }
```
headius:
```ruby
default_gems.collect(&:first)
```

lib/pom.rb
```diff
@@ -212,8 +203,8 @@ def installer.ensure_required_ruby_version_met; end
           end
         end
 
-        if g.default_spec
-          specfile_wildcard = "#{g.name}-#{version}*.gemspec"
+        if true # default_spec was always true for these?
```
headius: Why leave this `if` here?

enebo: yeah I did this last night and did not notice this until after I submitted but the intent looked like mkristian planned on this being run by more than default at some point.  I will remove though.

PR #5429, opened 2018-11-07
Attempt to fix #5421 windows oddity where some times a rename during bundle…

### LINKED ISSUE #5421: bundler on Windows - permission denied
### Environment
jruby 9.2.1.0 (2.5.0) 2018-11-06 7b14404 Java HotSpot(TM) 64-Bit Server VM 25.181-b13 on 1.8.0_181-b13 +jit [mswin32-x86_64]

### Expected Behavior
unfortunatelly https://github.com/jruby/jruby/commit/7b14404fb3ded10b4d07820aff5a4a809f0ba6f1 breaks bundler on windows, sorry I didn't catch it earlier!

### Actual Behavior
bundle update
Fetching https://github.com/rails/rails
The dependency byebug (>= 0) will be unused by any of the platforms Bundler is installing for. Bundler is installing for java but the dependency is only for ruby. To add those platforms to the bundle, run `bundle lock --add-platform ruby`.
There was an error while trying to write to
`C:/Users/pdaho/.bundle/cache/compact_index/rubygems.org.443.29b0360b937aa4d161703e6160654e47/versions`.
It is likely that you need to grant write permissions for that path.

enebo:
… install would fail.

This I believe fixes #5421.

The fix is I think if we depend on Files.move to do it's thing we should not
try and rename a second time using the old way.  This happened to never
manifest on linux or macos because bundler stores files in another FS so
we never hit this code path.

### DIFF
--- a/core/src/main/java/org/jruby/RubyFile.java
+++ b/core/src/main/java/org/jruby/RubyFile.java
@@ -1109,15 +1109,10 @@ public class RubyFile extends RubyIO implements EncodingCapable {
         Path destPath = Paths.get(dest.getAbsolutePath());
         try {
             Files.move(oldPath, destPath, StandardCopyOption.ATOMIC_MOVE);
+            return RubyFixnum.zero(runtime);
         } catch (IOException ioe) {
             throw Helpers.newIOErrorFromException(runtime, ioe);
         }
-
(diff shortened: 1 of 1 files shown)


ahorek:
LGTM, confirmed locally

PR #5432, opened 2018-11-09
Add a CI job for the "sequel" gem

headius:
This PR adds a job based on the CI runs for @jeremyevans Sequel library. I've basically just duplicated his .travis.yml in a shell script that's run along with our other jobs.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -44,48 +44,49 @@ jobs:
   fast_finish: true
 
   include:
-    - stage: test
-      env: PHASE='-Prake -Dtask=test:mri:core:int'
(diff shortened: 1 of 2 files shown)


headius:
Note this should be squashed when merged; there's a lot of fiddling around in there.

headius:
I've added in the remaining environment setup and reduced clone depth to 10.

@jeremyevans This is currently running against your master HEAD. Is there a stable branch or tag you'd recommend we use instead? Maybe we query rubygems.org for latest release?

headius:
@jeremyevans I think we'll just cross that bridge when we come to it. Thanks!

PR #5433, opened 2018-11-09
update jruby-readline de-ja-vu (for JRuby::Util.load_ext)

kares:
on top of GH-5225 fixes restoring stty after runtime tear-down

resolves GH-5387


enebo:
@kares land on monday or tuesday.  I think we made it past needing more quick point releases but I am not holding my breath.  I don't actually think this is a risk either...but at this point the smaller the delta the less chance there is to need another point release.

PR #5441, opened 2018-11-12
Push signal traps and at_exit blocks to a common thread for exec.

### LINKED ISSUE #5437: Using at_exit and Signal.trap possibly creates a race-condition
### Environment

Provide at least:
* JRuby version: `jruby 9.2.0.0 (2.5.0) 2018-05-24 81156a8 Java HotSpot(TM) 64-Bit Server VM 25.181-b13 on 1.8.0_181-b13 +jit [darwin-x86_64]`
* Operating system and platform (e.g. `uname -a`): `Darwin MB-ADPENTELA.local 17.4.0 Darwin Kernel Version 17.4.0: Sun Dec 17 09:19:54 PST 2017; root:xnu-4570.41.2~1/RELEASE_X86_64 x86_64`
* Installed or activated gems
* Environment variables

### Expected Behavior
* The `Signal.trap` and `at_exit` hooks are executed in different threads leading to a race-condition where both `at_exit` and `Signal.trap` are defined.

Example: Puma traps `SIGINT` and sucker_punch uses the `at_exit` hook leading to a `SystemExit` before sucker_punch workers are shutdown.
https://github.com/puma/puma/issues/1675
https://github.com/brandonhilkert/sucker_punch/issues/223

### Actual Behavior
MRI is using the same thread for both `at_exit` and `Signal.trap`. It would be nice if there is some synchronization.

```
### test.rb
at_exit do
  sleep(1)
  puts "Thread in at_exit: #{Thread.current}"
end

Signal.trap("SIGINT") do
 puts "Thread in trap: #{Thread.current}"
 exit
end

sleep

## JRuby
$ RUBYOPT=-d jruby test_in_jruby.rb 
Exception `Gem::MissingSpecError' at /Users/adpentela/.rvm/rubies/jruby-9.2.0.0/lib/ruby/stdlib/rubygems/dependency.rb:310 - Gem::MissingSpecError
^CThread in trap: #<Thread:0x19b33233>
Exception `SystemExit' at org/jruby/RubyKernel.java:735 - exit
Thread in at_exit: #<Thread:0x37d31475>

## MRI
 $ ruby test_in_jruby.rb 
^CThread in trap: #<Thread:0x00007fd94d085c40 run>
Thread in at_exit: #<Thread:0x00007fd94d085c40 run>
```
Let me know if I'm doing something wrong or if this is a duplicate issue.
I had a quick glance at https://github.com/jruby/jruby/issues/4956 but not sure if this is related.

Thanks!

headius:
This avoids races due to the signal traps running on the JVM's
signal-handling thread while we run at_exit blocks on JRuby's
main thread.

Fixes #5437.

### DIFF
--- a/core/src/main/java/org/jruby/Ruby.java
+++ b/core/src/main/java/org/jruby/Ruby.java
@@ -3306,25 +3306,40 @@ public final class Ruby implements Constantizable {
             context.pushScope(new ManyVarsDynamicScope(topStaticScope, null));
         }
 
+        // use signal-handling thread to avoid races with trap(INT) and similar shutdown signals
+        ExecutorService executor = (ExecutorService) getModule("Signal").getInternalVariable("executor");
+
+        int[] _status = {status};
         while (!atExitBlocks.empty()) {
             RubyProc proc = atExitBlocks.pop();
-            // IRubyObject oldExc = context.runtime.getGlobalVariables().get("$!"); // Save $!
-            try {
-                proc.call(context, IRubyObject.NULL_ARRAY);
-            } catch (RaiseException rj) {
-                RubyException raisedException = rj.getException();
-                if (!getSystemExit().isInstance(raisedException)) {
-                    status = 1;
-                    printError(raisedException);
-                } else {
-                    IRubyObject statusObj = raisedException.callMethod(context, "status");
-                    if (statusObj != null && !statusObj.isNil()) {
-                        status = RubyNumeric.fix2int(statusObj);
+            executor.submit((Runnable) () -> {
+                // IRubyObject oldExc = context.runtime.getGlobalVariables().get("$!"); // Save $!
+                try {
+                    proc.call(context, IRubyObject.NULL_ARRAY);
+                } catch (RaiseException rj) {
+                    RubyException raisedException = rj.getException();
+                    if (!getSystemExit().isInstance(raisedException)) {
+                        _status[0] = 1;
+                        printError(raisedException);
+                    } else {
+                        IRubyObject statusObj = raisedException.callMethod(context, "status");
+                        if (statusObj != null && !statusObj.isNil()) {
+                            _status[0] = RubyNumeric.fix2int(statusObj);
+                        }
                     }
+                    // Reset $! now that rj has been handled
+                    // context.runtime.getGlobalVariables().set("$!", oldExc);
                 }
-                // Reset $! now that rj has been handled
-                // context.runtime.getGlobalVariables().set("$!", oldExc);
+            });
+        }
+
+        try {
+            executor.shutdown();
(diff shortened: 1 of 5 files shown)


headius:
@Adithya-copart Can you provide an example script that causes a race and verify that this PR fixes it?

Adithya-copart:
@headius Will the code in the first comment of https://github.com/puma/puma/issues/1675 work?

There is a race between the `at_exit` in [sucker_punch](https://github.com/brandonhilkert/sucker_punch/blob/d06ba290be4d56fbbd45d37f73c47bd441ad0224/lib/sucker_punch.rb#L53-L55) and the `trap` in [puma](https://github.com/puma/puma/blob/2668597ec1dd9546d83db9f2ec5ad092add483e6/lib/puma/launcher.rb#L407-L412).

I can recreate something later today if it doesn't.

Thanks!

headius:
@Adithya-copart Ah yes, that's quite similar to how this PR works, except that we use a JDK "ExecutorService" that manages the native thread and queueing for us.

I will try to reproduce based on your instructions.

There are failures in CI, unfortunately. It seems that at least the CRuby/MRI tests expect `at_exit` to run on the main thread, affecting the main thread's state, raising new exceptions in the main thread, and so on. And as it turns out, CRuby avoids the races you describe by running the signal handlers on the main thread as well. That's right, it hijacks the main thread to run signal handlers.

```
$ rvm ruby-2.5 do ruby -e 'Signal.trap("INT") { p Thread.current }; p Thread.current; sleep'
#<Thread:0x00007fafdb87dc40 run>
^C#<Thread:0x00007fafdb87dc40 run>
```

I'm not sure we want to emulate that behavior.

I have a thought for a modification to my fix that just waits for all scheduled signal handlers to execute before proceeding on to the `at_exit` blocks. We'll see if that helps CI.

core/src/main/java/org/jruby/Ruby.java
```diff
@@ -3306,6 +3306,17 @@ public void tearDown(boolean systemExit) {
             context.pushScope(new ManyVarsDynamicScope(topStaticScope, null));
         }
 
+        // terminate signal-handling thread to avoid races with at_exit
+        ExecutorService executor = (ExecutorService) getModule("Signal").getInternalVariable("executor");
+        try {
+            executor.shutdown();
+            if (!executor.awaitTermination(1, TimeUnit.HOURS)) {
+                throw newRuntimeError("failed to execute at_exit blocks within 1h timeout");
```
kares: heh :) ... is this an arbitrary 1h?

headius:
This had to be reverted due to some unexpected behavior changes in at_exit. See https://github.com/jruby/jruby/issues/5437#issuecomment-481033326.

PR #5444, opened 2018-11-12
fix an edge case in gcd calculation

ahorek:
fixes
```
-9223372036854775808.gcd(-9223372036854775808)
```
should return 9223372036854775808 not -9223372036854775808

MRI uses FIXNUM_P macro instead of (x instanceof RubyFixnum), so -9223372036854775808 is properly delegated to a bignum logic

### DIFF
--- a/core/src/main/java/org/jruby/util/Numeric.java
+++ b/core/src/main/java/org/jruby/util/Numeric.java
@@ -552,7 +552,7 @@ public class Numeric {
      *
      */
     public static IRubyObject f_gcd(ThreadContext context, IRubyObject x, IRubyObject y) {
-        if (x instanceof RubyFixnum && y instanceof RubyFixnum) {
+        if (x instanceof RubyFixnum && y instanceof RubyFixnum && ((RubyFixnum)x).getLongValue() != Long.MIN_VALUE) {
             return RubyFixnum.newFixnum(context.runtime, i_gcd(((RubyFixnum)x).getLongValue(), ((RubyFixnum)y).getLongValue()));
         }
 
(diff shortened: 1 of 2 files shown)


headius:
Can we reduce the duplication here and introduce an equivalent to FIXNUM_P?

ahorek:
FIXNUM_P macro means a positive number in range 0...FIXNUM_MAX
but our i_gcd can handle negative numbers except both (x == Long.MIN_VALUE && y == Long.MIN_VALUE)
https://github.com/jruby/jruby/blob/master/core/src/main/java/org/jruby/util/Numeric.java#L514

any idea for a better name?

headius:
Hmm it occurs to me now this is a more general overflow problem, isn't it? Won't some values overflow into Long.MAX_VALUE+1 or +2 or so on?

headius:
Actually now see that you're covering the specific case where "x" is MIN_VALUE. I don't have any better suggestion for this. I think we'll merge what you have and then definitely need to put together some test cases.

ahorek:
Long.MIN_VALUE is -9223372036854775808
but Long.MIN_VALUE is 9223372036854775807
so `gcd(Long.MIN_VALUE, Long.MIN_VALUE)` should be 9223372036854775808, but it can't be represented as long

see the related discussion about this with @kares
https://github.com/jruby/jruby/pull/5364#discussion_r224950093

mri is unaffected because it doesn't allow to pass negative numbers to i_gcd at all, but their implementation of i_gcd has the same limitation

headius:
@ahorek Ok I think we're on the same page now.

I tweaked your changes a bit and went with "isLongMinValue" for a utility function in 2c00b59eb5.

ahorek:
thanks

Issue #5450, opened 2018-11-15
java.lang.NoSuchMethodError: java.nio.ByteBuffer.clear()


ahorek:
### Environment
```
java version "1.8.0_191"
Java(TM) SE Runtime Environment (build 1.8.0_191-b12)
Java HotSpot(TM) 64-Bit Server VM (build 25.191-b12, mixed mode)
```

### Actual behaviour
`rvm get head`

raises an error
```
    Installing rvm gem in 1 gemsetsUnhandled Java exception: java.lang.NoSuchMethodError: java.nio.ByteBuffer.clear()Ljava/nio/ByteBuffer;
java.lang.NoSuchMethodError: java.nio.ByteBuffer.clear()Ljava/nio/ByteBuffer;
           encode at org/jruby/RubyEncoding.java:306
     doEncodeUTF8 at org/jruby/RubyEncoding.java:197
   encodeBytelist at org/jruby/RubyString.java:6191
           <init> at org/jruby/RubyString.java:377
        newString at org/jruby/RubyString.java:468
        newString at org/jruby/Ruby.java:3563
  createFileClass at org/jruby/RubyFile.java:219
         initCore at org/jruby/Ruby.java:1547
        bootstrap at org/jruby/Ruby.java:1318
             init at org/jruby/Ruby.java:1217
      newInstance at org/jruby/Ruby.java:362
      internalRun at org/jruby/Main.java:273
              run at org/jruby/Main.java:234
             main at org/jruby/Main.java:206
```

the problem is that I have a jruby's binary that was compiled with java 9, but executed with java 8.

### Expected behaviour
this should be a workaround
https://jira.mongodb.org/browse/JAVA-2559
https://github.com/jruby/jruby/blob/master/core/src/main/java/org/jruby/RubyEncoding.java#L306

kares:
should be easy to put up a PR ... mostly about: `((Buffer) buf).clear();` with a comment why its there.

headius:
Hmm that is harder to avoid ☹️ If we wrap this in another method we may interfere with the VM intrinsic used to optimize these methods.

These cases should be pretty limited, though...we can just cast everything to long in those few places.

I hope there aren't going to be too many more of these; we are not ready to start depending on Java 11. Maybe we need to look into the build using JDK8 runtime library.

headius:
[...]

PR #5451, opened 2018-11-15
improve cross compile compatibility

ahorek:
The Java 9 ByteBuffer and CharBuffer classes introduces overloaded methods with covariant return types for the following methods used by the driver:

* position
* limit
* flip
* clear
* mark

Without casting, exceptions like this are thrown when executing on Java 8:

    java.lang.NoSuchMethodError: java.nio.ByteBuffer.limit(I)Ljava/nio/ByteBuffer

fixes https://github.com/jruby/jruby/issues/5450

### DIFF
--- a/core/src/main/java/org/jruby/RubyEncoding.java
+++ b/core/src/main/java/org/jruby/RubyEncoding.java
@@ -209,7 +209,7 @@ public class RubyEncoding extends RubyObject implements Constantizable {
     }
 
     private static byte[] getBytes(final ByteBuffer buffer) {
-        byte[] bytes = new byte[buffer.limit()];
+        byte[] bytes = new byte[((Buffer) buffer).limit()];
         buffer.get(bytes);
         return bytes;
     }
@@ -223,7 +223,7 @@ public class RubyEncoding extends RubyObject implements Constantizable {
             bytes = getBytes(buffer);
             off = 0;
         }
-        return new ByteList(bytes, off, off + buffer.limit(), enc, false);
+        return new ByteList(bytes, off, off + ((Buffer) buffer).limit(), enc, false);
     }
 
     public static byte[] encodeUTF16(String str) {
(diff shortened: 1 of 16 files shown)


headius:
I hate that we have to do this but I guess there's not really any good choice. However, perhaps it would be better if we put the cast + call for each of these into a utility method? Helpers.clearBuffer(buffer) for example?

kares:
my understanding was that only those having a changed return are problematic - so this seems a bit much.
`ByteBuffer#limit(int)` has an issue but not `ByteBuffer#limit()` with an `int` return, same for `position()`

headius:
@kares My concern is having this cast + call pattern all over the place. We would end up having to document every one of these lines so nobody changes it back (most IDEs will show it as an unnecessary cast). If we move them into utility methods and call them that way, we only have to document it there.

And with an import static, "clearBuffer(buf)" is shorter than "((Buffer) buf).clear()" anyway.

headius:
FWIW, it's also easier for us when we start depending on Java 11 (maybe in 2020) because we can just have the IDE inline all calls to the method and eliminate the cast.

headius:
@ahorek Feel like revisiting these and making them call utility methods, perhaps in org.jruby.util.io.ChannelHelper?

ahorek:
> ByteBuffer#limit(int) has an issue but not ByteBuffer#limit() with an int return, same for position()
@kares - the problem is ByteBuffer#limit is overloaded on java 9, but on java 8 this overloaded call-site doesn't exist.

> Feel like revisiting these and making them call utility methods, perhaps in org.jruby.util.io.ChannelHelper
@headius - sounds good. I'll do that later today.

ahorek:
after some investigation I'll close this one if you don't mind

other methods like `java.lang.Math.multiplyExact` has the same problem and it'll be a big maintainance burden

it can be fixed by compiling the source in compatibility mode, otherwise the binary compatibility with older java versions isn't guaranteed
`base.java.version` and `base.javac.version`

https://stackoverflow.com/questions/51692748/java-compilation-source-target-and-release-supported-versions/51692749#51692749

headius:
@ahorek Actually that post is referring to just the Java code and bytecode resulting from compiling; source specifies the maximum level of Java features allowed and target specifies what minimum JVM the bytecode will work on. The fixes you were making were separate and would require always compiling against Java 8 classes, necessitating having two JDKs installed to build properly.

I think the cast+call fix is still valid but I will look at doing it with a utility method.

ahorek:
ok, reopened https://github.com/jruby/jruby/pull/5459

PR #5452, opened 2018-11-19
Use default sg ITALY with getChronology here.

### LINKED ISSUE #5418: Date.today does not appear to take the local TZ into account
### Environment

jRuby 9.2.0.0+ (may also happen on prior versions)
Tested on OS X and Linux.

### Expected Behavior

```
irb(main):001:0> Time.now
=> 2018-11-05 23:32:54 -0600
irb(main):002:0> Date.today
=> #<Date: 2018-11-05 ((2458428j,0s,0n),+0s,2299161j)>
```

### Actual Behavior

```
irb(main):001:0> Time.now
=> 2018-11-05 23:32:54 -0600
irb(main):002:0> Date.today
=> #<Date: 2018-11-06 ((2458428j,0s,0n),+0s,2299161j)>
```
What appears to be happening here is that since the DST shift occurred on 11/04 at 2am, after 11pm each day thereafter (presumably until DST begins again in March), `Date.today` returns the next day.  This does not happen if doing `Time.now.to_date`.

headius:
Fixes #5418

### DIFF
--- a/core/src/main/java/org/jruby/ext/date/RubyDate.java
+++ b/core/src/main/java/org/jruby/ext/date/RubyDate.java
@@ -713,7 +713,7 @@ public class RubyDate extends RubyObject {
 
     @JRubyMethod(meta = true)
     public static RubyDate today(ThreadContext context, IRubyObject self) { // sg=ITALY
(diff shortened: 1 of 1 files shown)


headius:
@ahorek Is there any way to test this without setting the local system time?

Issue #5448, opened 2018-11-14
RubyStruct.java gives ArgumentError: unknown keywords: keyword


dr-itz:
### Environment

* jruby 9.2.4.0 (2.5.0) 2018-11-13 8faff06 Java HotSpot(TM) 64-Bit Server VM 25.162-b12 on 1.8.0_162-b12 +jit [darwin-x86_64]
* Running AR-JDBC 52-stable tests

### Behavior

Running:

```
RAILS=../rails52 rake rails:test_sqlite3 TEST=test/cases/type/adapter_specific_registry_test.rb
```

(../rails52 is checked out at v5.2.1)

Gives:

```
  1) Error:
ActiveRecord::AdapterSpecificRegistryTest#test_construct_args_are_passed_to_the_type:
ArgumentError: unknown keywords: keyword
    org/jruby/RubyStruct.java:403:in `initialize'
    /Users/dritz/devel/open/rails/rails52/activemodel/lib/active_model/type/registry.rb:12:in `block in register'
    /Users/dritz/devel/open/rails/rails52/activerecord/lib/active_record/type/adapter_specific_registry.rb:36:in `call'
    /Users/dritz/devel/open/rails/rails52/activemodel/lib/active_model/type/registry.rb:20:in `lookup'
    /Users/dritz/devel/open/rails/rails52/activerecord/test/cases/type/adapter_specific_registry_test.rb:85:in `block in test_construct_args_are_passed_to_the_type'


  2) Error:
ActiveRecord::AdapterSpecificRegistryTest#test_registering_adapter_specific_modifiers:
ArgumentError: unknown keywords: keyword
    org/jruby/RubyStruct.java:403:in `initialize'
    /Users/dritz/devel/open/rails/rails52/activemodel/lib/active_model/type/registry.rb:12:in `block in register'
    /Users/dritz/devel/open/rails/rails52/activerecord/lib/active_record/type/adapter_specific_registry.rb:36:in `call'
    /Users/dritz/devel/open/rails/rails52/activemodel/lib/active_model/type/registry.rb:20:in `lookup'
    /Users/dritz/devel/open/rails/rails52/activerecord/lib/active_record/type/adapter_specific_registry.rb:105:in `call'
    /Users/dritz/devel/open/rails/rails52/activemodel/lib/active_model/type/registry.rb:20:in `lookup'
[...]

PR #5455, opened 2018-11-19
Fix multi-arity method binding.

headius:
We have apparently not been binding multi-arity rest methods well
for some time; the logic to look for "hasVarargs" was not using
the correct class.getName() for an IRubyObject[], resulting in
any variable-arity-or-rest methods always getting routed to the
rest path, because the min/max logic breaks without hasVarargs.

This could regress easily without tests, but I'm not sure of the best way to test that all this generated code is working right.

This was discovered while investigating #5448, which weirdly was caused by this bug to go down a rest path, and then via JavaMethodN.call re-route to a specific-arity version that had a bug.


headius:
Pushed an additional commit that fixes bugs that cropped up in RubyStruct keyword_init handling. Because specific arities started getting routed to the correct places, they hit RubyStruct initialize methods that did not have proper error handling in place for kwargs-init without kwargs.

PR #5458, opened 2018-11-20
Tempfile.open should have arity 3, to allow IO options through.

### LINKED ISSUE #5456: Tempfile.open does not recognize the 3rd argument (options to pass to File.open)
### Environment

* JRuby version: `jruby 9.2.4.0 (2.5.0) 2018-11-13 8faff06 Java HotSpot(TM) 64-Bit Server VM 25.152-b16 on 1.8.0_152-b16 +jit [darwin-x86_64]`
* Operating system and platform: `Darwin inquanok 17.7.0 Darwin Kernel Version 17.7.0: Wed Oct 10 23:06:14 PDT 2018; root:xnu-4570.71.13~1/RELEASE_X86_64 x86_64`

### Expected Behavior

With the following script: (note, the block is largely irrelevant, since jRuby never actually gets to it, but I wanted to show that we are writing, rewinding, setting encoding and then yielding it to a caller that reads from it.):

```ruby
#!/usr/bin/env ruby

require 'Tempfile'

# aws-sdk S3.get_object reads/writes in byte chunks which may or may align with UTF-8 codepoints
Tempfile.open('prefix_', '/tmp', encoding: 'ASCII-8BIT:ASCII-8BIT') do |file|
  file.puts $$
  file.rewind
  file.set_encoding('UTF-8')
  # yield file
  puts file.read
end
```

This works in RMI 2.5.1.

### Actual Behavior

It appears that jRuby 9.2.4.0 does not honor the 3rd argument (9.1.x.0 all did):

```
% RBENV_VERSION=2.5.1 ruby /tmp/tempfile.rb
30919
% RBENV_VERSION=jruby-9.2.4.0 ruby /tmp/tempfile.rb
ArgumentError: wrong number of arguments (3 for 2)
  <main> at /tmp/tempfile.rb:5
% RBENV_VERSION=jruby-9.1.17.0 ruby /tmp/tempfile.rb
32394
```

headius:
Fixes #5456.

In #5456, @erikogan shows that his example code worked in 9.1.17.0, but I could find no changes to Tempfile during the past 9 months that would explain the regression.

### DIFF
--- a/core/src/main/java/org/jruby/ext/tempfile/Tempfile.java
+++ b/core/src/main/java/org/jruby/ext/tempfile/Tempfile.java
@@ -250,7 +250,7 @@ public class Tempfile extends RubyFile implements Finalizable {
         return open(context, recv, args, block);
     }
 
-    @JRubyMethod(required = 1, optional = 1, meta = true)
+    @JRubyMethod(required = 1, optional = 2, meta = true)
     public static IRubyObject open(ThreadContext context, IRubyObject recv, IRubyObject[] args, Block block) {
         RubyClass klass = (RubyClass) recv;
         Tempfile tempfile = (Tempfile) klass.newInstance(context, args, block);


headius:
Bisecting was inconclusive 👎

PR #5459, opened 2018-11-20
cross compile compatibility

ahorek:
https://github.com/jruby/jruby/pull/5451


headius:
Looking good! I also started doing this and my coverage for `clear` seems to hit more sites than yours: https://gist.github.com/headius/08a568f5cf662ff207145e54cc0dbeb2

Other suggestions:

* We can make the methods all generic to with T extends Buffer and just cast the return type to T as seen in my version. That avoids the grey "unnecessary cast" squiggly in most IDEs.
* Do an IDE search for callers to these methods so we make sure we get every call site.

Otherwise this is just fine. As @enebo was saying to me, there will still be cases where we regress, but we could set up a JDK9 build + JDK8 test run in CI to help avoid that.

headius:
Thanks for the help @ahorek! I made my additional small changes in #5499 and merged to master.

PR #5462, opened 2018-11-21
add java 10, 11 opcodes

ahorek:
fixes warnings
WARNING: Illegal reflective access by org.jruby.util.SecurityHelper to field java.lang.reflect.Field.modifiers
WARNING: Please consider reporting this to the maintainers of org.jruby.util.SecurityHelper
WARNING: Use --illegal-access=warn to enable warnings of further illegal reflective access operations
WARNING: All illegal access operations will be denied in a future release

### DIFF
--- a/core/src/main/java/org/jruby/RubyInstanceConfig.java
+++ b/core/src/main/java/org/jruby/RubyInstanceConfig.java
@@ -1889,7 +1889,9 @@ public class RubyInstanceConfig {
             case "9" :
                 return Opcodes.V9;
             case "10" :
-                return Opcodes.V9; // TODO: switch when `V10 = 54` added
+                return Opcodes.V10;
+            case "11" :
+                return Opcodes.V11;
         }
     }
 


kares:
oh really, does it fix those warnings, like all of them? now, I am a bit confused, do you know why is that?
... anyway, this makes sense either way, I might have left that TODO there myself :)

ahorek:
I know, you changed the default version to 1.8 in https://github.com/jruby/jruby/commit/1b10d61bc8bc4a2d6403f3485b3f3b90ad1249c5#diff-f1a3566fb02f3f96df7df320d1a2a669

this condition now fails on java 10, 11
https://github.com/jruby/jruby/blob/4e8bb2666a7257f0f5986800f96bb88efdd6acbd/core/src/main/java/org/jruby/util/SecurityHelper.java#L22

kares:
right, that makes sense - guess its one warning at a time than :)
but there's a but which I we might not need to care about atm, still for the record:

`-Xbytecode.version` should be used to specify version of generated byte-code.

... so maybe the `SecurityHelper` should not rely on that, we shall audit afterwards

headius:
Thank you!

PR #5468, opened 2018-11-25
Fix generated javadoc

pzygielo:


enebo:
@pzygielo hey I added this into our security release 9.2.4.1 since it has no risk impact.   So you should be set now.

pzygielo:
Thanks! I will use 9.2.4.1 then, and upgrade to next release as it will be published.

PR #5474, opened 2018-11-27
less sync-ing with thread registration

kares:
... here's an interesting find while trying to upgrade an app stuck on JRuby 1.7.4

the app uses celluloid's actor model and under load degraded considerably (~ 10x).
with celluloid all calls upon actors are executed in fibers, which are kind of heavy under JRuby,
the system ends up operating with ~3000 threads and interestingly thread registration (those are fiber backing threads) accounts for quite some delay as its synchronizing on `ThreadService`.

now, that locking shouldn't be needed since the backing map is already a synchronized collection.

screenshot_025

... poked around history and there's no explicit reason for `registerNewThread` to lock on `ThreadService`
pretty much the only thing I could think of that might not work 'exactly' as before is listing threads (`getActiveRubyThreads()`) so I changed the order for the registration logic to write to the map last (after the thread/context gets fully initialized).

have patched this on top of of 9.1.17.0 and put under load - the 10x degradation was gone and all seems good, so far. still could use a review whether I did not miss anything.

also how would you guys feel about backporting to 9.1, assuming there's going to be a 9.1.18.0 ?
(as the 9.2 upgrade isn't :green_book: yet and I would hate for this app to stay behind on an old 1.7)

### DIFF
--- a/core/src/main/java/org/jruby/ext/fiber/ThreadFiber.java
+++ b/core/src/main/java/org/jruby/ext/fiber/ThreadFiber.java
@@ -31,18 +31,21 @@ public class ThreadFiber extends RubyObject implements ExecutionContext {
     public ThreadFiber(Ruby runtime, RubyClass klass) {
         super(runtime, klass);
     }
-    
-    public static void initRootFiber(ThreadContext context) {
-        Ruby runtime = context.runtime;
-        
-        ThreadFiber rootFiber = new ThreadFiber(runtime, runtime.getClass("Fiber")); // FIXME: getFiber()
 
-        RubyThread currentThread = context.getThread();
+    public static void initRootFiber(ThreadContext context) {
+        initRootFiber(context, context.getThread());
+    }
+
+    public static void initRootFiber(ThreadContext context, RubyThread currentThread) {
+        Ruby runtime = context.runtime;
+
+        ThreadFiber rootFiber = new ThreadFiber(runtime, runtime.getFiber());
+
         rootFiber.data = new FiberData(new FiberQueue(runtime), currentThread, rootFiber);
         rootFiber.thread = currentThread;
         context.setRootFiber(rootFiber);
     }
-    
+
     @JRubyMethod(visibility = Visibility.PRIVATE)
     public IRubyObject initialize(ThreadContext context, Block block) {
         Ruby runtime = context.runtime;
--- a/core/src/main/java/org/jruby/ext/fiber/ThreadFiber.java
+++ b/core/src/main/java/org/jruby/ext/fiber/ThreadFiber.java
@@ -32,10 +32,6 @@ public class ThreadFiber extends RubyObject implements ExecutionContext {
         super(runtime, klass);
     }
 
-    public static void initRootFiber(ThreadContext context) {
-        initRootFiber(context, context.getThread());
-    }
-
     public static void initRootFiber(ThreadContext context, RubyThread currentThread) {
         Ruby runtime = context.runtime;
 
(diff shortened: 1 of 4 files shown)


core/src/main/java/org/jruby/ext/fiber/ThreadFiber.java
```diff
@@ -31,18 +31,21 @@
     public ThreadFiber(Ruby runtime, RubyClass klass) {
         super(runtime, klass);
     }
-    
+
     public static void initRootFiber(ThreadContext context) {
+        initRootFiber(context, context.getThread());
+    }
+
+    public static void initRootFiber(ThreadContext context, RubyThread currentThread) {
```
headius: Is this overload really necessary? The new signature is fine, but nobody except us should be calling initRootFiber. Just nuke the original.

core/src/main/java/org/jruby/internal/runtime/ThreadService.java
```diff
@@ -264,25 +261,24 @@ public void setMainThread(Thread thread, RubyThread rubyThread) {
 
                 rtList.add(entry.getValue());
             }
-
-            return rtList.toArray(new RubyThread[rtList.size()]);
         }
```
headius: Only state changes are against rubyThreadMap, which we sync on. This seems ok.

core/src/main/java/org/jruby/internal/runtime/ThreadService.java
```diff
@@ -264,25 +261,24 @@ public void setMainThread(Thread thread, RubyThread rubyThread) {
...
+        return rtList.toArray(new RubyThread[rtList.size()]);
     }
```
headius: Only state changes are against rubyThreadMap, which we sync on. This seems ok.

core/src/main/java/org/jruby/ext/fiber/ThreadFiber.java
headius: Is this overload really necessary? The new signature is fine, but nobody except us should be calling initRootFiber. Just nuke the original.

core/src/main/java/org/jruby/internal/runtime/ThreadService.java
```diff
@@ -264,25 +261,24 @@ public void setMainThread(Thread thread, RubyThread rubyThread) {
...
+        rubyThreadMap.put(threadOrFuture, rubyThread); // synchronized
     }
 
-    public synchronized void unregisterThread(RubyThread thread) {
-        rubyThreadMap.remove(Thread.currentThread());
+    public void unregisterThread(RubyThread thread) {
+        rubyThreadMap.remove(Thread.currentThread()); // synchronized
         getCurrentContext().setThread(null);
         localContext.set(null);
     }
```
headius: Really these three sync are the only major change here, right? Thinking through it now...

* registerNewThread: there's three state changes: ThreadContext gets its RubyThread, the localContext gets a soft ref to the ThreadContext, and the root fiber for the thread is initialized into the context. All of these are or should be thread-local at this point. This seems ok.
* initRootFiber, called from registerNewThread: all thread-local state. Ok.
* associateThread: only one state change against a synchronized collection. Ok.
* unregisterThread: one state changes against a synchronized collection and then thread-local or thread-specific changes. Ok.

kares: yep, that's the gist - the rest is just more explicit passing along stuff we work with ...

core/src/main/java/org/jruby/ext/fiber/ThreadFiber.java
kares: done

headius:
Go for it. And post some celluloid numbers 😀

headius:
Oh, ftr I am fine with this in 9.1 as well. 9.1.18 seems less and less likely though.

kares:
OK, picked this up to the [9.1 branch](https://github.com/jruby/jruby/commit/9726d8aaec345e0328cb7ba203b6762b669c4374) ... just in case there's a release :smiling_imp:

PR #5475, opened 2018-11-27
Update specs

eregon:
Note the tags added in 1dbb3da are due to using the old message (Ruby 2.2) for arity error:
```
ArgumentError: wrong number of arguments (2 for 1)
instead of
ArgumentError: wrong number of arguments (given 2, expected 1)
```
(See https://github.com/oracle/truffleruby/issues/1467 for more details)


eregon:
Specs are green, merging.

PR #5481, opened 2018-11-29
Never return the internal module name as a mutable string.

### LINKED ISSUE #5480: Class names are mutable
### Environment

* `jruby 9.2.0.0 (2.5.0) 2018-05-24 81156a8 Java HotSpot(TM) 64-Bit Server VM 25.172-b11 on 1.8.0_172-b11 +jit [darwin-x86_64]`
* `Darwin albus.local 17.7.0 Darwin Kernel Version 17.7.0: Wed Oct 10 23:06:14 PDT 2018; root:xnu-4570.71.13~1/RELEASE_X86_64 x86_64`

### Expected Behavior (MRI 2.5)

```
irb(main):001:0> class Foo; end
=> nil
irb(main):002:0> Foo.name
=> "Foo"
irb(main):003:0> Foo.name << 'bla'
=> "Foobla"
irb(main):004:0> Foo.name
=> "Foo"
```

### Actual Behavior (jRuby 9.2.0.0)

```
jirb(main):001:0> class Foo; end
=> nil
jirb(main):002:0> Foo.name
=> "Foo"
jirb(main):003:0> Foo.name << 'bla'
=> "Foobla"
jirb(main):004:0> Foo.name
=> "Foobla"
```

headius:
This duplicates logic in `to_s` that dups the module name before
returning it, but also ensures that any cached module name is
frozen so it can't be modified in place.

Fixes #5480.

### DIFF
--- a/core/src/main/java/org/jruby/RubyModule.java
+++ b/core/src/main/java/org/jruby/RubyModule.java
@@ -558,7 +558,7 @@ public class RubyModule extends RubyObject {
 
     /**
      * Generate a fully-qualified class name or a #-style name for anonymous and singleton classes which
-     * is properly encoded.
+     * is properly encoded. The returned string is always frozen.
      *
      * @return a properly encoded class name.
      *
@@ -618,6 +618,8 @@ public class RubyModule extends RubyObject {
             fullName.cat19(parent).cat19(colons);
         }
         fullName.cat19(rubyBaseName());
+
+        fullName.setFrozen(true);
 
         if (cache) cachedRubyName = fullName;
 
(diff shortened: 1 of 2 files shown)


enebo:
@headius this sort of makes sense how it broke.  Because we were a string before we would always make a RubyString from the string and we did not cache that finished string.  Not sure if we always marked it frozen or not before.

headius:
@enebo I didn't see any evidence that we froze it, but it makes sense to do so. That inner method was added by you in 9.2 so it seems safe enough to make it now return a frozen string reference. `to_s` already did a dup on it before returning, so `name` was just in error.

Issue #5288, opened 2018-08-22
Regression: Refinements Broken from 9.1 to 9.2


fny:
A Ruby [library of mine](https://github.com/fny/xorcist), which provides refinements, was working as of [JRuby 9.1.16.0](https://travis-ci.org/fny/xorcist/jobs/411575098). I recently added JRuby 9.2.0.0 and found that two tests which use refinements are now broken. Note all tests pass across all other Ruby versions, and I have not made any updates to the library.

 ```
 1) Error:
RefinementsTest#test_xor:
NoMethodError: undefined method `xor' for "w\x12\xC5_\xD7\xB8\x06\xF3\xD4\x9Bl\x909\xF8(M":String
    /home/travis/build/fny/xorcist/test/refinements_test.rb:10:in `test_xor'
  2) Error:
RefinementsTest#test_xor_in_place:
NoMethodError: undefined method `xor!' for "String":String
    /home/travis/build/fny/xorcist/test/refinements_test.rb:20:in `block in test_xor_in_place'
    /home/travis/build/fny/xorcist/test/test_helper.rb:17:in `frozen_strings_dependent'
    /home/travis/build/fny/xorcist/test/refinements_test.rb:19:in `test_xor_in_place'
```

Full trace available here:
https://travis-ci.org/fny/xorcist/jobs/418932838

I've included the relevant code below as a convenience:

```rb
# test/refinements_test.rb:
require 'test_helper'

if RUBY_ENGINE != 'rbx' && RUBY_VERSION >= '2.0.0'
  require 'xorcist/refinements'

  class RefinementsTest < Minitest::Test
    using Xorcist::Refinements

    def test_xor
      assert_equal ZERO, X.xor(X)
      assert_equal ONE,  X.xor(INVX)
      assert_equal X,    X.xor(ZERO)
      assert_equal INVX, X.xor(ONE)
    end

    def test_xor_in_place
      a = "String"
      b = a
      frozen_strings_dependent {
        b.xor!(X)
        assert_equal(a, b)
      }
    end
  end
end

# lib/xorcist/refinements.rb
require 'xorcist/string_methods'

module Xorcist
  module Refinements
    refine String do
      include Xorcist::StringMethods
    end
  end
end

# lib/xorcist/string_methods.rb
module Xorcist
  module StringMethods
    def xor(other)
      Xorcist.xor(self, other)
    end

    def xor!(other)
      Xorcist.xor!(self, other)
    end
  end
end
```

[...]

PR #5487, opened 2018-12-01
[#5288] Overwrite parent class with module wrapper

alexis779:
Fixes `Module#refine` regression when using `Module#include` to inject functions.

Compare latest refinement [commit](https://github.com/jruby/jruby/commit/1839880a1eb3c05904bb6f5a8a327ae02e12d59b#diff-b0dc42ec523c3c3de1ac9ff40bc56852L880) with CRuby [rb_using_refinement](https://github.com/ruby/ruby/blob/trunk/eval.c#L1342-L1347).

`RubyModule#setSuperClass` call was dropped. We add it back again.


kares:
nice work @alexis779, thanks!

PR #5488, opened 2018-12-02
[fix] RubyException#toJava(Object) shouldn't return null

kares:
... a regression since the exception hierarchy rewrite in **9.2**
spotted at: https://github.com/sparklemotion/nokogiri/issues/1818

for compat `rubyException.toJava(Object.class)` shall return self

### DIFF
--- a/core/src/main/java/org/jruby/RubyException.java
+++ b/core/src/main/java/org/jruby/RubyException.java
@@ -335,9 +335,9 @@ public class RubyException extends RubyObject {
      * @return the converted result
      */
     @Override
-    public Object toJava(Class target) {
-        if (target.isAssignableFrom(RaiseException.class)) {
-            return target.cast(throwable);
+    public <T> T toJava(Class<T> target) {
+        if (target != Object.class && target.isAssignableFrom(RaiseException.class)) {
+            return target.cast(toThrowable());
         }
         return super.toJava(target);
     }
--- a/core/src/test/java/org/jruby/test/TestRubyException.java
+++ b/core/src/test/java/org/jruby/test/TestRubyException.java
@@ -34,6 +34,7 @@ import org.jruby.Ruby;
 import org.jruby.RubyArray;
 import org.jruby.RubyException;
 import org.jruby.RubyString;
+import org.jruby.exceptions.RaiseException;
 
 import java.io.ByteArrayOutputStream;
 import java.io.PrintStream;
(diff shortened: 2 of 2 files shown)


core/src/main/java/org/jruby/RubyException.java
```diff
@@ -335,9 +335,9 @@ public IRubyObject cause(ThreadContext context) {
      * @return the converted result
      */
     @Override
-    public Object toJava(Class target) {
-        if (target.isAssignableFrom(RaiseException.class)) {
-            return target.cast(throwable);
+    public <T> T toJava(Class<T> target) {
+        if (target != Object.class && target.isAssignableFrom(RaiseException.class)) {
```
headius: Perhaps we should widen this to Thowable? Right now that, Exception, and RuntimeException all will return self, which is not any of those.

kares: not sure if I got that right, but passing `Throwable.class` returns the `RaiseException`
... see test which is part of the diff: https://github.com/jruby/jruby/pull/5488/commits/23934bfabc5ef4cada58a4901c38da4db01dd80c#diff-f9f68150a69d43ace7536ed08c60f821R68

my brain has always a hard-time reasoning `target.isAssignable(...)` but it does the job here.
`RubyException` isn't a `Throwable` so this is the best we get from `toJava(Throwable.class)`

headius: Ok, so basically anything you could assign a RaiseException value will return the throw able, *except* for Object.class in your patch now returns self. Yeah, isAssignableFrom is badly named...I always have to reread the docs.

kares: > so basically anything you could assign a RaiseException value will return the throw able, except for Object.class in your patch now returns self.

yep that's the regression, Nokogiri stores RubyException instances in a `RubyArray` and previously a `List#get` (`element.toJava(Object.class)`) on such array returned the instance itself.

kares:
NOTE: no strong opinions against changing the behaviour of `RubyException#toJava(Object.class)` to return a `RaiseException` ... Nokogiri was able to work-around the issue. 

this PR simply restores pre 9.2 ... with the new exception hierarchy it might make sense to convert by default. most Java convertable types do the coercion e.g. `RubyString#toJava()` - `java.lang.String`

enebo:
@headius @kares yeah I think 9.3 too.

kares:
okay - opened https://github.com/jruby/jruby/pull/5498 targeting 9.3 so its not forgotten

PR #5489, opened 2018-12-02
Autoload fixes

### LINKED ISSUE #5466: Weird `Kernel.autoload` behavior
### Environment

Provide at least:
  `jruby 9.2.5.0-SNAPSHOT (2.5.0) 2018-11-22 93f0df3 OpenJDK 64-Bit Server VM 25.181-b13 on 1.8.0_181-8u181-b13-1ubuntu0.16.04.1-b13 +jit [linux-x86_64]`
  Ubuntu 16.04
  `Linux max-u1604 4.4.0-139-generic #165-Ubuntu SMP Wed Oct 24 10:58:50 UTC 2018 x86_64 x86_64 x86_64 GNU/Linux`

### Problem

I expect `Kernel.autoload :Hi, 'somefile'` to do something useful. Right now, the only way found to trigger that autoload is do call `.singleton_class::Hi` on any class (It can be `String`, `Object`, etc). I really doubt there is any usecase for this.

In MRI, `Kernel#autoload` and `Kernel.autoload` always use the `Module.nested` of the calling context. 

In JRuby, `Kernel#autoload` uses the object of the calling code. (This difference only shows up if the `autoload` call is in an instance method of a module, which is included in a class, and the method is called from an instance of the class.. Not something to worry about)

However, JRuby seems to special case `Kernel.autoload` weirdly. Looking at the code:
https://github.com/jruby/jruby/blob/93f0df32f2c77e2fbc4929a5b01771ef85a47b25/core/src/main/java/org/jruby/RubyKernel.java#L222-L223

Ideally, `Kernel.autoload` would behave the same as calling `Kernel#autoload`.

headius:
The Kernel versions of `autoload` use the caller's frame to
determine the target module, where we were just detecting that
`autoload` was called against Kernel and then defining the
autoload on Object's singleton class.

test/jruby/test_autoload.rb has also been modified to match Ruby
2.5 behavior.

Fixes #5466.

### DIFF
--- a/core/src/main/java/org/jruby/RubyKernel.java
+++ b/core/src/main/java/org/jruby/RubyKernel.java
@@ -65,6 +65,7 @@ import org.jruby.exceptions.TypeError;
 import org.jruby.internal.runtime.methods.DynamicMethod;
 import org.jruby.internal.runtime.methods.JavaMethod.JavaMethodNBlock;
 import org.jruby.ir.interpreter.Interpreter;
+import org.jruby.ir.runtime.IRRuntimeHelpers;
 import org.jruby.java.proxies.ConcreteJavaProxy;
 import org.jruby.platform.Platform;
 import org.jruby.runtime.Arity;
@@ -197,7 +198,7 @@ public class RubyKernel {
     }
 
     static RubyModule getModuleForAutoload(ThreadContext context, IRubyObject recv) {
-        RubyModule target = context.getCurrentStaticScope().getModule();
+        RubyModule target = IRRuntimeHelpers.findInstanceMethodContainer(context, context.getCurrentScope(), recv);
 
         while (target != null && (target.isSingleton() || target.isIncluded())) {
             target = target.getSuperClass();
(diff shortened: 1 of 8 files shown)


headius:
I think I've got everything in place. A number of spec tags and a few test excludes have been removed.

core/src/main/java/org/jruby/RubyKernel.java
```diff
@@ -163,66 +164,46 @@ static void recacheBuiltinMethods(Ruby runtime) {
...
 
-    static RubyModule getModuleForAutoload(Ruby runtime, IRubyObject recv) {
-        RubyModule module = recv instanceof RubyModule ? (RubyModule) recv : recv.getMetaClass().getRealClass();
-        if (module == runtime.getKernel()) {
-            // special behavior if calling Kernel.autoload directly
-            module = runtime.getObject().getSingletonClass();
+    static RubyModule getModuleForAutoload(ThreadContext context, IRubyObject recv) {
+        RubyModule target = context.getCurrentStaticScope().getModule();
+
+        while (target != null && (target.isSingleton() || target.isIncluded())) {
```
enebo: @headius So if I autoload in a module which has been included then it migrates to where that was included?  If that module is included in two places then does only the first one fire and get stuck somewhere else?  This easily may be the semantics....

headius: Yes that appears to be the semantics. In MRI this logic is rb_class_real, our "getRealClass" but that method lives on RubyClass and not RubyModule.

headius:
This should also get fixed in this PR: https://bugs.ruby-lang.org/issues/14469

I'm tagging a couple MRI tests related to that fix.

PR #5491, opened 2018-12-03
WIP: improved error message for "wrong number of arguments"

ahorek:
see https://github.com/jruby/jruby/pull/5475


headius:
This should be squashed when merged. Let us know when you're satisfied 😀

enebo:
@ahorek travis is pretty unhappy with some Enumerator arity errors.  Perhaps something in this PR changed some arity checking?

ahorek:
yeah, there're still some cases that doesn't work as expected, I'll let you know when it's ready

ahorek:
@enebo @headius ready for review!

enebo:
@ahorek looks pretty straight-forward...just merged it

headius:
This appears to be breaking at least one spec that expects no method name in the message:

```
108)
Thread.each_caller_location doesn't accept positional and keyword arguments ERROR
Expected ArgumentError (wrong number of arguments (given 2, expected 0))
but got: ArgumentError (`each_caller_location': wrong number of arguments (given 2, expected 0))
```

We probably should just drop the method name.

PR #5494, opened 2018-12-03
Ruby 2.6: Add #to_h with block argument to Array, Enum, ENV, Hash and Struct

nomadium:
Hi folks,

This PR implements another Ruby 2.6 feature: #to_h method in Array, Enumerable, ENV, Hash and Struct classes now can map their elements to new keys and values to a given block.

For more information, please see feature #15143 [1].

I believe all the related tests (MRI and ruby/spec) are passing:i [2], [3], [4], [5], [6], [7], [8], [9], [10], [11].

For reference, I include a link to the commit introducing the functionality in MRI. [12]

Thanks in advance for any feedback.

 1. https://bugs.ruby-lang.org/issues/15143
 2. https://github.com/ruby/ruby/blob/v2_6_0_preview3/spec/ruby/core/array/to_h_spec.rb#L38-L41
 3. https://github.com/ruby/ruby/blob/v2_6_0_preview3/spec/ruby/core/enumerable/to_h_spec.rb#L47-L53
 4. https://github.com/ruby/ruby/blob/v2_6_0_preview3/spec/ruby/core/env/to_h_spec.rb#L7-L17
 5. https://github.com/ruby/ruby/blob/v2_6_0_preview3/spec/ruby/core/hash/to_h_spec.rb#L10-L14
 6. https://github.com/ruby/ruby/blob/v2_6_0_preview3/spec/ruby/core/struct/to_h_spec.rb#L16-L22
 7. https://github.com/ruby/ruby/blob/v2_6_0_preview3/test/ruby/test_array.rb#L1634-L1653
 8. https://github.com/ruby/ruby/blob/v2_6_0_preview3/test/ruby/test_enum.rb#L183-L202
 9. https://github.com/ruby/ruby/blob/v2_6_0_preview3/test/ruby/test_env.rb#L400-L404
10. https://github.com/ruby/ruby/blob/v2_6_0_preview3/test/ruby/test_hash.rb#L868-L876
11. https://github.com/ruby/ruby/blob/v2_6_0_preview3/test/ruby/test_struct.rb#L365-L370
12. https://github.com/ruby/ruby/commit/abe75149d14d3286d3051c9e961ab6473a243a19

### DIFF
--- a/core/src/main/java/org/jruby/RubyArray.java
+++ b/core/src/main/java/org/jruby/RubyArray.java
@@ -2037,8 +2037,13 @@ public class RubyArray<T extends IRubyObject> extends RubyObject implements List
     	return this;
     }
 
-    @JRubyMethod(name = "to_h")
+    @Deprecated
     public IRubyObject to_h(ThreadContext context) {
+        return to_h(context, Block.NULL_BLOCK);
+    }
+
+    @JRubyMethod(name = "to_h")
+    public IRubyObject to_h(ThreadContext context, Block block) {
         Ruby runtime = context.runtime;
         int realLength = this.realLength;
 
@@ -2047,12 +2052,15 @@ public class RubyArray<T extends IRubyObject> extends RubyObject implements List
         RubyHash hash = useSmallHash ? RubyHash.newSmallHash(runtime) : RubyHash.newHash(runtime);
 
         for (int i = 0; i < realLength; i++) {
-            IRubyObject elt = eltInternal(i).checkArrayType();
-            if (elt == context.nil) {
-                throw context.runtime.newTypeError("wrong element type " + eltInternal(i).getMetaClass().getRealClass() + " at " + i + " (expected array)");
+            IRubyObject e = eltInternal(i);
+            IRubyObject elt = block.isGiven() ? block.yield(context, e) : e;
+            IRubyObject key_value_pair = elt.checkArrayType();
+
+            if (key_value_pair == context.nil) {
+                throw context.runtime.newTypeError("wrong element type " + elt.getMetaClass().getRealClass() + " at " + i + " (expected array)");
             }
 
-            RubyArray ary = (RubyArray)elt;
+            RubyArray ary = (RubyArray)key_value_pair;
             if (ary.getLength() != 2) {
                 throw context.runtime.newArgumentError("wrong array length at " + i + " (expected 2, was " + ary.getLength() + ")");
             }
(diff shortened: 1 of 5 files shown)


core/src/main/java/org/jruby/RubyArray.java
```diff
@@ -2047,12 +2047,17 @@ public IRubyObject to_h(ThreadContext context) {
         RubyHash hash = useSmallHash ? RubyHash.newSmallHash(runtime) : RubyHash.newHash(runtime);
 
         for (int i = 0; i < realLength; i++) {
-            IRubyObject elt = eltInternal(i).checkArrayType();
-            if (elt == context.nil) {
-                throw context.runtime.newTypeError("wrong element type " + eltInternal(i).getMetaClass().getRealClass() + " at " + i + " (expected array)");
+            IRubyObject e = eltInternal(i);
+            IRubyObject elt, key_value_pair;
```
enebo: Could you move decls to their first assignment:

```java
  IRubyObject elt = block.isGiven() // ...
```

core/src/main/java/org/jruby/RubyGlobal.java
```diff
@@ -434,8 +435,9 @@ public IRubyObject to_s() {
         }
 
         @JRubyMethod
-        public RubyHash to_h(){
-            return to_hash();
+        public RubyHash to_h(ThreadContext context, Block block){
```
kares: we should consider keeping these public method (on RubyXXX types) signatures around (marked as deprecated)
due exts occasionally calling methods directly e.g. 
`@Deprecated public RubyHash to_h() { return to_h(getCurrentContext(), NULL_BLOCK) }`

core/src/main/java/org/jruby/RubyArray.java
nomadium: Fixed.

core/src/main/java/org/jruby/RubyGlobal.java
nomadium: Thanks @kares, I didn't think about that, good catch. I just updated only this method following your recommendation, but just to clarify: do you consider I should to the same backward compatible change for all the public ```to_h``` modified methods in this PR?

kares: yeah, that would be the same - to usually deprecate public methods on (core) RubyXXX.java types

lopex:
MRI seems to have refactored rb_hash_set_pair (used in enum, hash, struct) where they have centralized checks for "wrong element type" and "element has wrong array length".
We seem to duplicate a lot.

enebo:
@lopex perhaps a good follow up commit?

lopex:
@enebo yeah I'll do that, hopefuly on this weekend, and have some more things on the table regarding match data, and 2.6 String#split, etc

core/src/main/java/org/jruby/RubyGlobal.java
nomadium: @kares This should be fixed in the most recent version of this PR.

nomadium:
@enebo @kares I think I addressed your comments. Please take a look at this again when you have a chance.

kares:
LGTM

PR #5496, opened 2018-12-03
Bump JRuby version and Ruby compat level to 2.6.0

nomadium:
As title says.


kares:
@enebo slightly confusing having *ruby.revision* updated without importing stdlib?

headius:
Confusing but temporary. We also don't have any features implemented. All in good time 😀

nomadium:
@headius, @kares is it OK if I run the script ```tool/sync_ruby``` to import stdlib from MRI and open a new CR with that?

PR #5504, opened 2018-12-10
Also fix up return values when frame is present, for backref.

### LINKED ISSUE #5503: Problem when method returns Regexp capture group using special variable $1
### Environment

Initially we found the problem using Jruby 9.2.1.0, but it's also actual in 9.1.7.0 and in the latest version. 
* JRuby version (jruby -v) and command line (flags, JRUBY_OPTS, etc)
```
jruby 9.2.5.0 (2.5.0) 2018-12-06 6d5a228 Java HotSpot(TM) 64-Bit Server VM 25.121-b13 on 1.8.0_121-b13 +jit [darwin-x86_64]
```
```
Darwin JBA-MacBook-Pro.local 18.2.0 Darwin Kernel Version 18.2.0: Fri Oct  5 19:41:49 PDT 2018; root:xnu-4903.221.2~2/RELEASE_X86_64 x86_64
```

* Installed or activated gems
N/A
N/A
* Environment variables
N/A

### Test code:
```
def test_match(string)
  if string =~ /(.)/
    $1
  end
end

1000000.times { test_match("a") } 

test_match("b")
```

### Expected Behavior

* function `test_match` call always returns value (as in Ruby MRI or if `match` is used instead of `=~`):
```
jruby-9.2.5.0 :026 > test_match("b")
 => "b"
```

### Actual Behavior

After some iterations returned result always is `nil`:
```
jruby-9.2.5.0 :009 > test_match("b")
 => nil
jruby-9.2.5.0 :010 > test_match("b")
 => nil
```

Full console output:
[test_script.log](https://github.com/jruby/jruby/files/2662517/test_script.log)

headius:
This logic was used to ensure that any heap-scoped variable loads
happened before the "pop" logic for those scopes. However another
type of variable, frame-locals like $~ and $_, did not get the
same treatment. This patch ensures that any frame or binding
requirement triggers the return value copy fixup.

Fixes #5503.

### DIFF
--- a/core/src/main/java/org/jruby/ir/passes/AddCallProtocolInstructions.java
+++ b/core/src/main/java/org/jruby/ir/passes/AddCallProtocolInstructions.java
@@ -167,7 +167,8 @@ public class AddCallProtocolInstructions extends CompilerPass {
                     // Breaks & non-local returns in blocks will throw exceptions
                     // and pops for them will be handled in the GEB
                     if (!bb.isExitBB() && i instanceof ReturnInstr) {
-                        if (requireBinding) fixReturn(scope, (ReturnInstr)i, instrs);
+                        // Frame holds backref and lastline, binding holds heap scopes
+                        if (requireFrame || requireBinding) fixReturn(scope, (ReturnInstr)i, instrs);
                         // Add before the break/return
                         i = instrs.previous();
                         popSavedState(scope, bb == geb, requireBinding, requireFrame, savedViz, savedFrame, instrs);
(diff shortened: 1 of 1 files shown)


subbuss:
This might be conservative in the common case. It is not frame require that requires the return to be patch up .. but the fact that the return uses frame vars that requires the patch up. So, you could refine this if you wish to.

headius:
@subbuss Yeah that's kinda what I was thinking. Only returns that access the frame or binding need to be affected it. But I also was thinking...it's just one copy. Lack of visibility into which operands have "special" needs frustrates this a little.

PR #5505, opened 2018-12-10
Update stdlib to Ruby 2.5.3.

headius:


lib/ruby/stdlib/resolv.rb
```diff
@@ -743,66 +739,49 @@ class UnconnectedUDP < Requester # :nodoc:
         def initialize(*nameserver_port)
           super()
           @nameserver_port = nameserver_port
-          @mutex = Thread::Mutex.new
-          @initialized = false
-        end
-
-        def lazy_initialize
```
ahorek: just for confirmation, this was backported to ruby 2.3, 2.4 and 2.6?
isn't 2.5 also affected?
see https://bugs.ruby-lang.org/issues/14571 and https://github.com/jruby/jruby/pull/5074

test/mri/ruby/test_pack.rb
```diff
@@ -857,4 +860,20 @@ def test_unpack1
     assert_equal "hogefuga", "aG9nZWZ1Z2E=".unpack1("m")
     assert_equal "01000001", "A".unpack1("B*")
   end
+
+  def test_pack_infection
```
kares: fix for pack/unpack tainting is already on master: https://github.com/jruby/jruby/commit/1d1bb5ddf65b320efc4161f20e1ba8338e4de986
... but you will need to tag this one since JRuby does not do much with 'P' (or 'p')

lib/ruby/stdlib/resolv.rb
headius: Yes, this was originally filed against 2.5 it appears and backported as far as 2.3.

test/mri/ruby/test_pack.rb
headius: I think @enebo was looking into this...maybe it doesn't need to be done.

lib/ruby/stdlib/resolv.rb
ahorek: yeah, but this change reverts the fix. Maybe MRI guys forgot to backport it?

headius:
The final completed build failed because an earlier exclude was accidentally removed. I am merging after having restored that tag.

PR #5506, opened 2018-12-11
review thread sync and clear references

kares:
- a loop-less `runtime.getCurrentContext`
- do not re-create a (context) soft-ref as threads get finalized (unregistered)
- thread locals synchronization does not need to lock the RubyThread


core/src/main/java/org/jruby/RubyThread.java
```diff
@@ -382,8 +377,16 @@ public IRubyObject setErrorInfo(IRubyObject errorInfo) {
...
     }
 
-    public void setContext(ThreadContext context) {
-        this.contextRef = new WeakReference<ThreadContext>(context);
+    public void setContext(final ThreadContext context) {
+        if (context == null) {
+            WeakReference<ThreadContext> contextRef = this.contextRef;
+            if (contextRef != null) {
+                contextRef.clear();
+                this.contextRef = null;
```
headius: Maybe this should be moved to a "clearContext" method rather than special-casing null?

core/src/main/java/org/jruby/RubyThread.java
```diff
@@ -938,31 +950,28 @@ public boolean isAlive(){
 
     @JRubyMethod
     public IRubyObject fetch(ThreadContext context, IRubyObject key, Block block) {
-        Ruby runtime = context.runtime;
-
-        IRubyObject value = op_aref(key);
+        IRubyObject value = op_aref(context, key);
 
-        if (value.isNil()) {
+        if (value == context.nil) {
```
headius: Is this intended as an optimization? isNil is bimorphic, so it should inline everywhere, but context.nil will require a field access every time.

headius: Correction...isNil is bimorphic so it *could* inline everywhere. It may not because it's called against more than two different types and some JVMs don't do class hierarchy analysis.

core/src/main/java/org/jruby/internal/runtime/ThreadService.java
```diff
@@ -202,30 +197,34 @@ public void initMainThread() {
...
+        if ((context = ref.get()) != null) {
+            return context;
         }
 
-        return context;
+        // context is null, wipe out the SoftReference (this could be done with a reference queue)
+        localContext.set(null);
+
+        // loop until a context is available, to clean up soft-refs that might have been collected
+        return getCurrentContext();
```
headius: Isn't this recursing? Unlikely to be a problem but the comment claims it's a loop.

kares: there used to be one - so the comment was kept, except for the loop word its true :)

core/src/main/java/org/jruby/RubyThread.java
kares: generally, have been preferring `==` ... should have written a JMH before assuming

core/src/main/java/org/jruby/RubyThread.java
kares: good point, let's have a `clearContext`

core/src/main/java/org/jruby/RubyThread.java
kares:
```
Benchmark                               Mode  Cnt    Score     Error   Units
NilCheckBenchmark.benchIsNilFalse      thrpt    5  437.404 ±  44.222  ops/us
NilCheckBenchmark.benchIsNilFalseCast  thrpt    5  447.824 ± 109.204  ops/us
NilCheckBenchmark.benchIsNilTrue       thrpt    5  431.957 ±  63.293  ops/us
NilCheckBenchmark.benchIsNilTrueCast   thrpt    5  433.628 ±  57.552  ops/us
NilCheckBenchmark.benchNilCmpFalse     thrpt    5  462.117 ±  43.782  ops/us
NilCheckBenchmark.benchNilCmpTrue      thrpt    5  440.100 ±  35.774  ops/us
```
... seems it doesn't matter much

PR #5510, opened 2018-12-12
Backward traces

headius:
Work in progress to implement backward traces and newer MRI formatting and highlighting from Ruby 2.5.

The relevant tests will be tagged for the stdlib merge in #5505.


headius:
This appears to have been reverted in CRuby (thank goodness) according to @jeremyevans here: https://bugs.ruby-lang.org/issues/14558#note-5

The referenced hash does not appear to link to the reversion but I have asked for clarification. In any case, I am closing this and we will not support it (unless someone really demands we add the option in the future).

headius:
Proper hash for the reversion is ruby/ruby@487d0c99d53208594702bb3ce1c657130fb8d65f.

PR #5515, opened 2018-12-13
Add log.singletons and .verbose to track singleton class creation.

headius:
Singleton objects, usually created via `extend` or `def obj.meth`,
frequently waste CPU cycles and memory. It's also easy to
accidentally trigger an object to become a singleton. These
properties will make it easy to track all singleton classes
created by an application, along with an optional stack trace.

### DIFF
--- a/core/src/main/java/org/jruby/MetaClass.java
+++ b/core/src/main/java/org/jruby/MetaClass.java
@@ -31,6 +31,9 @@
 package org.jruby;
 
 import org.jruby.runtime.builtin.IRubyObject;
+import org.jruby.util.cli.Options;
+import org.jruby.util.log.Logger;
+import org.jruby.util.log.LoggerFactory;
 
 public final class MetaClass extends RubyClass {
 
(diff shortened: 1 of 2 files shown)


headius:
@guizmaii This is the flag I was mentioning. Run JRuby with `-Xlog.singletons` or `-Xlog.singletons.verbose` or add `jruby.` to the beginning of those and set them as JVM properties.

kares:
loos good - hopefully real-world (app) usable as well ... seems like it would be, back-trace makes it so

headius:
@kares Yeah I think it will be useful. If we see tons of new singleton classes being created long after an app has booted up, that's a red flag. We can flip the verbose flag and see where it's happening.

I am only logging the JVM trace here, though...perhaps I should log a "full" trace with interpreted Ruby frames intact? It's going to LONG either way.

enebo:
since kares also looked at this I just merged since I see no issues with this and think it will be useful

guizmaii:
How can I test this ? 

Is is something easy to compile locally JRuby and then use this locally compiled version ?

headius:
@guizmaii Yes, check out my branch (headius/jruby) and you can build and use it. Alternatively you can pick it up from nightly builds (https://www.jruby.org/nightly)

guizmaii:
Ok thanks @headius. I'll try with the nightly builds I think :)

Issue #5493, opened 2018-12-03
Difference of DateTimes with fractional milliseconds calculated incorrectly


ikaronen-relex:
The following appears to be a regression in DateTime handling in JRuby 9.2:

```
[1] pry(main)> JRUBY_VERSION
"1.7.27"
[2] pry(main)> (DateTime.new(2018, 1, 1, 0, 1, 29.000001) - DateTime.new(2018, 1, 1, 0, 0, 30)).to_f * 24 * 60 * 60
59.000001
```

```
[1] pry(main)> JRUBY_VERSION
"9.1.15.0"
[2] pry(main)> (DateTime.new(2018, 1, 1, 0, 1, 29.000001) - DateTime.new(2018, 1, 1, 0, 0, 30)).to_f * 24 * 60 * 60
59.000001
```

```
[1] pry(main)> JRUBY_VERSION
"9.2.4.1"
[2] pry(main)> (DateTime.new(2018, 1, 1, 0, 1, 29.000001) - DateTime.new(2018, 1, 1, 0, 0, 30)).to_f * 24 * 60 * 60
145.40000008881543
```

It appears that the DateTime subtraction code in JRuby 9.2 is incorrectly interpreting fractional _milliseconds_ as fractional _days_.  Thus, instead of the expected difference of 59.000001 seconds, the code is instead returning a difference of 59 + (60 * 60 * 24 * 0.001) = 59 + 86.4 = 145.4 seconds (plus some irrelevant floating point roundoff noise).

Ps. We've had issues with DateTime math in JRuby 9.2 before, and pull request #5369 by @massive was intended to address them. However, while commit 1360a4a appears to have made the specs included in that PR pass, it does not seem to have fully fixed the underlying problem(s), as the examples above demonstrate. Until this issue is properly fixed (or we find a reliable work-around), this is blocking JRuby 9.2 adoption for us.

ikaronen-relex:
Indeed, looking at the code in https://github.com/jruby/jruby/blob/9.2.4.1/core/src/main/java/org/jruby/ext/date/RubyDate.java#L1210, I'm pretty sure this bug occurs because `subMillisDiff()` returns a difference measured in (fractional) milliseconds, but `op_minus_date()` adds it to a difference measured in days. Adding a division by `DAY_MS` somewhere in there should fix this.

headius:
Thank you for the report and PR!

PR #5516, opened 2018-12-13
Fix sub-millisecond datetime difference handling

ikaronen-relex:
This patch fixes issue #5493 and adds the test case from the report as a new spec. It also fixes the existing DateTime add/subtract specs to actually use sub-millisecond differences like their descriptions say.

### DIFF
--- a/core/src/main/java/org/jruby/ext/date/RubyDate.java
+++ b/core/src/main/java/org/jruby/ext/date/RubyDate.java
@@ -1213,6 +1213,7 @@ public class RubyDate extends RubyObject {
 
         RubyNumeric subDiff = subMillisDiff(context, that);
         if ( ! subDiff.isZero() ) { // diff += diff_sub;
+            subDiff = subDiff.convertToRational().op_div(context, RubyFixnum.newFixnum(context.runtime, DAY_MS));  // #5493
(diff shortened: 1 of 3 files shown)


core/src/main/java/org/jruby/ext/date/RubyDate.java
```diff
@@ -1213,6 +1213,7 @@ private RubyNumeric op_minus_date(ThreadContext context, final RubyDate that) {
 
         RubyNumeric subDiff = subMillisDiff(context, that);
         if ( ! subDiff.isZero() ) { // diff += diff_sub;
+            subDiff = subDiff.convertToRational().op_div(context, RubyFixnum.newFixnum(context.runtime, DAY_MS));  // #5493
```
ikaronen-relex: It might be more efficient to replace `RubyFixnum.newFixnum(context.runtime, DAY_MS)` with `DAY_MS(context)` here, but if so, the same change should presumably be made to a lot of other places in this file as well.

headius: I think you're right, but that's unrelated to your fix. Feel free to do it in another PR, though! cc @kares

Issue #5511, opened 2018-12-12
Object shaping may be pulling in more variables than necessary


headius:
While investigating a memory leak with a user, I realized that a number of ActiveRecord model objects were being created with a very large number of fields, over 100. The fields were almost all `@after_remove_some_model` and `@before_remove_some_model` which I believe comes from the [state_machines](https://github.com/state-machines/state_machines) Ruby library.

I believe these are supposed to be class-level attributes, not instance-level, as they appear to point at hooks or event handlers metaprogrammed into the class itself. We need to investigate why these fields appear to be getting into the objects.

There is a small chance that these fields are being added on a per-instance basis, possibly by accident, because in this particular leaking heap all of these objects had also been turned into singletons. I was not able to determine where or why the singletons were being created.

PR #5518, opened 2018-12-13
Limit the width of reified objects to 50 by default.

headius:
This relates to #5511 but is not a fix. This only limits how wide
these object will be, so we're not generating extremely wide
classes for weirdly dynamic objects that create lots of new
instance variables.

This will be improved once we make objects know about their own
shapes, so they can evolve to handle the right number of variables
that are actually used.


headius:
cc @guizmaii

headius:
Ok looks like additional work needed...it's trying to grab vars at widths outside the max.

Issue #5520, opened 2018-12-15
Mutex gets 'broken' sporadically (since 9.2.1)


kares:
we're seen some mutex occasional failures in CI, for a while now.
turns out maybe not false alarms, but the implementation regressing (since **9.2.1**)

the issue can be spotted under concurrent threading with some user use-cases.
e.g. a .rb `CountDownLatch` https://raw.githubusercontent.com/benlangfeld/countdownlatch/master/lib/countdownlatch.rb ... guessing AR's connection pool might be also affected by this.

already bisected and find the cause to be: 66d2905ae233a39e36fcbe3ade6382c2892ade8e

which makes no sense ... going to try setup a PR with a concurrent test-case and a proposed fix.

kares:
CDL impl + occasional `Timeout::Error` failing script : 
```ruby
require 'thread'
require 'timeout'

class CountDownLatch
  def initialize(count)
    raise ArgumentError if count < 0
    @count = count
    @mutex = Mutex.new
    @conditional = ConditionVariable.new
  end

  def countdown!
    @mutex.synchronize do
      @count -= 1 if @count > 0
      @conditional.broadcast if @count == 0
    end
  end

  def count
    @mutex.synchronize { @count }
  end

  def wait(timeout = nil)
    # begin
      Timeout::timeout timeout do
        @mutex.synchronize do
          @conditional.wait @mutex if @count > 0
        end
      end
      true
    # rescue Timeout::Error
    #   false
    # end
  end
end

def wait_for_latch(count = 10, timeout = 5)
  latch = CountDownLatch.new count
  count.times do
    Thread.new { latch.countdown! }
  end
  latch.wait(timeout)
end

10_000.times do |i|
  puts i

  result = wait_for_latch
end
```

headius:
This seems to have manifested in a spec failure.

kares:
yy, it has been manifesting from time-to-time, for a while now.

headius:
I'm on it.

headius:
@kares Should we marked this fixed or move it to 9.2.7? My small patch addresses the base issue.

headius:
[...]

PR #5522, opened 2018-12-16
test, review & avoid (regressed) mutex behavior

kares:
these changes are based on reviewing code for #5520 

than there's the actual "wild" fix: https://github.com/jruby/jruby/compare/mutex-fix?expand=1#diff-330867c3ee2ffcb2a20e80af7bf9b4b3R161 ... as this line doing synchronization locking seems to be causing the problem (see explanation at GH-5520)

not entirely sure - why it's happening, yet, maybe it relates to delivering interrupts on the main thread?

to reproduce, just run the test that is added here : 
```
TIMES=100_000 bin/jruby test/jruby/test_thread.rb --name=/latch/
```

.. changing the line back to a blocking collection should make the test fail e.g.
```diff
    /** Stack of interrupt masks active for this thread */
-    private final List<RubyHash> interruptMaskStack = new CopyOnWriteArrayList<>();
+    private final List<RubyHash> interruptMaskStack = new Vector<>();
```
https://github.com/jruby/jruby/compare/mutex-fix?expand=1#diff-330867c3ee2ffcb2a20e80af7bf9b4b3R161

@headius feel free to push to the branch as/if needed. while I'd like to understand why that line causes such a regression I might not have the time to look in detail (next week).

### DIFF
--- a/core/src/main/java/org/jruby/RubyThread.java
+++ b/core/src/main/java/org/jruby/RubyThread.java
@@ -39,21 +39,15 @@ import java.nio.channels.Channel;
 import java.nio.channels.SelectableChannel;
 import java.nio.channels.SelectionKey;
 import java.nio.channels.Selector;
-import java.util.Arrays;
-import java.util.Collections;
 import java.util.Iterator;
 import java.util.Queue;
 import java.util.Vector;
 import java.util.WeakHashMap;
 import java.util.HashMap;
 import java.util.List;
-import java.util.ArrayList;
 import java.util.Map;
 import java.util.Set;
-import java.util.concurrent.ConcurrentLinkedQueue;
-import java.util.concurrent.ExecutionException;
-import java.util.concurrent.Semaphore;
-import java.util.concurrent.TimeUnit;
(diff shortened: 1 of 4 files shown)


kares:
interestingly, no longer passes the interrupt mask test - unexcluded from the related commit: https://github.com/jruby/jruby/commit/66d2905ae233a39e36fcbe3ade6382c2892ade8e

headius:
For 9.2.6 I have pushed a trivial version of the core fix here, the deadlock inside the copy of interrupt handlers. See adcd781332.

The deadlock came from having two synchronized lists try to `addAll` at the same time. I have not confirmed exactly how that situation happens, but this fix makes sure it never will.

kares:
rebased/redone on top of master - to also take a mask snapshot. 
... might move the test on master just to see if it reproduces.

kares:
okay, test is doing nice and w snapshot taking I am no longer reproducing any (mutex) issues.
have some interrupt logic for matching compatibility, will base those as this one gets merged.

Issue #5479, opened 2018-11-28
Update ERB stdlib from MRI 2.5.0 to at least 2.5.1


presidentbeef:
### Environment

jruby 9.2.4.0 (2.5.0) 2018-11-13 8faff06 Java HotSpot(TM) 64-Bit Server VM 10.0.1+10 on 10.0.1+10 +jit [darwin-x86_64]

### Issue

JRuby 9.2.4.0 uses the stdlib version of ERB from MRI 2.5.0. Unfortunately, that version introduced a backwards-incompatible change, as documented in https://redmine.ruby-lang.org/issues/14243. The change was reverted in MRI 2.5.1.

FWIW I tried the [current trunk version of ERB](https://github.com/ruby/ruby/blob/ab5a3db48e446a33f790fa5e8594a6dc24062987/lib/erb.rb) instead and it seemed to work fine with JRuby 9.2.4.0.

I'm not sure what the JRuby policy is for updating the stdlib, but I'm guessing it's to match the compatibility version that's reported with `jruby -v`. That would rule out just updating ERB to the 2.5.1 version.

So I suppose this ticket is just a little bit of motivation to update to the latest 2.5.x stdlib :)

headius:
@presidentbeef So you're saying that 2.5.1 has the fix? I agree, we should do a stdlib update for 9.2.6 (9.2.5 already has a lot of fixes in it) to the latest 2.5.x stdlib.

Need to figure out a better policy here...maybe I can get our patches into the MRI repo so we can just source stdlib directly from.there.

presidentbeef:
Yes, that's correct: 2.5.1 has the fix.

I was tracking down why I was having test failures with JRuby and I realized it was the same bug I had already seen and reported with MRI. It's *only* in 2.5.0.

headius:
Gotcha, thanks. We will roll stdlib to latest 2.5.x (and gems, if there's updates) for 9.2.6 some time in December. I'll start that soon...if it looks like not-too-many changes it might make 9.2.5.

headius:
This has been fixed in jruby/jruby#5505 and which was merged to master in d03c357. JRuby 9.2.6 will ship with Ruby 2.5.3 stdlib. We also fixed all relevant test failures in the updated MRI test suite, mostly outside of stdlib (in core).

PR #5524, opened 2018-12-17
backport fd leak in resolv

ahorek:
this patch restores the original patch in https://github.com/jruby/jruby/pull/5074
it was merged, but reverted later https://github.com/jruby/jruby/issues/5479

I think we should fork it to avoid regression until https://bugs.ruby-lang.org/issues/15414 is resolved


enebo:
@ahorek can you also submit to https://github.com/jruby/ruby/commits/jruby-ruby_2_5_0.  Not quite sure on branch name but it looks like that is 2.5.x where we maintain these extra changes.

Issue #5523, opened 2018-12-16
Regression with Marshal.load(a Hash)


larskanis:
### Environment

```sh
$ ruby -v
jruby 9.2.5.0 (2.5.0) 2018-12-06 6d5a228 OpenJDK 64-Bit Server VM 10.0.2+13-Ubuntu-1ubuntu0.18.04.2 on 10.0.2+13-Ubuntu-1ubuntu0.18.04.2 +jit [linux-x86_64]

$ uname -a
Linux netzbuch 4.15.0-36-generic #39-Ubuntu SMP Mon Sep 24 16:19:09 UTC 2018 x86_64 x86_64 x86_64 GNU/Linux
```

### Expected Behavior

This is what MRI 2.5.3 and JRuby <= 9.2.4.0 computes:
```ruby
h={:€a=>nil, :€c=>nil, :€h=>nil}
Marshal.load(Marshal.dump(h)) #  => {:€a=>nil, :€c=>nil, :€h=>nil}
```

### Actual Behavior

But with JRuby-9.2.5.0 wired things are happening:
```ruby
h={:€a=>nil, :€c=>nil, :€h=>nil}
Marshal.load(Marshal.dump(h)) #  => {:€a=>nil, :€c=>true, nil=>:€h}
```

This regression was raised by the [Eventbox](https://github.com/larskanis/eventbox) test suite.

headius:
Nice find! Maybe we should be running Eventbox's suite in our CI 😁

headius:
Ok, I see the problem. It looks like a flaw in how we register a symbol and its encoding in the marshaled data. The indices of the links are getting off, which causes it to ready too little data for `€c`, so the next value in the stream is the "T" for UTF-8 encoding, and then the next key appears to be the nil value that `€c` should have been assigned.

headius:
I believe I've fixed the issue in #5528, though there's a bit of a caveat listed there relating to visibility of partially-unmarshaled symbols.

The problem was largely as I described it above; however, the bug was in unmarshaling rather than marshaling. On the dump side, we register the symbol first and then its encoding, but on the unmarshal side we accidentally registered the encoding first. This led to it loading the wrong content for subsequent encodings since they all used the same encoding key of `:E`. The links for any encodings past the first would point at the first symbol, see it as not an encoding, and then not try to unmarshal the actual encoding. This shifted all values in the marshal stream around, so the keys and values were thrown off.

larskanis:
[...]

PR #5528, opened 2018-12-18
Handle symbol dumping properly

headius:
This PR is for #5523. In that issue, we found that the order a symbol and its encoding are being registered are correct, but when unmarshaling the symbol we register the encoding first. This led to the actual symbol content being treated as the encoding key, and the encoding value became the hash value associated to the original symbol. Then everything was off by one.

There's an issue here, though, in that the symbol is created and registered with our internal tables before its encoding is set correctly. This is also how MRI works, but they don't have issues for obvious reasons.

### DIFF
--- a/core/src/main/java/org/jruby/RubySymbol.java
+++ b/core/src/main/java/org/jruby/RubySymbol.java
@@ -39,7 +39,6 @@ package org.jruby;
 
 import org.jcodings.Encoding;
 import org.jcodings.specific.ASCIIEncoding;
-import org.jcodings.specific.ISO8859_1Encoding;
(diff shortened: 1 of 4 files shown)


headius:
Additional work brought about by @enebo realizing the previous logic could potentially change the encoding of an existing symbol. I added a new `newSymbol` path that accepts a "handler" lambda, called with the existing symbol + false (not a new symbol) or the new symbol + true (new symbol) *before* storage into the symbol table. This avoids both modifying an existing symbol's encoding *and* allows us to register the symbol link before setting the encoding.

headius:
I believe @enebo approves.

Issue #5525, opened 2018-12-17
High lock contention invalidating "failover global" invalidator


headius:
The FailoverSwitchPoint invalidator implements an invalidator that will eventually give up and always be failed. This is to reduce the overhead from global variables that are continually updated and read during the lifetime of the application.

The logic currently uses a synchronized method to do the update of the failure count and the invalidation of the existing switchpoint. As reported by @guizmaii on Gitter, some applications may see this contention in lock profiling if they are updating global variables frequently.

This invalidator is also used for invalidating the base Fixnum and Float methods, but generally I would not expect an application to be continuously redefining math operations.

I have requested that @guizmaii try to determine why global variables are being updated so much in his application, which embeds JRuby into a larger app.

I have an atomics-based improvement for FailoverSwitchPointInvalidator in progress.

Here's a result of a script updating the same global across many threads. The profiler I'm using is from https://github.com/jvm-profiling-tools/async-profiler.

Script:
```ruby
loop {
  100.times.map {
    Thread.new {
      1000000.times { $foo = 1 }
    }
  }.each(&:join)
}
```

Top lock contention:
```
          ns  percent  samples  top
  ----------  -------  -------  ---
3365580097350  100.00%  2297169  org.jruby.runtime.opto.FailoverSwitchPointInvalidator
```

headius:
With the fix in #5526:

Top lock contention:
```
          ns  percent  samples  top
  ----------  -------  -------  ---
   210857385   98.00%        4  java.util.Collections$SynchronizedSet
     4295257    2.00%      436  java.lang.ref.ReferenceQueue$Lock
```

guizmaii:
@headius impressive ! 

Is already there a nightly build including this fix I can try with my app ?

Issue #4808, opened 2017-10-09
Rework global variables, caching to eliminate race conditions


headius:
This issue spawned from discoveries in #4797, where @eregon discovered that various races in our global variable subsystem seem to be causing us to cache invalid values.

There are numerous races, including but not limited to:

* GlobalVariables uses a CHM but manually performs multi-step read/write operations that would race with other threads.
* UndefinedVariable modifies the GlobalVariable it was attached to, updating it with a different accessor. This makes caching the accessor impossible.
* The traces active for each global variable are managed in an unsynchronized ArrayList, which additionally can be modified from outside the subsystem.

There are likely others. The main problems lie in the fact that this part of JRuby is some of the oldest code, many files only bearing the copyright of the first committer to JRuby. None of this code was built for concurrency, and the many levels of abstraction and mutation make a simple fix impossible.

This bug will track a rework of the global variable subsystem for JRuby 9.2, which will eliminate some of the unnecessary abstraction, use appropriate thread-safe structures and synchronization for modification, and properly enable safe caching of values.

sumitmah:
I remember discussing similar stuff mruby with Matz in Rubyconf.

headius:
We did various amounts of work to improve globals during the 9.1 cycle, but I have not evaluated how well the current implementation deals with globals. The plumbing is still a bit messy and I'm sure there are remaining races, plus we have never really reenabled global optimization in indy the way I'd like.

Bumping to post 9.2 for additional work.

headius:
See #5525 for another example of problems with this code. It needs to be reworked from the ground up.

headius:
This is largely resolved by #5536 but latest additions there appear to be hanging some suites. I also need to benchmark "bad" cases like heavily-modified globals to ensure we're no worse than the existing logic. Bumped to 9.2.9.

headius:
Moved to 9.4 as part of larger perf work that will happen there.

headius:
This is now happening in #8003.

PR #5536, opened 2018-12-21
Convert all global variables to invokedynamic call sites.

headius:
Summary
=======

This is an experiment in using invokedynamic for global variable
storage. In the old logic for invokedynamic global variable
access, there were a number of challenges to maintaining a
SwitchPoint, invalidating it appropriately, and not racing with
value updates. This commit makes all internal "accessors" actually
be MutableCallSite instances via the new GlobalSite class. These
sites customize their own behavior, wiring up appropriate
transformations for read-only, lazily-calculated, non-global and
other types of global variables.

Performance
===========

It's unclear how much of a performance gain a typical case will
see, given the limited use of global variables in the wild.
However some variables see more frequent usage, like $VERBOSE or
$DEBUG for logging frameworks.

I've done limited experiments with a loop that checks a global:

```ruby
$foo = true
def foo
  i = 0
  while i < 10_000_000
    i+=1 # if $foo
  end
end
p 1000.times.map {
  t = Time.now
  foo
  Time.now - t }.sum / 1000 / 10_000_000
```

On JRuby 9.2.4.0 on C2, the loop dominates and each iteration
averages 5.50ns, while the version with the globals averages
5.67ns, for an increased cost of 0.17ns.

With this patch on C2, the same values are 5.50ns and 5.56sms,
increasing cost only 0.06ns, a savings of 0.11ns.

JRuby 9.2.4.0 on Graal JIT averages 0.91ns and 2.53ns, a more
obvious hit of 1.62ns.

This patch on Graal JIT averages 0.90ns and 0.92ns, an
increase of 0.02ns and a savings of 1.6ns

These numbers were gathered on my laptop, so throttling and such
may have an impact here, but global reads are consistently less
of a hit on the patch.

Caveats
=======

This initial patch also makes non-global variables such as $~ and
$_ into call sites, but they just permanently bind to a thread or
frame-local accessor. No attempt is made to cache their values.

I have not installed any sort of failover to the global call sites
to prevent constant deoptimization of methods that access
frequently-modified globals. This will likely still be needed,
since updating a MutableCallSite should have similar overhead to
a SwitchPoint invalidation. It should be possible to simply force
such call sites to bind permanently as with the non-globals above.


headius:
See #4808, #5525.

eregon:
Is a MutableCallSite correct here, or should it be a VolatileCallSite?
I guess MutableCallSite is fine when going to the fallback, as other threads would notice the SwitchPoint is invalid, but what when the global variable value changes and < maxfail? Those updates should be observed immediately by other threads.

headius:
@eregon I believe this is ok, since the contents of the call site are not themselves unguarded. The call site will always point to either the fallback, the uncached lookup, or a SwitchPoint-guarded constant value. The first two always return to the global variable accessor's volatile field, and the SwitchPoint will have volatile semantics preventing any threads from traversing it once invalidated. Even if a thread happens to cache the guarded constant in a non-volatile way, it will simply branch back into the fallback logic once the SwitchPoint has been invalidated.

eregon:
@headius Where is the SwitchPoint for the constant case?
I don't see it in https://github.com/jruby/jruby/pull/5536/files#diff-d527c65cb35e7631dd5519a097ca51a3

headius:
Moved to 9.4 as part of larger perf work there.

headius:
I attempted to rebase this one on master but it has gotten too far off during 9.4 changes. I will endeavor to recreate these optimizations after we release 9.4 later this month.

PR #5546, opened 2019-01-01
[refactor] should not use REALTIME for elapsed time

kares:
should properly use a MONOTONIC timer for measuring elapsed time


enebo:
@kares I believe we used to do this until people reported issues with paused virtual machines.

kares:
an option would be to have a cli switch for a monotonic timer, which would be curTimeMillis by default?
` -Djruby.timer.monotonic=false`

PR #5548, opened 2019-01-03
Define a proper annotation parameter for DynamicMethod.Version

donv:
This parameter is set to 0 in
InvocationMethodFactory.createJavaMethodCtor and in the precompiled
INVOKER classes.  This works somehow on the JVM, but fails in the
DalvikVM.

Source for defining annotation parameters:
https://docs.oracle.com/javase/tutorial/java/annotations/declaring.html


headius:
Super weird. Merging.

PR #5549, opened 2019-01-04
Use `invoke` instead of `invokeExact` to avoid casting error on Android

donv:


donv:
Fixed in Android upstream.

PR #5540, opened 2018-12-26
Ruby 2.6: Add endless range

nomadium:
Hi folks,

This PR implements another Ruby 2.6 language feature: Introduce endless ranges.

For more information, please see feature #12912. [1]

All its related tests are passing. [2] .. [19]

For reference, I include a link to the commit introducing the functionality in MRI. [20]

Thanks in advance for any feedback.

Note: Other methods in the Range class require changes to support or raise errors on endless ranges (e.g. #last, #max, #size and #to_a). I'll address that in subsequent commits or in a different PR. This PR is already long enough and I wanted to gather feedback to check if I'm going in the right direction.

1. https://bugs.ruby-lang.org/issues/12912
2. https://github.com/ruby/ruby/blob/v2_6_0/test/ruby/test_array.rb#L29-L55
3. https://github.com/ruby/ruby/blob/v2_6_0/test/ruby/test_array.rb#L186-L209
4. https://github.com/ruby/ruby/blob/v2_6_0/test/ruby/test_array.rb#L342-L385
5. https://github.com/ruby/ruby/blob/v2_6_0/test/ruby/test_array.rb#L387-L484
6. https://github.com/ruby/ruby/blob/v2_6_0/test/ruby/test_array.rb#L1428-L1469
7. https://github.com/ruby/ruby/blob/v2_6_0/test/ruby/test_range.rb#L9-L24
8. https://github.com/ruby/ruby/blob/v2_6_0/test/ruby/test_range.rb#L32-L39
9. https://github.com/ruby/ruby/blob/v2_6_0/test/ruby/test_range.rb#L41-L50
10. https://github.com/ruby/ruby/blob/v2_6_0/test/ruby/test_range.rb#L137-L146
11. https://github.com/ruby/ruby/blob/v2_6_0/test/ruby/test_range.rb#L152-L157
12. https://github.com/ruby/ruby/blob/v2_6_0/test/ruby/test_range.rb#L159-L178
13. https://github.com/ruby/ruby/blob/v2_6_0/test/ruby/test_range.rb#L180-L198
14. https://github.com/ruby/ruby/blob/v2_6_0/test/ruby/test_range.rb#L200-L206
15. https://github.com/ruby/ruby/blob/v2_6_0/test/ruby/test_range.rb#L208-L323
16. https://github.com/ruby/ruby/blob/v2_6_0/test/ruby/test_range.rb#L343-L405
[...]

PR #5552, opened 2019-01-09
Ruby 2.6: Add support for endless range in more methods

nomadium:
Hi folks,

This PR iterates on the changes introduced on #5540.

Summary:

- Make Range#min, max, include?, cover?, and === to support endless range

- Range#to_a now raises RangeError if it is endless

- Range#size now returns Float::INFINITY if it is endless

- Range#last and #max raises a RangeError if it is endless
    
    
For more information, please see feature #12912. [1]

For reference, I include links to similar commits in MRI: [2] [3] [4] [5].

Thanks for the feedback.

1. https://bugs.ruby-lang.org/issues/12912
2. https://github.com/ruby/ruby/commit/db1bdecb0d925b4668c0735158fce466333848f1
3. https://github.com/ruby/ruby/commit/c19ecf05b4c728951e1a1e223a40ae6883a4f8e0
4. https://github.com/ruby/ruby/commit/342619300ce26f9a134249002270c5d38d5993b3
5. https://github.com/ruby/ruby/commit/cae45174190b49ca3c67ca606c5f4b71e3847842

### DIFF
--- a/core/src/main/java/org/jruby/RubyRange.java
+++ b/core/src/main/java/org/jruby/RubyRange.java
@@ -763,6 +763,7 @@ public class RubyRange extends RubyObject {
         if (rangeLe(context, begin, obj) == null) {
             return context.fals; // obj < start...end
         }
+        if (isEndless) return context.tru;
         return context.runtime.newBoolean(isExclusive
                 ? // begin <= obj < end || begin <= obj <= end
                 rangeLt(context, obj, end) != null : rangeLe(context, obj, end) != null);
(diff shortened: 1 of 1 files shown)


enebo:
@nomadium hey fixed the :fast problem with spec runs...should be a new batch of errors!

nomadium:
@enebo Excellent, thanks for fixing it! I was a little puzzled because the error message is not evident at all (at least for me).

enebo:
@nomadium once I looked at the error message and the fact we have the same exact string on master I realized our spec config file had the ruby version in the name....seems somewhat unneeded in how we test now but in the past we would test multiple ruby versions in same release.

PR #5564, opened 2019-01-18
Fix time precision with fractional seconds as Rational

### LINKED ISSUE #5558: Time precision problem
### Environment

* JRuby 9.1.17.0, 9.2.5.0
* JDK:
    ```
    java version "1.8.0_192"
    Java(TM) SE Runtime Environment (build 1.8.0_192-b12)
    Java HotSpot(TM) 64-Bit Server VM (build 25.192-b12, mixed mode)
    ```

### Problem

Running AR-JDBC tests, we have intermittent failures in Rails Time precision tests. See https://github.com/jruby/activerecord-jdbc-adapter/pull/979 for the test exclusion

Extracted test follows. MRI (tested with 2.4.4 and 2.5.1) runs w/o failures, JRuby fails usually between 1000-2000 iterations. 

### Reproducer

The script contains all relevant parts from Rails (5.2.2 but 5.0.x is affected too), no dependencies

```ruby
# what active_record's Time.change does internally
def change_time_usec(time, new_usec)
  new_sec = time.sec
  new_sec += Rational(new_usec, 1000000)
  Time.new(time.year, time.month, time.day, time.hour, time.min, new_sec)
end

def change_time_nsec(time, new_nsec)
  change_time_usec(time, Rational(new_nsec, 1000))
end

# from 'activemodel/lib/active_model/type/helpers/time_value.rb'
def apply_seconds_precision(value, precision)
  number_of_insignificant_digits = 6 - precision
  round_power = 10**number_of_insignificant_digits
  new_usec = value.usec - value.usec % round_power
  change_time_usec(value, new_usec)
end

# basically what
#   activerecord/test/cases/date_time_precision_test.rb
# does in #test_datetime_precision_is_truncated_on_assignment
# note: need to repeat and sleep to reproduce, but not 100% reliable even then
5000.times do |i|
  now = Time.now
  time = change_time_nsec(now, 123456789)
  time2 = apply_seconds_precision(time, 6)

  unless time2.nsec == 123456000
    puts "fail after #{i} iterations"
    puts now.strftime "%H:%M:%S.%N"
    puts time.strftime "%H:%M:%S.%N"
    puts time2.strftime "%H:%M:%S.%N"
    break
  end

  sleep 0.01
end
```

dr-itz:
Fixes #5558

### DIFF
--- a/core/src/main/java/org/jruby/RubyTime.java
+++ b/core/src/main/java/org/jruby/RubyTime.java
@@ -1743,13 +1743,26 @@ public class RubyTime extends RubyObject {
 
             // 1.9 will observe fractional seconds *if* not given usec
             if (args[5] != context.nil && args[6] == context.nil) {
-                double secs = RubyFloat.num2dbl(context, args[5]);
-                if (secs < 0 || secs >= TIME_SCALE) {
-                    throw runtime.newArgumentError("argument out of range.");
+                if (args[5] instanceof RubyRational) {
+                    RubyRational rat = (RubyRational) args[5];
+                    if (rat.isNegative()) {
+                        throw runtime.newArgumentError("argument out of range.");
+                    }
+                    RubyRational nsec = (RubyRational) rat.op_mul(context, runtime.newFixnum(1_000_000_000));
+                    long full_nanos = nsec.getLongValue();
+                    long millis = full_nanos / 1_000_000;
+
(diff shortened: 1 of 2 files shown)


kares:
looking good, thanks Daniel ... targeting for next 9.2
@enebo for reference this is what @dr-itz has been hitting on AR-JDBC (tagging) from the Rails suite

PR #5566, opened 2019-01-20
Update specs

eregon:


eregon:
I tagged a few failures, should be green now, merging.

PR #5587, opened 2019-01-31
Time: minimal fix for Rails Time.change() problem

dr-itz:
When there is no timezone (empty string), return nil like MRI.

See also jruby/activerecord-jdbc-adapter#980


enebo:
@dr-itz have you ever contributed to ruby/spec before?  because this one is screaming for a spec...

dr-itz:
@enebo no, but I'll look into it..there's more than one spec needed here...but it'll probably be not before next weekend or so...

PR #5602, opened 2019-02-10
[fix] properly calculate warning ($VERBOSE) default

kares:
noticed this while implementing the deprecation warning `Object#=~` for 2.6
the 3 state warning level was only using true/false states so it ended up printing warnings by default
JRuby should keep a `nil` (falsy) warning level by default, as Verbosity enum was intended to

no one complained about this on 9.1/9.2 so I am targeting 2.6 for now, its present on 9.2 as well

### DIFF
--- a/core/src/main/java/org/jruby/RubyGlobal.java
+++ b/core/src/main/java/org/jruby/RubyGlobal.java
@@ -189,7 +189,7 @@ public class RubyGlobal {
         }
 
         RubyInstanceConfig.Verbosity verbose = runtime.getInstanceConfig().getVerbosity();
-        IRubyObject verboseValue = null;
+        IRubyObject verboseValue;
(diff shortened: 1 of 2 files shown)


enebo:
@kares did you happen to find any tests about this tri-state nature?  I think we can land this in 2.5 if you want as well.  I agree this is probably pretty minor since no one reported but if we got some coverage it can definitely go in for 9.2.7.0.

kares:
have not looked for a test but due warnings from `Object#=~` being deprecated on 2.6 - it caused failures

enebo:
@kares then we can 2.6 and figure out if we want to cherry pick later.

headius:
Have we seen this happening with 9.2? It seems to be doing the right thing right now.

kares:
@headius confused, what is doing the right thing ... is this redundant than for 9.2 or shall we back-port?

PR #5607, opened 2019-02-12
Multiple adjustments to the Exception and raise logic.

### LINKED ISSUE #5605: set_backtrace should disable backtrace generation on raise
Reported by @jeremyevans while improving perf of no-trace flow control exceptions.

The following benchmarks should perform roughly the same. Instead, the one with `set_backtrace` performs more like a normal exception with a full backtrace.

3-argument raise:
```ruby
loop do
  t = Time.now
  1000000.times do
    begin
      raise(Exception, '', [])
    rescue Exception
    end
  puts Time.now - t
end
```

set_backtrace:
```ruby
loop do
  t = Time.now
  1000000.times do
    begin
      e = Exception.new
      e.set_backtrace([])
      raise e
    rescue Exception
    end
  puts Time.now - t
end
```

There's some logic missing in JRuby to avoid generating a backtrace if one is manually set.

headius:
* Request backtrace from exception object before raise, so that
  pre-set backtrace or overridden #backtrace skip native trace
  gathering.
* Eliminate some redundant or unused backtrace-gathering methods.
* Fix Kernel#warn uplevel logic to use partial traces (Java 9).

The fixes here allow two additional ways of blunting the cost of
raising an exception (by eliminating the native stack trace):

* Call Exception#set_backtrace before raising
* Use an Exception subtype that overrides Exception#backtrace

This improves the performance of these scenarios to be roughly
equivalent to the three-arg form of Kernel#raise.

Fixes #5605.

### DIFF
--- a/core/src/main/java/org/jruby/NativeException.java
+++ b/core/src/main/java/org/jruby/NativeException.java
@@ -84,14 +84,6 @@ public class NativeException extends RubyException {
         return exceptionClass;
     }
 
-    @Override
-    public void prepareBacktrace(ThreadContext context) {
-        // if it's null, use cause's trace to build a raw stack trace
-        if (backtraceData == null) {
-            backtraceData = WALKER.walk(cause.getStackTrace(), stream -> TraceType.Gather.RAW.getBacktraceData(getRuntime().getCurrentContext(), stream));
-        }
-    }
-
     @JRubyMethod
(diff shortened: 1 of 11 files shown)


headius:
The three failures are the same three from master's `-ea` hang in MRI suite. I'm investigating that on master.

PR #5613, opened 2019-02-12
Autoload fixes

headius:
This restores #5489 for additional work before merging.


headius:
Addressed separately by #5764.

PR #5614, opened 2019-02-12
Terminate and await signal executor to avoid races with at_exit.

headius:
This restores #5441 for additional work before it is merged or dropped.

### DIFF
--- a/core/src/main/java/org/jruby/Ruby.java
+++ b/core/src/main/java/org/jruby/Ruby.java
@@ -3304,49 +3304,8 @@ public final class Ruby implements Constantizable {
             context.pushScope(new ManyVarsDynamicScope(topStaticScope, null));
         }
 
-        // use signal-handling thread to avoid races with trap(INT) and similar shutdown signals
+        // terminate signal-handling thread to avoid races with at_exit
         ExecutorService executor = (ExecutorService) getModule("Signal").getInternalVariable("executor");
-
-        int[] _status = {status};
-        while (!atExitBlocks.empty()) {
-            RubyProc proc = atExitBlocks.pop();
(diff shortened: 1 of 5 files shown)


headius:
This looks like it might go green, but it was responsible for a verbose error being logged on master when a Fiber executor job was rejected.

headius:
The issues with the original PR are explained here: https://github.com/jruby/jruby/issues/5437#issuecomment-481033326

We'll get this into 9.2.8

headius:
@Adithya-copart I'm not sure where we left this. I filed these changes before you came back with https://github.com/jruby/jruby/issues/5437#issuecomment-515134631 and suggested the original issue be closed. Is the additional work in this PR no longer necessary?

headius:
Detargeting since it's unclear if this is appropriate or necessary. Will be closed without further follow-up.

Adithya-copart:
@headius This PR is no longer necessary as the main thread actually waits for the signal executors.

The issue I came across is that the main thread is being shutdown within the trap handlers. The at_exit blocks were partially executed which I mistakenly assumed to be a race condition between traps and at_exit handlers.

It was later fixed in puma [here](https://github.com/puma/puma/pull/1961/files#diff-cdee464b1408f50d2df33316479cd064L426-L431).

```
Signal.trap('INT') {p 'In trap'; Thread.main.kill; exit}

at_exit do
  # This gets printed
  p 'Inside at exit...'
  sleep 2
  # Never gets printed
  p 'Finished..'
end
p 'sleeping...'

sleep
```

headius:
@Adithya-copart Ok thank you. We are debating whether to include PR #5918 which might help such situations by warning that there's unterminated threads at exit.

PR #5615, opened 2019-02-12
Make autoload dynamically dispatch to require. Fixes #5403.

### LINKED ISSUE #5403: `autoload` should delegate to `require`

headius:
This restored #5404 for additional work before merging.


headius:
This was addressed by #5764.

PR #5623, opened 2019-02-21
Update specs

eregon:


eregon:
Specs pass, merging.

kares:
hey @eregon, have a related question - see there's changes around LeakChecker from these updates.
recently fiddled with it due importing MRI tests, seems useless except for a few checks (such as ENV changes), was wondering whether TruffleRuby is able to use it e.g. for thread leak detection at all?

eregon:
@kares Very far from useless, this is basically the only efficient way to find leaked threads and file descriptors (the worse). The changes add checks for top-level constants, so e.g. there is no constant `A` used by 2 specs in different ways.

I think TruffleRuby could run all these checks last time I tried, the thread check is just Thread.list and alive?. These checks are enforced in the ruby/spec CI.

kares:
great, thanks. leaking files detection relies on ObjectSpace (and Tempfile internals) but I guess JRuby could have a target for that. thread detection seemed to give a lot of false positives (listed threads that were dying), but if it works fine for TR I might look into improving thread's dead status. 
thanks again for the feedback and all your RubySpec work, you're the best maintainer the Ruby community could have hoped for.

eregon:
@kares Thanks for the praise :smiley:

headius:
@kares Thread leak detection tends to run amuck of adopted JVM threads (finalizer etc) that MRI doesn't have to deal with, and also depends on Thread lifecycle completing within some timeframe that's never enough for all environments. That's likely not an issue for other impls.

eregon:
@headius 
> Thread leak detection tends to run amuck of adopted JVM threads (finalizer etc). That's likely not an issue for other impls.

If the Ruby-level finalizer thread is already started when specs start running, then it should already work, as the check compares the set of threads before & after specs.
We could probably exclude the Ruby-level finalizer thread otherwise, PRs welcome.
Adoption of Java threads becoming Ruby threads should not happen when running ruby/spec specs, isn't it?

> It also depends on Thread lifecycle completing within some timeframe that's never enough for all environments, so there's always going to be some flaky cases.

Each spec must always `Thread#join` all threads (otherwise it's a bug of that spec), and the check helps to ensure that. There is no time limit, so it's fully reliable and fast.

headius:
> If the Ruby-level finalizer thread is already started when specs start running, then it should already work, as the check compares the set of threads before & after specs.

The "Ruby-level" finalizer thread is simply the JVM finalizer thread, and it may be adopted into JRuby well after the specs start running. We don't automatically add all threads at JRuby startup time into the Ruby thread list, since the purpose of that list is to indicate threads that are actually running Ruby code.

> Adoption of Java threads becoming Ruby threads should not happen when running ruby/spec specs, isn't it?

It can happen at any time if there's a Java thread that later calls into Ruby. Finalizer threads are the main issue, but global executors or other background worker subsystems could get adopted at any time, perhaps during spec runs. Adoption is done lazily because of the extra JRuby runtime structures we need to set up for a thread to execute Ruby code.

headius:
@eregon ^

eregon:
> The "Ruby-level" finalizer thread is simply the JVM finalizer thread

Right, so if a Ruby finalizer ran before starting spec, it'd still work, unless the Thread is adopted and unadopted every time? I think it's OK to make an exception for the finalizer thread.

> It can happen at any time if there's a Java thread that later calls into Ruby.

But none of ruby/spec does that, isn't it? If other specs/tests do, it's likely a good idea to have some way to recognize these adopted threads because they are somewhat different than Ruby-created Threads, and then we could just filter them out.

PR #5625, opened 2019-02-22
[fix] DateTime (civil) Float sec rounding

kares:
noticed a lot of (new) DateTime failures on 2.6 branch from CSV tests
turns out they were all related to sec arg rounding when a `Float` is given.

this patch will handle such cases ~ as MRI, its not matching C logic but is quite close.


kares:
there seems to be a problem with `eql?` which fails in some cases (when `==` passes), looking into that

PR #5627, opened 2019-02-25
Rework method invocation to properly super

headius:
This PR reworks most of our method invocation pipeline to follow MRI's way to set up the frame class.

In MRI, when searching for a method in a class hierarchy, the method is returned along with the class in the hierarchy that held it. In the case of modules, this would be the included module wrapper inserted into the class hierarchy. By returning the wrapper, the "frame class" points at the proper place from which to start "supering" to the next level in the hierarchy.

JRuby's logic previously had to perform a search against the hierarchy every time a module method performed super, since we set up the frame class pointing at the module itself.

With the new logic, the CacheEntry returned from RubyModule.searchWithCache contains an additional reference to the appropriate class in the hierarchy from which to super. This eliminates the hierarchy search in all but a few places and should fix other bugs with double included modules, refinements, and prepends/includes.


core/src/main/java/org/jruby/internal/runtime/methods/AliasMethod.java
```diff
@@ -32,86 +32,116 @@
...
+        super(implementationClass, entry.method.getVisibility(), oldName);
+
+        this.entry = entry;
+
+        boolean findImplementer = true;
+
+        // This logic is an attempt to reduce the number of cases that must do an implementer search,
+        // since it is only needed for super calls (and possibly other features that require the caller's
+        // frame class to be available).
+        // TODO: general support for DynamicMethod.needsClass etc, so we can easily make this deterimnation.
```
olleolleolle:
```suggestion
        // TODO: general support for DynamicMethod.needsClass etc, so we can easily make this determination.
```

core/src/main/java/org/jruby/RubyModule.java
```diff
@@ -1763,6 +1749,20 @@ public DynamicMethod searchMethodInner(String id) {
...
+    public CacheEntry searchMethodEntryInner(String id) {
+        int token = generation;
+        // This flattens some of the recursion that would be otherwise be necessary.
+        // Used to recurse up the class hierarchy which got messy with prepend.
+        for (RubyModule module = this; module != null; module = module.getSuperClass()) {
+            // Only recurs if module is an IncludedModuleWrapper.
+            // This way only the recursion needs to be handled differently on
+            // IncludedModuleWrapper.
+            DynamicMethod method = module.searchMethodCommon(id);
+            if (method != null) return method.isNull() ? null : cacheEntryFactory.newCacheEntry(id, method, module, token);
```
enebo: the module.searchMethodCommon in RubyModule will return null, NullMethod, *Method, but in IncludedModuleWrapper.searchMethodCommon we look for Null method so it is only null, *Method.  I think we should have all versions behave as latter since we reduce number of cases we need to consider.

headius: Yeah I was just mimicking the existing logic throughout this stuff but it should be audited to see whether we still need all these representations of null.

headius: It looks like @tduehr introduced NullMethod when he did the first prepend impl for JRuby. @tduehr Was this just an attempt to start eliminating null from the searchMethod* logic? There's still only the one place that ever returns NullMethod so I'm wondering whether it's worth keeping.

tduehr: I needed to differentiate between a failed search and an empty result to break a recursion loop or something like that. I don't remember exactly what, but it shouldn't be too hard to swap it out for a null check and see if that breaks anything.

PR #5628, opened 2019-02-25
Set up CI with Azure Pipelines

headius:
This is a test PR to see if Azure Pipelines would make some sense for us to start using.


headius:
Azure seems to work pretty well. I'm going to merge this and perhaps we can start migrating some jobs off Travis to cut down on total CI time.

PR #5629, opened 2019-02-26
Deploy snapshots after successful builds

headius:
With the loss of Cloudbees' devops tooling, we have no place to host nightly builds. Our simplest option is probably to just deploy snapshots on every successful build, so there's always a current JRuby snapshot available. This PR attempts to do that on Travis.


headius:
Deleted merge and redid-it as a squashed merge in 1444fd63f5.

PR #5633, opened 2019-02-27
Always-on trace events

headius:
This PR attempts to make trace events always available, but low or zero cost when the user is not using them. The main changes are to make LineNumber operations also fire trace events (when someone's listening) and to eliminate all need for TraceInstr.

It will not be possible to fully eliminate a flag, since methods like `set_trace_func` provide the binding when the event fires, which we usually optimize away in jitted code. However, we can provide most of the rest of the event data and warn if the binding is accessed. With deoptimization, we may be able to provide the binding on demand as well.

We can make tracing nearly free in the JIT by making the events use invokedynamic; it will be the cost of a load from the current frame and a call that should both optimize away when not needed.


headius:
So the experiment seems largely successful. Without --debug most of the tracing and coverage tests pass; all of them pass under the interpreter. Performance of the red/black does not seem affected in either interpreter or compiled code, which seems to show that indy is doing its job.

However there's a few factors that keep me from moving forward with this:

* The TraceInstr I merged into LineNumberInstr had side effects, since it was always possible that it would trigger a call to arbitrary code. This limits how much optimization occurs across these boundaries, and prevents traces from being coalesced or elided. LineNumberInstr is not marked for side effects, and as a result it can be optimized away (this may explain why I can't get the JIT to run green).
* The indy trace logic reduces the bytecode for each trace call to an ALOAD followed by an INVOKEDYNAMIC, and the default no-op logic appears to optimize away. However, we have to insert these extra bytecodes for every line, as well as entry and exit events. Peak performance does not appear to be affected, but it feels like a lot to push on to the JVM and expect good startup and warmup.
* There are no known real-world libraries that use trace events in a production setting. It remains almost exclusively for tool use, which to me reduces the value of having it always on. The complexity of the code we feed the JVM and the demands we put on the optimizer may penalize non-trace usage more than the value of having tracing available.
* At the end of the day we still can't support the full tracing API. The binding provided to set_trace_func's proc and TracePoint#binding requires us to deoptimize most scopes so we can capture their frames, and we are not yet in a position where we can do that.

PR #5645, opened 2019-03-07
Only run deploy against actual pushes to our repo

headius:
It was also running for PRs against the master and maintenance
branches.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -104,7 +104,8 @@ matrix:
       script:
         - cp .travis-maven-settings.xml $HOME/.m2/settings.xml
         - ./mvnw clean deploy -Prelease
-      if: branch IN (master, jruby-9.1)
+      if: branch IN (master, jruby-9.1) # only master and maintenance branch
+      type: push # only pushes to those branches
 
   allow_failures:
     - env: COMMAND=test/check_versions.sh


headius:
This can be merged if the build does not have the "deploy" stage while still in PR form.

kares:
wouldn't the artifact deployment push become problematic when doing a release ?

headius:
@kares I believe it would just push to the snapshot repo as normal with a release version. I'm not sure if that would be a problem... @mkristian may know more. It should be possible to filter the deploy stage to only try to deploy when version ends in SNAPSHOT.

headius:
Latest tweak does not have the "deploy" stage on PRs, so I'm merging it.

mkristian:
@headius @kares deploy with SNAPSHOT in version will use snapshot repo and without SNAPSHOT it will use staging repo. you will see it on sonatype webinterface how deployed it and when. it is still a human which closes the release and pushes it to maven-central. so all good IMO

kares:
@mkristian agreed, still would have been confusing if you as a release manager would see two deployed artifacts in staging - one from CI and another from you running deploy.

mkristian:
@kares agree also about the confusing bit. let me see if we can deploy on snapshots, not sure yet.

PR #5646, opened 2019-03-09
Remove deprecated NativeException logic

headius:
The logic here has been deprecated and not really useful for quite some time, but after removing it we discovered that activerecord-jdbc still uses one of these methods. I have restored the deprecated code on master, and this PR removes them again for a major release (9.3) rather than breaking binary compatibility with 9.2.x.


dr-itz:
`jruby-openssl` is affected too:

```
src/main/java/org/jruby/ext/openssl/ASN1.java:1432:            return RaiseException.createNativeRaiseException(context.runtime, cause);
```

kares:
and it seems like there won't be an easy transition - ideally we would use the newer APIs but we need an easy way to support older JRubies as well, well maybe not that much unless e.g. jossl can just drop <= 9.1
(AR-JDBC 5.x can also declare its required Ruby version compatibility)

@headius what do you feel is a preferred API for constructing an arbitrary Ruby exception from Java?
... there's `exceptionClass.newInstance().toThrowable()` but that is fairly recent addition (only in 9.2)

headius:
@kares For most of the standard exceptions there's `runtime.newBlahException`. For re-raising a Java Throwable as a Ruby exception, I think we do need a nicer function, but in many cases the right thing to do is just keep the Throwable as a cause and throw a more appropriate Ruby or library-specific exception.

This PR only addresses the APIs that create NativeException, which has been deprecated for many years. The NativeException class and logic to rescue it (which also rescues and lazily wraps any java.lang.Throwable) would need to remain in place as long as there's any code out there rescuing NativeException.

headius:
I guess an alternative to removing these methods would be to make them just re-throw the Java exception in a Ruby-compatible way. Thoughts on that @enebo?

kares:
trying to throw custom Ruby exceptions ... from Java land, that was the use case (+setting up cause) e.g. raising an ActiveRecord::ActiveRecordError

dr-itz:
Something like this for AR-JDBC? https://github.com/jruby/activerecord-jdbc-adapter/pull/990

headius:
@kares @enebo Yeah my thought is that since this used to be the catch all exception-creator for Java exceptions, we could keep the same interface and make it just raise the Java exception instead of wrapping it in NativeException.

My justification is that everywhere you're currently rescuing NativeException it will still rescue the actual Java exception (as a NativeException, no less), but we can start migrating to proper handlers instead.

A related idea would make NativeException a module that java.lang.Throwable includes. The two would be mostly indistinguishable except that the natural type would now be the actual Java type, rather than NativeException.

enebo:
Can I understand the use cases better?  I think we have:

 1. Create Ruby exception from Java of a specific type (AR::SomeError)
 1. Wrap the cause in a Ruby exception for some purpose (maybe to prevent higher stacks from seeing original Java exception)

If we just assume we allow any Java exceptions to bubble into Ruby space then no API is needed but it is obviously more than that.  Right?

headius:
The PR by @dr-itz seems like a good change for ARJDBC and I believe @enebo has merged it.

My remaining ideas are for this PR... what should we do with these deprecated methods?

1. Delete them as in this PR today
2. Modify them to just provide the "raise"able Java exception in a Ruby form, as if you did `raise NullPointerException.new`

The former would break any ext code that still relies on these methods. Hopefully there's not much.

The latter would maintain API compatibility...mostly. It would not be raising NativeException, but the real exception it does raise is still rescuable by NativeException. Safe enough behavior change?

And the related idea to go with (2) would be to make NativeException into a module that all Java exceptions "implement" so they can be interchangeable with any code that handles either NativeException or the real one.

enebo:
Extra note...We probably should think specifically about what old users were using this method for and decide what APIs if any satisfy them.  @dr-itz  arjdbc code was pretty reasonable replacement for arjdbcs use where we just set the cause.  What is openssl's need?

headius:
> If we just assume we allow any Java exceptions to bubble into Ruby space then no API is needed but it is obviously more than that. Right?

We do allow this now...Java exceptions not otherwise handled in Java can be rescued in Ruby as themselves. They can also be rescued using "Exception" or "Object" or "NativeException", with the last case lazily wrapping them in a NativeException for backward compatibility.

It will be a case-by-case decision whether the Java exception is the right one to propagate. My reading of the ARJDBC code is that those exceptions are "fatal" and there's not really any value in propagating them as the Java version. RuntimeError + setCause seems appropriate there.

Other cases using newNativeRaiseException may intend to handle the exception *as its Java form* from Ruby. That can be supported without creating an actual NativeException by propagating the original Java exception as-is.

kares:
change set seems empty, has this been merged already?

headius:
@kares The commit still seems valid but I'm not sure why it shows zero files changed. For example, the only commit removes RaiseException.createNativeRaiseException, which is definitely still there on master.

I guess I will merge it and if we have to revert the revert again we'll know why.

PR #5647, opened 2019-03-09
Outline case bodies from unpack logic.

headius:
This mitigates a performance regression in GraalVM 0.13 where our
main unpack switch does not JIT for heavy use of unpack('m').

I had to do an additional method split in order to get things to
compile properly, so there's still a regression in GraalVM that
that needs to be addressed, but this at least gets a benchmark of
unpack('m') to run with similar performance to Hotspot.


headius:
Strangely the auto-refactoring seems to have introduced a regression. I'll give it another shot.

PR #5651, opened 2019-03-13
Additional pack/unpack optimizations

headius:
A couple more items for pack/unpack:

* Outline cases in pack logic to aid JVM JITs that have trouble with large switches or large methods.
* Reduce complexity of String#unpack1 to a single block, rather than using catch/throw.

This may produce a small improvement in pack performance, but mostly it's to preemptively deal with JVMs that have trouble with large switches or methods, and also to make the code a bit easier to read.


headius:
See #5647

Issue #4536, opened 2017-03-17
Failing "-x" command line option


donv:
Given a script `mixed_script`:

```ruby
#!/bin/sh
echo 'bash part'
'exec' "ruby" '-x' "$0" "$@"
#!ruby
puts 'ruby part'
```

Run with

```
./mixed_script
```

With MRI installed as current Ruby this produces

```
bash part
ruby part
```

With JRuby this produces

```
bash part
jruby: no Ruby script found in input (LoadError)
```

### Environment

Darwin Gratass.local 16.4.0 Darwin Kernel Version 16.4.0: Thu Dec 22 22:53:21 PST 2016; root:xnu-3789.41.3~3/RELEASE_X86_64 x86_64
jruby 9.1.8.0 (2.3.1) 2017-03-06 90fc7ab Java HotSpot(TM) 64-Bit Server VM 25.112-b16 on 1.8.0_112-b16 +jit [darwin-x86_64]
jruby 9.1.5.0 (2.3.1) 2016-09-07 036ce39 Java HotSpot(TM) 64-Bit Server VM 25.112-b16 on 1.8.0_112-b16 +jit [darwin-x86_64]
jruby 1.7.26 (1.9.3p551) 2016-08-26 69763b8 on Java HotSpot(TM) 64-Bit Server VM 1.8.0_112-b16 +jit [darwin-x86_64]

donv:
I could use some help with this.  Is is preventing my next deployment.

I can find where the error message comes from:

https://github.com/jruby/jruby/blob/b3c3613df6affdb8db1b891e60ad48c5cd2c5ced/core/src/main/java/org/jruby/Main.java#L299

This is based on `RubyInstanceConfig#hasShebangLine` which is set here:

https://github.com/jruby/jruby/blob/b1e84356a81097da873e437201f9bcf9540bd301/core/src/main/java/org/jruby/RubyInstanceConfig.java#L284

That is in `RubyInstanceConfig#parseShebangOptions`, but I cannot find where this is called.  Maybe nowhere?

kschoelhorn:
I also just stumbled upon this (exactly two years later :grin:).

[...]

PR #5659, opened 2019-03-21
Fix -x command line option to skip header

kschoelhorn:
The -x option has been broken since 9e2d6dccb9e481fe25184202ee5a1762bb0fba28, as `parseShebangOptions` is no longer run, which is responsible for calling `setHasShebangLine`. As this is no longer done, `hasShebangLine` always returns false, even if there is a valid shebang.

The fix is to remove the check at this location and move it into the `findScript` function, which already searches for the ruby shebang.

Like mentioned in #4536, there seems to be a test for this, but it is not run during CI (at least I couldn't find it). Also parseShebangOptions seems to be dead code, which I can also remove if you want.

### DIFF
--- a/core/src/main/java/org/jruby/RubyInstanceConfig.java
+++ b/core/src/main/java/org/jruby/RubyInstanceConfig.java
@@ -424,9 +424,18 @@ public class RubyInstanceConfig {
     private static InputStream findScript(InputStream is) throws IOException {
         StringBuilder buf = new StringBuilder(64);
         BufferedReader br = new BufferedReader(new InputStreamReader(is));
-        String currentLine = br.readLine();
-        while (currentLine != null && !isRubyShebangLine(currentLine)) {
-            currentLine = br.readLine();
+
+        boolean foundRubyShebang = false;
+        String currentLine;
+        while ((currentLine = br.readLine()) != null) {
+            if (isRubyShebangLine(currentLine)) {
+                foundRubyShebang = true;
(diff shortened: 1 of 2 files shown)


core/src/main/java/org/jruby/RubyInstanceConfig.java
```diff
@@ -424,9 +424,18 @@ public InputStream getScriptSource() {
     private static InputStream findScript(InputStream is) throws IOException {
         StringBuilder buf = new StringBuilder(64);
         BufferedReader br = new BufferedReader(new InputStreamReader(is));
-        String currentLine = br.readLine();
-        while (currentLine != null && !isRubyShebangLine(currentLine)) {
-            currentLine = br.readLine();
+
+        boolean foundRubyShebang = false;
+        String currentLine;
+        while((currentLine = br.readLine()) != null) {
```
olleolleolle: _One tiny thing_: This `while` statement has different whitespace.

It needs a space after the `while` keyword to keep it the same.

Same goes for the `if` statements below.

(Thanks for working to imrpove JRuby!)

kschoelhorn: Oops, you are totally right, fixed!

headius:
Failures looked like some intermittent issues we've had on all branches, so I think we can go forward with this. Did not make 9.2.7, but it's in for 9.2.8!

PR #5666, opened 2019-03-27
Inline first-level loads into "kernel" Ruby files.

headius:
This adds a trivial inliner to the build that reads in some of our
"kernel" Ruby source files and inlines any top-level "load" calls
with the target file's source. This reduces the number of loads
executed during JRuby boot from over twenty down to just five.

### DIFF
--- a/core/pom.rb
+++ b/core/pom.rb
@@ -154,6 +154,21 @@ project 'JRuby Core' do
                      'executable' =>  'java',
                      'classpathScope' =>  'compile' )
     end
+
+    plugin 'org.codehaus.mojo:exec-maven-plugin' do
+      execute_goals( :exec,
+                     id: 'condense-kernel-files',
+                     arguments: [ '-classpath',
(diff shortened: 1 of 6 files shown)


headius:
Startup time improvement seems in the neighborhood of noise, but it may show up more on a slow filesystem.

kares:
yep I also tried loading less (back a while) but it did not seem to make much difference ...

headius:
So I guess we can debate whether this is worth the added build complexity. I have not been able to measure much difference so far.

enebo:
It is not much code but I don't see much point if we cannot see a measurable difference.

Two thoughts on the basic idea though:
 1. if we could punch out all extraneous comments and whitespace but still leave lines intact for backtraces
 1. A full replay of kernel into the entire hierarchy as a single source stream

With either of those I may still not be so into them if it was like 1% improvement.  I feel like full AOT + AppCDS seems like a path we should be considering more now too.

PR #5668, opened 2019-03-28
Update specs

eregon:


eregon:
I added tags for the 10 new failing specs, it should be green now, so I'll merge.
BTW, there are 3 failing `autoload` specs, with the complex `require` of an `autoload` path case, curious to see how you solve that one (https://bugs.ruby-lang.org/issues/15663 discusses it a bit).

Issue #5671, opened 2019-03-29
Thread leak with the `biz` gem


guizmaii:
### Environment

```
➜  jruby -v
jruby 9.2.6.0 (2.5.3) 2019-02-11 15ba00b OpenJDK 64-Bit Server VM 11.0.2+9 on 11.0.2+9 +jit [darwin-x86_64]
```

```
➜  uname -a
Darwin Juless-MacBook-Pro 18.2.0 Darwin Kernel Version 18.2.0: Thu Dec 20 20:46:53 PST 2018; root:xnu-4903.241.1~1/RELEASE_X86_64 x86_64
```

`biz (1.8.2)` and `Rails 4.2.11.1`

The code: https://github.com/zendesk/biz/blob/master/lib/biz/periods/linear.rb#L23-L37
 creates an Enumerator which will instantiate a thread (actually a Fiber backed by a thread) and that fiber is never released.

headius:
I did a little experimenting with brutally creating fibers and Enumerator#next threads using a snippit of code similar to this:

Fiber:
```ruby
loop {
  e = Fiber.new {
    1.times { Fiber.yield 1 }
  }
  e.resume
  e.resume
  JRuby.gc
}
```

Enumerator:
```
loop {
  e = Enumerator.new {|y|
    1.times { y << 1 }
  }
  e.next
  JRuby.gc
}
```

In the Fiber case, we "resume" twice because the first time starts the fiber and yields "1", which pauses the fiber. The additional "resume" allows the fiber to terminate. Running with this code yields no thread leaks...only one worker thread is ever used because the fibers finish and return the thread to its pool.

If I remove one or both `resume` calls, the fibers *do* still get cleaned up, but this loop appears to be too fast for the JVM to keep up with. In a normal application I'd expect it to keep up better, but it's not advisable to spin up scads of fibers very rapidly on JRuby.

The Enumerator#next version only calls `next` once, because a subsequent all would raise `StopIteration` and that kills the loop. With the code as shown, it's just like the case of a Fiber that has not been allowed to completely drain: the old enums are getting collected and the threads are getting returned to the pool, but not rapidly enough to keep up with this benchmark.

[...]

PR #5672, opened 2019-03-30
Use pure-Ruby Fiber-based Enumerator#next

headius:
This code has been passed around multiple implementations as a
pure-Ruby version of Enumerator#next that just uses fibers. We
have had many issues with the separate fiber-like impl we used
for Enumerator#next, and the logical path forward is to unify that
with Fiber via this code.

There are a few caveats to this migration:

* Three new introduced failures appear to be shared with other
  impls using this logic (rbx, truffleruby).
* This commit breaks the fast-path Array logic for Enumerator#next
  so it's always going through a thread. However the protocol here
  would allow us to implement #to_generator on any object, which
  should make it possible to short-circuit more easily (in Ruby!).
* I have not measured performance. However the logic surrounding
  the Fiber calls should not be particularly heavy.
* I have not vetted the cleanup side of this, except to confirm
  that this implementation seems about as good at jruby/jruby#5671
  in that it does appear to GC, but for a high-churn benchmark it
  eventually collapses without explicit GCs. So...no worse?

This code is modified from TruffleRuby's copy of Rubinius's code,
with Rubinius's license being very liberal and TruffleRuby's being
identical to JRuby's.

I am proposing this for 9.2.7 because it seems like mostly a lateral move, although there were some tag changes. There are at least two todos before release:

* Green build obviously.
* Implement at least a few "fast" generators for core collections like Array and Hash.


core/src/main/java/org/jruby/RubyEnumerator.java
```diff
@@ -519,91 +524,26 @@ public IRubyObject with_index19(ThreadContext context, IRubyObject arg, final Bl
...
-        return ensureNexter(getRuntime()).hasNext();
+        ThreadContext context = runtime.getCurrentContext();
+        return sites(context).next_p.call(context, this, generator).isTrue();
     }
 
     @Override
     public Object next() {
-        return next( getRuntime().getCurrentContext() ).toJava( java.lang.Object.class );
+        ThreadContext context = getRuntime().getCurrentContext();
+        return sites(context).next_p.call(context, this, this).toJava( java.lang.Object.class );
```
kares: you might have wanted `sites(context).next.call(...)` here to invoke `next` (not `next?`)

core/src/main/ruby/jruby/kernel/enumerator.rb
```diff
@@ -7,6 +7,100 @@ class Yielder
...
+    end
+
+    begin
+      return @__generator__.next if @__generator__.next?
+    rescue StopIteration
+      nil # the enumerator could change between next? and next leading to StopIteration
+    end
+
+    exception = StopIteration.new 'iteration reached an end'
+    JRuby.ref(exception).result = @__generator__.result
```
kares: a bit ugly maybe worth extending `StopIteration.new` 
or having a `StopIteration.__new__('msg', result)`?
... you wouldn't need to have `RubyStopIteration#setResult` around

headius: Yeah this is troublesome. The result is not supposed to be mutable through normal Ruby APIs, if we want to be strictly compatible, but there's also no way to construct a StopIteration with a specific result from Ruby. Both the Rubinius and TruffleRuby versions of this code do something basically the same. I went with what was simplest for now.

core/src/main/java/org/jruby/RubyEnumerator.java
headius: You are correct sir, good catch. Must not be tested.

headius: Looks like this is tested and failed. I'll finish up the PR and make sure it's all green.

Issue #5650, opened 2019-03-12
decimal performance


stevenjwolfman:
### Environment

Provide at least:
* jruby 9.2.5.0 (2.5.0) 2018-12-06 6d5a228 Java HotSpot(TM) 64-Bit Server VM 25.121-b13 on 1.8.0_121-b13 +jit [darwin-x86_64]
* Darwin ip-192-168-30-11.ec2.internal 18.0.0 Darwin Kernel Version 18.0.0: Wed Aug 22 20:13:40 PDT 2018; root:xnu-4903.201.2~1/RELEASE_X86_64 x86_64

When implementing the Haversine formula (http://stackoverflow.com/a/12969617/3443587), if I converted latitude and longitude to a decimal before running it, it took more than 150 times longer than if I left it as a float. This seems excessive and an opportunity for tuning.

This slow performance was not found in JRuby 1.7

For just multiplying simple numbers
irb(main):002:0> Benchmark.measure{1000.times{5.0*5.0}}
#<Benchmark::Tms:0x41cec7b0 @stime=0.0, @label="", @cstime=0.0, @real=0.000554911996005103, @total=0.0, @cutime=0.0, @utime=0.0>
irb(main):003:0> Benchmark.measure{1000.times{5.0.to_d*5.0.to_d}}
#<Benchmark::Tms:0x6773c073 @stime=0.009999999999999787, @label="", @cstime=0.0, @real=0.015367278996563982, @total=0.04999999999999183, @cutime=0.0, @utime=0.03999999999999204>

stevenjwolfman:
This performance was not found in JRuby 1.7

kares:
that `measure` is really dump simple - you should usually measure longer times with JRuby.
`Decimal#*` seems to be on par with MRI - very likely to have the same performance under 1.7 :
```ruby
require 'benchmark'

TIMES = (ENV['TIMES'] || 100_000_000).to_i

Benchmark.bmbm do |bm|
  bm.report("int") { int = 5  ; TIMES.times { int * int } }
  bm.report("flt") { flt = 5.0; TIMES.times { flt * flt } }
  bm.report("dec") { dec = BigDecimal('5.0'); TIMES.times { dec * dec } }
end
```

MRI 2.5 : 
```
Rehearsal ---------------------------------------
int   2.540000   0.000000   2.540000 (  2.146279)
flt   2.730000   0.090000   2.820000 (  2.500193)
dec   9.440000   0.080000   9.520000 (  9.129224)
[...]

PR #5675, opened 2019-04-02
speedup BigDecimal division

kares:
in certain cases its really slow e.g. a `x / 2` as discovered from https://github.com/jruby/jruby/issues/5650

```ruby
TIMES = (ENV['TIMES'] || 10_000_000).to_i

Benchmark.bmbm do |bm|
  bm.report("/2 flt") {
    rad_per_deg = Math::PI/180
    loc1 = [46.3625, 15.114444]
    loc2 = [46.055556, 14.508333]
    val = ((loc2[1]-loc1[1]) * rad_per_deg)
    TIMES.times { val / 2 }
  }

  bm.report("/2 dec") {
    rad_per_deg = Math::PI/180
    loc1 = [BigDecimal('46.3625'), BigDecimal('15.114444')]
    loc2 = [BigDecimal('46.055556'), BigDecimal('14.508333')]
    val = ((loc2[1]-loc1[1]) * rad_per_deg)
    TIMES.times { val / 2 }
  }

  bm.report(" 1/1280") {
    one = BigDecimal(1)
    div = BigDecimal(1280)
    TIMES.times { one / div }
  }
end
```
here the `"/2 dec"` case is very slow - on JRuby **9.2.6** taking **400s**, while MRI takes **10s** : 
```
/2 flt       0.670000   0.030000   0.700000 (  0.351421)
/2 dec     444.870000   0.930000 445.800000 (434.352299)
 1/1280     38.520000   0.100000  38.620000 ( 36.827996)

/2 flt       0.433256   0.000000   0.433256 (  0.433262)
/2 dec       9.910286   0.000000   9.910286 (  9.910299)
 1/1280      7.118558   0.000000   7.118558 (  7.118567)
```
... with this patch JRuby gets close to MRI's performance (**over 30x improvement** for the above / 2 case) : 
```
/2 flt       0.740000   0.020000   0.760000 (  0.379111)
/2 dec      14.560000   0.240000  14.800000 ( 13.860334)
 1/1280      7.760000   0.030000   7.790000 (  6.865561)
```

Should be noted that **only division (backed by `java.math.BigDecimal#divide(another, context)`) seemed slow**, other decimal operations are either faster on on par with MRI. The 'new' division algorithm is based on Android's sources (ASF2 license) with a few tweaks and a major (in terms of performance) change of not down scaling exact results (remainder 0). Leaving that off to [`java.math.BigDecimal(BigInteger, int, MathContext)`](https://docs.oracle.com/javase/8/docs/api/java/math/BigDecimal.html#BigDecimal-java.math.BigInteger-int-java.math.MathContext-) which will reduce the scale `- n` (doing `BigInteger div 10^n`) faster than we can using public API.

### DIFF
--- a/core/src/main/java/org/jruby/ext/bigdecimal/Multiplication.java
+++ b/core/src/main/java/org/jruby/ext/bigdecimal/Multiplication.java
@@ -1,0 +1,111 @@
+/*
+ *  Licensed to the Apache Software Foundation (ASF) under one or more
+ *  contributor license agreements.  See the NOTICE file distributed with
+ *  this work for additional information regarding copyright ownership.
+ *  The ASF licenses this file to You under the Apache License, Version 2.0
+ *  (the "License"); you may not use this file except in compliance with
+ *  the License.  You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ *  Unless required by applicable law or agreed to in writing, software
+ *  distributed under the License is distributed on an "AS IS" BASIS,
+ *  WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ *  See the License for the specific language governing permissions and
+ *  limitations under the License.
+ */
+package org.jruby.ext.bigdecimal;
+
+import java.math.BigInteger;
+
+// NOTE: from Android's sources
+// https://android.googlesource.com/platform/libcore/+/refs/heads/master/luni/src/main/java/java/math/Multiplication.java
+
+/**
+ * Static library that provides all multiplication of {@link BigInteger} methods.
+ */
+class Multiplication {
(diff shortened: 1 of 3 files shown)


kares:
that the default delta is 0.01 or so - thus added an explicit delta to make those asserts meaningful

enebo:
@kares I just meant will we start seeing new errors from test suites like Rails which is assuming a particular rounding or something like that.  I am ok with that but I just am wondering what potential fallout is possible if we are not rounding exactly the same as we were.

kares:
we're not rounding exactly but weren't doing so before anyway. we do not match MRI with division, although have gotten better in 9.2 (compared to <= 9.1) and have some asserts guarding against regressions in terms of precision. so we should be doing the same (pbly even better regarding precision) as before.

enebo:
@kares thanks. lgtm!

PR #5677, opened 2019-04-04
Use close_on_exec method instead of using fcntl calls for pty. Fixes #5663

### LINKED ISSUE #5663: File::write uses nonblocking IO when MRI uses blocking
### Environment

* jruby 9.2.6.0 (2.5.3) 2019-02-11 15ba00b OpenJDK 64-Bit Server VM 25.201-b09 on 1.8.0_201-b09 +jit [linux-x86_64]
* RHEL 7.6

### Expected Behavior

MRI waits if nobody is reading the other side of a pty pair:
```
2.5.3 :001 > require 'pty'
 => true 
2.5.3 :002 > m,s = PTY.open
 => [#<IO:masterpty:/dev/pts/15>, #<File:/dev/pts/15>] 
2.5.3 :003 > 20.times{|i| puts "writing #{i}"; s.write("_" * 1000) }
writing 0
writing 1
writing 2
writing 3
writing 4
writing 5
writing 6
writing 7
writing 8
<hang until ctrl-c>
```

### Actual Behavior

However JRuby appears to use non-blocking IO and thus throws an exception:
```
jruby-9.2.6.0 :001 > require 'pty'
 => true 
jruby-9.2.6.0 :002 > m,s = PTY.open; nil
 => nil 
jruby-9.2.6.0 :003 > 20.times{|i| puts "writing #{i}"; s.write("_" * 1000) }
writing 0
writing 1
writing 2
writing 3
writing 4
writing 5
writing 6
writing 7
writing 8
writing 9
writing 10
writing 11
writing 12
        9: from /home/byteit101/.rvm/rubies/jruby-9.2.6.0/bin/irb:13:in `<main>'
        8: from org/jruby/RubyKernel.java:1179:in `catch'
        7: from org/jruby/RubyKernel.java:1179:in `catch'
        6: from org/jruby/RubyKernel.java:1411:in `loop'
        5: from org/jruby/RubyKernel.java:1047:in `eval'
        4: from (irb):3:in `evaluate'
        3: from (irb):3:in `block in evaluate'
        2: from org/jruby/RubyIO.java:1429:in `write'
        1: from org/jruby/RubyIO.java:1472:in `write'
SystemCallError (Unknown error (SystemCallError) - No message available)
jruby-9.2.6.0 :004 >
```
It took me a bit to figure out that the errno was `EAGAIN (Resource temporarily unavailable)` (not sure why that wasn't in the exception message, another issue?)

I found this as the surprise exception causes a compatibility issue with Pry in a Pty to exit on large output (think ENV inspection, ~12k on my machine) in JRuby, but not under MRI as MRI never throws here.

byteit101:
Fix as suggested in #5663. Provides same behavior as MRI now. 

Additionally, I ran this test: `jruby -r ./test/mri/lib/envutil.rb test/mri/test_pty.rb` and it passed. I wasn't sure if I needed to to anything else, or why https://github.com/jruby/jruby/blob/master/test/mri/excludes/TestPTY.rb suggests that it should fail.

This also doesn't address the opposite behavior of the nonblock flag as I noted in #5663 

Resolves #5663

### DIFF
--- a/lib/ruby/stdlib/pty.rb
+++ b/lib/ruby/stdlib/pty.rb
@@ -54,7 +54,7 @@ module PTY
       fds.each do |fd|
         fd.sync = true
 
-        hack_close_on_exec(fd)
+        fd.close_on_exec = true
       end
 
       return fds unless block_given?
@@ -106,8 +106,8 @@ module PTY
       pid = Process.spawn(*args, in: read, out: slave, err: slave, close_others: true, pgroup: true)
       [read, slave].each(&:close)
 
-      hack_close_on_exec(master)
-      hack_close_on_exec(write)
+      master.close_on_exec = true
+      write.close_on_exec = true
 
       ret = [master, write, pid]
 
@@ -130,16 +130,6 @@ module PTY
         return pwent.shell
       else
         "/bin/sh"
-      end
-    end
-
-    def hack_close_on_exec(fd)
-      # I assume this isn't actually supported for good reasons.
-      # but let's see how close to passing test_pty we can get.
-      fl = fd.fcntl(Fcntl::F_GETFL, 0)
-      fd.fcntl(Fcntl::F_SETFL, Fcntl::FD_CLOEXEC|fl)
-      fd.define_singleton_method(:close_on_exec?) do
-        true
       end
     end
 


headius:
👍 Thanks!

Issue #5656, opened 2019-03-18
Multicast Receive Broken On Windows 10


j2man:
### Environment
Logstash 6.6.2 Bundled JRuby
jruby 9.1.13.0 (2.3.3) 2017-09-06 8e1c115 Java HotSpot(TM) 64-Bit Server VM 25.162-b12 on 1.8.0_162-b12 +jit [mswin32-x86_64]

Windows 10 Enterprise 64-bit
RAM 32g
4 Ethernet Adapters

### Expected Behavior
JRuby should be receiving multicast data just like ruby receives multicast data. 
Verified data is received properly using ruby 2.5.3p105 (2018-10-18 revision 65156) [x64-mingw32]

### Actual Behavior
JRuby does not receive multicast data. 

JRuby multicast send works fine verified with wireshark and vanilla Ruby

################## Rx Code
require 'socket'
require 'ipaddr'
MCAST_GROUP = {
  :addr => '225.4.5.6',
  :port => 12345,
  :bindaddr => '0.0.0.0'
}
ip =  IPAddr.new(MCAST_GROUP[:addr]).hton + IPAddr.new(MCAST_GROUP[:bindaddr]).hton
sock = UDPSocket.new
sock.setsockopt(Socket::IPPROTO_IP, Socket::IP_ADD_MEMBERSHIP, ip)
sock.setsockopt(Socket::SOL_SOCKET, Socket::SO_REUSEADDR, 1)
sock.bind(Socket::INADDR_ANY, MCAST_GROUP[:port])
loop do
  msg, info = sock.recvfrom(1024)
  puts "MSG: #{msg} from #{info[2]} (#{info[3]})/#{info[1]} len #{msg.size}" 
end

################## Tx Code
require 'socket'
MCAST_GROUP = {
  :addr => '225.4.5.6',
  :port => 12345,
}
s = UDPSocket.new
s.setsockopt(Socket::IPPROTO_IP, Socket::IP_MULTICAST_TTL, 1)
s.send("hello", 0, MCAST_GROUP[:addr], MCAST_GROUP[:port])

j2man:
FYI I verified the bug still exists with the latest version of JRuby 9.2.6.0 poor logstash has no solution even if they bundle the latest release of JRuby.

enebo:
#5678 may or may not have solved this but since I am unable to not get the test to hang I am going with the assumption it is a problem of ours and not an environment issue.

[...]

PR #5678, opened 2019-04-05
multicast receive broken everywhere

enebo:
This is an attempt at fixing #5656.  I could not get the reporters server code to work on linux much less the other platforms referenced.  I believe it was something underneath this source line in original code:

```java
socket.multicastStateManager.getMulticastSocket().getChannel();
```

in Java source of getChannel() is literally 'return null'.  Based on changes over the years I suspect something changed in our implementation (like maybe we subclassed the channel?).  This version just uses the socket itself to receive and it blocks and also seems to work.  I also added the equivalent logic if somehow the socket is in a messed up state and tried to mark itself as non-blocking.

With that said I am not confident so I need some eyeballs...

### DIFF
--- a/core/src/main/java/org/jruby/ext/socket/RubyUDPSocket.java
+++ b/core/src/main/java/org/jruby/ext/socket/RubyUDPSocket.java
@@ -1,4 +1,5 @@
-/***** BEGIN LICENSE BLOCK *****
+/*
+ **** BEGIN LICENSE BLOCK *****
  * Version: EPL 2.0/GPL 2.0/LGPL 2.1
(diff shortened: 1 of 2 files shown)


headius:
Just to confirm, you tested that this works on Windows? I noticed the original test was masked on Windows.

enebo:
@headius I tried it and it does not seem to work so I should probably mask it out until I can figure that out.  This is plain old Java socket stuff so I half wonder if this ever worked on windows.  The reporter implied it used to work but I guess I need to try 1.7 and see.

PR #5679, opened 2019-04-06
Provide better thread name for enumerator/fiber workers.

### LINKED ISSUE #5670: Update fiber/enumerator threads with info about creation
Currently all fiber threads end up with a name something like "Ruby-Fiber-15" because we pool those thread objects and reuse them. However, this does not provide very good information about where or why the fiber thread was enlisted.

We could setName on those threads to update the name to something that reflects the origins of the fiber or enumerator using the given fiber thread, which would make it easier to debug when there's a leak or some other issue with those coroutine threads.

headius:
This patch sets the name of the threads used to drive fibers (or the similar Enumerator#next coroutines) in order to more easily audit and debug those threads. Relates to #5671 and fixes #5670.


headius:
There are a few minor changes here also:

* Actually loop when submitting the fiber runnable to the executor, with one attempt to GC.

Issue #5476, opened 2018-11-27
Mutex#lock and Mutex#synchronize are not interruptible by Thread#kill


eregon:
### Environment

* JRuby version: `jruby 9.2.5.0-SNAPSHOT (2.5.0) 2018-11-27 ebbb750 OpenJDK 64-Bit Server VM 25.171-b10 on 1.8.0_171-b10 +jit [linux-x86_64]`

Run:

```
ruby -e 'm=Mutex.new; m.lock; t=Thread.new{m.lock}; sleep 1; t.kill; t.join; p :done' 
```

### Expected Behavior

```
:done
```

### Actual Behavior

It hangs on JRuby.
3 Mutex specs have been tagged in https://github.com/jruby/jruby/pull/5475 to avoid hanging.

headius:
Huh. Must have regressed...I'm sure this worked before.

andy-twosticks:
Hangs for me on 9.1.8.0, 9.1.17.0, 9.2.0.0 and 9.2.1.0-dev.

I'd rather like this to work on 9.1.17.0, since that's the "stable" release.  I _think_ it might be tangentally responsible for a problem I have, where threads are being held onto between rspec sessions despite being supposedly killed at the end of the test?

eregon:
Couldn't the semantics just be try to lock in an interruptible manner, and if it is interrupted, check for `Thread#raise`, otherwise just try to lock (checking interrupts) again? That's essentially what we do in TruffleRuby.

headius:
@andy-twosticks: Do you have a real-world case freezing due to this? I am leaning toward going back to interruptible with some status check. As @eregon mentions, checking for an interrupt event may be sufficient, but I'm not sure I know all the cases that are expected to wake a lock-blocked thread.

headius:
@eregon The main concern I have with that approach is all the other JVM interrupt cases that aren't related to Ruby. Since we're a JVM language, we have to make decisions about whether to ignore non-Ruby interrupts or allow them to cancel the blocking operation being performed. There's no perfect answer.

We have had multiple user reports in the past where blocking operations were or were not interruptible via standard JVM mechanisms. Often we have to make a judgment call which way to go, but simply forcing Ruby semantics on users is not usually the first choice.

headius:
[...]

PR #5682, opened 2019-04-09
Allow contended Mutex to be interrupted by Ruby.

headius:
This PR includes a possible fix for #5476 wherein Ruby interrupts are allowed to wake a lock, but Java thread interrupts are considered spurious.

This should not be merged until we decide which of the various behaviors fits best.


headius:
Closing in favor of #5683.

PR #5612, opened 2019-02-12
Fix off-by-one error causing Dir.glob("foo" + "{bar}") to fail if Java asserts are enabled

ikaronen-relex:
When Java asserts are enabled, `Dir.glob` and `Dir[]` crash if given a pattern that ends in a curly brace. However, this _only_ happens if the pattern string has not been interned. (Specifically, the underlying ByteList must have `bytes.length == realSize` for the bug to trigger. Apparently that's not usually the case for interned string literals.)

Here is a simple command-line test case:

    JAVA_OPTS=-ea bin/jruby -J-ea -e 'Dir.glob("foo" + "{bar}")'

(I have no idea why _both_ `JAVA_OPTS=-ea` and `-J-ea` are needed to enable Java asserts, but that seems to be the case. That might be a separate issue.)

On JRuby 9.2.6.0, running the command above produces the following output:

```
Unhandled Java exception: java.lang.AssertionError: Invalid start: 8
java.lang.AssertionError: Invalid start: 8
               append at org/jruby/util/ByteList.java:542
          push_braces at org/jruby/util/Dir.java:473
            push_glob at org/jruby/util/Dir.java:312
                 glob at org/jruby/RubyDir.java:271
                 call at org/jruby/RubyDir$INVOKER$s$0$2$glob.gen:-1
                 call at org/jruby/internal/runtime/methods/DynamicMethod.java:204
                 call at org/jruby/internal/runtime/methods/DynamicMethod.java:200
         cacheAndCall at org/jruby/runtime/callsite/CachingCallSite.java:346
                 call at org/jruby/runtime/callsite/CachingCallSite.java:172
    invokeOther3:glob at -e:1
               <main> at -e:1
  invokeWithArguments at java/lang/invoke/MethodHandle.java:627
                 load at org/jruby/ir/Compiler.java:94
            runScript at org/jruby/Ruby.java:854
          runNormally at org/jruby/Ruby.java:777
          runNormally at org/jruby/Ruby.java:795
          runFromMain at org/jruby/Ruby.java:607
        doRunFromMain at org/jruby/Main.java:415
          internalRun at org/jruby/Main.java:307
                  run at org/jruby/Main.java:234
                 main at org/jruby/Main.java:206
```

[...]

PR #5693, opened 2019-04-12
pass -ea to JVM and fix asserts for test targets

kares:
... also incorporates https://github.com/jruby/jruby/pull/5612

there's one assert left that is failing related to frame dup-ing (for procs)


headius:
I believe I've fixed the null block duping on master, in 3095263.

headius:
ManyVarsDynamicScope is fixed on master.

headius:
I do not see any other failures on your branch that we haven't addressed, so I'm going to merge. We'll fix anything left on master, but hopefully it's green.

PR #5700, opened 2019-04-17
[ji] align Java array (with post 1.8 conventions)

kares:
- `j_ary = [ 1 ].to_java; j_ary[2]` shall raise `IndexError` (instead of `ArgumentError`)
- handle edge cases e.g. `[ "a", "b", "c" ].to_java[-10..2]` similar to a Ruby `Array` (`nil`)

the first one might be considered a bug as well as a breaking change -> maybe target 9.3?


headius:
This change seems ok to me even for a shorter-term merge. Someone *might* rely on it raising an error, but I think it would have to be a pretty weird case.

kares:
okay, its fine by me wherever it ends up.
@enebo any objections to have it next 9.2?

Issue #5699, opened 2019-04-16
sun.misc does not exist build issue.


FuncGuy:
Hi,
when i try to build jruby with jdk 11 getting a compiler error "java package sun.misc does not exist".
please have a look.

headius:
I can reproduce problems during the build.

```
Caused by: org.jruby.exceptions.RaiseException: (OpenSSL::X509::StoreError) setting default path failed: No password supplied for PKCS#12 KeyStore.
	at org.jruby.ext.openssl.X509Store.set_default_paths(org/jruby/ext/openssl/X509Store.java:165)
	at RUBY.SSLContext(uri:classloader:/META-INF/jruby.home/lib/ruby/shared/jopenssl19/openssl/ssl-internal.rb:31)
	at RUBY.SSL(uri:classloader:/META-INF/jruby.home/lib/ruby/shared/jopenssl19/openssl/ssl-internal.rb:22)
	at RUBY.OpenSSL(uri:classloader:/META-INF/jruby.home/lib/ruby/shared/jopenssl19/openssl/ssl-internal.rb:21)
	at RUBY.(root)(uri:classloader:/META-INF/jruby.home/lib/ruby/shared/jopenssl19/openssl/ssl-internal.rb:20)
```

We have attempted to fix this before: https://github.com/jruby/jruby/issues/4805, https://github.com/jruby/jruby/issues/4799

Looking at the backtraces, I can see that jruby-openssl is loading the 1.9-compatible logic, which means this could not possibly be using JRuby 9.x. Something in this process is attempting to use an older JRuby 1.7.x.

@mkristian You helped someone else stumbling into JRuby 1.7 in #4805, but in this case we are simply trying to build JRuby itself. I thought for sure we had full builds working under Java 9+.

This only appears to affect the "bootstrap" maven profile, so it should still be possible for you to build JRuby and get a usable distribution out of it. The "bootstrap" profile just installs tools necessary for working on JRuby itself.

headius:
@FuncGuy Please post full output of your build, adding -X to the mvn command line. That output will be very long, so please use gist.github.com or similar.

Now on a second attempt I can't get the error to fail. All I did in between was run `keytool -list -cacerts` and entered a blank password when prompted.

headius:
I cannot reproduce the sun.misc issue. We do reference that package in various places but it builds ok for me on Java 9+.

I still can't reproduce the keystore error.

FuncGuy:
[...]

PR #5702, opened 2019-04-18
#5699:adding the javax maven dependency to avoid "javax.annotation.Ge…

FuncGuy:
fixed the issue annotation not found compiler error.

### DIFF
--- a/core/pom.xml
+++ b/core/pom.xml
@@ -251,6 +251,11 @@ DO NOT MODIFIY - GENERATED CODE
       <artifactId>backport9</artifactId>
       <version>1.3</version>
     </dependency>
+      <dependency>
+          <groupId>com.google.code.findbugs</groupId>
+          <artifactId>jsr305</artifactId>
+          <version>3.0.2</version>
+      </dependency>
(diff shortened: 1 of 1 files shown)


headius:
👍

mkristian:
@FuncGuy dependency should be scope `provided` as we do not need this for runtime only for compile time.
BTW please change core/pom.rb, run build and then core/pom.xml will be generated with the changes from core/pom.rb.

headius:
Since @FuncGuy seems to have gone away I'll merge and fix @mkristian comments on master.

headius:
@FuncGuy Ok I was confused when I merged this...the changes in your PR don't actually do anything with the annotation API.

I did a subsequent commit that removed the dependency and plugin you added, and moved the annotation API dependency (which we already had) to the global space so it's active for all profiles of the build.

If this doesn't work you'll need to explain why these other two changes were needed. Please open a new issue for them that describes why you made them.

FuncGuy:
@headius Ok i will try to reproduce if at all possible i'm ok with your fix!!

PR #5703, opened 2019-04-18
[ji] to_time conversion for Java DateTime types

kares:
should be good to go, a further extension would be to have a `to_date` for Java date types


headius:
@kares This looks fine to me, and I know having stuff in Ruby hurts startup, but I am reluctant to keep adding these extensions in Java. Maybe it's not a huge deal...there's going to be a small number of core JDK types we want to be able to easily roundtrip.

kares:
it wont load on startup, we lazy add these as the Java class gets used. so adding more of these wont hurt startup performance

Issue #5709, opened 2019-04-25
Socket#close raises EBADF after unsuccessful connection


p0deje:
### Environment

* JRuby version (`jruby -v`) and command line (flags, JRUBY_OPTS, etc):

```
jruby 9.2.7.0 (2.5.3) 2019-04-09 8a269e3 Java HotSpot(TM) 64-Bit Server VM 25.172-b11 on 1.8.0_172-b11 +jit [darwin-x86_64]
```

* Operating system and platform (e.g. `uname -a`)

```
Darwin MacBoosha.local 18.2.0 Darwin Kernel Version 18.2.0: Thu Dec 20 20:46:53 PST 2018; root:xnu-4903.241.1~1/RELEASE_X86_64 x86_64
```

### Expected Behavior

I'm trying to check if the port `9250` is open and as a part of this exercise I execute the following code: 

```ruby
require 'socket'

sock = Socket.new(Socket::AF_INET, Socket::SOCK_STREAM)
addr = Socket.pack_sockaddr_in(9250, '127.0.0.1')

begin
  sock.connect_nonblock(addr)
rescue Errno::EINPROGRESS
  IO.select(nil, [sock], nil, 1) # select socket for writing
  sock.getsockopt(Socket::SOL_SOCKET, Socket::SO_ERROR) # read any errors from last connection
  retry
rescue Errno::ECONNREFUSED, Errno::EINVAL
  puts 'Open!'
end

sock.close
```

Running this code on MRI (`ruby 2.5.5p157 (2019-03-15 revision 67260) [x86_64-darwin18]`) raises no error:

```
$ ruby socket.rb
Open!
```

### Actual Behavior

Running the code on JRuby raises `Errno::EBADF` when it's trying to close socket:

```
$ ruby socket.rb
Open!
Errno::EBADF: Bad file descriptor - No message available
   close at org/jruby/ext/socket/RubySocket.java:694
  <main> at socket.rb:16
```

headius:
Hmm...our behavior seems more correct, to be honest. In your case the socket is not open (connection refused), so I'd expect `close` to error.

`sock.close rescue nil` is a simple workaround for the moment.

headius:
Yeah as usual I'm still missing some detail about when CRuby will raise this error and when it won't.

It doesn't appear to raise for repeat close on a Socket, pipe, or File:

```
[...]

PR #5713, opened 2019-04-25
Allow repeated close of any IO to succeed.

headius:
This relates to #5709 but breaks at least one spec:

```
File.open opens a file with a file descriptor d and a block FAILED
Expected Errno::EBADF but no exception was raised (nil was returned)
/Users/headius/projects/jruby/spec/ruby/core/file/open_spec.rb:173:in `block in <main>'
```

There may be other failures, but it's debateable whether they
represent behavior we need to emulate.


headius:
Looks like there's other failures, but I'm still unconvinced if the differing behavior is important.

PR #5716, opened 2019-04-25
Run rake-based test targets directly.

headius:
Running them through Maven causes an extra layer of buffering to
sometimes obscure where and when the suites hang, if they hang.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -44,54 +44,90 @@ matrix:
 
   include:
     - stage: test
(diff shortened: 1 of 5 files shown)


lopex:
So let's be clear here, layers of buffering or the suites are causing the problem ?

headius:
@lopex No, the buffering was just obscuring where the hang happened.

eregon:
Wooh, that's awesome :smiley: 
(I disliked spec running inside Rake inside Maven and Maven adding an `[exec]` prefix on every line, plus keeping Maven running for a while when there is no need)

PR #5719, opened 2019-04-27
Update specs

eregon:
* Specs for Ruby 2.3 are removed since MRI 2.3 is no longer supported.
* Timing specs should generally be more stable now with `TIME_TOLERANCE`.


eregon:
It looks like `Process.clock_gettime(Process::CLOCK_REALTIME)` gives microsecond-precise results in JRuby, but that `Process.clock_getres(Process::CLOCK_REALTIME)` pretends it's only millisecond-precise, that's most of the new tags added.

PR #5722, opened 2019-04-29
use same parameter for DNS.allocate_request_id/free_request_id to fix leak

colinsurprenant:
As described in logstash-plugins/logstash-filter-dns#52 

`resolv.rb` was updated from upstream Ruby stdlib starting at JRuby 9.2.0.0 which essentially re-introduced issue logstash-plugins/logstash-filter-dns#40 where all DNS request will eventually systematically time out.

This problem will only be triggered if there are multiple nameserver hosts specified OR if there is no nameserver host but there are more than one host in `/etc/resolv.conf` resulting in the usage of the `UnconnectedUDP` class in `resolv.rb`.

Essentially the problem is that the [`DNS.allocate_request_id(host, port)` call](https://github.com/jruby/jruby/blob/9.2.0.0/lib/ruby/stdlib/resolv.rb#L790) will cache the allocated id using the `host` and `port` parameters. 

But the cache cleanup will be done in the `close` method by the [DNS.free_request_id(service[0], service[1], id) call](https://github.com/jruby/jruby/blob/9.2.0.0/lib/ruby/stdlib/resolv.rb#L802) where the `service[0]` [will be `IPAddr.new(host)`](https://github.com/jruby/jruby/blob/9.2.0.0/lib/ruby/stdlib/resolv.rb#L789) and not the original `host` string used in the `sender` method leading to not finding the original cache key and not freeing it from the cache which is 64k elements wide and once full, the `allocate_request_id` method [will loop forever](https://github.com/jruby/jruby/blob/9.2.0.0/lib/ruby/stdlib/resolv.rb#L636-L638).

This fix just makes sure that the exact same parameters are used in both `DNS.allocate_request_id` and `DNS.free_request_id` using the `IPAddr#to_s` which will always yield the same host string. 

I am planning on submitting this issue upstream in Ruby but I figured I'd also submit here to provide a potential quicker turnaround for this nasty bug.

### DIFF
--- a/lib/ruby/stdlib/resolv.rb
+++ b/lib/ruby/stdlib/resolv.rb
@@ -783,7 +783,7 @@ class Resolv
           sock = @socks_hash[host.index(':') ? "::" : "0.0.0.0"]
           return nil if !sock
           service = [IPAddr.new(host), port]
-          id = DNS.allocate_request_id(host, port)
+          id = DNS.allocate_request_id(service[0].to_s, port)
           request = msg.encode
           request[0,2] = [id].pack('n')
           return @senders[[service, id]] =
@@ -795,7 +795,7 @@ class Resolv
             if @initialized
               super
               @senders.each_key {|service, id|
-                DNS.free_request_id(service[0], service[1], id)
+                DNS.free_request_id(service[0].to_s, service[1], id)
               }
               @initialized = false
             end


colinsurprenant:
I just checked in upstream Ruby and it actually [does not have this problem](https://github.com/ruby/ruby/blob/fc3e80cf6d8847fd5a2313dd7e2079d268f1ec84/lib/resolv.rb#L764-L786).  
This bug in the JRuby version of `resolv.rb` was actually introduced by https://github.com/jruby/jruby#4496

headius:
@colinsurprenant Ahh thanks for the footwork on this. We need to sort out what's the correct logic and do our best to sync it up between JRuby and CRuby.

headius:
@colinsurprenant Oh, so...if we pull CRuby's actual resolv.rb unmodified are we reintroducing that other issue?

colinsurprenant:
@headius yeah if you pull the latest CRuby `resolv.rb` it will revert #4496. I haven't really paid attention to what exactly #4496 does/fixes, I can certainly verify that and see if that fix is still needed and/or relevant in CRuby and to maybe submit it upstream?

headius:
@colinsurprenant Ah yes, @enebo refreshed my memory. So I believe #4496 was because our IPAddr and related structures were not quite matching CRuby...just another part of the complicated socket subsystem that we're not emulating exactly right. But things have changed in the meantime and a whole bunch of socket compat fixes got merged in within the last two years...we might be ok?

colinsurprenant:
@headius sounds good, I can certainly take a look!

headius:
@colinsurprenant Any status update on this? I don't have a problem merging this to fix JRuby but it would obviously be nicer if we could move toward MRI's version rather than further away.

colinsurprenant:
Dropped the ball on this :( 
Let's try for 9.2.9.0.

headius:
@colinsurprenant Ping again. 😁

I'm still unclear on whether we should merge this or update to a different version of MRI's resolv.rb (or whether we might already have the best version due to the [stdlib 2.7 update](#5917)  going into 9.2.9.)

headius:
Updating some details on my own...

The patch that @colinsurprenant proposes here does make our resolv.rb diverge further from the CRuby version. However the bug is caused by a separate divergence, and this just fixes that patch to work better.

If I'm understanding correctly, the change introduced in #4496 causes this to break because the `service` variable accessed in the `free_request_id` call is the version from `IPAddr.new`, which does not properly match the direct `host` version used by the call to `allocate_request_id`.

So I think the patch is is fine; the fix in #4496 was incomplete in that it only used the processed host along one path.

PR #5728, opened 2019-05-10
RFC: Do not add CLASSPATH to the module path

kschoelhorn:
960efcb13aa8fe6d44098ebe83ceeb7f90160445 and 7c4bf9c5d47aaa3f17d49636ee2dcb0dcded81b7 changed the starter so that the `$CLASSPATH` is added to the module path when running on java9+.

Treating the CLASSPATH as modules will create automatic modules for all entries that are not actually modules. This can lead to several problems, e.g. the automatic module name is determined by the name of the jar or directory, which fails when one component is not a valid java identifier. Another issue is that several automatic modules might export the same package, which is not allowed and leads to an error.

While the first problems is relatively easy to fix, the second one requires large changes and might not even possible if you are using third-party libraries (e.g. we are using UNO, which [has problems with this](https://bugs.documentfoundation.org/show_bug.cgi?id=117331)).

To fix these issues, we can use the module path just for jruby libraries and pass the CLASSPATH using `-classpath`, as both can be combined.

Note that I haven't done extensive testing, because I first wanted to hear what you think about this approach.

Unresolved questions:
 - [ ] Are there any unintended side effects?
 - [ ] Should $CP also go into the module path?
 - [ ] What about the other starters?

### DIFF
--- a/bin/jruby.bash
+++ b/bin/jruby.bash
@@ -374,10 +374,9 @@ if [ "$JRUBY_JSA" == "" ]; then
 fi
 
 # Determine whether to pass module-related flags
-classmod_flag="-classpath"
 if [ -d $JAVA_HOME/jmods ]; then
-  # Use module path instead of classpath
-  classmod_flag="--module-path"
+  # Use module path instead of classpath for the jruby libs
+  classpath_args=(--module-path "$JRUBY_CP" -classpath "$CP$CP_DELIMITER$CLASSPATH")
 
   # Switch to non-boot path since we can't use bootclasspath on 9+
   NO_BOOTCLASSPATH=1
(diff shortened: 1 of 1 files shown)


headius:
@kschoelhorn Good catch...I did not think about other libraries being on classpath since most Rubyists will require jars at runtime (which then load through URLClassLoader).

I will review. We will need an equivalent change for https://github.com/jruby/jruby-launcher.

iloveeclipse:
Regarding open questions: please note, classpath and modulepath are for different purpose and entries from one should not go to the another one. So as a user one should be able to specify both at same time without a fear that jruby does a happy mix from them.

headius:
Regarding your additional questions:

* We should not be depending on any CLASSPATH jars loading as modules, since that implies they should be loading with `-classpath`. In other words I think all effects of your change are intentional and appropriate.
* No, CP should be treated the same as CLASSPATH, as it is in your patch.
* Only thing is patching jruby-launcher. Shouldn't be hard...want to do a second PR for that?

klemens:
Thanks for merging this so fast! I will try to create a PR for jruby-launcher tomorrow.

(This is my private account)

PR #5739, opened 2019-05-20
Remove enable_feature calls in jruby.mspec

eregon:
* These features have been removed in https://github.com/ruby/spec/issues/667, so they don't do anything anymore.


headius:
Looks good, thanks.

PR #5744, opened 2019-05-22
first stab at a spec

ilikeorangutans:


kares:
Thanks Jakob, if you're just out for contributing ruby specs in general there's https://github.com/ruby/spec/
also fine to submit here but its more often the case when there's a JRuby bug/behavior that needs attention

kares:
we expected to have this passing in JRuby 9.2.9 (due https://github.com/jruby/jruby/pull/5834)

PR #5750, opened 2019-05-29
RbConfigLibrary: Set COUTFLAG to "-o "

### LINKED ISSUE #5749: mkmf: COUTFLAG does not have `-o`
jruby v9.2.7.0 produces Makefiles with invalid C compilation command.

Diffing a Makefile produced from the same [extconf.rb](https://github.com/sass/sassc-ruby/pull/127/files#diff-b3c266c0a40fd6ea5ce51093b147dbe6) with MRI vs jRuby

```diff
 OUTFLAG = -o $(empty)
-COUTFLAG = -o $(empty)
-CSRCFLAG = $(empty)
+COUTFLAG = $(empty)
```

Note that `COUTFLAG` in JRuby-produced Makefile is missing `-o`, resulting in an invalid `cc` command.

Work-around I'm using in the extconf.rb file:

```ruby
# COUTFLAG is not set correctly on jruby
if RUBY_PLATFORM == 'jruby'
  MakeMakefile.send(:remove_const, :COUTFLAG)
  MakeMakefile::COUTFLAG = '-o $(empty)'
end
```

glebm:
Fixes #5749

### DIFF
--- a/core/src/main/java/org/jruby/ext/rbconfig/RbConfigLibrary.java
+++ b/core/src/main/java/org/jruby/ext/rbconfig/RbConfigLibrary.java
@@ -471,6 +471,7 @@ public class RbConfigLibrary implements Library {
         setConfig(context, mkmfHash, "CPP", cpp);
         setConfig(context, mkmfHash, "CXX", cxx);
         setConfig(context, mkmfHash, "OUTFLAG", "-o ");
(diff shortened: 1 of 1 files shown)


kares:
looks good + it's been brought up twice already  - first at #4118 with more changes that are unclear to me

Issue #5754, opened 2019-05-30
EHOSTUNREACH when a FTP connection cannot be established


Adithya-copart:
### Expected Behavior

Is it possible to make the error convey that the remote host is unavailable or unreachable like MRI does with `EHOSTUNREACH`?

I think that `Errno::EBADF (Bad file descriptor - No message available)` do not convey that the FTP server/host is unreachable.

I looked at https://github.com/jruby/jruby/issues/5709#issuecomment-487128703 and wonder if it is even possible to raise a  `EHOSTUNREACH`.

Please feel free to close this if this cannot be considered an issue with JRuby.

**JRuby:**
```
jruby-9.2.7.0 :001 > require 'net/ftp'
 => true 
jruby-9.2.7.0 :002 > Net::FTP.open('example.com', 'test', 'password') do |ftp_conn|
jruby-9.2.7.0 :003 >     puts ftp_conn.nlst("*")
jruby-9.2.7.0 :004?>   end
Traceback (most recent call last):
       16: from /Users/adpentela/.rvm/rubies/jruby-9.2.7.0/lib/ruby/stdlib/net/ftp.rb:163:in `open'
       15: from /Users/adpentela/.rvm/rubies/jruby-9.2.7.0/lib/ruby/stdlib/net/ftp.rb:265:in `initialize'
       14: from /Users/adpentela/.rvm/rubies/jruby-9.2.7.0/lib/ruby/stdlib/net/ftp.rb:367:in `connect'
       13: from /Users/adpentela/.rvm/rubies/jruby-9.2.7.0/lib/ruby/stdlib/monitor.rb:226:in `mon_synchronize'
       12: from /Users/adpentela/.rvm/rubies/jruby-9.2.7.0/lib/ruby/stdlib/net/ftp.rb:369:in `block in connect'
       11: from /Users/adpentela/.rvm/rubies/jruby-9.2.7.0/lib/ruby/stdlib/net/ftp.rb:328:in `open_socket'
       10: from org/jruby/ext/timeout/Timeout.java:75:in `timeout'
        9: from org/jruby/ext/timeout/Timeout.java:82:in `timeout'
        8: from /Users/adpentela/.rvm/rubies/jruby-9.2.7.0/lib/ruby/stdlib/net/ftp.rb:333:in `block in open_socket'
        7: from /Users/adpentela/.rvm/rubies/jruby-9.2.7.0/lib/ruby/stdlib/socket.rb:638:in `tcp'
[...]

PR #5760, opened 2019-06-04
Socket error fixes

headius:
This fixes issues from #5709 and #5754.

* Unify logic for turning Java exceptions into Errno based on their messages. This allows the proper errors to be raised for ETIMEDOUT and EHOSTUNREACH within socket logic.
* Check for closed channel when doing the final flush logic in IO. This allows channels closed elsewhere or sockets not completely established to be closed again without error.
* Remove custom Socket#close since it should now work with the modified IO#close finalization logic.


headius:
I have reverted the general change to IO since it appears to break too many things and doesn't match MRI semantics. The unusual case here is that using JDK sockets, several connection failures leave the channel in a closed state, unusable for further connecting but not nulled out as is needed for close to silently move on. I am making a narrower change now that cleans up the socket after any of these fatal connection failures, so it looks to the Ruby close logic like it has already been closed cleanly elsewhere.

Issue #5590, opened 2019-02-06
LOADED_FEATURES internal index not working correctly


headius:
As part of work on eliminating all builtin libraries (so Shopify/bootsnap can cache require paths and lookups across runs) I ran into an issue with our logic to "provide" loaded features.

In MRI, rb_provide adds entries to the loaded features cache such that a simple require of the same name will be short-circuited.

We have similar logic, but it appears that the index where we store the short names either doesn't work properly or is getting wiped out.

This manifested in errors attempting to load the internal (provided) enumerator.rb, thread.rb, complex,rb, and rational.rb similar to the failures seen here: https://travis-ci.org/jruby/jruby/builds/489301097

Basically, because the loaded features cache didn't work right, it attempted to actually load those files. Since they don't exist, it produced an error.

I tracked one breakage down to the following lines in `test/jruby/test_file.rb`:

```ruby
      with_load_path("file:" + File.expand_path("test/jruby/dir with spaces/test_jar.jar") + "!/abc") do
        assert require('foo')
        assert $LOADED_FEATURES.pop =~ /foo\.rb$/
      end
```

Either the LOAD_PATH or the LOADED_FEATURES manipulation here prevents JRuby from seeing that enumerator.rb is already provided, and a later test fails when it tries to `require 'enumerator'`.

headius:
This also affects the non-existent (boot-loaded) `jruby/util`.

headius:
This has been fixed on the new load_service_redux branch by having a proper implementation of loaded features. The stubs were removed in 586b07c43341aac0601392f5cb4d10bb12942276.

Issue #5763, opened 2019-06-10
Refactor and optimize loaded features


headius:
For recent work on load/require and autoload, I ported over some logic from CRuby to manage the cache of loaded features. This cache builds up a mapping of all subpaths, considering load paths, canonicalized paths, and so on, in order to make future requires of any similar path detect the already-loaded feature. In CRuby, this cache has been carefully crafted to avoid allocating too many extra strings, sharing the backing store for the path among all entries. We currently do not do this, and I believe this may be slowing down load logic instead of speeding it up as the cache should do.

Before releasing JRuby 9.2.8, we will need to revisit this logic and ensure we're only doing the minimum allocation and work necessary to maintain and use the loaded features cache.

headius:
Addressed by #5764.

PR #5764, opened 2019-06-11
Rework load service, load path, loaded features, and autoload

headius:
This is a big one. This PR includes rewrites of how we handle load/require, load path lookups/caching/canonicalization, loaded feature caching, and autoload lifecycle relating to the rest.

Parts of this are loose ports from MRI, and still need some refinement and optimization. In particular, we need to make sure the load path and loaded features caches are tight, efficient, and not getting regenerated excessively. This also should be audited for concurrent state changes in autoload, since there's a few complicated interactions with threads and in-progress autoloads. We are passing most of the related specs (notably not the tags added in aea0e652866196e77b081fcfcf5f48c9c4e3752f) but the code is not proven safe in my eyes.

Hopefully others can help me audit this. I know it's a lot of code.

With this PR, here's status:

* MRI test_autoload: two excludes, for deprecation warning and a possibly bad test (threading)
* rubyspec kernel/module autoload_spec: 100%
* zeitwerk: 100%


headius:
Fixes #5590.

See #5763 for mentioned perf concerns.

headius:
Recent failure in an embedded/complete run of JRuby test suite...should be investigated for concurrency bug:

```
     [exec] ===============================================================================
     [exec] Error: test_require_fires_once_across_threads(TestRequire): RuntimeError: Concurrent requires caused double-loading
     [exec] /home/travis/build/jruby/jruby/test/jruby/test_require.rb:14:in `block in test_require_fires_once_across_threads'
     [exec]      11:       (1..(ENV['THREAD_COUNT'] || 5).to_i).map do
     [exec]      12:         Thread.new { require 'test/jruby/test_require_once_foo' }
     [exec]      13:       end.each { |t| t.join }
     [exec]   => 14:       raise "Concurrent requires caused double-loading" if $foo != 1
     [exec]      15:       $foo = 0
     [exec]      16:       $".pop
     [exec]      17:     end
     [exec] /home/travis/build/jruby/jruby/test/jruby/test_require.rb:10:in `test_require_fires_once_across_threads'
     [exec] org/jruby/RubyKernel.java:1150:in `catch'
     [exec] org/jruby/RubyKernel.java:1150:in `catch'
     [exec] ===============================================================================
```

https://travis-ci.org/jruby/jruby/jobs/544444301

headius:
Tweaks to get `Kernel#autoload` to find the right target (6d92298be04ec237ae6615f8dc9a6d448134e0f5) appear to have regressed a few things. This is related to the "JRUBY-6570" regression spec.

headius:
I know this is a big one...chat me up if you want to talk through some smaller pieces one at a time.

core/src/main/java/org/jruby/runtime/load/LibrarySearcher.java
```diff
@@ -359,20 +478,58 @@ public void clear() {
...
 
-        final CharSequence key;
-        final CopyOnWriteArrayList<IRubyObject> featurePaths;
+        final StringWrapper key;
+        final List<IRubyObject> featurePaths;
     }
 
-    protected void addSingleFeatureToIndex(String key, IRubyObject featurePath) {
-        Map<String, Feature> featuresIndex = this.loadedFeaturesIndex;
+    private final ThreadLocal<StringWrapper> keyWrapper = ThreadLocal.withInitial(() -> new StringWrapper(null, 0, 0));
```
kares: is this really worth the thread-local, do we allocate that many string instances on load service lookups?

alecgorge:
@headius is there anything I can do here to help get this over the finish line? Do you need more eyes on the code or is there functionality not finished yet?

headius:
@alecgorge The remaining work is to get autoloading with concurrent requires thread-safe. This logic was loosely ported from CRuby, which does not allow these operations to be performed in parallel, but the locking mechanisms we have available in JRuby are insufficient to make sure the case is safe. This is, I believe, the reason why the MRI suite has some sporadic failures; sometimes the parallel autoload tests work and sometimes they don't.

rickhull:
I see the Travis CI stuff is failing -- due to Travis CI issues.  FYI, Travis CI is basically dead for open source projects.

I just did a migration to GitHub Actions for a simple repo.  I used https://bibwild.wordpress.com/2020/11/12/deep-dive-moving-ruby-projects-from-travis-to-github-actions-for-ci/ as a guide, and it was fairly straightforward.  I found my way to this issue because my tests against jruby are showing the "circular requires" warning.

PR #5777, opened 2019-06-27
Update specs

eregon:


eregon:
Specs should be green now with the added tags, merging.

PR #5781, opened 2019-07-02
Support Time ranges in rand

davishmcclurg:
JRuby is currently converting Time objects to floats when generating a
random value from a range:

```
$ RBENV_VERSION=jruby-9.2.7.0 ruby -e 'puts rand(Time.now..Time.now)'
1562034267.423605
```

MRI returns a Time object:

```
$ RBENV_VERSION=2.6.3 ruby -e 'puts rand(Time.now..Time.now)'
2019-07-01 19:27:25 -0700
```

I believe the `RubyNumeric` check I added is the same as what MRI uses
in `rb_check_to_float`: https://github.com/ruby/ruby/blob/0b858425e1e4f2de40dc0d8e5dd105a2fd93e478/object.c#L3769

### DIFF
--- a/core/src/main/java/org/jruby/util/TypeConverter.java
+++ b/core/src/main/java/org/jruby/util/TypeConverter.java
@@ -297,6 +297,7 @@ public class TypeConverter {
     // rb_check_to_float
     public static IRubyObject checkFloatType(Ruby runtime, IRubyObject obj) {
         if (obj instanceof RubyFloat) return obj;
+        if (!(obj instanceof RubyNumeric)) return runtime.getNil();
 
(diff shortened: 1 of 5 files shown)


kares:
specs look good but if you look at the build the proposed fix introduced regressions: https://api.travis-ci.org/v3/job/553080882/log.txt ... the 'proper' fix likely won't involve a change a `TypeConverter`

davishmcclurg:
@kares I think the fix is correct (`TypeConverter.checkFloatType` is only used in `randomRand` as far as I can tell), but the `CustomRangeFloat` check I added wasn't right. I pushed up another commit to fix it. Let me know if I should squash them.

PR #5809, opened 2019-08-01
MRI compatibility with IO::sysopen for serial ports on windows

byteit101:
These changes make JRuby be compatible with MRI's API's and behaviors on Windows serial ports.

Fileno is now real in most cases, but lazily generated in most cases to avoid sucking up 2x the number of file resources.

I put the WinC interface in FilenoUtils arbitrarily, I wasn't sure the best place to put it

### DIFF
--- a/core/src/main/java/org/jruby/util/io/FilenoUtil.java
+++ b/core/src/main/java/org/jruby/util/io/FilenoUtil.java
@@ -2,10 +2,16 @@ package org.jruby.util.io;
 
 import com.headius.backport9.modules.Modules;
 import jnr.enxio.channels.NativeSelectableChannel;
+import jnr.ffi.LibraryLoader;
+import jnr.ffi.Pointer;
 import jnr.posix.FileStat;
+import jnr.posix.HANDLE;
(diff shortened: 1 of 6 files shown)


core/src/main/java/org/jruby/util/io/FilenoUtil.java
```diff
@@ -126,6 +132,43 @@ public static int filenoFrom(FileDescriptor fd) {
...
+            return winc._close(fd);// TODO: error handling
+        return -1;
+    }
+
+    public static interface WinC {
+        int _open_osfhandle(Pointer hndl, int flgs);
+        int _close(int fd);
+    }
+
+    private static final WinC winc = Platform.IS_WINDOWS ? LibraryLoader.create(WinC.class).load("msvcrt") : null;// TODO: string
```
headius: I think this might blow up if native support is diabled, yes? I think we need to add a check for that here and anywhere that tries to use this library.

byteit101: Pulled into ctor and accessed in a non-static manner now

PR #5814, opened 2019-08-05
fix custom call-site performance regression

kares:
restores interpreter 'fast' Fixnum/Float operations performance, which degraded after doing the `isBuiltin` checks (#4736)

there's a bit of refactoring needed for the CachingCallSite API used
we're adding a `CacheEntry setCache(CacheEntry entry, IRubyObject self)`
and checks such as `isBuiltin(RubyClass)` change to -> `isBuitlin(IRubyObject self)`
... this is the way JRuby intends to use these APIs anyway, so no harm done there

the gist of better performance is adding logic in custom sites when `setCache` happens,
do the builtin and target check there so that on the hot-path, taking the direct specific (Fixnum/Float) branch becomes 2 `instanceof` checks (edit: plus the obvious type "token" check)

### DIFF
--- a/core/src/main/java/org/jruby/ext/ffi/AutoPointer.java
+++ b/core/src/main/java/org/jruby/ext/ffi/AutoPointer.java
@@ -94,7 +94,7 @@ public class AutoPointer extends Pointer {
         ClassData classData = (ClassData) ffiHandle;
 
         // If no release method is defined, then memory leaks will result.
-        DynamicMethod releaseMethod = classData.releaseCallSite.retrieveCache(getMetaClass().getMetaClass(), classData.releaseCallSite.getMethodName()).method;
+        DynamicMethod releaseMethod = classData.releaseCallSite.retrieveCache((IRubyObject) getMetaClass()).method;
         if (releaseMethod.isUndefined()) {
             throw runtime.newRuntimeError("release method undefined");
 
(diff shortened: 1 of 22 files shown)


kares:
and here's some raw numbers : 

`jruby -e 'loop { t = Time.now; i = 0; while i < 100_000_000; i+=1; end; puts Time.now - t }'`

**JRuby 9.2.7.0**
```
0.791033
0.6960909999999999
0.5727019999999999
0.554544
0.54064
0.5588080000000001
0.5369900000000001
0.539406
0.542498
0.542969
```

**this branch**
```
0.879055
0.8296479999999999
0.85761
0.5957279999999999
0.583428
0.5834809999999999
0.5803499999999999
0.584803
0.587146
0.580974
```

things are much more interesting, when Fixnum gets re-opened (fairly common in gems and the whole point of doing #4736 in the first place) : 

`jruby -e 'class Fixnum; def some; end end; loop { t = Time.now; i = 0; while i < 100_000_000; i+=1; end; puts Time.now - t }'`

**JRuby 9.2.7.0**
```
2.400054
2.118736
1.931208
2.02758
2.0146930000000003
2.0489669999999998
1.943752
1.94749
1.93973
1.9508480000000001
```
**this branch** (re-opening has no effect)
```
0.889061
0.794413
0.6611440000000001
0.823025
0.5830369999999999
0.589919
0.589366
0.585882
0.583457
0.5890059999999999
```

Issue #4968, opened 2018-01-12
proc binding(local variables) is not garbage collected (memory leak)


yamam:
Local variables which is binded by a proc is not garbage collected, if the proc is passed to java function.

### Environment
jruby 9.1.0.0 (2.3.0) 2016-05-02 a633c63 Java HotSpot(TM) 64-Bit Server VM 25.151-b12 on 1.8.0_151-b12 +jit [mswin32-x86_64]

### Expected Behavior

JavaClass.java
```java
public class JavaClass
{
    public JavaClass(Runnable proc)
    {
        proc.run();
    }
}
```

memoryleak.rb
```ruby
import 'JavaClass'

class BigClass
    # 100MB class
    def initialize
        @a = "a" * (100 * 1024 ** 2)
    end
end

def func
    a = BigClass.new
    JavaClass.new do
        puts "callback"
    end
end

func

# GC by Java VisualVM

gets
```

1.
$ javac JavaClass.java
$ jruby-9.0.5.0/bin/jruby memoryleak.rb

2.
run GC by jvisualvm

after running GC, memory usage is reduced about 100MB.

### Actual Behavior
1.
$ javac JavaClass.java
$ jruby-9.1.0.0/bin/jruby memoryleak.rb  (9.1.0.0 or above)

2.
run GC by jvisualvm

after running GC, memory usage is not reduced.

kares:
just try the latest off JRuby 9.1 and you should be good to do, recall this has been fixed along the way

yamam:
I tried jruby 9.1.15 and 9.1.16-SNAPSHOT. The result is the same as 9.1.0.0.
I also tried jruby 9.2.0.0-SNAPSHOT, but another error occured, and the script couldn't run
```
jruby-9.2.0.0-SNAPSHOT/bin/jruby memoryleak.rb
unknown encoding name - MS932
```

headius:
This is likely because the result of func gets stored into a temporary local variable but never used or cleared. This causes the JavaClass object, the block and binding it captures, and the BigClass stored in that binding to be retained.

The IR for the top-level method, from the func call down:

```
  10:  %v_7 := call_0o(self<%self>, callType: VARIABLE, name: func, potentiallyRefined: false)
  11:          line_num(lineNumber: 20)
  12:  %v_8 := call_0o(self<%self>, callType: VARIABLE, name: gets, potentiallyRefined: false)
  13:          pop_binding
  14:          pop_method_frame
  15:          return(%v_8)
[...]

PR #5820, opened 2019-08-08
Dereference proc from interface impl singleton

headius:
By holding a reference to the proc, we anchor the proc's binding to the singleton class. When methods from the singleton class are cached elsewhere, such as for Java interface proxies (see GH-4968) we end up keeping the proc's binding alive longer than we'd like.

This patch detaches the proc object from the singleton, instead pointing it at the Proc class. The side effects from this include(at least) that any new singleton methods defined on that proc's singleton class will dispatch to hook methods (like singleton_method_defined) on Proc, rather than on the proc object. It's unknown whether this would affect any existing code.


headius:
Approved by @enebo and @kares.

PR #5825, opened 2019-08-09
Allow open, nonpublic classes to bind as if accessible

headius:
The existing logic will fail to define any Ruby proxy class + method stubs when a given class is nonpublic and a generic setAccessible (unrelated to the class) would fail. This makes it impossible to allow JRuby to bind the class when it has been explicitly opened up by e.g. `add-opens` flags.

This patch uses the module's open status to determine whether the package has been opened up to JRuby. If it has, we proceed to bind the class as if it were possible to set it accessible.


headius:
I have not tested this yet. Also seeking input from @kares @enebo.

core/src/main/java/org/jruby/javasupport/binding/Initializer.java
```diff
@@ -489,7 +489,9 @@ private static void setJavaClassFor(final Class<?> javaClass, final RubyModule p
         for ( Class<?> klass = javaClass; klass != null; klass = klass.getSuperclass() ) {
             // only add class's methods if it's public or we can set accessible
             // (see JRUBY-4799)
-            if (Modifier.isPublic(klass.getModifiers()) || JavaUtil.CAN_SET_ACCESSIBLE) {
+            if (Modifier.isPublic(klass.getModifiers())
+                    || JavaUtil.CAN_SET_ACCESSIBLE
+                    || Modules.getModule(klass).isOpen(klass.getPackage().getName(), Modules.getModule(Ruby.class))) {
```
kares: will NPE on Java 8 -> `getPackage()` returns null for default package while 9 returns a package object
recall, we have logic dealing with this in `com.headius.backport9.modules` itself, so maybe having a 
`Modules.isOpen(Class target, Class from)` there handling the details would fit well ...

headius: Good call, I'll force push that change.

headius:
Ok...I made the change @kares suggested, plus an additional fix that sets methods accessible if their host class is nonpublic. Now I'm not sure we should make this change.

Generally methods that are on non-public class, whether the methods themselves are public or not, can't be seen by Java callers. Also typically, when we have a *child* class that's non-public, it is intended to be a hidden implementer or overrider of the parent methods, and should not be bound directly.

This PR may be a solution in search of a problem. I initiated it mostly because I noticed that we check `trySetAccessible` for methods but not for classes, and that seemed incongruent. Now, I'm not sure.

headius:
I'm going to drop this for now; invisible classes will open a can of worms I don't think we want to mess with, and if people really need to access them they can certainly use reflection directly.

Issue #1806, opened 2014-07-11
Fiber#new and current Thread


robin850:
Hello,

There is a tiny problem with JRuby when dealing with a `Fiber` and a `Thread` object. For instance, with this code:

``` ruby
thread = Thread.current

Fiber.new {
  thread == Thread.current
}.resume
```

The fiber will yield `false` because JRuby spawns a new Thread calling [`Fiber#new`](https://github.com/jruby/jruby/blob/4fc75aed44c3b882b9f4791b39b51db4106fe30c/core/src/main/java/org/jruby/truffle/runtime/core/RubyFiber.java#L97). The fiber will yield `true` on MRI though.

Looking at the logs, the commit introducing this thread creation (i.e. [0a8680cf](https://github.com/jruby/jruby/commit/0a8680cf51e31874f8efb15395d592fe5385ef57)) is the initial commit of the `RubyFiber.java` file so we have no information about this.

This has been tested with JRuby 1.7.13 and current master.

Have a nice day.

zanker:
I can confirm this still exists on JRuby 1.7.19/9.0.0.0.pre1. It can lead to some fun issues when you try to reuse threads using `Thread.current`, like  [statsd-ruby](https://github.com/reinh/statsd/blob/master/lib/statsd.rb#L393) or [net-http-persistent](https://github.com/drbrain/net-http-persistent/blob/master/lib/net/http/persistent.rb#L635) and something that uses Fibers like Celluloid/Sidekiq.

chuckremes:
This issue is also still causing headaches. See [issue 2 in `ruby-io` for details](https://github.com/chuckremes/ruby-io/issues/2).

headius:
This has come up many times, and it's not an easy thing to fix. The main issue is that our Fiber subsystem uses normal Ruby synchronization mechanisms (specifically, Queue) which depends on the Fiber and the Thread it's synchronizing against to have their own notions of current thread.

However, since I wrote this implementation, we have improved our internal logic for synchronizing threads with non-Ruby mechanisms without interfering in Ruby signalling. This can be revisited, but most likely not until 9.2.

zanker:
Good to hear anyway, thanks for an update!

headius:
FWIW there is now a working coroutine implementation for OpenJDK: [Project Loom](http://openjdk.java.net/projects/loom/). Given a bit of time (or help from the community) we could ship pre-release support for it in JRuby any day now. Maybe for 9.2.2?

[...]

PR #5827, opened 2019-08-10
Use parent thread for Thread.current within a fiber

headius:
This is the start of experiments in making Thread.current, among other things, pretend that a Fiber is equivalent to its master thread, rather than being its own thread.

This work was triggered by me running the socketry/async specs and seeing that there's some locking done across fibers that's expected to treat them all as owning the lock. In this case, it's
an RSpec reentrant lock implementation that aggregates a Mutex and the thread that acquired it and compares that thread with Thread.current to avoid relocking.

For JRuby the following changes were needed to make the lock work:

* Thread.current must point at the "Fiber current thread" rather than the "Fiber thread". We have attempted this in the past and run afoul of system-level locking that saw through to the real
locks.
* Mutex#owned? had to do a similar check, aggregating a reference   to the Fiber current thread that originally locked it and using that for comparison with the current Fiber current thread.

With these changes the async specs run to completion with 16 errors that I have not looked into yet.

See #1806.

### DIFF
--- a/core/src/main/java/org/jruby/RubyThread.java
+++ b/core/src/main/java/org/jruby/RubyThread.java
@@ -925,7 +925,7 @@ public class RubyThread extends RubyObject implements ExecutionContext {
 
     @JRubyMethod(meta = true)
     public static RubyThread current(IRubyObject recv) {
-        return recv.getRuntime().getCurrentContext().getThread();
+        return recv.getRuntime().getCurrentContext().getFiberCurrentThread();
     }
 
     @JRubyMethod(meta = true)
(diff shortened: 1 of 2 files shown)


headius:
This didn't make the cut for 9.3, which we wanted to get out the door to work on Ruby 3.0 support in 9.4. Punting to 9.4 to reevaluate.

headius:
Rebased on master. With recent changes to fibers, we'll see how this goes in CI. It's not something I'd like to drop in an update release, so if we are green with this change now might be the time to push it through.

headius:
There appears to be several hanging tests, so something is not right here.

headius:
Going to reboot this on a separate branch and PR.

PR #5826, opened 2019-08-09
Add support for .jruby.java_opts home and cwd files

headius:
This implements the dotfile idea from #5824 in our bash launcher. The order of precedence for options, indicated by their position in the java command line, goes as follows:

* home dir .jruby.java_opts
* current dir .jruby.java_opts
* JAVA_OPTS environment variable
* command line flags

This allows users to set up global, local, and environment-based java options but still override those at the command line.

An equivalent change will need to go into jruby-launcher.

PR #5831, opened 2019-08-13
Further improvements to dev mode

headius:
This is a work-in-progress to find additional settings and tweaks we can associate with the `--dev` flag to speed things up at boot for developer environments.


headius:
Will merge now and prepare to start utilizing the `@file` syntax as in #5826.

PR #5834, opened 2019-08-14
enumerator review for better compatibility

kares:
this work builts upon https://github.com/jruby/jruby/pull/5672, which introduced the enumerator.rb library used by Rubinius ... 
here, we're now adjusting parts for better compatibility with MRI's implementation.

just a note that the `enumerator.rb` introduced wasn't a drop in replacement and (already) required modifications on JRuby's end (Fiber state sharing) to make it usable. the gist of the lib stays the same.

changes are motivated by getting compatibility improved - **50 passing specs, 11 untagged MRI tests**

also, due some `Enumerator` code being triggered during boot by RGs - we rather avoid reflecting on JRuby's internal classes (hasn't triggered the parts with JI loading but still)


headius:
Wow, lots of commits. I'll have a look.

Issue #3333, opened 2015-09-16
handle non-public inner class retrieval


kares:
users still might use work-around `JavaUtilities.get_proxy_class('path.to.Outer$Inner')` e.g. #933 

they can get an improvement to act similar to how `private_constant` does - no showing up in `Outer.constants` but with `Outer::Inner` working (inside `Outer`)

PR #5835, opened 2019-08-15
Fail gracefully when inner class const_missing finds nothing

headius:
This should fix the issue reported as a comment in https://github.com/ruboto/JRuby9K_POC/issues/7.

The issue here appears to be that non-existent inner classes are not handled properly; the resulting null or nil values fail to cast to expected types and the superclass const_missing fallback never triggers.

The change I have here uses a different path for acquiring a reference to the proxy class, which then allows the fallback to fire properly.

The original change was made by @kares in #3333 but it appears the test cases added did not test the case where there is no inner class for the constant.


headius:
Test added in a118fe8ce61b2d7292217a3d92e31b160f939919.

PR #5837, opened 2019-08-17
Tweaks to reduce memory use of short commands

headius:
This is a set of memory/alloc-reducing experiments that can be merged or cherry-picked as appropriate.

### DIFF
--- a/core/src/main/java/org/jruby/ast/ListNode.java
+++ b/core/src/main/java/org/jruby/ast/ListNode.java
@@ -255,7 +255,6 @@ public class ListNode extends Node implements Iterable<Node> {
 
     @Override
     public Iterator<Node> iterator() {
-        Thread.dumpStack();
         return new Iterator<Node>() {
             int i = 0;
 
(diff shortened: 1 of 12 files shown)


headius:
The failures appear to be due to 3dc4a9c, the first commit in this PR. For some reason the line number is off by one. @enebo Can you see why that would be?

headius:
I have re-pushed this with the failing source position patch moved to #5838.

core/src/main/java/org/jruby/ast/ListNode.java
```diff
@@ -174,11 +246,16 @@ public boolean isEmpty() {
...
+        Node single = this.single;
+
+        if (idx == 0 && single != null) return single;
+
         return list[idx];
     }
 
     @Override
     public Iterator<Node> iterator() {
+        Thread.dumpStack();
```
kares: :scissors:

headius: Fixed! We might want to consider removing this since it doesn't appear to be used.

headius:
I'm going to go ahead and merge this. @enebo any review issues you find I will patch later.

PR #5838, opened 2019-08-17
Reuse existing source position if it matches

headius:
This seems simple enough but causes some off-by-one issues for reasons I have not managed to figure out.

If this change can work, it reduces SimpleSourcePosition allocation during `-e 1` from 46240 (1110264 bytes) to 13984 instances (335616 bytes).


headius:
@enebo Ahh yes, I remember the change you made for 2.6. This patch was a substantial reduction but eliminating them altogether is obviously better. We can wait for your change or backport it but I'll close this one.

PR #5839, opened 2019-08-18
Move more tests to Azure from Travis

headius:
This PR will move as many jobs as possible from Travis to Azure, to increase our coverage on Windows.


headius:
We'll approach this more directly by testing stuff on Windows first.

PR #5840, opened 2019-08-18
Consistently resolve path of jruby.bash

Mrnoname1000:
The bash script sometimes returns an incorrect path for `JRUBY_HOME`. I came up with a bunch of cases where it would return the wrong directory, in no particular order:

1. a directory in PATH is a symlink
2. a directory in PATH is in a symlink
3. the executable is a symlink in PATH
4. the executable is in a directory that is a symlink
5. the executable is a symlink that points to an executable in a directory that is a symlink
6. the executable is being read from a pipe

Simply replacing `$0` with `$BASH_SOURCE` fixes case 3. Obviously this is bash-specific, so this can't be ported to the sh script.
Case 4 are solved by using readlink on the executable until it isn't a symlink. The script already did this, so I made some changes to solve other issues.
Cases 1, 2, and 5 are solved by running `pwd -P` in the directory of the executable.
I don't know how to solve case 6 without hardcoding a directory to check.

Because `pwd -P` always returns the absolute working directory, it seems running `dirname "$SELF_PATH"` is unnecessary, so I removed it.
If anyone can think of any other torture tests, let me know.

### DIFF
--- a/bin/jruby.bash
+++ b/bin/jruby.bash
@@ -49,7 +49,7 @@ while [ -h "$SELF_PATH" ]; do
     SELF_PATH="$(cd "$BASE_DIR" && cd $(dirname -- "$SYM") && pwd -P)/$(basename -- "$SYM")"
 done
 
-JRUBY_HOME="$SELF_PATH"
+JRUBY_HOME="${SELF_PATH%/*/*}"
 
 if [ -z "$JRUBY_OPTS" ] ; then
   JRUBY_OPTS=""
--- a/bin/jruby.bash
+++ b/bin/jruby.bash
@@ -36,28 +36,20 @@ if [ -z "$JAVA_VM" ]; then
 fi
 
 # get the absolute path of the executable
-SELF_PATH=$(builtin cd -P -- "$(dirname -- "$0")" >/dev/null && pwd -P) && SELF_PATH=$SELF_PATH/$(basename -- "$0")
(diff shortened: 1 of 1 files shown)


headius:
Looks good, though I've never seen that syntax you used for JRUBY_HOME. What is that?

Thanks for the patch!

bin/jruby.bash
```diff
@@ -36,28 +36,20 @@ if [ -z "$JAVA_VM" ]; then
...
-PRG=$SELF_PATH
-
-JRUBY_HOME_1=`dirname "$PRG"`           # the ./bin dir
-if [ "$JRUBY_HOME_1" = '.' ] ; then
-  cwd=`pwd`
-  JRUBY_HOME=`dirname $cwd` # JRUBY-2699
-else
-  JRUBY_HOME=`dirname "$JRUBY_HOME_1"`  # the . dir
-fi
+JRUBY_HOME="${SELF_PATH%/*/*}"
```
headius: What does this syntax do?

Mrnoname1000: That's just a quick and dirty way of removing /bin/jruby.bash. `${parameter%pattern}` returns the shorted possible expansion of `parameter` without the suffix `pattern`. Since `SELF_PATH` will always return the absolute path of jruby.bash, we can safely remove those last two slashes.
Also, it's POSIX compatible, so this and most of the rest of this could probably be ported to the shell script.

Issue #5821, opened 2019-08-09
9.2.8 -- java.util.logging.ConsoleHandler.setOutputStream works in java 8 but not 11 (impacts trinidad)


cshupp1:
On Windows 10 using  jruby-complete-9.2.8.0-20190808.165920-96.jar,

consider the following irb session:
```
C:\work\uts-web\uts-web-editor>c:\languages\Java\jdk1.8.0_121\bin\java -jar %JRUBY_JAR% -S jirb
irb(main):001:0> JRUBY_VERSION
=> "9.2.8.0-SNAPSHOT"
irb(main):002:0>     JUL = Java::JavaUtilLogging
=> Java::JavaUtilLogging
irb(main):003:0> handler = JUL::ConsoleHandler.new
=> #<Java::JavaUtilLogging::ConsoleHandler:0x64942607>
irb(main):005:0> handler.setOutputStream(Java::JavaLang::System.out)
=> nil
```

The same in 11:
```
irb(main):006:0> JRUBY_VERSION
=> "9.2.8.0-SNAPSHOT"
irb(main):007:0>  JUL = Java::JavaUtilLogging
=> Java::JavaUtilLogging
irb(main):008:0>  handler = JUL::ConsoleHandler.new
=> #<Java::JavaUtilLogging::ConsoleHandler:0x315cf170>
irb(main):009:0>  handler.setOutputStream(Java::JavaLang::System.out)
?[1mTraceback?[m (most recent call last):
        6: from uri:classloader:/META-INF/jruby.home/bin/jirb:13:in `<main>'
        5: from org/jruby/RubyKernel.java:1193:in `catch'
        4: from org/jruby/RubyKernel.java:1193:in `catch'
        3: from org/jruby/RubyKernel.java:1425:in `loop'
        2: from org/jruby/RubyKernel.java:1061:in `eval'
        1: from (irb):9:in `evaluate'
?[1mNoMethodError (?[4mundefined method `setOutputStream' for #<Java::JavaUtilLogging::ConsoleHandler:0x315cf170>?[0;1m)?[m
irb(main):010:0> java.lang.System.getProperty('java.version')
=> "11.0.4"
```

This affects Trinidad.

Thanks,

Cris

cshupp1:
Chicanery like this seems to work:
```
JUL = Java::JavaUtilLogging

class Poo < JUL::ConsoleHandler

  def setOutputStream(stream)
    super(stream)
  end

end

[...]

PR #5843, opened 2019-08-21
Move all JI setAccessible to trySetAccessible.

### LINKED ISSUE #5841: Can't call methods of private subclasses with java_method in java 12
### Environment

jruby 9.2.8.0 (2.5.3) 2019-08-12 a1ac7ff OpenJDK 64-Bit Server VM 12.0.2+10 on 12.0.2+10 +jit [mswin32-x86_64]

JavaTest.java
```java
public class JavaTest {

    public interface PublicInterface {
        public void publicMethod(Integer a);
        public void publicMethod(Long a);
    }

    public PublicInterface publicMember = new PrivateSubclass();

    private class PrivateSubclass implements PublicInterface {
        public void publicMethod(Integer a) {System.out.println("publicMethod(Integer)");}
        public void publicMethod(Long a) {System.out.println("publicMethod(Long)");}
    }
}
```

test.rb
```ruby
java_import 'JavaTest'
JavaTest.new.publicMember.java_method(:publicMethod, [java.lang.Integer.java_class]).call(1)
```

### Expected Behavior
You can call methods of private subclasses with java_method in java 8.

```
$ javac JavaTest.java
$ jruby test.rb
publicMethod(Integer)
```

### Actual Behavior
You can't call methods of private subclasses with java_method in java 12.
```
$ jruby test.rb
2019-08-20T09:36:29.010+09:00 [main] WARN FilenoUtil : Native subprocess control requires open access to sun.nio.ch
Pass '--add-opens java.base/sun.nio.ch=org.jruby.dist' or '=org.jruby.core' to enable.
TypeError: illegal access on 'publicMethod': class org.jruby.javasupport.JavaMethod (in module org.jruby.dist) cannot access a member of class JavaTest$PrivateSubclass with modifiers "public"
    call at org/jruby/RubyMethod.java:120
  <main> at test.rb:2
```

headius:
The logic previously was depending on the ji.setAccessible
property, which by default would simply refuse to set accessible
*at all* when the Java version was >= "1.9" (also showing this
logic was introduced very early in the Java 9 development cycle.)

That default is removed here, along with a the CAN_SET_ACCESSIBLE
static field in RubyInstanceConfig deprecated and replaced with
the original meaning of this property, simply whether to attempt
setAccessible.

The direct calls to setAccessible have been replaced by the
trySetAccessible in backport9 library that checks if the target
member's module is open to JRuby, allowing users to opt-in to that
reflective access if their application needs it. Previously these
cases still skipped setAccessible just due to being on Java 9+.

Most of the changes here were masked by the property, but I have
also fixed several others that were not guarded by the property
or that were otherwise used in booting JRuby or extending Java
integration.

This fixes #5841 by allowing `java_method` Method objects to
go ahead and trySetAccessible when the module is open.

This relates to #5821, which fixed in 9.2.8 a similar issue where
even an open module would not set fields and methods accessible.

### DIFF
--- a/core/src/main/java/org/jruby/RubyInstanceConfig.java
+++ b/core/src/main/java/org/jruby/RubyInstanceConfig.java
@@ -1837,7 +1837,7 @@ public class RubyInstanceConfig {
 
     public static final boolean JIT_LOADING_DEBUG = Options.JIT_DEBUG.load();
 
-    public static final boolean CAN_SET_ACCESSIBLE = Options.JI_SETACCESSIBLE.load();
+    public static final boolean SET_ACCESSIBLE = Options.JI_SETACCESSIBLE.load();
 
     // properties for logging exceptions, backtraces, and caller invocations
     public static final boolean LOG_EXCEPTIONS = Options.LOG_EXCEPTIONS.load();
@@ -2044,4 +2044,6 @@ public class RubyInstanceConfig {
      */
     @Deprecated
     public static final boolean NATIVE_NET_PROTOCOL = Options.NATIVE_NET_PROTOCOL.load();
+    @Deprecated
+    public static final boolean CAN_SET_ACCESSIBLE = Options.JI_SETACCESSIBLE.load();
 }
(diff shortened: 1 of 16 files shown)


headius:
> minor suggestion: have a JRuby `trySetAccessible` helper
> ... so we do not have to keep in mind to pass `Ruby.class`

Yeah that might be a good idea. I hate to add something more to Ruby.java, but maybe to our Java class? It's already pretty fat and not really public API. `Java.trySetAccessible` would just do `Modules.trySetAccessible(..., Java.class)` and have the same effect.

Issue #1050, opened 2013-09-30
Timeout::timeout not throwing exception when timeout period reaches


navaneeth:
_/tmp/loop.sh_

``` shell
#!/bin/sh
while :
do
    sleep 1
done
```

Now start the above script from a Ruby script.

_/tmp/test.rb_

``` ruby
require 'timeout'
begin
    Timeout::timeout(1) {
        `/tmp/loop.sh`.chomp
    }
rescue Exception => e
    puts 'inside rescue'
rescue Error => e1
    puts "inside error"
end
```

Execute the above script in a JRuby `ScriptingContainer`.

``` java
ScriptingContainer container = new ScriptingContainer();
container.runScriptlet(PathType.ABSOLUTE, "/tmp/test.rb");
```

The rescue block is not getting executed. Tested this with MRI and it works well there.

rtyler:
FWIW, I've reproduced that this is still the case with JRuby 9.0.0.0 whereas MRI 2.1.5 puts "inside rescue"

Adithya-copart:
I faced the same issue when I tried to do something like this on :

```
require 'timeout'
begin
  Timeout.timeout(5) do
    pid = spawn('ruby -e "sleep"')
    Process.wait pid
  end
rescue Timeout::Error => e
  'timed out'
end
```

headius:
I suspect MRI is using a non-blocking form of wait here, that allows them to give up on the subprocess. I will look into whether we can do the same.

headius:
Fixed by #5848.

PR #5848, opened 2019-08-27
Wrap Process.wait* calls with pthread_kill logic.

headius:
This addresses #1050 in the same way CRuby implements this. In
CRuby, the wait functions are called directly, and the
`pthread_kill` function is used to force the call to be
interrupted. It turns out the JVM also uses `pthread_kill` to
interrupt blocking native calls, so we use the same binding they
do.

Note that this will only be used on non-Windows, but the JDK's
binding of `NativeThread.signal` might already be stubbed out
there.


headius:
@enebo Any concerns here?

PR #5856, opened 2019-09-02
jffi archive mips64el

jackwang713:
[jffi.archive.tar.gz](https://github.com/jruby/jruby/files/3565089/jffi.archive.tar.gz)


headius:
I am going to move this archived binary over to a jffi PR.

headius:
Not sure why this is getting constant updates. Closing because it's not valid anymore anyway. We have updated master jffi to 1.2.20 with support for mips64el-Linux

PR #5860, opened 2019-09-08
automatic-module-name

### LINKED ISSUE #5855: Issues using bundle plugin with Java modules
I've been trying to figure out how to get the jruby-core and jruby-complete jar files to include Java module information, but so far I'm stuck dealing with the Apache Felix bundle plugin's weak support for filtering files.

@mkristian I need help here!

I want to get at LEAST Automatic-Module-Name in the manifests for all forms of jar files we export, a la jruby-core.jar should be "org.jruby.core", jruby-complete.jar should be "org.jruby.complete" and the jruby.jar in the dist should be "org.jruby.dist". Only this last one works currently.

The core jar refuses to take the Automatic-Module-Name manifest entry no matter what I do. I've tried modifying various places in core/pom.rb and in maven/jruby/pom.rb and it never picks it up.

The situation with jruby-complete is even worse. Because it relies on the maven-bundle-plugin (OSGi bundler) it pulls in the module-info.class files from joni, jcodings, and invokebinder. This last one "wins" because it's later in dependencies and there's no way to omit this file using the bundle plugin.

I did manage to get Automatic-Module-Name to be in the manifest by adding it to the "archive" subsection of the bundle plugin config, but because module-info.class is present this is ignored. The jar ends up looking like it's just an invokebinder module.

Hints online are sparse. The best recommendation I found was to ditch the bundle plugin and go back to just a shade plugin, which allows fine-grained filtering and renaming we want.

I'm not sure how to proceed here, but it's really a mess that our published jruby-complete jar advertises itself as the invokebinder module.

mkristian:
fix #5855

### DIFF
--- a/core/pom.rb
+++ b/core/pom.rb
@@ -307,6 +307,9 @@ project 'JRuby Core' do
                            {pattern: 'org.objectweb', shadedPattern: 'org.jruby.org.objectweb' },
                            {pattern: 'me.qmx.jitescript', shadedPattern: 'org.jruby.me.qmx.jitescript'},
                        ],
+                       transformers: [ {'@implementation' => 'org.apache.maven.plugins.shade.resource.ManifestResourceTransformer',
+                                         'mainClass' => 'org.jruby.Main',
+                                         'manifestEntries' => {'Automatic-Module-Name' => 'org.jruby.core'}}],
                        filters:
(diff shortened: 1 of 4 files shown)


headius:
Awesome!

PR #5803, opened 2019-07-27
Switch readline to reline and irb to irb gem.

headius:
This addresses #5802 and replaces the built-in stdlib IRB with the new improved IRB by @aycabta.

Notes:

* This is based on current releases of these two gems, reline 0.0.1 and irb 1.1.0.pre.2.
* There's one patch missing for irb: ruby/irb#20. I have patched around this by aliasing `RubyVM::InstructionSequence.compile` to `JRuby.compile_ir` in our own core/src/main/ruby/jruby/jruby.rb. This change should be removed once irb has been updated.
* There's one patch missing for reline on Windows: ruby/reline#44. I have not patched around that in this PR.
* reline uses Fiddle on Windows, and I have not tested that our Fiddle is sufficient.

I believe the plan is for MRI to move to reline and the new irb in Ruby 2.7, which will correspond to JRuby 9.3.

PR #5861, opened 2019-09-08
keep executable bin/exe executable

mkristian:
ad https://github.com/jruby/jruby/pull/5803#issuecomment-515658854

@headius not sure about the bin scripts. for example if I install `gem install xmlrpc` then there are no scripts in $GEM_HOME/bin but with `gem install bundle` I do get them copied to bin.

the only difference I see that we miss the executable bits on for example lib/ruby/gems/shared/gems/xmlrpc-0.3.0/bin/*
which I fix with this PR

do I miss something ?

### DIFF
--- a/lib/pom.rb
+++ b/lib/pom.rb
@@ -233,7 +233,10 @@ project 'JRuby Lib Setup' do
 
     # fix file permissions of installed gems
     ( Dir[ File.join( jruby_gems, '**/*' ) ] + Dir[ File.join( jruby_gems, '**/.*' ) ] ).each do |f|
-      File.chmod( 0644, f ) rescue nil if File.file?( f )
+      if File.file?( f )
+        mod = f.include?('/bin/') || f.include?('/exe/') ? 0755 : 0644
+        File.chmod( mod, f ) rescue nil
+      end
     end
   end
 


headius:
@mkristian I'll have a look!

headius:
@mkristian Ah I think I see the confusion. What I meant was that when we preinstall a gem using our maven build, the bin scripts are not installed. For example, if I remove `bin/rake`, it does not get replaced when the build fetches and installs rake.

```
[] ~/projects/jruby $ gem uninstall rake
Remove executables:
	rake

in addition to the gem? [Yn]  y
Removing rake
Successfully uninstalled rake-12.3.2

[] ~/projects/jruby $ git st
On branch master
Your branch is up to date with 'origin/master'.

Changes not staged for commit:
  (use "git add/rm <file>..." to update what will be committed)
  (use "git restore <file>..." to discard changes in working directory)
	deleted:    bin/rake

Untracked files:
  (use "git add <file>..." to include in what will be committed)
	blah.rb

no changes added to commit (use "git add" and/or "git commit -a")

[] ~/projects/jruby $ mvn -q
/Users/headius/projects/jruby/core/target/generated-sources/annotated_classes.txt - not found. skip generator.
using jruby 9.1.17.0
install gems unless already installed
/Users/headius/.m2/repository/rubygems/rake/12.3.2/rake-12.3.2.gem
fix the gem stub files
generating missing .bat files

[] ~/projects/jruby $ git st
On branch master
Your branch is up to date with 'origin/master'.

Changes not staged for commit:
  (use "git add/rm <file>..." to update what will be committed)
  (use "git restore <file>..." to discard changes in working directory)
	deleted:    bin/rake

Untracked files:
  (use "git add <file>..." to include in what will be committed)
	blah.rb

no changes added to commit (use "git add" and/or "git commit -a")
```

I would like to remove all bundled gems' bin scripts from our repository, but I can't because of this.

headius:
I'm not sure if the executable bits are important, since those files aren't run directly...but it would be good to preserve those too.

Issue #5863, opened 2019-09-09
ConditionVariable#wait causes thread to wait forever in 9.2.8.0


asppsa:
### Environment

+ `jruby 9.2.8.0 (2.5.3) 2019-08-12 a1ac7ff OpenJDK 64-Bit Server VM 25.222-b05 on 1.8.0_222-b05 +jit [linux-x86_64]`
+ `Linux managua 5.2.9-arch1-1-ARCH #1 SMP PREEMPT Fri Aug 16 11:29:43 UTC 2019 x86_64 GNU/Linux`
+ This issue appears to affect only 9.2.8.0.  From my testing, the issue is not present in 9.2.7.0, or in 9.1.x.
+ It does not involve any gems - just core libraries.

### Expected Behaviour

Running [this file](https://gist.github.com/asppsa/c74b654a006475c44729cc57cd3fd86e#file-test-rb) should result in "ping" and "pong" being printed repeatedly to the screen.  This works on MRI and on older versions of JRuby.

### Actual Behaviour

"ping" and "pong" are printed some number of times (it varies), and then output stops.  [Output from Ctrl+\\](https://gist.github.com/asppsa/c74b654a006475c44729cc57cd3fd86e#file-output-from-ctrl) indicates that both the main thread and Thread-1 are in a waiting state.

### Additional info

This issue has been filed on advice from @headius, after a conversation on IRC. He has suggested that https://github.com/jruby/jruby/commit/bea9ad42f62e24a3963e135f2138c75a716f7790 is the source of the issue.

headius:
Thank you for filing this one! This definitely should be fixed for 9.2.9.

headius:
Likely the same general issue as #5875.

headius:
> Likely the same general issue as #5875.

This does not appear to be the case. That issue is caused by us giving up on the lock immediately upon interrupt and raising an error. The correct behavior is up for debate, but probably should be checking for thread events (raise, kill) and if none exist try to acquire the lock again.

headius:
I've pushed a fix for this that actually uses the Mutex's JDK Lock to do the sleeping, via a JDK Condition. Your script runs properly now!

I'm still debating the complexity of the patch, since it required us to start using the native JDK thread state rather than our tightly-controlled synthetic state. This may be fine for general use cases, but there's a few tests that have become unreliable as a result.

Anyway, you can check out #5942 for the current iteration of the fix.

PR #5866, opened 2019-09-10
Test passing control between two threads in the ConditionVariable spec

asppsa:
This tests (J)Ruby's ability to safely switch back and forth between two threads that both have locks on a single mutex, as identified in #5863.

I'm happy to make changes to this spec if needed.  In particular, I wasn't sure what a suitable timeout should be (went with 5 seconds), and also whether to add this to the spec for `#signal` or `#wait`.  I went with `#signal` because it seems that the failure is in the ability to wake up after going to sleep.

### DIFF
--- a/spec/ruby/library/conditionvariable/signal_spec.rb
+++ b/spec/ruby/library/conditionvariable/signal_spec.rb
@@ -66,4 +66,51 @@ describe "ConditionVariable#signal" do
     # released in the same order
     r2.should == r1
   end
+
+  it "allows control to be passed between a pair of threads" do
+    m = Mutex.new
+    cv = ConditionVariable.new
+    repeats = 100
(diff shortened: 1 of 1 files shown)


asppsa:
Also, I should note that I used MRI 2.6 to check that this spec can pass.

asppsa:
@headius, I'll rebase this onto master to see if it passes now.

headius:
Excellent, thank you! Assuming this goes green I'll merge it in.

headius:
Two suites failed despite my best efforts, but I'm convinced they're just caching problems with the jnr-constants maven artifact. Merging this in...thank you for your thoroughness @asppsa!

eregon:
@asppsa Thanks for this good spec, it also caught a bug in TruffleRuby!

PR #5868, opened 2019-09-10
Make Symbol#to_s return frozen strings  

rdubya:
This is a PR to explore the possibility of adding support for returning frozen strings. We'll probably want to have a JRuby flag until the spec is hammered out for the ruby language in general.

This relates to https://bugs.ruby-lang.org/issues/16150

### DIFF
--- a/core/src/main/java/org/jruby/RubySymbol.java
+++ b/core/src/main/java/org/jruby/RubySymbol.java
@@ -94,6 +94,7 @@ public class RubySymbol extends RubyObject implements MarshalEncoding, EncodingC
     private final ByteList symbolBytes;
     private final int hashCode;
     private Object constant;
+    private RubyString rubyString;
 
     /**
      *
@@ -484,6 +485,10 @@ public class RubySymbol extends RubyObject implements MarshalEncoding, EncodingC
     }
 
     final RubyString to_s(Ruby runtime) {
+        if (true) { // Add check for runtime flag here?
+            if (rubyString == null) rubyString = runtime.freezeAndDedupString(RubyString.newStringShared(runtime, symbolBytes));
+            return rubyString;
+        }
         return RubyString.newStringShared(runtime, symbolBytes);
     }
 


rdubya:
There are at least a few places in rails where this breaks already. For example: https://github.com/rails/rails/blob/ae2cbf40449c270feec6fa0aa82b3d5fc2350be7/activesupport/lib/active_support/ordered_options.rb#L43

headius:
Passed all the tests we run 😮

headius:
@rdubya I'm game to add the flag. I'm starting to think we're not going to get ANY help from ruby-core on this though. They seem to be afraid of breaking one line out of a million. @enebo thoughts?

enebo:
@headius @rdubya yeah I am game to try this as it can be used to demonstrate what problems would exist if MRI decided to make this change.  I don't think they will but if there is any chance it will be needed with a ton of empirical evidence (and this flag could help provide that).

If we add this and realize it doesn't work with lots of stuff we can remove it too.

rdubya:
Any thoughts on naming or an example I can look at? Also, should I keep it symbols only for now or expand to some of the other primitives? Seems like the other primitives are probably safer than symbols, can't imagine too many people are trying to call `false.to_s.gsub!(...)`

ahorek:
fyi, it was approved as an experimental 2.7 feature
https://github.com/ruby/ruby/pull/2437
see also
https://github.com/ruby/ruby/pull/2487
https://github.com/ruby/ruby/pull/2490

eregon:
> @rdubya I'm game to add the flag. I'm starting to think we're not going to get ANY help from ruby-core on this though. They seem to be afraid of breaking one line out of a million. @enebo thoughts?

What do you mean? matz agreed to experiment with it: https://bugs.ruby-lang.org/issues/16150#note-18
(and I spent some time to push for this on your own feature request)

enebo:
@eregon you are aking @headius but I think the whole twitter thread last week on not having frozen strings by default in Ruby 3 has taken away our optimism.  Historically, changing Ruby has been much less likely than adding to it.  I do agree though that we should accept Matz moving forward with the experiment as a positive sign.

headius:
@eregon Thank you for pushing the issue forward...I have not had time to circle back to it in the past week or so.

My frustration above is that we still seem no closer to frozen strings in 3.0 and even freezing `to_s` strings for the "obvious" cases (Symbol, NilClass, True/False) is only being added as "experimental".

My interpretation of "experimental" was that it wouldn't be enabled by default in the 2.7 release...but I guess this is not the case? I'm very confused about calling it an "experimental" feature if everyone will have to adapt to it anyway (I'm glad they'll have to adapt to it, but I question what "experimental" means here).

eregon:
@headius experimental for MRI just means it might be removed before the Ruby 2.7 if it's too problematic, but otherwise it'll be part of the release. Some experimental features have a warning when they are used, e.g. Pattern matching.

headius:
@eregon Experimental for us means if it gets into a release it will be guarded by an option, which is what @rdubya and I were discussing above. We do releases more frequently than CRuby, and would like to get this feature out there for people to test against ASAP (i.e. in JRuby 9.2.9).

headius:
@rdubya Given that this will be part of a larger push to freeze strings for a few core singleton objects, let's go with a property name that reflects that.

Perhaps "frozen.core.to_s" since this applies to the `to_s` result of several "core" types?

headius:
@rdubya Oh I'd also say go ahead and do it for the other items marked as "experimental" in the ruby-lang issue, so that's nil, true, false, and Module#name.

nomadium:
@headius I continued this PR at: https://github.com/jruby/jruby/pull/6052

headius:
Closing in favor of #6052 (and because it seems like this feature will not be accepted in CRuby).

PR #5870, opened 2019-09-10
respect jit.max to stop compilation

kares:
... after some time esp. for big (Rails) apps that seemingly always execute some 'new-hot' methods

JIT parts also seemed ripe for a cleanup, so there's a few bits such as :
- not having multiple `callCount` in a hierarchy
- negated JIT condition - JITing when JIT meant to be disabled
- get JIT excludes to also exclude block compilations
- more useful ji logging - seeing exception.toString failures

aside from these, there's 2 simple but important changes : 

- `-Xjit.max=10000` default, to compilations and not leak meta-space
  even for a large Rails app this should be conservative as it takes several hours to reach the limit
  from a production app 10_000 JIT compiles means having around 550 of allocated meta-space

- `Xjit.maxsize` reduced from 2000 -> 1000 again based on real world production data
  there's very little JIT eligible large methods in Rails, so this further helps a bit saving on meta-space

also, since I will like to add JMX setters for the jit params, they are being read from the instance config

### DIFF
--- a/core/src/main/java/org/jruby/compiler/BlockJITTask.java
+++ b/core/src/main/java/org/jruby/compiler/BlockJITTask.java
@@ -26,13 +26,13 @@
 
 package org.jruby.compiler;
 
+import org.jruby.Ruby;
 import org.jruby.ast.util.SexpMaker;
 import org.jruby.ir.IRClosure;
 import org.jruby.ir.targets.JVMVisitor;
 import org.jruby.ir.targets.JVMVisitorMethodContext;
 import org.jruby.runtime.CompiledIRBlockBody;
 import org.jruby.runtime.MixedModeIRBlockBody;
-import org.jruby.util.OneShotClassLoader;
 
 import static org.jruby.compiler.MethodJITTask.*;
 
@@ -62,27 +62,16 @@ class BlockJITTask extends JITCompiler.Task {
                 return;
             }
 
-            String key = SexpMaker.sha1(body.getIRScope());
-            JVMVisitor visitor = new JVMVisitor(jitCompiler.runtime);
-            BlockJITClassGenerator generator = new BlockJITClassGenerator(className, methodName, key, jitCompiler.runtime, body, visitor);
+            final String key = SexpMaker.sha1(body.getIRScope());
+            final Ruby runtime = jitCompiler.runtime;
+            JVMVisitor visitor = new JVMVisitor(runtime);
+            BlockJITClassGenerator generator = new BlockJITClassGenerator(className, methodName, key, runtime, body, visitor);
 
             JVMVisitorMethodContext context = new JVMVisitorMethodContext();
             generator.compile(context);
 
-            // FIXME: reinstate active bytecode size check
-            // At this point we still need to reinstate the bytecode size check, to ensure we're not loading code
-            // that's so big that JVMs won't even try to compile it. Removed the check because with the new IR JIT
-            // bytecode counts often include all nested scopes, even if they'd be different methods. We need a new
-            // mechanism of getting all body sizes.
-            Class sourceClass = visitor.defineFromBytecode(body.getIRScope(), generator.bytecode(), new OneShotClassLoader(jitCompiler.runtime.getJRubyClassLoader()));
-
-            if (sourceClass == null) {
-                // class could not be found nor generated; give up on JIT and bail out
-                jitCompiler.counts.failCount.incrementAndGet();
-                return;
-            } else {
-                generator.updateCounters(jitCompiler.counts, body.ensureInstrsReady());
-            }
+            Class<?> sourceClass = defineClass(generator, visitor, body.getIRScope(), body.ensureInstrsReady());
+            if (sourceClass == null) return; // class could not be found nor generated; give up on JIT and bail out
 
             // successfully got back a jitted body
 
--- a/core/src/main/java/org/jruby/compiler/BlockJITTask.java
+++ b/core/src/main/java/org/jruby/compiler/BlockJITTask.java
@@ -56,7 +56,7 @@ class BlockJITTask extends JITCompiler.Task {
         if (excludeModuleName != null) {
             body.setCallCount(-1);
             if (jitCompiler.config.isJitLogging()) {
-                JITCompiler.log(body.getImplementationClass(), body.getFile(), body.getLine(), methodName, "skipping block in " + excludeModuleName);
+                JITCompiler.log(body, methodName, "skipping block in " + excludeModuleName);
             }
             return;
         }
@@ -95,7 +95,7 @@ class BlockJITTask extends JITCompiler.Task {
 
     @Override
     protected void logImpl(final String message, Object... reason) {
-        JITCompiler.log(body.getImplementationClass(), body.getFile(), body.getLine(), methodName, message, reason);
+        JITCompiler.log(body, methodName, message, reason);
     }
 
 }
--- a/core/src/main/java/org/jruby/compiler/BlockJITTask.java
+++ b/core/src/main/java/org/jruby/compiler/BlockJITTask.java
@@ -50,52 +50,52 @@ class BlockJITTask extends JITCompiler.Task {
     }
 
     @Override
-    public void exec() {
-        try {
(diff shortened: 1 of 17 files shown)


core/src/main/java/org/jruby/compiler/BlockJITTask.java
```diff
@@ -34,71 +34,79 @@
...
-                    generator.updateCounters(jitCompiler.counts, body.ensureInstrsReady());
+    @Override
+    public void exec() {
+        try {
+            // Check if the method has been explicitly excluded
+            String excludeModuleName = checkExcludedMethod(jitCompiler.config, className, methodName, body.getImplementationClass());
+            if (excludeModuleName != null) {
+                body.setCallCount(-1);
+                if (jitCompiler.config.isJitLogging()) {
+                    JITCompiler.log(body.getImplementationClass(), body.getFile(), body.getLine(), methodName, "skipping block in " + excludeModuleName);
```
headius: Why did this move in here? We can't really exclude blocks from JIT by method name, can we?

core/src/main/java/org/jruby/internal/runtime/AbstractIRMethod.java
```diff
@@ -32,8 +34,8 @@
     protected final IRScope method;
     protected final StaticScope staticScope;
     protected InterpreterContext interpreterContext = null;
-    protected int callCount = 0;
-    private MethodData methodData;
+    protected volatile int callCount = 0;
+    private transient MethodData methodData;
```
headius: Is the volatile here for disabling? Since our count is always checking `> threshold` I never bothered to make this volatile since someone will eventually cross the line. If it's volatile it will force a volatile read for every call.

And why transient?

core/src/main/java/org/jruby/internal/runtime/AbstractIRMethod.java
```diff
@@ -42,21 +44,40 @@ public AbstractIRMethod(IRScope method, Visibility visibility, RubyModule implem
         this.staticScope.determineModule();
         this.signature = staticScope.getSignature();
 
-        // -1 jit.threshold is way of having interpreter not promote full builds.
-        if (Options.JIT_THRESHOLD.load() == -1) callCount = -1;
+        final Ruby runtime = implementationClass.getRuntime();
+        if (runtime.getInstanceConfig().isJitDisabled()) setCallCount(-1);
```
headius: Would this be better as `! isJitEnabled()` and just reverse the logic? Something about `isSomethingNegative` methods bothers me.

core/src/main/java/org/jruby/ir/runtime/IRRuntimeHelpers.java
```diff
@@ -624,12 +624,9 @@ public static void checkArity(ThreadContext context, StaticScope scope, Object[]
...
      */
-    public static boolean shouldPrintIR(Ruby runtime) {
-        boolean booting = runtime.isBooting();
-        boolean print = Options.IR_PRINT.load();
-        boolean printAll = Options.IR_PRINT_ALL.load();
-
-        return (print && !booting) || (booting && printAll);
+    public static boolean shouldPrintIR(final Ruby runtime) {
+        final boolean booting = runtime.isBooting();
+        return (Options.IR_PRINT.load() && !booting) || (booting && Options.IR_PRINT_ALL.load());
```
headius: I kinda like the original better since the condition reads nice.. "if print and not booting, or booting and print all..."

core/src/main/java/org/jruby/runtime/InterpretedIRBlockBody.java
```diff
@@ -144,10 +144,19 @@ protected IRubyObject commonYieldPath(ThreadContext context, Block block, Block.
...
+        if (this.callCount < 0) return;
+        // we don't synchronize callCount++ it does not matter if count isn't accurate
+        if (this.callCount++ >= runtime.getInstanceConfig().getJitThreshold()) {
+            synchronized (this) { // disable same jit tasks from entering queue twice
+                if (this.callCount >= 0) {
+                    this.callCount = Integer.MIN_VALUE; // so that callCount++ stays < 0
+
+                    runtime.getJITCompiler().buildThresholdReached(context, this);
+                }
+            }
```
headius: This logic is very similar to that inside the method/block jit stuff, isn't it?

core/src/main/java/org/jruby/runtime/MixedModeIRBlockBody.java
```diff
@@ -15,29 +17,24 @@
...
-
 public class MixedModeIRBlockBody extends IRBlockBody implements Compilable<CompiledIRBlockBody> {
     private static final Logger LOG = LoggerFactory.getLogger(MixedModeIRBlockBody.class);
 
     protected boolean pushScope;
     protected boolean reuseParentScope;
     private boolean displayedCFG = false; // FIXME: Remove when we find nicer way of logging CFG
-    private volatile int callCount = 0;
     private InterpreterContext interpreterContext;
+    private volatile int callCount = 0;
```
headius: Same thoughts about volatility 🤔

headius:
default JIT max seems reasonable. I wonder if we should warn in some mode (maybe always, maybe only verbose mode, I don't know) that the limit has been reached? 10k methods is a lot, especially if a large Rails app is taking hours to get there.

2000 to 1000 limit seems reasonable because this is IR count and IR is much more dense than bytecode. I'll be doing some review of all IR jit logic to see what else we can shrink, also.

Live JIT flags seem good too...useful for tuning exploration etc.

core/src/main/java/org/jruby/runtime/MixedModeIRBlockBody.java
kares: yy, do not think it needs to be volatile.
but it kind of was before for the MixedModel impls

core/src/main/java/org/jruby/internal/runtime/AbstractIRMethod.java
kares: transient just a marker that its a cached field. 
method pbly can't serialize state due the IR scope but if it did this field is not needed

core/src/main/java/org/jruby/compiler/BlockJITTask.java
kares: believe we can - if a method is excluded its blocks should also not JIT. however I did not test if that case works - just noticed on aprod instance that blocks are being JITed for an excluded class. that case, an excluded mod or class, is reasonable and should be handled here ...

core/src/main/java/org/jruby/internal/runtime/AbstractIRMethod.java
kares: +1 will update

core/src/main/java/org/jruby/runtime/InterpretedIRBlockBody.java
kares: yy - would need a common super type if we're to keep field read write (this.callCount++)

kares:
now updated with more work, parts of which I am going to need to test out JITing in production.
(namely the possibility to change jit parameters from JMX - code should be fine to handle that)

also, have tested locally - block JIT excludes work as well as reaching jit.max halts compilation

headius:
I'm fine merging this any time. @enebo should probably coordinate with you because I think he was playing with some stuff that might overlap.

kares:
actually do have @enebo's *counter_nuke* work merged on top of this (in another branch) with some fine tuning (ability to disable jit.time.delta)

enebo:
@kares @headius I think merging this will be fine even it counter_nuke ends up solving the original issue as well.  if aging works really well then a jit.max may never be used but I could see it as an insurance policy.

A second thought is we should consider decrementing jit count if an oneshotclassloader unloads because the method went away.  I do not think that is important for this PR and that may be more complicated with the chunked classloading I am looking at.  Still something to think about.

kares:
good, was also thinking of having an insurance policy (jit.max) regardless of other work ...
merging, will put up further work for review (on top of this) - the class-loader sharing piece.

PR #5874, opened 2019-09-12
Reduce bytecode size from JIT

headius:
First commit in this branch explains things best, but in general this PR reduces JIT bytecode size and complexity by initializing lazy constant values (literals, call sites objects, etc) using invokedynamic.


headius:
I've requested reviews from @kares and @enebo. I know there's a lot of changes here but they're all pretty small individually. Mostly I'm looking for more validation that this does not hurt us, and ideally helps us wrt bytecode size and perhaps some startup/warmup/peak performance improvements.

I did run some numbers against a trivial Roda app and this branch performs a good 15% better than master.

Issue #5773, opened 2019-06-26
SecureRandom does not implement #bytes like MRI


kreynolds:
### Environment

Provide at least:
* jruby 9.2.6.0 (2.5.3) 2019-02-11 15ba00b OpenJDK 64-Bit Server VM 11.0.2+9 on 11.0.2+9 +jit [darwin-x86_64]
* Darwin Kelleys-MBP.neo.rr.com 18.6.0 Darwin Kernel Version 18.6.0: Thu Apr 25 23:16:27 PDT 2019; root:xnu-4903.261.4~2/RELEASE_X86_64 x86_64

### Expected Behavior

* SecureRandom should implement `#bytes` to match MRI. For example, MRI implements `#bytes` and `#gen_random`, and the remaining methods are extended with a formatter.

### Actual Behavior

* It does not implement `#bytes` at all.

headius:
We implement `random_bytes`...perhaps this is just a new alias?

headius:
Looks like it was an alias that became a wrapper method, and we just never added it (or mistakenly removed it).

headius:
Trivial fix coming.

It looks like there's specs for `Random#bytes` [here](https://github.com/jruby/jruby/blob/d850ba54e4c11b93ed59da341325eb74457602f0/spec/ruby/core/random/bytes_spec.rb) but nothing for `SecureRandom.bytes` [here](https://github.com/jruby/jruby/tree/d629c9464791d01f20f5cb805c45f287b06de1c1/spec/ruby/library/securerandom). Could you have a look and see if we can share or duplicate these specs? cc @eregon for help moving this to something shared, if that's necessary.

headius:
FWIW this following workaround basically does what I did:

```ruby
class << SecureRandom
  def bytes(n); random_bytes(n); end
  alias gen_random bytes
end
```

kreynolds:
Yeah, I know its a trivial alias, I just filed in the interest of maintaining general compatibility. Thanks for knocking this one out :)

PR #5878, opened 2019-09-15
add Random.random_number

ahorek:
random_number is just an alias to rand

fixes
```
Random.new.random_number
```

`Random.random_number` (and `SecureRandom.bytes` https://github.com/jruby/jruby/issues/5773) is actually a ruby 2.6 feature, should I keep it or move it to 2.6 branch?

### DIFF
--- a/core/src/main/java/org/jruby/RubyRandom.java
+++ b/core/src/main/java/org/jruby/RubyRandom.java
@@ -273,7 +273,7 @@ public class RubyRandom extends RubyObject {
         return this;
     }
 
-    @JRubyMethod(name = "rand", meta = true, optional = 1)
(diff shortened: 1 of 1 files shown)


ahorek:
looking at the source again, it looks like SecureRandom dynamically defines this method after
```require 'securerandom'```
https://github.com/ruby/ruby/blob/master/random.c#L1567

ahorek:
SecureRandom should automatically extend the original Random class https://github.com/ruby/ruby/blob/c5eb24349a4535948514fe765c3ddb0628d81004/lib/securerandom.rb#L135

do we wanna support extending Random via Random::Formatter? anyway this PR is wrong, so closing for now.

PR #5887, opened 2019-09-24
[jit] replace method invocation counter with jit.time.delta

kares:
this is @enebo's work from https://github.com/jruby/jruby/tree/counter_nuke
rebased + with a few touches for usability (able to specify *jit.time.delta* as long)

might need more work as we still do not have real-world numbers (hopefully soon).
... or we can add it into next 9.2 but having *jit.time.delta = -1* (disabled) by default?


enebo:
I believe this is safe but without pushback I am skipping for somewhat imminent 9.2.10.0.

headius:
I merged latest master into this to see where we stand on it, but I don't see us making this change right before 9.3.

kares:
we did not do a lot of experiments with `jit.time.delta`, but the feature could be off by default (as presented in the PR).
the added overhead (when off) is simply the `nanoTime()` call on every method instantiation ...

PR #5889, opened 2019-09-25
[feat] jit.exclude based on path/to/impl_file.rb

kares:
since there's currently no way to exclude dynamic evals (e.g. from Rails)

allows us to handle excludes as long as FILE is set ... e.g. [excluding generated attribute methods](https://github.com/rails/rails/blob/4-2-stable/activerecord/lib/active_record/attribute_methods.rb#L48)


kares:
tested the exclude manually - working fine.

PR #5903, opened 2019-10-02
less locking in interpreter mode (due IRMethod)

kares:
a load test profiled showed some lower (but likely often unnecessary) numbers around:
```
org.jruby.ir.IRMethod.lazilyAcquireInterpreterContext()	
org.jruby.internal.runtime.methods.MixedModeIRMethod.ensureInstrsReady()	
org.jruby.internal.runtime.methods.MixedModeIRMethod.call(ThreadContext, IRubyObject, RubyModule, String, IRubyObject, IRubyObject, Block)	
org.jruby.RubyClass.finvoke(ThreadContext, IRubyObject, String, IRubyObject, IRubyObject, Block)	
org.jruby.RubyBasicObject.send(ThreadContext, IRubyObject, IRubyObject, IRubyObject, Block)	
org.jruby.RubyKernel.send(ThreadContext, IRubyObject, IRubyObject, IRubyObject, IRubyObject, Block)
org.jruby.RubyKernel$INVOKER$s$send.call(ThreadContext, IRubyObject, RubyModule, String, IRubyObject, IRubyObject, IRubyObject, Block)	
org.jruby.runtime.callsite.CachingCallSite.call(ThreadContext, IRubyObject, IRubyObject, IRubyObject, IRubyObject, IRubyObject, Block)	
jruby.$2_dot_3_dot_0.gems.actionview_minus_4_dot_2_dot_11_dot_1.lib.action_view.template.invokeOther2:send(ThreadContext, IRubyObject, IRubyObject, IRubyObject, IRubyObject, IRubyObject, Block)
jruby.$2_dot_3_dot_0.gems.actionview_minus_4_dot_2_dot_11_dot_1.lib.action_view.template.RUBY$block$render$1(ThreadContext, Block, StaticScope, IRubyObject, IRubyObject[], Block)	
```
in a sample 3 threads were blocked for a total of 400ms (and more samples showed a similar trace)

### DIFF
--- a/core/src/main/java/org/jruby/ir/IRMethod.java
+++ b/core/src/main/java/org/jruby/ir/IRMethod.java
@@ -43,8 +43,7 @@ public class IRMethod extends IRScope {
     }
 
     private synchronized void buildMethodImpl() {
-        DefNode defNode = this.defNode;
-        if (defNode == null) return;
+        if (hasBeenBuilt()) return;
 
         IRBuilder.topIRBuilder(getManager(), this).
                 defineMethodInner(defNode, getLexicalParent(), getFlags().contains(IRFlags.CODE_COVERAGE)); // sets interpreterContext
--- a/core/src/main/java/org/jruby/ir/IRMethod.java
+++ b/core/src/main/java/org/jruby/ir/IRMethod.java
@@ -15,13 +15,13 @@ public class IRMethod extends IRScope {
     // Argument description
     protected ArgumentDescriptor[] argDesc = ArgumentDescriptor.EMPTY_ARRAY;
 
-    private DefNode defn;
+    private volatile DefNode defNode;
 
     public IRMethod(IRManager manager, IRScope lexicalParent, DefNode defn, ByteList name,
             boolean isInstanceMethod, int lineNumber, StaticScope staticScope, boolean needsCodeCoverage) {
         super(manager, lexicalParent, name, lineNumber, staticScope);
 
-        this.defn = defn;
+        this.defNode = defn;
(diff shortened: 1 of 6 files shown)


core/src/main/java/org/jruby/ir/IRMethod.java
```diff
@@ -33,25 +33,28 @@ public IRMethod(IRManager manager, IRScope lexicalParent, DefNode defn, ByteList
...
+        if (!hasBeenBuilt()) buildMethodImpl();
 
         return interpreterContext;
     }
 
-    public synchronized BasicBlock[] prepareForCompilation() {
-        if (!hasBeenBuilt()) lazilyAcquireInterpreterContext();
+    private synchronized void buildMethodImpl() {
+        DefNode defNode = this.defNode;
+        if (defNode == null) return;
```
enebo: You removed use of commont hasBeenBuilt here.  Was it a perf thing?

kares: defNode==nul  does it + since we read the field ...
not completely happy about this but it works fine

enebo: @kares so this was changed because of worry that this.defNode will end up getting nulled if it kept using the helper method?  Since ti is synchronized that is not an issue though is it?  

My reason for these comments is I like to see hasBeenBuilt vs the code used to check it because it explains what that check is for.

kares: not really, been just lazily following read volatile var once in a method automatically.
will change it to suit your needs. thanks for the review.

PR #5907, opened 2019-10-07
Properly override base env with supplied values for spawn

### LINKED ISSUE #3428: Process.spawn doesn't remove nil env variables
`Process.spawn` is supposed to unset environment variables that are explicitly specified as `nil`. Instead, the environment variable is being set as `""`. This happens even if the environment variable didn't exist before calling `Process.spawn`.

Below is a sample script demonstrating the problem:

``` ruby
#!/usr/bin/env ruby
if ARGV.first != "spawned"
  Process.wait Process.spawn({ "SHOULD_BE_NIL" => nil }, __FILE__, "spawned")
  exit
end

if ENV["SHOULD_BE_NIL"].nil?
  puts "success!"
else
  puts "failure: #{ENV["SHOULD_BE_NIL"].inspect}"
end
```

When run on `Ruby 2.2.3` and `JRuby 1.7.22`, it works as expected... I get `success!`, but on `JRuby 9.0.3.0`, I get `failure: ""`.

I'd be happy to work on a fix for this if it hasn't already been solved (I found some similar looking issues open, but couldn't find one that quite matched it).

matthewd:
Hi!

I have no idea what I'm doing (and don't have a local dev environment, so I'm making things up without even a typecheck to back me up :pensive:), so this is probably disastrously wrong in detail, but...

I think something along these lines fixes #3428, as well as a wider issue:

```
% env | grep EDITOR
EDITOR=vim

% ruby -e 'system({ "EDITOR" => "emacs" }, "env")' | grep EDITOR
EDITOR=emacs

% rbenv shell jruby-9.2.8.0
% ruby -e 'system({ "EDITOR" => "emacs" }, "env")' | grep EDITOR
EDITOR=vim
EDITOR=emacs
```

POSIXly speaking (without having checked the relevant docs), that seems Bad. Even without the not-removing-nils issue, it appears to be a bit hit-or-miss which entry a process will choose to honour... I'm struggling to show it in isolation, but I got here because I was seeing the "old" value win in my real usage.

### DIFF
--- a/core/src/main/java/org/jruby/util/ShellLauncher.java
+++ b/core/src/main/java/org/jruby/util/ShellLauncher.java
@@ -234,32 +234,12 @@ public class ShellLauncher {
         try {
             // dup for JRUBY-6603 (avoid concurrent modification while we walk it)
             RubyHash hash = null;
-            if (!clearEnv) {
+            if (clearEnv) {
+                hash = RubyHash.newHash(runtime);
+            } else {
                 hash = (RubyHash) runtime.getObject().getConstant("ENV").dup();
             }
-            String[] ret;
 
-            if (mergeEnv != null) {
-                ret = new String[hash.size() + mergeEnv.size()];
-            } else {
-                ret = new String[hash.size()];
-            }
-
-            int i = 0;
-            if (hash != null) {
-                for (Map.Entry<String, String> e : (Set<Map.Entry<String, String>>)hash.entrySet()) {
-                    // if the key is nil, raise TypeError
-                    if (e.getKey() == null) {
-                        throw runtime.newTypeError(runtime.getNil(), runtime.getStructClass());
-                    }
-                    // ignore if the value is nil
-                    if (e.getValue() == null) {
-                        continue;
-                    }
-                    ret[i] = e.getKey() + '=' + e.getValue();
-                    i++;
-                }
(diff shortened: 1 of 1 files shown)


headius:
This slipped through the cracks but can probably be merged for 9.2.10. Note that ShellLauncher is only used when we can't (or don't) use true native process lauching via PopenExecutor, so this code will only affect a subset of process logic.

headius:
Remarkably, I didn't even know it was possible to have the same entry twice in env.

matthewd:
> Note that ShellLauncher is only used when we can't (or don't) use true native process lauching via PopenExecutor

Ah, this had me confused for a moment -- I'm on macOS, so would expect popen to be doing the thing. But it turns out:

https://github.com/jruby/jruby/blob/439acebb7da3ef0fc815029877324bcea65ec65c/core/src/main/java/org/jruby/util/io/PopenExecutor.java#L1016


> Remarkably, I didn't even know it was possible to have the same entry twice in env.

Me neither! In retrospect it follows from the "series of contiguous `\0`-separated key-value pairs" definition that it's _possible_, but I suspect it's not _valid_, and very much doubt most applications have given conscious thought to the eventuality.

headius:
@matthewd Can you come up with a spec for this behavior?

Issue #5906, opened 2019-10-04
Java::JavaLang::Throwable missing full_message method


davishmcclurg:
I don't know if this is expected behavior, but I was surprised that `Java::JavaLang::Throwable` does not implement `full_message`:

```
irb(main):011:0> JRUBY_VERSION
=> "9.2.8.0"
irb(main):012:0> Java::JavaLang::Throwable.new('ugh').full_message
Traceback (most recent call last):
        6: from /Users/dharsha/.rbenv/versions/jruby-9.2.8.0/bin/irb:13:in `<main>'
        5: from org/jruby/RubyKernel.java:1193:in `catch'
        4: from org/jruby/RubyKernel.java:1193:in `catch'
        3: from org/jruby/RubyKernel.java:1425:in `loop'
        2: from org/jruby/RubyKernel.java:1061:in `eval'
        1: from (irb):12:in `evaluate'
NoMethodError (undefined method `full_message' for java.lang.Throwable: ugh:Java::JavaLang::Throwable)
Did you mean?  message
irb(main):013:0> StandardError.new('ugh').full_message
=> "(irb):12: ugh (StandardError)\n"
```

kares:
throwable is the more general Java object here and Ruby exceptions are throwables :)
but yeah, the concern is valid as JRuby makes throwables Ruby Exception-like.

you could setup `full_message` as an alias to `inspect`, it would likely get merged ...
https://github.com/jruby/jruby/blob/master/core/src/main/java/org/jruby/javasupport/ext/JavaLang.java#L244 (the format would be slightly different but I think its a fine start)

PR #5908, opened 2019-10-08
Implement Java::JavaLang::Throwable#full_message

davishmcclurg:
Since Java throwables are treated similarly to regular Ruby exceptions,
I think it makes sense to implement `full_message` for easier logging
and debugging.

This moves the `RubyException` implementation to `TraceType` so that it
can be shared with `Throwable`. I had to change the `exception` argument
type in `printBacktraceMRI` in order to support the exception-like
object that `Throwable` provides.

Closes: https://github.com/jruby/jruby/issues/5906

### DIFF
--- a/core/src/main/java/org/jruby/RubyException.java
+++ b/core/src/main/java/org/jruby/RubyException.java
@@ -39,7 +39,6 @@ package org.jruby;
 
 import org.jruby.anno.JRubyClass;
 import org.jruby.anno.JRubyMethod;
-import org.jruby.ast.util.ArgsUtil;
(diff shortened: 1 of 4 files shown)


headius:
The original intent of actually propagating Java exceptions was indeed to make them duck type as Ruby exceptions, so this makes sense.

headius:
Looks good to me. The whitespace changes are a little annoying, but the whitespace that was there shouldn't have been. If you do any future PRs, try to keep the whitespace changes in a separate commit, so they don't show up in git blame as being related to the actual change.

Thank you for the patch!

Issue #3205, opened 2015-08-01
Dir.foreach, Dir.new, Dir.open: Missing/incompatible encoding support.


Freaky:
Ruby 2.2.2:

```
 Dir.new("/var/empty", encoding: Encoding.find("filesystem")).to_a # => [".", ".."]
 Dir.open("/var/empty", encoding: Encoding.find("filesystem")).to_a # => [".", ".."]
 Dir.foreach("/var/empty", encoding: Encoding.find("filesystem")).to_a # => [".", ".."]
```

JRuby 9.0.0.0 yields, respectively:

```
ArgumentError: wrong number of arguments calling `initialize` (2 for 1)
ArgumentError: wrong number of arguments calling `open` (2 for 1)
TypeError: wrong argument type Hash (expected Encoding)
```

The following works in JRuby:

```
Dir.foreach("/var/empty", Encoding.find("filesystem")).to_a # => [".", ".."]
Dir.new("/var/empty").each(Encoding.find("filesystem")).to_a # => [".", ".."]
```

But in Ruby 2.2.2 these raise:

```
ArgumentError: wrong number of arguments (2 for 1)
ArgumentError: wrong number of arguments (1 for 0)
```

It would seem the resolution to #2547 in 26ed114efb12f1b59a2f3f7915b876e80c349a31 is incorrect.

MariuszCwikla:
I would like to help on this. I did already some initial investigation and coding.
This also look like a duplicate of issue #4495, isn't it?

MariuszCwikla:
I prepared initial fix but I have some observations and questions.

1. `Dir.open`, `Dir.new`, `Dir.foreach` support encoding both as string and as Encoding in CRuby:

```
p Dir.open('.', encoding: 'UTF-8').entries
p Dir.open('.', encoding: Encoding.find('UTF-8')).entries
p Dir.new('.', encoding: 'UTF-8').entries
p Dir.new('.', encoding: Encoding.find('UTF-8')).entries
Dir.foreach('.', encoding: 'UTF-8'){|f| p f}
Dir.foreach('.', encoding: Encoding.find('UTF-8')){|f| p f}
```

2. I found MRI test `test_dir_enc` and implemented changes to make the test pass (i.e. encoding as String).
I didn't find any MRI test with `Encoding` type.
I would like to add them, but I'm not sure where to add. Should this be `test_dir.rb` under `test/mri` or under `test/jruby`?
[...]

PR #5915, opened 2019-10-14
encoding parameter for Dir.open, Dir.new #3205 #4495

### LINKED ISSUE #4495: Dir.open missing optional second keyword argument
test_dir.rb:test_unknown_keywords shows one failure but no point adding the test for bogus keywords until we have any keyword argument support for it at all.

MariuszCwikla:
Fixes #4495 and partial fix for #3205

### DIFF
--- a/core/src/main/java/org/jruby/RubyDir.java
+++ b/core/src/main/java/org/jruby/RubyDir.java
@@ -304,8 +304,9 @@ public class RubyDir extends RubyObject implements Closeable {
      * @return all entries for this Dir
      */
     @JRubyMethod(name = "entries")
-    public RubyArray entries() {
-        return RubyArray.newArrayMayCopy(getRuntime(), JavaUtil.convertJavaArrayToRuby(getRuntime(), snapshot));
+    public RubyArray entries() {    	
+        JavaUtil.StringConverter converter = new JavaUtil.StringConverter(encoding);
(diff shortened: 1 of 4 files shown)


MariuszCwikla:
In f5b78ad I've added fix for `foreach` and more tests

kares:
looking good, still could we maybe have : `Dir.mkdir("./testDir_1")` in a test setup
and also do a test teardown part to cleanup all of the created dir's content and remove it.

MariuszCwikla:
@kares thanks for hints. Normally I would do like you suggested, but `teardown` already performs cleanup this in it's own "interesting" ;) way:
- `setup` deletes `testDir_1` up to` testDir_5`
- `teardown` calls `setup` ;) 

Do you agree to do more refactoring (including setup/teardown and other existing tests not related to my change?)

kares:
@MariuszCwikla oh I did not check, in that case all :+1: from me

headius:
The modified `open` and `initialize` could be split into one- and two-argument versions, which will usually avoid the `args[]` boxing, but it's a minor item. I'll make the fix quickly on master since we'd like this in 9.2.9.

headius:
I notice that one particular piece of logic seems to be lost: in `foreach`, if an `Encoding` was passed in instead of an options hash, it would be used without trying to dig it out of options.

Was that behavior in error? It appears to have been there since 1.9.

I will commit a change to remove the [] bits, but we do need confirmation about that logic.

PR #5917, opened 2019-10-15
Update stdlib to 2.5.7.

headius:


lib/ruby/stdlib/net/http.rb
```diff
@@ -633,10 +633,6 @@ class << HTTP
     # is required to use the proxy, and p_no_proxy hosts which do not
     # use the proxy.
     #
-    # In JRuby, this will default to the JSE proxy settings provided in the
-    # 'http.proxyHost' and 'http.proxyPort' Java system properties, if they
-    # are set, falling back on environment variables otherwise.
-    #
```
olleolleolle: This comment - is it to be removed or kept?

headius: Yeah I'm confused why this didn't get picked up when I updated our fork. I'm re-pushing without this and some other non-functional OpenSSL changes.

headius:
The complete diff from Ruby 2.5.7 is here: https://github.com/jruby/ruby/compare/v2_5_7...jruby-ruby_2_5_7

Issue #5909, opened 2019-10-09
Concurrent File#flock calls of the same file results in Errno::EINVAL exception on Windows


meineerde:
### Environment

* `jruby 9.2.8.0 (2.5.3) 2019-08-12 a1ac7ff OpenJDK 64-Bit Server VM 25.222-b10 on 1.8.0_222-ojdkbuild-b10 +jit [mswin32-x86_64]`
* Operating System: Windows 10 1803

### Expected Behavior

Consider the following code:

```ruby
require 'tempfile'
tmpfile = Tempfile.new('')

10.times do
  Thread.new do
    mode = File::WRONLY | File::APPEND | File::CREAT
    file = File.new(tmpfile.path, mode: mode)

    50.times do
      file.flock(File::LOCK_EX)
      file.puts Thread.current.object_id
      file.flock(File::LOCK_UN)
    end
  end
end
```

Each Thread should write its `object_id` 50 times to the file while exclusively flocking it around each write. The flocks should not interfere with each other and should not result in an exception. Instead, they should serialize the writes of the different threads.

This works as expected on both Linux (e.g. on Travis CI) as well as macOS with JRuby and MRI. It also works correctly with all MRI versions between 2.1 - trunk, even on Windows.

### Actual Behavior

With the current versions of JRuby 9.0, 9.1, and 9.2 on Windows (e.g. on AppVeyor), the above code results in the following error in some of the concurrent threads when attempting the `file.flock(File::LOCK_EX)`:

```
Errno::EINVAL Invalid argument - C:/path/to/file
    flock at org/jruby/RubyFile.java:342
```

Note: Since this appears to be a race condition, you might have to adjust the number of threads and the inner loop a bit to reliably reproduce this depending on your environment or try the code multiple times.

headius:
I'm guessing it's the EINVAL from here: https://github.com/jruby/jruby/blob/e255434b319780d05c607f0b45e2d0ae4c8eef50/core/src/main/java/org/jruby/util/io/PosixShim.java#L229-L232

When on Windows, I believe we are not using a true native `flock` functions. Instead we are leaning on the JDK's own file locking logic, which has some differing semantics.

In this case, assuming the EINVAL is from the above code, it's happening because the file appears to already be locked by this process and we attempt to lock it again. That could be a race in our code or in the JDK code for file locking. I'm poking around a bit now

headius:
[...]

Issue #5519, opened 2018-12-14
teardown with threads waiting gets ugly


kares:
tried to run a suite for [adhearsion](https://github.com/adhearsion/adhearsion)

- `git clone git@github.com:adhearsion/adhearsion.git`
- `bundle install`
- `rake spec`

all is fine, except for an ugly warning printed from a thread (stucked) on teardown.

... so far haven't tried reducing the case to a minimal piece, but its fairly understandable
there's an initializer thread, seemingly, [doing signal handling](https://github.com/adhearsion/adhearsion/blob/develop/lib/adhearsion/initializer.rb#L87-L99) started on boot:

```ruby
      self_read, self_write = IO.pipe

      %w(INT TERM HUP ALRM ABRT USR2).each do |sig|
        trap sig do
          self_write.puts sig
        end
      end

      Thread.new do
        begin
          while readable_io = IO.select([self_read])
            signal = readable_io.first[0].gets.strip
            handle_signal signal
          end
        rescue => e
          logger.error "Crashed reading signals"
          logger.error e
          exit 1
        end
      end
```

... and by the time the process exits (`runtime.teardown()`) : 
```java
    public void teardown() {
        // clear all thread-local context references
        localContext = new ThreadLocal<SoftReference<ThreadContext>>();

        // clear main context reference
        mainContext = null;

        // clear thread map
        rubyThreadMap.clear();
   }
```

the thread is still there trying to raise an exception, obviously hitting a **null** at `mainContext`

thinking JRuby should know its shutting down and not attempt to raise from threads that are still around ... as an improvement for the GH-3313 fix: https://github.com/jruby/jruby/commit/9f72b0c83ec0087ef100a1eefee75d615e9f5544

JRuby 9.2.5.0 ends and prints a stack-trace :  
```
Finished in 4 minutes 16.6 seconds (files took 9.34 seconds to load)
2541 examples, 0 failures, 27 pending

[Coveralls] Outside the Travis environment, not sending data.
Exception in thread "Ruby-0-Thread-3871: /home/kares/workspace/oss/adhearsion/lib/adhearsion/initializer.rb:95" Exception in thread "Ruby-0-Thread-3928: /home/kares/workspace/oss/adhearsion/lib/adhearsion/initializer.rb:95" Exception in thread "Ruby-0-Thread-3909: /home/kares/workspace/oss/adhearsion/lib/adhearsion/initializer.rb:95" java.lang.NullPointerException
[...]

PR #5918, opened 2019-10-16
Warn when teardown is called with active threads.

headius:
This has come up many times; Ruby threads are daemon threads by
default, which means they will not keep the main script from
exiting. As a result, if there's any threads still runnign when
the main script exits, we may tear down resources they need to
run properly. Even though they're about to die, they may produce
errors on the way out as shown in https://github.com/jruby/jruby/issues/5909#issuecomment-542470809
and other bug reports over the years like #5519, #3316, #3313 and
others.

This is not a fix for the behavior, but it introduces a
non-verbose warning if the JRuby runtime is torn down while
there's still active threads.

### DIFF
--- a/core/src/main/java/org/jruby/Ruby.java
+++ b/core/src/main/java/org/jruby/Ruby.java
@@ -3297,6 +3297,12 @@ public final class Ruby implements Constantizable {
     public void tearDown(boolean systemExit) {
         int status = 0;
 
+        // Check if there's live threads, since teardown may interfere with them
+        RubyThread[] activeRubyThreads = getThreadService().getActiveRubyThreads();
+        if (activeRubyThreads.length > 1) {
+            warnings.warn("teardown while live threads exist:\n" + Arrays.toString(activeRubyThreads));
+        }
+
         // clear out old style recursion guards so they don't leak
         mriRecursionGuard = null;
 


headius:
Example warning, from the flock script in #5909 

```
(unknown):0: warning: teardown while live threads exist:
[Thread[Ruby-0-Thread-12: flocker.rb:1,5,main], Thread[Ruby-0-Thread-18: flocker.rb:1,5,main], Thread[Ruby-0-Thread-1: flocker.rb:1,5,main], Thread[Ruby-0-Thread-8: flocker.rb:1,5,main], Thread[Ruby-0-Thread-7: flocker.rb:1,5,main], Thread[main,5,main], Thread[Ruby-0-Thread-4: flocker.rb:1,5,main], Thread[Ruby-0-Thread-5: flocker.rb:1,5,main], Thread[Ruby-0-Thread-15: flocker.rb:1,5,main], Thread[Ruby-0-Thread-19: flocker.rb:1,5,main], Thread[Ruby-0-Thread-9: flocker.rb:1,5,main], Thread[Ruby-0-Thread-11: flocker.rb:1,5,main], Thread[Ruby-0-Thread-17: flocker.rb:1,5,main], Thread[Ruby-0-Thread-16: flocker.rb:1,5,main], Thread[Ruby-0-Thread-20: flocker.rb:1,5,main], Thread[Ruby-0-Thread-2: flocker.rb:1,5,main], Thread[Ruby-0-Thread-13: flocker.rb:1,5,main], Thread[Ruby-0-Thread-6: flocker.rb:1,5,main], Thread[Ruby-0-Thread-14: flocker.rb:1,5,main], Thread[Ruby-0-Thread-10: flocker.rb:1,5,main], Thread[Ruby-0-Thread-3: flocker.rb:1,5,main]]
```

enebo:
@headius I think we will see this occasionally on Rails because of rb-inotify.  It went away for me locally but people still see it so in what I thought was fixed on Rails side somehow is something weirder.

headius:
@enebo So what's your verdict on including this? Perhaps 9.2.10 and we can let it bake for a while?

enebo:
I think we should bake this for .10

enebo:
@headius I notice we fail a few things with this PR but it would be nice to land earlier than later.

headius:
This is moot now; on teardown we attempt to terminate any remaining running Ruby threads. See #6176.

Issue #5910, opened 2019-10-10
'Cannot cast org.jruby.gen.RubyObject7 to org.jruby.gen.RubyObject4' errors


jordan-thoms:
### Environment

Provide at least:
* JRuby version (`jruby -v`) and command line (flags, JRUBY_OPTS, etc)
jruby 9.2.8.0 (2.5.3) 2019-08-12 a1ac7ff OpenJDK 64-Bit Server VM 25.222-b10 on 1.8.0_222-8u222-b10-1ubuntu1~18.04.1-b10 +jit [linux-x86_64]

java -XX:+UseCompressedOops -XX:+CMSClassUnloadingEnabled -XX:+UseConcMarkSweepGC -Dfile.encoding=UTF-8 -Xmx7000m -Xss512k -Djffi.boot.library.path=/opt/jruby-9.2.8.0/lib/jni -Djava.security.egd=file:/dev/urandom -Xbootclasspath/a:/opt/jruby-9.2.8.0/lib/jruby.jar -classpath : -Djruby.home=/opt/jruby-9.2.8.0 -Djruby.lib=/opt/jruby-9.2.8.0/lib -Djruby.script=jruby -Djruby.shell=/bin/sh org.jruby.Main /opt/jruby-9.2.8.0/bin/bundle exec sidekiq -q pdf -q pdf_auto_sync -c 80 -r ./jworker.rb

* Operating system and platform (e.g. `uname -a`)
Linux kami-jworker-5cd6444676-ll7kx 4.14.137+ #1 SMP Thu Aug 8 02:47:02 PDT 2019 x86_64 x86_64 x86_64 GNU/Linux

Other relevant info you may wish to add:
* Installed or activated gems
* Application/framework version (e.g. Rails, Sinatra)
* Environment variables

### Expected Behavior
JRuby should not have internal exceptions

### Actual Behavior

On rare occasions (we run ~80 k8s pods with this workload at peak each day, and this situation happens for a single one of those processes about once a week), suddenly all requests we attempt to make fail with a ClassCastException  (stacktrace below). Interestingly, whenever this happens, it's the stacktrace seems consistent - it's always relating to 'gems/http-cookie-1.0.3/lib/http/cookie_jar/abstract_store.rb:45 '.  It seems to also always be specifically failing trying to cast a 'org.jruby.gen.RubyObject7' to 'org.jruby.gen.RubyObject4' object - as in those same class names every time - I'm not sure what significance that has.

[...]

PR #5919, opened 2019-10-16
Synchronize variable reification against real class.

headius:
This is likely at least a partial fix for #5910, where it appears
that multiple threads all trigger reificatin of a class's
variables at the same time, leading to some of those threads
choosing a different reified type and ultimately class casting
exceptions.

The change here synchronizes against the class's "real" class,
which should be the one on which we attach the allocated. This
should force multiple threads all allocating the first instance
at once to wait for each other, eventually falling back on the
new allocator that does not re-reify the type.

This will eventually be moot once individual object instances hold
a reference to their own layout managers.

### DIFF
--- a/core/src/main/java/org/jruby/RubyObject.java
+++ b/core/src/main/java/org/jruby/RubyObject.java
@@ -153,28 +153,32 @@ public class RubyObject extends RubyBasicObject {
     public static final ObjectAllocator IVAR_INSPECTING_OBJECT_ALLOCATOR = new ObjectAllocator() {
         @Override
         public IRubyObject allocate(Ruby runtime, RubyClass klass) {
-            synchronized (klass.getRealClass()) {
-                ObjectAllocator allocator = klass.getAllocator();
+            ObjectAllocator allocator = klass.getAllocator();
 
-                if (allocator == this) {
-                    // proceed, we are the first one to get here
-                    Set<String> foundVariables = klass.discoverInstanceVariables();
+            if (allocator == this) {
+                // eagerly gather variables outside of sync
+                Set<String> foundVariables = klass.discoverInstanceVariables();
 
-                    if (Options.DUMP_INSTANCE_VARS.load()) {
-                        System.err.println(klass + ";" + foundVariables);
+                synchronized (klass.getRealClass()) {
+                    // check again before reifying
+                    allocator = klass.getAllocator();
+
+                    if (allocator == this) {
+                        // proceed, we are the first one to get here
+
+                        if (Options.DUMP_INSTANCE_VARS.load()) {
+                            System.err.println(klass + ";" + foundVariables);
+                        }
+
+                        allocator = RubyObjectSpecializer.specializeForVariables(klass, foundVariables);
+
+                        // invalidate metaclass so new allocator is picked up for specialized .new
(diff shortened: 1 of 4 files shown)


headius:
@enebo @kares I have concerns about the locking.

kares:
looks reasonable (given no further details #5910 - still weird that there's only one pathological case).

would it make sense to reduce the lock duration to the bare minimum of (double-checking) and setting the generated class-and-allocator?, moving 'only' these parts under the critical section: https://github.com/jruby/jruby/blob/master/core/src/main/java/org/jruby/specialized/RubyObjectSpecializer.java#L152-L153

core/src/main/java/org/jruby/RubyObject.java
```diff
@@ -152,18 +153,28 @@ public IRubyObject allocate(Ruby runtime, RubyClass klass) {
...
-            klass.metaClass.invalidateCacheDescendants();
+                    // invalidate metaclass so new allocator is picked up for specialized .new
+                    klass.metaClass.invalidateCacheDescendants();
 
-            return allocator.allocate(runtime, klass);
+                    return allocator.allocate(runtime, klass);
+                } else {
+                    // someone else replaced allocator while we waited for lock, use it
+                    return allocator.allocate(runtime, klass);
+                }
```
rdubya: Maybe it would affect readability, but would it be possible to get rid of the `else` and always call `return allocator.allocate(runtime, klass);` at the bottom? Not sure if it would have any benefits over just not having duplicated code though (if it is even possible.)

enebo:
I have been pondering impact of this and I believe the ordinary classes will not be much of an impact because there are just not that many types in a typical application.   A few thousand locks when most happen early in an app does not seem like a big deal.

Singleton types perhaps in some DCI pattern?  Two threads singletonizing the same object at the same time already hurts my head.  We must already synch on that?   Trying to think of patterns where a framework would endlessly make new types.  Seems like the main worry to more locking here.

core/src/main/java/org/jruby/RubyObject.java
headius: Agreed.

headius:
> would it make sense to reduce the lock duration to the bare minimum of (double-checking) and setting the generated class-and-allocator?

The only thing that could move out would be the walking of all classes and methods in the type's hierarchy. Seems like that should be ok; the structures involved are all safe.

headius:
It turns out that the related Ruby code in #5910 was actually buggy to begin with; it allowed uninitialized classes to be constructed, which led to us seeing multiple views of instance variables in defined methods. I've detailed that here: https://github.com/jruby/jruby/issues/5910#issuecomment-543429744

We can't fix that, but it did expose some weaknesses in our variable reification that I have fixed in the most recent commit:

* We were forcing all methods to compile, breaking the startup-time promises of our lazy IRMethod compilation. That is now fixed; if the `defNode` is still present, we will walk it rather than forcing a compile. @enebo please review. We should investigate whether we're forcing methods to compile too early for other things.
* The walk of methods should produce the same result for the same unchanging method hierarchy. It did not only because the hierarchy was actually changing in #5910. Because the walk should produce the same result every time, I've moved it outside the synchronization block to reduce contention (but possibly waste some work). @kares please review.

PR #5924, opened 2019-10-18
inner classes should be references with dot #2639

### LINKED ISSUE #2639: support method access for nested Java classes
assuming Java integration already works in a Java-like manner : 
`java.lang.management.MemoryType.value_of('HEAP')`

shall support Java-like `java.awt.Desktop.Action.valueOf('OPEN')`
instead of "only" Ruby-like `java.awt.Desktop::Action.valueOf('OPEN')`

MariuszCwikla:
Fixes #2639

### DIFF
--- a/core/src/main/java/org/jruby/RubyBasicObject.java
+++ b/core/src/main/java/org/jruby/RubyBasicObject.java
@@ -1705,8 +1705,19 @@ public class RubyBasicObject implements Cloneable, IRubyObject, Serializable, Co
         if (args.length == 0 || !(args[0] instanceof RubySymbol)) {
             throw context.runtime.newArgumentError("no id given");
         }
+        
+        RubySymbol symbol = (RubySymbol)args[0];
+        
+        if (recv instanceof RubyClass) {
+            String probablyClassName = symbol.idString();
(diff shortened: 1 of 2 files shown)


kares:
Thanks Mariusz, pbly should have closed #2639 ... at least my opinions having this feature changed.
not sure we still want more magic with like these, since we should follow Ruby-like semantics (you can not access nested constants this way in vanilla Ruby) esp. as nesting rules in MRI have gotten stricter.

back to the PR - this is touching method_missing functionality in BasicObject to provide a Java feature. this will easily have side effects, so even if the feature is desired we should approach it differently.

headius:
Yeah I concur with @kares here...the method missing is a bit too invasive, especially since many Java integration use cases will want to monkey-patch `method_missing` to do other cool things and that will break this behavior.

I am still of the opinion that inner classes should be treated like constants, and referenced using namespace scoping (`Foo::Bar`) just like constants are. Using `.` should only be for methods, or minimally for Java fields exposed as methods.

headius:
Also, thanks @MariuszCwikla for giving this a shot! 👏

Issue #5923, opened 2019-10-18
Exception inspect in wrong encoding


jlahtinen:
### Environment

Seems to be problem since 1.7.27 (probably earlier also).
My console encoding:

    $ echo $LANG
    en_GB.utf8

### Expected Behavior

cruby gives excepted output for inspect:

    $ ruby -e 'e = Exception.new "å"; puts "#{RUBY_PLATFORM} - #{RUBY_VERSION} - #{e} - #{e.inspect}"'
    x86_64-linux - 2.5.3 - å - #<Exception: å>

### Actual Behavior

jruby behaviour:

    $ ruby -e 'e = Exception.new "å"; puts "#{RUBY_PLATFORM} - #{RUBY_VERSION} - #{e} - #{e.inspect}"'
    java - 2.5.3 - å - #<Exception: Ã¥>

Seems like "å" is read inside exception as  ISO-8859-1 and ouputted as encoded "UTF-8"

    $ ruby -e 'puts "å".force_encoding("ISO-8859-1").encode("UTF-8")'
    Ã¥

enebo:
Note: Our inspect code internally is building a String up using a StringBuilder and not our newer str() methods (which assumes all string data will end up being in Java's default charset).  We just need to change this code to use our newer APIs.

PR #5927, opened 2019-10-20
Use new RubyStringBuilder.str api to create inspected string for Exception

jlahtinen:
pull request fix for #5923

### DIFF
--- a/core/src/main/java/org/jruby/RubyException.java
+++ b/core/src/main/java/org/jruby/RubyException.java
@@ -303,8 +303,7 @@ public class RubyException extends RubyObject {
 
         if (exception.isEmpty()) return context.runtime.newString(rubyClass);
 
-        String exceptionStartStr = new StringBuilder(2 + rubyClass.length() + 2).append("#<").append(rubyClass).append(": ").toString();
-        String exceptionStr = RubyStringBuilder.str(context.runtime, exceptionStartStr, exception, ">");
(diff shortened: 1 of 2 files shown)


core/src/main/java/org/jruby/RubyException.java
```diff
@@ -302,10 +303,9 @@ public RubyString inspect(ThreadContext context) {
 
         if (exception.isEmpty()) return context.runtime.newString(rubyClass);
 
-        return RubyString.newString(context.runtime,
-                new StringBuilder(2 + rubyClass.length() + 2 + exception.size() + 1).
-                    append("#<").append(rubyClass).append(": ").append(exception.getByteList()).append('>')
-        );
+        String exceptionStartStr = new StringBuilder(2 + rubyClass.length() + 2).append("#<").append(rubyClass).append(": ").toString();
```
kares: you should nuke the `StringBuilder` here - it shouldn't be needed

jlahtinen: I think nuking is done now.

kares: the last round-trip (Ruby -> Java -> Ruby string) is redundant, but we can fix that later ...

enebo:
Super cool.  I may change the signature to be ruby,string,irubyobject,string,irubyobject,string but thanks for this work.  We really appreciate it!

Issue #5929, opened 2019-10-20
NPE json nested-array regression


ahorek:
### Environment

Provide at least:
jruby 9.2.9.0-SNAPSHOT (2.5.7) 2019-10-20 ebab3d4 Java HotSpot(TM) 64-Bit Server VM 11.0.1+13-LTS on 11.0.1+13-LTS +jit [mswin32-x86_64]

bisected to https://github.com/jruby/jruby/commit/b9bd393834f9a66fbd452199d8ca96806a217c60

cc @kares 

### Expected Behavior
```
require 'json'

class SubArray2 < Array
  def to_json(*a)
    {
      JSON.create_id => self.class.name,
      'ary'          => to_a,
    }.to_json(*a)
  end
end

JSON(SubArray2["foo"])
```
```
=> "{\"json_class\":\"SubArray2\",\"ary\":[\"foo\"]}"
```

### Actual Behavior

```
java.lang.NullPointerException: null
           eltInternal at org/jruby/RubyArray.java:905
              generate at json/ext/Generator.java:306
              generate at json/ext/Generator.java:272
                 visit at json/ext/Generator.java:377
                 visit at org/jruby/RubyHash.java:676
          visitLimited at org/jruby/RubyHash.java:698
              visitAll at org/jruby/RubyHash.java:2744
              generate at json/ext/Generator.java:356
              generate at json/ext/Generator.java:329
           generateNew at json/ext/Generator.java:194
          generateJson at json/ext/Generator.java:37
               to_json at json/ext/GeneratorMethods.java:70
                  call at json/ext/GeneratorMethods$RbHash$INVOKER$s$0$0$to_json.gen:-1
          cacheAndCall at org/jruby/runtime/callsite/CachingCallSite.java:343
                  call at org/jruby/runtime/callsite/CachingCallSite.java:76
  invokeOther5:to_json at testik.rb:6
               to_json at testik.rb:6
                  call at org/jruby/internal/runtime/methods/CompiledIRMethod.java:154
               finvoke at org/jruby/RubyClass.java:788
                invoke at org/jruby/runtime/Helpers.java:459
            callMethod at org/jruby/RubyBasicObject.java:373
           generateNew at json/ext/Generator.java:451
          generateJson at json/ext/Generator.java:59
              generate at json/ext/GeneratorState.java:209
                  call at json/ext/GeneratorState$INVOKER$i$1$0$generate.gen:-1
[...]

PR #5930, opened 2019-10-20
dupImpl specializations

ahorek:
fixes https://github.com/jruby/jruby/issues/5929


kares:
oh cool, you already fixed it ... thanks again!

PR #5939, opened 2019-10-23
Restructure indy-based math binding.

### LINKED ISSUE #5938: regressed indy fixnum op with custom equality (==)
tracking a regression from https://github.com/elastic/logstash/issues/11196#issuecomment-544985427

... which boiled down to be a separate JRuby 9.2.8.0 issue (with `-Xcompile.invokedynamic`)

the regression was introduced with https://github.com/jruby/jruby/pull/4736 and relates to `==` comparison, sample reproducer:

```ruby
class Primitive

  def _value
    @__value ||= 0
  end

  def ==(other)
    other == _value
  end

  # def <=>(other)
  #   _value <=> other
  # end

  # def eql?(other)
  #   # double dispatch
  #   other.eql?(snapshot)
  # end

end

class Klass; end

threads = []

3.times do |i|
  threads << Thread.start do

    begin

      header = Klass.new

      def header.version
        Primitive.new
      end

      header.version == 5
    rescue java.lang.Exception => ex
      puts("FAILED " + java.lang.Thread.currentThread.name + " - " + ex.toString)
      raise ex
    end

  end
end

threads.each &:join

p :done
```

```
Unhandled Java exception: java.lang.ClassCastException: org.jruby.gen.RubyObject1 cannot be cast to org.jruby.RubyFixnum
java.lang.ClassCastException: org.jruby.gen.RubyObject1 cannot be cast to org.jruby.RubyFixnum
            fixnum_op_equal at org/jruby/runtime/invokedynamic/MathLinker.java:234
             fixnumOperator at org/jruby/runtime/invokedynamic/MathLinker.java:168
  bug_bindata_reproducer.rb at bug_bindata_reproducer.rb:41
                 callDirect at org/jruby/runtime/CompiledIRBlockBody.java:136
                       call at org/jruby/runtime/IRBlockBody.java:77
                       call at org/jruby/runtime/Block.java:129
                       call at org/jruby/RubyProc.java:295
                       call at org/jruby/RubyProc.java:274
                       call at org/jruby/RubyProc.java:270
                        run at org/jruby/internal/runtime/RubyRunnable.java:105
                        run at java/lang/Thread.java:748
```

headius:
This primarily adds a hard RubyFixnum or RubyFloat type check on
self, to ensure we're not even attempting fast bindings for other
types. It also reduces overall binding overhead in that case by
moving direct indy logic out of the fail path and moves a couple
more adapted method handles to constants.

Fixes #5938.

### DIFF
--- a/core/src/main/java/org/jruby/runtime/invokedynamic/MathLinker.java
+++ b/core/src/main/java/org/jruby/runtime/invokedynamic/MathLinker.java
@@ -91,6 +91,8 @@ public class MathLinker {
     private static final CallType[] CALL_TYPES = CallType.values();
 
     private static final int[] ARG_2_TO_0 = new int[] { 2 }; // MethodHandles.permuteArguments clones the array
+    public static final MethodHandle FLOAT_TEST_ARG_2_TO_0 = permuteArguments(FLOAT_TEST, methodType(boolean.class, ThreadContext.class, IRubyObject.class, IRubyObject.class), ARG_2_TO_0);
+    public static final MethodHandle FIXNUM_TEST_ARG_2_TO_0 = permuteArguments(FIXNUM_TEST, methodType(boolean.class, ThreadContext.class, IRubyObject.class, IRubyObject.class), ARG_2_TO_0);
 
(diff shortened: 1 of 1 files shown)


kares:
:+1: this should resolve the issue.
got confused a bit since I expected a check on the indy path (before) reaching math-linker ...
as for non-indy `self instanceof RubyFixnum/RubyFloat` is done before reaching this logic.

PR #5940, opened 2019-10-23
Improve how we acquire ThreadContext to eliminate null refs.

headius:
For #5936 we saw some cases where the SoftReference created here
came back as null, even after the "adopt" logic had run. While I
have not isolated an exact cause, there are thread locals, soft
references, and GC interacting here. Since the original logic did
not guarantee non-null references throughout the code, I've
modified it to always check and only proceed once it truly has
non-null references and eventually context in hand.

### DIFF
--- a/core/src/main/java/org/jruby/internal/runtime/ThreadService.java
+++ b/core/src/main/java/org/jruby/internal/runtime/ThreadService.java
@@ -197,10 +197,10 @@ public class ThreadService {
      * collected.
      */
     public final ThreadContext getCurrentContext() {
-        ThreadContext context = null;
+        ThreadContext context;
 
         // keep trying until we have a context
-        while (context == null) {
+        do {
             SoftReference<ThreadContext> ref = localContext.get();
             if (ref == null) {
                 context = adoptCurrentThread().getContext(); // registerNewThread will localContext.set(...)
@@ -210,7 +210,7 @@ public class ThreadService {
                     localContext.set(null);
                 }
             }
-        }
+        } while (context == null);
 
         return context;
     }
--- a/core/src/main/java/org/jruby/internal/runtime/ThreadService.java
+++ b/core/src/main/java/org/jruby/internal/runtime/ThreadService.java
@@ -197,21 +197,22 @@ public class ThreadService {
      * collected.
      */
     public final ThreadContext getCurrentContext() {
-        SoftReference<ThreadContext> ref = localContext.get();
-        if (ref == null) {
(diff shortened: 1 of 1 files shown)


kares:
previous version ended up recursing instead of always reaching a loop - otherwise seems the same?
not sure what changed here - we never ended up with `null` on production apps for a long time.
really lack an understanding what precisely is fixed here ...

headius:
@kares In the later comments of #5936 both @ahorek and I saw NPE when attempting to get the ThreadContext out of the SoftReference we acquired from the ThreadLocal. Somehow, that SoftReference was coming out as null even after calling `adoptCurrentThread`. I could not suss out how that would be possible, exactly, but there's a complicated dance of references and threads and GC in play.

The new logic basically just adds in extra null checking, in case the stars align to allow such to happen, and avoids recursing to re-attempt the acquisition of the context. It eliminates the NPE we saw.

headius:
@enebo Yeah I couldn't trace a path that would result in getting a null SoftReference after adoption, but I could not exactly rule it out.

There's two situations where that ThreadLocal might contain null:

* The current thread has never been adopted
* The current thread was adopted but the reference went away (soft ref, memory pressure trigger)

In both cases the current thread should go into the logic that assigns a new SoftReference before proceeding.

headius:
@kares I will make that change, thank you.

I want to understand if there was a reason you preferred recursion...something I'm missing in my version?

My only concern with the recursion was that under some heavy GC scenario it may recurse many times, failing to acquire a context. It is not likely to happen or to cause stack issues, since we're talking soft references instead of weak references, but it nagged at the back of my brain a bit. The change was not altogether related to this patch, but it made sense to me.

I don't like the overall complexity of the adoption logic, so I want to be absolutely sure I understand your version of this method.

headius:
I've made the change and rebased to pick up fixes from master.

kares:
> I want to understand if there was a reason you preferred recursion...something I'm missing in my version?

think I did a micro-benchmark where it showed up that always entering a loop was 'slower'.
was looking for that benchmark but did not find (not sure what I actually did - whether trying to stress this from a known getCurrentContext .rb site or simply a JMH). 

anyway, its more important to have this working (even in edge cases) rather than 'optimized' to the ms which won't matter much in the end.

headius:
@kares That seems plausible, but I *think* it shouldn't make much difference these days. It boils down to a branch versus a recursive invocation, and in both cases the rare path that fails to acquire a context should end up being profiled out. If you do see a difference, though, we can revisit it.

PR #5941, opened 2019-10-23
WIP: travis hang

ahorek:
don't merge, let's see if it hangs on travis or not...


ahorek:
ok, skiping that suspicious test didn't help

Issue #5875, opened 2019-09-13
Mutex behavior mismatch with MRI since JRuby 9.2.8.0


walro:
### Environment

```
$ jruby -v
jruby 9.2.8.0 (2.5.3) 2019-08-12 a1ac7ff Java HotSpot(TM) 64-Bit Server VM 9.0.1+11 on 9.0.1+11 +jit [darwin-x86_64]
```

### Expected Behavior

The following executes "forever" without errors on MRI 2.5.3 and JRuby 9.2.7.0:

```ruby
mutex = Mutex.new

t1 = Thread.new do
  loop do
    mutex.synchronize do
      mutex.sleep 1 # not needed to provoke the error, but just to motivate the t1.run below
    end
  end
end

loop do
  mutex.synchronize do
  end

  t1.run
end
```

This is a real-world problem cooked down to its core. The real-world app (a busy web-app) uses https://github.com/sharpstone/rack-timeout, where one can find similar [mutex usage](https://github.com/sharpstone/rack-timeout/blob/fa4af253370611489356e6badc9b5a02a751661c/lib/rack/timeout/support/scheduler.rb#L58-L130).

### Actual Behavior

```
$ jruby mutex.rb 
warning: thread "Ruby-0-Thread-1: mutex.rb:1" terminated with exception (report_on_exception is true):
ConcurrencyError: interrupted waiting for mutex
         lock at org/jruby/ext/thread/Mutex.java:101
  synchronize at org/jruby/ext/thread/Mutex.java:158
     mutex.rb at mutex.rb:5
         loop at org/jruby/RubyKernel.java:1425
     mutex.rb at mutex.rb:4
ThreadError: killed thread
    wakeup at org/jruby/RubyThread.java:1279
       run at org/jruby/RubyThread.java:1437
  mutex.rb at mutex.rb:15
      loop at org/jruby/RubyKernel.java:1425
    <main> at mutex.rb:11
```

After some digging around I would assume this is connected to the change in #5683. What I find a bit curious is that the main thread must also try to synchronize to force the error to happen.

headius:
I'm on it! The problem here is that for whatever reason I implemented mutex sleeping to sleep based on some arbitrary lock, rather than on the lock associated with the mutex.

mrmargolis:
[...]

PR #5942, opened 2019-10-23
Mutex fixes for interruptibility

headius:
This PR contains fixes for some interruptibility issues introduced in JRuby 9.2.8:

* A thread woken up while attempting to acquire a contended lock would raise an error rather than continuing to try to acquire the lock.
* A thread sleeping via a `ConditionVariable` would be uninterruptible.

WIP until I confirm it's ready.

### DIFF
--- a/core/src/main/java/org/jruby/RubyThread.java
+++ b/core/src/main/java/org/jruby/RubyThread.java
@@ -50,6 +50,7 @@ import java.util.Set;
 import java.util.concurrent.*;
 import java.util.concurrent.atomic.AtomicIntegerFieldUpdater;
 import java.util.concurrent.atomic.AtomicReference;
+import java.util.concurrent.locks.Condition;
 import java.util.concurrent.locks.Lock;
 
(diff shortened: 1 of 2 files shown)


headius:
Fix for #5875 is in.

headius:
Fix for #5863 is in.

The second fix introduces a new RubyThread status called "NATIVE", which indicates that the native JDK thread status should be translated into a Ruby thread status. This allows us to use the provided JDK Condition.await, which does the appropriate lock release followed by going into a wait state (via LockSupport.park). This in turn allows the specs to pass that depend on the lock being released by the time the thread status is "sleep".

This fix is also a stepping stone toward eliminating the synthetic Ruby thread statuses. As we go forward we can modify more of these state changes to just defer to the NATIVE status, leaning on the JDK to handle state changes without racing.

headius:
Assuming that specs pass, I consider this one ready to go. Probably should create some specs for the cases in #5942  and #5863.

PR #5943, opened 2019-10-23
properly synchronize globalInlineCount

ahorek:
synchronization using a boxed primitive isn't reliable, because the field's value could be changed while a thread is in a synchronized block.


kares:
nice find, this piece is just quite weird (there's still `globalInlineCount++` against the boxed int).
@ahorek how do you feel about refactoring to simply use an `AtomicInt` (without synchronize) ?

enebo:
with PR as it is globalInlineCount should be made an int.  The usage of this synchronize was so narrow I don't think in practice we would ever see the field change while it is synched.  That said, either suggested fix is better (I would prefer AtomicInt myself).

Issue #5947, opened 2019-10-28
Update FFI scripts and other files from FFI gem


headius:
Ok, it's time we got JRuby's FFI and the FFI gem on the same page, so we can share updates.

Currently a large number of files we have in common with the FFI gem are not up-to-date. There are many more type configs in the gem and I'm sure there's fixes and tweaks on the JRuby side that have not been propagated to the gem.

I'll start a PR and we can get this stuff synced up. Once we match, I'd like to get the ffi "java" gem populated with the shared files so we can preinstall it and allow users to update it between JRuby releases (which is not possible right now).

cc @larskanis @ahorek @olleolleolle @eregon

larskanis:
Hi @headius ! It's been some months since [this comment](https://github.com/ffi/ffi/issues/660#issuecomment-456558171), but to that time I migrated jruby's ffi java files to the ffi gem as a jruby extension [here](https://github.com/larskanis/ffi/commits/join-jruby). It already worked pretty well, but a bunch of tests failed. I didn't push this attempt forward, since I wasn't sure, whether moving the ffi code out of the jruby repository into the ffi gem repository is the direction you would like to go.

This way we could make sure, that every addition to the ffi gem also pass on JRuby. But I'm not sure, how backward compatible the JRuby files are, since the ffi gem should be usable on various JRuby versions.

eregon:
Agreed it would be good to share more of the FFI code between implementations: https://github.com/ffi/ffi/issues/660

Currently TruffleRuby has a copy of FFI Ruby files (= frontend), and reimplements the C extension methods with Ruby + Truffle NFI code (= backend).
We have a script to update the frontend files as well as specs easily in TruffleRuby:
https://github.com/oracle/truffleruby/blob/853c3ab0194ed04cf4b3fb7b528bf360e66407fe/tool/import-ffi.sh

TruffleRuby passes almost all specs of FFI, there are just a few exclusions.
It would be a good idea to add TruffleRuby in FFI's CI and just exclude these not-yet-working specs.

I'm not sure what's best for the future.
Maybe we should have the TruffleRuby backend in the FFI gem?
Or just use the frontend Ruby code from the FFI gem, but use the backend in TruffleRuby?
Or just keep the current situation for TruffleRuby since it's already very easy to synchronize?

If we move the TruffleRuby backend in the FFI gem, I would want push access to ffi/ffi so I can maintain and fix the TruffleRuby backend conveniently.
[...]

PR #5948, opened 2019-10-28
Update shared FFI code

headius:
This PR will track the process of getting files we have in common with the [ffi gem](https://github.com/ffi/ffi) updated so we can share updates and maintenance effort.

See #5947


core/src/main/java/org/jruby/ext/ffi/Platform.java
```diff
@@ -317,49 +317,12 @@ public static void createPlatformModule(Ruby runtime, RubyModule ffi) {
         OS_TYPE os = platform.getOS();
         module.defineConstant("ADDRESS_SIZE", runtime.newFixnum(platform.addressSize));
         module.defineConstant("LONG_SIZE", runtime.newFixnum(platform.longSize));
-        module.defineConstant("OS", runtime.newString(OS.toString()));
-        module.defineConstant("ARCH", runtime.newString(platform.getCPU().toString()));
-        module.defineConstant("NAME", runtime.newString(platform.getName()));
-        module.defineConstant("IS_WINDOWS", runtime.newBoolean(os == OS.WINDOWS));
-        module.defineConstant("IS_BSD", runtime.newBoolean(platform.isBSD()));
-        module.defineConstant("IS_FREEBSD", runtime.newBoolean(os == OS.FREEBSD));
         module.defineConstant("IS_DRAGONFLYBSD", runtime.newBoolean(os == OS.DRAGONFLYBSD));
```
ahorek: https://github.com/ffi/ffi/pull/724

lib/ruby/stdlib/ffi/platform/sysconf.rb.ffi
```diff
@@ -1,139 +0,0 @@
-module Platform;end
```
ahorek: jruby specific?

headius: I think these are template files that originally lived outside of the generation logic but now live in some form in the gen/ dir?

core/src/main/java/org/jruby/ext/ffi/Platform.java
headius: I nuked all but DragonFly from JRuby's `FFI::Platform` java code. Can nuke DragonFlyBSD once I see it in the ruby code.

headius:
I have had to disable the caching of the Struct layout as part of this PR, but I believe this is due to a misfeature in the C FFI. I have filed an issue here: https://github.com/ffi/ffi/issues/734

headius:
> I have had to disable the caching of the Struct layout

ffi/ffi#734 is fixed now (deprecated behavior to be removed in 2.0) so I reverted the removal of the caching logic.

headius:
Retargeted to 9.3. Help wanted!

core/src/main/java/org/jruby/ext/ffi/Platform.java
larskanis: @headius `IS_DRAGONFLYBSD` is part of ffi's ruby code now.

ahorek:
the remaining failure is
```
  Struct tests :enum field r/w
     Failure/Error: @orig_field.put(ptr, type.to_native(value, nil))

     NoMethodError:
       undefined method `put' for #<FFI::StructLayout::Number:0x71a509c>
       Did you mean?  puts
                      putc
```

this change https://github.com/jruby/jruby/pull/5948/commits/221b11d3ff9d247d4ca2646be2a68be543d7185a is in conflict with
https://github.com/ffi/ffi/blob/21dc3d48f672ecf334058b36f485292721de8a87/lib/ffi/struct_layout.rb#L92

should we fork the ruby version?

headius:
> should we fork the ruby version?

Ideally, no, we should match the Ruby code exactly.

I'm not sure what you mean by a conflicts. It seems that Field simply doesn't define `put` and `get`. I'm adding those now.

ahorek:
the previous java version used to delegate the put method to MappedFieldIO
https://github.com/jruby/jruby/blob/221b11d3ff9d247d4ca2646be2a68be543d7185a/core/src/main/java/org/jruby/ext/ffi/StructLayout.java#L1417

your fix looks good https://github.com/jruby/jruby/pull/5948/commits/f40efb1270a867278bf3054149ad1d815543db0a thanks

headius:
With these last changes from @ahorek and the missing Field `put` and `get` in place, it appears we are green!

headius:
> the previous java version used to delegate the put method to MappedFieldIO

Ahh yes, but it was never bound in Ruby. All good now!

Issue #5937, opened 2019-10-22
port CGI.escapeHTML optimization


ahorek:
one of the recent ruby 2.7 optimization is https://github.com/ruby/ruby/pull/2226

jruby is a bit behind
```
ruby 2.7 dev
      CGI.escapeHTML      2.614k (± 9.3%) i/s -     13.108k in   5.065551s

ruby 2.5.3
      CGI.escapeHTML    761.177  (± 6.7%) i/s -      3.825k in   5.050563s

jruby 9.2.9.0 dev
      CGI.escapeHTML    224.765  (± 6.7%) i/s -      1.122k in   5.013482s
```

the previous java port is at https://github.com/jruby/jruby/blob/4e8bb2666a7257f0f5986800f96bb88efdd6acbd/core/src/main/java/org/jruby/ext/cgi/escape/CGIEscape.java#L64

porting the new algorithm should improve numbers

the optimization is especially important for rails, erb engine does use it for html escaping:
https://api.rubyonrails.org/classes/ERB/Util.html#method-c-h

headius:
As long as there's no functionality change, I think we can pull this into 9.2.9 (still 2.6 compat, etc).

headius:
Oops, for some reason I thought this was a PR. We'll merge it whenever someone does it.

ahorek:
sry, no PR yet, but I can work on it.

just one question, the optimized java version doesn't seem to be used now
https://github.com/jruby/jruby/blob/4e8bb2666a7257f0f5986800f96bb88efdd6acbd/core/src/main/java/org/jruby/ext/cgi/escape/CGIEscape.java#L352

was it removed intentionally in https://github.com/jruby/jruby/issues/4531 ?

headius:
There's still issues with supering up a complicated include+prepend+extend hierarchy, so we'll aim to get that and this wrapped up in 9.2.10.

enebo:
@ahorek https://github.com/jruby/jruby/pull/6777  

Can you try this port on this branch?  I believe it will work now.  If so we can land this for 9.3 instead of 9.4.

enebo:
Fixed prepend issue in PR #6777 and impld in #6778

Issue #4678, opened 2017-06-19
Prepending a module to a module extended by a class gives NPE when calling `super`


donv:
jruby 9.1.12.0 (2.3.3) 2017-06-15 33c6439 Java HotSpot(TM) 64-Bit Server VM 25.112-b16 on 1.8.0_112-b16 +jit [darwin-x86_64]
Darwin Gratass.local 16.6.0 Darwin Kernel Version 16.6.0: Fri Apr 14 16:21:16 PDT 2017; root:xnu-3789.60.24~6/RELEASE_X86_64 x86_64

Given the following code:

```ruby
module A
  def self.included(clazz)
    clazz.extend ClassMethods
  end

  module ClassMethods
    def foo
      puts 'super'
    end
  end
end

class B
  include A
end

module C
  def foo
    super
  end
end
A::ClassMethods.prepend C

B.foo
```

JRuby 9.1.12.0 gives the following stack trace:

```
Unhandled Java exception: java.lang.NullPointerException
java.lang.NullPointerException: null
                      unresolvedSuper at org/jruby/ir/runtime/IRRuntimeHelpers.java:1014
  invokeSuper6:-unknown-super-target- at jruby_example.rb:19
                                  foo at jruby_example.rb:19
                                 call at org/jruby/internal/runtime/methods/CompiledIRMethod.java:155
                         cacheAndCall at org/jruby/runtime/callsite/CachingCallSite.java:318
                                 call at org/jruby/runtime/callsite/CachingCallSite.java:131
                    invokeOther10:foo at jruby_example.rb:24
                               <main> at jruby_example.rb:24
                  invokeWithArguments at java/lang/invoke/MethodHandle.java:627
                                 load at org/jruby/ir/Compiler.java:95
                            runScript at org/jruby/Ruby.java:827
                          runNormally at org/jruby/Ruby.java:746
                          runNormally at org/jruby/Ruby.java:764
                          runFromMain at org/jruby/Ruby.java:577
                        doRunFromMain at org/jruby/Main.java:417
                          internalRun at org/jruby/Main.java:305
                                  run at org/jruby/Main.java:232
                                 main at org/jruby/Main.java:204
```

With MRI 2.4.1 it prints "super".

donv:
Workaround is to postpone the `include` call to after the `prepend`call.

Possible duplicate of #4531.

headius:
This no longer NPEs but still fails:

```
$ jruby blah.rb
[...]

PR #5949, opened 2019-10-28
use optimized CGIEscape

ahorek:
as commented in https://github.com/jruby/jruby/issues/5937

`CGI.escapeHTML` does always use a pure ruby version which is slower

https://github.com/ruby/ruby/blob/master/ext/cgi/escape/escape.c#L408

```
ruby
CGI.escapeHTML    242.472  (? 7.0%) i/s -      1.220k in   5.061783s
optimized
CGI.escapeHTML    804.555  (? 5.6%) i/s -      4.032k in   5.027866s
```

### DIFF
--- a/core/src/main/java/org/jruby/ext/cgi/escape/CGIEscape.java
+++ b/core/src/main/java/org/jruby/ext/cgi/escape/CGIEscape.java
@@ -426,10 +426,10 @@ public class CGIEscape implements Library {
     public void load(Ruby runtime, boolean wrap) {
         RubyClass rb_cCGI = runtime.defineClass("CGI", runtime.getObject(), runtime.getObject().getAllocator());
         RubyModule rb_mEscape = rb_cCGI.defineModuleUnder("Escape");
+        RubyModule rb_mUtil   = rb_cCGI.defineModuleUnder("Util");
         rb_mEscape.defineAnnotatedMethods(CGIEscape.class);
-        // We do this in cgi/util.rb to work around jruby/jruby#4531.
-//        rb_mUtil.prependModule(rb_mEscape);
-//        rb_mEscape.extend_object(rb_cCGI);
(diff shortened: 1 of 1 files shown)


ahorek:
If I understand this correctly, the original issue was fixed in https://github.com/jruby/jruby/pull/5627 and the CGI tweak was partly reverted in https://github.com/jruby/jruby/commit/3d8847e0dc656a0b2b072d0bd642128202ae66d3 . Is it safe to use the CGIEscape.java library now? I didn't find any issues and the CI is green. ping @headius ?

headius:
Looking into it. I want to understand what it was disabled.

headius:
I attempted to uncomment both of those lines and that results in errors like this during test/mri/cgi tests:

```
CGIUtilTest#test_cgi_unescapeHTML:UTF-32LE:
NoMethodError: super: no superclass method `unescapeHTML' for CGI:Class
Did you mean?  unescape_html
               escape_html
               unescapeHTML
               escapeHTML
    org/jruby/RubyBasicObject.java:1709:in `method_missing'
    org/jruby/ext/cgi/escape/CGIEscape.java:377:in `unescapeHTML'
    org/jruby/ext/cgi/escape/CGIEscape.java:377:in `unescapeHTML'
    /Users/headius/projects/jruby/test/mri/cgi/test_cgi_util.rb:135:in `block in CGIUtilTest'
```

And the issue in #4678 still causes a similar error attempting to call a super `foo`. Will poke at this a bit but it may be a deeper problem than we can fix for 9.2.9 on short notice.

ahorek:
ok, if it's too risky, feel free to postpone it

headius:
I realized a simple way we can integrate this: implement it as a separate ext that does not have to be prepended and supered in the ways MRI does. Looking into this quickly as an option.

Issue #5905, opened 2019-10-03
String#gsub given a String can avoid Regexp


headius:
It appears in Ruby 2.2 a change by @nobu modified the logic to just call `String#index` when passing a String to `String#gsub`, rather than always using a Regexp object and paying the cost to compile it every time.

We still do the old behavior, making this form of `gsub` slower than it could be.

```
rvm truffleruby do ruby bench_gsub.rb 
Warming up --------------------------------------
         gsub-string    24.074k i/100ms
          gsub-regex    79.710k i/100ms
    gsub-regex-block    65.247k i/100ms
Calculating -------------------------------------
         gsub-string    354.958k (±34.9%) i/s -      1.420M in   5.036696s
          gsub-regex      2.097M (±34.5%) i/s -      8.290M in   5.010322s
    gsub-regex-block      2.007M (±27.7%) i/s -      8.417M in   5.004553s

$ rvm ruby-2.6.4 do ruby bench_gsub.rb 
Warming up --------------------------------------
         gsub-string    88.943k i/100ms
          gsub-regex    51.561k i/100ms
    gsub-regex-block    49.026k i/100ms
Calculating -------------------------------------
         gsub-string      1.115M (± 5.2%) i/s -      5.603M in   5.041119s
          gsub-regex    596.028k (± 6.3%) i/s -      2.991M in   5.041490s
    gsub-regex-block    546.740k (± 3.7%) i/s -      2.745M in   5.028991s

$ jruby -Xcompile.invokedynamic bench_gsub.rb 
Warming up --------------------------------------
         gsub-string    92.520k i/100ms
          gsub-regex   142.823k i/100ms
    gsub-regex-block   109.341k i/100ms
Calculating -------------------------------------
         gsub-string      2.188M (± 8.8%) i/s -     10.825M in   5.001965s
          gsub-regex      3.631M (± 9.5%) i/s -     17.996M in   5.024391s
    gsub-regex-block      2.315M (± 7.6%) i/s -     11.481M in   4.999877s
```

The change in question is here: https://github.com/ruby/ruby/commit/5752b61d8649f0b4128721753dc1b38669efdb9c

Probably not a difficult one to duplicate in JRuby, for anyone who wants to do some spelunking.

rdubya:
Can you you throw your bench_gsub.rb file in here or is it in the codebase somewhere?

headius:
Oops, that's a good idea.

fidothe:
[...]

PR #5952, opened 2019-10-30
Use string index in String #sub/#gsub when String pattern passed without creating a Regexp

fidothe:
This PR addresses #5905. It's still pretty naive, and I'm really not a Java programmer, so there's probably a few dumb mistakes and style things aside from the big MatchData problem.

The big known problem remaining is how to populate the `$~` `MatchData`. In MRI the `MatchData` from a string-only `String#sub` or `#gsub` will return a `Regexp` from `MatchData#regexp`. I don't think we currently have a way of doing that, so that means changes to `MatchData`.

MRI is clearly doing something odd here, because if you call `#inspect` on the `MatchData` returned from a string pattern `#sub`, it's in a different form to what you get from invoking it with a Regexp pattern (this from MRI 2.6.3):

```
~ $ irb
irb(main):001:0> "a".sub("a", "b")
=> "b"
irb(main):002:0> $~
=> #<MatchData: a>
irb(main):003:0> "a".sub(/a/, "b")
=> "b"
irb(main):004:0> $~
=> #<MatchData "a">
```

Once I know what the best approach to handling that is, I can fill the hole in the `#sub` implementation and `#gsub`'s returned `MatchData`s  will behave as expected.

### DIFF
--- a/bench/core/string/bench_sub.rb
+++ b/bench/core/string/bench_sub.rb
@@ -1,0 +1,18 @@
+str1 = 'white chocolate'
+str2 = 'a1'
+str3 = 'dog'
+
+regex2 = /\d/
+regex3 = /\w+/
+
+benchmark 'sub-string' do
+  r1 = str1.sub('white', 'dark')
+end
+
+benchmark "sub-regex" do
+  r2 = str2.sub(regex2, '2')
+end
+
+benchmark "sub-regex-block" do
+  r3 = str3.sub(regex3) { |animal| animal == 'dog' ? 'cat' : 'dog' }
+end
--- a/core/src/main/java/org/jruby/RubyString.java
+++ b/core/src/main/java/org/jruby/RubyString.java
@@ -2884,7 +2884,7 @@ public class RubyString extends RubyObject implements CharSequence, EncodingCapa
         Ruby runtime = context.runtime;
         frozenCheck();
 
-        if (block.isGiven()) return subBangIter(context, asRegexpArg(runtime, arg0), null, block);
+        if (block.isGiven()) return subBangIter(context, arg0, null, block);
         throw runtime.newArgumentError(1, 2);
     }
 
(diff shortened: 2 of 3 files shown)


fidothe:
Question for @enebo / @headius. We're failing 4 Rubyspec specs for `#sub` and 4 for `#gsub`. (They're failing for me on master too, so it looks like no new problems there.) Are those failures expected?

fidothe:
I can provide a `gsubFast` method overload that takes a `RubyString` and not a `RubyRegexp`. Is that worth doing for symmetry, will anyone use it? (It's only user is in `RubyDate`, where it is very definitely using a `Regexp`.)

fidothe:
(Also, I'm going out on a limb and assuming that we don't want to replicate the inconsistency in output from MRI's `MatchData#inspect`)

fidothe:
Squashed the code changes down, left the new benchmark as its own commit to make before/after running easier for others.

Benchmark results before/after from my machine https://gist.github.com/fidothe/7d76f60cf2adbb8a2aab8adf21f40192

fidothe:
As I said before, this is the first Java code I've written, so I imagine it's full of style problems. (I found redundant casting and stuff, but assume there's more stuff like that I didn't recognise.) Let me know if there's stuff you want changed (and why, so I learn...) and I will happily oblige.

enebo:
@fidothe Awesome job!  Getting some more perf out of a common method hopefully is a pretty satisfying start in hacking JRuby.  

As far as style I made a few very minor changes: 457d12b13d.  Unfortunately, RubyString is not so consistent from a single-style perspective.  Some people more heavily favor marking anything which will not change as final whereas others (myself included) feel final is more useful as a marker when that finalness important or aids in helping signal intent.  We have some other oddities in this file like following MRI naming vs using more obvious full word variable names.

There was 3 places with the else was not joined with the previous { in that commit above.    For conditionals we use {} always and they are on same line as if/else.  For single line ifs they require {} if you want if and body on different lines or no {} if the body is on the same line as the if.

kares:
Matt, your Java is very nice. Would certainly assumed you're an experienced programmer based on this!

fidothe:
> Matt, your Java is very nice. Would certainly assumed you're an experienced programmer based on this!

Aw, thanks 😊 @kares

PR #5953, opened 2019-11-02
Fix Array.pack with @ directive and buffer #4727

### LINKED ISSUE #4727: Array#pack buffer should start from beginning of buffer?
This is confusing, but we're definitely not passing it. This spec was added recently with the addition of a "buffer" kwarg for Array#pack.

It appears from the result that we are supports to take the "@4" as being from the beginning of the given buffer, leaving the existing four bytes in place and writing the new four non-zero bytes at offset 4. Instead, we append the nulls and the new values.

From test/mri/ruby/test_pack.rb in test_pack_with_buffer:
```ruby
    buf = String.new(capacity: 100)
    #...
    [0xDEAD_BEEF].pack('N', buffer: buf)
    assert_equal "\xDE\xAD\xBE\xEF", buf

    [0xBABE_F00D].pack('@4N', buffer: buf)
    assert_equal "\xDE\xAD\xBE\xEF\xBA\xBE\xF0\x0D", buf
```

This is new logic but there's more work needed than I want to put in right now, and this is an unexpected behavior of the buffer.

See #4687 and #4293.

MariuszCwikla:
Fixes #4727


headius:
Did not notice this fix came in, thank you!

PR #5957, opened 2019-11-05
Add async-profiler smarts to --sample flag

headius:
This is a work in progress. It uses the upcoming jruby-async-profiler gem to implement `--sample`, which originally used the `-Xprof` flag removed as of Java 10.


headius:
https://github.com/jruby/jruby-async-profiler

PR #5959, opened 2019-11-05
Tweak getCurrentContext for SPEED

headius:
This change reduces the overhead of calling Ruby.getCurrentContext
since we still call it in a number of places.

* ThreadService now extends ThreadLocal rather than aggregating a ThreadLocal in a field. This eliminates one hop.
* All hot-path methods in ThreadService are static.
* Restore @kares recursion logic since it does appear faster than an explicit null check loop.

On OpenJDK 8 C2 this reduces single-thread getCurrentContext time from around 4ns to around 3.2ns. Other VMs have similar gains.

Tested with a trivial benchmark:

```java
import org.jruby.Ruby;

public class ContextGetter {
  public static void main(String[] args) {
    Ruby runtime = Ruby.newInstance();
    while (true) {
      long nanos = System.nanoTime();
      for (int i = 0; i < 100_000_000; i++) {
        runtime.getCurrentContext();
      }
      System.out.println((System.nanoTime() - nanos) / 100_000_000.0);
    }
  }
}
```

### DIFF
--- a/core/src/main/java/org/jruby/Ruby.java
+++ b/core/src/main/java/org/jruby/Ruby.java
@@ -2762,7 +2762,7 @@ public final class Ruby implements Constantizable {
     }
 
     public ThreadContext getCurrentContext() {
-        return threadService.getCurrentContext();
+        return ThreadService.getCurrentContext(threadService);
     }
 
     /**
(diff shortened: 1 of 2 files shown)


headius:
Assembly output from this benchmark and this branch on OpenJDK8 C2 is here: https://gist.github.com/headius/ac6ae20bcaa33c9795c3df4decb88dfd

Note that the hot path appears to have inlined completely down to line 167, returning the non-null result at that point. This is a Very Good Thing.

Issue #5956, opened 2019-11-05
dlload throws LoadError instead of Fiddle::DLError


ankane:
### Environment

```
jruby 9.2.8.0 (2.5.3) 2019-08-12 a1ac7ff Java HotSpot(TM) 64-Bit Server VM 25.192-b12 on 1.8.0_192-b12 +jit [darwin-x86_64]
```

### Expected Behavior

```ruby
require "fiddle/import"

module Hello
  extend Fiddle::Importer
  dlload "hi"
end
```

raises `Fiddle::DLError` like MRI

### Actual Behavior

Raises `LoadError`

Issue #5955, opened 2019-11-05
NameError: uninitialized constant Fiddle::CParser::TYPE_SIZE_T


ankane:
Hi, I'm running  into an error with `fiddle` with the latest JRuby on MacOS 10.14.6 (may affect other platforms as well).

### Environment

- jruby 9.2.8.0 (2.5.3) 2019-08-12 a1ac7ff Java HotSpot(TM) 64-Bit Server VM 25.192-b12 on 1.8.0_192-b12 +jit [darwin-x86_64]

### Expected Behavior

No error when `size_t` is used. Here's a simple script to reproduce:

```ruby
require "fiddle/import"

module Hello
  extend Fiddle::Importer
  dlload "libc.dylib" # change extension on other platforms
  extern "size_t puts(char* str)"
end
```

(this is not the actual interface for `puts` but works for demonstration)

### Actual Behavior

Raises

```txt
NameError: uninitialized constant Fiddle::CParser::TYPE_SIZE_T
Did you mean?  Fiddle::TYPE_INT
    const_missing at org/jruby/RubyModule.java:3746
      parse_ctype at /Users/andrew/.rbenv/versions/jruby-9.2.8.0/lib/ruby/stdlib/fiddle/cparser.rb:164
  parse_signature at /Users/andrew/.rbenv/versions/jruby-9.2.8.0/lib/ruby/stdlib/fiddle/cparser.rb:92
           extern at /Users/andrew/.rbenv/versions/jruby-9.2.8.0/lib/ruby/stdlib/fiddle/import.rb:166
    <class:Hello> at repro.rb:6
           <main> at repro.rb:3
```

ahorek:
hi @ankane !
do you have a real world example that you would like to support on jruby?

ankane:
Hey @ahorek, thanks for the response! I've created a `fiddle` branch in a number of projects to see if it was possible to switch from FFI. Here are a few that are affected by this issue.

- [Vowpal Wabbit](https://github.com/ankane/vowpalwabbit/tree/fiddle)
- [NGT](https://github.com/ankane/ngt/tree/fiddle)

And a few that are affected by #5956.

- [LightGBM](https://github.com/ankane/lightgbm/tree/fiddle)
- [xLearn](https://github.com/ankane/xlearn/tree/fiddle)

ankane:
Also, seeing errors with `malloc`.

```ruby
require "fiddle"

Fiddle::Pointer.malloc(100)
```

Throws

```txt
[...]

PR #5960, opened 2019-11-05
fiddle fixes

ahorek:
fixes https://github.com/jruby/jruby/issues/5956 and https://github.com/jruby/jruby/issues/5955

but this test case doesn't work yet
https://github.com/jruby/jruby/blob/4e8bb2666a7257f0f5986800f96bb88efdd6acbd/test/mri/fiddle/test_func.rb#L73
I'm experimanting for now :)

### DIFF
--- a/lib/ruby/stdlib/fiddle/jruby.rb
+++ b/lib/ruby/stdlib/fiddle/jruby.rb
@@ -5,15 +5,38 @@
 require 'ffi'
 
 module Fiddle
+  def self.malloc(size)
+    Fiddle::Pointer.malloc(size)
+  end
+
+  def self.free(ptr)
+    Fiddle::Pointer.__freefunc__(ptr)
+    nil
+  end
+
+  def self.dlwrap(val)
+    Pointer.to_ptr(val)
+  end
+
   TYPE_VOID         = 0
   TYPE_VOIDP        = 1
   TYPE_CHAR         = 2
+  TYPE_UCHAR        = -2
   TYPE_SHORT        = 3
+  TYPE_USHORT       = -3
   TYPE_INT          = 4
+  TYPE_UINT         = -4
   TYPE_LONG         = 5
+  TYPE_ULONG        = -5
   TYPE_LONG_LONG    = 6
+  TYPE_ULONG_LONG   = -6
   TYPE_FLOAT        = 7
   TYPE_DOUBLE       = 8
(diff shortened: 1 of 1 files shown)


headius:
Nice! We definitely want to get Fiddle working better. Long term it might make more sense to do a direct implementation atop jnr-ffi, but for now getting the FFI version working is probably best.

ahorek:
I splited two obvious errors to separate PRs
https://github.com/jruby/ruby/pull/2 https://github.com/jruby/ruby/pull/3

ahorek:
one more interesting failure
https://github.com/jruby/jruby/blob/4e8bb2666a7257f0f5986800f96bb88efdd6acbd/test/mri/fiddle/test_func.rb#L44

a string pointer #to_s should return a ruby string. That would be tricky to implement :)

headius:
Do you think there's any work remaining here or could we go ahead with a merge?

headius:
I have merged jruby/ruby#2 and jruby/ruby#3 across the two active branches `jruby-ruby_2_5_7` and `jruby-ruby_2_6_5` and incorporated their changes into the JRuby branches `master` and `ruby-2.6`.

I also figured out why the libc changes keeps disappearing: we were versioning `lib/ruby/stdlib/fiddle/jruby.rb` in both places when it's specific to JRuby, without any equivalent in CRuby, and subject to frequent updates. I modified all branches to reflect this, with messages about the proper versioning location added to jruby [here](https://github.com/jruby/jruby/commit/18082826ff06a10d71ec38c7648531210692790e) and [here](https://github.com/jruby/jruby/commit/b2b9e7d81c0e051d0c1ec919fb9f8946d9955c11).

(I see the spelling error now and will fix that).

I've also made sure that all changes from jruby/ruby's `jruby-ruby_2_5_7` and jruby/jruby's `master` branch have been incorporated back into the jruby/ruby `jruby-ruby_2_6_5` branch upon which the jruby/jruby `ruby-2.6` branch is based. See jruby/ruby@afa0b2506a928b6958163d74972ac6c94b2d9f7e.

We need a better way to manage our patch set against CRuby's stdlib.

ahorek:
> Do you think there's any work remaining here or could we go ahead with a merge?

there're still many broken parts, but I think it's ready for review

headius:
Everything here seems appropriate to me. I will merge and we can do other fixes in other PRs.

PR #5963, opened 2019-11-07
improve BUILDING.md

k77ch7:
following commands does not work for me.
`mvn -Pbootstrap`
`mvn -Pbootstrap-no-launcher`
`mvn -pl core`

My Env
```
$ ./mvnw -v
Apache Maven 3.6.0 (97c98ec64a1fdfee7767ce5ffb20918da4f719f3; 2018-10-25T03:41:47+09:00)
Maven home: /Users/xxx/.m2/wrapper/dists/apache-maven-3.6.0-bin/3rgjh30jneo7541hun7uggltkb/apache-maven-3.6.0
Java version: 11.0.2, vendor: Oracle Corporation, runtime: /Library/Java/JavaVirtualMachines/openjdk-11.0.2.jdk/Contents/Home
Default locale: ja_JP, platform encoding: UTF-8
OS name: "mac os x", version: "10.15", arch: "x86_64", family: "mac"
```

### DIFF
--- a/BUILDING.md
+++ b/BUILDING.md
@@ -78,13 +78,13 @@ environment. This will do the following:
  needed to run integration tests.
 
 ```
-mvn -Pbootstrap
+./mvnw -Pbootstrap
 ```
 
 In case there is a problem with installing the jruby-launcher (due to missing compiler or so) use
 
 ```
-mvn -Pbootstrap-no-launcher
+./mvnw -Pbootstrap-no-launcher
 ```
 
 This only needs to be run once to install these gems or if you update
@@ -96,7 +96,7 @@ After changing Java code, you can recompile quickly by running one of the
 jar files by
 
 ```
-mvn -pl core
+./mvnw -pl core
 ```
 
 ### Day to Day Testing


headius:
Hmm I notice we do use `./mvnw` at the beginning of the doc, and at some point we switch back to regular `mvn` assuming it's installed in PATH somewhere. It seems like we should be consistent, or else make the `mvn` command lines clearly demarcate that this must be your working maven command, a la something like this:

```
<mvn command> -Pbootstrap
```

Thoughts? @enebo @kachick @mkristian

enebo:
I do not feel abstracting this is not helpful.  If you are a maven user 'mvn' is a known thing.  Seeing ./mvnw for them will know they can just type 'mvn'.  If not then possibly just using ./mvnw everywhere will help the non-maven users more.

mkristian:
for me using `./mvnw` also means that we know the maven version and all works. using different maven version may or may not have issues

enebo:
@mkristian ah yeah good point.  Another thing a maven user may figure out but a new user will likely get tripped up on.

k77ch7:
Thanks for your reply.
I had overlooked the following WARNING.

```
[WARNING] Rule 0: org.apache.maven.plugins.enforcer.RequireMavenVersion failed with message:
Detected Maven Version: 3.2.1 is not in the allowed range [3.3.0,).
```
After changing mvn version 3.2.1 to 3.3.9, it now works.

What do you think about this  diff?
```diff
diff --git a/BUILDING.md b/BUILDING.md
index ea56d7b2cc..d778304133 100644
--- a/BUILDING.md
+++ b/BUILDING.md
@@ -5,7 +5,7 @@ Prerequisites:

 * A [Java 8-compatible (or higher) Java development kit (JDK)](http://www.oracle.com/technetwork/java/javase/downloads/index.html).
   * If `JAVA_HOME` is not set on Mac OS X: `export JAVA_HOME=$(/usr/libexec/java_home)`
-* [Maven](https://maven.apache.org/download.cgi) 3+ (Maven Wrapper provided with `./mvnw`)
+* [Maven](https://maven.apache.org/download.cgi) 3.3.0+ (Maven Wrapper provided with `./mvnw`)
 * [Apache Ant](https://ant.apache.org/bindownload.cgi) 1.8+ (see https://github.com/jruby/jruby/issues/2236)
 * [Make](https://www.gnu.org/software/make/) and a C++ compiler for installing the jruby-launcher gem
```

headius:
@k77ch7 Yes I think that is a better option. We only mention `mvnw` as a way for people with no maven (or the wrong version) to get a working version, and I think it's cleaner for us to just refer to `mvn` from then on. Your change will help make it clear what our minimum version is.

If you are ok with it, I will commit your proposed version-update patch with attribution and close this PR. Ok?

k77ch7:
@headius  It’s OK.

headius:
Committed the minimal version update in 9970478ca6. Thanks!

PR #5966, opened 2019-11-08
Remove usages of deprecated ConfigMap

deivid-rodriguez:
Rubygems deprecated it in https://github.com/rubygems/rubygems/pull/2857.


headius:
Good find, thanks!

PR #5967, opened 2019-11-08
Open addressing hash redux

headius:
Ongoing work to make @ChrisBr's open-addressing hash impl thread-safe (enough) to merge.


core/src/main/java/org/jruby/java/proxies/MapJavaProxy.java
```diff
@@ -265,7 +253,7 @@ public RubyHashEntry internalDeleteEntry(final RubyHashEntry entry) {
             // NOTE: this is here to make maps act similar to Hash-es which allow modifications while
             // iterating (meant from the same thread) ... thus we avoid iterating entrySet() directly
             final Map<Object, Object> map = mapDelegate();
-            final Map.Entry[] entries = map.entrySet().toArray(NULL_MAP_ENTRY);
+            final Map.Entry[] entries = map.entrySet().toArray(new Map.Entry[map.size()]);
```
kares: we might have had a AIOoBE here - end using `Entry[0]` array resolved the issue

headius: How could it be AIOOBE? The map's size can't be less than zero, and constructing a zero-length array is ok.

Though really it should use the same empty array when it's zero-length...

kares: https://github.com/jruby/jruby/commit/40f40fb9fae31578ba8c5b2a026df883ef9df7d7 it was a NPE really when the size is bigger than actual contents (in Rhino's case)

ChrisBr:
That's great 🎉 Anything I could help out?

headius:
@ChrisBr Can you poke around and find (or make?) some concurrent-access specs? There may be something we can borrow from TruffleRuby (I believe they make more thread-safety guarantees for their hash than we do) or repurpose from concurrent-ruby (which ships some thread-safe collections they must be testing).

I'm going to try out some simple approaches to synchronizing, like an atomic integer spin lock.

headius:
Here's a simple synchronization patch around all places where "extents" is accessed. It mostly passes tests, but fails a threaded iterator test in test/mri/test_hash.rb around line 1203. The deadlock is due to one thread starting an iteration and then another thread also tries to acquire the first thread's hash's lock.

https://gist.github.com/headius/340794413b33688f78325fc6f8398f40

The semantics of concurrent iteration need to be explored. I also did a partial experiment to spinlock around these same operations, but also did not have a good answer for concurrent deadlocking iterators.

headius:
The problem with `test_threaded_iter_level` is that it tries to get both threads going before proceeding, but having a hard lock around iteration means the second thread never can get going.

I poked around TruffleRuby a bit and it appears they have two lock-free strategies for iterating:

* For packed arrays (small) they make a defensive copy of the entry array and iterate against that.
* For chained buckets (large) they iterate the bucket chain which is not sensitive to threading.

We can certainly do the first strategy in the case of linear search table. The second will probably require some ingenuity to avoid making large defensive copies (e.g. maybe read locking to allow concurrent iteration without modification).

headius:
This experiment is unlikely to ever be revisited, and with the possibility of using purpose-built collections like HashSmith (https://github.com/jruby/jruby/issues/9394) I think we can safely close this.

The work to be done for #9394 is now supported in 10.1.x by RubyHash being abstract. We just need to work through how to efficiently present the HashSmith collections as JRuby collections.

PR #5977, opened 2019-11-22
fix 'jruby: line 88: cd: ../libexec/bin: No such file …'

pythias:
MacOS
```
brew install jruby
jruby -v
/usr/local/bin/jruby: line 88: cd: /usr/local/bin/jruby: Not a directory
Error: --module-path requires module path specification
```


headius:
Request review from @Mrnoname1000.

Mrnoname1000:
Some testing shows incorrect paths in the original logic due to some weird issues with symlinked directories. I think the cd to `BASE_DIR` in the loop was a typo, and this fixes that. Looks good to me.

headius:
@mrnoname1000 Thanks!

PR #5978, opened 2019-11-22
Cache method_missing target

headius:
This logic will cache the target class's method_missing impl when
a method call fails to find a natural method table entry. This
scenario was a source of failed call sites in Redmine's json
builder and in the Rails StringInquirer class used for Rails.env.
With this change in place Redmine json issue requests improved by
roughly 15%.


headius:
Note that this caching is still doing an indirect call, via `DynamicMethod#call`. There's an additional opportunity here by specializing the bound target and making it possible to inline `method_missing`.

headius:
I did not expect it to pass, but that's great!

PR #5984, opened 2019-11-26
Bind only public interface methods for private classes

headius:
This includes two fixes introduced (at least in part) by #5832:

* Private classes were considered for method binding even though
  we no longer set them accessible on Java 9+ when they have not
  already been opened. This resulted in Java 9+ environments
  producing method errors when they attempted to dispatch to those
  methods on private classes.
* In order to fix the above for private classes with public
  interfaces, this patch also fixes how we traverse a class's
  implemented interfaces. Possibly from changes in #5832, we were
  not walking an interface's superinterfaces, causing us to only
  walk one level in such cases.

In my local tests, this caused failures in `spec:ji` for specs accessing the following classes; I do not believe that these methods should be callable, since they are on private inner classes without any public interface method:

https://github.com/jruby/jruby/blob/b35397b767ccf9b8a0f29e23f9b338576802acc9/spec/java_integration/fixtures/InnerClasses.java#L78-L95

### DIFF
--- a/core/src/main/java/org/jruby/javasupport/binding/MethodGatherer.java
+++ b/core/src/main/java/org/jruby/javasupport/binding/MethodGatherer.java
@@ -23,8 +23,10 @@ import java.util.ArrayList;
 import java.util.Arrays;
 import java.util.Collections;
 import java.util.HashMap;
+import java.util.HashSet;
 import java.util.List;
 import java.util.Map;
+import java.util.Set;
 import java.util.function.Function;
 
 import static org.jruby.runtime.Visibility.PUBLIC;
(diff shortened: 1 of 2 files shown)


headius:
@kares @enebo I do not think the failing spec should pass. I have not looked into its lineage yet, but the `capsMethod` in question is only defined on those private inner classes. It likely passed previously because we were aggressively setting the methods accessible.

Java would not be able to compile against or invoke these methods for a normal invocation.

headius:
Confirmed in both cases that Java cannot call the `capsMethod` in question, and so we should not be binding it. The general rule of thumb is that we don't bind methods on non-visible classes.

kares:
makes sense but what about compatibility for users on 8,
they might expect private methods to be callable, isn't this a breaking change?

enebo:
@kares I will make an argument that binding to private without an explicit setAccessible in their code was never intended behavior.  In this case we can maybe just point out this was unintended behavior (e.g. a bug).  I guess the bigger question with that answer is how common this is in reality.

PR #5987, opened 2019-12-01
Starting a French translation

liliantucho:


headius:
Thank you! We welcome translations of the base docs in our distribution. I do wonder what's typical for managing these documents in other projects. Do they typically keep a bunch of localized readmes at the root or is it more common to put them in a localized documentation subdirectory?

headius:
Also I did not look at your Code of Conduct translation, but there's existing official translations here: https://www.contributor-covenant.org/translations

liliantucho:
We are actually students, we are thinking on how to manage the different languages and will come back to you later about it.

headius:
@liliantucho Thank you! We will happily accept translations once there's a good place to put them.

headius:
@liliantucho We are trying to finish up 9.3 so I will need to remove this issue from that release. We would still love to incorporate a French translation if there is interest.

PR #5992, opened 2019-12-07
Make many fields on Ruby final.

headius:
Many recent JITs have started to optimize repeat final field
accesses. This patch moves almost all core class and module init
into the Ruby constructor, so that those fields (as well as some
structural references) can be made final. This should help any
core classes that look up their own type on each construction
(which includes almost all of them) when those constructors are
inlined together with other accesses of the same field.

### DIFF
--- a/core/src/main/java/org/jruby/AbstractRubyMethod.java
+++ b/core/src/main/java/org/jruby/AbstractRubyMethod.java
@@ -88,7 +88,7 @@ public abstract class AbstractRubyMethod extends RubyObject implements DataType 
 
     @JRubyMethod(name = "eql?", required = 1)
     public IRubyObject op_eql(ThreadContext context, IRubyObject other) {
-        return context.runtime.newBoolean( equals(other) );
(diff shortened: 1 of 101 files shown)


headius:
This is largely complete. There are certainly other fields in JRuby that could be made final but it's starting to get far away from the Ruby class at this point.

Based on some simple looping benchmarks, there appears to be roughly a 5-10% improvement (rather noisy on this machine) on a loop that constructs five arrays each containing just the current iteration value. On a JIT like Graal that can eliminate the loop and the arrays, the PR has a larger impact, in the neighborhood of 40%+ performance improvement. This is largely due to the final Boolean values coming from ThreadContext, but the Fixnum class access also comes into play.

In any case this is difficult to show in a microbenchmark, since the fields in question get cached in L1 very quickly regardless of whether they're final or not. The difference becomes a matter of accessing the value from L1 or from a register, assuming it doesn't spill to memory. But after looking at IGV output from Graal it's clear that the final values are getting accessed from memory only once when final.

PR #5999, opened 2019-12-12
ensure javase version in osgi meta is 8 and not 9

rmannibucau:
side note: it fixes the javase version but there are still a lot of warning so some refinements can be nice later


headius:
Thank you!

PR #6007, opened 2019-12-21
Bash cleanup

Mrnoname1000:
Here are a bunch of changes cleaning up the Bash launcher. Most of it is assisted by [shellcheck](https://www.shellcheck.net).
Changes include syntax improvements, removing unused variables, and some minimal syntax changes to make it easier to port to POSIX shell in the future.
See commit messages for more details.

### DIFF
--- a/bin/jruby.bash
+++ b/bin/jruby.bash
@@ -284,15 +284,18 @@ do
      -J*)
          java_args=("${java_args[@]}" "${1#-J}") ;;
      # Pass -X... and -X? search options through
-     -X*\.\.\.|-X*\?)
+     -X*...|-X*\?)
         ruby_args=("${ruby_args[@]}" "$1") ;;
      # Match -Xa.b.c=d to translate to -Da.b.c=d as a java option
      -X*.*)
         java_args=("${java_args[@]}" "-Djruby.${1#-X}") ;;
      # Match switches that take an argument
-     -C|-e|-I|-S) ruby_args=("${ruby_args[@]}" "$1" "$2"); shift ;;
+     -C|-e|-I|-S)
+        ruby_args=("${ruby_args[@]}" "$1" "$2")
+        shift ;;
      # Match same switches with argument stuck together
-     -e*|-I*|-S*) ruby_args=("${ruby_args[@]}" "$1" ) ;;
+     -e*|-I*|-S*)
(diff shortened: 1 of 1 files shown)


enebo:
If there ends up being some esoteric platform which somehow disagrees with any of this we should land it sooner than later.

headius:
I discussed with @mrnoname1000 the remaining steps for POSIX shell support and most of them boil down to leveraging sh argument arrays to replace the += bash arrays we use heavily. Historically the main trouble we had with sh was losing proper quoting when forced to reinterpolate arguments during arg processing. The other challenge is that there's no readlink in posix so we'd either have to require that (probably not a big deal) or import a non-trivial sh-based implementation of it. Thanks for the deep investigation @mrnoname1000!

enebo:
@headius @Mrnoname1000 sounds great.  Using simplest sh that is portable sounds like a dream to me :)

bin/jruby.bash
```diff
@@ -125,41 +115,40 @@ add_log "  JAVA_OPTS: $JAVA_OPTS"
...
-    JAVA_HOME=$(dirname $(dirname $expanded_javacmd))
+  expanded_javacmd="$(command -v "$JAVACMD")"
+  if [ -z "$JAVA_HOME" ] && [ -x "$expanded_javacmd" ] ; then
+    JAVA_HOME="$(dirname "$(dirname "$expanded_javacmd")")"
   fi
 fi
 
 # Detect Java 9+ by the presence of a jmods directory in JAVA_HOME
-if [ -d $JAVA_HOME/jmods ]; then
+if [ -d "$JAVA_HOME"/jmods ]; then
```
eregon: Note: this actually fixes a warning running on Windows in a Bash shell:
```
/c/Users/runneradmin/.rubies/jruby-9.2.9.0/bin/jruby: line 155: [: C:\Program: binary operator expected
```
This happens in GitHub Actions.

headius: Makes sense with spaces in path.

eregon: Yeah, exactly. Using the Bash launcher Windows is likely unusual though (Just `jruby` would give precedence to the `.exe`, only `/full/path/to/jruby` or `jruby.bash` in a Bash shell would use the Bash launcher).

PR #6010, opened 2019-12-26
Ruby 2.6 : Add String#split with a blcok

k77ch7:
This PR implements String#split with a blcok.

 See [Feature #4780](https://bugs.ruby-lang.org/issues/4780).


kares:
we might prefer keeping the old signatures - as it might be used by some ext.
and adding a new split overloads with the `IRubyObject` return and a block arg (annotated with @JRubyMethod) ...

k77ch7:
@kares Thanks for your advice. 
I tried refactoring. 
Is my understanding correct?

kares:
thanks - the idea for having split with block might have been to avoid creating the array. 
still good enough as an initial implementation!

PR #6011, opened 2019-12-27
Ruby 2.6 : Add Random.bytes

k77ch7:
This PR implements Random.bytes method.

See Feature [#4938](https://bugs.ruby-lang.org/issues/4938)

### DIFF
--- a/core/src/main/java/org/jruby/RubyRandom.java
+++ b/core/src/main/java/org/jruby/RubyRandom.java
@@ -683,12 +683,16 @@ public class RubyRandom extends RubyObject {
     // c: random_s_bytes
     @JRubyMethod(meta = true)
     public static IRubyObject bytes(ThreadContext context, IRubyObject recv, IRubyObject arg) {
-        return ((RubyRandom) (context.runtime.getRandomClass()).getConstant("DEFAULT")).bytes(context, arg);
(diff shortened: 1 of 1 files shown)


core/src/main/java/org/jruby/RubyRandom.java
```diff
@@ -680,6 +680,12 @@ public IRubyObject marshal_load(ThreadContext context, IRubyObject arg) {
         return this;
     }
 
+    // c: random_s_bytes
+    @JRubyMethod(meta = true)
+    public static IRubyObject bytes(ThreadContext context, IRubyObject recv, IRubyObject arg) {
+        return ((RubyRandom) (context.runtime.getRandomClass()).getConstant("DEFAULT")).bytes(context, arg);
```
kares: MRI works even if you change the `Random::DEFAULT` constant
so we should follow that, use `context.runtime.getDefaultRand()` to retrieve the internal (default) random.
the existing bytes should be refactored to an internal static helper that receives a `RandomType random` arg

k77ch7: @kares i will try refactoring.

k77ch7: @kares Please review. Thank you.

PR #6014, opened 2019-12-31
correct/revisit exception backtrace (and stack-trace)

kares:
original motivation for these changes was that the internal "lightweight" forced backtrace didn't have any reduced impact over a normal `raise` but also the fast that the legacy Java parts got neglected (not having the proper forced stack-trace as before).

last part is probably due having two `TestRaiseException` cases - one of which wasn't run since the recent changes (adding the Java exception hierarchy for Ruby ones) ...

than I realized `RaiseException` usually replaces a custom Java trace so not need to initially `fillInStackTrace` (which turned out to be potentially incorrectly on with `-Xjump.backtrace`)

p.s. also added some minor pre-filtering (e.g. `getStackTrace` from top - has been bothering me for a while :see_no_evil:) - we could take it further to skip the Java stack trace head to not include all of the exception `<init>` (all the way to `java.lang.Throwable.<init>`) ... maybe later.


kares:
numbers confirm that a forced back-trace now is much faster (`raise RuntimeError, '', nil`) : 
```
       raise default   841.000  i/100ms
 raise nil backtrace   898.000  i/100ms
       raise default      9.296k (± 1.2%) i/s -     93.351k in  10.044026s
 raise nil backtrace      9.290k (± 1.8%) i/s -     93.392k in  10.056207s
```
```
       raise default   867.000  i/100ms
 raise nil backtrace    78.471k i/100ms
       raise default      9.434k (± 1.6%) i/s -     94.503k in  10.019629s
 raise nil backtrace      1.626M (± 2.2%) i/s -     16.322M in  10.040849s
```

kares:
Java 11 is half the speed for 'normal' raises (no change and I re-call this is a known thing) : 
```
jruby 9.2.9.0 (2.5.7) 2019-10-30 458ad3e Java HotSpot(TM) 64-Bit Server VM 11.0.2+9-LTS on 11.0.2+9-LTS +jit [linux-x86_64]

       raise default   449.000  i/100ms
 raise nil backtrace   466.000  i/100ms
       raise default      4.695k (± 1.7%) i/s -     47.145k in  10.045269s
 raise nil backtrace      4.665k (± 1.8%) i/s -     47.066k in  10.092519s
```
```
jruby 9.2.10.0-SNAPSHOT (2.5.7) 2019-12-31 fe4b188 Java HotSpot(TM) 64-Bit Server VM 11.0.2+9-LTS on 11.0.2+9-LTS +jit [linux-x86_64]

       raise default   464.000  i/100ms
 raise nil backtrace    72.106k i/100ms
       raise default      4.962k (± 1.1%) i/s -     49.648k in  10.006191s
 raise nil backtrace      1.489M (± 1.2%) i/s -     14.926M in  10.027093s

```
... maybe something to tackle with later :crossed_fingers:

core/src/main/java/org/jruby/exceptions/RaiseException.java
```diff
@@ -128,7 +128,19 @@ private void preRaise(ThreadContext context, IRubyObject backtrace, boolean capt
...
+                skip = 5;
+            } else if ("preRaise".equals(rubyTrace[3].getMethodName())) {
+                skip = 4;
+            }
+            // NOTE: we could skip more useless Throwable.<init> hierarchy constructor trace
+            //  up to org.jruby.exceptions.RaiseException.from ?
+            rubyTrace = Arrays.copyOfRange(rubyTrace, skip, rubyTrace.length);
+        }
+        setStackTrace(javaTraceFromRubyTrace(rubyTrace));
     }
```
headius: I wonder if we can just filter out all frames where the package starts with "java." might be safe too? All traces we initiate should be rooted as close to the Ruby code or core class code as possible.

headius: I wonder if we can just filter out all frames where the package starts with "java." might be safe too? All traces we initiate should be rooted as close to the Ruby code or core class code as possible.

core/src/main/java/org/jruby/exceptions/RaiseException.java
```diff
@@ -91,28 +102,71 @@ public final RubyException getException() {
...
+            }
         } else {
             exception.setBacktrace(backtrace);
+            if (!backtrace.isNil() && !isEmptyArray(backtrace)) {
+                exception.captureBacktrace(context);
+            } else { // !backtrace.isNil()
+                //fillInStackTraceSkip(1 + 2); // TODO: we fill the stack due backwards compatibility
+                // we likely want a flag like -Xexception.java.trace to fine control filling in the stack-trace
+                // (but ONLY for sub-classes of RaiseException?)
+            }
```
headius: I think we should just remove this todo and discuss filtering separately. It would be a bonus if we can move the actual trace acquisition closer to the relevant code or convince Java not to fill in things like "fillInStackTrace".

headius:
That's a 👍 from me even though I submitted the review as a general comment.

core/src/main/java/org/jruby/exceptions/RaiseException.java
kares: leftover, definitely shouldn't be in. will cleanup

core/src/main/java/org/jruby/exceptions/RaiseException.java
kares: in that case I think we would reach JRuby's `JumpException` constructor as top.
was wondering whether to go further - in relation to the Throwable ctor hierarchy.
we might want to skip all of those ctors so a "java." prefix won't do ... or maybe we do not :yin_yang:  
intentionally did not push this too far and rather check at exact places (which is quite ugly but working fine), for now to avoid more costs in looping the whole stack-trace for another time.

headius: Yeah I'm fine pushing any further filtering to a later discussion.

Issue #6004, opened 2019-12-19
Lots of parked "Fiber thread for block at" threads


Unknown-Guy:
**Environment**
 
* `JRUBY_OPTS='--dev --headless -J-XX:+UseStringDeduplication'`
* `jruby 9.2.9.0 (2.5.7) 2019-10-30 458ad3e OpenJDK 64-Bit Server VM 25.212-b03 on 1.8.0_212-b03 [darwin-x86_64]`
* `Darwin Me-MBP-3.local 19.2.0 Darwin Kernel Version 19.2.0: Sat Nov  9 03:47:04 PST 2019; root:xnu-6153.61.1~20/RELEASE_X86_64 x86_64`

Other relevant info you may wish to add:
*  does not happen on 9.2.7.0, same issue on 9.2.8.0
* [thread dump](https://gist.github.com/Unknown-Guy/36dad4b61e53b5366b49208baafd9283) 

**Expected Behavior**

Does not generate excessive amount of threads

**Actual Behavior**

VisualVM shows lots of threads in for `Fiber thread for block at: uri:classloader:/jruby/kernel/enumerator.rb:62` in `Park` state and running our apps full test suite ends up with running out of threads at around 4k.

headius:
So something changed in 9.2.8 that's causing these to not get cleaned up.

Fibers are used to implement `Enumerator#next`, and on JRuby each fiber requires a native thread. If you do not `next` your way completely through the collection, that means the fiber thread is sitting there waiting for another call. We have taken pains to make sure that fibers that are abandoned will eventually get cleaned up (usually via GC triggering finalization) but perhaps something broke along the way that's causing this to fail now.

headius:
Ok, so the big thing that landed in 9.2.8.0 was the actual move to a pure-Ruby `Enumerator#next` based on Fiber. This was done to eliminate a second "fiber-like" bit of logic we had been maintaining separately.

So we need to figure out why these fibers are sticking around.

1. Do you know what code is causing them to be created? It would be `Enumerator#next` logic, clearly. You may be able to pass `-Xjit.threshold=0` to JRuby to force more code to JIT compile and show up in the stack.
2. Does forcing GC clean them up? You can connect with `jconsole` or `jvisualvm` and force a GC (or two or three) and see if they clear up.
[...]

PR #6021, opened 2020-01-07
Fix Enumerator fiber leaks

headius:
This PR will include commits to reduce the chance of `Enumerator#next` fibers leaking.

The original issue is #6004.

So far most of the leaks appear to be caused by combining `Enumerator::Lazy` with `Enumerator#next`, since many lazy paths construct new Lazy instances in such a way that they are hard-referenced by the eventual fiber.


headius:
In the interest of improving things for 9.2.10 I have merged this. There's still known leaks in lazy enumerators and the root cause of #6004 remains elusive but it is a step forward.

Issue #6025, opened 2020-01-09
SyntaxError (Rational (174532925199432957/1E+19) out of range)


284km:
**Environment**
  
```
$ jruby -v
jruby 9.2.9.0 (2.5.7) 2019-10-30 458ad3e OpenJDK 64-Bit Server VM 12.0.2+10 on 12.0.2+10 +jit [darwin-x86_64]

$ uname -a
Darwin Apple197 17.7.0 Darwin Kernel Version 17.7.0: Sun Jun  2 20:31:42 PDT 2019; root:xnu-4570.71.46~1/RELEASE_X86_64 x86_64
```

**Expected Behavior**

I expect the same result as CRuby below.

```
$ ruby -v
ruby 2.6.5p114 (2019-10-01 revision 67812) [x86_64-darwin17]
$ irb
irb(main):001:0> 0.0174532925199432957r
=> (174532925199432957/10000000000000000000)

$ ruby -v
ruby 2.7.0p0 (2019-12-25 revision 647ee6f091) [x86_64-darwin17]
cirb(main):001:0> 0.0174532925199432957r
=> (174532925199432957/10000000000000000000)
```

Here is our use case. We want to convert from Degree to Radian.

https://github.com/red-data-tools/red-colors/blob/9d97f82043c05b5a6d460efc84cb4cf5e8c4d1bf/lib/colors/convert.rb#L56

**Actual Behavior**

```
$ jruby -v
jruby 9.2.9.0 (2.5.7) 2019-10-30 458ad3e OpenJDK 64-Bit Server VM 12.0.2+10 on 12.0.2+10 +jit [darwin-x86_64]
$ irb
irb(main):001:0> 0.0174532925199432957r
Traceback (most recent call last):
        5: from /Users/kazuma.furuhashi/.rbenv/versions/jruby-9.2.9.0/bin/irb:13:in `<main>'
        4: from org/jruby/RubyKernel.java:1193:in `catch'
        3: from org/jruby/RubyKernel.java:1193:in `catch'
        2: from org/jruby/RubyKernel.java:1446:in `loop'
        1: from org/jruby/RubyKernel.java:1052:in `eval'
SyntaxError ((irb):1: Rational (174532925199432957/1E+19) out of range.)
0.0174532925199432957r

# FYI, This is the expected result.
irb(main):002:0> 0.017453292519943295r
=> (17453292519943295/1000000000000000000)
```

PR #6027, opened 2020-01-10
[fix] SyntaxError rational overflow

ahorek:
https://github.com/jruby/jruby/issues/6025


headius:
Thank you!

kares:
the test added might end up lost in the next test update from MRI
so the test should be pushed upstream or rather become a spec ...

ahorek:
done, thanks

PR #6028, opened 2020-01-11
revert rational test

ahorek:
https://github.com/ruby/spec/pull/750


enebo:
Coolio.  Thanks for adding it as a spec!

PR #6031, opened 2020-01-12
[fix] restore rescue nil performance 

kares:
initially I thought `rescue nil` back-trace capturing is disabled on purpose
but seems it simply got lost at f64a3e0

follow-up on https://github.com/jruby/jruby/pull/6014

should introduce some perf.testing to avoid loosing performance

### DIFF
--- a/bench/core/kernel/bench_raise.rb
+++ b/bench/core/kernel/bench_raise.rb
@@ -1,0 +1,48 @@
+require 'benchmark/ips'
+
+Benchmark.ips do |x|
+
+  x.config(:time => 10, :warmup => 10)
+
+  x.report("raise(msg) default") do
+    begin
+      raise 'default'
+    rescue => ex
+      ex && true
+    end
+  end
+
+  x.report("raise(class, msg)") do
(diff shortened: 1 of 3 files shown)


headius:
I wish there were a better way to test this but everything I can think of in Ruby would defeat the optimization.

Perhaps we could introduce something that tests the IR directly? Confirm that something like `eval "foo rescue nil"` includes the appropriate instruction to disable backtrace?

Looks good in any case, thank you!

kares:
hey, thanks for the merge and response.

had the idea of doing like a 'performance' unit test (haven't looked into it).
establish a Java stack-trace-less exception baseline cost and assert our stack-trace less pass are not order of magnitude slower.

headius:
That could work too. My main worry about performance tests is that Travis will sporadically be slow, or become slow, for a given build. It would be better than nothing though.

We could also look at just setting up a periodic dedicated-server run of all benchmarks we've committed, tracking them over time like the Pypy project has done. Then at least we could get some mild notification if something regresses seriously.

Issue #5576, opened 2019-01-27
Ruby 2.6 Support


anubhav8421:
This issue will track Ruby 2.6 support for JRuby. This will be released in JRuby 9.3.0.0

As with previous releases, we will base our checklist on MRI's NEWS file: https://github.com/ruby/ruby/blob/ruby_2_6/NEWS
This bug will track all NEWS items new in Ruby 2.6. This list is based off https://github.com/ruby/ruby/blob/ruby_2_6/NEWS.

Only changes relevant to JRuby are listed here. MRI-specific internal changes and features are not included.

NOTE: Pull requests should be done against the ```ruby-2.6``` branch. That branch will be merged via #6037.

## Language changes

- [ ] Refinements take place at block passing [Feature #14223](https://bugs.ruby-lang.org/issues/14223)
- [ ] Refinements take place at Kernel#public_send [Feature #15326](https://bugs.ruby-lang.org/issues/15326)
- [ ] Refinements take place at Kernel#respond_to? [Feature #15327](https://bugs.ruby-lang.org/issues/15327)
- [x] ```else``` without ```rescue``` causes a syntax error [Experimental] [Feature #14606](https://bugs.ruby-lang.org/issues/14606)
- [x] Constant names may start with a non-ASCII capital letter [Feature #13770](https://bugs.ruby-lang.org/issues/13770)
- [x] Endless ranges are introduced. Range that has no end, like ```(0..)``` or ```(0...)``` [Feature #12912](https://bugs.ruby-lang.org/issues/12912)
- [ ] Non-Symbol keys in a keyword arguments hash cause an exception.
- [x] The “shadowing outer local variable” warning is removed.  [Feature #12490](https://bugs.ruby-lang.org/issues/12490)
- [ ] Print ```cause``` of the exception if the exception is not caught and printed its backtraces and error message. [Feature #8257](https://bugs.ruby-lang.org/issues/8257)
- [x] The flip-flop syntax is deprecated. [Feature #5400](https://bugs.ruby-lang.org/issues/5400)

### Array

- [x] Add Array#union and Array#difference instance methods. [Feature #14097](https://bugs.ruby-lang.org/issues/14097)
- [x] Array#to_h now accepts a block that maps elements to new key/value pairs. [Feature #15143](https://bugs.ruby-lang.org/issues/15143)
- [x] Array#filter is a new alias for Array#select. [Feature #13784](https://bugs.ruby-lang.org/issues/13784)
[...]

PR #6033, opened 2020-01-13
Ruby 2.6 : Enumerator::Chain class, Enumerable#chain, Enumerator#+

k77ch7:
This PR implements Enumerator::Chain class, Enumerable#chain and Enumerator#+.

See Feature [#15144](https://bugs.ruby-lang.org/issues/15144)


enebo:
@k77ch7 if you want to you can also use Java lambdas when possible (e.g.   return new SizeFn() {).  We have not required it but it is an option if you want to in future commits.

headius:
I have updated the checklist at https://github.com/jruby/jruby/issues/5576

k77ch7:
@enebo Understood. thanks for reviewing.

PR #6040, opened 2020-01-15
make jruby-core-source jar has no absolute path from home dir

### LINKED ISSUE #5979: Maven source artifact contains full build path
The current source artifact published on maven central ( https://repo1.maven.org/maven2/org/jruby/jruby-core/9.2.9.0/jruby-core-9.2.9.0-sources.jar ) contains absolute paths for some files, such as `/home/enebo/work/release/core/target/classes/META-INF/LEGAL`.

The expected behavior would be to either exclude these files completely, or to move them to their appropriate location such as `META-INF/LEGAL`.

The files in question can be found by running `unzip -l` on the source jar file linked above, and then scrolling to the very bottom of the output.

mkristian:
fixes #5979

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -46,6 +46,7 @@ glob_test
 jruby-complete.jar
 jruby-findbugs.html
 latest_source_cache
+core/src/main/java/META-INF/
 lib/jni
 lib/jruby*.jar
 lib/native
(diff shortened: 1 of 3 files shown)


mkristian:
@enebo please have a look. copying the legal stuff into core/src/main/java/META-INF is a bit hacky but did not get it to work otherwise

PR #6052, opened 2020-02-08
Make Symbol#to_s, TrueClass#to_s, FalseClass#to_s and Module#name return frozen strings

nomadium:
This is a continuation of the work started by @rdubya in https://github.com/jruby/jruby/pull/5868.

For more information, take a look at https://bugs.ruby-lang.org/issues/16150.


core/src/main/java/org/jruby/RubyNil.java
```diff
@@ -144,7 +144,14 @@ public static RubyFloat to_f(ThreadContext context, IRubyObject recv) {
      */
     @JRubyMethod
     public static RubyString to_s(ThreadContext context, IRubyObject recv) {
-        return RubyString.newEmptyString(context.runtime);
+        Ruby runtime = context.runtime;
+        RubyString emptyString = RubyString.newEmptyString(runtime);
+
+        if (runtime.getInstanceConfig().isFrozenCoreToS()) {
+            return runtime.freezeAndDedupString(emptyString);
```
lopex: does it make sense to dedup an emty string here ?

nomadium: Is there an issue with deduplication because its cost? Is it a better idea to just return a frozen copy?

lopex: I was more thinking about of just returning a frozen dedupped singleon for that runtime

nomadium: Right, I see that singleton could be created by the runtime. I just updated the PR accordingly.

core/src/main/java/org/jruby/RubyBoolean.java
```diff
@@ -163,7 +163,14 @@ public static IRubyObject false_xor(ThreadContext context, IRubyObject fals, IRu
 
         @JRubyMethod(name = "to_s", alias = "inspect")
         public static RubyString false_to_s(ThreadContext context, IRubyObject fals) {
-            return RubyString.newStringShared(context.runtime, FALSE_BYTES);
+            Ruby runtime = context.runtime;
+            RubyString toRubyString = RubyString.newStringShared(runtime, FALSE_BYTES);
+
+            if (runtime.getInstanceConfig().isFrozenCoreToS()) {
+                return runtime.freezeAndDedupString(toRubyString);
```
kares: unless its cached we're just doing more work on every `to_s` - this wasn't the intention of the original PR

nomadium: My understanding is that the string is being cached but I could be wrong, of course: https://github.com/jruby/jruby/blob/master/core/src/main/java/org/jruby/Ruby.java#L4587-L4662

What you mean by original PR, #5868 or MRI one?

kares: right, the point being that there was a field check instead of a map lookup.

kares:
MRI ditched the feature, for now, does it make sense for this to be pushed forward anyway?

... and if so this could use some better naming on the switch flag and introduced methods (not that I was able to come up with any - as long as there was only the symbol case: `-Xsymbol.to_s.cached`) :wink:

core/src/main/java/org/jruby/RubyModule.java
```diff
@@ -2387,7 +2387,16 @@ public IRubyObject name() {
 
     @JRubyMethod(name = "name")
     public IRubyObject name(ThreadContext context) {
-        return getBaseName() == null ? context.nil : rubyName().strDup(context.runtime);
+        if (getBaseName() == null) return context.nil;
+
+        Ruby runtime = context.runtime;
+        RubyString name = rubyName().strDup(runtime);
```
kares: name is already cached in a field. thus dup-ing and than freezing, not sure what the added benefit is.
this one either needs more work or should be dropped.
for now, I would vote for the later (and wait for Ruby to decide) ... we would real-world usage numbers.
like in case of the original PR @rdubya actually looked into the `Symbol#to_s` case being used a lot (in Rails).

headius:
Since it sounds like CRuby will not continue with this feature, I guess there's not much more we can do here. Changing this unilaterally on JRuby would be nice but if CRuby will never support it people won't fix their libraries.

I guess we can give up now. 🤷‍♂️

nomadium:
Should we really give up on this? My understanding is that MRI reverted the change on Symbol class but the feature is available in Module#name, TrueClass#to_s, FalseClass#to_s and NilClass#to_s as you can see below:

```
miguel@shirley:~$ ruby2.7 --version
ruby 2.7.0p0 (2019-12-25 revision 647ee6f091) [x86_64-linux-gnu]
miguel@shirley:~$ irb2.7
irb(main):001:0> RUBY_VERSION
=> "2.7.0"
irb(main):002:0> true.to_s.frozen?
=> true
irb(main):003:0> false.to_s.frozen?
=> true
irb(main):004:0> nil.to_s.frozen?
=> true
irb(main):005:0> URI.name.frozen?
=> true
irb(main):006:0> :hello.to_s.frozen?
=> false
irb(main):007:0> ^D
miguel@shirley:~$ 
```

So, I believe there is value in having this feature available behind a flag since jruby master branch is supporting until 2.5. If you think that approach is OK, I can address the comments from @kares and iterate on the PR.

If you think it's not worth to work on this at this time, that's OK as well and thanks for reviewing the changes so far anyway.

kares:
@nomadium interesting, I know they shipped it an than reverted. not sure which release got what.

if it got into 2.7 why not target **2.7** (3.0) I know its a long shot but we would not need the runtime flag and possibly any maintenance overhead if this is shipped in a flag and thus poorly tested.
we circle all the way back to the original PR which only had `Symbol#to_s` patched and needs a `-Xsymbol.to_s.cached` flag. wdyt?

nomadium:
That makes sense to me, to drop from this PR what will be included eventually in 2.7 or 3.0 and have a frozen Symbol#to_s variant with a flag like the original PR intended. If that's OK, I'll update it later with the proposed flag name.

nomadium:
As we chatted before, even when MRI ditched the feature, it could be useful to collect relevant data points. I dropped all 2.7 features from this PR and introduced the flag -Xsymbol.to_s.frozen to enable the optional frozen version of Symbol#to_s.

Feel free to review when you have time.

kares:
looking good, however this being kind of an experimental flag I am not sure we should expose the setting per runtime. maybe just start with checking `Options.SYMBOL_TO_STRING_FROZEN.load()` at the one place its relevant directly, wout the `Ruby` accessors. any concerns there?

nomadium:
That's right, with this being an experiment I should not be exposing related public getters and setters in RubyInstanceConfig class. I just updated the PR.

PR #6054, opened 2020-02-12
Refined send

### LINKED ISSUE #5945: Refinements are not activated in send
Hey! It's me again 🙂

### Environment

Provide at least:
* JRuby version: jruby 9.2.8.0 (2.5.3) 2019-08-12 a1ac7ff OpenJDK 64-Bit Server VM 25.222-b10 on 1.8.0_222-b10 +jit [linux-x86_64]
* Operating system and platform: Linux 562bfe67be53 4.9.125-linuxkit #1 SMP Fri Sep 7 08:20:28 UTC 2018 x86_64 GNU/Linux

### Expected Behavior

The code below prints `"foo"`.

```ruby
class A
end

module Ext
  refine A do
    def foo
      "foo"
    end
  end
end

using Ext

p A.new.send(:foo)
```

### Actual Behavior

```
NoMethodError: undefined method `foo' for #<A:0x305fd85d>
Did you mean?  fork
  <main> at test.rb:14
```

### Note

Refinements have been working with `send` in MRI since 2.5. In 2.6, the support for `public_send` and `respond_to?` has been added (already mentioned in https://github.com/jruby/jruby/issues/5576).

headius:
Kernel#send and BasicObject#__send__ should honor refinements from the surrounding scope.

Note this forces sends in the presence of refinements to require a scope, which currently means a dynamic scope.

Fixes #5945

### DIFF
--- a/core/src/main/java/org/jruby/RubyBasicObject.java
+++ b/core/src/main/java/org/jruby/RubyBasicObject.java
@@ -31,7 +31,7 @@ package org.jruby;
 import org.jcodings.Encoding;
 import org.jruby.ast.util.ArgsUtil;
 import org.jruby.ir.interpreter.Interpreter;
-import org.jruby.runtime.CallSite;
+import org.jruby.parser.StaticScope;
 import org.jruby.runtime.JavaSites;
 import org.jruby.runtime.JavaSites.BasicObjectSites;
 import org.jruby.runtime.callsite.CacheEntry;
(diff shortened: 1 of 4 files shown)


core/src/main/java/org/jruby/RubyBasicObject.java
```diff
@@ -1718,27 +1718,35 @@ public static IRubyObject method_missing19(ThreadContext context, IRubyObject re
     public IRubyObject send(ThreadContext context, IRubyObject arg0, Block block) {
         String name = RubySymbol.objectToSymbolString(arg0);
 
-        return getMetaClass().finvoke(context, this, name, block);
+        StaticScope staticScope = context.getCurrentStaticScope();
+
+        return getMetaClass().finvokeWithRefinements(context, this, staticScope, name, block);
```
enebo: Are these changes needed?  You changed finvoke to finvokeWithRefinements.

core/src/main/java/org/jruby/RubyClass.java
```diff
@@ -484,7 +485,11 @@ static boolean notVisibleAndNotMethodMissing(DynamicMethod method, String name,
     }
 
     public IRubyObject finvoke(ThreadContext context, IRubyObject self, String name, Block block) {
-        CacheEntry entry = searchWithCache(name);
+        return finvokeWithRefinements(context, self, null, name, block);
+    }
+
+    public IRubyObject finvokeWithRefinements(ThreadContext context, IRubyObject self, StaticScope staticScope, String name, Block block) {
```
enebo: If all finvoke calls *withRefinements should we just put the body of *withRefinements into finvoke and not have these extra signatures?

core/src/main/java/org/jruby/RubyBasicObject.java
headius: finvoke will pass null for the scope when calling finvokeWithRefinements. This is the pattern they use in MRI... the non-refined logic calls the refined logic with a null scope.

core/src/main/java/org/jruby/RubyClass.java
headius: It's a signature issue; a new signature is needed to pass in the scope to use for refined lookup.

core/src/main/java/org/jruby/RubyBasicObject.java
enebo: ok yeah I missed that

PR #6056, opened 2020-02-13
[fix] retore $! on closure nested non-local returns

kares:
```ruby
def foo(e)
  fail("foo") if $! != e
  yield
end

def bar
  e = StandardError.new 'foo'
  begin
    raise e
  rescue
    fail("bar") if $! != e
    #1.times do
      foo(e) { return } # PROBLEM
    #end
  end
end

bar

fail("$! #{$!} not cleared!") if $! # FAILS - JRuby retains $!
puts "DONE"
```

### DIFF
--- a/core/src/main/java/org/jruby/ir/IRBuilder.java
+++ b/core/src/main/java/org/jruby/ir/IRBuilder.java
@@ -130,7 +130,7 @@ public class IRBuilder {
         }
     }
 
-    private static class RescueBlockInfo {
+    static class RescueBlockInfo {
         Label      entryLabel;             // Entry of the rescue block
         Variable   savedExceptionVariable; // Variable that contains the saved $! variable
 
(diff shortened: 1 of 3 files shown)


kares:
hmm, so this works for the provided sample script above but not for the spec (essentially same script but within some describe/it mspec structure ... :blowfish:) : 
```
1)
Predefined global $! in bodies without ensure should be cleared when an exception is rescued even when a non-local return is present ERROR
Java::JavaLang::ArrayIndexOutOfBoundsException: 4
org.jruby.ir.operands.TemporaryLocalVariable.retrieve(TemporaryLocalVariable.java:74)
org.jruby.ir.instructions.PutGlobalVarInstr.interpret(PutGlobalVarInstr.java:70)
org.jruby.ir.interpreter.StartupInterpreterEngine.processOtherOp(StartupInterpreterEngine.java:178)
org.jruby.ir.interpreter.StartupInterpreterEngine.interpret(StartupInterpreterEngine.java:104)
org.jruby.ir.interpreter.Interpreter.INTERPRET_BLOCK(Interpreter.java:116)
org.jruby.runtime.MixedModeIRBlockBody.commonYieldPath(MixedModeIRBlockBody.java:136)
org.jruby.runtime.IRBlockBody.yieldSpecific(IRBlockBody.java:70)
org.jruby.runtime.Block.yieldSpecific(Block.java:153)
org.jruby.ir.runtime.IRRuntimeHelpers.yieldSpecific(IRRuntimeHelpers.java:503)
org.jruby.ir.instructions.YieldInstr.interpret(YieldInstr.java:76)
org.jruby.ir.interpreter.StartupInterpreterEngine.processOtherOp(StartupInterpreterEngine.java:178)
org.jruby.ir.interpreter.StartupInterpreterEngine.interpret(StartupInterpreterEngine.java:104)
org.jruby.ir.interpreter.InterpreterEngine.interpret(InterpreterEngine.java:86)
org.jruby.internal.runtime.methods.MixedModeIRMethod.INTERPRET_METHOD(MixedModeIRMethod.java:156)

```

PR #6058, opened 2020-02-14
Change triple to single backquotes

deivid-rodriguez:
I don't think triple backquotes are meant to be used inline, at least they broken syntax highlighting of my editor :)


headius:
Yeah overzealous quoting there. Thanks!

PR #6062, opened 2020-02-14
Mark custom classloaders as parallel capable

headius:
These classloaders apparently led to a deadlock inside Azul Zing
when using the ReadyNow feature that saves class and JIT data
betwen runs. This should be a benign change, since these loaders
typically only ever load one class, but we'll let it back for a
while.


headius:
I have also marked JRubyClassLoader as parallel capable, since all its internal state appears to be protected by synchronized sections.

PR #6067, opened 2020-02-15
optimize roundUpToPowerOfTwo

ahorek:
fewer lines of code + a little bit faster


headius:
The Java version of `numberOfLeadingZeros` has a similar implementation, but I suspect it's faster because it's also an "intrinsic" on Hotspot (meaning it will use whatever is optimal on each platform.

I'm :+1: on this.

headius:
Just under the wire for 9.2.10!

PR #6071, opened 2020-02-15
socket addrinfo fixes

ahorek:
Addrinfo.ip doesn't accept null chars
accept Addrinfo.ip("\<any\>")
protocol family should be set


enebo:
@ahorek thanks...you made it just under the wire!

Issue #6060, opened 2020-02-14
Globbed jar contents with expand path failing in RubyGems


headius:
In #5996 we discovered that a RubyGems update seems to break our ability to use a jar file as a GEM_PATH.

The failures look like this:

```
TestJarredGemsWithSpacesInDirectory: 
  test_jarred_gem_on_gem_path:					E
================================================================================
org/jruby/RubyKernel.java:1187:in `catch'
org/jruby/RubyKernel.java:1192:in `catch'
org/jruby/RubyKernel.java:1187:in `catch'
org/jruby/RubyKernel.java:1192:in `catch'
     26:     Gem.clear_paths
     27:     old = ENV['GEM_PATH']
     28:     ENV['GEM_PATH'] = File.expand_path('../gem.jar', __FILE__)
  => 29:     assert require('mygem')
     30:   ensure
     31:     ENV['GEM_PATH'] = old
     32:     Gem.clear_paths
/home/travis/build/jruby/jruby/test/jruby/test_jarred_gems.rb:29:in `test_jarred_gem_on_gem_path'
/home/travis/build/jruby/jruby/lib/ruby/stdlib/rubygems/core_ext/kernel_require.rb:54:in `require'
org/jruby/RubyKernel.java:977:in `require'
Error: test_jarred_gem_on_gem_path(TestJarredGemsWithSpacesInDirectory): LoadError: no such file to load -- mygem
================================================================================
: (0.018789)
  test_list_gem_from_jar_with_spaces_in_directory:		F
================================================================================
org/jruby/RubyKernel.java:1187:in `catch'
org/jruby/RubyKernel.java:1192:in `catch'
org/jruby/RubyKernel.java:1187:in `catch'
org/jruby/RubyKernel.java:1192:in `catch'
      6: 
      7:   def test_list_gem_from_jar_with_spaces_in_directory
      8:     out = jruby(%q{-r"./test/jruby/dir with spaces/testgem.jar" -S jgem list})
  =>  9:     assert(out =~ /testgem/)
     10: 
     11:     cp = ENV['CLASSPATH']
     12:     begin
/home/travis/build/jruby/jruby/test/jruby/test_jarred_gems.rb:9:in `test_list_gem_from_jar_with_spaces_in_directory'
Failure: test_list_gem_from_jar_with_spaces_in_directory(TestJarredGemsWithSpacesInDirectory): <nil> is not true.
================================================================================
```

[...]

PR #6085, opened 2020-02-19
Always use the filesystem encoding for expand_path output

headius:
This relates to the changes in #6060. We could think of no valid
reason for File.expand_path not to return strings encoded the
same as system file paths are encoded, but it's not clear if this
semantic change would cause any incompatibility.


headius:
The history behind this has gotten fuzzy and I do not recall why we felt the need to make this change. I'm just going to get rid of it, since much of this logic is based on CRuby and no clear reason was given why we should diverge.

PR #6095, opened 2020-02-23
C dash return

headius:
I'm as confused as you are. Apparently these two events use dashes, while the others have underscores. This failed the spec suite for https://github.com/rg-3/trip.rb because the specs wanted dashes.


headius:
Hmm ok something's off. The tests clearly seem to want the underscore name.

headius:
I've cherry-picked the test changes so they'll be active on jruby-9.2 and master, but it seems like the dash change was the wrong one. Needs deeper investigation.

Issue #6093, opened 2020-02-22
CGI.escapeHTML crashes on invalid byte sequence where CRuby does not


PragTob:
**Environment Information**
  
```
tobi@speedy:~/github/simplecov/tmp/aruba/project(convert-utf8)$ jruby -v
jruby 9.2.10.0 (2.5.7) 2020-02-18 fffffff OpenJDK 64-Bit Server VM 12.0.1+12 on 12.0.1+12 +jit [linux-x86_64]
tobi@speedy:~/github/simplecov/tmp/aruba/project(convert-utf8)$ uname -a
Linux speedy 5.3.0-40-generic #32~18.04.1-Ubuntu SMP Mon Feb 3 14:05:59 UTC 2020 x86_64 x86_64 x86_64 GNU/Linux
```

**Expected Behavior**

```
tobi@speedy:~/github/simplecov/tmp/aruba/project(convert-utf8)$ ruby -v
ruby 2.7.0p0 (2019-12-25 revision 647ee6f091) [x86_64-linux]
tobi@speedy:~/github/simplecov/tmp/aruba/project(convert-utf8)$ irb
irb(main):001:0> require 'cgi'
=> true
irb(main):002:0> CGI.escapeHTML "\xA4\xAA\xA4Ϥ褦"
=> "\xA4\xAA\xA4Ϥ褦"
irb(main):003:0> 
```

**Actual Behavior**

```
tobi@speedy:~/github/simplecov/tmp/aruba/project(convert-utf8)$ ruby -v
jruby 9.2.10.0 (2.5.7) 2020-02-18 fffffff OpenJDK 64-Bit Server VM 12.0.1+12 on 12.0.1+12 +jit [linux-x86_64]
tobi@speedy:~/github/simplecov/tmp/aruba/project(convert-utf8)$ irb
WARNING: An illegal reflective access operation has occurred
WARNING: Illegal reflective access by com.headius.backport9.modules.Modules to method sun.nio.ch.NativeThread.signal(long)
WARNING: Please consider reporting this to the maintainers of com.headius.backport9.modules.Modules
WARNING: Use --illegal-access=warn to enable warnings of further illegal reflective access operations
WARNING: All illegal access operations will be denied in a future release
irb(main):001:0> require 'cgi'
=> true
irb(main):002:0> CGI.escapeHTML "\xA4\xAA\xA4Ϥ褦"
Traceback (most recent call last):
[...]

PR #6097, opened 2020-02-25
CGI.escapeHTML restore the original encoding

ahorek:
fixes https://github.com/jruby/jruby/issues/6093


headius:
@enebo Perhaps this is safe for 9.2.11?

enebo:
Fixes something I know which happens in web apps (bogus encoded text being source from users).  Landing this seems like a net loss in new user reports from this behavioral difference with MRI. Seems reasonably safe too.

PR #6106, opened 2020-03-04
Ruby 2.6 : Enumerator::ArithmeticSequence class, Numeric#step, Range#step

k77ch7:
This PR implements Enumerator::ArithmeticSequence class , Numeric#step and Range#step.

See MRI's NEWS file https://github.com/ruby/ruby/blob/ruby_2_6/NEWS

### DIFF
--- a/core/src/main/java/org/jruby/RubyArithmeticSequence.java
+++ b/core/src/main/java/org/jruby/RubyArithmeticSequence.java
@@ -43,11 +43,14 @@ import org.jruby.util.Numeric;
 import static org.jruby.RubyEnumerator.enumeratorizeWithSize;
 import static org.jruby.RubyEnumerator.SizeFn;
 
+import static org.jruby.RubyNumeric.fixable;
+import static org.jruby.RubyNumeric.fix2long;
 import static org.jruby.RubyNumeric.floatStep;
 import static org.jruby.RubyNumeric.floatStepSize;
 import static org.jruby.RubyNumeric.dbl2num;
 import static org.jruby.RubyNumeric.int2fix;
 import static org.jruby.RubyNumeric.num2dbl;
+import static org.jruby.RubyNumeric.num2long;
 
 import static org.jruby.runtime.Helpers.hashEnd;
 import static org.jruby.runtime.Helpers.hashStart;
@@ -57,7 +60,7 @@ import static org.jruby.runtime.Helpers.safeHash;
 /**
  * Implements Enumerator::ArithmeticSequence
  */
-@JRubyClass(name = "Enumerator::ArithmeticSequence")
+@JRubyClass(name = "Enumerator::ArithmeticSequence", parent = "Enumerator")
 public class RubyArithmeticSequence extends RubyObject {
 
     private IRubyObject begin;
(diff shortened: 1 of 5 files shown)


headius:
Woohoo! Thank you!

core/src/main/java/org/jruby/RubyArithmeticSequence.java
```diff
@@ -0,0 +1,565 @@
...
+            return this;
+        }
+
+        if (end.isNil()) {
+            while (true) {
+                block.yield(context, c);
+                c = ((RubyNumeric)c).op_plus(context, s);
+            }
+            //FIXME : unreachable statement
+            //return this;
```
headius: Maybe just return this with an inline comment stating that it is never reached? There's not really any "fix" possible here since it's an infinite loop unless terminated by an exception.

core/src/main/java/org/jruby/RubyArithmeticSequence.java
```diff
@@ -0,0 +1,565 @@
...
+        }
+
+        if (Helpers.rbEqual(context, step, int2fix(context.runtime, 0)).isTrue()) {
+            while (true) {
+                block.yield(context, c);
+            }
+            //FIXME : unreachable statement
+            //return this;
```
headius: Same comment as above about infinite loop.

core/src/main/java/org/jruby/RubyArithmeticSequence.java
```diff
@@ -0,0 +1,565 @@
...
+            while (RubyNumeric.numFuncall(context, c, new GeCallSite(), last).isTrue()) {
+                block.yield(context, c);
+                c = ((RubyNumeric)c).op_plus(context, s);
+            }
+        } else {
+            while (RubyNumeric.numFuncall(context, last, new GeCallSite(), c).isTrue()) {
+                block.yield(context, c);
+                c = ((RubyNumeric)c).op_plus(context, s);
+            }
+        }
```
headius: These call sites should be cached in JavaSites. This logic will construct a new one every time which will be slower than doing a slow path invocation.

This should ideally be done for *all* dynamic calls from Java logic. There's several other places in this PR that could benefit.

See following for examples of how to define and access those sites:

Definition: https://github.com/jruby/jruby/blob/master/core/src/main/java/org/jruby/runtime/JavaSites.java

Easy access via a static method: https://github.com/jruby/jruby/blob/master/core/src/main/java/org/jruby/RubyString.java#L6673

Usage: https://github.com/jruby/jruby/blob/master/core/src/main/java/org/jruby/RubyString.java#L4022-L4032

core/src/main/java/org/jruby/RubyArithmeticSequence.java
```diff
@@ -0,0 +1,565 @@
...
+        Ruby runtime = context.runtime;
+        IRubyObject b = begin, e = end, s = step;
+        RubyArray ary;
+        long n;
+        boolean x;
+
+        if (num == null) {
+            if (!e.isNil()) {
+                IRubyObject zero = int2fix(runtime, 0);
+                int r = RubyComparable.cmpint(context, ((RubyNumeric)step).coerceCmp(context, context.sites.Numeric.op_cmp, zero), s, zero);
```
headius: Making a site local to this class would improve caching.

core/src/main/java/org/jruby/RubyArithmeticSequence.java
```diff
@@ -0,0 +1,565 @@
...
+                if (r > 0 && Helpers.invokePublic(context, b, ">", e).isTrue()) {
+                    return context.nil;
+                }
+
+                if (r < 0 && Helpers.invokePublic(context, b, "<", e).isTrue()) {
+                    return context.nil;
+                }
```
headius: Example of other places where ArithmeticSequence-specific JavaSites sites would be useful.

core/src/main/java/org/jruby/RubyArithmeticSequence.java
```diff
@@ -0,0 +1,565 @@
...
+
+                str = new ByteList(str, 0, str.length() - 2); /* drop the last ", " */
+                str.append(')');
+            }
+        }
+
+        str.append(')');
+
+        return RubyString.newStringLight(context.runtime, str);
```
headius: Probably needs a specific encoding? `new ByteList` defaults to `ASCII-8BIT` and I believe `inspect` is supposed to always be `US-ASCII`. If there's any characters not in that character range they'd need to be encoded.

k77ch7: I have confirmed same behavior as ruby 2.6.3. :confused:❓ 

```
% ruby -v
ruby 2.6.3p62 (2019-04-16 revision 67580) [x86_64-darwin19]
2.6.3 :001 > (1..).step.inspect.encoding
 => #<Encoding:ASCII-8BIT>
```

k77ch7:
@headius thank you for reviewing. I updated this PR.

core/src/main/java/org/jruby/RubyArithmeticSequence.java
headius: Ok, I stand corrected!

PR #6109, opened 2020-03-04
Lazy RubyGems

headius:
RubyGems by default will parse all default gemspecs and load a large amount of code that many simple command-line use cases will never need. This patch, based on code from TruffleRuby, uses autoload and other tricks to lazily initialize RubyGems only when a gem is actually required.

This speeds up startup by around 30% on my system (around 440 gems installed) but has some limitations:

* did_you_mean must be a default gem so that it can be loaded
  from stdlib without activating RubyGems.
* Because default gemspecs are not parsed at startup, it is not
  possible to upgrade any of the default gems in stdlib.
* There may be threading concerns with loading RubyGems lazily.

Results of this patch:

```
BEFORE:

[] ~/projects/jruby $ time jruby --dev -e 1

real	0m1.407s
user	0m1.883s
sys	0m0.193s

AFTER:

[] ~/projects/jruby $ time jruby --dev -e 1

real	0m1.047s
user	0m1.242s
sys	0m0.148s

did_you_mean still works:

[] ~/projects/jruby $ time jruby --dev -e putz
NameError: undefined local variable or method `putz' for main:Object
Did you mean?  puts
               putc
  <main> at -e:1

real	0m1.083s
user	0m1.346s
sys	0m0.144s
```


headius:
Apparently did_you_mean is now a default gem in Ruby 2.7, so the first limitation is moot. But it won't **act** like a default gem (can't be upgraded) with this patch in place.

headius:
Note that these limitations apply to current versions of TruffleRuby, i.e. no default gems can be upgraded without forcing RubyGems to load fully.

headius:
Note: MRI tests run with `--disable-gems` by default, so we can't infer from them whether this is a safe change.

The other failures seem mostly to be from tests that expect RubyGems constants to be present or otherwise confirm RubyGems has booted before proceeding.

headius:
TruffleRuby has added code that uses a list of default gems and their sources to activate RubyGems lazily while still supporting default gem upgrading. This code has not been contributed back to RubyGems or to CRuby, but should work fine under JRuby.

headius:
The commit in f28245b seems to be causing an early access Java 16 to crash with the following output, after generating the CDS archive using the `generate_appcds` command in the jruby-startup gem:

https://gist.github.com/headius/60ecfc3e120e10e8a3e1f801539520e6

If I comment out the gathering of options into `ALL_OPTIONS` and hardcode the `lazy_rubygems` check to `true` it no longer crashes, but it works fine without CDS or with CDS on Java 15.

headius:
Two remaining issues may doom this from working properly.

The first, #6506, shows an issue with IRB attempting to access features from the `Gem` module lazily. For some reason, just accessing the `Gem` constant is not triggering a full load of RubyGems, resulting in a method IRB needs not being found. I have worked around that issue by having `Binding#irb` eagerly load RubyGems, but it is still an open issue.

The second is the cause of the remaining failure here during the build of the jruby-jars gem. One of the files loaded to verify that artifact is the `maven/tools/pom.rb` file from the ruby-tools Java library. That file needs RubyGems to function, but instead of requiring `rubygems` it starts by loading `rubygems/specification`, and that is not sufficient to get the `Gem::VERSION` constant defined. The resulting error looks like this:

```
NameError: uninitialized constant Gem::VERSION
Did you mean?  Gem::Version
          const_missing at org/jruby/RubyModule.java:3899
  <class:Specification> at uri:classloader:/META-INF/jruby.home/lib/ruby/stdlib/rubygems/specification.rb:157
                 <main> at uri:classloader:/META-INF/jruby.home/lib/ruby/stdlib/rubygems/specification.rb:41
                require at org/jruby/RubyKernel.java:1014
                require at uri:classloader:/jruby/kernel/lazy_rubygems.rb:118
                 <main> at /Users/headius/.m2/repository/de/saumya/mojo/ruby-tools/1.0.7/ruby-tools-1.0.7.jar!/maven/tools/pom.rb:26
                require at org/jruby/RubyKernel.java:1014
                require at uri:classloader:/jruby/kernel/lazy_rubygems.rb:118
                 <main> at -e:1
```

I am working on a hack for the second item, but it may be a symptom of a larger problem: namely that there are many libraries that do not work with a partially-defined `Gem` module.

iklam:
> The commit in [f28245b](https://github.com/jruby/jruby/commit/f28245bf0ba4ecdb17675ef37910294e32bf4757) seems to be causing an early access Java 16 to crash with the following output, after generating the CDS archive using the `generate_appcds` command in the jruby-startup gem:
> 
> https://gist.github.com/headius/60ecfc3e120e10e8a3e1f801539520e6

I created the following issue for tracking this in OpenJDK: https://bugs.openjdk.java.net/browse/JDK-8259275

@headius could you help us reproduce the issue, ideally with the exact JRuby binaries that you used?

headius:
@iklam I will try to get together a commit hash + build instructions that crashes.

headius:
@iklam I posted a comment on the JDK issue with steps to reproduce. Please let me know if you are unable to trigger the crash!

iklam:
> @iklam I posted a comment on the JDK issue with steps to reproduce. Please let me know if you are unable to trigger the crash!

We are able to reproduce the crash and are working on fixing it. Thanks for the bug report.

PR #6123, opened 2020-03-16
Fixes include_package to work from any (sub)module or (sub)class

AndyObtiva:
[Ready for review]

While recently working on my project Glimmer (https://github.com/AndyObtiva/glimmer), I noticed that include_package does not work from everywhere. I implemented a fix to make include_package work from any (sub)module or (sub)class.

### DIFF
--- a/core/src/main/ruby/jruby/java/core_ext/module.rb
+++ b/core/src/main/ruby/jruby/java/core_ext/module.rb
@@ -1,37 +1,5 @@
 # Extensions to the standard Module package.
 class Module
-  alias const_missing_without_jruby const_missing
-  def const_missing(constant)
-    all_included_packages = instance_exec(&JRUBY_PRIVATE.method(:included_packages_from_ancestors_namespaces))
-    return const_missing_without_jruby(constant) if all_included_packages.empty?
-    real_name = instance_exec(&JRUBY_PRIVATE.method(:java_aliases_from_ancestors_namespaces))[constant] || constant
-
-    java_class = nil
-    last_error = nil
-
-    all_included_packages.each do |package|
-      begin
-        java_class = JavaUtilities.get_java_class("#{package}.#{real_name}")
-      rescue NameError => e
-        # we only rescue NameError, since other errors should bubble out
-        last_error = e
-      end
-      break if java_class
-    end
-
-    if java_class
-      return JavaUtilities.create_proxy_class(constant, java_class, self)
-    else
-      # try to chain to super's const_missing
-      begin
-        return const_missing_without_jruby(constant)
-      rescue NameError => e
-        # super didn't find anything either, raise our Java error
-        raise NameError.new("#{constant} not found in packages #{all_included_packages.join(', ')}; last error: #{(last_error || e).message}")
-      end
-    end
-  end
-
   private
 
   module JRUBY_PRIVATE
@@ -143,6 +111,44 @@ class Module
     package = package.package_name if package.respond_to?(:package_name)
     the_included_packages = instance_exec(&JRUBY_PRIVATE.method(:included_packages))
     the_included_packages << package unless the_included_packages.include?(package)
+    method(:const_missing_without_jruby)
+  rescue NameError => e
+    if e.message.start_with?("undefined method `const_missing_without_jruby'")
+      self.class.class_eval do
+        alias const_missing_without_jruby const_missing
+        def const_missing(constant)
+          all_included_packages = instance_exec(&JRUBY_PRIVATE.method(:included_packages_from_ancestors_namespaces))
+          return const_missing_without_jruby(constant) if all_included_packages.empty?
+          real_name = instance_exec(&JRUBY_PRIVATE.method(:java_aliases_from_ancestors_namespaces))[constant] || constant
+
+          java_class = nil
+          last_error = nil
+
+          all_included_packages.each do |package|
+            begin
+              java_class = JavaUtilities.get_java_class("#{package}.#{real_name}")
+            rescue NameError => e
+              # we only rescue NameError, since other errors should bubble out
+              last_error = e
+            end
+            break if java_class
+          end
+
+          if java_class
+            return JavaUtilities.create_proxy_class(constant, java_class, self)
+          else
+            # try to chain to super's const_missing
+            begin
+              return const_missing_without_jruby(constant)
+            rescue NameError => e
+              # super didn't find anything either, raise our Java error
+              raise NameError.new("#{constant} not found in packages #{all_included_packages.join(', ')}; last error: #{(last_error || e).message}")
+            end
+          end
+        end
+      end
+    end
+  ensure
     nil
   end
 
--- a/core/src/main/ruby/jruby/java/core_ext/module.rb
+++ b/core/src/main/ruby/jruby/java/core_ext/module.rb
@@ -5,8 +5,14 @@ class Module
   end
 
   def included_packages_from_namespaces
-    namespaces.uniq.map do |klass|
+    namespaces.map do |klass|
       klass.method(:included_packages).call rescue []
+    end.map(&:to_a).reduce([], :+).uniq
+  end
+
+  def included_packages_from_ancestors
+    ancestors.map do |klass|
+      klass.method(:included_packages_from_namespaces).call rescue []
     end.map(&:to_a).reduce([], :+).uniq
   end
 
@@ -15,8 +21,14 @@ class Module
   end
 
   def java_aliases_from_namespaces
-    namespaces.uniq.map do |klass|
+    namespaces.map do |klass|
       klass.method(:java_aliases).call rescue {}
+    end.reverse.reduce({}, :merge)
+  end
+
+  def java_aliases_from_ancestors
+    ancestors.map do |klass|
+      klass.method(:java_aliases_from_namespaces).call rescue {}
     end.reverse.reduce({}, :merge)
   end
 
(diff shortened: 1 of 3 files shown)


kares:
Thanks Andy, could you show us an example piece of reproducer that you're trying to get working?
The amount of method "pollution" in Module seems to have increased quite a lot and its hard to justify wout any tests/examples.

AndyObtiva:
Thanks for your feedback.

I just added specs for namespaced nested use of include_package to demonstrate cases that used to fail before.

I'll look into reducing method pollution.

Otherwise, I think I just thought of a few more cases I didn't handle. Please give me some time to implement them, test-first this time.

AndyObtiva:
Below are the new cases I thought of and implemented.

About method pollution, I made all new methods I added to `Module` private. Is that good enough or would you rather we have less methods while accepting the trade-offs with maintainability, readability, and understandability? I'd like to do what's best if possible.

**Specs:**

```ruby
  it "can be imported from a superclass using 'include_package package.module" do
    class SuperClass
      include_package java.lang
      class AnotherSuperClass
      end
    end
    class SubClass < SuperClass
      include_package java.util
    end
    class AnotherSubClass < SuperClass::AnotherSuperClass
    end
    class SubSubClass < SubClass
    end
    expect(SuperClass::System).to respond_to 'getProperty'
    expect(SubClass::System).to respond_to 'getProperty'
    expect(AnotherSubClass::System).to respond_to 'getProperty'
    expect(SubSubClass::System).to respond_to 'getProperty'
    expect(SubSubClass::Arrays).to respond_to 'asList'
  end

  it "can be imported from a supermodule using 'include_package package.module" do
    module SuperModule
      include_package java.lang
      module AnotherSuperModule
      end
    end
    class SuperClass
      include SuperModule
    end
    class SubClass < SuperClass
    end
    class AnotherSuperClass
      include SuperClass::AnotherSuperModule
    end
    class AnotherSubClass < AnotherSuperClass
    end
    module SubModule
      include SuperModule
      include_package java.util
    end
    module SubSubModule
      include SubModule
    end
    expect(SuperModule::System).to respond_to 'getProperty'
    expect(SuperClass::System).to respond_to 'getProperty'
    expect(SubClass::System).to respond_to 'getProperty'
    expect(AnotherSuperClass::System).to respond_to 'getProperty'
    expect(AnotherSubClass::System).to respond_to 'getProperty'
    expect(SubModule::System).to respond_to 'getProperty'
    expect(SubSubModule::Arrays).to respond_to 'asList'
  end
```

headius:
I can appreciate what you're doing but this feels like, to use the vernacular, "pouring perfume on a pig". The `include_package` functionality is problematic for many reasons, mainly because it has to rely on the `const_missing` mechanism which is itself rather problematic:

The main problem here is that `const_missing` only gets called on the immediate target of the lookup, which means it does not cascade to containing namespaces or superclasses. As a result you must include every package you want to access into every namespace that accesses classes from it.

This might be doable but the patch as written is too invasive (as @kares mentioned).

Some possible improvements...

Module nesting
---------------

Your patch modifies the `include_package` `const_missing` behavior to also search namespaces that wrap the current one by adding a `namespaces` feature.

I think what you actually want here is `Module.nesting`:

```
$ jruby -e "module A; module B; module C; p Module.nesting; end; end; end"
[A::B::C, A::B, A]
```

That would eliminate all the nested namespace management from your patch. A similar change could also search parent classes for included packages.

Overriding default Module#const_missing
------------------------------------------

I don't think this change can go in. Overriding the default const_missing in this way almost certainly interfere with other systems that try to hook missing constants. We'll be fighting over who gets to be the default const_missing. It also causes all const_missing dispatches to walk nested namespaces and build a list of included packages.

For now we should continue to install the specialized const_missing only packages that request it (i.e. lazily decorate only that specific module with this logic).

headius:
BTW, thanks for continuing to work on Glimmer... I have been seeing your steady stream of releases latelt!

headius:
Also...

> would you rather we have less methods

Yes, this :-)

I think a guiding principal here should be that modules untouched by Java integration look exactly like they do in CRuby. We should *only* add this extra logic when requested, and should minimize how much we add as much as possible.

test/mri/ruby/test_module.rb
```diff
@@ -588,7 +588,7 @@ def test_s_constants
     c2 = Module.constants
     assert_equal([:WALTER], c2 - c1)
 
-    assert_equal([], Module.constants(true))
+    assert_equal([:JRUBY_PRIVATE], Module.constants(true))
```
kares: this is actually something that shows the implementation is leaking into all modules.
understand it might be demotivating but this raises a :red_circle: for the proposed changes ...

AndyObtiva: Since the tests aren't passing yet, this obviously isn't ready for review though.

kares: Hey Andy, what I meant is what was mentioned in comments already.
The changes should NOT affect plain-old Ruby modules in any way ...
Meaning all existing Ruby tests/specs should pass as before (wout changes).

Will hold off any reviews, was just a hint to not chase a dead end potentially.

AndyObtiva:
@headius Thank you very much for the feedback. JRuby really has everything I need, and any `include_package` improvements are only "nice to have". This pull request is just a fun learning experience for me while hopefully contributing something helpful to others.

In case you'd change your mind, I have a solution that avoids method/constant pollution, and restricts `const_missing` definition to `include_package` just like before. The main catch is that the code is very ugly. 

One thing I am concerned about though, which I am sure you're all concerned about as well, is backwards compatibility with previous versions of JRuby to avoid breaking existing codebases. As a result, I plan to undo all my code from `include_package`, restoring to how it originally was, and then add a new `include_package_in_scope` instead, which implies clearly that the package included will be available to the scope of the class/module, that is all contained classes/modules and subclasses/submodules.

I will report back once I have `include_package_in_scope` ready in case there is any interest in it. If not, no hurt feelings. I am only trying to help, and if I am not helping with my contribution, there is no point in it.

Sincerely,

Andy 

p.s. By the way, I didn't know about Module.nesting, and yet it doesn't work for my case because its documentation says it returns module nesting from the point of call, meaning if namespaces have been declaring already, and I have the nested class/module, I couldn't use Module.nesting anymore. Only in the declaration of the class can I use it. In any case, I learned something new. Thanks.

AndyObtiva:
OK, on further experimentation and thinking, it seems like the idea of an additional `include_package_in_scope` is not feasible because its differing `const_missing` implementation conflicts with that of `include_package`. As such, I am ditching. 

An enhanced `include_package` is simpler from a code design point of view provided it does not break any existing codebases. Also, it does alias `const_missing` to ensure that it does not conflict with any existing uses of `const_missing` in a codebase. Perhaps it's a good idea to try beta-testing the enhanced `include_package`, and if the results turn out negative, revert it back to the original `include_package`.

Otherwise, if you'd rather avoid all together, I can understand given the complications. Thanks for giving me the chance to contribute anyways even if nothing comes to fruition out of this.

Best regards,

Andy

enebo:
@AndyObtiva Another idea would be to rolling these changes into a ruby gem.  We are more than happy to have an addition on the calling java from ruby wiki page linking to a gem which provides this top-level behaviour.   As far as breaking any old behavior a gem would insulate from that since it would be opt-in behaviour.  It just may end up as an issue in your gem...

headius:
@AndyObtiva Thanks very much for your exploration on this. At the very least we'll have this issue to point at to show some reasons why we have not made this change sooner.

The `const_missing` behavior change was what concerned me most. We've had issues with `include_package` and other libraries' `const_missing` implementations stepping on each other, even though we're largely using it as designed. You only get one `const_missing`, and it's devilishly hard to get multiple libraries to cooperate.

I did have a realization last night that we *could* implement this feature as a part of the constant lookup logic we run for Ruby. Basically, it would add an additional check at each level of the search that would use any included packages to do another attempt to find the constant.

Functionally, it would look as those all the classes in a given package were actually defined as constants; the package would just be another way constants might be "stored" on a module.

Practically, it might slow down constant missing a bit in the best case (no packages included) and slow it down greatly in the worst case (when there's one or more packages included).

In any case, it's a non-trivial addition and would probably be simplest as an "always-on" feature.

@enebo or @kares might have some thoughts on this approach, but it's not something we'd attempt to do before JRuby 9.3 in any case.

@AndyObtiva If you'd like to keep exploring, this might be a more acceptable implementation, since we can keep all the logic on the native (Java) side, avoid polluting namespaces or altering `const_missing` behavior, and support your intended feature in a simpler way. We may end up deciding it's too invasive, but working code could change that.

headius:
FWIW I also support the idea of turning your Ruby implementation into a gem. It would allow you to opt-into the feature and we can continue exploring whether something similar would make sense in JRuby proper.

AndyObtiva:
Sounds good. 

Here is the [gem](https://rubygems.org/gems/nested_inherited_jruby_include_package):
https://github.com/AndyObtiva/nested_inherited_jruby_include_package

I look forward to new developments in JRuby.

Cheers and thanks all for your time.

headius:
@AndyObtiva Epic gem name! ❤️

AndyObtiva:
Thank you. I figured if I make it generic enough, people will know it’s just a temporary fix and won’t be too attached to it until there is a built in solution in JRuby. :) Been using on a couple of projects. So far so good though I fixed two bugs in it. Also, I added to the JRuby WIKI under Calling Java from JRuby: https://github.com/jruby/jruby/wiki/CallingJavaFromJRuby

I recently added MacOS DMG bundle packaging support to Glimmer via a rake task (https://andymaleh.blogspot.com/2020/04/glimmer-051-packaging-glimmer-apps-as.html?m=1), and thought the combination of Warbler and javapackager was great. Light years ahead of some Java packaging work I was involved with as a junior dev early in my career.

PR #6133, opened 2020-03-24
Enhanced exit blocks

headius:
This adds the ability to register an exit hook from Java using a simple function object. It also switches the Timeout library from using `at_exit` to using this API for executor shutdown.

See #6127 which, due to the lack of such an API, was forced to dig around in JRuby internals from Ruby code.

There may be other places where this could be used but I did not audit other `at_exit` consumers.


kares:
:+1: somehow the name confused me `pushExitFunction` evoked a FIFO order assumption.
maybe `addExitFunction` sounds clearer but not strong feelings about naming ...

headius:
I went with that name mostly because we already had `pushExitProc` and it seemed to make sense that way. `push` fits FILO ordering to me as well because it implies a stack, no?

PR #6135, opened 2020-03-24
Bump `json` version provided by default

deivid-rodriguez:
The 2.2.0 version gemspec provided by prints an annoying warning when read:

```
NOTE: Gem::Specification#rubyforge_project= is deprecated with no replacement. It will be removed on or after 2019-12-01.
Gem::Specification#rubyforge_project= called from /home/deivid/.rbenv/versions/jruby-9.2.11.0/lib/ruby/gems/shared/specifications/default/json-2.2.0-java.gemspec:18.
```


headius:
Thank you for the PR! We will update all stdlibs for JRuby 9.3, but it seems this one has not gotten updated so far.

PR #6136, opened 2020-03-24
Fix another unintended triple backquote

deivid-rodriguez:
This is another overzealous quoting I found in docs. Same as https://github.com/jruby/jruby/pull/6058.


headius:
Thank you!

Issue #6137, opened 2020-03-24
`sprintf` adds extra buffer slots when precision arg truncates string slices


Magisus:
**Environment Information**
* JRuby version: 9.2.7.0+
* Operating system and platform: All

**Expected Behavior**

The `sprintf` function should correctly truncate string slices when a precision is specified that is smaller than the slice length.

**Actual Behavior**
The `sprintf` function misbehaves when using a precision number with a string argument, if that string is a slice of a larger string. It will add the wrong number of spaces to the buffer and then fill it with extra characters from the larger string, or leave the buffer spaces uninitialized if there are too few characters in the source string to fill it.

Script to repro:

```
long_string = "aabbccddhelloddccbbaa"
start_index = 8
sub_str_length = 5
precision = 3
sub_string = long_string[start_index, sub_str_length]
puts sprintf("%.#{precision}s", sub_string)
# => helloddccbb
# When calculating the precision, it adds (start_index (8) + precision (3) = 11) slots to the output buffer,
# then fills it with that many characters from the original larger string.
# If there are not enough characters to fill the number of slots that were added,
# it will output gibberish in the remaining slots (I assume these are just uninitialized bits)
```
I was able to fix this bug by adding this simple patch at https://github.com/jruby/jruby/blob/master/core/src/main/java/org/jruby/util/Sprintf.java#L622:
```
  len = StringSupport.nth(enc, bytes.getUnsafeBytes(), bytes.begin(), bytes.begin() + bytes.getRealSize(), precision);
+ len = len - bytes.begin(); // subtract the starting index to get the number of bytes to add
```

headius:
cc @kares since I think the original commit was your work.

We do not really consider this a general security issue, but given the impact to Puppet we will put out a 9.2.11.1 release this week.

kares:
for the record, anyone reaching here this only concerns: `%s` with precision e.g. `%.1s`
https://github.com/jruby/jruby/commit/bb90d3b7644316f8ae6b92e02defdf3838854fb5 introduced the issue (while attempting to fix multi-byte precision in `%s`).

PR #6138, opened 2020-03-24
Subtract begin offset for partial substring in sprintf 's'

headius:
See #6137 for the original bug report. This PR includes the change proposed there along with some new doco for `nth` and a spec.


headius:
Moving forward with this since we want to put a release out tomorrow.

PR #6141, opened 2020-03-26
Tie become_java! and concrete extension (fixes #4165)

### LINKED ISSUE #4165: java_field causes NPE when using derived class of a java object
### Environment
- jruby 9.1.5.0 (2.3.1) 2016-09-07 036ce39 Java HotSpot(TM) 64-Bit Server VM 25.101-b13 on 1.8.0_101-b13 +jit [linux-x86_64]
### Test Case

``` rb
require 'jruby/core_ext'
class A < Java::javafx::scene::layout::BorderPane
  java_field "java.lang.Object label"
end
A.become_java!
```

I would expect this to succeed, as it does if you swap the parent with a ruby class
### Unexpected Error

```
Unhandled Java exception: java.lang.NullPointerException
java.lang.NullPointerException: null
      generateFieldAccessors at org/jruby/java/addons/ClassJavaAddons.java:77
                  becomeJava at org/jruby/java/addons/ClassJavaAddons.java:69
                 become_java at org/jruby/java/addons/ClassJavaAddons.java:40
                        call at org/jruby/java/addons/ClassJavaAddons$INVOKER$s$become_java.gen:-1
                cacheAndCall at org/jruby/runtime/callsite/CachingCallSite.java:318
                        call at org/jruby/runtime/callsite/CachingCallSite.java:131
  invokeOther16:become_java! at jrubyfx-fxmlloader2.rb:4
                      <main> at jrubyfx-fxmlloader2.rb:4
                        load at org/jruby/ir/Compiler.java:111
                   runScript at org/jruby/Ruby.java:834
                 runNormally at org/jruby/Ruby.java:749
                 runNormally at org/jruby/Ruby.java:767
                 runFromMain at org/jruby/Ruby.java:580
               doRunFromMain at org/jruby/Main.java:425
                 internalRun at org/jruby/Main.java:313
                         run at org/jruby/Main.java:242
                        main at org/jruby/Main.java:204
```

byteit101:
Still in progress, but allow concrete class extension to define non-extended fields and signatures (required for jrubyfx to use the built in FXMLLoader that uses reflection on passed in classes)

Notable questions:

- [ ] ClassJavaAddons: how to check for java proxies vs failed reification
- [x] Is adding two arguments everywhere the best way to send them down to the asm layer?
- [x] MethodData generalization: are the LocalMethodData constants reasonable?
- [x] MethodData.getDeclaringClass ???


headius:
Putting this on my todo list to review... perhaps early next week? Maybe you have some thoughts @kares?

byteit101:
While investigating annotations, it appears that jrubyc and become_java! have different syntaxes:

```rb
java_annotation 'Deprecated'
java_signature 'void my_method()'
```
vs
```rb
java_signature '@Deprecated void my_method()'
```

jrubyc uses at least the former and become_java! only saves the latter. As such, I've implemented parameters for the latter, but that brings up what to do with fields. Should `java_field` support annotation parsing or should `java_annotation` be used instead, and probably expanded for methods too?

byteit101:
As per conversation on matrix with @enebo, This PR now has the following API changes:
`java_annotation` is deprecated and warns on useless calls. https://github.com/jruby/jruby/wiki/JRuby-Reference#java_annotation should be updated
`java_signature` now supports annotations that have parameters (previously only supported unnamed parameters)
`java_field` now supports annotations (previously didn't support annotations)

lib/ruby/stdlib/jruby/core_ext/class.rb
```diff
@@ -153,22 +153,31 @@ def add_method_signature(name, classes)
...
+    add_field_signature signature.name, signature.return_type
 
-    type, name = signature
-    add_field_signature(name, type)
+    annotations = signature.annotations
+    add_field_annotation signature.name, annotations if annotations
   end
 
+  def java_annotation(anno)
+    STDERR.puts "java_annotation is deprecated. Use java_signature '@#{anno} ...' instead. Called from: #{caller.first}"
```
kares, line 173: minor: we probably just want to use `warn` or some kind of warnings on the native end.

kares:
Great work so far Patrick!

Think it makes sense to have as its going, I did like the previous `java_annotation 'Deprecated'` wout the `@` prefix but for `java_field` etc it's more reasonable to have Java-like syntax `@Deprecated ...`.

Not sure if current scope imports are respected (or whether they worked using `java_annotation`) but it would be nice to not have to always use fully-qualified Java class names in annotations.

core/src/main/java/org/jruby/ast/java_signature/AnnotationVisitor.java
```diff
@@ -0,0 +1,54 @@
...
+ * Software distributed under the License is distributed on an "AS
+ * IS" basis, WITHOUT WARRANTY OF ANY KIND, either express or
+ * implied. See the License for the specific language governing
+ * rights and limitations under the License.
+ *
+ * Copyright (C) 2001-2002 Benoit Cerrina <b.cerrina@wanadoo.fr>
+ * Copyright (C) 2002 Jan Arne Petersen <jpetersen@uni-bonn.de>
+ * Copyright (C) 2002-2004 Anders Bengtsson <ndrsbngtssn@yahoo.se>
+ * Copyright (C) 2004 Thomas E Enebo <enebo@acm.org>
+ * Copyright (C) 2004 Stefan Matthias Aust <sma@3plus4.de>
```
headius, line 18: Remove attributions unrelated to this file. If you really want to add yourself, go ahead, but since the audit trail is in git I don't believe we need to have these.

core/src/main/java/org/jruby/ast/java_signature/BoolLiteral.java
```diff
@@ -0,0 +1,36 @@
+/*
+ * To change this template, choose Tools | Templates
+ * and open the template in the editor.
+ */
```
headius, line 4: Fix license header

core/src/main/java/org/jruby/ast/java_signature/NumberLiteral.java
```diff
@@ -0,0 +1,40 @@
+/*
+ * To change this template, choose Tools | Templates
+ * and open the template in the editor.
+ */
```
headius, line 4: Fix license header

core/src/main/java/org/jruby/ast/java_signature/NumberLiteral.java
```diff
@@ -0,0 +1,40 @@
...
+package org.jruby.ast.java_signature;
+
+/**
+ *
+ * @author enebo
```
headius, line 9: Author @enebo?

core/src/main/java/org/jruby/javasupport/proxy/JavaProxyClassFactory.java
```diff
@@ -155,15 +160,43 @@ public JavaProxyClass newProxyClass(final Ruby runtime, ClassDefiningClassLoader
...
+        Map<String, Map<Class, Map<String, Object>>> fieldannos;
+        Map<String, Class> fields;
+        if (rclass != null)
+        {
+        	fieldannos = rclass.getFieldAnnotations();
+        	fields = rclass.getFieldSignatures();
+        	
+        }
+        else 
+        {
```
headius, line 182: Style issues here and other places... close brace and open brace should be on same line as else, if's open brace should be on same line, etc.

lib/ruby/stdlib/jruby/core_ext/class.rb
headius: Yeah use Kernel#warn or Warnings#warn so it can be silenced if necessary.

headius:
Addressing the original 4 question bullets:

- ClassJavaAddons: how to check for java proxies vs failed reification

I'm not sure exactly what you mean. What you have seems fine to me, maybe?

- Is adding two arguments everywhere the best way to send them down to the asm layer?

Usually if I feel like I'm passing around the same arguments too many places, it might be time for a data object that contains all of them together. It's not a huge deal either way for this PR, but perhaps we can refactor this to be more OO and not have to pass so many params all over.

- MethodData generalization: are the LocalMethodData constants reasonable?

I think these are fine as-is.

- MethodData.getDeclaringClass ???

If this is no longer being used then it's fine to remove.

core/src/main/java/org/jruby/java/addons/ClassJavaAddons.java
```diff
@@ -65,7 +66,13 @@ private static IRubyObject becomeJava(final ThreadContext context, final RubyCla
...
         klass.reifyWithAncestors(dumpDir, useChildLoader);
 
-        final Class<?> reifiedClass = klass.getReifiedClass();
+        Class<?> reifiedClass = klass.getReifiedClass();
+        if (reifiedClass == null) { // java proxies can't be reified, but they deserve field accessors too
+            reifiedClass = JavaProxyClass.getProxyClass(context.getRuntime(), klass).getJavaClass();
+        }
+        if (reifiedClass == null) {
+            throw context.runtime.newTypeError("requested class " + klass.getName() + " was not reifiable");
+        }
```
byteit101, line 75: @headius This bit of code does not distinguish between Failed Reified Ruby classes and Concretized Classes. getReified is null if either it's a failure to reify or a java proxy class. Should the logic be split between those two cases, and if so, how?

headius: Aha, so you think maybe we need another way to know a given class is a java proxy class. Perhaps we add another boolean flag to RubyClass like `isJavaSubclass` or `isJavaProxy`?

byteit101: >  so you think maybe we need...
No, I am asking if we do. I have no strong opinion on these internals, but I do think the error message could be a misleading if we care to tell users about the difference between concrete vs reified, so I'm wondering if *you* think we should disambiguate. 
Secondarily, I think it's safe for this code path to be retried for failed reified and failed concrete multiple times, but that might have implications I'm not aware of

byteit101:
> Usually if I feel like I'm passing around the same arguments too many places, it might be time for a data object that contains all of them together. It's not a huge deal either way for this PR, but perhaps we can refactor this to be more OO and not have to pass so many params all over.

The reason I hesitated was because all of the items came from the ruby class, and so the obvious thing was to just pass the ruby class down, but I wasn't sure if that would break existing encapsulation design too much. Creating a small mirror of RubyClass things also seemed kind of silly if we could just use RubyClass

headius:
I am fine merging this once review comments are addressed and CI is green. Thank you for your help and patience @byteit101!

headius:
This might help the failures (annotation API is no longer part of JDK):

https://gist.github.com/headius/cde16ee9fe14767549ed0036ec0a04dc

Issue #6142, opened 2020-03-27
Thread didn't execute?


ioquatix:
```
koyoko% jruby -v             
jruby 9.2.9.0 (2.5.7) 2019-10-30 458ad3e OpenJDK 64-Bit Server VM 25.242-b08 on 1.8.0_242-b08 +jit [linux-x86_64]
koyoko% uname -a
Linux koyoko.local 5.5.9-arch1-2 #1 SMP PREEMPT Thu, 12 Mar 2020 23:01:33 +0000 x86_64 GNU/Linux
```

In my live stream, talking about thread safety no less, I encountered some kind of issue in JRuby when using threads in code that should have been fine.

https://youtu.be/Wu9LRNOc5pQ?t=630

It appears that while we did construct a thread, it never actually executed any code.

Here is the code from the live stream above:

```ruby
#!/usr/bin/env ruby

def test(repeats = 1000)
	mutex = Mutex.new
	
	mutexes = {}
	threads = []
	
	threads = repeats.times.map do |i|
		Thread.new do
			mutex.synchronize do
				mutexes[i] ||= Mutex.new
			end
		end
	end
	
	threads.each(&:join)
	
	if threads.size != mutexes.size
		puts "There were #{threads.size} threads and #{mutexes.size} mutexes."
		
		repeats.times do |i|
			unless mutexes.include?(i)
				puts "Missing #{i}!"
			end
		end
	end
	
	missing = repeats.times.to_a - mutexes.keys
	
	unless missing.empty?
		puts "Missing: #{missing.inspect}"
	end
end

100.times{test}
```

I have been discussing the issue with @headius and he suggested it might be a problem with the native thread being GCed before starting:

```java
    public void start() {
        Thread thread = getThread();
        if (thread != null) thread.start();
        else throw new RuntimeException("WTF");
    }
```

and he was able to get it to fail by increasing GC frequency.

```
[] ~/projects/jruby $ jruby thread_mutexes.rb 
Unhandled Java exception: java.lang.RuntimeException: WTF
```

However I am not able to reproduce the issue, it seems very infrequent.

headius:
Yeah my theory is just a theory, but it seems the most likely candidate right now.

In NativeThread, we hold a weak reference to the actual JVM Thread object. I believe this is getting collected at just the right moment in the video link provided (at 10:34) such that it silently fails to run the thread's body.

[...]

PR #6143, opened 2020-03-27
Hold a hard reference to Ruby threads

headius:
The old logic maintained a weak reference to the native thread
associated with a Ruby thread, in order to avoid keeping its
resources alive longer than the thread's lifecycle. However in
the case shown in #6142, it seems likely that under heavy load the
native thread gets collected before it starts running. Because of
the logic we have to silently ignore collected references, the
NativeThread.start method simply returns if the thread has gone
away, giving us no indication that the thread never actually ran.

This patch splits our NativeThread into two types:

* RubyNativeThread, which holds a hard reference to the native
  Thread and implements start to call Thread#start.
* AdoptedNativeThread, which holds a weak reference to the native
  Thread and errors (BUG!) if we attempt to call call start on
  that adopted thread.

I believe this will fix the issue in #6142. This may explain other
"one in a million" failures we have never quite tracked down.

### DIFF
--- a/core/src/main/java/org/jruby/RubyThread.java
+++ b/core/src/main/java/org/jruby/RubyThread.java
@@ -61,10 +61,11 @@ import org.jruby.exceptions.RaiseException;
 import org.jruby.exceptions.ThreadKill;
 import org.jruby.exceptions.Unrescuable;
 import org.jruby.ext.thread.Mutex;
-import org.jruby.internal.runtime.NativeThread;
+import org.jruby.internal.runtime.RubyNativeThread;
 import org.jruby.internal.runtime.RubyRunnable;
 import org.jruby.internal.runtime.ThreadLike;
 import org.jruby.internal.runtime.ThreadService;
+import org.jruby.internal.runtime.AdoptedNativeThread;
 import org.jruby.java.proxies.ConcreteJavaProxy;
 import org.jruby.javasupport.JavaUtil;
 import org.jruby.runtime.Arity;
@@ -447,7 +448,7 @@ public class RubyThread extends RubyObject implements ExecutionContext {
         RubyThread rubyThread = new RubyThread(runtime, threadClass, true);
 
         // TODO: need to isolate the "current" thread from class creation
-        rubyThread.threadImpl = new NativeThread(rubyThread, Thread.currentThread());
+        rubyThread.threadImpl = new AdoptedNativeThread(rubyThread, Thread.currentThread());
         runtime.getThreadService().setMainThread(Thread.currentThread(), rubyThread);
 
         // set to default thread group
@@ -592,7 +593,7 @@ public class RubyThread extends RubyObject implements ExecutionContext {
                                           final RubyClass recv, final Thread thread) {
         final RubyThread rubyThread = new RubyThread(runtime, recv, true);
 
-        rubyThread.threadImpl = new NativeThread(rubyThread, thread);
+        rubyThread.threadImpl = new AdoptedNativeThread(rubyThread, thread);
         ThreadContext context = service.registerNewThread(rubyThread);
         service.associateThread(thread, rubyThread);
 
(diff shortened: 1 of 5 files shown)


headius:
The second commit here may actually be the only fix needed. We hold a hard reference to the thread but once we are no longer using that local variable the JVM may consider it dereferenced. Doing the Thread.start via the WeakReference opens up the possibility that GC will have claimed the object. Instead, I modified the logic to start the native Thread directly, so it remains referenced until (at least) the Thread#start call has completed.

headius:
I got some confirmation from our JVM friends that this theory could very well be true. Even though we hold a reference to the Thread object in a local variable, by the time we call start we are no longer using that local variable. The start call itself happens after traversing the WeakReference, so if the JVM happens to GC before that point but after the last use of the local variable, we might see the WR get cleared.

Incredible "luck" to catch this on film.

headius:
Strange failure in concurrent-ruby here makes me reluctant to merge this just yet. The CyclicBarrier test tries to wait for all threads to finish, but at least one of them seems to still be waiting with this stack trace:

```
  #number_waiting
    without any waiting thread
      should be equal to zero
    with waiting threads
java.lang.Object.wait(Native Method)
java.lang.Object.wait(Object.java:460)
org.jruby.RubyThread$SleepTask.run(RubyThread.java:1531)
org.jruby.RubyThread.executeBlockingTask(RubyThread.java:1584)
org.jruby.RubyThread.wait_timeout(RubyThread.java:2132)
com.concurrent_ruby.ext.SynchronizationLibrary$JRubyLockableObject.nsWait(SynchronizationLibrary.java:254)
com.concurrent_ruby.ext.SynchronizationLibrary$JRubyLockableObject$INVOKER$i$0$1$nsWait.call(SynchronizationLibrary$JRubyLockableObject$INVOKER$i$0$1$nsWait.gen)
org.jruby.internal.runtime.methods.JavaMethod$JavaMethodN.call(JavaMethod.java:815)
org.jruby.internal.runtime.methods.DynamicMethod.call(DynamicMethod.java:203)
org.jruby.runtime.callsite.CachingCallSite.call(CachingCallSite.java:172)
org.jruby.ir.interpreter.InterpreterEngine.processCall(InterpreterEngine.java:316)
org.jruby.ir.interpreter.StartupInterpreterEngine.interpret(StartupInterpreterEngine.java:72)
org.jruby.ir.interpreter.InterpreterEngine.interpret(InterpreterEngine.java:86)
org.jruby.internal.runtime.methods.MixedModeIRMethod.INTERPRET_METHOD(MixedModeIRMethod.java:156)
org.jruby.internal.runtime.methods.MixedModeIRMethod.call(MixedModeIRMethod.java:143)
org.jruby.runtime.callsite.CachingCallSite.call(CachingCallSite.java:182)
org.jruby.runtime.callsite.CachingCallSite.callIter(CachingCallSite.java:191)
org.jruby.ir.interpreter.InterpreterEngine.processCall(InterpreterEngine.java:337)
org.jruby.ir.interpreter.StartupInterpreterEngine.interpret(StartupInterpreterEngine.java:72)
org.jruby.ir.interpreter.Interpreter.INTERPRET_BLOCK(Interpreter.java:116)
org.jruby.runtime.MixedModeIRBlockBody.commonYieldPath(MixedModeIRBlockBody.java:137)
org.jruby.runtime.IRBlockBody.doYield(IRBlockBody.java:172)
org.jruby.runtime.BlockBody.yield(BlockBody.java:108)
org.jruby.runtime.Block.yield(Block.java:184)
com.concurrent_ruby.ext.SynchronizationLibrary$JRubyLockableObject.rubySynchronize(SynchronizationLibrary.java:232)
```

I'm not sure if this is a problem related to my change (which seems unlikely) or something flaky about the concurrent-ruby tests. I will make the simpler change on master (ensure the thread remains hard referenced until `start` returns) and use this PR to work on additional cleanup and refactoring of this thread lifecycle.

headius:
The concurrent-ruby failure appears to be a bad test, which I filed in https://github.com/ruby-concurrency/concurrent-ruby/issues/862

PR #6145, opened 2020-03-27
Update specs

eregon:


eregon:
Tags added, should be green now, merging.

headius:
Thank you!

headius:
Spec updates and other maintenance items not related to a specific release should get the Non-Release milestone, for future reference. We don't want resolved issues to have a blank milestone.

PR #6146, opened 2020-03-28
Ruby 2.6 : Kernel#Integer with exception option

k77ch7:
This PR implements Kernel#Integer with exception option.
See MRI's NEWS file https://github.com/ruby/ruby/blob/ruby_2_6/NEWS


headius:
Ahh nuts, I went ahead and implemented this and the other numerics earlier today, before I saw this PR ☹️

Thanks for the help, though!

PR #6150, opened 2020-03-29
Source FFI Ruby sources from the gem

headius:
This PR will remove FFI Ruby sources from our repository and source them from the gem, so they can be upgraded as any other default gem.

It doesn't pass now because the "java" version of the gem contains no sources.


headius:
I've rebased off current master and this should be ready to go now.

We still have to update specs manually.

headius:
So close! The main ffi.rb file unfortunately uses `Gem::Version` to parse the JRuby version when checking for >= "9.3.pre", but if we are running with RubyGems disabled it blows up. That prevents JRuby from booting at all, since some internal functionality depends on FFI to be loadable.

I have pushed ffi/ffi#823 to fix this and avoid touching the Gem namespace while loading ffi.rb.

larskanis:
https://github.com/ffi/ffi/pull/823 is merged and released as ffi-1.14.0. I also added JRuby tests on github actions, but [they fail](https://github.com/ffi/ffi/runs/1579481882?check_suite_focus=true#step:8:27) since this PR is not yet on jruby-head.

headius:
@larskanis Thank you for the update! I will fix this to the release and merge.

headius:
@larskanis I am in the process of working out the last kinks, but noticed that there are platform files only in JRuby for things like syslog constants. I believe these files were generated by Rubinius many years ago and came along with us adding FFI to JRuby. Some of them even have Rubinius property names. I am fine keeping these files versioned in JRuby for now but I will be looking into a way to remove them from the `ffi` subdirs.

Example: https://github.com/jruby/jruby/tree/862ebf4ea162d44870fc06272020529285842a56/lib/ruby/stdlib/ffi/platform/x86_64-linux

headius:
A quick search seems to indicate that only the syslog files are being used by any libraries in JRuby. The `platform.conf` files also appear to be unnecessary, likely just another place Rubinius stored platform-specific constants for a few other libraries they implemented with FFI.

headius:
@larskanis I have created https://github.com/jruby/jruby/pull/6503 to address the files not found in the FFI gem.

PR #6176, opened 2020-04-14
Shut down threads on teardown

headius:
This change modifies our shutdown process to asynchronously interrupt and wait for all registered threads (excluding main) to allow them to complete execution.

Previously our behavior was to walk away from running threads. This behavior appears to have either changed in CRuby long ago, or we never implemented this behavior.

Note that this implementation does not currently differentiate between threads started from Ruby code and threads started by other code in the JVM that happened to get adopted by JRuby. We probably should not attempt to kill those adopted threads, and certainly don't want to try to join on them.

The following script illustrates some behavior of thread shutdown. This may be a good start at writing a spec.

```ruby
# normal Ruby thread should be interrupted and awaited
t = Thread.new {
  p 1
  begin
    sleep
  ensure
    p 2
  end
}

# at_exit blocks should be run after thread termination
at_exit { p 3 }

def four(*); p 4; end

# finalization at shutdown runs after after thread termination
o = Object.new
ObjectSpace.define_finalizer(o, method(:four))

# java daemon thread should not prevent shutdown and also not be interrupted
java.lang.Thread.new do
  java.lang.Thread.currentThread.daemon = true
  begin
    sleep
  ensure
    p 5
  end
end

exit # process should output "1 2 3 4" but not "5".
```

### DIFF
--- a/core/src/main/java/org/jruby/Ruby.java
+++ b/core/src/main/java/org/jruby/Ruby.java
@@ -3282,6 +3282,9 @@ public final class Ruby implements Constantizable {
             }
         }
 
+        // Shut down thread service after at_exit hooks have run
+        threadService.teardown();
+
         // Fetches (and unsets) the SIGEXIT handler, if one exists.
         IRubyObject trapResult = RubySignal.__jtrap_osdefault_kernel(this.getNil(), this.newString("EXIT"));
         if (trapResult instanceof RubyArray) {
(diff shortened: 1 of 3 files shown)


headius:
Note a few issues reference our non-shutdown of running threads:

* In #5519 @kares mentions it might be good for us to start shutting down threads at termination.
* In #3313 we discuss whether or not to shut down running threads and the behavior difference from CRuby.
* #5437 does not directly deal with thread lifecycle, but relates to it via signal and at_exit handling.

headius:
Reviewed by @enebo with no major concerns. We'll merge and let this bake.

There's one additional change I considered and rejected: don't shut down all threads if we're not the "main" runtime (i.e. check the `systemExit` argument to `Ruby.tearDown`). At a glance it seems better to not shut down threads if the process is not exiting, but this would mean we have Ruby threads still running against a torn-down runtime. For now I have opted to always attempt to shutdown child Ruby threads during runtime teardown.

PR #6178, opened 2020-04-14
Align 8 and 11 in github actions CI

headius:
Trying to get the remaining maven suites to pass in 11 so we can matrix these.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -6,77 +6,23 @@ env:
   JAVA_OPTS: '-XX:+TieredCompilation -XX:TieredStopAtLevel=1'
 
 jobs:
-  maven-test-8:
+
+  # Maven test suites that pass on both Java 8 and 11
+  maven-test:
     runs-on: ubuntu-latest
 
     strategy:
-      fail-fast: false
-
-    steps:
-    - name: checkout
-      uses: actions/checkout@v2
-    - name: set up java 8
-      uses: actions/setup-java@v1
-      with:
-        java-version: 8
-    - name: cache dependencies
-      uses: actions/cache@v1
-      with:
-        path: ~/.m2
-        key: ${{ runner.os }}-m2-${{ hashFiles('**/pom.xml') }}
-        restore-keys: ${{ runner.os }}-m2
-    - name: bootstrap
-      run: mvn -Pbootstrap clean package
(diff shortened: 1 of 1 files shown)


headius:
@kares The `jruby-jars` suites seem to be failing with that old "no password supplied for keystore" issue. What was the resolution/workaround for that?

The other failure is the `test` profile which fails to start up testing because it can't find the `javax.scripting` package. We will need to pass an appropriate module flag to enable that module when running this suite.

headius:
@mkristian I'm not sure how to specify the appropriate module flag when running `-Ptest` on Java 11.

I'm going to go ahead and merge this but it would be really nice if we could get those last two suites working on 11.

mkristian:
I will have look today, hopefully via some maven 'module'

kares:
> The jruby-jars suites seem to be failing with that old "no password supplied for keystore" issue. What was the resolution/workaround for that?

some of the ruby maven stuff must be using an old jruby (with old jruby-openssl) under the hood.
have tried forcing to use newer jruby with the plugin but it just keeps coming back, the plugins could use an updated release. and now I am not sure which one exactly ... was to blame here.

headius:
I get a different error locally running `mvn -Dinvoker.skip=false -Pjruby-jars`under adoptopenjdk 11:

```
[INFO] --- gem-maven-plugin:1.0.7:package (default-package) @ jruby-jars ---
[INFO] include dependencies? true
/Users/headius/projects/jruby/lib/pom.rb:2: warning: already initialized constant MORE_QUIET
[INFO] use repository layout? false
[INFO]  -- include -- org.jruby:jruby-stdlib:jar:9.3.0.0-SNAPSHOT:compile
[INFO] Exception in thread "main" java.lang.ExceptionInInitializerError
[INFO] 	at org.jruby.Main.<init>(Main.java:99)
[INFO] 	at org.jruby.Main.main(Main.java:193)
[INFO] Caused by: java.lang.RuntimeException: unsupported Java version: 11
[INFO] 	at org.jruby.RubyInstanceConfig.initGlobalJavaVersion(RubyInstanceConfig.java:1835)
[INFO] 	at org.jruby.RubyInstanceConfig.<clinit>(RubyInstanceConfig.java:1585)
[INFO] 	... 2 more
```

PR #6180, opened 2020-04-17
Implement oneshot_lines for coverage

headius:
This commit also makes some substantial changes to how coverage is
compiled in IR and JIT.

* Coverage is now independent of tracing and adds behavior to the
  LineNumberInstr
* Coverage is always available
* Each covered LineNumber does its own update of coverage data
* Oneshot mode only updates once and then clears coverage flag
* JIT uses indy to bind the coverage updating logic, with oneshot
  clearing it to a no-op after the first execution

### DIFF
--- a/core/src/main/java/org/jruby/ir/IRBuilder.java
+++ b/core/src/main/java/org/jruby/ir/IRBuilder.java
@@ -6,6 +6,7 @@ import org.jruby.RubySymbol;
 import org.jruby.ast.*;
 import org.jruby.ast.types.INameNode;
 import org.jruby.compiler.NotCompilableException;
+import org.jruby.ext.coverage.CoverageData;
 import org.jruby.parser.StaticScope;
 import org.jruby.runtime.ArgumentDescriptor;
 import org.jruby.runtime.ArgumentType;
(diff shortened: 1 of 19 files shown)


core/src/main/java/org/jruby/ir/IRBuilder.java
```diff
@@ -336,12 +337,15 @@ public void addArgumentDescription(ArgumentType type, RubySymbol name) {
     public void addInstr(Instr instr) {
         if (needsLineNumInfo) {
             needsLineNumInfo = false;
-            addInstr(manager.newLineNumber(_lastProcessedLineNum));
+
+            if (needsCodeCoverage()) {
+                addInstr(new LineNumberInstr(_lastProcessedLineNum, true, (coverageMode & CoverageData.ONESHOT_LINES) != 0));
```
enebo, line 342: This is an implicit true in constructor isn't it?

headius: So you are saying remove that true since we'll always provide oneshot = true | false, which disambiguates enough?

headius: I modified the code to pass the coverage mode instead of booleans.

PR #6182, opened 2020-04-18
[ji] java_import now raise NameError on invalid names

kares:
given invalid import names `java_import` will raise `NameError`, 

removing `constant =~ /^[A-Z].*/` capitalization validation, leaving that part to `const_set`
the `[A-Z]` validation did not handle multi-byte capital (UTF-8) characters that well ...
in theory Java class names support the full Unicode set not just ascii characters.


enebo:
Looks like some nice error messages to me

Issue #6157, opened 2020-04-01
CSV parsing and encoding errors in 2.6


headius:
Not sure why these tests are failing on JRuby's 2.6 branch, since we use the exact same code for the `csv` library. However, here they are.

```
[ 36/897] TestCSV::Parsing#test_malformed_csv = 0.02 s                                          
  1) Failure:
TestCSV::Parsing#test_malformed_csv [/Users/headius/projects/jruby/test/mri/csv/test_csv_parsing.rb:171]:
<"Unquoted fields do not allow \\r or \\n (line 4)."> expected but was
<"Unquoted fields do not allow new line <\"\\r\"> in line 4.">.

[ 89/897] TestCSVEncodings#test_encoding_is_upgraded_during_writing_as_needed = 0.01 s                             
  2) Failure:
TestCSVEncodings#test_encoding_is_upgraded_during_writing_as_needed [/Users/headius/projects/jruby/test/mri/csv/test_encodings.rb:233]:
<"UTF-8"> expected but was
<"US-ASCII">.

[ 90/897] TestCSVEncodings#test_encoding_is_upgraded_for_ascii_content_during_writing_as_needed = 0.01 s
  3) Failure:
TestCSVEncodings#test_encoding_is_upgraded_for_ascii_content_during_writing_as_needed [/Users/headius/projects/jruby/test/mri/csv/test_encodings.rb:241]:
<"UTF-8"> expected but was
<"ISO-8859-1">.

[108/897] TestCSVEncodings::DifferentOFS#test_encoding_is_upgraded_during_writing_as_needed = 0.01 s             
  4) Failure:
TestCSVEncodings::DifferentOFS#test_encoding_is_upgraded_during_writing_as_needed [/Users/headius/projects/jruby/test/mri/csv/test_encodings.rb:233]:
<"UTF-8"> expected but was
<"US-ASCII">.

[109/897] TestCSVEncodings::DifferentOFS#test_encoding_is_upgraded_for_ascii_content_during_writing_as_needed = 0.00 s
  5) Failure:
TestCSVEncodings::DifferentOFS#test_encoding_is_upgraded_for_ascii_content_during_writing_as_needed [/Users/headius/projects/jruby/test/mri/csv/test_encodings.rb:241]:
<"UTF-8"> expected but was
<"ISO-8859-1">.

[467/897] TestCSVParseInvalid#test_ignore_invalid_line = 0.00 s                                                          
  6) Error:
[...]

PR #6190, opened 2020-04-24
Restart dead enumerator

headius:
This fixes the remaining failure in CSV as discussed in #6157. Enumerators that fail prematurely due to an exception should be re-started on the next call to `next`.


headius:
Failure is due to my removing an exclude that doesn't fail on MacOS.

headius:
I do not agree with the restarting behavior, so I have filed https://bugs.ruby-lang.org/issues/16816

Issue #6191, opened 2020-04-24
Indy fails on some parts of Ruby on Rails


Mahaswami:
**Environment Information**

JRuby 9.2.11 / Ubuntu 18.04 / JDK 14
  
Flags: -Djruby.compile.invokedynamic=true
Rails: 4.2.x

**Expected Behavior**

Rails app to function normally with Indy turned on

**Actual Behavior**

When this flag is turned on, the app blows up with the following exception

Java::ComHeadiusInvokebinder::InvalidTransformException (arguments matching arg.* are not all of the same type):

Please see the attached trace.

[jruby indy rails exception.txt](https://github.com/jruby/jruby/files/4531526/jruby.indy.rails.exception.txt)

We also receive one more exception which we believe Indy related

```
gems/activerecord-jdbc-adapter-1.3.23/lib/arjdbc/jdbc/adapter.rb:528:in `block in exec_query_raw': ArgumentError: wrong number of arguments (given 1, expected 2..3):         SELECT a.attname, format_type(a.atttypid, a.atttypmod),
               pg_get_expr(d.adbin, d.adrelid), a.attnotnull, a.atttypid, a.atttypmod
          FROM pg_attribute a LEFT JOIN pg_attrdef d
            ON a.attrelid = d.adrelid AND a.attnum = d.adnum
         WHERE a.attrelid = '"cg_community"."denormalized_activity_streams"'::regclass
           AND a.attnum > 0 AND NOT a.attisdropped
         ORDER BY a.attnum
 (ActiveRecord::StatementInvalid)
    from gems/activerecord-4.2.8/lib/active_record/connection_adapters/abstract_adapter.rb:484:in `block in log'
    from gems/activesupport-4.2.8/lib/active_support/notifications/instrumenter.rb:20:in `instrument'
```

[jruby indy rails another exception.txt](https://github.com/jruby/jruby/files/4531549/jruby.indy.rails.another.exception.txt)

Though exception points to a SQL, this could be because of some internal Indy wiring is behaving diffeerntly.

headius:
The first issue was fixed after 9.2.11 was released. See #6130 

I don't believe there's any workaround for this. The `method_missing` optimization was applied in JRuby 9.2.11 and I did not make it possible to disable it.

@enebo Perhaps we need a mechanism, either through Options or something finer-grained and more hidden, to be able to tweak all indy optimizations?

[...]

PR #6192, opened 2020-04-24
Merge variadic method_missing indy fix from master

headius:
This was previously fixed only on master in #6132, but in order to support #6191 I am merging it to the jruby-9.2 branch. We won't release a 9.2.12, but users can build this branch to get the patched version.


headius:
@enebo Do you have any objections to merging this fix to jruby-9.2 branch?

headius:
@enebo I went ahead with it... not sure how you would like this milestone set.

enebo:
If we ever make another release for some particular reason we will remark this to that release.  For now it is a fix in no release so this feels like the way to make it.

PR #6194, opened 2020-04-26
Remove deprecated classes

headius:
Any classes deprecated and unused for more than two years will be deleted in JRuby 9.3. Most of these have been deprecated much longer.


headius:
I have removed what I thought was safe to remove. There are some remaining items:

* RuntimeHelpers was deprecated in 2012, but was still in use in Nokogiri until Feb 2019. It was fixed to use Helpers by @jvshahid in sparklemotion/nokogiri#1874. This happened in the middle of JRuby 9.2 versions, so we'll have to wait a while
* Several classes in LoadService from an older implementation of file searching were not removed due to upcoming #5764. They could be removed after that has been merged, since they've been deprecated since around 2012.
* RubyContinuation may still be in use in the racc library's parser extension. That usage was fixed in ruby/racc#92 but with ruby/racc#91 still open I'm not sure if it ever got into a release.
* ClassVarDeclNode is deprecated, but as part of the NodeVisitor interface I'm not sure we can ever remove it. Deleting the method would break any implementers that explicitly `@Override` since the method will have disappeared. We may consider introducing a new interface at some point and having the old one extend it.

PR #6195, opened 2020-04-27
Use all the tricks to properly eliminate illegal access warnings

headius:
This PR attempts to tweak all of our uses of `setAccessible` for accessing internal JDK classes to do the "right thing", be that using method handles, `addOpens`, or other tweaks to open up these classes if we legally can or fall back gracefully to degraded functionality *without warnings*.

The general strategy is described in the first commit to FilenoUtil, but I summarize it again here:

* Reflecting against these classes should always succeed if they are present, so this does not change.
* Instead of using the Field or Method directly, we unreflect it into a MethodHandle. This lets us know we have a fully-accessible way to deal with the reflected access, and since it raises errors instead of silently warns, we can fall back for inaccessible structures.
* When we are dealing with an inaccessible structure, use `Module.addOpens` to make it *really* open, which then gives us a workable method handle and no warning output.

Much of this is to deal with the unfortunate fact that our current check, `Module.isOpen`, will always return true for the unnamed module *but still warn when you set it accessible*. This "kinda-sorta-open" state is a bug in my opinion, but at least explicitly using `addOpens` promotes it to a fully-open state with no warnings.

I will endeavor to eliminate all warnings in all modes with as little fallback as necessary.

### DIFF
--- a/core/pom.rb
+++ b/core/pom.rb
@@ -73,7 +73,7 @@ project 'JRuby Core' do
 
   jar 'me.qmx.jitescript:jitescript:0.4.1', :exclusions => ['org.ow2.asm:asm-all']
 
-  jar 'com.headius:backport9:1.6'
+  jar 'com.headius:backport9:1.7'
 
   jar 'javax.annotation:javax.annotation-api:1.3.1', scope: 'compile'
 
--- a/core/pom.rb
+++ b/core/pom.rb
@@ -73,7 +73,7 @@ project 'JRuby Core' do
 
   jar 'me.qmx.jitescript:jitescript:0.4.1', :exclusions => ['org.ow2.asm:asm-all']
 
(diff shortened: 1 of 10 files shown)


kares:
:+1: very interesting, falsely assumed the runtime `addOpens` would still print warnings ...

headius:
@kares It's pretty silly if you ask me. If you're running from the classpath (not the module path) these core JDK classes will claim to be open, but then actually warn because you're not supposed to access them. Adding via `addOpens` in this case is safe, because they're "kinda sorta open", and then the warnings won't happen. I'm frustrated by this silly choice, but at least we have a path forward.

headius:
Retarget to 9.2.12 since this is largely zero-cost and cleans up most remaining warnings in JRuby proper. See #6285.

Issue #6197, opened 2020-04-30
Missing method that is public but defined on a non-public superclass


fidothe:
This is the same issue seen in #5885, but now extends to OpenJDK 8 as well as OpenJDK 12 (these are all I've tested on) under > 9.2.10.

**Environment Information**
JRuby 9.2.10.0, 9.2.11.1 invoked with --dev --debug

x86_64, macOS 10.15.4 (Darwin Kernel Version 19.4.0)

**Expected**
Calling the `setInitialMode` on `net.sf.saxon.s9api.Xslt30Transformer` (which is defined on the abstract class `net.sf.saxon.s9api.AbstractXsltTransformer`) should not explode with a `NoMethodError`

**Actual**
On 9.2.8, 9.2.10, and 9.2.11 under OpenJDK 12 this goes boom. On 9.2.8 under OpenJDK 8, it doesn't explode and works as expected (as reported in #5585). On 9.2.10 and 9.2.11 under OpenJDK 8 it fails in the same way as it does under OpenJDK 12.

I'll add a test case and see if I can reproduce under master too.

headius:
We did not get this fixed for 9.2.12 but there's a workaround use `java_alias` to explicitly bind the missing methods:

```ruby
class net::sf::saxon::s9api::AbstractXsltTransformer
  java_alias :setInitialMode, :setInitialMode, [<parameter types for setInitialMode here>]
end
```

It's not pretty but it works for now.

headius:
The general issue here, to help searches and users that run into this in the future, is that we are completely ignoring the methods on non-public classes even if they would be visible from a child class. This is the simple form that represents the original reported issue:

```java
abstract Foo {
  public void foo() {}
}

public Bar extends Foo {}
```

The `foo` method can't be called against a `Foo` instance since `Foo is not public, but it can be called against a `Bar` instance, since both the method and the target class are public in that case. Because `Foo` is not public, we do not bind *any* methods from it, so it ends up being missing from `Bar` instances.

The attached PR attempts to fix this, but it has some problems.

Morozzzko:
I've met this issue while trying to integrate github.com/structurizr/java and JRuby.

Top-level issue: a method was missing even though it existed in a .jar and was accessible via Java and Kotlin. It returned a simple `NoMethodError`

When I tried to introspect the class via `#methods` and `#instance_methods` I couldn't see _any_ methods which were defined in a class.

[...]

PR #6198, opened 2020-04-30
Add test for #6197

fidothe:
Test exercising dispatch to a method on a Java object which is inherited from an
abstract class is failing with NoMethodError. Illustrates #6197


enebo:
I see no CI here but it appears that we should pass the spec as written so I am extremely belatedly merging this.

PR #6211, opened 2020-05-11
Ruby 2.6 : fix failling transcode tests.

k77ch7:
This fixes failling TestTranscode#test_invalid_replace_string and #test_scrub_encode_with_coderange.

see https://github.com/ruby/ruby/blob/v2_6_6/string.c#L10152-L10172
      https://github.com/ruby/ruby/blob/v2_6_6/transcode.c#L2703


ahorek:
nice, this also fixes https://github.com/jruby/jruby/blob/0b478ce79674c400af028827563b045938b0efe6/spec/tags/ruby/core/string/encode_tags.txt#L1

headius:
Looks great!

headius:
Oops

PR #6220, opened 2020-05-14
fix copy & paste error

ahorek:
fixes a wrong reference introduced by https://github.com/jruby/jruby/commit/e8dbe9c269f7723badb89d137b2a1d52b3d3400b


headius:
Good catch!

Issue #6153, opened 2020-03-30
Passing `:chdir` to `Kernel.system` doesn't work reliably


deivid-rodriguez:
**Environment Information**

```
$ jruby -v
jruby 9.2.11.0 (2.5.7) 2020-03-02 612d7a05a6 OpenJDK 64-Bit Server VM 11.0.6+10-post-Ubuntu-1ubuntu118.04.1 on 11.0.6+10-post-Ubuntu-1ubuntu118.04.1 [linux-x86_64]
$ env | grep JRUBY
JRUBY_OPTS= --dev --debug -J-Xmx1024m
$ env | grep JAVA
JAVA_HOME=/usr/lib/jvm/java-11-openjdk-amd64
$ uname -a
Linux chaba 5.3.0-42-generic #34~18.04.1-Ubuntu SMP Fri Feb 28 13:42:26 UTC 2020 x86_64 x86_64 x86_64 GNU/Linux
```

**Expected Behavior**

I expect that the following code would behave the same as common ruby, and print "hola" to the screen.

```
system(Gem.ruby, "-e", "puts('hola')", :chdir => Dir.pwd)
```

**Actual Behavior**

Instead, I get

```
sh: 1: Syntax error: "(" unexpected
```

Without the `:chdir` option, it works as expected.

headius:
The root problem here is likely not the options but the `:chdir`.

Our native process launching is based on `posix_spawn`, which does not support doing `chdir` in the new child process. So this ends up falling back on JDK process launching, which does support `chdir`.

This means we have to hand-assemble a command line to pass to `sh -c` (Ruby launches via sh for these sorts of command lines), and that resulting command line probably does not properly escape the parentheses in `puts('hola')`.

The `chdir` behavior is not easy for us to fix; normally, if we could fork, make changes, and then exec, we'd chdir in the forked subprocess before calling exec (as CRuby and the JDK process APIs do). Unfortunately we can't safely fork from Java, so our options are no chdir and real native process launching, or chdir and Java's cumbersome process launching.

There are a couple workarounds for you:

* You can restructure your command line to be a valid sh script that does a chdir before starting.
* If you know that changing the directory in the parent process won't interfere with other executing threads, you can do a native chdir yourself before calling `system`.

deivid-rodriguez:
Thank you!

I already workarounded it, but I thought it would be good to report it since it seemed like a clear bug.

eregon:
[...]

PR #6226, opened 2020-05-15
Fixes to improve support for chdir in spawn functions

headius:
This PR will be an aggregate of fixes to make process launching with chdir work more like it does in CRuby.

The main issue here is in our native process launching, which uses `posix_spawn`. There's no ability in this function to do a chdir before launching the new process, and we have worked around this in a number of ways:

* If we are launching with the JDK APIs, we can use the JDK's ProcessBuilder support for chdir. This backends at the JVM level on a fork+exec pattern. Note that as mentioned in #6153, the keyword argument `chdir` option is not yet supported along this path, but it does already handle our virtual CWD.
* If we are launching using native APIs and the command is a shell script, we prepend a `cd somedir` to that script. This may be sufficient, since the expectation is that the remainder of the script is already valid shell. We may wish to pass the requested dir in using argument passing as mentioned in #6153.
* If we are launching using native APIs and the command and arguments are separate parameters to e.g. `system`, we must construct a new command line using sh and chdir without the actual program and arguments getting interpreted as part of a shell script. The current JRuby code just naively joins the command and its arguments into a single string, which is the cause of #6153.

The bulk of work in this PR will make use of the examples provided by @Mrnoname1000 to use exec within our intermediate shell script, ensuring the actual command and arguments are handled without the shell attempting to treat them as part of a shell script.

There may be other combinations not described above that become apparent while working on these fixes.

TO DO:

* [ ] chdir support for non-native system calls
* [ ] confirm chdir support when running a shell script (single argument with shell characters)
* [ ] native launch with chdir should avoid exposing program arguments to sh
  * [x] multi-arg launch with chdir (program plus args as separate parameters)
  * [ ] single-arg launch with chdir (just program)
  * [ ] still validate target program when using intermediate shell to do chdir


headius:
The first fix here utilizes the following pattern, suggested by @Mrnoname1000, to create a multi-argument process launch that safely does a chdir followed by an exec.

```c
execlp("sh", "/bin/sh", "-c", "cd \"$0\"; exec \"$@\"", path_to_cd, arguments, NULL);
```

This fixes the original case provided in #6153, since the embedded Ruby script is passed along to the requested Ruby interpreter without shell intervention.

I have done a somewhat "brute force" implementation of this change. When we determine that a chdir is necessary and we have multiple arguments (indicating an execl-style process launch), the additional elements from the C call example above are prepended to the collection of program and arguments and the chdir flag is cleared. This avoids falling back on existing logic to launch with `sh`, which is really meant for handling system calls where the sole command is clearly a shell script.

There are a few items missing here:

* It's not clear to me yet why we don't use this logic for argc = 1, or a single program command. That case also needs to be handled.
* The target program is no longer validated on the filesystem before proceeding, since the new command is `/bin/sh` and we know it will exist.

headius:
As a note for future folks, a recap of what has been added:

* chdir with system with multiple arguments should be working well now, which fixes #6153.
* The native system logic is now being used whenever it is available, rather than only when not doing chdir.
* I have done some minimal cleanup of the ported popen code
* A number of excludes have fallen off due to this work.

There are a few possible improvemements in the future:

* Because bash exec can accept a few flags (POSIX exec does not) we may want to detect bash and pass -- to avoid the potential for a command to pass exec args.
* An alternative fix would be to not exec, but then that leaves an intermediate sh process in the process list, waiting for the termination of our subcommand.

Thanks to @mrnoname1000 for helping with these arcane shell oddities.

Issue #6219, opened 2020-05-14
JRuby thrown NullPointerException on File.open of nonexistent uri


JohnPhillips31416:
JRuby version: 9.2.11.1

For the folllowing file

```
fname = "uri:classloader:/a/b/c/d.e"

puts "fname is #{fname} = #{File.path(fname)}"

File.open(fname) do |file|
  puts "Opened #{file.name}"
end
```

I get NullPointerException not 'No such file or directory' if the file does not exist

```
$ ruby -v
jruby 9.2.11.1 (2.5.7) 2020-03-25 b1f55b1a40 OpenJDK 64-Bit Server VM 11.0.6+10 on 11.0.6+10 +jit [darwin-x86_64]
$ ruby foo.rb 
fname is uri:classloader:/a/b/c/d.e = uri:classloader:/a/b/c/d.e
Unhandled Java exception: java.lang.NullPointerException
java.lang.NullPointerException: null
      openInputStream at org/jruby/util/URLResource.java:153
          openChannel at org/jruby/util/URLResource.java:158
                 open at org/jruby/util/io/PosixShim.java:438
          cloexecOpen at org/jruby/RubyIO.java:1269
          sysopenFunc at org/jruby/RubyIO.java:1257
      sysopenInternal at org/jruby/RubyIO.java:1249
              sysopen at org/jruby/RubyIO.java:1234
      fileOpenGeneric at org/jruby/RubyFile.java:1471
             openFile at org/jruby/RubyFile.java:1448
           initialize at org/jruby/RubyFile.java:365
                 call at org/jruby/RubyFile$INVOKER$i$0$3$initialize.gen:-1
                 call at org/jruby/runtime/callsite/CachingCallSite.java:84
          newInstance at org/jruby/RubyClass.java:939
                 open at org/jruby/RubyIO.java:1156
                 call at org/jruby/RubyIO$INVOKER$s$0$0$open.gen:-1
                 call at org/jruby/internal/runtime/methods/DynamicMethod.java:204
         cacheAndCall at org/jruby/runtime/callsite/CachingCallSite.java:386
                 call at org/jruby/runtime/callsite/CachingCallSite.java:184
             callIter at org/jruby/runtime/callsite/CachingCallSite.java:191
    invokeOther4:open at foo.rb:6
               <main> at foo.rb:6
  invokeWithArguments at java/lang/invoke/MethodHandle.java:710
                 load at org/jruby/ir/Compiler.java:89
[...]

PR #6227, opened 2020-05-15
NPE File.open of nonexistent uri

ahorek:
fixes: https://github.com/jruby/jruby/issues/6219


headius:
Wow, somehow in my bug-fixing fever I did not notice this PR. Thank you, and sorry this fell off the radar!

PR #6228, opened 2020-05-15
Default to space padding for "special" formats

### LINKED ISSUE #6169: strftime precision against combinated formats (ex. %F, %r, ...)
`Time.strftime` has some combinated formats, such as `%F` equal to `%Y-%m-%d`, and `%r` equal to `%I:%M:%S %p`.

Users can apply options even for those combinated formats, for example `%20F`. But the behavior is different between CRuby and JRuby.

**Environment Information**

* Confirmed with JRuby 9.1.15.0 and 9.2.10.0.
* Mac OS X 10.13. (The OS does not sound related, though.)

**Expected Behavior**

This is CRuby's behavior. (2.4, 2.5, 2.6, 2.7)

```
irb(main):001:0> require 'time'
=> true
irb(main):002:0> Time.now.strftime("%20F")
=> "          2020-04-06"
```

**Actual Behavior**

JRuby is working like :
```
irb(main):001:0> require 'time'
=> true
irb(main):002:0> Time.now.strftime("%20F")
=> "00000000000000002020-04-06"
```

headius:
This fixes #6169 and I believe matches CRuby for the "special" formats handled here, but I am unsure if this is the right way to fix it. CRuby has something similar in its complicated strftime code, but they do not have a separate parser for date formats like we do.

@enebo I feel like this is something that should be done in the strftime parser; it should see that the next token is going to want space padding and insert the FORMAT_OUTPUT token with appropriate flags. What do you think?

### DIFF
--- a/core/src/main/java/org/jruby/util/RubyDateFormatter.java
+++ b/core/src/main/java/org/jruby/util/RubyDateFormatter.java
@@ -275,12 +275,37 @@ public class RubyDateFormatter {
         lexer.yyreset(reader);
 
         Token token;
+        RubyTimeOutputFormatter formatter = null;
         try {
             while ((token = lexer.yylex()) != null) {
-                if (token.format != Format.FORMAT_SPECIAL) {
+                if (token.format == Format.FORMAT_OUTPUT) {
+                    formatter = (RubyTimeOutputFormatter) token.data;
+                } else if (token.format != Format.FORMAT_SPECIAL) {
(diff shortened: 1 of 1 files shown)


enebo:
@headius yeah maybe that works and is only a single pass over the format string.  I guess it seems reasonable if that is really a hard and fast rule.

headius:
We will go with this fix for now but may want to rework this and add a format cache in the future.

PR #6232, opened 2020-05-19
Avoid binding Kotlin Companion classes

### LINKED ISSUE #6196: "warning: already initialized constant Companion" with Kotlin companion objects
When using Kotlin companion objects from JRuby, we get an "already initialized constant Companion" warning simply by loading the class.  The companion object works correctly and is accessible as a constant from the parent class.  

We're seeing this on JRuby 9.2.11.1.  I tested prior versions and this warning first appears in JRuby 9.2.9.0.

**Detailed Description**

Given a Kotlin class like:
```
package kotlintest

class Test {
	companion object {
		const val FOO = "foo"
	}
}
```

And a ruby class like:
```
require 'java'
require './build/libs/test-all.jar'

p Java::Kotlintest::Test::FOO
```

We get the following output:
```
$ ruby test.rb
test.rb:4: warning: already initialized constant Companion
"foo"
```

The warning is triggered just by loading the class.  So if you just reference `Java::Kotlintest::Test`, without the companion object, you still get the warning.

Gradlefile used to build the jar:
```
plugins {
    id("org.jetbrains.kotlin.jvm") version "1.3.70"
    id("com.github.johnrengelman.shadow") version "5.2.0"
}

repositories {
    jcenter()
}

dependencies {
    implementation("org.jetbrains.kotlin:kotlin-stdlib-jdk8")
}
```

headius:
This logic is intended to fix #6196.

Kotlin companion objects end up creating both an inner class named "Companion" and a static final field named "Companion", which leads to the warning in #6196. This PR attempts to only bind the field, since the class is unlikely to be useful from non-Kotlin languages.

There are a few issues here I'd like some input on.

* Should we be binding anything for this?

Perhaps neither the field nor the class should be made visible to Ruby? The example in #6196 is trivial and seems to bind the one companion field as a static final field on the parent class, making the companion class and object largely useless. I don't know how other more elaborate companion objects get structured.

* I had to put this filtering in two places because as a side effect of acquiring the static final Companion field, the inner class Companion also gets bound and defined.

A second attempt to define the constant comes later in `installInnerClasses`, but filtering here was not sufficient to guarantee the class did not bind. It would be preferable to only every assign inner class constants in **one place**, but making that change is way out of scope for this PR.

* This is another check added to every class in the system.

I'm using a ClassValue, but it still requires initializing some data storage for the class specific to this Kotlin check. I'm not sure if there's a cleaner way; we have to acquired this bit somehow.

### DIFF
--- a/core/src/main/java/org/jruby/javasupport/Java.java
+++ b/core/src/main/java/org/jruby/javasupport/Java.java
@@ -58,6 +58,7 @@ import org.jcodings.Encoding;
 
 import org.jruby.*;
 import org.jruby.javasupport.binding.Initializer;
+import org.jruby.javasupport.binding.MethodGatherer;
 import org.jruby.javasupport.proxy.JavaProxyClass;
 import org.jruby.javasupport.proxy.JavaProxyConstructor;
 import org.jruby.runtime.Arity;
@@ -773,8 +774,8 @@ public class Java implements Library {
 
         final RubyModule parentModule; final String className;
 
+        Class<?> declaringClass = clazz.getDeclaringClass();
         if ( fullName.indexOf('$') != -1 ) { // inner classes must be nested
-            Class<?> declaringClass = clazz.getDeclaringClass();
             if ( declaringClass == null ) {
                 // no containing class for a $ class; treat it as internal and don't define a constant
                 return;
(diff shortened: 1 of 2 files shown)


aryeh-looker:
Could it make sense to write a test that actually compiles a simple Kotlin file and asserts correct behavior (there are a couple cases: constants in the companion class, methods, renamed companion, and the `@JvmStatic` annotation)? Would be glad to lend a hand with that if feasible.

headius:
Unfortunately this PR is not correct; the "Companion" name is just a default if you don't specify a different one, as @kalenp described in #6196.

I'm not sure anymore that there's a clean way to detect the companion object.

PR #6234, opened 2020-05-20
Improved integration with Ruby and Java exceptions

### LINKED ISSUE #6203: Raise Ruby exception in JRuby
I would like to raise an exception class defined in Ruby from Java code. How would one do this?

The examples given in https://github.com/jruby/jruby/wiki/JRuby-Reference#Exceptions all show how to work with Java exceptions, but I would like to raise a Ruby exception.

We have this code:

<pre>
    try {
      string = (RubyString) value;
    } catch (ClassCastException e) {
      throw context.runtime.newArgumentError(e.toString());
    }
</pre>

I tried:

<pre>
        RubyClass cls = this.runTime.getClass("Mongo::GssapiNative::Error");
        RubyString msg = RubyString.newString(context.runtime, message);
        throw cls.newInstance(context, msg, Block.NULL_BLOCK);
</pre>

This failed with:

<pre>
src/main/org/mongodb/sasl/GSSAPIAuthenticator.java:141: error: incompatible types: IRubyObject cannot be converted to Throwable
        throw cls.newInstance(context, msg, Block.NULL_BLOCK);
                             ^
</pre>

So I need to convert my Ruby class instance to a Throwable?

Related question: our [existing code](https://github.com/mongodb/mongo-ruby-kerberos/blob/master/src/main/org/mongodb/sasl/MongoSecurityException.java) supports exception causes, how do I attach those to Ruby exception classes? Is this handled for me automatically somehow?

Thanks in advance.

headius:
* Make a best effort to sync up `cause` between Java and Ruby.
* Provide more convenience forms of RaiseException.from that accept a String path to the exception class and varargs constructor parameters.

Fixes #6203.

### DIFF
--- a/core/src/main/java/org/jruby/Ruby.java
+++ b/core/src/main/java/org/jruby/Ruby.java
@@ -4243,9 +4243,14 @@ public final class Ruby implements Constantizable {
     }
 
     /**
-     * @param exceptionClass
-     * @param message
-     * @return
+     * Construct a new RaiseException wrapping a new Ruby exception object appropriate to the given exception class.
+     *
+     * There are additional forms of this construction logic in {@link RaiseException#from}.
+     *
+     * @param exceptionClass the exception class from which to construct the exception object
+     * @param message a simple message for the exception
+     * @return a new RaiseException wrapping a new Ruby exception
+     * @see RaiseException#from(Ruby, RubyClass, String)
      */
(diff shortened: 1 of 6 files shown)


headius:
Failure seems unrelated so I'm going to rebase and see how it looks.

headius:
I believe the MRI and spec failures are fixed with latest commit, though the root cause (exposed by this PR) still remains in #6277.

Issue #6225, opened 2020-05-15
NoMethodError: undefined method `[]' for <ClassName>:Class


mogman1:
I've created a test script with some specs to demonstrate this bug at https://github.com/mogman1/protobuf_error_script.

* Jruby versions: `>= 9.2.7.0`
* ActiveSupport `5.0` and `5.1`, but NOT `4.2` or `5.2` or `6.0`
* protobuf (v3.10.3)
* OSX and Ubuntu Xenial (see travis builds), probably others

The issue is that when attempting to check the value of a protobuf message attribute's truthiness by appending `?` to the end of it (e.g. `x.is_working?`), for certain versions of JRuby and ActiveSupport 5, you get something like ``` NoMethodError: undefined method `[]' for FailingProtobufMessage:Class```.  When the bug is occurring, the protobuf library is checking the class and not the instance itself when it calls `self` (see https://github.com/ruby-protobuf/protobuf/blob/v3.10.3/lib/protobuf/field/base_field.rb#L354 and error in Travis build).  This does not happen on any MRI >= 2.4 or JRuby <= 9.2.6.0.

I was able to pinpoint the exact commit where this started happening as at https://github.com/jruby/jruby/commit/3a56fcb3576fab1276a8ae51f0ea9d30a85690b9.  Everything before that my tests will all pass, and then everything from then on I get the error I mentioned.

mogman1:
Playing around with it a little more this morning, it looks like it's susceptible to the order in which certain things are being loaded.  The protobuf gem allows you to deprecate a field and in order to trigger the bug, there must be a deprecated field but it must occur before the boolean field.  I added a test to demonstrate that.

headius:
This is good news:

```
$ jruby -v
jruby 9.3.0.0-SNAPSHOT (2.6.5) 2020-06-09 e5d98e4865 OpenJDK 64-Bit Server VM 25.222-b10 on 1.8.0_222-b10 +jit [darwin-x86_64]

$ rake spec
/Users/headius/projects/jruby/bin/jruby -I/Users/headius/projects/jruby/lib/ruby/gems/shared/gems/rspec-support-3.9.3/lib:/Users/headius/projects/jruby/lib/ruby/gems/shared/gems/rspec-core-3.9.2/lib /Users/headius/projects/jruby/lib/ruby/gems/shared/gems/rspec-core-3.9.2/exe/rspec --pattern spec/\*\*\{,/\*/\*\*\}/\*_spec.rb
......

[...]

PR #6235, opened 2020-05-20
Revert "Perform implementer search for aliases with super."

mogman1:
This reverts commit 3a56fcb3576fab1276a8ae51f0ea9d30a85690b9.

For what it's worth, simply reverting this change (and relevant changes after) seems to resolve the issue in https://github.com/jruby/jruby/issues/6225.

### DIFF
--- a/core/src/main/java/org/jruby/internal/runtime/methods/AliasMethod.java
+++ b/core/src/main/java/org/jruby/internal/runtime/methods/AliasMethod.java
@@ -33,10 +33,8 @@
 package org.jruby.internal.runtime.methods;
 
 import org.jruby.RubyModule;
-import org.jruby.internal.runtime.AbstractIRMethod;
 import org.jruby.runtime.Arity;
 import org.jruby.runtime.Block;
-import org.jruby.runtime.Helpers;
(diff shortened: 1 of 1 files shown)


headius:
Looking into this today.

headius:
Ok, so I'm willing to believe that this logic is not exactly right, but reverting it does bring about an alias+super failure in the CRuby tests:

```
  1) Error:
TestAlias#test_super_in_aliased_module_method:
NoMethodError: super: no superclass method `foo' for #<TestAlias::SuperInAliasedModuleMethod::Derived:0x247a8411>
    /home/travis/build/jruby/jruby/test/mri/ruby/test_alias.rb:102:in `foo'
    /home/travis/build/jruby/jruby/test/mri/ruby/test_alias.rb:121:in `test_super_in_aliased_module_method'
```

This is largely what the change was intended to fix.

headius:
It seems that current JRuby master, which will be JRuby 9.3, no longer fails the case given by @mogman1. Hooray!

@mogman1 I will close this as invalid, and please let us know when you verify a recent build of JRuby 9.3.

Issue #5857, opened 2019-09-03
Even JRuby 9.2.8.0/ 9.2.11.1 terribly slow


rovf:
### Environment

Provide at least:
JRuby 9.2.8.0, Java 1.8, Windows 10

When stepping up from JRuby 1.7.27 to 9.2.8, I hoped to see some performance increasements, but on contrary, 9.2.8 turned out to be slower than 1.7.27:

One CPU-bound application, which runs constantly in about 31 minutes under 1.7, takes 36 minutes with 9.2. I am aware that I have to take into account a longer warm-up time with JRuby 9.x, but I don't think that this will have a noticeable influence, if the total running time is more than half an hour.

In a different application, which is using the `caller` method frequently in its logging package, I had expected JRuby 9.2 to be much faster, because it is said that this has improved the speed of this method. What I see instead, is:
- Running in JRuby 1.7.27 WITHOUT using `caller()`: 90 Seconds
- Running in JRuby 1.7.27 WITH using `caller` in many places, but otherwise processing the same data: 29 minutes (this shows that `caller` is really a bottleneck)
- Running in JRuby 9.2.8 WITHOUT `caller`: 2 minutes
- Running in JRuby 9.2.8 WITH logging (i.e. `caller`): 31 minutes

I also considered the heap size (-J-Xmx4g), but monitoring the applications shows that both actually use less than 70% of the heap.

I also tried to enable frozen string literals (because we have many string literals) and "invoke dynamic", but both did not have any noticeable effect.

Any idea why the new JRuby is so much slower than the old one?

enebo:
@rovf so the bulk of your second application essentially is time to perform caller with 90s-120s (depending on version) doing things which is not caller?  If so the next step would be understanding what the non-caller part of the application is doing.  We have generally beat performance in 9.x over 1.7.x for any reasonably warmed up app.  Our main issue for 9.x has been short-lived scripts which take extra time. 

Since you are on windows that also makes me wonder about IO.  In 1.7 we used pure-Java IO but in 9 we switched to native IO for all but Windows.  It is possible we have introduced some extra overhead on IO for windows perhaps (guessing)?

In both apps perhaps there is something in common which explains why both lose to 1.7?

rovf:
[...]

PR #6236, opened 2020-05-21
Small optimization roundup

headius:
This PR contains a number of small optimizations from work on #5857.


headius:
Link to #6238.

headius:
I have finished the work I intended to do on this.

With these changes, the following forms are now as fast or faster than JRuby 1.7:

* Interpolation of a Fixnum into a string
* Interpolation of a String in general (limited testing, but heuristics now match 1.7's optimizations better)
* Symbol#to_proc for simple no-arg calls (most common form)

These changes appear to have improved the script from #6238 such that JRuby 9.3 is generally faster than 1.7:

```
[] ~/projects/jruby $ jruby b_aext.rb 10000
Rehearsal ------------------------------------------------------------------------
10000 iterations on 9.3.0.0-SNAPSHOT  13.250000   0.850000  14.100000 ( 11.956069)
-------------------------------------------------------------- total: 14.100000sec

10000 iterations on 9.3.0.0-SNAPSHOT  11.160000   0.100000  11.260000 ( 10.807729)

[] ~/projects/jruby $ jruby17 b_aext.rb 10000
Rehearsal --------------------------------------------------------------
10000 iterations on 1.7.27  12.660000   0.910000  13.570000 ( 12.029000)
---------------------------------------------------- total: 13.570000sec

10000 iterations on 1.7.27  11.710000   0.160000  11.870000 ( 11.354000)
```

Even with the great variability in this script (it is very object-heavy and subject to memory pressure) we are at least on par where we were 20%+ slower before.

There's additional work that could be done here:

* Use invokedynamic to improve size estimation and caching/inlining of interpolated elements
* Incorporate the simple regexp optimization from #6237

For purposes of addressing #6238, however, I will consider this work finished.

Request review from @enebo and @kares.

headius:
@enebo I did that rename mostly because it was obviously being used as a supertype for literal strings, but FrozenString did not (could not) implement it. By moving it to an interface there is now a common supertype for both mutable and frozen string literals, which simplifies code in other places.

PR #6237, opened 2020-05-21
Optimize simple regexp split

headius:
This is the same optimization that CRuby 2.7 did for simple regexp in String#split.

This has a tremendous impact on performance of patterns like `str.split(/,/)` but the code isn't passing all specs yet. Pushing this PR now so we don't lose the code.


headius:
The logic I ported over here is fine, but then it drops into older logic that turns a freshly-unwrapped string delimeter *back* into a regexp and doesn't really save us a great deal. The performance issue for which I explored this (#5857) has also been resolved by making other improvements.

 I'm going to ditch this, since it was a 2.7 change anyway and just doing a proper re-port of the split logic would be a better option. Low priority, will refile as a 9.4 item.

PR #6240, opened 2020-05-23
Fix failing jruby-jars CI build in GH actions

headius:
Trying 🙄


headius:
I fixed this in bc9867abc7245b72ae951dbf17089fc52624de20 by forcing the gem-maven-plugin to use a 9.3 snapshot. I also filed #6263 for us to switch to the released 9.3 when it's out.

Issue #6171, opened 2020-04-10
webrick tests fail on jruby


deivid-rodriguez:
We're getting [a random test failure we're getting when testing rubygems against jruby](https://github.com/rubygems/rubygems/issues/3451), which shows a backtrace inside webrick's code.

Since the failure is specific to jruby, I tried running `webrick` tests against `jruby`.

**Expected Behavior**

webrick tests pass under jruby.

**Actual Behavior**

You can see the current failures on [this WIP PR](https://github.com/ruby/webrick/pull/40) I opened at the webrick repo, where I just set up some changes so that the test suite at least runs properly.

current stable: https://github.com/ruby/webrick/pull/40/checks?check_run_id=577693801
current head: https://github.com/ruby/webrick/pull/40/checks?check_run_id=577693789

headius:
Hmm yes, we should get those tests running again and green. So many projects, so little time!

deivid-rodriguez:
Yeah... it happens :)

headius:
With #6246 I believe I've made the tests as green as possible. A few issues remain:

* OpenSSL callbacks are not being called in all the right places, leading to failures in test_ssl.rb's `test_sni`. See jruby/jruby-openssl#201 and jruby/jruby-openssl#202.
* Log-testing logic from `start_server` in WEBrick's test/lib/webrick/unit.rb is not thread-safe, and attempts to read two arrays that are being concurrently modified. This sometimes leads to a concurrency error.
* Server setup in the `test_restart_after_stop` test from test_server.rb is not properly ensuring threads are started and variables initialized in proper sequence, leading to races that sometimes cause the test to fail.

The first case will need work in jruby-openssl that's probably unlikely to happen without some help from the community. The other two cases require fixes in the WEBrick tests. Only the `test_sni` failure happens consistently, at this point.

I'm going to close this since there's new issues filed for the SSL problems and the other issues are in WEBrick's tests.

headius:
Oops, reopen until #6246 has been reviewed and merged.

headius:
#6246 was merged some time ago and resolved as many of these failures as possible. The remaining issues are not in JRuby.

Issue #6124, opened 2020-03-17
StackOverflow after several hours after upgrading from JRuby 9.2.9 to 9.2.11


highb:
We're seeing an issue with one of our Sinatra apps after upgrading from JRuby 9.2.9 to 9.2.11 where all of our threads start throwing StackOverflow exceptions after around 10-16 hours of uptime. [Here's a gist of some sample logs.](https://gist.github.com/highb/6100dc35efca935d436dcc6a48a34279) [The project is VMPooler (it's open source.)](https://github.com/puppetlabs/vmpooler) I haven't been able to track down what originally caused the error, but once it starts happening the service unable to complete any of it's work.

I have suspicions that it's related to a failed attempt to connect to our VSphere API that's transient in nature and the call is so deep in our call stack that when the exception happens and the new exception handling code in #6014 triggers, it ends up going over the maximum stack size. I wish I had more conclusive information but I haven't been able to find any at this time.

If you would like any additional info or have some troubleshooting steps you would like us to take, please ping me. Thank you!

**Environment Information**

We're running in a Docker container on Mesos.  

* JRuby version (`jruby -v`) and command line (flags, JRUBY_OPTS, etc)
This version has been running in production for several days with no issues:
```
root@4059599ae4cf:/# jruby -v
jruby 9.2.9.0 (2.5.7) 2019-10-30 458ad3e OpenJDK 64-Bit Server VM 25.242-b08 on 1.8.0_242-b08 +jit [linux-x86_64]
```

This is the version that is getting the StackOverflow after running for a few hours:
```
root@01bcd27113d4:/# jruby -v
jruby 9.2.11.0 (2.5.7) 2020-03-02 612d7a05a6 OpenJDK 64-Bit Server VM 25.242-b08 on 1.8.0_242-b08 +jit [linux-x86_64]
```

* Operating system and platform (e.g. `uname -a`)
The mesos agents that the containers are running on:
```
[brandon.high@cinext-mesosagent-prod-139 ~]$ uname -a
Linux cinext-mesosagent-prod-139 3.10.0-1062.9.1.el7.x86_64 #1 SMP Fri Dec 6 15:49:49 UTC 2019 x86_64 x86_64 x86_64 GNU/Linux
```

Other relevant info you may wish to add:
* Installed or activated gems
```
root@01bcd27113d4:/# cat Gemfile.lock
PATH
  remote: .
  specs:
    vmpooler (0.11.0)
      connection_pool (~> 2.2)
[...]

PR #6246, opened 2020-05-27
Fixes to get WEBrick tests greener

headius:
These are small fixes to address failing tests in WEBrick, as reported in #6171.

### DIFF
--- a/core/src/main/java/org/jruby/util/IOChannel.java
+++ b/core/src/main/java/org/jruby/util/IOChannel.java
@@ -36,6 +36,7 @@ import java.nio.channels.WritableByteChannel;
 
 import org.jruby.Ruby;
 import org.jruby.RubyString;
+import org.jruby.exceptions.ReadPartialBufferOverflowException;
 import org.jruby.runtime.CallSite;
 import org.jruby.runtime.ThreadContext;
 import org.jruby.runtime.builtin.IRubyObject;
@@ -193,9 +194,4 @@ public abstract class IOChannel implements Channel {
         }
     }
 
-    private class ReadPartialBufferOverflowException extends IOException {
-        public ReadPartialBufferOverflowException(String message) {
-            super(message);
-        }
-    }
 }
--- a/core/src/main/java/org/jruby/util/IOChannel.java
+++ b/core/src/main/java/org/jruby/util/IOChannel.java
@@ -84,12 +84,20 @@ public abstract class IOChannel implements Channel {
     }
 
     protected int read(CallSite read, ByteBuffer dst) throws IOException {
-        IRubyObject readValue = read.call(runtime.getCurrentContext(), io, io, runtime.newFixnum(dst.remaining()));
+        int remaining = dst.remaining();
+        IRubyObject readValue = read.call(runtime.getCurrentContext(), io, io, runtime.newFixnum(remaining));
         int returnValue = -1;
         if (!readValue.isNil()) {
             ByteList str = ((RubyString)readValue).getByteList();
-            dst.put(str.getUnsafeBytes(), str.getBegin(), str.getRealSize());
-            returnValue = str.getRealSize();
+            int realSize = str.getRealSize();
+
+            if (realSize > remaining) {
+                throw new ReadPartialBufferOverflowException(
+                        "error calling " + io.getType() + "#readpartial: requested " + remaining + " bytes but received " + realSize);
+            }
+
+            dst.put(str.getUnsafeBytes(), str.getBegin(), realSize);
+            returnValue = realSize;
         }
         return returnValue;
     }
@@ -182,6 +190,12 @@ public abstract class IOChannel implements Channel {
 
         public int write(ByteBuffer src) throws IOException {
             return write(write, src);
+        }
+    }
+
+    private class ReadPartialBufferOverflowException extends IOException {
+        public ReadPartialBufferOverflowException(String message) {
+            super(message);
         }
     }
 }
--- a/core/pom.rb
+++ b/core/pom.rb
@@ -41,11 +41,11 @@ project 'JRuby Core' do
 
   # exclude jnr-ffi to avoid problems with shading and relocation of the asm packages
   jar 'com.github.jnr:jnr-netdb:1.1.6', :exclusions => ['com.github.jnr:jnr-ffi']
-  jar 'com.github.jnr:jnr-enxio:0.27', :exclusions => ['com.github.jnr:jnr-ffi']
-  jar 'com.github.jnr:jnr-unixsocket:0.31', :exclusions => ['com.github.jnr:jnr-ffi']
-  jar 'com.github.jnr:jnr-posix:3.0.56', :exclusions => ['com.github.jnr:jnr-ffi']
+  jar 'com.github.jnr:jnr-enxio:0.28', :exclusions => ['com.github.jnr:jnr-ffi']
+  jar 'com.github.jnr:jnr-unixsocket:0.32', :exclusions => ['com.github.jnr:jnr-ffi']
+  jar 'com.github.jnr:jnr-posix:3.0.57', :exclusions => ['com.github.jnr:jnr-ffi']
   jar 'com.github.jnr:jnr-constants:0.9.15', :exclusions => ['com.github.jnr:jnr-ffi']
-  jar 'com.github.jnr:jnr-ffi:2.1.14'
+  jar 'com.github.jnr:jnr-ffi:2.1.15'
   jar 'com.github.jnr:jffi:${jffi.version}'
   jar 'com.github.jnr:jffi:${jffi.version}:native'
 
--- a/core/pom.xml
+++ b/core/pom.xml
@@ -96,7 +96,7 @@ DO NOT MODIFIY - GENERATED CODE
     <dependency>
       <groupId>com.github.jnr</groupId>
       <artifactId>jnr-enxio</artifactId>
-      <version>0.27</version>
+      <version>0.28</version>
       <exclusions>
         <exclusion>
           <artifactId>jnr-ffi</artifactId>
@@ -107,7 +107,7 @@ DO NOT MODIFIY - GENERATED CODE
     <dependency>
       <groupId>com.github.jnr</groupId>
       <artifactId>jnr-unixsocket</artifactId>
-      <version>0.31</version>
+      <version>0.32</version>
       <exclusions>
         <exclusion>
           <artifactId>jnr-ffi</artifactId>
@@ -118,7 +118,7 @@ DO NOT MODIFIY - GENERATED CODE
     <dependency>
       <groupId>com.github.jnr</groupId>
       <artifactId>jnr-posix</artifactId>
-      <version>3.0.56</version>
+      <version>3.0.57</version>
       <exclusions>
         <exclusion>
           <artifactId>jnr-ffi</artifactId>
@@ -140,7 +140,7 @@ DO NOT MODIFIY - GENERATED CODE
     <dependency>
       <groupId>com.github.jnr</groupId>
       <artifactId>jnr-ffi</artifactId>
-      <version>2.1.14</version>
+      <version>2.1.15</version>
     </dependency>
     <dependency>
       <groupId>com.github.jnr</groupId>
--- a/core/src/main/java/org/jruby/RubyIO.java
+++ b/core/src/main/java/org/jruby/RubyIO.java
@@ -4233,11 +4233,12 @@ public class RubyIO extends RubyObject implements IOEncodable, Closeable, Flusha
         Channel channel1 = null;
         Channel channel2 = null;
 
-        if (args.length >= 3) {
+        if (args.length >= 3 && !args[2].isNil()) {
             length = args[2].convertToInteger();
-            if (args.length == 4) {
-                offset = args[3].convertToInteger();
-            }
+        }
+
+        if (args.length == 4 && !args[3].isNil()) {
+            offset = args[3].convertToInteger();
         }
 
         IOSites sites = sites(context);
--- a/core/src/main/java/org/jruby/exceptions/ReadPartialBufferOverflowException.java
+++ b/core/src/main/java/org/jruby/exceptions/ReadPartialBufferOverflowException.java
@@ -1,0 +1,17 @@
+package org.jruby.exceptions;
+
+import java.io.IOException;
+import java.nio.ByteBuffer;
+
+/**
+ * Represents a buffer overflow (too much data returned) when calling `readpartial` on an IO-like object.
+ *
+ * See jruby/jruby#6246 and ruby/webrick#43
+ *
+ * @see org.jruby.util.IOChannel.IOReadableByteChannel#read(ByteBuffer)
+ */
+public class ReadPartialBufferOverflowException extends IOException {
+    public ReadPartialBufferOverflowException(String message) {
+        super(message);
+    }
+}
--- a/core/src/main/java/org/jruby/ext/socket/RubyBasicSocket.java
+++ b/core/src/main/java/org/jruby/ext/socket/RubyBasicSocket.java
@@ -95,7 +95,7 @@ public class RubyBasicSocket extends RubyIO {
 
     public RubyBasicSocket(Ruby runtime, RubyClass type) {
         super(runtime, type);
-        doNotReverseLookup = true;
+        doNotReverseLookup = runtime.isDoNotReverseLookupEnabled();
     }
 
     @JRubyMethod(meta = true)
--- a/core/src/main/java/org/jruby/ext/timeout/Timeout.java
+++ b/core/src/main/java/org/jruby/ext/timeout/Timeout.java
@@ -89,7 +89,7 @@ public class Timeout {
 
     @JRubyMethod(module = true)
     public static IRubyObject timeout(final ThreadContext context, IRubyObject recv, IRubyObject seconds, IRubyObject exceptionType, Block block) {
-        return timeout(context, recv, seconds, exceptionType, defaultTimeoutMessage(context), block);
+        return timeout(context, recv, seconds, exceptionType, context.nil, block);
     }
 
     @JRubyMethod(module = true)
@@ -106,9 +106,11 @@ public class Timeout {
         final AtomicBoolean latch = new AtomicBoolean(false);
 
         final Object id = exceptionType.isNil() ? new Object() : null;
+        final RubyString exceptionMessage = message.isNil() ? defaultTimeoutMessage(context) : message.convertToString();
+
         Runnable timeoutRunnable = id != null ?
-                TimeoutTask.newAnonymousTask(currentThread, timeout, latch, id, message.convertToString()) :
-                TimeoutTask.newTaskWithException(currentThread, timeout, latch, exceptionType, message.convertToString());
+                TimeoutTask.newAnonymousTask(currentThread, timeout, latch, id, exceptionMessage) :
+                TimeoutTask.newTaskWithException(currentThread, timeout, latch, exceptionType, exceptionMessage);
 
         ScheduledThreadPoolExecutor executor = (ScheduledThreadPoolExecutor) timeout.getInternalVariable("__executor__");
 
--- a/core/src/main/java/org/jruby/ir/instructions/UnresolvedSuperInstr.java
+++ b/core/src/main/java/org/jruby/ir/instructions/UnresolvedSuperInstr.java
@@ -1,5 +1,6 @@
 package org.jruby.ir.instructions;
 
+import org.jcodings.specific.USASCIIEncoding;
 import org.jruby.RubyInstanceConfig;
 import org.jruby.RubySymbol;
 import org.jruby.ir.IRFlags;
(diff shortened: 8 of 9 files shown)


headius:
The failure in webrick's test_cgi appears to be a bug in how we handle multibyte content passing through environment variables.

The test tries to pass through a URI like "/webrick.cgi/%A4%DB%A4%B2/%A4%DB%A4%B2" using a CGI subprocess. The URI is parsed out and passed via an environment variable, which in JRuby ends up passing through a Java String. That String conversion appears to mangle this string, probably because it doesn't seem to be valid UTF-8.

The code below is responsible, but I'm not sure how to get these env values to pass through to setenv without getting mangled.

https://github.com/jruby/jruby/blob/6c0f3370f5b550cfd857483e4c82fefff8daf27f/core/src/main/java/org/jruby/RubyGlobal.java#L525

headius:
I don't believe we'll be able to fix the ENV issues without moving completely away from Java's own `System.getenv`. In order to provide strings, that API must decode the environment entries, which wipes out any entries that can't be decoded using the system's default character encoding.

I will file a separate issue for this and likely expand the WEBrick test to also exclude the "java" platform.

headius:
See #6248 for the ENV issue.

headius:
The next failure, test_big_bodies in test_httpproxy, took some untangling. There's many layers to this test, but ultimate I found a WEBrick bug here:

https://github.com/ruby/webrick/blob/6b6990ec81479160d53d81310c05ab4dc508b199/lib/webrick/httprequest.rb#L276-L277

This is a pretty big assumption about how `copy_stream` is implemented, and at least in the case of JRuby, it's not valid. The contract of `readpartial` is not properly maintained by WEBrick here, which causes JRuby internals to overflow a buffer (because too much data came back) and ultimately fail the test.

I will file a WEBrick bug (and PR if possible) to fix this.

headius:
The final CJK issue appears to be a bug in our `File.expand_path` logic. The test creates a temp dir with the "あ" character similar to below:

```
/var/folders/cq/ylcgmnn556x33f5hsqd0h54h0000gn/T/あ20200528-99322-ael0q9
```

It proceeds to run with that directory as the server's file base, and writes another file with the "あ" character in it. However when attempting to search for that file in the server's base, it uses `File.expand_path`, which results in a path like this:

```
/var/folders/cq/ylcgmnn556x33f5hsqd0h54h0000gn/T/?20200528-99322-ael0q9
```

We then fail to see this as a directory and fail the overall file search.

headius:
Reduced case:

```
$ jruby -rtmpdir -e "Dir.mktmpdir('あ') {|dir| p File.expand_path(dir.force_encoding('ASCII-8BIT'))}"
"/var/folders/cq/ylcgmnn556x33f5hsqd0h54h0000gn/T/?20200528-99474-ov6cnb"
```

The base path for the server appears to be ASCII-8BIT, which ends up breaking JRuby's logic for `expand_path` that tries to decode it. I'm not sure if this is our bug or WEBrick's.

headius:
WEBrick does indeed appear to be forcing this path to be "binary", which leads to the `expand_path` bug. However if I remove the `b` call here I get an invalid UTF-8 error later on. There's clearly some issues negotiating string and path encodings here.

https://github.com/ruby/webrick/blob/83cf440858b46ce91866165bf0a6db580f6e6215/lib/webrick/httpservlet/filehandler.rb#L327

headius:
I've rebased this branch off master after merging in #6124 

That PR represents a number of fixes to avoid mangling file paths that are marked as binary but which actually contain legitimate multibyte characters. There are two strategies taken:

* When the actual character content of the strings does not matter, as in `File.expand_path` (all interesting elements there are single-byte characters like `/` or `..`), we treat binary strings as raw byte data and transcode as ISO-8859-1 characters. This allows all multibyte characters to remain untouched and only the interesting path characters are manipulated.
* When the character content of the strings does matter, such as when we are caling JDK functions that demand a properly-decoded String, we treat binary paths as though they are actually encoded like the system default (filesystem) encoding. This will not always be the right call, but since these paths are intended to interact with the local system it should generally be a safe bet that they will be filesystem-encoded.

With these additional changes the CJK failure is now passing. I will run some final checks but I believe I've done all I can on the JRuby side for the WEBrick test suite.

headius:
With all these changes and the fix from ruby/webrick#45 I am now down to Only a couple failures. I include them below. The test_sni failure appears to happen consistently, and the other two are sporadic (only saw the concurrency issue once in dozens of runs).

```
TestWEBrickHTTPProxy#test_proxy [/Users/headius/projects/webrick/test/webrick/utils.rb:72]:
exceptions on 1 threads:
#<Thread:0x4a3329b9@/Users/headius/projects/webrick/test/webrick/utils.rb:65 dead>:
org/jruby/RubyArray.java:3311:in `+': Detected invalid array contents due to unsynchronized modifications with concurrent users (ConcurrencyError)
	from /Users/headius/projects/webrick/test/webrick/utils.rb:49:in `block in start_server'
	from /Users/headius/projects/webrick/test/webrick/test_httpproxy.rb:69:in `block in test_proxy'
	from /Users/headius/projects/jruby/lib/ruby/stdlib/net/http.rb:1529:in `block in transport_request'
	from /Users/headius/projects/jruby/lib/ruby/stdlib/net/http/response.rb:165:in `reading_body'
	from /Users/headius/projects/jruby/lib/ruby/stdlib/net/http.rb:1528:in `transport_request'
	from /Users/headius/projects/jruby/lib/ruby/stdlib/net/http.rb:1490:in `request'
	from /Users/headius/projects/jruby/lib/ruby/stdlib/net/http.rb:1483:in `block in request'
	from /Users/headius/projects/jruby/lib/ruby/stdlib/net/http.rb:924:in `start'
	from /Users/headius/projects/jruby/lib/ruby/stdlib/net/http.rb:1481:in `request'
	from /Users/headius/projects/webrick/test/webrick/test_httpproxy.rb:68:in `block in test_proxy'
	from /Users/headius/projects/webrick/test/webrick/utils.rb:67:in `block in start_server'

TestWEBrickHTTPS#test_sni [/Users/headius/projects/webrick/test/webrick/utils.rb:72]:
exceptions on 1 threads:
#<Thread:0x6b54655f@/Users/headius/projects/webrick/test/webrick/utils.rb:65 dead>:
/Users/headius/projects/webrick/test/lib/minitest/unit.rb:201:in `assert': Expected /\A([.+*]+\n)+\z/ to match "". (MiniTest::Assertion)
	from /Users/headius/projects/webrick/test/lib/test/unit/assertions.rb:37:in `assert'
	from /Users/headius/projects/webrick/test/lib/minitest/unit.rb:293:in `assert_match'
	from /Users/headius/projects/webrick/test/webrick/test_https.rb:90:in `block in test_sni'
	from /Users/headius/projects/webrick/test/webrick/utils.rb:67:in `block in start_server'

TestWEBrickServer#test_restart_after_stop [org/jruby/internal/runtime/methods/CompiledIRMethod.java:109]:
exceptions on 1 threads:
#<Thread:0x52b6319f dead>:
java/lang/Thread.java:1559:in `getStackTrace': killed thread (ThreadError)
	from org/jruby/Ruby.java:4251:in `newRaiseException'
	from org/jruby/Ruby.java:3915:in `newThreadError'
	from org/jruby/RubyThread.java:1312:in `wakeup'
	from /Users/headius/projects/webrick/test/webrick/test_server.rb:143:in `block in test_restart_after_stop'
	from /Users/headius/projects/webrick/lib/webrick/server.rb:332:in `call_callback'
	from /Users/headius/projects/webrick/lib/webrick/server.rb:161:in `block in start'
	from /Users/headius/projects/webrick/lib/webrick/server.rb:32:in `start'
	from /Users/headius/projects/webrick/lib/webrick/server.rb:157:in `start'
	from /Users/headius/projects/webrick/test/webrick/test_server.rb:153:in `block in test_restart_after_stop'
```

And some full traces below:
```
TestWEBrickHTTPS#test_sni [org/jruby/internal/runtime/methods/CompiledIRMethod.java:109]:
exceptions on 1 threads:
#<Thread:0x693e4d19 dead>:
java/lang/Thread.java:1559:in `getStackTrace': Expected /\A([.+*]+\n)+\z/ to match "". (MiniTest::Assertion)
	from org/jruby/runtime/backtrace/TraceType.java:242:in `getBacktraceData'
	from org/jruby/runtime/backtrace/TraceType.java:53:in `getBacktrace'
	from org/jruby/RubyException.java:371:in `captureBacktrace'
	from org/jruby/exceptions/RaiseException.java:115:in `preRaise'
	from org/jruby/exceptions/RaiseException.java:65:in `<init>'
	from org/jruby/exceptions/Exception.java:38:in `<init>'
	from org/jruby/RubyException.java:151:in `constructThrowable'
	from org/jruby/RubyException.java:349:in `toThrowable'
	from org/jruby/RubyKernel.java:952:in `raise'
	from org/jruby/RubyKernel$INVOKER$s$0$3$raise.gen:-1:in `call'
	from org/jruby/internal/runtime/methods/DynamicMethod.java:215:in `call'
	from org/jruby/internal/runtime/methods/DynamicMethod.java:211:in `call'
	from org/jruby/runtime/callsite/CachingCallSite.java:396:in `cacheAndCall'
	from org/jruby/runtime/callsite/CachingCallSite.java:205:in `call'
	from Users/headius/projects/webrick/test/lib/minitest//Users/headius/projects/webrick/test/lib/minitest/unit.rb:201:in `invokeOther5:raise'
	from /Users/headius/projects/webrick/test/lib/minitest/unit.rb:201:in `assert'
	from org/jruby/internal/runtime/methods/CompiledIRMethod.java:109:in `call'
	from org/jruby/internal/runtime/methods/MixedModeIRMethod.java:70:in `call'
	from org/jruby/ir/runtime/IRRuntimeHelpers.java:1229:in `unresolvedSuper'
	from org/jruby/ir/runtime/IRRuntimeHelpers.java:1212:in `unresolvedSuperSplatArgs'
	from Users/headius/projects/webrick/test/lib/test/unit//Users/headius/projects/webrick/test/lib/test/unit/assertions.rb:37:in `invokeSuper10:-dynamic-super_target-'
	from /Users/headius/projects/webrick/test/lib/test/unit/assertions.rb:37:in `assert'
	from org/jruby/internal/runtime/methods/CompiledIRMethod.java:109:in `call'
	from org/jruby/internal/runtime/methods/CompiledIRMethod.java:145:in `call'
	from org/jruby/internal/runtime/methods/MixedModeIRMethod.java:175:in `call'
	from org/jruby/internal/runtime/methods/DynamicMethod.java:211:in `call'
	from org/jruby/runtime/callsite/CachingCallSite.java:396:in `cacheAndCall'
	from org/jruby/runtime/callsite/CachingCallSite.java:205:in `call'
	from Users/headius/projects/webrick/test/lib/minitest//Users/headius/projects/webrick/test/lib/minitest/unit.rb:293:in `invokeOther8:assert'
	from /Users/headius/projects/webrick/test/lib/minitest/unit.rb:293:in `assert_match'
	from org/jruby/internal/runtime/methods/CompiledIRMethod.java:109:in `call'
	from org/jruby/internal/runtime/methods/CompiledIRMethod.java:145:in `call'
	from org/jruby/internal/runtime/methods/MixedModeIRMethod.java:175:in `call'
	from org/jruby/internal/runtime/methods/DynamicMethod.java:211:in `call'
	from org/jruby/runtime/callsite/CachingCallSite.java:396:in `cacheAndCall'
	from org/jruby/runtime/callsite/CachingCallSite.java:205:in `call'
	from Users/headius/projects/webrick/test/webrick//Users/headius/projects/webrick/test/webrick/test_https.rb:90:in `invokeOther12:assert_match'
	from /Users/headius/projects/webrick/test/webrick/test_https.rb:90:in `block in test_sni'
	from org/jruby/runtime/CompiledIRBlockBody.java:141:in `callDirect'
	from org/jruby/runtime/IRBlockBody.java:64:in `call'
	from org/jruby/runtime/Block.java:143:in `call'
	from org/jruby/RubyProc.java:291:in `call'
	from org/jruby/RubyProc$INVOKER$i$call.gen:-1:in `call'
	from org/jruby/internal/runtime/methods/JavaMethod.java:349:in `call'
	from org/jruby/runtime/callsite/CachingCallSite.java:172:in `call'
	from Users/headius/projects/webrick/test/webrick//Users/headius/projects/webrick/test/webrick/utils.rb:67:in `invokeOther11:yield'
	from /Users/headius/projects/webrick/test/webrick/utils.rb:67:in `block in start_server'
	from org/jruby/runtime/CompiledIRBlockBody.java:141:in `callDirect'
	from org/jruby/runtime/IRBlockBody.java:64:in `call'
	from org/jruby/runtime/IRBlockBody.java:58:in `call'
	from org/jruby/runtime/Block.java:139:in `call'
	from org/jruby/RubyProc.java:318:in `call'
	from org/jruby/internal/runtime/RubyRunnable.java:105:in `run'
	from java/lang/Thread.java:748:in `run'

TestWEBrickServer#test_restart_after_stop [org/jruby/internal/runtime/methods/CompiledIRMethod.java:109]:
exceptions on 1 threads:
#<Thread:0x52b6319f dead>:
java/lang/Thread.java:1559:in `getStackTrace': killed thread (ThreadError)
	from org/jruby/runtime/backtrace/TraceType.java:242:in `getBacktraceData'
	from org/jruby/runtime/backtrace/TraceType.java:53:in `getBacktrace'
	from org/jruby/RubyException.java:371:in `captureBacktrace'
	from org/jruby/exceptions/RaiseException.java:115:in `preRaise'
	from org/jruby/exceptions/RaiseException.java:65:in `<init>'
	from org/jruby/exceptions/Exception.java:38:in `<init>'
	from org/jruby/exceptions/StandardError.java:38:in `<init>'
	from org/jruby/exceptions/ThreadError.java:38:in `<init>'
	from org/jruby/RubyThreadError.java:51:in `constructThrowable'
	from org/jruby/RubyException.java:349:in `toThrowable'
	from org/jruby/exceptions/RaiseException.java:81:in `from'
	from org/jruby/Ruby.java:4251:in `newRaiseException'
	from org/jruby/Ruby.java:3915:in `newThreadError'
	from org/jruby/RubyThread.java:1312:in `wakeup'
	from org/jruby/RubyThread$INVOKER$i$0$0$wakeup.gen:-1:in `call'
	from org/jruby/runtime/callsite/CachingCallSite.java:141:in `call'
	from Users/headius/projects/webrick/test/webrick//Users/headius/projects/webrick/test/webrick/test_server.rb:143:in `invokeOther0:wakeup'
	from /Users/headius/projects/webrick/test/webrick/test_server.rb:143:in `block in test_restart_after_stop'
	from org/jruby/runtime/CompiledIRBlockBody.java:141:in `callDirect'
	from org/jruby/runtime/IRBlockBody.java:64:in `call'
	from org/jruby/runtime/Block.java:143:in `call'
	from org/jruby/RubyProc.java:275:in `call'
	from org/jruby/RubyProc$INVOKER$i$call.gen:-1:in `call'
	from org/jruby/internal/runtime/methods/DynamicMethod.java:177:in `call'
	from org/jruby/runtime/callsite/CachingCallSite.java:74:in `call'
	from Users/headius/projects/webrick/lib/webrick//Users/headius/projects/webrick/lib/webrick/server.rb:332:in `invokeOther2:call'
	from /Users/headius/projects/webrick/lib/webrick/server.rb:332:in `call_callback'
	from org/jruby/internal/runtime/methods/CompiledIRMethod.java:109:in `call'
	from org/jruby/internal/runtime/methods/CompiledIRMethod.java:132:in `call'
	from org/jruby/internal/runtime/methods/MixedModeIRMethod.java:140:in `call'
	from org/jruby/internal/runtime/methods/DynamicMethod.java:203:in `call'
	from org/jruby/runtime/callsite/CachingCallSite.java:172:in `call'
	from Users/headius/projects/webrick/lib/webrick//Users/headius/projects/webrick/lib/webrick/server.rb:161:in `invokeOther28:call_callback'
	from /Users/headius/projects/webrick/lib/webrick/server.rb:161:in `block in start'
	from /Users/headius/projects/webrick/lib/webrick/server.rb:32:in `start'
	from org/jruby/internal/runtime/methods/CompiledIRMethod.java:122:in `call'
	from org/jruby/internal/runtime/methods/MixedModeIRMethod.java:105:in `call'
	from org/jruby/runtime/callsite/CachingCallSite.java:151:in `call'
	from org/jruby/runtime/callsite/CachingCallSite.java:160:in `callIter'
	from Users/headius/projects/webrick/lib/webrick//Users/headius/projects/webrick/lib/webrick/server.rb:157:in `invokeOther84:start'
	from /Users/headius/projects/webrick/lib/webrick/server.rb:157:in `start'
	from org/jruby/internal/runtime/methods/CompiledIRMethod.java:122:in `call'
	from org/jruby/internal/runtime/methods/MixedModeIRMethod.java:105:in `call'
	from org/jruby/internal/runtime/methods/DynamicMethod.java:195:in `call'
	from org/jruby/runtime/callsite/CachingCallSite.java:141:in `call'
	from Users/headius/projects/webrick/test/webrick//Users/headius/projects/webrick/test/webrick/test_server.rb:153:in `invokeOther1:start'
	from /Users/headius/projects/webrick/test/webrick/test_server.rb:153:in `block in test_restart_after_stop'
	from org/jruby/runtime/CompiledIRBlockBody.java:141:in `callDirect'
	from org/jruby/runtime/IRBlockBody.java:64:in `call'
	from org/jruby/runtime/IRBlockBody.java:58:in `call'
	from org/jruby/runtime/Block.java:139:in `call'
	from org/jruby/RubyProc.java:318:in `call'
	from org/jruby/internal/runtime/RubyRunnable.java:105:in `run'
	from java/lang/Thread.java:748:in `run'
```

headius:
The line of code in question for the concurrency error is adding two arrays that are being concurrently modified.

From test/lib/webrick/unit.rb:

```ruby
  def start_server(klass, config={}, log_tester=DefaultLogTester, &block)
    log_ary = []
    access_log_ary = []
>>> log = proc { "webrick log start:\n" + (log_ary+access_log_ary).join.gsub(/^/, "  ").chomp + "\nwebrick log end" }
    config = ({
      :BindAddress => "127.0.0.1", :Port => 0,
      :ServerType => Thread,
      :Logger => WEBrick::Log.new(log_ary, WEBrick::BasicLog::WARN),
      :AccessLog => [[access_log_ary, ""]]
    }.update(config))
    ...
```

The arrays here blew up just this one time, most likely due to the backing array or offset values getting changed out mid-addition. I will not attempt to fix this, but it's good justification for us to finally make concurrent array access robust enough to avoid locks.

headius:
The (first) failure in `test_sni` is due to `OpenSSL::PKey::RSA#initialize` not passing the block through to the generation process, which is supposed to yield each byte in turn.

I filed jruby/jruby-openssl#201 for this issue, but after some investigation I've determined there's no way to implement this callback on top of the Bouncy Castle RSA key pair generation. I believe this particular test will need to be skipped on JRuby.

headius:
With the key pair generator callback assertion commented out, the remaining assertions also fail due to missing `verify_callback` behavior in jruby-openssl. This is not a new problem and it is known that this callback does not fully implement the Ruby openssl library spec. I have opened jruby/jruby-openssl#202 for this case.

I believe that `test_sni` will not be fixable with JRuby's current openssl implementation.

headius:
The final issue here, in `test_restart_after_stop`, seems likely to be a race in the setup and teardown of the test server.

https://github.com/ruby/webrick/blob/2c13beca2c25dc099b7ed20dbe4d7ede77dfef86/test/webrick/test_server.rb#L134-L163

First, I have seen this error:

```
test_server.rb:143:in `block in test_restart_after_stop': undefined method `wakeup' for nil:NilClass
```

This is caused when the `client_thread` variable is not assigned before the server thread starts the server, which calls the start callback. It can be fixed with this diff, I believe:

```diff
diff --git a/test/webrick/test_server.rb b/test/webrick/test_server.rb
index 8162a18..0d944a0 100644
--- a/test/webrick/test_server.rb
+++ b/test/webrick/test_server.rb
@@ -149,14 +149,15 @@ class TestWEBrickServer < Test::Unit::TestCase
       :Port => 0,
       :Logger => warn_flunk)
     2.times {
-      server_thread = Thread.start {
-        server.start
-      }
       client_thread = Thread.start {
         sleep 0.1 until server.status == :Running || !server_thread.status
         server.stop
         sleep 0.1 until server.status == :Stop || !server_thread.status
       }
+      server_thread = Thread.start {
+        server.start
+      }
+
       assert_join_threads([client_thread, server_thread])
     }
   end
```

The second failure I have seen is the "killed thread" failure mentioned above, which also appears to be happening during the start callback. I suspect this would either be fixed by the above, or by adding some confirmation that the threads in question have actually started. In short, it's another race on thread status.

There are no further changes I can make in JRuby to make this test green. It fails sporadically due to races in the test itself.

headius:
This PR is ready for review!

core/src/main/java/org/jruby/util/IOChannel.java
```diff
@@ -184,4 +192,10 @@ public int write(ByteBuffer src) throws IOException {
             return write(write, src);
         }
     }
+
+    private class ReadPartialBufferOverflowException extends IOException {
```
kares, line 196: nitpick: make the class static

headius: You're right, I'll fix it.

headius: I opted to make it a public, top-level class in our `exceptions` package, since there may also be a need to catch this specific exception some day.

PR #6249, opened 2020-05-29
Encoding experiments with file path negotiation

headius:
This PR holds some experiments in making file path negotiation (expand_path, fnmatch) work properly with strings that are encoded as "binary" or "ASCII-8BIT".

In these cases it appears we should largely ignore the encoding of the characters and treat it as raw bytes, since it's not possible for us to guess the encoding. However some of the logic in these methods uses Java strings, and therefore needs to be decoded into UTF-16 characters. We will need to balance the binary cases with the properly-encoded cases as best we can, or otherwise rewrite this logic altogether to use only the raw byte form.

### DIFF
--- a/core/src/main/java/org/jruby/RubyDir.java
+++ b/core/src/main/java/org/jruby/RubyDir.java
@@ -171,7 +171,7 @@ public class RubyDir extends RubyObject implements Closeable {
 
         this.encoding = encoding;
 
-        String adjustedPath = RubyFile.adjustRootPathOnWindows(runtime, newPath.toString(), null);
+        String adjustedPath = RubyFile.getAdjustedPath(context, newPath);
         checkDirIsTwoSlashesOnWindows(getRuntime(), adjustedPath);
 
         this.dir = JRubyFile.createResource(context, adjustedPath);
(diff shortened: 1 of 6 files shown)


headius:
The first fixes here relate to #6246 in the failing "CJK" test that depends on treating a local file path and a remote URI path as binary and still locate the file in question.

My changes treat binary or ASCII-8BIT paths provided to `File.expand_path` as ISO-8859-1 bytes, eliminating any chance for their multibyte characters to be improperly decoded. The first commit does this treatment to all paths passed to `expand_path`, and the second narrows it to just the cases where one or both incoming paths are already marked as ASCII-8BIT. This fixes part of the failing CJK test by allowing the `expand_path` negotiation to proceed without mangling.

headius:
After force-pushing I have committed to the narrowed binary-only change and also renamed the confusing "decodeISO" logic to "decodeRaw" to more clearly indicate that it's decoding raw byte content.

headius:
The additional fixes deal with places where we use Java strings for paths and can't easily avoid doing so, because those strings feed into JDK APIs that expect properly-decoded characters. When we have such a string in hand, marked as binary, this logic will attempt to decode it as the default system encoding. This may not be correct, if the string is both marked as binary and has a different encoding from the system, but when dealing with file paths it seems a reasonable assumption. It will at least be correct more often than treating the path as ISO0-8859-1 bytes when it may actually have multibyte characters.

This fixes a large number of APIs that previous just decoded binary strings as ISO-8859-1:

* Dir entries and globbing
* File test methods like `directory?`
* File-opening APIs like `File.new`

The logic has largely been centralized in RubyFile.getAdjustedPath, which does all of the following:

* Extracts an already-decoded path if the given argument is an IO object
* Coerces a path using the `get_path` logic for other object types
* Checks for null bytes
* Handles adjusting the CWD and windows path elements
* Decodes binary strings as though they are system-encoded (the fix described above)

PR #6252, opened 2020-05-30
Fix failling TestUTF16#test_succ

k77ch7:
This fixes failling [TestUTF16#test_succ ](https://github.com/jruby/jruby/blob/master/test/mri/ruby/enc/test_utf16.rb#L267-L273) .


headius:
Thank you!

PR #6256, opened 2020-06-02
Raise proper error for missing URLResource target

headius:
See #6219.


headius:
Retargeted to 9.2.12 and cherry-picked into #6285.

PR #6257, opened 2020-06-02
Split up main test and clean up output

headius:
MainTestSuite included a large set of test cases and took the bulk of time to run, but because it wrapped those cases we got no output from them and had to dig in whenever something failed.

In addition, several tests printed output that was not necessary for testing, and Red Bridge's `EmbedEvalUnitImpl` unconditionally called for the runtime to print any errors raised **in addition to actually raising those errors to the user**.

This PR cleans up both situations.


headius:
Failing suite doesn't seem to be related to my changes.

PR #6262, opened 2020-06-03
Fix jruby jars deploy

headius:
Wrestling with CI


headius:
Not sure if this will run into problems, but 9.2.11.1 was having problems as seen here:

https://travis-ci.org/github/jruby/jruby/jobs/694452697

We should change this to the non-snapshot 9.3 version when it is released. I will file a separate issue for that.

PR #6264, opened 2020-06-04
Expose Java's built-in monitor synchronization as a JRuby API

headius:
I'm not sure why we've never done this before, but since there's
really no way to synchronize against an object's Java monitor from
Ruby this API seems like a gap in JRuby. Here I've added the
following JRuby utility functions:

* JRuby.synchronized(arg) {} which works like Java's synchronized
  statement
* JRuby.wait(arg [, millis [, nanos]]) like Object.wait
* JRuby.notify(arg) and notifyAll(arg) like Object.notify[All]

This question came up in the JRuby chat, and it seems like a
useful and appropriate addition to the JRuby module.

### DIFF
--- a/core/src/main/java/org/jruby/ext/jruby/JRubyUtilLibrary.java
+++ b/core/src/main/java/org/jruby/ext/jruby/JRubyUtilLibrary.java
@@ -39,6 +39,7 @@ import org.jruby.anno.JRubyMethod;
 import org.jruby.ast.util.ArgsUtil;
 import org.jruby.exceptions.RaiseException;
 import org.jruby.javasupport.Java;
+import org.jruby.runtime.Block;
 import org.jruby.runtime.ThreadContext;
 import org.jruby.runtime.builtin.IRubyObject;
 import org.jruby.runtime.load.BasicLibraryService;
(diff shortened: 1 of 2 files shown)


enebo:
@headius I thought we had it but I realize now if was the module to allow all methods to synch.   Seems like a long overdue feature.

headius:
@enebo Yeah, kind of a "duh" moment for me too. The Synchronized module is still out there, but you have to include it into a class or extend it into an object, and then it synchronizes every method call.

kares:
indeed very useful, wonder if these should be handling Java wrappers in a special way or not.
I think proxies have `wait`/`notify` inherited from java.lang.Object so that should do, users passing a Java proxy here just get semantics on the wrapper (a know maybe even desired "feature" for now).

headius:
@kares No I think your first instinct was right...it should synchronize against the wrapped object. If a user calls synchronize against a Java object, it must actually be locked for other consumers elsewhere in the JVM trying to synchronize against the same object. As far as anyone using JRuby is concerned, there is no wrapper.

headius:
I've fixed sync logic for wrapped Java objects. All that remains is writing some tests for this API.

Issue #6218, opened 2020-05-13
Errors running multiple independent ScriptingContainers in parallel


jsolomon8080:
jruby 9.2.11.1 using the jruby-complete-9.2.11.1.jar
```
% java -version
openjdk version "11.0.4" 2019-07-16
OpenJDK Runtime Environment AdoptOpenJDK (build 11.0.4+11)
OpenJDK 64-Bit Server VM AdoptOpenJDK (build 11.0.4+11, mixed mode)

% uname -a
Linux localhost 3.10.0-1062.12.1.el7.x86_64 #1 SMP Tue Feb 4 23:02:59 UTC 2020 x86_64 x86_64 x86_64 GNU/Linux
% lsb_release -a
LSB Version:    :core-4.1-amd64:core-4.1-noarch
Distributor ID: CentOS
Description:    CentOS Linux release 7.6.1810 (Core) 
Release:        7.6.1810
Codename:       Core
```
Our environment is ScriptingContainer only. We had a ruby on rails web app that was ported to pure java but we have ~100 ERB templates and we use embedded jruby to evaluate them. In production, we have basically very few instances and threads using a ScriptingContainer. However, our unit tests use gradle and many tests touch the ERB pipeline and start ScirptingContainers.

We are finally updating to java11 from java8 and that has forced us to upgrade from jruby-1.6.7 (from 2011!) to the latest jruby (jruby-9.2.11.1). We have been using the same embedded architecture for 3-4 years and never had this problem with jruby-1.6.7.

Our environment is Linux and Windows and we see the same error on both platforms. In fact, probably more on Windows than Linux. I'm showing this error with CentOS 7.6 and java11 but it also repros with other versions CentOS and Windows and on java8.

This multi-threaded issue isn't super common in our CI/CD environment but I crafted a test case to make it common. I start X ScriptingContainers on Y threads where X=500 and Y=40. Again, not something we'd ever do, but it shows the behavior easily. Each ScriptingContainer is only accessed by the single thread that created it. Here is the java test case:

```
import java.util.ArrayList;
import java.util.Collection;
import java.util.List;
import java.util.concurrent.Callable;
import java.util.concurrent.Executors;

import org.jruby.embed.LocalContextScope;
import org.jruby.embed.LocalVariableBehavior;
import org.jruby.embed.ScriptingContainer;
import org.junit.Test;

import com.google.common.util.concurrent.Futures;
import com.google.common.util.concurrent.ListenableFuture;
[...]

Issue #3657, opened 2016-02-11
jruby-9.0.5.0-complete.jar: leaves stale jffi*.dll files in temp folder


perlun:
This seems very similar to #1237, which was resolved on the JRuby 1.7 track a while ago.

Symptoms: on a customer Windows server, the C: folder was filled up and the application(s) stopped working correctly. Upon careful examination, we discovered loads of temp files.

![image](https://cloud.githubusercontent.com/assets/630613/12975576/791b6f38-d0c5-11e5-9cff-dc9d78912cc1.png)

The jruby-\* subfolders have content that looks like this:

![image](https://cloud.githubusercontent.com/assets/630613/12975587/87e6d822-d0c5-11e5-9a05-fa4e8c4e5878.png)

(I treat this as two different issues, just leaving the images here for the sake of completeness. Will follow this up a bit more myself but if anyone has any obvious ideas, _please let me know_.)

---

The jffi files turned out to be very simple to reproduce locally:

```
$ java -jar jruby-complete-9.0.5.0.jar -e "require 'ffi'"
```

This will create a jffi file in the temp folder (in my case, C:\msys64\tmp).

I also tried downloading `jruby-bin-9.0.5.0.zip` and running with that. Surprisingly enough, _the result is different_:

```
plundberg@DESKTOP-NVMNDAM MSYS /c/jruby-9.0.5.0
$ bin/jruby -v
jruby 9.0.5.0 (2.2.3) 2016-01-26 7bee00d Java HotSpot(TM) 64-Bit Server VM 25.66-b18 on 1.8.0_66-b18 +jit [Windows 10-amd64]

plundberg@DESKTOP-NVMNDAM MSYS /c/jruby-9.0.5.0
$ bin/jruby -e "require 'ffi'"
```

This does _not_ create any stale temp files whatsoever. :astonished: 

Any clues?

perlun:
(I looked at the other problem, with the stale `.jar` files left in a temp folder. It was very similar in nature: `require 'openssl'` and you get it. Same behavior there; it appears _only_ with jruby-complete, not with the "normal" jruby.)

mkristian:
the original problem you are referring to was when you kill the jruby
process hard, then the JVM can not clean up files marked as
```deleteOnExit()``` since the process is gone. this is true only for
jruby-complete or any scenario where you need to have native libraries or
jars loaded from within a jar-file (in your case the jruby-complete.jar)
which are copied to an temp file and used from there. you are saying that
[...]

PR #6269, opened 2020-06-05
Avoid using URL.openStream due to JDK cache

headius:
The caching done by OpenJDK's jar connection logic appears to be
sensitive to heavy concurrent use. When requesting that a URL open
a stream on itself, the default behavior is for it to use an
internal cache of previously opened jar files. A map from the URL
to the JarFile is maintained, where JarFile eventually bottoms out
in an open ZipFile.

Unfortunately URLClassLoader uses the same cache to reuse open jar
files, while also adding those files to its list of closables.
When the URLClassLoader itself is closed, all closables are
closed, which may end up closing a cached JarFile.

So if different URLClassLoader instances are used to access the
same jar files, it is likely that a shared JarFile instance will
be used and eventually closed, triggering errors.

This commit alters our use of URL.openStream to always instead
open a URLConnection and then disable caching before proceeding to
actually open the stream. This avoids the use of cached jar files,
ensuring we don't hit this bug.

A similar change will likely also need to be done for all places
where we do ClassLoader.getResourceAsStream, since if the
classloader is a URLClassLoader it will backend on this same cache
and once again eat the poison pill.

Part of, and possibly sufficient for, issue #6218.

### DIFF
--- a/core/src/main/java/org/jruby/embed/internal/EmbedRubyRuntimeAdapterImpl.java
+++ b/core/src/main/java/org/jruby/embed/internal/EmbedRubyRuntimeAdapterImpl.java
@@ -37,6 +37,7 @@ import java.io.InputStream;
 import java.io.Reader;
 import java.io.StringReader;
 import java.net.URL;
+import java.net.URLConnection;
 
 import org.jruby.Ruby;
 import org.jruby.RubyInstanceConfig.CompileMode;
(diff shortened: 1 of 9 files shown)


headius:
I've merged a less-invasive fix in #6273, which leaves system jar files open but closes any that we created ourselves (for nested jars). This PR would be necessary to fully isolate us from the broken JarFile cache in OpenJDK, but I think #6273 will effectively avoid the bug for all JRuby use cases.

headius:
There may be a renewed reason to use this patch: keeping these jar files open, even if they don't leak, means they will never get deleted on Windows. That has led to "leaking" these temp files as reported in #3657

headius:
After much work on trying to get temp files to delete on Windows, I've determined that it's largely an unsolvable problem.

Files can be deleted on Windows if opened with FILE_SHARE_DELETE flag, but most files opened by the JDK do not use this. In addition, we can't control files opened by a URLClassLoader because it's deep in the bowels of JDK code. Experiments to mark the jars as delete-on-exit have also not helped, since at the time of exit the jars are still held in a global cache.

So given that there's a larger, more complicated challenge on Windows, and our localized fix in #6273 seems to have addressed all reported issues, I'm going to close this. I do hope to see the JDK get a fix for it in the future, and we may explore whether we can turn off caching reliably on Java 9 and higher.

PR #6276, opened 2020-06-10
Fix typo to combine flags in the parser singleton class rule

eregon:
See https://github.com/ruby/ruby/blob/46a5d1b4a63f624f2c5c5b6f710cc1a176c88b02/parse.y#L2585

I'm not sure how to regenerate the parser (and ripper maybe), could you do it?

### DIFF
--- a/core/src/main/java/org/jruby/parser/RubyParser.y
+++ b/core/src/main/java/org/jruby/parser/RubyParser.y
@@ -1558,7 +1558,7 @@ primary         : literal
                     support.setIsInClass($<Boolean>4.booleanValue());
                 }
                 | k_class tLSHFT expr {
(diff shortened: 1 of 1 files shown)


enebo:
@eregon Thanks for submitting this.  I am doing a lot of parser work on a branch and will merge it there (as well as generating new versions of the parsers).  If you ever need to generate it you can run ./tool/generate_parser and ./tool/generate_ripper.  Leaving it open to make sure I address it.

enebo:
I added it to my branch which will make it into 9.3/master soonish.  I also "upgraded" it to use 0b notation to hopefully help with the intent in 33e3fd74769d23b2f46c8dd02d2e4c2902713330 (as well as generate the .java files).

eregon:
Thanks!

PR #6279, opened 2020-06-14
Some fixes, tweaks, and new tools for working with IR

chrisseaton:
Possibly best reviewed by looking at the individual commits.


enebo:
d47b5da76290513704bbc518f192fece132d7c22 is weird but not from your change.  getInstructions should never return null so the first half of that if should not exist (unless there is concurrent access and then we should ask why that is happening).  The other thing which bothers me is that checking for no instructions seems like a weird mechanism for signalling it is already a full build.  We should at least probably check the ic itself instead of what instrs it happens to have.

None of that involves you PR though :)

PR #6288, opened 2020-06-18
Always treat non-full IC as not having protocol

### LINKED ISSUE #6282: NPE in ruby block called concurrently from Java
**Environment Information**
  
* JRuby 9.2.7.0

**Expected Behavior**

No error when calling ruby block concurrently from Java

I have not been able to reproduce this error locally at all but seems to happen sporadically in our testing pipeline. The failing tests are different but a common pattern is passing a ruby block to java which is called concurrently. An example similar to our tests could be:

```ruby
 results = java.util.concurrent.ConcurrentHashMap.new
 calculate_something_in_parallel(input1, input2, -> (id, result) { results[id] = result })
```
The stack trace then indicates that it fails when calling the lambda. 

**Actual Behavior**

Returns an NPE with the following stack trace:

```
java.lang.NullPointerException 
     org.jruby.runtime.InterpretedIRBlockBody.commonYieldPath(InterpretedIRBlockBody.java:127)
     org.jruby.runtime.IRBlockBody.call(IRBlockBody.java:79)
     org.jruby.runtime.Block.call(Block.java:124)
     org.jruby.RubyProc.call(RubyProc.java:295)
     org.jruby.RubyProc.call(RubyProc.java:274)
     org.jruby.RubyProc.call(RubyProc.java:270)
     org.jruby.javasupport.Java$ProcToInterface.callProc(Java.java:1131)
     org.jruby.javasupport.Java$ProcToInterface.access$300(Java.java:1108)
     org.jruby.javasupport.Java$ProcToInterface$ConcreteMethod.call(Java.java:1169)
     org.jruby.gen.InterfaceImpl630006984.consume(org/jruby/gen/InterfaceImpl630006984.gen:13)
```

headius:
The logic here used to go to interpreterContext to check for the
call protocol logic. The hasExplicitCallProtocol method returns a
cached value acquired from the encapsulated scope, via its flags.
However the enum of flags in IRScope may be mutated concurrently,
leading to a situation where we have seen the scope flip to having
call protocols installed, but the InterpreterContext in hand is
still the one without protocol instructions. As a result, we may
start interpreting using an unoptimized IR along the "direct" path
which does not push any scope state. This lead to the AIOOB
reported later in #6282.

The NPE reported in #6282 has not reproduced for me on the
jruby-9.2 branch, and may have been fixed through other means, but
this modified logic should also help any such cases that remain.
Instead of overwriting the two InterpreterContext fields, the non-
full version will always remain the non-full version, and the full
version will only be set into that field. Once set, neither are
ever cleared or re-set to a new value, avoiding a transient null
in the instructionContext field.

I ran the new reproduction from #6282 for a few hours and saw no
errors. Adding a test that guarantees this is fixed may be
difficult, but I believe we've got it here.

This logic will be overhauled for 9.3, so we'll go with this
workaround for now and call it fixed.

Fixes #6282.

### DIFF
--- a/core/src/main/java/org/jruby/runtime/InterpretedIRBlockBody.java
+++ b/core/src/main/java/org/jruby/runtime/InterpretedIRBlockBody.java
@@ -76,7 +76,6 @@ public class InterpretedIRBlockBody extends IRBlockBody implements Compilable<In
 
             ic = closure.getInterpreterContext();
             interpreterContext = ic;
-            fullInterpreterContext = interpreterContext;
         }
         return ic;
     }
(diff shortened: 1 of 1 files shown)


core/src/main/java/org/jruby/runtime/InterpretedIRBlockBody.java
```diff
@@ -93,31 +92,28 @@ public String getName() {
...
+        ensureInstrsReady(); // so we get debugging output
+        return Interpreter.INTERPRET_BLOCK(context, block, self, fullInterpreterContext, args, block.getBinding().getMethod(), Block.NULL_BLOCK);
     }
 
     @Override
     protected IRubyObject commonYieldPath(ThreadContext context, Block block, Block.Type type, IRubyObject[] args, IRubyObject self, Block blockArg) {
         if (callCount >= 0) promoteToFullBuild(context);
 
-        InterpreterContext ic = ensureInstrsReady();
+        ensureInstrsReady();
```
enebo, line 114: My only question is it wrong for 115 to stay as it is while still removing line 120 from before?  I see a field lookup after ensureInstrsReady() in new patch and I am trying to reason through concurrency aspects of that vs returning from ensureInstrsReady.  All I can think of it if somehow two ensureInstrsReady happen at the same time you will just get one. They both at least logically should be the same.

headius: You're right, and I've pushed a change that uses the result of calling `ensureInstrsReady`. Since there's no crossover on the two IC fields, this is now safe (and there's no reason to re-read `interpreterContext`).

PR #6289, opened 2020-06-18
Only define static field constant when inner class collides

### LINKED ISSUE #6196: "warning: already initialized constant Companion" with Kotlin companion objects

headius:
This commit makes two changes to fix #6196:

* When both an inner class and a static final field have the same
  name, only define the field. Typically this is used for a
  singleton pattern where the field points at a singleton instance
  of the inner class, as with Kotlin companion classes. Since we
  can't define both, we prefer the field.
* Do not define the inner class constant as a side effect of
  initializing that class's proxy, like we do for package
  constants for normal classes. Defer that constant definition to
  the initialization of the declaring class, which will make the
  decision whether to define (or not) the constant at that point.

If both the field and the class are needed, it's still possible to
import the inner class under a specific name using `java_import`:

```ruby
java_import("OuterClass$InnerClass") { "MyWeirdInnerClass" }
MyWeirdInnerClass.whatever()
```

### DIFF
--- a/core/src/main/java/org/jruby/javasupport/Java.java
+++ b/core/src/main/java/org/jruby/javasupport/Java.java
@@ -773,14 +773,13 @@ public class Java implements Library {
 
         final RubyModule parentModule; final String className;
 
-        if ( fullName.indexOf('$') != -1 ) { // inner classes must be nested
-            Class<?> declaringClass = clazz.getDeclaringClass();
-            if ( declaringClass == null ) {
-                // no containing class for a $ class; treat it as internal and don't define a constant
-                return;
-            }
-            parentModule = getProxyClass(runtime, JavaClass.get(runtime, declaringClass));
-            className = clazz.getSimpleName();
+        if ( fullName.indexOf('$') != -1 ) {
+            /*
+             We don't want to define an inner class constant here, because it may conflict with static fields.
+             Instead, we defer that constant definition to the declaring class's proxy initialization, which will deal
+             with naming conflicts appropriately. See GH-6196.
+             */
+            return;
         }
         else {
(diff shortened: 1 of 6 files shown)


headius:
I will add a trivial mechanical test case using Java code, and I believe @kalenp will try to come up with some actual Kotlin test cases we can incorporate.

headius:
Note that this PR really *just* avoids the warning, and does nothing special with the companion classes. We may want to consider a bit more magic, similar to Scala singleton objects where make the object's methods appear to be on the class like they would appear from Scala code.

Issue #6283, opened 2020-06-15
Travis CI: testing on arm64


mohamedhafez:
Travis CI now supports testing on `arm64` as an alpha feature (see https://docs.travis-ci.com/user/multi-cpu-architectures/), it would be great if JRuby could be tested on that architecture in addition to the default `amd64`. 

The reason I ask is we would like to start running our JRuby app on AWS's new Graviton2 ARM instances due to the big cost savings, and it would be great to know that JRuby would work without issues on ARM

monkstone:
That's interesting, I am already using JRuby for my PiCrate gem, which so far seems to run without issues https://discourse.processing.org/t/experiments-with-64-bit-os-on-raspberrypi4/20628.

headius:
Good idea! I don't expect will have many troubles, since the heavy lifting has mostly been done by the JVM. All our native bits already support arm64 as well.

headius:
This is a non release item, but I marked it for 9.3 since it would be nice to say we pass tests on aarch64.

headius:
I have attempted to do this but the support is rather "raw" at this point. See my PR at #6285 

I managed to basically get it to run, but I could only install OpenJDK 11 which had some failures. That may be resolved now with recent work on JRuby master to get newer JDKs in CI, but the environment is still a bit tricky to work with.

For what it's worth, other than the Java 11 issues, tests appear to run just fine on AArch64. We will be expanding our testing over the coming weeks.

headius:
Huzzah! I rebased on master and managed to get a passing run of the smaller jruby "smoke test" suite on ARM64 + JDK11.

I'm now poking at expanding to the other more comprehensive suites, but this is a good sign.

headius:
Ok, I think this seals the deal: I managed to get the Ruby specs running on Travis ARM64, and we fail just 0.13% of the specs we pass on x86_64. Many of the remaining failures look like they are environment issues (missing syslog, for example) or possibly bad specs.

We will continue to expand testing on ARM64, but I think it's safe to say that JRuby on ARM64 is working very well already!

headius:
We have left TravisCI because they ended our free OSS support. We are fully on GHA for the 9.3 and master branches, and will be adding some jobs for the 9.2 branch.

We would like to add some other platforms, like aarch64, but we could use some help!

PR #6292, opened 2020-06-20
Add more travis platforms

headius:
Travis now supports (in an experimental way) three additional hardware platforms: `arm64`, `ppc64le`, and `s390x`. I will use this PR to try out some of our suites on those platforms and keep the ones that work.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -305,3 +305,4 @@ addons:
   apt:
     packages:
       - haveged
+      - ant
--- a/.travis.yml
+++ b/.travis.yml
@@ -13,7 +13,7 @@ cache:
 
 before_install:
   - unset _JAVA_OPTIONS
-  - rm ~/.m2/settings.xml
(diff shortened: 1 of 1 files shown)


headius:
See #6283 for users asking about AArch64 support.

headius:
One of the test:jruby:jit ARM64 builds failed with an unusual EAGAIN message... I've restarted it, we'll see if it happens again.

This Travis environment is definitely incomplete... more recently I had to tell it to install `ant`, even though that's supposed to be there already. Previously I had to modify the builds to use our `mvnw` script, since Maven is also not available.

headius:
This was an interesting experiment, but I feel like we are migrating away from Travis more and more. Github Actions does not yet have support for other hardware platforms, but ARM64 is all but inevitable, and we may be able to utilize https://github.com/uraimo/run-on-arch-action in the short term or for ppc64 and s390x.

PR #6294, opened 2020-06-23
[deps] update Rubygems to latest 3.1 (3.1.6)

kares:
targeting for 9.2.12

failing 2 ruby specs atm - needs investigation as RGs changed path handling:
```
1)
Kernel#require (path resolution) raises a TypeError if passed an object that has #to_s but not #to_str FAILED
Expected TypeError but no exception was raised (true was returned)
/home/travis/build/kares/jruby/spec/ruby/core/kernel/shared/require.rb:80:in `block in <main>'
org/jruby/RubyDir.java:439:in `chdir'
/home/travis/build/kares/jruby/spec/ruby/core/kernel/shared/require.rb:79:in `block in <main>'
org/jruby/RubyBasicObject.java:2622:in `instance_eval'
org/jruby/RubyEnumerable.java:1809:in `all?'
org/jruby/RubyArray.java:1809:in `each'
org/jruby/RubyArray.java:1809:in `each'
/home/travis/build/kares/jruby/spec/ruby/core/kernel/require_spec.rb:5:in `<main>'
org/jruby/RubyKernel.java:1009:in `load'
org/jruby/RubyBasicObject.java:2622:in `instance_eval'
org/jruby/RubyArray.java:1809:in `each'

2)
Kernel#require ($LOADED_FEATURES) does not load a ./ relative path that is already stored FAILED
Expected true to be false
/home/travis/build/kares/jruby/spec/ruby/core/kernel/shared/require.rb:397:in `block in <main>'
org/jruby/RubyDir.java:439:in `chdir'
/home/travis/build/kares/jruby/spec/ruby/core/kernel/shared/require.rb:396:in `block in <main>'
org/jruby/RubyBasicObject.java:2622:in `instance_eval'
org/jruby/RubyEnumerable.java:1809:in `all?'
org/jruby/RubyArray.java:1809:in `each'
org/jruby/RubyArray.java:1809:in `each'
/home/travis/build/kares/jruby/spec/ruby/core/kernel/require_spec.rb:5:in `<main>'
org/jruby/RubyKernel.java:1009:in `load'
org/jruby/RubyBasicObject.java:2622:in `instance_eval'
org/jruby/RubyArray.java:1809:in `each'
```


deivid-rodriguez:
I think this might be because of some bugs that I introduced in rubygems 3.1 require :disappointed:.

Cherry-picking recent PRs fixing require issues should probably do the trick:

* https://github.com/rubygems/rubygems/pull/3124
* https://github.com/rubygems/rubygems/pull/3133
* https://github.com/rubygems/rubygems/pull/3153
* https://github.com/rubygems/rubygems/pull/3639

kares:
Thanks David, will those PRs get a backport for 3.1 or will the only land in 3.2?
Need to check Ruby 2.7 (the only one on RGs 3.x) whether they applied any patching over the default install.

deivid-rodriguez:
I think there's an intention to backport these fixes to rubygems 3.1.5 and include them with ruby 2.7.2. At least @hsbt assigned [this ruby-core ticket to himself](https://bugs.ruby-lang.org/issues/16926) after I suggested it.

headius:
@deivid-rodriguez For context, we want to release JRuby 9.2.12 with RubyGems 3.1, but we need these fixes present. We are naturally reluctant to patch our shipped version of RubyGems if it's possible an upgrade will wipe them out.

deivid-rodriguez:
Thanks! @kares already explained that in https://github.com/rubygems/rubygems/pull/3751, and it sounds totally reasonable to me.

headius:
@kares We're reaching endgame for 9.2.12 so I think we need to make a decision. Updating right before release makes me and @enebo a bit nervous.

Can you explain why you believe we need to do this update as part of the release, rather than just recommending that users `gem update --system` themselves?

deivid-rodriguez:
FWIW, 3.1.6 was released with these backports: https://blog.rubygems.org/2021/01/26/3.1.6-released.html.

headius:
@deivid-rodriguez Thank you!

@enebo Updating RubyGems is probably way out of scope for 9.2.17.0 but perhaps .18? cc @kares

lib/ruby/stdlib/rubygems/command_manager.rb
```diff
@@ -169,12 +169,6 @@ def process_args(args, build_args=nil)
     when '-v', '--version' then
       say Gem::VERSION
       terminate_interaction 0
-    when '--no-ri', '--no-rdoc' then
```
headius, line 172: Minor concern here as these flags will now error.

headius (approved):

lib/ruby/stdlib/rubygems/command_manager.rb
headius: Actually nevermind... they errored before. The difference is that they use the deprecation mechanism instead of having special code.

PR #6295, opened 2020-06-23
Add Kotlin Integration tests for Companion objects

kalenp:
This starts a Kotlin integration spec.  The integration of Kotlin into
the build still needs a bit of work, but if you have `kotlinc` installed
in your path, you can compile the fixtures and run the tests with

```
jruby -S rake test:compile
jruby bin/rspec spec/java_integration/kotlin
```

### DIFF
--- a/rakelib/test.rake
+++ b/rakelib/test.rake
@@ -30,6 +30,9 @@ namespace :test do
     java_home = [ ENV_JAVA['java.home'], File.join(ENV_JAVA['java.home'], '..') ] # in case of jdk/jre
     javac = java_home.map { |home| File.expand_path('bin/javac', home) }.find { |javac| File.exist?(javac) } || 'javac'
     sh "#{javac} -cp #{classpath} -d test/target/test-classes #{Dir['spec/java_integration/fixtures/**/*.java'].to_a.join(' ')}"
+    # TODO: What about environments without this installed?
+    kotlinc = 'kotlinc'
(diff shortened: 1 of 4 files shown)


rakelib/test.rake
```diff
@@ -30,6 +30,9 @@ namespace :test do
     java_home = [ ENV_JAVA['java.home'], File.join(ENV_JAVA['java.home'], '..') ] # in case of jdk/jre
     javac = java_home.map { |home| File.expand_path('bin/javac', home) }.find { |javac| File.exist?(javac) } || 'javac'
     sh "#{javac} -cp #{classpath} -d test/target/test-classes #{Dir['spec/java_integration/fixtures/**/*.java'].to_a.join(' ')}"
+    # TODO: What about environments without this installed?
+    kotlinc = 'kotlinc'
+    sh "#{kotlinc} -cp #{classpath} -include-runtime -d test/target/test-classes/kotlin_fixtures.jar #{Dir['spec/java_integration/fixtures/**/*.kt'].to_a.join(' ')}"
```
kalenp, line 35: I built this as a jar using the `-include-runtime` parameter to ensure that the necessary kotlin libraries would be available when running the tests.  Otherwise, you'll need to arrange to have the kotlin runtime included in the classpath when running the kotlin tests, since the compiled fixtures end up referencing kotlin types.

headius:
We will want to make this work using Maven to do the Kotlin compile but it's a bit involved. I will get to it eventually, or @mkristian might be able to help in the short term.

Issue #6287, opened 2020-06-18
com.sun.javafx.collections not exported in javafx.base


GoldenWeiss:
.: Environment :.
jruby 9.2.11.1

.: Overview :.
It prevents access to the inherited methods of _ObservableMapWrapper_ in the jrubyfx DSL. Hence, the method add_change_listener throws an _Exception_ : 

TypeError: illegal access on 'addListener': class org.jruby.javasupport.JavaMethod cannot access class com.sun.javafx.collections.ObservableMapWrapper (in module javafx.base) because module javafx.base does not export com.sun.javafx.collections to unnamed module @abc0923
  add_change_listener at uri:classloader:/jrubyfx/core_ext/observable_value.rb:121

.: Reproduction :.
x =Java::javafx.collections.FXCollections.observable_map({})
x.addListener Java::javafx.collections.MapChangeListener.impl{|a,b| puts "z"}

.: Small patch by Headius :.
In run configurations, add the line : 
--add-exports javafx.base/com.sun.javafx.collections=ALL-UNNAMED

.: Note :.
observable_array_list.add_change_listener runs just fine, as it doesn't inherit from collections.

headius:
The problem here is that `ObservableMapWrapper` is a public class, but in a non-exported package. Because it's public we proceed to set up its methods, assuming they and the class itself will be accessible. However when a class is not exported, accessing it directly this way triggers the error reported above.

We will also need to check whether a given class has been exported to the module JRuby is running as before binding its methods. If it has not been exported, we may try to export it, but failing that we should treat it the same way we treat private classes.

headius:
Merged to master and 9.2.

PR #6296, opened 2020-06-24
Fixes to improve Java 11 testing

headius:
This PR will fix issues preventing some of our test suites from passing on Java 11, in order to help move toward full compatibility in the presence of Java modules.

This PR is based on 9.3 but most fixes should be mergeable to 9.2, and we may want to consider some or all of them for 9.2.12.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -51,6 +51,10 @@ jobs:
         run: tool/maven-ci-script.sh
         env:
           PHASE: '-Pmain,test -Dinvoker.test=extended'
+      - name: test profile
+        run: tool/maven-ci-script.sh
+        env:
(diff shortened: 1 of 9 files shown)


headius:
How boring. I only got to make one fix!

With this change, the following test suites are green and running on JDK11 in CI:

* JRuby core (test:jruby)
* JRuby IR and bytecode compilation w/ invokedynamic (spec:compiler)
* Java integration (spec:ji)
* ruby/spec core (spec:ruby:fast)
* CRuby core (test:mri:core:jit)

The only fix here was to avoid binding non-exported classes' methods, which fixes #6287. It's simple, so I'll cherry-pick it for 9.2.12.

headius:
I tossed FFI in as well since that rounds out the set of features being tested.

PR #6298, opened 2020-06-25
Do not bind methods from classes that are not exported

headius:
This cherry-picks the fix for #6287 for 9.2.12.


headius:
Something strange is up with the Travis failures unrelated to this PR. It seems to be picking up a filename from master even though the only commit makes no such change. Will merge and see if it happens on the branch.

PR #6301, opened 2020-06-26
Reconfigure some CI builds and add JDK14

headius:
This PR makes a few changes to our Travis CI runs:

* Modify the "int" builds to encourage a full build for code called twice, using threshold=1 and jit.background=false.
* Remove "fullint" builds from JRuby and MRI suites.
* Add indy flag to a few JDK11 builds.
* Duplicate some of the JDK11 builds for JDK14. These builds should be kept up with latest OpenJDK release (so move to 15 this fall).

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -50,12 +50,6 @@ matrix:
         - jruby -S rake test:mri:core:int
       env: JRUBY_OPTS='$JRUBY_OPTS -J-Xmx1G'
 
-    - name: MRI core fullint
-      script:
-        - mvn clean package -Pbootstrap
-        - jruby -S rake test:mri:core:fullint
-      env: JRUBY_OPTS='$JRUBY_OPTS -J-Xmx1G'
-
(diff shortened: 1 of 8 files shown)


headius:
Two things failed on JDK14:

* -Xverify:none and related flags are now deprecated and print a warning at boot, leading to failures in several subprocess tests that received unexpected output. I have patched around this by adding `-XX:-PrintWarnings` to our `--dev` flags, but we should start trying to migrate away from using these flags.
* OpenJDK 13 and higher will now raise BindException (instead of IOException) for EACCES when binding a socket. This caused a failure in `test_udp_socket_bind` in `test/jruby/test_socket.rb`. I've fixed this by checking the BindException message, as we do for other IOExceptions.

headius:
The two minor patches here (handle BindException based on message and properly assign errno for a couple others) have been cherry-picked to 9.2.12, to improve compatibility with Java 13+.

PR #6303, opened 2020-06-28
Run security specs

eregon:
* Inherit from the default MSpec configuration
* Run security specs

[`security/` specs](https://github.com/ruby/spec/commits/master/security) exist since a while in ruby/spec but it seems JRuby didn't run them in CI yet.

### DIFF
--- a/spec/jruby.mspec
+++ b/spec/jruby.mspec
@@ -4,6 +4,9 @@ require 'rbconfig'
 require 'java'
 require 'jruby'
 
+# Inherit from the default configuration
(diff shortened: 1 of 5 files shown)


eregon:
Some security specs are failing, see https://travis-ci.org/github/jruby/jruby/jobs/702869188

headius:
The failing security specs seem mostly to be randomized hashes. I'll review tomorrow and we'll patch them up for 9.2.12. Will file a separate issue.

Issue #6304, opened 2020-06-29
Align behavior with CRuby for security specs


headius:
In #6303 @eregon noticed a few of the security specs do not pass currently.

Tags are here: https://github.com/jruby/jruby/pull/6303/commits/48c5e8552f404be3153391cacb7de54e2476c587

An example run is here: https://travis-ci.org/github/jruby/jruby/jobs/702869188

The failures are mostly due to us not randomizing hash calculation, but there's a few others that warrant a second look. I believe it would be best for us to patch these for the 9.2.12 release this week.

headius:
The hash failures are all due to us not using siphash plus a random seed in the numeric classes. As with other places this was patched, we have not generally agreed with the "fix" of randomizing hashes, and instead allow users to opt into siphash if they have concerns. That decision could be revisited, but at least we should update the numeric classes to also have optional siphash calculation. We may also be able to use our existing "perlhash" with a random seed and meet the spirit of these fixes.

The WEBrick failures appear to have never been made to the ruby/webrick repository, or at least Ruby 2.6 claims to ship WEBrick 1.4.2, but there are extensive diffs from the ruby_2_6 branch and the WEBrick v1.4.2 tag. It appears CRuby has diverged here without properly updating release versions.

The Array#pack failure shows a place where we are not propagating tainting. We have never fully supported tainting, as it is a broken security mechanism. Nevertheless, we should be able to patch this easily to pass the spec. I assume this spec will no longer be relevant when tainting is removed as a feature.

The String#unpack failure simply raises the wrong error. There's no vulnerability here (both because of the error and because the JVM will prevent such memory issues), but we could add the logic to raise the RangeError the spec expects.

headius:
The numeric hash failures are addressed in #6305.

The String#unpack error is addressed in #6306.

headius:
The Array#pack failure is due to us not tainting the result when any of the elements of the array are themselves tainted.

The modifications to fix this would be somewhat invasive.

Given the fact that:

* tainting is deprecated in 2.7 and to be removed in 2.8/3.0
* tainting has never been fully supported in JRuby
* tainting is in fact a broken security mechanism

...I vote that we don't bother with this change.

headius:
The remaining WEBrick failures will be addressed once we figure out what version of WEBrick is actually being shipped in Ruby 2.5.x and 2.6.x. CRuby appears to be shipping heavily patched sources for both releases, while claiming they're both WEBrick 1.4.2.

[...]

PR #6305, opened 2020-06-29
Mimic CRuby's numeric hashing

headius:
This PR makes changes to RubyFixnum, RubyBignum, and RubyFloat hashing to incorporate the per-runtime random seed, in an effort to mimic CRuby's fixes for CVE-2011-4815.

This CVE was part of a flurry of "hashDOS" fixes that happened in the early 2010s to address poor hash use by application frameworks like Rails. Because untrusted data from web requests was blindly used for internal hashing, these frameworks could be DOS'ed by predicting the hashes of strings and other data and tricking Ruby into putting all elements in the same linear-scanned hash box.

The change here roughly mimics the CRuby logic for combining a random seed with the 64-bit long representation of these core numeric types, which in turn propagates that randomness to the composite types like Complex and Rational.

We may want to make this randomization optional, since we have had requests like #590 to allow users to opt into repeatable hashing.

This PR can safely merge to master.

See #6304 for the security specs failing without this change. This PR addresses all the failing numeric hash security specs.

### DIFF
--- a/core/src/main/java/org/jruby/RubyBignum.java
+++ b/core/src/main/java/org/jruby/RubyBignum.java
@@ -41,6 +41,7 @@ import org.jruby.anno.JRubyClass;
 import org.jruby.anno.JRubyMethod;
 import org.jruby.common.IRubyWarnings.ID;
 import org.jruby.runtime.ClassIndex;
+import org.jruby.runtime.Helpers;
 import org.jruby.runtime.JavaSites;
 import org.jruby.runtime.ThreadContext;
 import org.jruby.runtime.builtin.IRubyObject;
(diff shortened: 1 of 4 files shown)


headius:
I will modify this to put the random seed logic behind the same "consistent hash" option, so it can be disabled.

headius:
Aha, the random seed property `jruby.consistent.hashing` actually modifies how we acquire the seed; when on, it uses a static value. This is ready to go.

```
[] ~/projects/jruby $ jruby -e "1000.upto(1010) {|i| puts i.hash}"
5850910177678188037
-4122927456021001207
4349978983989361182
-5623858649709827614
2849047790300534774
-7124789843398654470
1348116596611707855
-8625721037087481389
-152814597077119065
8320091842933243307
-1653745790765945921

[] ~/projects/jruby $ jruby -e "1000.upto(1010) {|i| puts i.hash}"
-7227042336866544064
-2770157313365699149
1686727710135145767
6143612733635990664
-7846246316572715779
-3389361293071870866
1067523730428974035
5524408753929818948
-8465450296278887767
-4008565272778042854
448319750722802060

[] ~/projects/jruby $ jruby -Xconsistent.hashing -e "1000.upto(1010) {|i| puts i.hash}"
-2980970509909832607
-1858700091923459759
-736429673937086975
385840744049285873
1508111162035658657
2630381580022031505
3752651998008404289
4874922415994777137
5997192833981149921
7119463251967522769
8241733669953895553

[] ~/projects/jruby $ jruby -Xconsistent.hashing -e "1000.upto(1010) {|i| puts i.hash}"
-2980970509909832607
-1858700091923459759
-736429673937086975
385840744049285873
1508111162035658657
2630381580022031505
3752651998008404289
4874922415994777137
5997192833981149921
7119463251967522769
8241733669953895553
```

PR #6307, opened 2020-06-30
Only close temporary jars to avoid poisoning global jar file cache

headius:
Cherry-pick of #6273.


headius:
After discussion on Matrix, we will pull this change into 9.2.12, but default the "close URLClassLoader" option to behave as in JRuby 9.2.11.

Issue #4194, opened 2016-09-30
"LoadError: no such file to load" if IOException is raised in loaded file


olleolleolle:
### Environment

`jruby 9.1.5.0 (2.3.1) 2016-09-07 036ce39 Java HotSpot(TM) 64-Bit Server VM 25.92-b14 on 1.8.0_92-b14 +jit [darwin-x86_64]`

`Darwin xxxxx.local 15.6.0 Darwin Kernel Version 15.6.0: Mon Aug 29 20:21:34 PDT 2016; root:xnu-3248.60.11~1/RELEASE_X86_64 x86_64`

But we originally had the issue on Linux too (same JRuby version).
### Expected Behavior

To see `Java::JavaIo::IOException` and the stacktrace
### Actual Behavior

`LoadError: no such file to load -- test.rb`
### Reproduction

https://gist.github.com/olleolleolle/c56a31f32401d944c87d6c37a82c0f14

jensnockert:
https://github.com/jruby/jruby/blob/master/core/src/main/java/org/jruby/runtime/load/LibrarySearcher.java#L246 is the issue, trying to find a fix.

headius:
I noticed the other day that MRI does not appear to wrap raised exceptions in LoadError anymore, while we still do. I am not certain when it DOES through LoadError, but I suspect it's only for actual load/require problems.

dekellum:
This just bit me.   Before finding this issue I managed to produce a minimized reproduction, in case it helps, shown below.  @headius : some further notes on this.

* As best I can tell, this only happens with IOException and derivatives, not other java exception classes or ruby exceptions.
* You suggest the LoadError is wrapping the IOException, but wouldn't jruby's normal exception dumping show the wrapped exception in this case? On jruby 9.2.7 it doesn't.   
* On jruby 1.7.26 (bottom) it does at least show the message (but not the stack) of the IOException which I guess is why this never bothered me before on 1.7.x
* This load scenario would effect almost any CLI utility written in jruby that uses java.io and is installed as gem and executed via the normal gem wrappers. 
* Workarounds would include wrapping these exceptions in something else, or to catch and handle the exceptions via the loaded ruby file.

~~~ txt
% jruby -v
jruby 9.1.7.0 (2.3.1) 2017-01-11 68056ae OpenJDK 64-Bit Server VM 25.121-b13 on 1.8.0_121-b13 +jit [linux-x86_64]

% cat ./raises.rb
require 'java'

class CLI
  def run
[...]

PR #6313, opened 2020-07-09
Do the read of the full file contents separately

headius:
This allows us to propagate IOException rather than wrongly
interpreting it as an error caused by the load of the file itself.
It still gets wrapped in a LoadError, but at least it does not
provide a misleading message.

Relates to #4194.


olleolleolle:
This will communicate clearer to the user, and will solve my reported problem. Thanks for your tenacity!

Issue #1585, opened 2014-03-26
jit.codeCache doesn't work


areman:
Hello,

I'm using **jruby 1.7.11 (1.9.3p392) 2014-02-24 86339bb on Java HotSpot(TM) Client VM 1.7.0_10-b18 [Windows XP-x86]**

I run 
**jruby -Xcompile.mode=FORCE -Xjit.debug=true -Xjit.codeCache=c:\tmp test.rb**
and many 

```
... Ruby: no jitted code in classloader for ...
... JITCompiler: writing jitted code to to ...
```

will be printed. And in the rerun the same, but the cached files exists in the directory.
I looked into the source and I found that the **JRubyClassLoader** is used to load the cached classes. But I can't find the location in the code where the **codeCache** directory is added to the **JRubyClassLoader**.
After I patched the getJRubyClassLoader method in the Ruby class, and add the code cache directory to the Classloader, everything works fine for me.

``` ruby
public synchronized JRubyClassLoader getJRubyClassLoader() {
    // FIXME: Get rid of laziness and handle restricted access elsewhere
    if (!Ruby.isSecurityRestricted() && jrubyClassLoader == null) {
        jrubyClassLoader = new JRubyClassLoader(config.getLoader());

        // if jit code cache is used, we need to add the cache directory to the classpath
        // so the previously generated class files can be reused.
        if( config.JIT_CODE_CACHE != null && !config.JIT_CODE_CACHE.trim().isEmpty() ) {
            File file = new File( config.JIT_CODE_CACHE );

            if( file.exists() == false || file.isDirectory() == false ) {
                getWarnings().warning("The jit.codeCache '" + config.JIT_CODE_CACHE + "' directory doesn't exit.");
            } else {
                try {
                    URL url = file.toURI().toURL();
                    jrubyClassLoader.addURL( url );
                } catch (MalformedURLException e) {
                    getWarnings().warning("Unable to add the jit.codeCache '" + config.JIT_CODE_CACHE + "' directory to the classpath." + e.getMessage());
                }
            }
        }
    }

    return jrubyClassLoader;
}   
```

Maybe this is an possible patch for the Ruby class.

Thanks,
André

MSNexploder:
I guess this one can be closed.
"jit.codeCache" is deprecated in 9k and (as far as I can see) does nothing nowadays.

kares:
@headius @enebo for 9.2 we shall move options that do nothing from being printed on `--properties` ?

headius:
[...]

PR #6315, opened 2020-07-10
Deprecate and isolate all Options that are no longer used

headius:
Most of these were completely unused. A few were still used but
the features they referred to did not work anymore, or were
otherwise nonnfunctional in some way.

There's some related cleanup of RubyInstanceConfig here but I have
not done a full audit of unused fields and methods there.

At least a few of these may have meaning in the future, if the
functionality they used to refer to is restored:

* REFLECTED_HANDLES: we may want to add back reflected
  DynamicMethod handles to core class methods to support platforms
  that do not allow bytecode generation (e.g. GraalVM AOT).
* FIBER_COROUTINES: The Loom project is moving fast and we will
  definitely take advantage of it. We will probably want a more
  Loom-specific property name at that time, though.
* FFI_COMPILE_INVOKEDYNAMIC: We are not currently binding Ruby to
  FFI calls as directly as we could, and we should probably
  restore this functionality at some point.
* THREAD_DUMP_SIGNAL: We used to register a trap for dumping all
  Ruby stack traces. We may want to restore this functionality.
* ENUMERATOR_LIGHTWEIGHT: We may want to make it possible to turn
  off lightweight enumerators, of which we currently have none?
  Seems unlikely.
* INVOKEDYNAMIC*: most of the capabilities of invokedynamic have
  been always-on (in indy mode) since they stabilized in Java 8,
  but being able to configure them at a finer grain may be useful
  in the future.
* COMPILE*: as with INVOKEDYNAMIC*, most optimizations are always-
  on, and we may want to restore the ability to turn them off.

This cleanup was inspired by #1585.


headius:
The sole failure in the last build was a regression on master.

PR #6318, opened 2020-07-11
remove redundant Process.uid patch on windows

kares:
have cleaned up unused parts in RG's *defaults/jruby.rb*, this is the only questionable piece.

WindowsPOSIX impl already `return 0` from `Process.uid` the patch should not be needed, 
the only place RGs uses `Process.uid` is for the cache check: https://github.com/rubygems/rubygems/blob/v3.0.6/lib/rubygems/source.rb#L120 which should not matter.

was thinking maybe this is there simply as a desired side-effect? https://github.com/jruby/jruby/commit/8a898fb50edc3af78c5832c484b3af5774384040


headius (approved):
Seems fine to me.

PR #6321, opened 2020-07-13
Add self to SizeFn

headius:
This is a cleaned-up merge of @anukin's changes in #4714.

The addition of `context` happened separately as part of #5958, and as a result this PR only adds the "self" receiver to the SizeFn interface.

This PR may be fine to include as-is, but none of the changes involved actually use the self object for anything. This PR as submitted only updates #4714 to merge properly.

Additional work to follow, which would then actually complete work described in #4688, would be to make all SizeFn objects static instances, using the passed in "self" to do the stateful calls they make now against the captured self. This would avoid the repeated allocation of the SizeFn instances for each enumerator produced by a given enumerable object.


headius:
@enebo @kares the most recent commit is the most invasive, mostly because it adds a generic `<T extends IRubyObject>` parameter to RubyEnumerator.SizeFn.

kares (approved):
:+1: believe this also provides better readability.
despite the bigger change-set it reads nicely ...

headius:
There's some quirks I have to work through here around Enumerator, since it seems the size function there needs to always operate against the enumerator rather than the object it wraps. Will return to this soon, but if someone else wants to try it you can see quick failures running `test/jruby/test_enumerator.rb` or `rake test:jruby`.

Issue #4845, opened 2017-11-09
You can crash irb by defining method `!`


aschmied:
Defining a method `!` causes a crash on `!true`.

### Environment

Windows 10

    > jruby --version
    jruby 9.1.13.0 (2.3.3) 2017-09-06 8e1c115 Java HotSpot(TM) 64-Bit Server VM 25.144-b01 on 
    1.8.0_144-b01 +jit [mswin32-x86_64]

    > java -version
    java version "1.8.0_144"
    Java(TM) SE Runtime Environment (build 1.8.0_144-b01)
    Java HotSpot(TM) 64-Bit Server VM (build 25.144-b01, mixed mode)

### Actual Behaviour

```
> jirb
irb(main):001:0> def !
irb(main):002:1> end
=> :!
irb(main):003:0> !true
NoMethodError: private method `!' called for false:FalseClass
                               gets at C:/jruby/jruby-9.1.13.0/lib/ruby/stdlib/irb/input-method.rb:152
                block in eval_input at C:/jruby/jruby-9.1.13.0/lib/ruby/stdlib/irb.rb:469
                      signal_status at C:/jruby/jruby-9.1.13.0/lib/ruby/stdlib/irb.rb:623
                block in eval_input at C:/jruby/jruby-9.1.13.0/lib/ruby/stdlib/irb.rb:468
                          buf_input at C:/jruby/jruby-9.1.13.0/lib/ruby/stdlib/irb/ruby-lex.rb:189
                               getc at C:/jruby/jruby-9.1.13.0/lib/ruby/stdlib/irb/ruby-lex.rb:104
                           match_io at C:/jruby/jruby-9.1.13.0/lib/ruby/stdlib/irb/slex.rb:206
                              match at C:/jruby/jruby-9.1.13.0/lib/ruby/stdlib/irb/slex.rb:76
                              token at C:/jruby/jruby-9.1.13.0/lib/ruby/stdlib/irb/ruby-lex.rb:283
                                lex at C:/jruby/jruby-9.1.13.0/lib/ruby/stdlib/irb/ruby-lex.rb:265
  block in each_top_level_statement at C:/jruby/jruby-9.1.13.0/lib/ruby/stdlib/irb/ruby-lex.rb:236
                               loop at org/jruby/RubyKernel.java:1292
  block in each_top_level_statement at C:/jruby/jruby-9.1.13.0/lib/ruby/stdlib/irb/ruby-lex.rb:232
[...]

PR #6333, opened 2020-07-18
Use topSelf's class as scope's module

headius:
This appears to allow TOPLEVEL_BINDING to properly define methods on `main` rather than on `Object` as described in recent comments at #4845.


headius:
This clearly doesn't pass, and further investigation seems to show that CRuby actually does use `rb_cObject` as the toplevel cref class:

https://github.com/ruby/ruby/blob/86c338d6e5a9ce6b2c3c5ad0769020bc41952f73/vm.c#L3169

I'll close this and we're back to the drawing board.

Issue #3031, opened 2015-06-10
Match data missing in concurrent proc


donv:
Hi!

Here is a simplified example from an actual production app.  It fails every time for me on JRuby 1.7.20.  Am I doing something wrong?

```ruby
chars = ("a".."z").to_a

block = proc do |i|
  s = 5.times.map {chars.sample}.join
  if /(a)/ =~ s
    raise "Match is missing (#{i}): #{s}: #{$1.inspect}" unless $1
  end
end

threads = 20.times.map{Thread.start{400.times{|i| block.call i}}}
threads.each(&:join)
```

donv:
From IRC #jruby:

```
14:41 <donV> Hi all! Can anybody reproduce? https://github.com/jruby/jruby/issues/3031 
14:46 <bbrowning> donV: RuntimeError: Match is missing (145): gnrak: "a" 
14:46 <ebarrett> RuntimeError: Match is missing (155): sakaw: nil 
14:47 <ebarrett> ^ on a couple of days ago master branch
14:47 <donV> Thanks! 
...
18:46 <enebo> donV: yeah it seems like somehow $~ is getting set and that dynamic scope where it lives is comopeting with another thread 
18:46 <enebo> donV: Should not be doing that so we need to fix it 
```

billdueber:
Making an explicit match object works fine. (Just trying to be helpful while being ignorant)

```ruby

chars = ("a".."z").to_a

block = ->(i) do
  s = 5.times.map {chars.sample}.join
  m = /(a)/.match(s)
  if m
    raise "Match is missing (#{i}): #{s}: #{$1.inspect}" unless m[1]
  end
end

threads = 20.times.map{Thread.start{400.times{|i| block.call i}}}
threads.each(&:join)

```

headius:
There does appear to be a behavioral difference from MRI here:

```
[] ~/projects/jruby $ rvm ruby-2.2 do ruby -e 'p = proc { p $~; "foo" =~ /foo/ }; Thread.new {p.call}.join; Thread.new{p.call}.join'
nil
nil

[] ~/projects/jruby $ jruby -e 'p = proc { p $~; "foo" =~ /foo/ }; Thread.new {p.call}.join; Thread.new{p.call}.join'
nil
#<MatchData "foo">
```

So the $~ are both frame-local and thread-local; we may need some logic in our proc/binding to duplicate the frame if it did not original in the same thread.

headius:
[...]

PR #6336, opened 2020-07-19
Force backref/lastline writes to captured frame into thread-local

headius:
Traditionally the $~ and $_ variables have been frame-local but not thread-local, meaning that if the frame is uprooted by capturing a block, that frame-local storage is shared across threads. That leads to issues like #3031, which has been present in JRuby and CRuby both for at least the past 15 years.

This patch attempts to make writes to backref or lastline on a captured frame become thread-local without losing the direct access for simple single-thread accesses.

We already track whether a frame has been captured in order to eagerly raise LocalJumpError for any non-local breaks or returns that may originate from that block. This also turns out to be useful for knowing that the block and its associated frame are now potentially visible across threads. We reuse this flag to make backref and lastline safe to read and write across threads.

Once a block has been captured, the frame is marked captured, and subsequent writes will switch to using a ThreadLocal. Reads will continue to return the direct value as long as it is not a ThreadLocal, and otherwise will return the thread-local value. The ThreadLocal itself will have an initial value for each thread equal to the value at the time when the first captured write happens.

I believe this allows all current expectations about behavior to work properly, with minimal added overhead:

* For uncaptured reads, the nonvolatile read is now volatile and there's a type-check and cast to IRubyObject.
* For captured reads, the nonvolatile read is now volatile, the IRubyObject type-check will fail, a ThreadLocal type-check will succeed, and a thread-local get will be performed.
* For uncaptured writes, the captured flag will be found to be false before performing a direct write. Since captured can only be set by the thread that owned the frame originally, until captured a thread can freely write directly to the field.
* For captured writes, the 'captured' flag will be found to be true before proceeding to set the existing thread-local or atomically switch to using a new thread-local.

The primary performance issue will most likely be due to the creation of a ThreadLocal at the moment of the first captured write. For code sharing procs across threads, this overhead is unavoidable (and the existing behavior is just a bug that has not been found yet). However for code using procs alongside backref or lastline but only on a single thread, this may be a substantial performance hit with no behavioral improvement.

### DIFF
--- a/core/src/main/java/org/jruby/runtime/Frame.java
+++ b/core/src/main/java/org/jruby/runtime/Frame.java
@@ -36,6 +36,8 @@ package org.jruby.runtime;
 import org.jruby.RubyModule;
 import org.jruby.runtime.builtin.IRubyObject;
 
+import java.util.concurrent.atomic.AtomicReferenceFieldUpdater;
+
 /**
  * A Frame holds per-call information that needs to persist outside the
  * execution of a given method. Currently a frame holds the following:
@@ -83,10 +85,10 @@ public final class Frame {
     private Visibility visibility = Visibility.PUBLIC;
 
     /** backref **/
-    private IRubyObject backRef;
+    private volatile Object backRef;
 
     /** lastline **/
-    private IRubyObject lastLine;
+    private volatile Object lastLine;
 
     /** whether this frame has been captured into a binding **/
     boolean captured;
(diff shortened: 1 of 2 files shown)


headius:
As mentioned in the description and primary commit, the main performance impact here will probably come from single-thread uses of captured procs alongside backref/lastline updating, since that will force the creation of a new ThreadLocal and subsequent accesses through it without need.

An additional change here could be used to reduce this cost for single-thread captured blocks with a moderate increase in cost for multi-thread captured blocks.

The captured flag could be used to indicate that this state is now "thread-risky" and additional measures taken to ensure only one thread is actually accessing the values in question. This would probably require tracking the original capturing thread in some way, and perhaps only allowing that thread to update the field directly. Once a second thread attempts to update the field, it flips to a thread-local.

Unfortunately the cost of tracking that thread may eliminate any gains: we could use a thread ID, but it may get reused, and if we use the thread itself we'd need to hold that reference weakly.

Ultimately we want to do as much as possible to keep the normal, non-captured, single-threaded case as fast as possible, so it may be acceptable to sacrifice some performance for captured procs that also set backref.

(Note that the ThreadLocal construction will only ever happen once per frame, so that further reduces the impact.)

headius:
Note also this makes some frame elimination impossible, but we currently don't fully eliminate frames used by blocks anyway.

headius:
The failure is in a regression test that ensures JRuby runtimes can be fully cleaned up by GC once abandoned. It seems like weak references will indeed be necessary within the ThreadLocal, since otherwise the value (and along with it the entire JRuby runtime) get rooted by any thread that touches this logic.

This will substantially increase the cost of this fix.

headius:
Ok, after some discussion I am killing this idea for now.

It would be nice to make these frame fields thread-local (and weak) but I feel the cost is too high and we are patching around a problematic design. A better approach would be to introduce a separate stack or structure for maintaining thread-local + frame-local backref and lastline values rather than allowing them to live in a heap structure that gets shared across threads. Basically, if we never want backref and lastline to be seen across threads, they should live outside of the Frame object that does get shared across threads.

We can revisit this PR in the future if needed, but I would like to think we can do a better implementation of backref/lastline that does not (cannot!) leak data across threads.

PR #6340, opened 2020-07-23
[ci] an attempt to run tests on Windows

kares:
a bit of Windows CI de-ja-vu, this time trying to run `test:jruby` 
chosen JRuby suite as it is very stable, and in out control, compared to say ruby-specs

motivated by: https://github.com/jruby/jruby/issues/6322 (can re-target to `jruby-9.2` if needed)

PR #6357, opened 2020-08-14
RubySpec fixes for Windows

headius:
This PR is a WIP to get specs running green on Windows.


headius:
The specs are now running to completion (for the "fast" subset, anyway) with around 150 failures and errors. The failures are largely in areas we expect: process launching, low-level IO operations, and a large number from win32ole which does not appear to boot properly for these specs.

I will merge what I have here since rake and mspec are working. Once #6340 lands we can revisit the failing specs and get a tagged run going.

PR #6359, opened 2020-08-16
Tag off remaining json failures.

k77ch7:
These seem to have fixed in JRuby 9.2.8.0 .


kares:
Nice, good to know json is progressing. Thanks.

PR #6360, opened 2020-08-17
Fix inconsistency in #5837

T45K:
In #5837, only `getSymbol(ByteList bytes, boolean hard)` method was modified, and its similar method, `getSymbol(ByteList bytes, ObjBooleanConsumer<RubySymbol> handler, boolean hard)` was not modified.
I think the method has a potential bug, so I fixed it.


headius:
Seems good to me, and we should probably try to move away from using ByteList.toString since it is confusing that it returns a raw ISO-8859-1 string regardless of the encoding.

PR #6361, opened 2020-08-20
Remove redundant code

T45K:
To judge from #5384, there is redundant code in `LoadFrameClosureInstr#clone`.


enebo:
@T45K Can you explain why that NOP should be removed?    My memory is foggy on details but I think this inline will happen to make a new inlined version of a method which is having the closure (callClosure) inlined into it.  In that case I think the literal closure will be instrs in the new method and NOP is ok....except if something else tries to consume it as a proc within that method.  I think that was what the message @subbuss had added was about.

So if I understand this correctly just having it always copy the closure means proc consumption of the block will still work?

T45K:
@enebo Thank you for your comment.

Why I made this modification was that `ReifyClosureInstr` and `LoadImplicitClosureInstr` were modified (NOP was removed) and `LoadFrameClosureInstr`, their similar class, have not been modified in #5384.
I have made this modification without understanding their behavior, sorry.

If you feel this change is unnecessary, please close this PR．

enebo:
I will close this because atm I broke inliner on master.  I will jot a note to consider this when I work on the inliner again though.  Part of me thinks we should remove the NOP but I will run into it when we hit a case where the NOP breaks code?

PR #6372, opened 2020-08-29
move socket message to constants

ahorek:
depends on https://github.com/jnr/jnr-constants/pull/71


headius:
jnr-constants has been updated on master, so we'll just push this on through. It's trivial once the jnr-constants values are there.

PR #6375, opened 2020-08-30
[ffi] update ffi to 1.13.1

ahorek:


larskanis:
I'm glad that ffi implementations are properly aligned by now. Hope it's not too much work to do these updates?

headius:
The one failure is the same sporadic failure we have seen in other builds. I will open a separate issue for that.

PR #6378, opened 2020-08-30
fix failing TestM17N#test_utf_without_bom_valid

k77ch7:
This fixes failling [TestM17N#test_utf_without_bom_valid](https://github.com/k77ch7/jruby/blob/master/test/mri/ruby/test_m17n.rb#L272-L277) .


headius:
The failure is a known sporadic failure that I've just now filed as #6382.

PR #6383, opened 2020-09-02
Fixes for non-native execution

headius:
Several key commands no longer work with native support disabled. We would like to allow as much of Ruby to run as possible when running in pure-Java mode.

These are fixes to better support the pure-Java runtime mode.


headius:
With the fixes here I am able to install gems including Rails, generate a Rails app, bundle install that app, boot the server and handle requests. There will be other fixes to be made but this is a good point to merge.

headius:
@enebo We should consider backporting these to the 9.2 line if we expect to do a 9.2.15.0 or higher, since they are small and may prevent issues like rubygems/rubygems#4129 from affecting future 9.2 maintenance updates.

PR #6388, opened 2020-09-10
Update JNR to Java 8+ versions

headius:


core/src/main/java/org/jruby/ext/socket/RubySocket.java
```diff
@@ -115,25 +117,25 @@ static void createSocket(Ruby runtime) {
...
 
-        rb_mConstants.setConstant("AI_PASSIVE", runtime.newFixnum(1));
-        rb_mConstants.setConstant("AI_CANONNAME", runtime.newFixnum(2));
-        rb_mConstants.setConstant("AI_NUMERICHOST", runtime.newFixnum(4));
-        rb_mConstants.setConstant("AI_ALL", runtime.newFixnum(256));
-        rb_mConstants.setConstant("AI_V4MAPPED_CFG", runtime.newFixnum(512));
-        rb_mConstants.setConstant("AI_ADDRCONFIG", runtime.newFixnum(1024));
-        rb_mConstants.setConstant("AI_V4MAPPED", runtime.newFixnum(2048));
-        rb_mConstants.setConstant("AI_NUMERICSERV", runtime.newFixnum(4096));
+        rb_mConstants.setConstant("AI_PASSIVE", runtime.newFixnum(AddressInfo.AI_PASSIVE));
```
ahorek, line 120: why not just add ```runtime.loadConstantSet(rb_mConstants, AddressInfo.class);``` etc. ?

headius:
The failed build at https://travis-ci.org/github/jruby/jruby/jobs/726387023 is a sporadic failure that may be a bad test. I have filed #6391 to deal with that at some point.

All libraries now appear to be working properly with releases targeted at "8" and automatic module names. I will proceed with releasing them.

PR #6389, opened 2020-09-11
[socket] remove misused SOL_ constants

ahorek:
fixes a freeze in ```test_multicast_send_and_receive``` https://github.com/jruby/jruby/pull/6388


headius:
Cherry-picked into #6388 to get that JNR stuff green. Thank you!

PR #6390, opened 2020-09-11
jrubyc: make valid java code when using annotations in java_signature

byteit101:
### Setup

```rb
 require 'java'
java_import 'javax.ws.rs.Path'
java_import 'javax.ws.rs.GET'
java_import 'javax.ws.rs.Produces'

java_package 'com.headius.demo.jersey'
class HelloWorld

  java_signature "@GET public java.lang.String cliched_message2()"
  def cliched_message2
    "Hello World"
  end
end
```

### Expected jrubyc --java output
```txt
...
@GET public java.lang.String cliched_message2() {
...
```
### Actual jrubyc --java output prior to this PR

```txt
...
@GETpublic java.lang.String cliched_message2() {
...
```
Note no space

This PR fixes the space and formats all annotations alike

### DIFF
--- a/lib/ruby/stdlib/jruby/compiler/java_class.rb
+++ b/lib/ruby/stdlib/jruby/compiler/java_class.rb
@@ -610,7 +610,7 @@ JAVA
     end
 
     def annotations_string
-      annotations.map { |anno| "@#{anno}" }.join("\n")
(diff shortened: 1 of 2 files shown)


headius:
Failure is same issue as #6391. I will review.

Issue #6386, opened 2020-09-08
Comparing Date object with nil sets global $! variable to NoMethodError


amitsuryavanshi:
`Date.today.compare_with_coercion nil` or `Date.today <=> nil` sets the ruby global variable `$!` to `#<NoMethodError: undefined method coerce for nil:NilClass>`. It does not throw the exception but sets the global `$!` variable. Now when I raise the exception in my app code the raised exception's `cause` is set to `#<NoMethodError: undefined method `coerce' for nil:NilClass>`. 

**Environment Information**
* JRuby version - jruby-9.2.13.0
* Operating system - MaC OS Catalina
* Rails - 6.0.3.2

**Expected Behavior**
- Raise the exception `#<NoMethodError: undefined method coerce for nil:NilClass>` rather than just setting it in `$!`

**Actual Behavior**

* Sets the ruby's `$!` variable to `#<NoMethodError: undefined method coerce for nil:NilClass>` but does not raises it.

**Script**
```
require 'rubygems'
require 'active_support'
require "date"

puts 'Before $! => '
puts  $!
Date.today <=> nil
puts 'After $! => '
puts  $!

```

Script Output :
```
Before $! => 

After $! => 
undefined method `coerce' for nil:NilClass
```

klobuczek:
Another way to illustrate the problem:
```
require "date"
require 'active_support'
begin
  Date.today <=> nil
  raise Exception
rescue Exception => e
  puts e.cause
end
```

The comparison does not raise any non rescued exceptions but it wrongfully sets the next Exception's `cause`. This does not happen on any MRI >= 2.5.0.

Unfortunately, I was not able to reproduce this behavior with plain exceptions without the Date comparison. There is something special about Date implementation.

headius:
Thanks @ahorek for the fix!

PR #6392, opened 2020-09-13
restore $! on date comparisons

ahorek:
fixes https://github.com/jruby/jruby/issues/6386


headius:
Fix is correct. Test could probably be made a spec, since CRuby likely also covers up the failed coercion, but I'll merge the fix now.

PR #6401, opened 2020-09-18
define available constants

ahorek:


headius:
I think this should probably check if the enum in question implements Constant before casting, and only use the defined logic in that case. It doesn't look like we use those methods with any non-Constant Enums but since we can't change the public API we should make it work in that case.

Otherwise it looks fine.

headius:
I realize now that master already did the blind cast to Constant, and I have a different change in mind. This is fine to merge.

headius:
I decided to go ahead and fix the constant set logic to enforce the Constants interface at the generic level. Since this method never worked unless the enum in question implemented Constant, I think it's best to just make that restriction explicit.

PR #6402, opened 2020-09-19
Clean up compile warnings and deprecations on Java 14

headius:
Lots of deprecation cleanup and a few more notable changes:

* The BigDecimal constants for rounding are now binding the ordinals of RoundingMode, since the integer values from java's BigDecimal are deprecated. Places that handle these constant values are updated to use `values()[ordinal]` instead of the legacy `valueOf(int)`
* `MulticastSocket.joinGroup(InetAddress)` is replaced with a call to `joinGroup(SocketAddress, NetworkInterface)` that should be equivalent.
* There are plans to make `yield` a reserved identifer, so bare calls to `yield` methods must be qualified.
* Two packages that were marked deprecated to trick javac into emitting a class file (and therefore not recompiling every time) are undeprecated, due to the many deprecation warnings they produce on recent javac versions.


headius:
The remaining deprecations include the following:

* `Object#finalize` which should be replaced by using the `Cleaner` class in Java 9+. No equivalent replacement exists for Java 8.
* `Package.getPackage(String)` is replaced by `ClassLoader.getDefinedPackage` but that is also Java 9+.
* There are still calls to our own `FileResource.inputStream` which was deprecated without a replacement.

Issue #6405, opened 2020-09-21
Thread kill does not release lock


JohnPhillips31416:
Hit a deadlock where jstack shows a number of threads waiting on a lock but no thread is holding the lock.  We were able to reproduce it with the code snippet below.  This snippet works (runs indefinitely) on jruby-9.1.17.0 and MRI 2.5.8, but deadlocks on jruby-9.2.9.0 and jruby-9.2.13.0

The suspicion is that the `t.kill` is killing the thread without releasing the mutex, possible related to this change https://github.com/jruby/jruby/commit/cc6bf49f8075e8de3acb0031753340bac994f179#diff-7768018ea9b120637b1da8461cb5df2b

```
def test_it
  mutex = Mutex.new
  sleep_div = 500
  starts = 0
  ends = 0
  2000.times do |i|
    ts = 3.times.map do
      Thread.new do
        sleep(rand / sleep_div)
        mutex.synchronize{ starts += 1; sleep(rand / sleep_div); ends += 1 }
      end
    end
    3.times do
      sleep(rand / sleep_div)
      t = ts.sample
      t.kill
    end
    if i % 10 == 0
      mutex.synchronize do
        puts "iter #{i}, starts: #{starts}, ends: #{ends}"
      end
    end
  end
end

if __FILE__ == $0
  test_it
end
```

JohnPhillips31416:
We only use thread kill in a few places so right now it looks like this isn't a blocker for us, we're going to try to refactor it out.

headius:
Commit cc6bf49 was also implicated by me in #6326 but the reporter was unable to provide a reproduction. Maybe now we can fix it!

stalbot:
I got lucky finding this, as we had just one instance of an actually exercised `Thread#kill` in our production code, right around a point where we were hitting a "re-locking by same thread" exception in a traceback. 

Interestingly, cc6bf49f8075e8de3acb0031753340bac994f179 was my guess about a culprit but I could also repro this on 9.2.8.0 but not 9.2.7.0. So that suggests it might be f1c7836096c81896a825fdaef283a69f74fb0ef6 instead? Pretty sure I double-checked 9.2.8.0 really did have the issue.

headius:
@stalbot Given your excellent reproduction script, I found the issue in about five minutes! Thank you! I will push a PR shortly.

headius:
@stalbot PR is in, but clearly we should add some kind of test for this.

[...]

Issue #6326, opened 2020-07-15
ActiveRecord ConnectionPool locking issue with 9.2.12.0


dgolombek:
**Environment Information**
  
Provide at least:
* JRuby version: jruby 9.2.12.0 (2.5.7)
* OS: Linux foo-i0b23b0e45022bfe9a 4.4.0-1110-aws #121-Ubuntu SMP Fri Jun 12 04:21:59 UTC 2020 x86_64 x86_64 x86_64 GNU/Linux
* ActiveRecord *: 4.2.11.3

**Expected Behavior**

ActiveRecord ConnectionPool locks should be acquired and released quickly, as happened with JRuby 9.1.17.0

**Actual Behavior**

Threads seem to block trying to acquire the ActiveRecord::ConnectionAdapters::ConnectionPool and being very slow to do so, OR something hidden is holding that lock. Our app explodes from ~115 threads to ~350 threads, with most of them either trying to acquire or release the lock on that class. The three locations we see most threads in:

1. https://github.com/rails/rails/blob/4-2-stable/activerecord/lib/active_record/connection_adapters/abstract/connection_pool.rb#L262
2. https://github.com/rails/rails/blob/4-2-stable/activerecord/lib/active_record/connection_adapters/abstract/connection_pool.rb#L280
3. https://github.com/rails/rails/blob/4-2-stable/activerecord/lib/active_record/connection_adapters/abstract/connection_pool.rb#L91

The full 15M Thread dump is at https://gist.githubusercontent.com/dgolombek/8b3ce3c68e7cd53438b758f1825061f5/raw/8aaa8524fd032772b97a5674f3667a1003921a0e/gistfile1.txt

We discussed this on #jruby:matrix.org starting at 2:31 EST on 2020-07-15.

headius:
From Matrix discussion, it seems like nearly all of these threads are blocked waiting for the connection pool lock in ActiveRecord. The unusual finding is that we could find no thread in the dump that has actually acquired the lock, raising the possibility that due to some logic error in either Rails or JRuby the lock has been locked and never unlocked by one of the other threads or by a thread which has exited.

A cursory inspection of `Mutex#lock`, `Mutex#unlock`, `Monitor#synchronize`, and related `Thread` code did not lead me anywhere. `synchronize` unlocks the acquired lock in an ensure block, and thread shutdown appears to have no way to skip unlocking all locks the thread has acquired. There's a possibility that the `handle_interrupt` logic in `synchronize` could be related.

[...]

PR #6407, opened 2020-09-21
Make sure locks are added immediately after acquisition

headius:
There's a race doing the lock add after the executeTask, since a thread event could interrupt the flow and bypass that logic. A finally would not be appropriate, since an interrupted attempt to acquire the lock would then stil add it. Instead we just move the lock add immediately after lockInterruptibly. If we reach this point, the lock is locked by the current thread, and must be added.

I am looking over the rest of the lock management code in Thread to see if there's other places that could be a risk, but this appears to fix the issue reported in #6405 and probably fixes the issue in #6326.


headius:
Tracked down the regression to bea9ad42f62e24a3963e135f2138c75a716f7790 (released in 9.2.8.0).

headius:
I made the unlocking at thread disposition more robust, but did not see any other risky pieces of code related to locking. We'll go with this.

PR #6418, opened 2020-09-29
Add zcat support based on CRuby logic

### LINKED ISSUE #6417: GzipReader fails to uncompress past first file in multi-file gzip
The basic issue is spelled out at the following CRuby issues and PRs:

https://bugs.ruby-lang.org/issues/9790

https://github.com/ruby/zlib/issues/7

https://github.com/ruby/zlib/pull/13

Basically, a file compressed using multi-file gzip will only decompress the first file using our GzipReader. Because our GzipReader is a loose port of CRuby code, which wraps a pure-Java zlib port (jzlib) we exhibit the same behavior as CRuby.

```
$ cat blah1.txt
hello

$ cat blah2.txt
world

$ gzip -c blah1.txt blah2.txt > blah.gz

$ gunzip -l blah.gz
  compressed uncompressed  ratio uncompressed_name
          72            6 -99.9% blah

$ jruby gzipreader.rb blah.gz
blah.gz line 1
"hello\n"
end-of-file reached
size: 6

$ gunzip --stdout blah.gz
hello
world
```

The solution for CRuby is to expose the `zcat` function to allow users to get that behavior. We will need to do something similar, though jzlib does not (I think) currently have such a function.

headius:
See ruby/zlib#13

Fixes #6417

This is a Ruby 3.0 feature, but it would be low-impact to merge it earlier.


headius:
Note this adds a test to CRuby's test_zlib that would come in when we update tests anyway, so that part could be omitted.

Issue #5270, opened 2018-08-02
Concrete JRuby Implementations of Abstract Java Classes Can't be Constructed With Reflection


ScottHaney:
### Environment
* JRuby: jruby 9.2.0.0 (2.5.0) 2018-05-24 81156a8 Java HotSpot(TM) 64-Bit Server VM 25.181-b13 on 1.8.0_181-b13 +jit [linux-x86_64]
* Operating system and platform: Linux shoes 4.15.0-23-generic #25-Ubuntu SMP Wed May 23 18:02:16 UTC 2018 x86_64 x86_64 x86_64 GNU/Linux

### Expected Behavior
Creating a JRuby class from an abstract Java class should have a constructor that can be called through reflection. This means that the following code should work `JRubyConcreteClass.java_class.constructor.new_instance` when `JRubyConcreteClass` derives from an abstract Java class.

### Actual Behavior
It complains that it can't instantiate the concrete class because it can't make an instance of the abstract base class. The error says "TypeError: can't make instance of AbstractClass". I have created a [repository that reproduces the issue](https://github.com/ScottHaney/examples/tree/master/jruby_abstract_class_error). Also overriding abstract java methods in JRuby doesn't seem to work either which is something that would be helpful to have.

### Potential Work Around for now
As a work around I am going to create concrete derived classes that directly implement the abstract methods and then add new setter type methods that JRuby code can then call to set data on. The data populated through these setters will then be used inside of the overridden abstract methods to get the desired behavior. This is a hack, but I think it will work for now.

I am also open to helping out with some development depending on what might be involved in fixing improving the behavior of implementations of abstract Java classes.

Thank you for your time and help!

monkstone:
I don't know whether it is any help to you but I've used abstract classes in JRubyArt/propane and this works for me https://ruby-processing.github.io/propane/classes/library_proxy

headius:
Ok the usual way we improve these features is by getting some new specs in place. Class extension has needed an overhaul for years but the specs are very weak. Can you folks collab to figure out both what we do now and what we *want* to do? The new impl is not terribly hard if we have good specs and can throw away the current impl.

monkstone:
OK just to clarify works for me as expected, but may'be that's my peculiar/particular case, I'd be happy to collaborate on a spec.

ScottHaney:
[...]

Issue #2369, opened 2014-12-29
Can't subclass a Java class that calls an abstract function in its constructor


headius:
From http://jira.codehaus.org/browse/JRUBY-2748
# Original report

I have found that while subclassing a Java class that  calls an abstract function in its constructor will cause the initialization of the subclass in JRuby to throw an ArgumentError with message "Constructor invocation failed: null." 

Here is my Java base class:

```
package com.allstontrading.logalyzer;
import java.util.List;

public abstract class AbstractTest {
     protected String obj;

     public AbstractTest() {
          this("dog");
          this.run();
     }
     public AbstractTest(String obj) {
          this.obj = obj;
     }

     protected abstract List<String> myAbstractFunc();
     public void run() {
          if (this.myAbstractFunc() != null)
               for (String string : this.myAbstractFunc())
                    System.out.println("Big " + this.obj + "s Eat: " + string);
          else
               System.out.println("OMG");
     }
}
```

Here is my example JRuby script:

```
include Java

import 'com.allstontrading.logalyzer.AbstractTest'
import 'java.util.ArrayList'

class RubyTest < com.allstontrading.logalyzer.AbstractTest
    def initialize()
        super("Pink Elephant")
    end

    def myAbstractFunc()
        list = ArrayList.new
        list << "Cancerous Rocks"
        list << "Candy Apples"
        list << "Moldy Bread"
        return list
    end
end

puts "This test does NOT call an abstract function in the constructor."
test = RubyTest.new
test.run

class AnotherRubyTest < com.allstontrading.logalyzer.AbstractTest
    def initialize()
        super
    end

    def myAbstractFunc()
        list = ArrayList.new
        list << "Cancerous Rocks"
        list << "Candy Apples"
        list << "Moldy Bread"
        return list
    end 
end     

puts ""
puts ""
puts "This test does call an abstract function in the constructor."
anotherTest = AnotherRubyTest.new
```

Here is the output:

```
This test does NOT call an abstract function in the constructor.
Big Pink Elephants Eat: Cancerous Rocks
Big Pink Elephants Eat: Candy Apples
Big Pink Elephants Eat: Moldy Bread

This test does call an abstract function in the constructor.
/home/mbohn/jruby-1.1.2/lib/ruby/site_ruby/1.8/builtin/javasupport/utilities/base.rb:26:in `new_instance2': Constructor invocation failed: null (ArgumentError)
[...]

Issue #449, opened 2012-12-16
support newInstance() with arguments


byteit101:
Java cannot call newInstance on a ruby class with arguments (rubyClass.newInstance(object) will throw an exception)

byteit101:
Context from JavaFX 7 jar dump:
```java
private static void launchApplication1(Class<? extends Application> paramClass, Class<? extends Preloader> paramClass1, String[] paramArrayOfString)
    throws Exception
  {
...
          Constructor localConstructor1 = paramClass1.getConstructor(new Class[0]);
          localPreloader = (Preloader)localConstructor1.newInstance(new Object[0]); /// THIS is where JavaFX fails to call ruby when paramClass is a JRuby class

          ParametersImpl.registerParameters(localPreloader, new ParametersImpl(paramArrayOfString));
```

If we pass javafx.Application::launch() a JRuby class (based on javafx.Application), it will fail at the above line in com.sun.javafx.application.LauncherImpl

byteit101:
And here is the java 8 source, which is similar, but I've not tested it and it might work: http://hg.openjdk.java.net/openjfx/8/master/rt/file/f89b7dc932af/modules/graphics/src/main/java/com/sun/javafx/application/LauncherImpl.java#l791

byteit101:
Here is a quick test case and the full error source:
Original test case:
```rb
require 'jruby/core_ext'
class App2 < javafx.application.Application; def start(stage); end; end
App2.become_java!
javafx.application.Application.launch(App2.java_class.to_java(java.lang.Class))
```
That case error:
```
Exception in Application constructor
Java::JavaLang::RuntimeException: Unable to construct Application instance: class javafx.application.Application
	from com.sun.javafx.application.LauncherImpl.launchApplication1(com/sun/javafx/application/LauncherImpl.java:907)
	from com.sun.javafx.application.LauncherImpl.lambda$launchApplication$155(com/sun/javafx/application/LauncherImpl.java:182)
	from java.lang.Thread.run(java/lang/Thread.java:745)

```

Reduced case:
```rb
require 'jruby/core_ext'
class App2 < javafx.application.Application; def start(stage); end; end
App2.become_java!
App2.java_class.to_java(java.lang.Class).getConstructor([].to_java(java.lang.Class)).newInstance([].to_java java.lang.Object)
```

The error I get
```
Java::JavaLang::InstantiationException: 
	from java.lang.reflect.Constructor.newInstance(java/lang/reflect/Constructor.java:423)
[...]

PR #6422, opened 2020-10-02
Java-constructable Concrete Ruby Classes

byteit101:
Work on #449, duplicate #5270, and related #2369. See #449 for latest discussion.

Missing: Working code, Tests, and a good design. 

Right now this is very hacky.

Notes to watch out for:

https://github.com/jruby/jruby/wiki/FAQs#when-i-implement-a-java-interface-and-provide-my-own-initialize-method-i-cant-construct-it-anymore

interface methods must be in :methods (vs all)


byteit101:
I'm now moving the concrete extension code into the reification code. So far I have the following pattern implement:
Normal Ruby class:

```java
public class NormalReify2 extends RubyObject implements Reified { // this is unchanged, and we are a IRubyObject
   private static Ruby ruby;
   private static RubyClass rubyClass;

   public static void clinit(Ruby var0, RubyClass var1) {
      ruby = var0;
      rubyClass = var1;
   }

   public NormalReify2(Ruby var1, RubyClass var2) { //initialized via the RubyAlllocator, which is a change from before to allow java/init calls to not collide
      super(var1, var2);
// ruby called us, we will be initialized later
   }

   public NormalReify2() { //java-called
      super(ruby, rubyClass);
      this.callMethod("initialize"); // initialize as java called us
   }

   public int mymethod(String var1) {
      Ruby var2 = ruby;
      return ((Number)this.callMethod("mymethod", new IRubyObject[]{JavaUtil.convertJavaToUsableRubyObject(var2, var1)}).toJava(Integer.TYPE)).intValue();
   }
```

Concrete Extension:

```java
public class RubyTest extends AbstractTest implements Reified { // Unsure of which interfaces should be implemented
   private static Ruby ruby;
   private static RubyClass rubyClass;
   private RubyBasicObject rubyObject; // in theory should be IRubyObject, but not worrying about the threadcontext is nice :-)

   public static void clinit(Ruby var0, RubyClass var1) { // to be replaced by static{} block
      ruby = var0;
      rubyClass = var1;
   }

   public RubyTest(Ruby var1, RubyClass var2) { // ruby facing ctor
      if (this.rubyObject == null) {
         this.rubyObject = new ConcreteJavaProxy(var1, var2, this); //note this pulls allocation out of an Allocator object, is that fine? what references, doc should there be?
      }
// no init, as ruby will call it later. this is part of the alloc
   }

   public RubyTest() { // java-facing ctor
      if (this.rubyObject == null) { // eh, lazy codegen that should be cleaned up
         (this.rubyObject = new ConcreteJavaProxy(ruby, rubyClass, this)).callMethod("initialize"); // init here, as java called us
      }
   }

   public List myAbstractFunc() { // an overridden function
      if (this.rubyObject == null) { // I think this could use some optimization, but this is here in case the parent calls methods in our super
         this.rubyObject = new ConcreteJavaProxy(ruby, rubyClass, this);
      }

      Ruby var1 = ruby;
      return (List)this.rubyObject.callMethod("myAbstractFunc", IRubyObject.NULL_ARRAY).toJava(List.class); // Should the method lookup be there every time?
   }

   public static IRubyObject __allocate__(Ruby var0, RubyClass var1) { //uses StaticRubyAllocator
      return (new RubyTest(var0, var1)).rubyObject; // Note: rubyObject has a ref to `this`, so no GC worries
   }
```

Ruby API:
```rb
class MyClass
   configure_java_class methods: :explicit/:all, call_init: true/false, java_constructable: true/false
   # sets JavaClassConfiguration
end
```

byteit101:
Of note, given this java class:
```java
class MyClass {
  MyClass(String s){
    doAbstract();
  }
  protected abstract doAbstract();
}
```

and this JRuby class:
```rb
class MyRuby < MyClass
  def initialize(s)
    puts "in init"
  end
  def doAbstract
    puts "in abstract"
  end
end
```

the resulting output, with my current code is:
```txt
in abstract
in init
```
which is weird. 

Also, super() is a no-op in ctors on concrete classes now, which also strikes me as weird

byteit101:
I have three sets of questions remaining: Architecture, Design, and Implementation, in order of importance

## Architecture

### Java Init Sequence == Ruby Init Sequence?
As of the way I've implemented it right now, a ruby init sequence is `.reify() unless reified` + a call to the java init sequence. This is:
```txt
ReifiedClass.<init>(*args)
  super(*args)
  this.rubyObject = new ConcreteJavaProxy(this)
  rubyObject#initialize(*args)
    # super() calls are *SILENTLY* no-ops
```
Alloc no longer exists [1]
I am unsure if __jreate! still exists in a useable form

I debated making super NOT a no-op, but that would involve a lot of code gen + super-splitting that I wanted to ask about before I went down that rabbit hole. I also didn't like only making it a no-op in java init sequences, as it seemed too context sensitive and prone to confusion. Note this is a behavioral change, and I don't like it.

In theory, we could separate args == 0, as super is a no-op there (probably[2]), but I feel like it could lead to a confusing mental model of args == 0 being different from args > 0.

1: I have no idea if this is enforced/rumps are still around
2: Weird static init stuff before super that the super depends on makes this not 100% true

### Super-splitting
This is an idea I had, but am unsure of how to go about this or if it's an ok idea.
in essence, to allow `#initialize` to call `super`, we need to dispatch the types in code-gen. `this` isn't usable before `super`.

The general format of generated code will turn into:

```txt
ReifiedClass.<init>(*args)
  this.rubyObject = new ConcreteJavaProxy()
  superargs, continuation = rubyObject#initialize_with_split_super(*args)
  switch(guessSignatureFromArgs(*superargs))
    when V -> super()
    when LString -> super(*convert(superargs))
    when LString;Z -> super(*convert(superargs))
    etc...
   end
  rubyObject.setJavaObject(this)
  continuation.call
```

There are 3 ways I though of implementing this:

 1. AST/IR transforms as `super` is a keyword and can't be hidden in other methods (I think?) 
 2. Explicit new method/api (see comments in #449 )
 3. Fibers

I think #3 is out due to overhead, though it would work today and is the most minimally invasive. Howerver `ThreadLocal` Storage would be disrupted.
2 is the most user-hostile in terms of backwards compatibility and extra stuff to think about, but is the simplest to implement
Despite some poking around the parser, I am unsure of the ability of 1, and it needs to be re-computable in case someone changes the definition of `#initialize`. It would work roughly by transforming as async/await do:

```rb
def initialize(*args)
     variable = OtherClass.call_static(*args)
     super(variable, keep_me = args[0] + 5)
     self.to_java.java_method(keep_me, variable) # values retained
end
def initialize_transformed(*args) # No need to change the return type, as initialize doesn't return values
     variable = OtherClass.call_static(*args)
     return [ [variable, keep_me = args[0] + 5], lambda {
         self.to_java.java_method(keep_me, variable) # values retained
    }
end
```

### New Method names => new generated class
The old concrete extension code (now deleted in this PR) used to regenerate when a new method was added. The reify code doesn't appear to do this. Why the difference?

## Design
### Reify options
Right now things are configured as such:
```rb
class MyClass
   configure_java_class methods: :explicit/:all, call_init: true/false, java_constructable: true/false
end
```
I was also thinking of supporting the same options (to override the ones set there) in `become_java!(methods: ...etc)`

Where should the options be configured?

Also, are those reasonable options? (for all: https://github.com/jruby/jruby/pull/6422/files#diff-fa6783cb1e5bbf966f341330a90265a3aa04682625c103500fe61d5e8602ef96 and for what they do (this area of the code): https://github.com/jruby/jruby/pull/6422/files#diff-4823dd97065ce1b80e235b191b03497a85a35a0acd1e8089e09eef1365b699ecR1564-R2165)

## Implementation

### `new` method
I was having issues setting up the correct new method, which leads to double dispatch. Is there a way to preemptively separate these concerns: https://github.com/jruby/jruby/pull/6422/files#diff-ee5b3e6e296d6bf1315743db55b99867762ade5bdfedec34c19bef47ffc8ae98R113

### Performance
Where should I worry about it? I see some places use `DynamicMethod`s, others `CachingCallSite`s. Should I support InDy in any way?


Not mentioned: the smaller and less important issues and todo's scattered around the code right now. I want to get the big questions answered first :-)

byteit101:
Added super-splitting proposal code-gen, and some tests on how option 1 holds up. Answer: super sketchy with it being enforced to be just like java for that method to work (no super in if's, etc). I though about maybe an alternative to fibers, but that would require us to implement a continuation in super, precluding java calls on the stack frame between <init> and the super call.

byteit101:
A summary of matrix conversation with @headius:

 * Init sequence in Java and Ruby should be the same
 * Use split-super + AST/IR transforms (option 1)
   * Talk to @enebo about doing IR transforms cleanly for this
 * in the future, look into hard-jitting initialize into the java <init>
 * "hook-in" methods as proposed in #449 are mostly off the table as the break the initialize API (likely used a lot)

Of note this:
 * Doesn't break API for `super()` in ruby
 * Doesn't break API for `#initialize` allowing `super()` calls in ruby
 * Allows dynamic method changing of `#initialize`
 * Is a lot more work than other options
 * Is necessary because you can't pass uninitializedThis from `<init>` outside of the current frame in bytecode

byteit101:
Things to do before this is finished, roughly in order:

 * [ ] Fix AST/IR transforms with @enebo 's help
 * [ ] Fix broken tests
 * [ ] Add new tests
 * [ ] Ensure inheritance isn't broken
 * [ ] Make sure JRubyFX & FXML are supported with the changes
 * [ ] Ensure error messages are correct
 * [ ] Remove debugging println's and printStackTrace
 * [ ] Reformat code

enebo:
From todays chat:

```text
We have an AST solution but AST is not desirable as a finished product.  It is good enough in WIP/PR form to debug and figure out what other problems we may encounter
The basic problem is we want to execute a front and back portion of some initialize methods and in IR we need to demarcate and split those apart and be able to execute them separately
The basic outline of what your generated code does now would likely be very similar but we might need to return temp array and maybe a couple of other things
so we can pass them to however we execute the second half
We need to consider error reporting and if we can make that descriptive or generic
As for splitting I think we can use a special instr (or add to runtimehelperinstr) to mark the super during IRBuild if the method is named initialize
Then the split itself should not be difficult we use startup instrs and just cleave it into two
I think that is the current state from this conversation and most of it is just considering an IR version
I think also we can initially do this as interp-only to reduce the scope
byteit101
Things I just realized would be really hard to support: super in a loop
But I will need to think through try/catch semantics
enebo (@enebo:matrix.org)
heh impossible in Java so there is that
byteit101
yea, if you try to compile the decompiled bytecode in javac, it doesn't like having code before the super() :-)
^ from the balloon example I pasted above
enebo (@enebo:matrix.org)
It may be that IR helps us a little bit that if we can detect a branch instr which jumps past the super instr or we see a jump instr which targets something above a super instr we can error
```

byteit101:
Most existing tests pass, but a few brought up questions:

## Design

### Matching non-`java_signature`ed methods to java methods
There are 2 obvious-to-me solutions:

 1. All java method signatures in the hierarchy of a given name are overloaded
 2. Only the correct arity method signatures in the hierarchy of a given name are overloaded

Two tests show that the current behavior is a fail-through of 2 then 1:

From spec/j_i/types/extension_spec.rb:
  it "dispatches to other-arity superclass methods if arities mismatch" do
  it "raises argument error if no matching arity method has been implemented on class or superclass" do  # JRUBY-4788

The first test makes sense to me, but the second surprises me slightly that the reified class should generate the java proxy for the wrong arity. Is this intentional? I can't seem to find much about that old issue beyond `JRUBY-4788 Stackoverflow when calling code from java with wrong parameters` which suggests to me that other exceptions thrown could be acceptable?

This issue brings up the broader question of should we generate everything possible to override in java land or only generate what we have defined right now from ruby land? The latter is mostly what is on this branch, and not def'ing an abstract method or interface method will lead to exciting java AbstractMethodException's, but test #1 above implies it shouldn't be the former to me.

Also leads directly into...

### Generate all or only necessary?
```
rspec ./spec/java_integration/types/classgen_spec.rb:90 # A Ruby subclass of a Java concrete class with custom methods generates new methods usable by ruby & java
```
That test fails with "too few methods found" if I generate only the necessary methods, and "too many methods" if I generate all (internal class methods are exported as well). I saw the old concrete code filtered out the internals. Should that still be done in the exact same way, or should it be changed, along with that test

### JavaClass.clone

JavaClass.clone backs up a level, in that java classes are known by their reified (java-variable) value, which isn't copied on clone. All method signatures are also lost. Is this intentional?

Concretely, this code generates the below defs
```rb
java.util.ArrayList.new
Cloned = java.util.ArrayList.clone
Cloned.new
```
```java
class Cloned extends java.util.AbstractList {
 // default ctors generated, but omitted for brevity
}
```
Obviously, attempting to use the cloned class fails with AbstractMethod errors.

```txt
jruby-9.2.7.0 :031 > JRuby.reference0(Cloned).getReifiedClass
 => nil 
jruby-9.2.7.0 :032 > JRuby.reference0(java.util.ArrayList).getReifiedClass
 => class java.util.ArrayList 
```

(the ".new" worked previously as __jcreate! had a copy of the class in it's instance, but nobody else did. The JCreate method was copied in the clone)

This can be solved with some implications by copying the java fields to the new class in the clone. Is that ok?

byteit101:
I still need to reformat the code, but I have several questions first:

# Ruby Syntax
This is the current usage of how to configure what is included in a class, and what methods are called when invoked from java:
```
class Foo
# normal usage
  configure_java_class ctor_name: "j_initialize"

  # All options, defaults
  configure_java_class methods: :all, ctors: :all, java_constructable: true, ruby_constructable: true, ctor_name: :initialize, proxy_dispatches: {}, includes: [], excludes: ["see-sources, it's a long list"]

  # All options, alternates
  configure_java_class methods: :explicit, ctors: :explicit, java_constructable: false, ruby_constructable: false, ctor_name: "a_string_or_symbol", proxy_dispatches: {a: "hash"}, includes: ["initialize", ...], excludes: ["my_method"]

  # block config
  configure_java_class methods: :all, ctors: :all, java_constructable: true, ruby_constructable: true, ctor_name: :initialize do
    dispatch "initialize", "not_initialize
    dispatch key: :value
    include :initialize
    exclude :initialize
  end
```
Thoughts?

# General/Not File Specific
How should I set allocators, the new method, and initialize? ...

# RubyClass.java

## General
ReifiedJavaProxy is-a Reified. Is that a good idea, or bad?

## initialize_copy()
See clone question above: https://github.com/jruby/jruby/pull/6422#issuecomment-734996219

# reify()
See the lines in the 1350's "parents differ!"
Should I be using `reifiedParent` or `parentReified`? I don't quite understand the difference, but the method/concrete difference in usage seems to work for all cases I've tested.


RubyClass.Reifier.Interfaces //TODO: filter interfaces as per old code?


For this version, only concrete extension allows java to pass args // TODO: allow java to pass args?

All fields are public? // TODO: public?

What methods/ctors should we define? (lots of TODOs
	RubyClass.ConcreteReifier.findTarget: //TODO: final? (Should we allow overwriting finals?)

Where to stick the static int lookup

generated code style, SkinnyMethodAdapter, vs GeneratorAdapter?


--
RubyModule.searchMethodLateral: a way to get stuff on up to the superclass, but not includeing. Good? caching?

--
SkinnyMethodAdapter: I changed it, but see it's also in jnr-ffi. Are those copied in any way or can they diverge?



---
RealClassGenerator
What exception should "no matching supercalls" be?

RCG: toJavaArrayUnsafe? toJavaArray? toJavaArrayMaybeUnsafe?


--
RubyToJavaInvoker: Performance? (will pull out)

--
ConcreteJavaProxy/general: initialize, jinitialize, jcreate?
CJP is a mess because I don't know how to install the proper handlers

--
JavaProxyClass.getConstructors: filter out non-ruby ctors here?/isExportable

JavaProxyClass.get: removed! remove?
JavaProxyClass setProxyClassReified: have this here? is this how to save the proxy class? no use of method keys?
JavaProxyClass: new/initialize install checking big issue

JavaProxyClass.getProxyClass: now doesn't generate, only retreives. Is that ok? Are any of the mixin exclusions needed to be ported?

-----
Enebo:

AbstractIRMethod: Where should the impl be? Abstract? each subclass? currently later, CompiledIR sticks out

---------

byteit101:
@headius: This is all of the initialization flow options as per this PR. Let me know if you have questions.

jrubysequencepr

(Related: my original idea at https://github.com/jruby/jruby/issues/449#issuecomment-700315740)

headius:
Ok finally circling back to final review and questions. I will do a first pass over things I know I can answer now.

Ruby Syntax
-------------

The syntax and options seem logical to me but I would like to get some other opinions from @kares and @enebo at least.

General/Not File Specific
-------------------------

> How should I set allocators, the new method, and initialize?

If this is a concern about overwriting the ones that are there that is fine to do for allocate and new. Writing them directly to the proxy class is correct. If you can be more specific about your concern I can answer other questions.

> I created __jallocate!, different from initialize and __jcreate! in that it doesn't call any ruby methods. I'm unsure how to get __jcreate! to not call initialize

The `_jallocate!` name seems fine to me, is there any other question here?

> What generated code style to use: SkinnyMethodAdapter vs GeneratorAdapter?

No preference. When modifying existing code, obviously continue to use the same things. Unifying all our code generation to use one adapter would be nice (we also use jitescript in a few places which is similar) but that is a separate project.

> I'm seeing an extra stack frame with the name of the ctor, but the line number of the new call, unsure what do to?

In Java or Ruby? If this is a Ruby frame being mined from the Java stack trace, then it just uses the class and method name to find frames that are interesting in Ruby. Eliminating a duplicate frame would require a different method name we don't include in the Ruby trace, or new logic in the exception stack trace generator to ignore these duplicate frames. An example would help me understand what's happening.

RubyClass.java
---------------

> ReifiedJavaProxy is-a Reified. Is that a good idea, or bad?

It seems logical to me. The reified marker is so we know that there is a specific concrete class that goes with this Ruby class, which is the case here.

> For this version, only concrete extension allows java to pass args, not ruby Object-based classes. I think that's fine for now?

Yes I think that is fine for now. So basically this means that reified classes can only be constructed using a no-arg constructor, yes?

> What methods & ctors should we define? See above comment as well

In general I am leaning toward just defining the methods that the user wants to override, rather than defining all (or most) of them just in case the class overrides one in the future. So the question becomes whether to preserve the current behavior, which we discussed on Matrix... you have added filtering to keep methods like `java_class` and such from getting exported, and I think that is fine.

> What should we default to calling for normal ::new calls? IRO constructor? allocator via allocate?

Standard protocol would be allocate, which is usually just hooked up to the IRO constructor (and increasingly using lambda method references for the constructor, so there's not even an intermediate).

headius:
More answers...

RubyClass.java
---------------

> What should we default to calling for normal ::new calls? IRO constructor? allocator via allocate?

Revisiting this since you clarified what the IRO constructor is...

Calling via the allocator still seems like the logical default since that is the standard behavior. Do you have any reasons why we should use the IRO constructor instead?

> Where to stick the static int lookup? Right now it is spread across several files, and it seems hacky. JavaProxyClass.setProxyClass/RubyClass.reify

As discussed on Matrix, switch this to using a ThreadLocal to pass these "out of band data" objects through to the class init and clear the threadlocal after `<clinit>` has run (or clear it during <clinit> once the final fields have been initialized).

headius:
> initialize_copy()
> See clone question above: #6422 (comment)

Discussed on Matrix...

The problem here is that when cloning the Java proxy class we do not propagate the reified class, so the new class looks like a new sibling class of the original, except that it doesn't override any of the methods the original did and we generate a new concrete subclass. I think it should carry the reified class along and just be a duplicate of the proxy that can be further modified in isolation.

This is not really a use case we have actively tried to deal with. I guess I would say we should try not to break current behavior, but outside of that I am not confident I even know what behavior users would expect.

byteit101:
> Calling via the allocator still seems like the logical default since that is the standard behavior. Do you have any reasons why we should use the IRO constructor instead?

I don't know how to override the new/initialize/jcreate/etc as per a previous question to use allocate, but I did successfully plumb via the IRO constructor :-)
If you would like to move back to the allocator method, I will need help figuring out how to do that

headius:
> I don't know how to override the new/initialize/jcreate/etc as per a previous question to use allocate, but I did successfully plumb via the IRO constructor :-)

That is fine then. This is a very specialized initialization path so it is understandable that it would deviate from the typical allocate sequence.

headius:
> Should I be using reifiedParent or parentReified?

Yeah this is confusing to me as well. The only difference between the two is that one (parentReified) is from `superClass.getRealClass()` which will skip over prepended modules and singletons and such. I think we should only be using this one, since singletons do not reify separately and prepends should only be meaningful for object dispatch against the same class.

The behavioral difference between these two reified classes probably never showed up because it would be very rare to see a class that is both a concrete extension AND prepended/singletonize (and probably impossible to singletonize since usually that would need an instance of the class already).

So basically they are identical unless prepends/singletons are in play, in which case we should always be using `parentReified`.

> searchMethodLateral

I don't believe this functionality exists anywhere currently.

> SkinnyMethodAdapter

Ignore the jnr-ffi version.

This comes back to the project of unifying bytecode generation to use only one of these (which should probably be the adapter shipped with ASM).

> RealClassGenerator

> What exception should "no matching supercalls" be?

NoMethodError I guess? There's not much variety here in the Ruby exceptions.

> RubyArray.toJavaArrayUnsafe? RubyArray.toJavaArray? RubyArray.toJavaArrayMaybeUnsafe? Which should I use?

`toJavaArray` always gives you a copy of the relevant range of the internal array. If you know more about the contents you can use one of the others to avoid the allocation and copying.

`toJavaArrayMaybeUnsafe` gives you the direct array only if it is not shared and the valid range being viewed is the entire array (i.e. begin == 0 and RubyArray.realLength == array.length). If those conditions are not met you get a copy of the relevant range. Use this to guarantee you get a Java array that holds exactly only the viewed elements, and possibly avoid allocation by using the actual direct array.

`toJavaArrayUnsafe` ignores the viewed range and gives you the direct array unless it is shared, in which case you get a copy. Use this if you know it is not shared and want to work with the raw array directly. You must deal with the possibility that `begin` is not zero.

I have added some doco.

> RubyToJavaInvoker...

> Is this a valid optimization?

Seems valid to me!

> Should there be something for the primitives?
> For the specialized case?

Perhaps? I would focus on the most common cases, which might mean varargs from 0, 1, 2, and 3 offsets. Primitives maybe? Mostly the thing that stops us from specializing for everything is the amount of duplication that results.

If you can show perf improvements as high as 15% in a real app, that is a pretty big win. Be pragmatic and don't overoptimize if it doesn't provide obvious gains for non-contrived cases.

> CJP is a mess because I don't know how to install the proper handlers, see that in General above

Perhaps we should discuss this again...

headius:
> JavaProxyClass

> getConstructors()

This only appears to be used within the existing concrete subclass logic. I wouldn't worry about it too much. If you don't need it, it could probably be removed.

See the existing logic in JCreateMethod that uses it to pick a constructor to call.

> get()

This appears to do nothing but get the java.lang.Class for the concrete subclass. Do with it as you will, since the resulting class was not very useful on its own before.

> setProxyClassReified()

Need to discuss this to understand what you need.

> getProxyClass()

> This method now doesn't generate, only retrieves. Is that ok?

Seems fine. Lots of places in Java integration were made lazy unnecessarily years ago.

> Are any of the mixin exclusions needed to be ported?

Not sure what you mean, let's chat about it.

...

Other than things I need to discuss with you for clarification, I think I have addressed everything. Sorry for the delay in getting these questions resolved! Let's try to get this thing landed ASAP.

byteit101:
> > I'm seeing an extra stack frame with the name of the ctor, but the line number of the new call, unsure what do to?
> 
> In Java or Ruby? If this is a Ruby frame being mined from the Java stack trace, then it just uses the class and method name to find frames that are interesting in Ruby. Eliminating a duplicate frame would require a different method name we don't include in the Ruby trace, or new logic in the exception stack trace generator to ignore these duplicate frames. An example would help me understand what's happening.

I think this is fixed, but now I realized that exceptions are wrapped & stacks tossed through a java ctor. Should they not be wrapped?

```rb
require 'java'
require 'jruby/core_ext'

class Test < java.lang.Object
	def initialize
		doit
	end
	
	def doit
		raise "my error"
	end
end

Test.new
# ArgumentError: Constructor invocation failed: (RuntimeError) my error
#  <main> at Test.rb:14
class TestR < Object
	def initialize
		doit
	end
	
	def doit
		raise "my error"
	end
end

TestR.new
#
#RuntimeError: my error
#        doit at test.rb:23
#  initialize at test.rb:19
#         new at org/jruby/RubyClass.java:884
#      <main> at test.rb:27
```



> > For this version, only concrete extension allows java to pass args, not ruby Object-based classes. I think that's fine for now?
>
> Yes I think that is fine for now. So basically this means that reified classes can only be constructed using a no-arg constructor, yes?

Concrete classes can be constructed with any super-arg ctor, plus manually specified, from java. Non-concrete can only be constructed with a no-arg ctor.

> >    RubyArray.toJavaArrayUnsafe? RubyArray.toJavaArray? RubyArray.toJavaArrayMaybeUnsafe? Which should I use?

I'll stay with toJavaArrayUnsafe for my usage then

> >  get()
>
> This appears to do nothing but get the java.lang.Class for the concrete subclass. Do with it as you will, since the resulting class was not very useful on its own before.

Removed!

> > This method now doesn't generate, only retrieves. Is that ok?
>
> Seems fine. Lots of places in Java integration were made lazy unnecessarily years ago.

To clarify: this used to eagerly generate before my PR, but my PR no longer makes it generate, only retreives any previously-generated data

@enebo `super(..) if cond` is not being caught as an error. Is there something I need to tie in to catch that?

headius:
> I realized that exceptions are wrapped & stacks tossed through a java ctor

As in other places we call through Java code, we should allow all exceptions to proceed unwrapped. This should restore the Ruby trace for your first example.

We used to post-process Java exceptions coming from Java integration calls but it required us to be catching every one and rewriting it, which then broke Java consumers of those exceptions further up the stack.

headius:
Here are the Windows GHA failures as of this moment:

```
 4) Error: test_interface_methods_seen(TestHigherJavasupport): TypeError: failed to coerce rubyobj.TestHigherJavasupport.A to org.jruby.javasupport.test.Interface2
D:/a/jruby/jruby/test/jruby/test_higher_javasupport.rb:1366:in `test_interface_methods_seen'
     1363:      ci = org.jruby.javasupport.test.ConsumeInterfaces.new
     1364:      ci.addInterface1(A.new)
     1365:      ci.addInterface1(B.new)
  => 1366:      ci.addInterface2(B.new)
     1367:      ci.addInterface1(C.new)
     1368:      ci.addInterface2(C.new)
     1369: 

 6) Error: test_lowercase_colon_package_syntax(TestHigherJavasupport): TypeError: failed to coerce rubyobj.TestHigherJavasupport.LCTestA to org.jruby.javasupport.test.Interface2
D:/a/jruby/jruby/test/jruby/test_higher_javasupport.rb:1403:in `test_lowercase_colon_package_syntax'
     1400:     assert_equal('boo!', LCTestB.new.boo)
     1401:     ci.addInterface1(LCTestA.new)
     1402:     ci.addInterface1(LCTestB.new)
  => 1403:     ci.addInterface2(LCTestB.new)
     1404:   end
     1405: 
     1406:   def test_marsal_java_object_fails

 7) Error: test_varargs_constructor_in_ruby_subclass(TestHigherJavasupport): ArgumentError: Constructor invocation failed: (TypeError) cannot convert instance of class org.jruby.RubyString to class [Ljava.lang.String;
D:/a/jruby/jruby/test/jruby/test_higher_javasupport.rb:203:in `initialize'
D:/a/jruby/jruby/test/jruby/test_higher_javasupport.rb:203:in `test_varargs_constructor_in_ruby_subclass'
     200:   end
     201: 
     202:   def test_varargs_constructor_in_ruby_subclass
  => 203:     var_args = RubyVarArgsCtor1.new 'foo', 'bar'
     204:     assert_equal 'foo', var_args.constants[0]
     205:     assert_equal 'bar', var_args.constants[1]
     206: 

 8) Error: test_varargs_overloaded_method(TestHigherJavasupport): ArgumentError: Constructor invocation failed: java.lang.reflect.InvocationTargetException
D:/a/jruby/jruby/test/jruby/test_higher_javasupport.rb:255:in `test_varargs_overloaded_method'
     252:     assert_equal 'foo', var_args.constants[0]
     253:     assert_equal 'bar', var_args.constants[1]
     254: 
  => 255:     var_args = RubyVarArgsOnlyCtor2.new
     256:     var_args.setConstants 'foo', 'bar' # (String, String)
     257:     assert_equal 'foo', var_args.constants[0]
     258:     assert_equal 'bar', var_args.constants[1]
```

I will try to confirm these should be passing. Not sure why Windows would have any differing behavior here.

headius:
Running `rake test:jruby` locally I get five failures that are similar to the Windows GHA build: https://gist.github.com/headius/13324fcf3b4ada1985eac7262388e77c

JI specs have three failures for me locally:

```
  1) JRuby class reification supports reification of java classes with interfaces
     Failure/Error: DEFAULT_FAILURE_NOTIFIER = lambda { |failure, _opts| raise failure }
     
       expected: [:success]
            got: [:failSO, :fail1, "fail3"]
     
       (compared using eql?)
     
       Diff:
       @@ -1 +1 @@
       -[:success]
       +[:failSO, :fail1, "fail3"]

  2) JRuby class reification supports reification of ruby classes with interfaces
     Failure/Error: DEFAULT_FAILURE_NOTIFIER = lambda { |failure, _opts| raise failure }
     
       expected: [:success]
            got: [:failSO, :fail1, "fail3"]
     
       (compared using eql?)
     
       Diff:
       @@ -1 +1 @@
       -[:success]
       +[:failSO, :fail1, "fail3"]

  3) A Ruby subclass of a Java concrete class with custom methods generates new methods usable by ruby & java
     Failure/Error: DEFAULT_FAILURE_NOTIFIER = lambda { |failure, _opts| raise failure }
     
       expected: 1
            got: 2
     
       (compared using ==)
```

This is MacOS, Java 11.

byteit101:
What remains to be done?

 * Fix the jenkins tests
   * Random openj9 hang in tests
   * File not found when it exists in :aot tests
 * Review the TODO's scattered around the code to determine if they are important things or not
 * Get @kares and @enebo 's views on the ruby api, so I can doc it (https://github.com/jruby/jruby/pull/6422#issuecomment-775394796)
 * Review the code

PR #6423, opened 2020-10-03
update jruby-openssl to 0.10.5

kares:
expected to avoid left-over reflection warnings on Java 9+

https://github.com/jruby/jruby-openssl/releases/tag/v0.10.5


kares:
on hold due: https://github.com/elastic/logstash/issues/12299

kares:
elastic/logstash#12299 was a false alarm (build specific `bundle install` issue), 0.10.5 should be :green_circle:  to go!

headius:
Huzzah! I think this is the last reflective access warning we know of!

Issue #6385, opened 2020-09-03
DateTime loses timezone offsets on marshal/unmarshal


paulkerschen:
On JRuby 9.2, marshaling and unmarshaling a DateTime object with a nonzero timezone offset outputs an “invalid offset is ignored” warning, and returns a DateTime with the timezone offset set to zero.

I believe the marshal dump and load methods in org.jruby.ext.date.RubyDate may be representing nonzero timezone offsets in an incompatible way.

**Environment Information**
  
* JRuby 9.2.13.0
* Fresh install with no additional gems
* No additional flags or JRUBY_OPTS

**Expected Behavior**

On MRI 2.5.7, marshaling and unmarshaling a DateTime object preserves timezone offset without error. Minimal example in irb:

```
:001 > require 'date'
 => true 
 :002 > d = DateTime.now
 => #<DateTime: 2020-09-03T10:00:56-07:00 ((2459096j,61256s,323456000n),-25200s,2299161j)> 
 :003 > Marshal.load(Marshal.dump(d))
 => #<DateTime: 2020-09-03T10:00:56-07:00 ((2459096j,61256s,323456000n),-25200s,2299161j)>
```

**Actual Behavior**

Minimal irb example in JRuby 9.2.13.0. Note “invalid offset is ignored” warning and loss of timezone offset information after marshaling and unmarshaling.

```
jruby-9.2.13.0 :001 > require 'date'
 => true 
jruby-9.2.13.0 :002 > d = DateTime.now
 => #<DateTime: 2020-09-03T09:53:30-07:00 ((2459096j,60810s,196000000n),-25200s,2299161j)> 
jruby-9.2.13.0 :003 > Marshal.load(Marshal.dump(d))
invalid offset is ignored
 => #<DateTime: 2020-09-03T16:53:30+00:00 ((2459096j,60810s,196000000n),+0s,2299161j)> 
```

A DateTime object with zero offset does not trigger the warning.

```
jruby-9.2.13.0 :001 > require 'date'
 => true 
jruby-9.2.13.0 :002 > d = DateTime.now.new_offset(0)
 => #<DateTime: 2020-09-03T17:19:31+00:00 ((2459096j,62371s,904000000n),+0s,2299161j)> 
jruby-9.2.13.0 :003 > Marshal.load(Marshal.dump(d))
 => #<DateTime: 2020-09-03T17:19:31+00:00 ((2459096j,62371s,904000000n),+0s,2299161j)> 
```

[...]

PR #6424, opened 2020-10-04
[fix] Date/DateTime loosing offset on marshal

kares:
This is a regression since `Date`/`DateTime` where moved to native in 9.2.
In JRuby 9.1 we used to store the offset as it's returned by `Date#offset` (usually a `Rational`), 
however on the native side we now store `* DAYS_IN_SECONDS` an an integer.

The change here restores 9.1 behaviour.

resolves GH-6385


headius (approved):
I love a good one-liner. 👍

Issue #6019, opened 2020-01-07
Sporadic error in socket ruby spec on Travis


headius:
I restarted the build and this went away. Happened during spec:ruby:fast suite:

```
1)
Socket.udp_server_loop_on when no connections are available blocks the caller ERROR
NoMethodError: undefined method `each' for nil:NilClass
/home/travis/build/jruby/jruby/lib/ruby/stdlib/socket.rb:979:in `udp_server_recv'
/home/travis/build/jruby/jruby/lib/ruby/stdlib/socket.rb:1008:in `block in udp_server_loop_on'
org/jruby/RubyKernel.java:1442:in `loop'
/home/travis/build/jruby/jruby/lib/ruby/stdlib/socket.rb:1006:in `udp_server_loop_on'
/home/travis/build/jruby/jruby/spec/ruby/library/socket/socket/udp_server_loop_on_spec.rb:16:in `block in <main>'
```

headius:
This came up again today on a branch build: https://travis-ci.org/github/jruby/jruby/jobs/710485092

Digging into it, it appears that the select inside `Socket.udp_server_loop_on` is returning a `readable` array with a nil value, which should not happen (select should only return lists of ready streams, or nil).

This situation has come up before and it's a bug in `select`.

I will tag this spec for now.

headius:
May be fixed by https://github.com/jnr/jnr-enxio/commit/9037dc524b647e50cc47a2b99ca689e699e14fd5 in enxio 0.32.1.

headius:
I reverted this in 753ed1561f, fingers crossed that the enxio fix also fixed this.

headius:
The enxio fix does not appear to be sufficient, so I have restored the tag in 9cf5061196.

https://travis-ci.org/github/jruby/jruby/jobs/733407492

PR #6425, opened 2020-10-05
Update enxio to omit poll fd from key list

headius:
This long-standing bug allows a null key to get into the list of
keys when using the `poll` selector. One of the elements in the
internal array of keys is expected to be null to reserve a spot
for an additional poll file descriptor, but null elements should
never be returned in the keys list.


headius:
This update may fix https://github.com/jruby/jruby/issues/6019.

Issue #6421, opened 2020-09-30
Dir.glob with `classpath:` resource as a base is broken


Magisus:
**Environment Information**
  
Provide at least:
* JRuby version: 9.2.11+ (still broken on 9.2.13.0)
* MacOS 10.15, Centos 7 (3.10.0-862.3.2.el7.x86_64)

Other relevant info you may wish to add:
* Used within a clojure app
* issue can be reproduced running `irb` in a leiningen REPL
* Also reproducible inside a jar

**Expected Behavior**

`Dir.glob` should be able to take a `classpath:<filename>` resource as a base, since it apparently supports `uri:classloader:<filename>`.

**Actual Behavior**

We observed the following, starting a REPL by running `lein irb` in a checkout of a the puppetserver repo. Apologies that I'm not sure how to properly set up a classpath outside of this environment for a simpler repro:

```
# this is the correct behavior
irb(main):006:0> Dir.glob("*", base: "./src/ruby/puppetserver-lib")
=> ["puppet"]
 
# This started breaking with 9.2.11.0 (or possibly 9.2.10, but we skipped that version)
irb(main):009:0> Dir.glob "*", base:  "classpath:/puppetserver-lib"
Traceback (most recent call last):
16: from org.jruby.ir.interpreter.InterpreterEngine.processCall(InterpreterEngine.java:325)
15: from org.jruby.runtime.callsite.CachingCallSite.call(CachingCallSite.java:205)
14: from org.jruby.runtime.callsite.CachingCallSite.cacheAndCall(CachingCallSite.java:396)
13: from org.jruby.internal.runtime.methods.DynamicMethod.call(DynamicMethod.java:208)
12: from org.jruby.internal.runtime.methods.DynamicMethod.call(DynamicMethod.java:212)
11: from org.jruby.RubyDir$INVOKER$s$0$2$glob.call(RubyDir$INVOKER$s$0$2$glob.gen)
10: from org.jruby.RubyDir.glob(RubyDir.java:310)
9: from org.jruby.util.Dir.push_glob(Dir.java:310)
8: from org.jruby.util.Dir.push_braces(Dir.java:449)
7: from org.jruby.util.Dir.push_globs(Dir.java:481)
6: from org.jruby.util.Dir.glob_helper(Dir.java:693)
5: from org.jruby.util.Dir.glob_helper(Dir.java:745)
4: from org.jruby.util.JRubyFile.createResource(JRubyFile.java:82)
[...]

PR #6427, opened 2020-10-06
Use normalized URI for glob

headius:
The Dir.glob logic did not use the normalized URI form when calling glob internals, and since those internals do not support "classpath:" it ended up raising an IllegalArgumentException as reported in #6421.

This change should be benign, but my concern lies in having to call "getAbsolutePath" to get the normalized URI back.

This may be worth including in 9.2.14.


headius:
Clearly this needs another look.

PR #6429, opened 2020-10-07
Fixes #6428: Date._strptime raises Java exception on invalid argument

### LINKED ISSUE #6428: Date._strptime raises Java exception on invalid argument
... while it should properly raise a Ruby `TypeError`, such as with MRI:

`ruby -rtime -e "Time.strptime(0, '%Y-%m-%d')"`
```
time.rb:432:in `_strptime': no implicit conversion of Integer into String (TypeError)
```

`ruby -rtime -e "Time.strptime(nil, '%Y-%m-%d')"`
```
time.rb:432:in `_strptime': no implicit conversion of nil into String (TypeError)
```

JRuby (9.2.13.0) ends up with:

```
        4: from org.jruby.runtime.callsite.CachingCallSite.call(CachingCallSite.java:203)
        3: from org.jruby.ext.date.RubyDate$INVOKER$s$_strptime.call(RubyDate$INVOKER$s$_strptime.gen)
        2: from org.jruby.ext.date.RubyDate._strptime(RubyDate.java:1685)
        1: from org.jruby.ext.date.RubyDate.parse(RubyDate.java:1691)
Java::JavaLang::ClassCastException (org.jruby.RubyNil cannot be cast to org.jruby.RubyString)

```

MariuszCwikla:
Fixes #6428

### DIFF
--- a/core/src/main/java/org/jruby/ext/date/RubyDate.java
+++ b/core/src/main/java/org/jruby/ext/date/RubyDate.java
@@ -1685,7 +1685,8 @@ public class RubyDate extends RubyObject {
 
     @JRubyMethod(meta = true)
     public static IRubyObject _strptime(ThreadContext context, IRubyObject self, IRubyObject string, IRubyObject format) {
-        format = TypeConverter.checkStringType(context.runtime, format);
+        string = string.convertToString();
+        format = format.convertToString();
         return parse(context, string, ((RubyString) format).decodeString());
     }
 
--- a/test/jruby/test_time.rb
+++ b/test/jruby/test_time.rb
@@ -213,4 +213,10 @@ class TestTimeNilOps < Test::Unit::TestCase
     end
   end
 
+  def test_strptime_type_error
+    assert_raise(TypeError) { Time.strptime(0, '%Y-%m-%d') }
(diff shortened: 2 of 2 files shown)


test/mri/test_time.rb
```diff
@@ -454,6 +454,13 @@ def test_strptime_empty
     assert_raise(ArgumentError) { Time.strptime('+09:00', '%z') } # [ruby-core:62351] by matz.
   end
 
+  def test_strptime_type_error
```
kares, line 457: it's a bit confusing but the MRI suite is owner and imported from Ruby itself.
thus these could get lost easily on the next import, preferably use Ruby Specs under *spec/ruby* (maybe there is one - it's just tagged to be skipped) or simply move this to *test/jruby/test_time.rb*

kares (changes requested):
:+1: except we should move tests

test/mri/test_time.rb
MariuszCwikla: Sure, will do. Thanks

MariuszCwikla: Done, moved to `test/jruby/test_time.rb`

PR #6431, opened 2020-10-08
Fix platform type inititilization

MariuszCwikla:
Accidentally I found a bug in `Platform.initPlatform()`.  On Windows it returns `UnixPlatform` even though it has following logic:

```
    private static Platform initPlatform(){
        try {
            if (IS_WINDOWS)
                return new NTPlatform();
```

It turns out that ordering of static variables is matters.
When debugging it turns out that `Platform.OS` is null and IS_WINDOWS is not correctly computed.

Simplest solution is to just move `INSTANCE` below `OS` and `IS_WINDOWS`.
Tested on Oracle JDK 8 and Open JDK 11.

### DIFF
--- a/core/src/main/java/org/jruby/platform/Platform.java
+++ b/core/src/main/java/org/jruby/platform/Platform.java
@@ -40,7 +40,6 @@ import org.jruby.runtime.Helpers;
  * Platform specific constants.
  */
 public abstract class Platform {
-    private static final Platform INSTANCE = initPlatform();
     public static Platform getPlatform() {
         return INSTANCE;
     }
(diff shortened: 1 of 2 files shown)


headius:
Wow, this is a great find! I am amazed it did not break anything in the handful of Windows CI tests we run.

headius:
@enebo @kares Perhaps this should go into 9.2.14 as well? I guess we do not know of any bugs caused by it, so I'm not sure how to weigh the small risk involved.

kares:
@headius this is not risky, .java parts only use the `Platform.*` constants - there's only [one place](https://github.com/jruby/jruby/blob/9.2.13.0/core/src/main/java/org/jruby/RubyProcess.java#L608) where the instance is retrieved -> `Process.groups` will be broken on Windows (and Solaris) but it's not-implemented in Windows anyways.

headius:
@kares Yeah at a glance the exposure looked pretty small, and now it seems largely non-existent. Probably better to get it into 9.2 line just in case some future patch exposes this issue.

headius:
Cherrypicked to master in cce44feecc.

kares:
using `Process.groups` will likely go heavy on warnings (due relying on JDK internals). 
I'll take a stub at using POSIX-y groups instead.

PR #6438, opened 2020-10-15
Fixes #6404 jrubyc: mangled ("--target") path on Windows

### LINKED ISSUE #6404: jrubyc has issues with mangled (target) path on Windows
not 100% sure but seems like `jrubyc` is broken with `--target` path specified,
the path seems to end up being invalid on Windows, here's 2 (pended) reproducers: https://github.com/jruby/jruby/commit/851629a431b4c5b9fe63d78723cd32980f030c85

MariuszCwikla:
Fixes #6404

### DIFF
--- a/test/jruby/compiler/test_jrubyc.rb
+++ b/test/jruby/compiler/test_jrubyc.rb
@@ -3,6 +3,7 @@ require 'stringio'
 require 'tempfile'
 require 'fileutils'
 require 'test/jruby/test_helper'
+require 'pathname'
 
 class TestJRubyc < Test::Unit::TestCase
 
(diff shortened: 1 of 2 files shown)


kares:
Thanks Mariusz, :trophy: find!

kares (approved):

test/jruby/compiler/test_jrubyc.rb
```diff
@@ -15,8 +16,8 @@ def setup; require 'jruby/jrubyc'
 
     @tempfile_stderr = Tempfile.open("test_jrubyc_stderr")
     @old_stderr = $stderr.dup
-    $stderr.reopen @tempfile_stderr
-    $stderr.sync = true
+#    $stderr.reopen @tempfile_stderr
+#    $stderr.sync = true
```
headius, line 20: Perhaps we can just remove these lines?

headius (approved):
Looks good to me.

test/jruby/compiler/test_jrubyc.rb
headius: Actually I suspect these lines were commented out solely for debugging and should be restored. I'll take care of it.

PR #6439, opened 2020-10-15
[refactor] use posix for Process.groups

kares:
instead of relying on *com.sun.security.auth.module* internals

this change drops `org.jruby.platform.Platfom.getPlatform` and its (private) sub-classes,
`Process.groups` on Windows is still "not implemented" as before


core/src/main/java/org/jruby/RubyProcess.java
```diff
@@ -605,12 +605,16 @@ public static IRubyObject exit_bang(IRubyObject recv, IRubyObject[] args) {
 
     @JRubyMethod(name = "groups", module = true, visibility = PRIVATE)
     public static IRubyObject groups(IRubyObject recv) {
-        long[] groups = Platform.getPlatform().getGroups(recv);
-        RubyArray ary = RubyArray.newArray(recv.getRuntime(), groups.length);
+        final Ruby runtime = recv.getRuntime();
+        long[] groups = runtime.getPosix().getgroups();
+        if (groups == null) { // not-implemented for the given platform (e.g. Windows)
+            throw runtime.newNotImplementedError("groups() function is not implemented");
```
ahorek, line 611:
```suggestion
            throw runtime.newNotImplementedError("groups() function is unimplemented on this machine");
```
\* to match MRI's message

ahorek (approved):
nice cleanup

headius:
This cleanup seems fine, but it's not zero-sum. On systems where the user has chosen to disable native support, or if native support can't be loaded for some other reason (security etc) there's no fallback to the old logic.

I would suggest that perhaps the invasive reflection-based logic be added to JavaPosix in jnr-posix so we can still support this functionality without native support.

headius:
Status here @kares? Can we restore the old logic as a fallback and get this merged in for 9.3?

kares:
sorry haven't had the time to do this as the jnr-posix end, yet

headius:
@kares Now's the time to merge or punt to 9.4.1.0.

kares:
believe, it's a good idea to merge and not rely on JDK internals, despite not having the jnr-posix work in place.
the only regression I can see potentially happening from this would be `Process.groups` not working on Solaris (haven't checked if it's actually the case).

PR #6444, opened 2020-10-22
Fix date parse with week number

joshuacronemeyer:
I noticed that parsing a date string with a week number (https://en.wikipedia.org/wiki/ISO_week_date) in jruby will raise NoMethodError (undefined method `nonzero?' for nil:NilClass) I looked into it and have attempted a fix and included a couple unit tests.

### DIFF
--- a/lib/ruby/stdlib/date.rb
+++ b/lib/ruby/stdlib/date.rb
@@ -406,7 +406,7 @@ class Date
       return jd
     end
 
-    if cwyear = elem[:cwyear] and cweek = elem[:cweek] and cwday = (elem[:cwday] || elem[:wday].nonzero? || 7) and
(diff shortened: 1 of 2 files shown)


joshuacronemeyer:
Ok. Yay tests. I'll have another look.

joshuacronemeyer:
I was hoping I could fix this in ruby, but looks like the issue goes a layer deeper into RubyDate.java. I have a temporary fix in the ruby layer that i hope will demonstrate the idea of my fix. If all tests pass here I will see if I can fix RubyDate _parse_impl accordinly.

headius:
Sounds great, thank you for looking into this! If you have questions feel free to stop by our Matrix chat via https://www.jruby.org/chat.

headius:
@joshuacronemeyer This one has been idle for quite a while and I'm not sure where it stands, so we're going to de-target it and hopefully you can give us an update.

Issue #5896, opened 2019-09-29
Hash#each raises ArgumentError with a method of arity 2 converted to a block


mvz:
### Environment

Provide at least:
* JRuby version (`jruby -v`) and command line (flags, JRUBY_OPTS, etc)
```
jruby 9.2.6.0 (2.5.3) 2019-02-11 15ba00b OpenJDK 64-Bit Server VM 11.0.5-ea+6-post-Debian-2 on 11.0.5-ea+6-post-Debian-2 [linux-x86_64]
JRUBY_OPTS=--dev
JAVA_OPTS=" --add-opens java.base/java.io=ALL-UNNAMED --add-opens java.base/java.lang=ALL-UNNAMED --add-opens java.base/java.util=ALL-UNNAMED --add-opens java.base/java.lang.reflect=ALL-UNNAMED --add-opens java.base/java.security.cert=ALL-UNNAMED --add-opens java.base/java.security=ALL-UNNAMED --add-opens java.base/javax.crypto=ALL-UNNAMED --add-opens java.base/sun.nio.ch=ALL-UNNAMED --add-opens java.base/java.nio.channels.spi=ALL-UNNAMED --illegal-access=warn"
```
* Operating system and platform (e.g. `uname -a`)
```
Linux bean 5.2.0-2-amd64 #1 SMP Debian 5.2.9-2 (2019-08-21) x86_64 GNU/Linux
```

### Expected Behavior

Given the following Ruby script:
```ruby
class Bar
  def foo
    { a: 1, b: 2}.each(&method(:bar))
  end

  def bar(a, b)
    puts "#{a}: #{b}"
  end
end

Bar.new.foo
```

CRuby has the following output:
```
a: 1
b: 2
```

### Actual Behavior

JRuby has the following output:
```
ArgumentError: wrong number of arguments (given 1, expected 2)
     bar at foo.rb:3
    each at org/jruby/RubyHash.java:1419
     foo at foo.rb:3
  <main> at foo.rb:11
```

eregon:
Ref: https://github.com/ruby/spec/pull/811#issuecomment-717891062

headius:
As mentioned on #6451, we'll wait a bit and give @marshalium a chance to sort out the right fix and behavior here. The confusing combination is Hash#each plus a proc-ified Map of arity 2; CRuby actually passes two arguments, rather than a splattable Array, but we pass a splattable Array and since our Method is strict (perhaps correctly so) it rejects it as a single argument.

[...]

Issue #6165, opened 2020-04-02
Lambda fails to arity check incoming single-arg array in JIT


headius:
The arity-checking logic for lambdas differs between the unoptimized startup interpreter and the optimized full interpreter or JIT modes. As a result, the following code does not error when it should:

```ruby
[[1, 2, 3]].collect(&->(a, b, c) {[a, b, c]})
```

Along the optimized paths for this lambda block, the logic is splatting the incoming array into the arguments before checking arity. The simple interpreter just checks arity and properly errors.

I tried a simple fix for this by deleting the `toAry` logic in IRRuntimeHelpers.prepareBlockArgs, but it led to the following error:

```
Error: test_zip(TestEnumerator): ArgumentError: wrong number of arguments (given 1, expected 2)
/home/travis/build/jruby/jruby/test/jruby/test_enumerator.rb:-1:in `block in test_zip'
/home/travis/build/jruby/jruby/test/jruby/test_enumerator.rb:253:in `each'
org/jruby/RubyEnumerable.java:1275:in `each_with_index'
org/jruby/RubyEnumerator.java:396:in `each'
org/jruby/RubyEnumerable.java:1936:in `zip'
/home/travis/build/jruby/jruby/test/jruby/test_enumerator.rb:272:in `test_zip'
     269: 
     270:     a = []
     271:     cond = ->((x, i), y) { a << [x, y, i] }
  => 272:     @obj.each_with_index.zip([:a, :b, :c], &cond)
     273:     assert_equal([[1,:a,0],[2,:b,1],[3,:c,2],[1,nil,3],[2,nil,4]], a)
     274: 
     275:     a = []
```

This is not a new issue but it came up due to a new test in CRuby 2.6's test suite. I will tag it for now to get the 2.6 branch merged, but we need to unify how all these dispatch paths work.

headius:
Note that the simple interpreter uses RubyProc.prepareArgs to set up the argument list, which does no splatting before lambda arity check. However slightly different logic is used along different yield paths, and the second example above (which fails after my patch) does things differently.

headius:
The bad fix also led to dozens of failures in enumerator logic, of the following forms:

```
13) Error:
TestEnumerable#test_max_by:
ArgumentError: wrong number of arguments (given 1, expected 2)
    /home/travis/build/jruby/jruby/test/mri/ruby/test_enum.rb:13:in `block in test_max_by'
[...]

PR #6451, opened 2020-10-27
Fix yielding to methods that take two params from Hash#each

marshalium:
This fixes issue #5896 by always splatting out the arguments if possible.

I was surprised that `Helpers.restructureBlockArgs()` is only used here. Not sure why this is a special case.

Overall I'm not 100% sure about this fix. But since I couldn't get the tests to all run locally I'm opening up this PR in hopes that CI can reveal issues or someone here can tell me what I'm missing.

### DIFF
--- a/core/src/main/java/org/jruby/RubyHash.java
+++ b/core/src/main/java/org/jruby/RubyHash.java
@@ -1513,7 +1513,11 @@ public class RubyHash extends RubyObject implements Map {
     private static final VisitorWithState<Block> YieldKeyValueArrayVisitor = new VisitorWithState<Block>() {
         @Override
         public void visit(ThreadContext context, RubyHash self, IRubyObject key, IRubyObject value, int index, Block block) {
-            block.yield(context, context.runtime.newArray(key, value));
+            if (block.getSignature().arityValue() > 1) {
+                block.yieldSpecific(context, key, value);
+            } else {
+                block.yield(context, context.runtime.newArray(key, value));
+            }
         }
     };
 
--- a/core/src/main/java/org/jruby/runtime/Helpers.java
+++ b/core/src/main/java/org/jruby/runtime/Helpers.java
@@ -2439,16 +2439,11 @@ public class Helpers {
     // . Array with multiple values and NO rest should extract args if there are more than one argument
 
     static IRubyObject[] restructureBlockArgs(ThreadContext context,
-        IRubyObject value, Signature signature, Block.Type type, boolean needsSplat) {
+        IRubyObject value, Signature signature, Block.Type type) {
 
         if (!type.checkArity && signature == Signature.NO_ARGUMENTS) return IRubyObject.NULL_ARRAY;
 
         if (value == null) return IRubyObject.NULL_ARRAY;
-
-        if (needsSplat) {
-            IRubyObject ary = Helpers.aryToAry(context, value);
-            if (ary instanceof RubyArray) return ((RubyArray) ary).toJavaArrayMaybeUnsafe();
-        }
 
         return new IRubyObject[] { value };
     }
--- a/core/src/main/java/org/jruby/runtime/MethodBlockBody.java
+++ b/core/src/main/java/org/jruby/runtime/MethodBlockBody.java
@@ -64,7 +64,7 @@ public class MethodBlockBody extends ContextAwareBlockBody {
 
     @Override
     protected IRubyObject doYield(ThreadContext context, Block block, IRubyObject value) {
-        IRubyObject[] realArgs = Helpers.restructureBlockArgs(context, value, getSignature(), block.type, false);
+        IRubyObject[] realArgs = Helpers.restructureBlockArgs(context, value, getSignature(), block.type);
         return entry.method.call(context, receiver, entry.sourceModule, originName, realArgs, Block.NULL_BLOCK);
     }
 
--- a/spec/tags/ruby/core/enumerable/collect_tags.txt
+++ b/spec/tags/ruby/core/enumerable/collect_tags.txt
@@ -1,1 +1,0 @@
-fails:Enumerable#collect yields 2 arguments for a Hash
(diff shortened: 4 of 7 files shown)


marshalium:
For sure! I've opened a PR to merge a spec that tests the same thing into ruby/spec: https://github.com/ruby/spec/pull/811

What should I do with this PR while I wait for the ruby/spec one to get approved? Should I remove the test from this commit or will we merge this as is and remove the duplication later when JRuby syncs up with ruby/spec?

eregon:
See https://github.com/ruby/spec/pull/811#issuecomment-717891062, I believe Method#to_proc is always strict and does not auto-splat. It seems specific behavior of Hash#{each/each_pair/map} rather, originating from an optimization not intentionally changing semantics. Note in Ruby 3 Hash#{each/each_pair} don't exhibit this strange behavior anymore

marshalium:
Thanks @eregon! You're correct. I will try to come up with a better fix

headius:
@marshalium I will wait to hear back from you about the correct behavior and fix for this.

These special splatting behaviors have always been a mystery, and have gone back and forth over the years. In this case I'm getting that it's the combination of a proc-ified Method fed to Hash#each and friends resulting in unexpected behavior... where we are strictly enforcing the arity and not splatting the incoming argument array, CRuby passes two arguments explicitly rather than two arguments as an argument array.

Needless to say these exceptions are confusing and frustrating, and I'm glad this will be more normalized for 3.0.

headius:
@marshalium My five minute patch may be along the right lines:

```
diff --git a/core/src/main/java/org/jruby/RubyHash.java b/core/src/main/java/org/jruby/RubyHash.java
index 40f7170f1b..59b6f4ec55 100644
--- a/core/src/main/java/org/jruby/RubyHash.java
+++ b/core/src/main/java/org/jruby/RubyHash.java
@@ -1513,7 +1513,11 @@ public class RubyHash extends RubyObject implements Map {
     private static final VisitorWithState<Block> YieldKeyValueArrayVisitor = new VisitorWithState<Block>() {
         @Override
         public void visit(ThreadContext context, RubyHash self, IRubyObject key, IRubyObject value, int index, Block block) {
-            block.yield(context, context.runtime.newArray(key, value));
+            if (block.getSignature().arityValue() == 2) {
+                block.yieldSpecific(context, key, value);
+            } else {
+                block.yield(context, context.runtime.newArray(key, value));
+            }
         }
     };
 
```

I have not confirmed whether this breaks anything but from discussions it sounds like this is close to the proper behavior.

marshalium:
Thx! I kept the removal of the unnecessary `needsSplat` argument and have pushed up the arity check so we can see what CI says. It seems right because Hash is a special case and so should be doing that check explicitly but it still feels like I'm not understanding something about how the JRuby code works.

marshalium:
Ah, here's the thing I don't understand. And it looks like the opposite bug where lambdas are too lenient. And it's why this looked like a Method#to_proc issue and not a general arity enforcing issue.

On Ruby 2.5.8 you don't get auto-splatting when yielding two args to a lambda outside of special cases like Hash#each:
```
marshall@lappy7 jruby % RBENV_VERSION=2.5.8 irb
irb(main):001:0> RUBY_DESCRIPTION
=> "ruby 2.5.8p224 (2020-03-31 revision 67882) [x86_64-darwin19]"
irb(main):002:0> m = lambda { |x, y| [x, y] }
=> #<Proc:0x00007fe8d98917b8@(irb):2 (lambda)>
irb(main):003:0> [[1,2],[3,4]].each(&m)
        4: from /Users/marshall/.rbenv/versions/2.5.8/bin/irb:11:in `<main>'
        3: from (irb):3
        2: from (irb):3:in `each'
        1: from (irb):2:in `block in irb_binding'
ArgumentError (wrong number of arguments (given 1, expected 2))
```

In JRuby it works without error:
```
marshall@lappy7 jruby % RBENV_VERSION=jruby-9.2.13.0 irb
irb(main):001:0> RUBY_DESCRIPTION
=> "jruby 9.2.13.0 (2.5.7) 2020-08-03 9a89c94bcc OpenJDK 64-Bit Server VM 11.0.8+11 on 11.0.8+11 +jit [darwin-x86_64]"
irb(main):002:0>  m = lambda { |x, y| [x, y] }
=> #<Proc:0x16c3ca31@(irb):2 (lambda)>
irb(main):003:0> [[1,2],[3,4]].each(&m)
=> [[1, 2], [3, 4]]
```

And what's weird is that calling the lambda directly doesn't have this magic behavior:
```
marshall@lappy7 jruby % RBENV_VERSION=jruby-9.2.13.0 irb
irb(main):001:0> RUBY_DESCRIPTION
=> "jruby 9.2.13.0 (2.5.7) 2020-08-03 9a89c94bcc OpenJDK 64-Bit Server VM 11.0.8+11 on 11.0.8+11 +jit [darwin-x86_64]"
irb(main):002:0> m = lambda { |x, y| [x, y] }
=> #<Proc:0x53dfacba@(irb):2 (lambda)>
irb(main):003:0> m.call(1, 2)
=> [1, 2]
irb(main):004:0> m.call([1,2])
        6: from /Users/marshall/.rbenv/versions/jruby-9.2.13.0/bin/irb:13:in `<main>'
        5: from org/jruby/RubyKernel.java:1189:in `catch'
        4: from org/jruby/RubyKernel.java:1189:in `catch'
        3: from org/jruby/RubyKernel.java:1442:in `loop'
        2: from org/jruby/RubyKernel.java:1048:in `eval'
        1: from (irb):4:in `evaluate'
ArgumentError (wrong number of arguments (given 1, expected 2))
```

So this makes me more confident that the arity checking was just always missing and it wasn't very noticeable because this other lambda arity checking bug is here.

headius:
The behavior of a lambda fed to `each` is likely a bug in how we unwrap the lambda to use as a block. It is losing its "lambda-ness" and being treated as a normal `proc`. You are probably right that it is the same issue as #6165. Don't worry about that case here, since I think it is a different cause and already a known issue.

headius:
@marshalium The only failure in your latest push appears to be a JDK-downloading glitch which I have restarted. I am glad to see so many tags go away! We will review and get this merged soon.

headius (approved):
Looks good to me. The only real change is the separate yield in Hash#each, which matches CRuby logic. The other changes just remove dead code.

Issue #5876, opened 2019-09-13
JSR 223 different behaviour between compiled execution and evaluated


andsel:
### Environment

Provide at least:
* JRuby version (`jruby -v`) and command line (flags, JRUBY_OPTS, etc)
jruby-9.2.8.0
* Operating system and platform (e.g. `uname -a`)
Linux 5.0.0-27-generic #28-Ubuntu SMP Tue Aug 20 19:53:07 UTC 2019 x86_64 x86_64 
x86_64 GNU/Linux
JVM: OpenJDK 12.0.2

Other relevant info you may wish to add:
* Installed or activated gems
* Application/framework version (e.g. Rails, Sinatra)
* Environment variables

### Expected Behavior
Both `eval` invoked on `ScriptEngine` and on `CompiledScript` prints same value of local variable

* Describe your expectation of how JRuby should behave, perhaps by showing how CRuby/MRI behaves.
Both ways of execution should output the same string

* Provide an executable Ruby script or a link to an example repository.
NB put the tests in two separate unit test _or_ run one at a time
```java
@Test
public void testEvaluateCompiled() throws ScriptException {
    ScriptEngineManager manager = new ScriptEngineManager();
    ScriptEngine engine = manager.getEngineByName("jruby");

    Bindings bindings = new SimpleBindings();
    bindings.put("message", "my variable");

    CompiledScript compiled = ((Compilable) engine).compile("puts \">>>>>>\" + message");
    compiled.eval(bindings);
}

@Test
public void testEvaluateInterpreted() throws ScriptException {
    ScriptEngineManager manager = new ScriptEngineManager();
    ScriptEngine engine = manager.getEngineByName("jruby");

    Bindings bindings = new SimpleBindings();
    bindings.put("message", "my variable");

    engine.eval("puts \">>>>>>\" + message", bindings);
}
```

### Actual Behavior

* Describe or show the actual behavior.
the test `testEvaluateCompiled` fails with error while the other print correctly the string `>>>>>>global variable`

* Provide text or screen capture showing the behavior.
```
WARNING: An illegal reflective access operation has occurred
WARNING: Illegal reflective access by com.headius.backport9.modules.Modules (file:/home/andrea/.gradle/caches/modules-2/files-2.1/org.jruby/jruby-complete/9.2.8.0/2dbdff9d9f4142e4a0bc2b743ded9487a80459ef/jruby-complete-9.2.8.0.jar) to field java.io.FileDescriptor.fd
[...]

PR #6457, opened 2020-11-08
[embed] refactor (JSR223) exception wrapping

kares:
seems it's been a while since this code was visited, there's some cleanup/refactoring along the line.

the motivation is trying to use `javax.script` APIs and finding things a bit cumbersome, namely:
- Ruby raises get double packed into an "embed failed exception" which than is wrapped into a `javax.script.ScriptException` (when using `ScriptEngine#invokeMethod`),
  believe this is just ineffective - e.g. Nashorn actually [exports](https://docs.oracle.com/javase/8/docs/jdk/api/nashorn/jdk/nashorn/api/scripting/NashornException.html) it's specific exceptions and doesn't always wrap.
 JRuby, since 9.2, has the Java chain for Ruby exceptions - which is very useful with JSR223 (unless it's burried to causes down the line), JRuby should really consider doing a breaking change here ... as proposed.

  also, any throwable gets wrapped (into a `InvokeFailedException`) even `java.lang.Error`s, which feels weird.
- null returns around JRuby's JSR-223 script engine impl, e.g. it was legit to `callMethod("", foo)`with an empty name that pretty much doesn't actually invoke a Ruby method - the impl simply returned null (and this was actually tested :gun:).

- (silent) exception printout(s) and runtime std-out printing ruby exceptions by in embed mode

**CHANGE SUMMARY:**
- javax.script embed APIs will no longer wrap with intermediate `InvokeFailedException`, `RaiseException`s will simply by wrapped one level into a (checked) `javax.script.ScriptException`
- `ScriptingContainer` continues to use the embed exception wrapping by default, but there's a way to disable wrapping using the added constructor
- "breaking change": only `java.lang.Exception` types will be wrapped - never `java.lang.Error` regardless of whether it's coming from a Ruby raise (no way to detect)

~~Potential Open Questions~~: 
- ~~do we want to keep `org.jruby.embed.InvokeFailedException` **deprecated** given ScriptingContainer will still wrap into those by default~~
- ~~maybe a better `ScripingContainer` API is needed e.g. factory methods where we always turn off wrapping by default (esp. if we keep `InvokeFailedException` deprecated)~~


headius:
Pre-review comments...

> do we want to keep org.jruby.embed.InvokeFailedException deprecated

Since this is still used by default in ScriptingContainer invokes I think we probably can't deprecate it unless we also deprecate the invoke paths that use it. This leads to the second question...

> maybe a better ScripingContainer API is needed

I am less excited about adding to the factory API than I would be about deprecating current invoke/eval methods in favor of new ones. Those eval methods have not been touched in over a decade and there are better forms of invoke and eval we could add as a new API. That would allow us to deprecate the old paths and provide a migration path away from old patterns.

For purposes of this PR, which mainly seeks to reduce the wrapping happening inside the javax.script API, I would say let's **not** deprecate InvokeFailedException until we can work on making a new ScriptingContainer API that doesn't use it at all.

kares:
still find some (untested) cases related to exceptions - there is proper `javax.script.ScriptException` wrapping around `eval`
but not if a script got "pre-compiled" using `((Compilable) engine).compile("...");` (e.g. in case of a syntax error the non-wrapped exception was thrown previously), should be fixed and properly tested ...

kares:
:green_circle: after merge, note that the PR does NOT resolve https://github.com/jruby/jruby/issues/5876 but we at least explicitly test existing behaviour with compiled scripts.

headius:
This can be merged any time and I will look into enhancing the 223 exception messages with file and line info once it has landed (#6560).

Issue #6458, opened 2020-11-10
bundle install on Windows fails with NotImplementedError: stat.st_gid unsupported or native support failed to load


eregon:
Happens on jruby-head with Bundler 2.
Maybe a recent change caused this? I never saw it before.

https://github.com/ruby/setup-ruby/runs/1378857770?check_suite_focus=true
https://github.com/ruby/setup-ruby/runs/1378857565?check_suite_focus=true
```
  C:\windows\system32\cmd.exe /D /S /C "D:\jruby-head\bin\bundle.bat config --local path vendor/bundle"
  C:\windows\system32\cmd.exe /D /S /C "D:\jruby-head\bin\bundle.bat lock"
  Fetching gem metadata from https://rubygems.org/.
  Resolving dependencies...
  Writing lockfile to D:/a/setup-ruby/setup-ruby/Gemfile.lock
  Cache key: setup-ruby-bundler-cache-v2-windows-latest-jruby-head-Gemfile.lock-095290c5236b11deb5dd04d5add69dccaac6031c6c7519ebe74c6a916382b678
  C:\windows\system32\cmd.exe /D /S /C "D:\jruby-head\bin\bundle.bat install --jobs 4"
```  
  ## Backtrace
  
  ```
  NotImplementedError: stat.st_gid unsupported or native support failed to load; see https://github.com/jruby/jruby/wiki/Native-Libraries
    org/jruby/RubyFileStat.java:301:in `gid'
    D:/jruby-head/lib/ruby/gems/shared/gems/bundler-2.1.4/lib/bundler/vendor/fileutils/lib/fileutils.rb:1433:in `copy_metadata'
    D:/jruby-head/lib/ruby/gems/shared/gems/bundler-2.1.4/lib/bundler/vendor/fileutils/lib/fileutils.rb:500:in `block in copy_entry'
    D:/jruby-head/lib/ruby/gems/shared/gems/bundler-2.1.4/lib/bundler/vendor/fileutils/lib/fileutils.rb:1518:in `wrap_traverse'
    D:/jruby-head/lib/ruby/gems/shared/gems/bundler-2.1.4/lib/bundler/vendor/fileutils/lib/fileutils.rb:494:in `copy_entry'
    D:/jruby-head/lib/ruby/gems/shared/gems/bundler-2.1.4/lib/bundler/vendor/fileutils/lib/fileutils.rb:551:in `block in mv'
[...]

PR #6460, opened 2020-11-10
[windows] fix gid

ahorek:
https://github.com/jruby/jruby/issues/6458

https://github.com/jnr/jnr-posix/commit/c5a412a8c80502f99f2baabf784ac4fa9a43aa36


eregon:
Thanks for the PR.
Looks good to me, assuming there is no sensible way to get those on Windows.
MRI seems to do the same: https://github.com/eregon/actions-shell/runs/1381456341?check_suite_focus=true#step:4:649

Could you merge? :)

headius (approved):

headius:
@ahorek Thanks fo the fix!

PR #6462, opened 2020-11-11
Update webrick to 1.6.1

joshbressers:
Upgrade webrick to 1.6.1 which fixes CVE-2020-25613


headius:
Thanks! This should also be applied to the jruby-9.2 branch but I can take care of that.

joshbressers:
Awesome, thanks!

Issue #6466, opened 2020-11-16
java.lang.ArrayIndexOutOfBoundsException in 9.2.13.0


AlexWayfer:
**Environment Information**
  
Provide at least:
* Cirrus CI `jruby 9.2.13.0 (2.5.7) 2020-08-03 9a89c94bcc OpenJDK 64-Bit Server VM 25.275-b01 on 1.8.0_275-b01 +jit [linux-x86_64]`
* Arch Linux (`5.9.8-arch1-1`) `jruby 9.2.13.0 (2.5.7) 2020-08-03 9a89c94bcc OpenJDK 64-Bit Server VM 14.0.2+12 on 14.0.2+12 +jit [linux-x86_64]`

Other relevant info you may wish to add:
* It's in [`filewatcher-cli`](https://github.com/filewatcher/filewatcher-cli) project.

**Expected Behavior**

No errors.

**Actual Behavior**

In Cirrus CI:

https://cirrus-ci.com/task/5614335160483840?command=test#L63

```
warning: thread "Ruby-0-Thread-2: /tmp/cirrus-ci-build/spec/filewatcher/cli/spec_helper/shell_watch_run.rb:68" terminated with exception (report_on_exception is true):
java.lang.ArrayIndexOutOfBoundsException: -1
	at org.jruby.runtime.ThreadContext.getCurrentFrame(ThreadContext.java:541)
	at org.jruby.runtime.ThreadContext.getFrameName(ThreadContext.java:675)
	at org.jruby.exceptions.RaiseException.doCallEventHook(RaiseException.java:171)
	at org.jruby.exceptions.RaiseException.preRaise(RaiseException.java:111)
	at org.jruby.exceptions.RaiseException.<init>(RaiseException.java:65)
	at org.jruby.exceptions.Exception.<init>(Exception.java:38)
	at org.jruby.exceptions.StandardError.<init>(StandardError.java:38)
	at org.jruby.exceptions.SystemCallError.<init>(SystemCallError.java:38)
	at org.jruby.RubySystemCallError.constructThrowable(RubySystemCallError.java:133)
	at org.jruby.RubyException.toThrowable(RubyException.java:349)
	at org.jruby.exceptions.RaiseException.from(RaiseException.java:81)
	at org.jruby.Ruby.newRaiseException(Ruby.java:4125)
	at org.jruby.Ruby.newErrnoFromInt(Ruby.java:3746)
[...]

PR #6467, opened 2020-11-16
Simplify detached process thread logic

headius:
In #6466 we see that the thread from a Process::detach is raising
an error when the waitpid fails with an error. This does not
appear to match the logic in MRI and also breaks if there is a
"raise" hook installed in the system.

This change should better match MRI logic for the detach thread
and avoid the hook issue reported in #6466. It does not address a
separate bug in the trace hook logic when no Ruby frames are
available.


headius:
This PR went against master but the same commit is in 9.2 at 4acd76ae110ec6f1adccc0c0b88faebb9465b36c.

Issue #6471, opened 2020-11-19
Bundled did_you_mean gem v1.2.0 has a CC NonCommercial license in included file


colinsurprenant:
[Bundled gem did_you_mean v1.2.0](https://github.com/jruby/jruby/blob/47d0c634f6289bf469b9c7d691abe969d96f5390/lib/pom.rb#L49) has a Creative Commons Attribution-NonCommercial-ShareAlike 3.0 license [on an included file](https://github.com/ruby/did_you_mean/blob/v1.2.0/evaluation/incorrect_words.yaml).

Per https://github.com/ruby/did_you_mean/issues/105 starting in v1.2.1 the whole `evaluation/` is now excluded from the gem which now ships without non-production code that has a non-commercial license.

I am leaving the trivial code-change to the team as I am unsure if a version more recent than v1.2.1 would be preferred here.

headius:
I don't see any reason we can't upgrade this in 9.3, and possibly in 9.2.14 if the gem is still compatible with Ruby 2.5.

@enebo Any concerns about updating did_you_mean for 9.2.14?

headius:
The latest is 1.4.0. We could either go to that and be current or just bump to 1.2.1 to get the CC stuff out of there. Bumping to 1.2.1 would clearly be the least impact.

colinsurprenant:
Perfect, thanks @headius!

headius:
Fixed by #6473.

PR #6473, opened 2020-11-20
Update did_you_mean

headius:
This updates the did_you_mean gem to version 1.2.1 to remove the CC BY-NC-SA yaml content reported in #6471.

The latest version is 1.4.0 but we have not decided whether to make that leap in a minor release of 9.2.


headius:
The CI failures appear to be Travis env changes and are unrelated to this PR.

PR #6482, opened 2020-12-02
Update JNR projects to latest

headius:
This picks up changes to the automatic module name from
jnr/jffi#67 but also updates JNR projects across the board to grab
all recent bug fixes.

### DIFF
--- a/core/pom.rb
+++ b/core/pom.rb
@@ -40,12 +40,12 @@ project 'JRuby Core' do
   jar 'org.ow2.asm:asm-util:${asm.version}'
 
   # exclude jnr-ffi to avoid problems with shading and relocation of the asm packages
-  jar 'com.github.jnr:jnr-netdb:1.1.6', :exclusions => ['com.github.jnr:jnr-ffi']
-  jar 'com.github.jnr:jnr-enxio:0.25', :exclusions => ['com.github.jnr:jnr-ffi']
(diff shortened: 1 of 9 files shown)


headius:
@enebo This is a larger change than the master update in #6481 but I think it should go into 9.2.14 to pick up various fixes to the JNR projects that have only been applied to master so far.

ahorek:
https://github.com/jruby/jruby/pull/6460 should be backported with this change

headius:
> #6460 should be backported with this change

Good call, I can take care of it.

headius:
@ahorek This PR appears to be hanging in a multicast socket test. Maybe you remember another fix we did on master that needs to be backported?

headius:
Aha... perhaps the constant fixes here: https://github.com/jruby/jruby/pull/6388

headius:
The fixes in #6388 do not appear to address the hanging test.

headius:
Ok possibly false negative since this test appears to hang on macOS even on master. I have pushed cherry-picks of all the changes that seemed relevant.

headius:
This is ready to merge, @enebo. The hangs were apparently related to the updated socket constants that @ahorek and I applied to master; when jnr-constants was updated those same changes needed to be applied to the 9.2 branch. The risk here seems fairly low but I would like a review before merging.

ahorek:
could you release https://github.com/jnr/jnr-constants/commits/master ? it should be safe.

I think that's all. +1

headius:
@ahorek Ahh it has been a while for that one. I will release it and the projects that use it.

Issue #6479, opened 2020-11-29
Java Interface Default methods not properly supported


byteit101:
**Definition**
```rb
class Bad
	include java.util.Iterator
end
class Iter
	include java.util.Iterator
	def hasNext
		false
	end
	def next
		nil
	end
	def remove
		super
	end
end
```

**Expected Behavior**
```rb
Bad.new.remove # => Java::JavaLang::UnsupportedOperationException (remove)
Bad.new.forEachRemaining {|x| p x} #=> NoMethodError (undefined method `hasNext' for ...)
Iter.new.remove # => Java::JavaLang::UnsupportedOperationException (remove)
Iter.new.forEachRemaining {|x| p x} #=> nil
```

**Actual Behavior**
```rb
Bad.new.remove # => NoMethodError (undefined method `remove' for ...)
Bad.new.forEachRemaining {|x| p x} #=> NoMethodError (undefined method `forEachRemaining' for ...)
Iter.new.remove # => NoMethodError (super: no superclass method `remove' for ...)
Iter.new.forEachRemaining {|x| p x} #=> NoMethodError (undefined method `forEachRemaining' for ...)
```

headius:
So it is the `super` that is the problem yeah? Hmmm... if we are properly binding the default methods on the interface module, I would expect this to work.

byteit101:
It's what I first noticed, but the other bit is that `Bad.new.remove` is a no method error, until you (today's workaround) call `to_java`

headius:
Ok if `Bad.new` doesn't work there may be a larger issue. This definitely needs to be fixed for 9.3.

byteit101:
To be clear on the to_java, it's per instance, not per-class:
```rb
class A; include java.util.Iterator; end.new.remove # error, NoMethodError
A.new.remove # error, NoMethodError
class B; include java.util.Iterator; end.new.to_java.remove # works, Java::JavaLang::UnsupportedOperationException
B.new.remove # error, NoMethodError
```

kares:
historically JRuby never bound abstract methods and thus skipped interfaces completely - believe this is the case here:
```
java.util.Iterator.methods(false)
> [:impl, :[], :java_method, :java_send, :append_features, :java_class, :extended]
```
... should not be hard to fix, let's see if we hit some old stuff which I am forgetting.

PR #6484, opened 2020-12-04
[ji] bind interface default methods

kares:
JRuby simply ignored any instance methods on interfaces, however this is incorrect due default methods.

resolves https://github.com/jruby/jruby/issues/6479


headius (approved):
Nice, not much of a change required!

headius:
Looks great to me, nice simple change. Merge as you see fit.

kares:
realized we need another piece of specs here to test interface method invocation with a Ruby target.

Issue #6452, opened 2020-10-27
super with kwargs splat causes HashMap to convert to Hash


headius:
The kwargs handling for the following code appears to be too aggressive, and results in the HashMap getting converted into a normal Ruby Hash:

```ruby
h = java.util.HashMap.new
class X
  def blah(*args, **kwargs)
    args
  end
end
class Y < X
  def blah(*args)
    super
  end
end
arr = Y.new.blah(h)
p arr[0].class # Hash, should be java.util.HashMap
```

This is the cause of recent CI failures in `spec:ji`, due to the recent rspec-expectations 3.9.3 release. The specific change was https://github.com/rspec/rspec-expectations/pull/1187/commits/23c1de7c94908b759d85cecf91f307b81f3aa422 as part of https://github.com/rspec/rspec-expectations/pull/1187 which appears to have been made to better support keyword arguments in the `method_missing` based matchers (like `be_equal?`, which is what started failing in our JI specs).

I will pin our Gemfile to rspec-expectations 3.9.2 to get the build green again, but this needs to be addressed for 9.3 at least and possibly 9.2.14 as well.

headius:
The example given might be reducible, but some form of super call appears necessary. If the kwargs method is called directly, rather than through super, it does not end up doing the conversion. So this seems to be a bug specific to super + kwargs splat in target args.

kares:
woow, maybe another good reason to consider NOT treating Java Maps "special" by being wrapped in a Hash-like proxy.
changing that will be breaking behavior though ... but maybe acceptable for 9.3?

headius:
It would be pretty big breaking behavior. I don't think it is really a bad thing that we have all the Hash methods on HashMap but the fact that it flips to Hash so easily is a bit of an issue. Should `to_h` not exist on HashMap? That seems too extreme, but is modifying it to return self (a HashMap) better? Is it reasonable to have a `to_h` that doesn't produce a natural Hash?

headius:
FWIW I **think** this issue means we are trying **too hard** to coerce arguments into kwargs within this code path, but it I have not confirmed whether CRuby also tries to coerce here.

headius:
[...]

Issue #3790, opened 2016-04-09
[ji] java.util.Collection should not be to_ary convertible


kares:
currently only Java arrays provide `to_ary` ... it _maybe_ makes sense for `java.util.List`
... but not for generic collection instances such as sets (even Ruby's Set doesn't provide `to_ary`)

proposing only having `to_a` in Collection and providing `to_ary` on List (at least for some compatibility)

Issue #459, opened 2012-12-21
Multiple assignment deconstruction fails with Java Lists


adamonduty:
I've come across an issue with multiple assignment that is similar to [JRUBY-6404](http://jira.codehaus.org/browse/JRUBY-6404).

I show current and expected behavior in [this gist](https://gist.github.com/4348505), but the idea is that Java ArrayList's (and presumably any Java List) don't deconstruct with multiple assignment the same as a ruby array:

``` ruby
  a = [1,2]
  b, c = a
  # => b = 1, c =2
```

With a Java ArrayList, `b = a` and `c = nil`.

In the gist, I used a small workaround by aliasing `to_ary` => `to_a`, but I'm guessing this isn't very efficient.

PR #6490, opened 2020-12-11
Remove to_hash from HashMap proxy

headius:
This addresses #6452 by removing `to_hash` coercion from our HashMap proxy, which avoids it being coerced to keyword arguments. It causes new Java integration failures in specs that expect the `to_hash` behavior, and I have left those failures intact as a demonstration of what will be lost by removing this behavior.

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -1,4 +1,3 @@
 source 'https://rubygems.org'
 
 gem 'rspec', '3.9.0'
-gem 'rspec-expectations', '3.9.2' # pinned to work around jruby/jruby#6452
--- a/core/src/main/java/org/jruby/java/proxies/MapJavaProxy.java
+++ b/core/src/main/java/org/jruby/java/proxies/MapJavaProxy.java
@@ -457,9 +457,9 @@ public final class MapJavaProxy extends ConcreteJavaProxy {
     /** rb_hash_to_hash
      *
      */
-    @JRubyMethod(name = { "to_hash", "to_h" })
-    public RubyHash to_hash(ThreadContext context) {
-        return getOrCreateRubyHashMap(context.runtime).to_hash(context);
+    @JRubyMethod(name = { "to_h" })
+    public RubyHash to_h(ThreadContext context) {
+        return getOrCreateRubyHashMap(context.runtime).to_h(context);
     }
 
     /** rb_hash_aset


kares:
should JRuby be "consistent" here - also not provide `to_ary` on array proxies and `java.util.List` ? 
```
>> java.util.ArrayList.new([1, 2]).to_ary
=> [1, 2]
>> ary = Java::int[2].new; ary[0] = 1; ary.to_ary
=> [1, 0]
```
... guess we would still want to keep `java.lang.String.new('foo').to_str` on Java strings since that seems legit.

headius:
> should JRuby be "consistent" here

I think I will try to determine the lineage of these methods as far as JI goes, and see if they were intentionally added to make these sorts of conversions easier, or if we added them "just because". Given the various tests and specs that failed it seems likely that these methods would be missed.

The original reason for proposing that we remove `to_hash` was due to the potential for it to be converted as part of normal keyword arguments handling, but based on https://github.com/rspec/rspec-expectations/issues/1241 the questionable behavior in RSpec does indeed appear to be incorrect. If they end up fixing it, then the original reason for removing `to_hash` goes away.

headius:
Unfortunately it looks like the primitive array `to_ary` goes back so far that most of the lineage behind that decision has been lost: c93e9184ca8e6452719372a5c036cf29bc8ff4c6

It may have been added for to resolve the mentioned issues, but the content of those issues was lost when Codehaus went away.

headius:
`List#to_ary` was added by you as a narrowing of `Collection#to_ary`, which indeed was inappropriate (since not all Collections are array-like): https://github.com/jruby/jruby/issues/3790

`Collection#to_ary` was added to support multiple assignment, which perhaps still makes sense to have for Lists? 80954e92cf45156a1320eb14fc2e7b5d72e56a0d https://github.com/jruby/jruby/issues/459 https://github.com/jruby/jruby/pull/460

And `to_hash` appears to have been added at the same time as custom Map support, so it wasn't for any other reason than to make Map work as much like Hash as possible: 61aca7697ce544ef03577bac7084feca359a5aa7

So given that the RSpec issue does indeed appear to be a bug with a possible fix, `to_ary` for deconstruction seems appropriate (also consider arg splatting), and `to_hash` is not actually causing any user-reported issues, I am leaning back toward the status quo and leaving things alone.

headius:
Any opinion here @enebo?

enebo:
@headius I am ok with this.  Some other issue I asked @kares what he thought could break but I keep coming up blank.  We could end up surprised but we should do this sooner than later to hopefully see a problem if there is one.

headius:
There is no reason to do this and several reasons not to. I think we should just leave the status quo in place until we have a recent to do differently.

Issue #6491, opened 2020-12-13
Undefine rlimit features when unsupported


headius:
In https://github.com/rubygems/rubygems/issues/4129 we were reminded of a recently fixed bug in our `Process.getrlimit` implementation (#6383) which has so far only been applied to master for 9.3. While it does correct the original issue and would also correct the Bundler and net-http-persistent issues, the actual fix would be to leave the rlimit-related methods and constants undefined (or marked as unimplemented) when their features cannot be provided in a given environment.

Doing this for Windows seems acceptable, since that matches how CRuby behaves: the related C functions and constants are not defined for Windows environments, so the Ruby methods and constants are unimplemented. Making this change for all platforms where we cannot support native libraries might be an overreach, however, since users on e.g. Linux may be more likely to blindly call these methods.

I will push a fix that disables this only for WIndows and we can discuss the ramifications of doing it for all non-native environments as well.

PR #6492, opened 2020-12-13
Disable rlimit features on Windows

headius:
See #6491 and https://github.com/rubygems/rubygems/issues/4129 for discussions of this issue.

We *could* also do this disabling when running on a platform that does not support our FFI, but that would increase the compatibility gap on such platforms.

### DIFF
--- a/core/src/main/java/org/jruby/RubyProcess.java
+++ b/core/src/main/java/org/jruby/RubyProcess.java
@@ -105,10 +105,17 @@ public class RubyProcess {
         process_sys.defineAnnotatedMethods(Sys.class);
 
         runtime.loadConstantSet(process, jnr.constants.platform.PRIO.class);
-        runtime.loadConstantSet(process, jnr.constants.platform.RLIM.class);
-        for (RLIMIT r : RLIMIT.values()) {
-            if (!r.defined()) continue;
-            process.defineConstant(r.name(), runtime.newFixnum(r.intValue()));
+
+        if (Platform.IS_WINDOWS) {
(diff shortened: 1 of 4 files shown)


headius:
I am unsure if there's a ruby/spec to add for this, since typically ruby/spec does not specify when features should **not** be implemented.

eregon:
> I am unsure if there's a ruby/spec to add for this, since typically ruby/spec does not specify when features should **not** be implemented.

It's totally fine to add a ruby/spec for this, i.e. testing that `Process.respond_to?(:getrlimit)` is false, and/or that the constants are not defined on Windows, especially since it's clear this affects compatibility in this case.

slonopotamus:
Would be great if this fix was quickly released as a 9.2.14.1 or something like that. Because currently latest JRuby + Windows + latest Bundle results in https://github.com/rubygems/rubygems/issues/4129.

headius:
@slonopotamus There will be a release of some sort but not until January. We are getting too close to the holiday season to do releases of stable JRuby versions.

It sounds like @deivid-rodriguez will put out Bundler 2.2.1 with the fix for JRuby (among others) and I think that will solve this problem automatically.

headius:
> It's totally fine to add a ruby/spec for this

Thank you for clarifying... I will also add some specs.

Past RubySpec policy was to avoid adding tests that "proved a negative". However if users can observe that a feature is implemented on some platforms and not implemented on others, it seems clear to me that we should have specifications for that behavior.

headius:
This is merged and there should be snapshots available soon at https://www.jruby.org/nightly

PR #6493, opened 2020-12-14
[ji] "usable" inspect for core Java types

kares:
This is, I believe, my 3rd attempt on Java types having usable `inspect` (last one [2 years ago](https://github.com/jruby/jruby/pull/5219) actually got reverted).

It's frustrating working with Java integration out-of-the-box (unless you alias all `alias inspect to_s`).
For example a rspec failure on a Java type will not get you anything useful: `#<Java::JavaTime::Instant:0xeded048>`

Even with a simple `inspect` -> `toString` alias (a previous attempt) I still find myself figuring out types manually.
e.g. `"[]"` is it an empty Ruby Array instance or a Java collection instance?

The proposal here would be to define simple inspect (Java#toString) on intermediate Java types: primitives, strings, enums (and maybe BigInteger/BigDecimal). This is already a compromise as sometimes it's useful to know it's a Java (`java.lang.String`) but since these are 'Ruby' convertible (`to_str`, `to_int`) it kind of makes sense to not go verbose.

The format is not set in stone - aligning all requires some format changes on Java arrays and Maps. 

JRuby obviously can not cover everything out there, but should be able to provide a nice experience with Java SDK types.
Loading too much won't be an issue - JI extensions already load lazily on first occurence of the type in Ruby scripted Java
(should refactor these further to not even initialize Java classes but use class names instead).

The default would be either a reflective (might require unlocking ALL for field reflection) or a bean-style inspect attempt.
Having an optional `reflective_inspect`, even if it isn't a default for generic Java objects, around would be really helpful for `jirb` sessions poking around Java APIs.

Current status:

```diff
1,20c1,21
< #<Java::JavaLang::String:0x4ad896f7>
> "fó\"bař\""
< #<Java::JavaLang::Character:0x7d61468c>
> 'c'
< #<Java::JavaLang::Float:0x4b98225c>
> 10.01
< #<Java::JavaMath::BigInteger:0x3c488b34>
> 1111111111111111111
< #<Java::JavaUtilConcurrent::TimeUnit:0x3866c96e>
> SECONDS
< byte[0, 127, -128]@5c76fce4
> #<Java::byte[3]: [0, 127, -128]>
< float[]@5aab5b31
> #<Java::float[0]: []>
< java.lang.String[bar]@5408d4b3
> #<Java::JavaLang::String[1]: ["bar"]>
< #<Java::JavaTime::Instant:0xeded048>
> #<Java::JavaTime::Instant: 2020-12-14T05:19:14.785116Z>
< #<Java::JavaTime::LocalDateTime:0x655621fd>
> #<Java::JavaTime::LocalDateTime: 2020-12-14T06:19:14.790763>
< #<Java::JavaTime::LocalDate:0xfcd0e8d>
> #<Java::JavaTime::LocalDate: 2020-12-14>
< #<Java::JavaTime::Year:0x1a4083f6>
> #<Java::JavaTime::Year: 2020>
< #<Java::JavaUtil::Date:0x44abdb1f>
> #<Java::JavaUtil::Date: Mon Dec 14 06:19:14 CET 2020>
< #<Java::JavaSql::Time:0x7cd1ec54>
> #<Java::JavaSql::Time: 01:00:00>
< #<Java::JavaUtilRegex::Pattern:0x6d420cdd>
> #<Java::JavaUtilRegex::Pattern: /.*?/>
< #<Java::JavaUtil::ArrayList:0x78128dd0>
> #<Java::JavaUtil::ArrayList: [:symbol]>
< #<Java::JavaUtil::Arrays::ArrayList:0x28dd81ad>
> #<Java::JavaUtil::Arrays::ArrayList: [[#<Java::JavaLang::StringBuilder: "str">], #<Java::JavaTime::LocalDateTime: 2020-12-14T06:19:14.841896>, #<Foo:0x3a5e2525 @foo=true>]>
< #<Java::JavaUtil::HashSet:0x5c9a4f3b>
> #<Java::JavaUtil::HashSet: [0, 1, 2]>
< #<Java::JavaUtilConcurrent::LinkedBlockingDeque:0x6f24ce45>
> #<Java::JavaUtilConcurrent::LinkedBlockingDeque: [1, "2"]>
< {2=>"val2", 10=>"val1"}
> #<Java::JavaUtilConcurrent::ConcurrentHashMap: {2=>"val2", 10=>"val1"}>
```

### DIFF
--- a/core/src/main/java/org/jruby/java/proxies/MapJavaProxy.java
+++ b/core/src/main/java/org/jruby/java/proxies/MapJavaProxy.java
@@ -136,9 +136,9 @@ public final class MapJavaProxy extends ConcreteJavaProxy {
             RubyStringBuilder.cat(runtime, buf, SPACE);
 
             if (size() == 0) {
-                RubyStringBuilder.cat(runtime, buf, EMPTY_HASH_BYTES);
+                RubyStringBuilder.cat(runtime, buf, EMPTY_HASH_BL);
             } else if (runtime.isInspecting(map)) {
-                RubyStringBuilder.cat(runtime, buf, RECURSIVE_HASH_BYTES);
+                RubyStringBuilder.cat(runtime, buf, RECURSIVE_HASH_BL);
             } else {
                 try {
                     runtime.registerInspecting(map);
--- a/core/src/main/java/org/jruby/java/proxies/MapJavaProxy.java
+++ b/core/src/main/java/org/jruby/java/proxies/MapJavaProxy.java
@@ -51,6 +51,9 @@ import java.lang.reflect.InvocationTargetException;
 import java.util.Collection;
 import java.util.Map;
 import java.util.Set;
+
+import org.jruby.util.Inspector;
+import org.jruby.util.RubyStringBuilder;
 import org.jruby.util.TypeConverter;
 
 /**
@@ -125,7 +128,25 @@ public final class MapJavaProxy extends ConcreteJavaProxy {
 
         @Override
         public IRubyObject inspect(ThreadContext context) {
-            return super.inspect(context);
+            final Ruby runtime = context.runtime;
+
+            RubyString buf = Inspector.inspectStart(context, receiver.getMetaClass());
+            RubyStringBuilder.cat(runtime, buf, Inspector.SPACE);
+
+            if (size() == 0) {
+                RubyStringBuilder.cat(runtime, buf, EMPTY_HASH_BYTES);
+            } else if (runtime.isInspecting(this)) {
+                RubyStringBuilder.cat(runtime, buf, RECURSIVE_HASH_BYTES);
+            } else {
+                try {
+                    runtime.registerInspecting(this);
+                    buf.cat19(inspectHash(context));
+                } finally {
+                    runtime.unregisterInspecting(this);
+                }
+            }
+
+            return RubyStringBuilder.cat(runtime, buf, Inspector.GT);
         }
 
         @Override
--- a/core/src/main/java/org/jruby/java/proxies/MapJavaProxy.java
+++ b/core/src/main/java/org/jruby/java/proxies/MapJavaProxy.java
@@ -461,12 +461,10 @@ public final class MapJavaProxy extends ConcreteJavaProxy {
         return (RubyProc) newProc;
     }
 
-    /** rb_hash_to_s
-     *
-     */
+    // NOTE: keep Map#to_s -> toString as with other Java types
     @JRubyMethod(name = "to_s")
     public IRubyObject to_s(ThreadContext context) {
-        return getOrCreateRubyHashMap(context.runtime).to_s(context);
+        return RubyString.newString(context.runtime, getMapObject().toString());
     }
 
     /** rb_hash_rehash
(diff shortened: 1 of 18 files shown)


headius:
I am still on board with the idea. My issue in the past has been that `toString` is frequently a bad choice, since there are large data objects that will essentially render everything. Your approach here is cleaner in that regard; you are doing one-off inspects for specific types rather than just dumping `inspect` calls right into `toString`.

As long as we avoid invasive reflection and accept that most user types will not have a safe, generic way to inspect, I think this will be good and useful.

core/src/main/java/org/jruby/java/proxies/MapJavaProxy.java
```diff
@@ -438,12 +461,10 @@ public RubyProc to_proc(ThreadContext context) {
...
     }
 
-    /** rb_hash_to_s
-     *
-     */
+    // NOTE: keep Map#to_s -> toString as with other Java types
     @JRubyMethod(name = "to_s")
     public IRubyObject to_s(ThreadContext context) {
-        return getOrCreateRubyHashMap(context.runtime).to_s(context);
+        return RubyString.newString(context.runtime, getMapObject().toString());
```
kares, line 467: we keep `to_s` -> `toString` for other collections so this seemed inconsistent.
do we want to keep it since we changed `inspect` already to not be a `to_s` alias?

always a bit 2 sided between consistency and "improvements", JRuby re-defs `to_s` on:
- `java.lang.Throwable` - `to_s` -> `message`
- `java.lang.Class` - `to_s` -> `name` (toString is `"class java.lang.String"`)
- on Java arrays as the default `toString` is useless

would be personally inclined towards `java.util.Map` not being wrapped in an internal `Hash` (like the rest of proxies),
in which case `to_s` just ends up as `toString`

headius: > would be personally inclined towards java.util.Map not being wrapped in an internal Hash

Me too. I think our goal should be to get MapJavaProxy as close as possible to just patching Map rather than all the places it gives up and turns it into a "special" Hash.

headius: If it wasn't clear I support this change. 👍

core/src/main/java/org/jruby/util/Inspector.java
```diff
@@ -0,0 +1,56 @@
...
+
+import static org.jruby.util.io.EncodingUtils.encStrBufCat;
+
+public class Inspector {
+
+    public static final ByteList EMPTY_ARRAY_BYTES = new ByteList(new byte[] { '[',']' }, USASCIIEncoding.INSTANCE, false);
+    public static final ByteList RECURSIVE_ARRAY_BYTES = new ByteList(new byte[] { '[','.','.','.',']' }, USASCIIEncoding.INSTANCE, false);
+
+    public static final ByteList EMPTY_HASH_BYTES = new ByteList(new byte[] { '{','}' }, USASCIIEncoding.INSTANCE, false);
+    public static final ByteList RECURSIVE_HASH_BYTES = new ByteList(new byte[] { '{','.','.','.','}' }, USASCIIEncoding.INSTANCE, false);
```
headius, line 17: I know "BYTELIST" is longer than "BYTES"... maybe we can use "BL" to indicate that these are more than just the bytes?  Just a minor nitpick.

headius (approved):
Looks good to me.

kares:
:taco: thanks, going to merge this one and follow up with a few more core `inspect`s that would be nice to have in another PR.

Issue #6499, opened 2020-12-16
Array shuffle method should fail if prng is passed without being a named argument


jmalves:
**Environment Information**
  Tested on Jruby 9.2.14.0

**Expected Behavior**

> 2.5.0 :001 > prng = Random.new(1000); (0..10).to_a.shuffle(prng)
> Traceback (most recent call last):
>         3: from /Users/jalves/.rvm/rubies/ruby-2.5.0/bin/irb:11:in `<main>'
>         2: from (irb):1
>         1: from (irb):1:in `shuffle'

**Actual Behavior**

> jruby-9.2.14.0 :004 > prng = Random.new(1000); (0..10).to_a.shuffle(prng)
>  => [3, 10, 9, 6, 7, 2, 4, 5, 1, 0, 8]

So I would expect the shuffle method to throw an exception, which does not happen in the JRuby case.

headius:
Fixed for whatever release comes next, along with `Array#sample`.

PR #6500, opened 2020-12-16
Fix Array shuffle and sample arity checking

headius:
Fixing shuffle at first for #6499 but sample appears to have the same problem.


headius:
This fixes the missing arity error for shuffle and sample but I did not add any specs for these error conditions.

headius:
Well crap, that was supposed to go to 9.2. I will cherry-pick I guess.

PR #6503, opened 2020-12-21
Cleanup FFI platform files

headius:
A number of files under the `ffi/platform` subdir appear to have lineage back to the original addition of FFI support to JRuby and were used by Rubinius to support their own core and standard library classes that used FFI for implementation. Most of these files are no longer used by JRuby, and none of them are versioned by the FFI gem, so this PR will transplate or eliminate them.

Most files are FFI constant declarations, while the `platform.conf` files appear to have been used internally by Rubinius. Of these files, only the `syslog.rb` constant definitions and the `fcntl` related files under x86-64 solaris appear to be in use in JRuby.

Two platform `types.conf` files exist that do not appear to be in the FFI gem:

* an enigmatic `i486-gnu` platform
* a `powerpc64le-linux` platform

Of these two files, only the ppc file appears to be relevant, and should probably be contributed back to the FFI gem.


headius:
@larskanis The commit core/src/main/ruby/jruby/kernel/file.rb deletes two platform `types.conf` files that do not exist in the FFI gem. The `i486-gnu` is probably too weird to bother with, but perhaps the `powerpc64le-linux` config should be transferred to the gem?

larskanis:
@headius I added the `powerpc64le-linux/types.conf` to ffi gem in https://github.com/ffi/ffi/commit/cf520d748a28d36c00019765eedb227368f3b371 .

PR #6513, opened 2020-12-29
[ji] eliminate JavaClass usage in favor of Java wrappers

kares:
JRuby's Java integration has been using a custom `JavaClass` ruby class to represent a `java.lang.Class` instance.
However this isn't super easy to work with (personally I have always find it quite confusing given the existing complexity of proxy modules/classes) nor has been 100% consistent e.g. `obj.java_class` depending on the proxy type (interface impl, Java sub-class in Ruby) might have returned a proxy instead of the `JavaClass` instance (returned from `java.lang.String.java_class`).

The proposal here is to start returning `java.lang.Class` instance Java proxy wrappers everywhere (all `java_class` calls), for compatibility all of the custom `JavaClass` Ruby API is supported on `java.lang.Class` and others (there's  `JavaField`, `JavaConstructor` instances `JavaClass` methods where returning which will be 'emulated' using `java.lang.reflect.Field`, `java.lang.reflect.Constructor`).

### DIFF
--- a/core/src/main/java/org/jruby/RubyClass.java
+++ b/core/src/main/java/org/jruby/RubyClass.java
@@ -1841,8 +1841,8 @@ public class RubyClass extends RubyModule {
             if (reifiedClass == null) reifyWithAncestors(); // possibly auto-reify
             // Class requested; try java_class or else return nearest reified class
             final ThreadContext context = runtime.getCurrentContext();
-            IRubyObject javaClass = JavaClass.java_class(context, this);
-            if ( ! javaClass.isNil() ) return javaClass.toJava(target);
+            Class<?> javaClass = JavaClass.getJavaClassIfProxy(context, this);
+            if (javaClass != null) return (T) javaClass;
 
             Class reifiedClass = nearestReifiedClass(this);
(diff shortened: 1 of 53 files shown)


headius:
I have made quick attempts at this in the past and I support this change. Anything we can do to eliminate that legacy hierarchy of JavaClass and friends would be great!

kares:
@headius definitely - this has been very difficult to work with.
I think all in all this is quite close and existing specs/tests pass for the most part with `java_class` simply returning `java.lang.Class`instance proxies. the only compatibility 'issue' atm is not mapping exceptions to a Ruby hierarchy.

anyhow, we will need to make a decision point whether, despite no longer returning `JavaClass` instances, we keep the `Java::JavaClass` constant around for (Ruby) consumers or rather point to `JavaClass = java.lang.Class` in which case there's the opportunity to remove all of the JavaClass JavaMethod native implementations. or would those be considered semi-blessed JRuby API we should keep around? I think not but thought I rather double check.

headius:
I would love to see this land soon. Time to start wrapping up 9.3 for reals.

PR #6522, opened 2021-01-11
optimize Array#unshift

ahorek:
before
```
[1,2,3] # new array
[_,1,2,3,_,_] # extend the array, move ptr by 1
[5,1,2,3,_,_] # unshift
[5,5,1,2,3,_] # unshift 2 = sliding array, but we have to copy all elements
```

after
```
[1,2,3] # new array
[_,_,_,1,2,3] # extend the array, move elements to the right
[_,_,5,1,2,3] # unshift
[_,5,5,1,2,3] # unshift 2, if there's a free capacity, just shift pointers and skip reallocations
```

results
```
before
             unshift    172.393  (? 0.6%) i/s -    867.000  in   5.029295s
after
             unshift      3.896k (? 2.5%) i/s -     19.760k in   5.075590s (22.8x faster)
ruby 2.7.2
             unshift      2.332k (± 4.8%) i/s -     11.809k in   5.078204s
```

script
```
array = []
10_000.times do |i|
  array.unshift(i)
end
```

### DIFF
--- a/core/src/main/java/org/jruby/RubyArray.java
+++ b/core/src/main/java/org/jruby/RubyArray.java
@@ -1081,6 +1081,19 @@ public class RubyArray<T extends IRubyObject> extends RubyObject implements List
         values = vals;
     }
 
+    private void unshiftRealloc(int valuesLength) {
+        final long newLength = valuesLength << 1;
+        if (newLength >= Integer.MAX_VALUE) {
+            throw getRuntime().newIndexError("index " + newLength + " too big");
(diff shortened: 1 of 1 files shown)


headius (approved):
Looks fine to me.

This does assume that `unshift` and `shift` will not typically be mixed with `push` and `pop`, since we would want to slide the array different directions for the two use cases. Perhaps it is ok if these two patterns optimize in opposite ways?

PR #6525, opened 2021-01-12
Make Pathname#inspect behave like other Ruby implementations (fixes #6519)

### LINKED ISSUE #6519: Pathname#inspect behaviour is not consistent with other Ruby implementations
I'm running JRuby 9.2.13.0 on Mac OS X (`Darwin Kernel Version 18.7.0`).

While debugging a JRuby-specific failure in `sshkit` (https://github.com/capistrano/sshkit/issues/488), I attempted to run the `sshkit` test suite, and noticed 3 failures caused by the way JRuby currently implements `Pathname#inspect`.

Only with JRuby, a heading hash (`#`) is missing.

```
$ RBENV_VERSION=2.3.8 ruby -e "require 'pathname'; puts Pathname.new('/tmp').inspect"
#<Pathname:/tmp>
$ RBENV_VERSION=3.0.0 ruby -e "require 'pathname'; puts Pathname.new('/tmp').inspect"
#<Pathname:/tmp>
$ RBENV_VERSION=truffleruby-20.2.0 ruby -e "require 'pathname'; puts Pathname.new('/tmp').inspect"
#<Pathname:/tmp>
$ RBENV_VERSION=jruby-9.2.13.0 ruby -e "require 'pathname'; puts Pathname.new('/tmp').inspect"
<Pathname:/tmp>       <---- missing # at beginning of line
```

While it may appear like a minor issue, or one that could cause breakage if we change it in JRuby right now, I believe it hinders JRuby use a bit, by causing small discrepancies in libraries test suites.

If possible, it would be great to align the behaviour on what MRI does.

thbar:
Opening a PR to verify if the tests pass via CI, because I haven't been quickly able to run the tests locally.

I also wonder if I need to "sync" the Ruby specs (see #6519 and https://github.com/ruby/spec/pull/822), or add something in the test suite for this.

### DIFF
--- a/core/src/main/java/org/jruby/ext/pathname/RubyPathname.java
+++ b/core/src/main/java/org/jruby/ext/pathname/RubyPathname.java
@@ -298,7 +298,7 @@ public class RubyPathname extends RubyObject {
 
     @JRubyMethod
     public IRubyObject inspect(ThreadContext context) {
-        return context.runtime.newString("<Pathname:" + getPath() + ">");
+        return context.runtime.newString("#<Pathname:" + getPath() + ">");
     }
 
     @JRubyMethod(required = 1, optional = 1, reads = BACKREF, writes = BACKREF)


headius:
The fix is trivial and the failure is not your fault.

Specs are currently being updated by @eregon on a semi-regular basis, so it will get into our repo sooner or later.

PR #6532, opened 2021-01-13
Improvements for popen process launch in Windows

### LINKED ISSUE #6516: When extension is omitted, IO.popen does not find the executable in JRuby on Windows
**Environment Information**
  
JRuby version and operating system:
* jruby 9.2.14.0 (2.5.7) 2020-12-08 ebe64bafb9 OpenJDK 64-Bit Server VM 25.275-b01 on 1.8.0_275-b01 +jit [mswin32-x86_64]
* jruby 9.3.0.0-SNAPSHOT (2.6.5) 2020-12-28 08642361e6 OpenJDK 64-Bit Server VM 25.275-b01 on 1.8.0_275-b01 +jit [mswin32-x86_64]

Other relevant info:
* I think this is another instance of #5018

**Expected Behavior**

GIVEN the following Ruby code running in JRuby on Windows

AND the following statement succeeds:
```Ruby
output = IO.popen(["gem.bat", "build", "-V", "jruby_test.gemspec"], &:read)
```

AND the PATHEXT environment variable is set to `.COM;.EXE;.BAT;.CMD;.VBS;.VBE;.JS;.JSE;.WSF;.WSH;.MSC;.CPL`

Then following statement (which omits the ".bat" extension) should also succeed:
```Ruby
output = IO.popen(["gem", "build", "-V", "jruby_test.gemspec"], &:read)
```

I noticed this in my CI job for my gem.  I do not have a windows box to test this on, but was able to duplicate this using GitHub Actions using a Windows VM:

https://github.com/jcouball/jruby_test/blob/main/Rakefile#L7

In JRuby on Windows, the `popen_test_1` rake task succeeds and the `popen_test_2` rake task fails.

**Actual Behavior**

The second command fails with the error:

```
IOError: Cannot run program "gem" (in directory "D:\a\jruby_test\jruby_test"): CreateProcess error=2, The system cannot find the file specified
```

headius:
The primary fix here is to do the same argument processing and executable searching for `popen` that we do for `system` by aligning the relevant logic. The other fix is to handle another special case JDK error when launching a process using a path that does not exist.

This will fix #6516 once finished.

### DIFF
--- a/core/src/main/java/org/jruby/runtime/Helpers.java
+++ b/core/src/main/java/org/jruby/runtime/Helpers.java
@@ -261,6 +261,12 @@ public class Helpers {
             return errnoFromMessage(dnee);
         } catch (BindException be) {
             return errnoFromMessage(be);
+        } catch (IOException ioe) {
+            String message = ioe.getMessage();
+            // Raised on Windows for process launch with missing file
+            if (message.endsWith("The system cannot find the file specified")) {
+                return Errno.ENOENT;
+            }
         } catch (Throwable t2) {
             // fall through
         }
--- a/core/src/main/java/org/jruby/util/ShellLauncher.java
+++ b/core/src/main/java/org/jruby/util/ShellLauncher.java
@@ -811,33 +811,17 @@ public class ShellLauncher {
             }
 
             String[] args = parseCommandLine(runtime.getCurrentContext(), runtime, strings);
-            LaunchConfig lc = new LaunchConfig(runtime, strings, false);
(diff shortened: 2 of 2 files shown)


headius:
This change is a bit of a shot in the dark, since our testing on Windows is a little weak. I was able to execute the early stages of @enebo's verification script, so things seem to be working, but a more complete verification is warranted.

headius:
This has been verified by me testing basic gem installs and @enebo running his pre-release Windows verification scripts that build and run a Rails app. We'll go with it.

PR #6534, opened 2021-01-14
Move Monitor to native

headius:
See https://github.com/ruby/ruby/pull/2576 which did the same for CRuby 2.7.

This eliminates the need for the monitor.rb fixes in #6527 and ruby/monitor#2, plus synchronize will perform significantly better than the nested block implementation.

This may be safe to include in 9.3 even though the compatibility level is 2.6.


headius:
The most likely vector for compatibility issues here is the removal of MonitorMixin from Monitor's includes. Before, Monitor included MonitorMixin to get most of its functionality, but Monitor is the captain now.

https://github.com/jruby/jruby/pull/6534/files#diff-932b5c8ed1b8e197d26de4132979a13401fa78b07cbb1faf1c742851748674d8L306

Issue #6541, opened 2021-01-27
Errno::ENOPROTOOPT when connecting to Redis with JRuby-9.2.14.0


ThomasKoppensteiner:
Connecting to Redis with `tcp_keepalive` option set raises an `Errno::ENOPROTOOPT` error in `JRuby-9.2.14.0`. It works in `JRuby-9.2.13.0`. I think this is because [this PR](https://github.com/jruby/jruby/pull/6388) removed `SOL` constants from `RubyBasicSocket`, but didn't add it to `RubyTCPSocket` (and others).

### Example
```ruby
#!/usr/bin/env ruby
# frozen_string_literal: true

require "redis"

class ConnectionFactory
  def self.build_connection_without_keepalive
    ::Redis.new(host: REDIS_HOST, port: REDIS_PORT, db: REDIS_DB)
  end

  def self.build_connection_with_keepalive
    ::Redis.new(host: REDIS_HOST, port: REDIS_PORT, db: REDIS_DB, tcp_keepalive: 60)
  end

  REDIS_HOST="127.0.0.1"
  REDIS_PORT=6379
  REDIS_DB=0

  private_constant :REDIS_HOST, :REDIS_PORT, :REDIS_DB
end

begin
  puts "Connection without TCP keepalive..."
  workaround_con = ConnectionFactory.build_connection_without_keepalive
  workaround_con.set("foo1", "bar1")
  val = workaround_con.get("foo1")
  puts "Value fetched: #{val}"
rescue StandardError => e
  puts e.class
  puts e.message
  puts e.backtrace
end

puts "---"

begin
  puts "Connection with TCP keepalive..."
  broken_con = ConnectionFactory.build_connection_with_keepalive
  broken_con.set("foo2", "bar2")
  val = broken_con.get("foo2")
  puts "Value fetched: #{val}"
rescue StandardError => e
  puts e.class
  puts e.message
  puts e.backtrace
end
```
##### JRuby-9.2.13.0 Output
```sh
Connection without TCP keepalive...
Value fetched: bar1
---
Connection with TCP keepalive...
Value fetched: bar2
```

##### JRuby-9.2.14.0 Output
```sh
Connection without TCP keepalive...
Value fetched: bar1
---
Connection with TCP keepalive...
Errno::ENOPROTOOPT
Protocol not available - Protocol not available
org/jruby/ext/socket/RubyBasicSocket.java:327:in `setsockopt'
[...]

PR #6542, opened 2021-01-28
[socket] improve keep alive support

ahorek:
this PR improves Redis support for their TCP keep-alive implementation https://github.com/jruby/jruby/issues/6541

also fixes a bug, getsockopt returns zero if the passed constant isn't supported. It should always be a SocketOption or an exception.


headius (approved):
Looks good, thanks!

PR #6543, opened 2021-01-28
Update specs

eregon:


eregon:
See https://github.com/jruby/jruby/issues/6544

Issue #6547, opened 2021-02-02
NoMethodError for Concurrent::Hash#deep_merge!


mohamedhafez:
In JRuby version 9.2.14.0 `Concurrent::Hash#deep_merge!` is undefined, but it is defined for ~~Ruby 2.6.6~~ Ruby 2.5.8. This is causing an error in my Rails app when upgrading to version 1.8.8 of the `i18n` gem, which uses the method (see https://github.com/ruby-i18n/i18n/issues/554)

```
NoMethodError: undefined method `deep_merge!' for {}:Concurrent::Hash
Did you mean?  deep_merge
                      store_translations at /Users/mohamed/.rvm/gems/jruby-9.2.14.0/gems/i18n-1.8.8/lib/i18n/backend/simple.rb:45
                               load_file at /Users/mohamed/.rvm/gems/jruby-9.2.14.0/gems/i18n-1.8.8/lib/i18n/backend/base.rb:230
                                    each at org/jruby/RubyHash.java:1415
                               load_file at /Users/mohamed/.rvm/gems/jruby-9.2.14.0/gems/i18n-1.8.8/lib/i18n/backend/base.rb:230
                       load_translations at /Users/mohamed/.rvm/gems/jruby-9.2.14.0/gems/i18n-1.8.8/lib/i18n/backend/base.rb:18
                                    each at org/jruby/RubyArray.java:1809
                       load_translations at /Users/mohamed/.rvm/gems/jruby-9.2.14.0/gems/i18n-1.8.8/lib/i18n/backend/base.rb:18
                       init_translations at /Users/mohamed/.rvm/gems/jruby-9.2.14.0/gems/i18n-1.8.8/lib/i18n/backend/simple.rb:80
                             eager_load! at /Users/mohamed/.rvm/gems/jruby-9.2.14.0/gems/i18n-1.8.8/lib/i18n/backend/simple.rb:65
                             eager_load! at /Users/mohamed/.rvm/gems/jruby-9.2.14.0/gems/i18n-1.8.8/lib/i18n.rb:77
                                    each at org/jruby/RubyArray.java:1809
                                Finisher at /Users/mohamed/.rvm/gems/jruby-9.2.14.0/gems/railties-6.0.3.4/lib/rails/application/finisher.rb:123
                           instance_exec at org/jruby/RubyBasicObject.java:2694
[...]

PR #6549, opened 2021-02-03
Make sync and profiling method wrapper logic work with refinements

headius:
This PR will track work to make our two types of "transparent" delegating method wrappers compatible with refinements.

As described in #6547, the wrapper methods that the Synchronized module uses interfere with refinement logic due to the latter's dependence on its own wrapper. When the two are active at the same time, the method lookup process may fail to locate methods that are clearly there, due to not knowing how to access the original unwrapped method properly.

The first fix here is a one-off that pre-unwraps all DelegatingDynamicMethod wrappers before proceeding to the rest of the refined logic, but the issue is systemic; the synchronized wrappers cannot be perfectly transparent, and so they interfere with method table logic that depends on seeing the original method. An alternative solution is needed.

One such solution might be to move the synchronization logic to a specialized CacheEntry, that will provide the wrapper only when being used to invoke. All other accesses would continue to see the original method. This would avoid eagerly wrapping the retrieved method and hiding its true nature, but it would open up the possibility that the unsynchronized version would leak out and be invokable without synchronization further down the line.

The TruffleRuby equivalent to this eagerly replaces all methods with new methods that actively synchronize; they are normal methods, not wrappers, so refinements do not interfere with them. However this only affects methods defined on the class at the point at which the class is flipped to being synchronized; new methods will not be synchronized unless the switch is again flipped. This does not seem to be an acceptable alternative to adding synchronization at lookup or invocation time.

### DIFF
--- a/core/src/main/java/org/jruby/RubyModule.java
+++ b/core/src/main/java/org/jruby/RubyModule.java
@@ -75,6 +75,7 @@ import org.jruby.internal.runtime.methods.AliasMethod;
 import org.jruby.internal.runtime.methods.AttrReaderMethod;
 import org.jruby.internal.runtime.methods.AttrWriterMethod;
 import org.jruby.internal.runtime.methods.DefineMethodMethod;
+import org.jruby.internal.runtime.methods.DelegatingDynamicMethod;
 import org.jruby.internal.runtime.methods.DynamicMethod;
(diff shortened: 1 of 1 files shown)


headius:
The larger fixes to change how Synchronized and profiling works will need to come in a separate PR, and probably only applied to master due to the invasiveness of the change.

PR #6551, opened 2021-02-03
Dir file walking encoding improvements

headius:
A set of tweaks to align encoding handling across the Dir filesystem-walking methods.


headius:
This aligns encoding logic to help bring resolution to #3205.

PR #6553, opened 2021-02-08
getResource() should never be called with a starting slash when called on a classloader

ahus1:
Calling it with a slash might or might not work depending on the classloader.
A starting slash is only used when calling getResource() on a class.

JRuby using a "/" at the beginning cased an exception when running this inside the plugin class loader of IntelliJ (https://github.com/asciidoctor/asciidoctor-intellij-plugin/issues/664). 
I'd be happy if this could be included in an upcoming patch release, so I can remove a workaround in the IntelliJ AsciiDoc plugin.

See https://stackoverflow.com/questions/47900677/where-does-leading-slash-in-java-class-loader-getresource-leads-to for more information on slashes and classloaders.

### DIFF
--- a/core/src/main/java/org/jruby/ext/rbconfig/RbConfigLibrary.java
+++ b/core/src/main/java/org/jruby/ext/rbconfig/RbConfigLibrary.java
@@ -277,7 +277,7 @@ public class RbConfigLibrary implements Library {
         setConfig(context, CONFIG, "target_cpu", arch);
 
         String jrubyJarFile = "jruby.jar";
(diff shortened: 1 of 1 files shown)


headius:
This is a simple fix. I will cherry-pick it to the 9.2 branch and merge that to master so it is in 9.2.15.0 or 9.3, whichever comes next (hopefully within the next month).

headius:
Cherry-picked to jruby-9.2 in 0013ec58ba855d002c7c4ee8f631d0be5648a6ac and will be merged to master shortly. Thank you!

PR #6558, opened 2021-02-12
[deps] update rspec version to 3.10.0

kares:


headius:
The gem plugin stopped working some time back so `-Pbootstrap` has been broken and not used. That's why it doesn't use that property anymore.

Probably need to get the gem plugin fixed some day.

PR #6562, opened 2021-02-15
[feat] support RubyString#split for Java (exts)

kares:
currently, there's no easy way to do a `RubyString#split` from a Java extension (due split writing a back-ref for the frame).
the idea here is to provide a "strongly" typed API that Java consumers could leverage wout shooting themselves in the :foot:


headius:
Big +1 for making the backref stuff more explicit. You have probably seen some Regexp functions where you can pass in a holder object or let it do the implicit operation, but this is a nice way to make it more typed.

PR #6567, opened 2021-02-18
CI: quick test on how ARM64 is doing

kares:
... (sorry I do not have credits to test on my own fork :crying_cat_face: )

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -32,6 +32,9 @@ jdk:
 os:
   - linux
 
+arch:
+  - arm64
+
 env:
   global:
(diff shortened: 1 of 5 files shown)


headius:
Linux AArch64 should be pretty good but I have not made an effort to tag off things that do fail there. Apple Arm64 still has issues.

kares:
for whatever reason CI stopped working recently - now only runs AppVeyor but not Travis nor GH actions :crying_cat_face:

headius:
It's not because it is a draft is it?

kares:
moved to https://github.com/jruby/jruby/pull/6567

PR #6571, opened 2021-02-19
Fixes #6563. Rails 6.1.2.x dynamic finders cause Java::JavaLang::ClassCastException

### LINKED ISSUE #6563: Rails 6.1.2.x dynamic finders cause Java::JavaLang::ClassCastException
**Environment Information**
  
* JRuby version (`jruby -v`): `jruby 9.2.14.0 (2.5.7) 2020-12-08 ebe64bafb9 Java HotSpot(TM) 64-Bit Server VM 25.45-b02 on 1.8.0_45-b14 +jit [darwin-x86_64]`
* Operating system and platform: initially discovered on MacOS 10.13.6, but replicated on Travis under Ubuntu 18.04.5 LTS (Bionic), see below
* Application/framework version: Rails 6.1.2.1
* Code available at: https://github.com/sosol/sosol/tree/rails-6.1.2.1

**Expected Behavior**

* Dynamic Rails finder methods e.g. `find_by_custom_model_attribute()` should continue to work

**Actual Behavior**

* Dynamic Rails finder methods result in `Java::JavaLang::ClassCastException: org.jruby.ir.operands.TemporaryClosureVariable cannot be cast to org.jruby.ir.operands.DepthCloneable` ([full backtrace from Travis](https://travis-ci.com/github/sosol/sosol/jobs/483537961) under OpenJDK 8 [attached here](https://github.com/jruby/jruby/files/5990744/classcastexception.txt))

I was able to isolate this down to dynamic finders by running a single test and adding debug logging until I could pin it down. With e.g. [this change to use `where` instead](https://github.com/sosol/sosol/commit/3e71cdc31c9a87a34d19fd764c5fce966eeaf5d6), the single test I chose passes instead of resulting in `Java::JavaLang::ClassCastException` (single test invoked with `rails test test/unit/publication_test.rb:91`).

Exception occurs on both Rails 6.1.2.1 and Rails 6.1.2.

Exception does **not** occur under Rails 6.1.1 with nothing else changed: https://travis-ci.com/github/sosol/sosol/builds/217275737

I also expanded tests from OpenJDK 8 up through OpenJDK 15 to verify that it's not related to the older JVM with Rails 6.1.2.1: https://travis-ci.com/github/sosol/sosol/builds/217273487

Only difference between the two is the Rails version (+ dependencies): https://github.com/sosol/sosol/compare/rails-6.1.2.1...rails-6.1.1

Please let me know if you think I should file a regression on Rails or whichever `action/active*` gem change is responsible instead of or in addition to here.

enebo:
Fixes #6563

90% of the changes here are for JRuby to dynamically raise an exception if we try and use zsuper from within a define_method block.

The important part is instead of building up a list of argument operands at each level so we can support zsuper we just delete all that.  So removing the logic we should not have been doing in the first place means we do not hit a bug in that code in the first place.


core/src/main/java/org/jruby/ir/IRBuilder.java
```diff
@@ -3980,75 +3982,77 @@ public Operand buildZArray(Variable result) {
...
-                    /* Should we add an exception throw here just in case? */
-                //}
+        return zsuperResult;
+    }
 
-                addInstr(new LabelInstr(allDoneLabel));
-                return zsuperResult;
-            }
-        };
+    private boolean isMethodDefine() {
```
headius, line 4024: Not `isDefineMethod`?

core/src/main/java/org/jruby/ir/IRBuilder.java
```diff
@@ -3980,75 +3982,77 @@ public Operand buildZArray(Variable result) {
...
+                exceptionClass,
+                new Operand[]{new StringLiteral(message)},
+                null));
+        Operand kernel = searchModuleForConst(new ObjectClass(), runtime.newSymbol("Kernel"));
+        addResultInstr(CallInstr.create(scope,
+                createTemporaryVariable(),
+                runtime.newSymbol("raise"),
+                kernel,
+                new Operand[] { exception },
+                null));
```
headius, line 4049: You could just call this as `Kernel.raise WhateverException, "message"` rather than doing the call to `new`.

headius (approved):

core/src/main/java/org/jruby/ir/IRBuilder.java
enebo: yeah I yoda named that for no reason.   I will change that.

core/src/main/java/org/jruby/ir/IRBuilder.java
enebo: sure.  I can make this change...less generated instrs + bytecode.

enebo:
Spurious failure on travis on some MT test writing IO.  Unrelated to this.

PR #6572, opened 2021-02-23
Optimization for calling java vararg methods on hotspot. ~10x speedup

byteit101:
Array.set is much slower than raw array access, and leads to a very
laggy ui in certain JRubyFX workloads that involve lots of vararg
calls to java apis. This patch optimizes that case, and leads to a
resulting, user-visible ~10x speedup

This change was pulled out of https://github.com/jruby/jruby/pull/6422 work


kares (approved):
LGTM

headius (approved):

headius:
Very nice!

PR #6576, opened 2021-02-24
Unrevert racc gem

headius:
It was not actually the cause of any issues and we would like to have it be an upgradable default gem.


headius:
See #6517 and #6537.

Issue #6579, opened 2021-02-25
Upgrade to 9.2.15.0 causes Kernel#system to hang. 


jonathanswenson:
**Environment Information**
  
Provide at least:
* JRuby version (`jruby -v`) and command line (flags, JRUBY_OPTS, etc)
`jruby 9.2.15.0 (2.5.7) 2021-02-24 aa05fda562 OpenJDK 64-Bit Server VM 11.0.9+0-adhoc..source on 11.0.9+0-adhoc..source +jit [linux-x86_64]`

* Operating system and platform (e.g. `uname -a`)
`Linux hostname 5.7.17-1rodete5-amd64 #1 SMP Debian 5.7.17-1rodete5 (2021-01-08) x86_64 GNU/Linux`

Other relevant info you may wish to add:
I’m using nix to manage my env, but others in my org who aren’t using nix are running into the same issue. 

**Expected Behavior**

We have a bit of a complex build step that uses a bash script to call a rake task, uses system to call a few ruby scripts, which uses the `system` method to launch a gradle wrapper script ultimately running a java gradle process. Under 9.2.13.0 (and 9.2.14.0), this would launch successfully, waiting for the gradle task to complete and return. 

**Actual Behavior**

However, with 9.2.15.0, the gradle process hangs in the TL process state forever. It is unresponsive to jstacks, ultimately causing the parent build process to hang. 

Seems as though the system command is launching differently from 9.2.14.0 due to this commit: https://github.com/jruby/jruby/pull/6568/commits/0cb917093ec28ce571fc5f7b0707b318531d9ca5

Was unable to reproduce with a simple target script (instead of gradle) thus far. Trying to use a custom build from @headius to validate a potential fix.

headius:
This is likely due to the introduction of `exec` into the `sh` command line we run to intercept the `chdir` before running the expected command. I do not know why this would cause it to hang versus not using exec, but a PR is coming to try removing it.

headius:
Attempting to simulate this locally... have not been able to trigger a hang outside of JRuby using a similar command structure:

```
$ sh -c 'cd lib; exec "$@"' sh gradle

> Task :help

Welcome to Gradle 6.8.3.

To run a build, run gradle <task> ...

To see a list of available tasks, run gradle tasks

To see a list of command-line options, run gradle --help

To see more detail about a task, run gradle help --task <task>

For troubleshooting, visit https://help.gradle.org

[...]

PR #6580, opened 2021-02-25
Fix system with chdir hanging

headius:
It seems like the change in #6226 (backported as #6568) to use `exec` to run the subcommand after doing an `sh` to `chdir` may be causing `system` with `chdir` to hang under some circumstances (#6579).

This PR removes the exec, which should put us back to the old behavior other than how we handle the actual command args by evaluating `"$@"`.

cc @mrnoname1000 for assistance


headius:
Might be more correct to replace `exec` with `eval` but I will confer with @mrnoname1000.

headius:
The exec was not at fault. See 6582.

PR #6582, opened 2021-02-26
Remove process group to avoid hanging daemons

### LINKED ISSUE #6579: Upgrade to 9.2.15.0 causes Kernel#system to hang.

headius:
In #5383 we added logic to create a new process group when doing
chdir in spawn, to avoid orphaning the eventual process spawned
by the intermediate sh. Unfortunately this has a side effect of
hanging on waitpid if the process spawns a long-running daemon.

This patch removes the process group and instead uses exec in the
generated sh script so that the sh process is replaced. The forked
daemon lives on, as it should, but only the main process is
waited on.

Fixes #6579

### DIFF
--- a/core/src/main/java/org/jruby/util/io/PopenExecutor.java
+++ b/core/src/main/java/org/jruby/util/io/PopenExecutor.java
@@ -125,13 +125,16 @@ public class PopenExecutor {
         if (eargp.chdirGiven) {
             // we can'd do chdir with posix_spawn, so we should be set to use_shell and now
             // just need to add chdir to the cmd
-            prog = (RubyString)prog.strDup(runtime).prepend(context, newString(runtime, "cd '" + eargp.chdir_dir + "'; "));
+            String script = "cd '" + eargp.chdir_dir + "'; ";
+
+            // use exec to eliminate extra sh process if we do not need to run command as a shell script
+            if (!searchForMetaChars(prog)) {
+                script = script + "exec ";
+            }
+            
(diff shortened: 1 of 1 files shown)


headius:
We have a winner!

headius:
No idea how to test this one.

PR #6584, opened 2021-02-27
Update to ruby/spec@37e52e5

eregon:


eregon:
Specs seem to get stuck, I'll add some flag to investigate which one.

eregon:
The failing spec is:
https://travis-ci.com/github/jruby/jruby/jobs/486719975
```
Signal.trap
- returns the previous handler
- accepts a block in place of a proc/command argument
- registers an handler doing nothing with :IGNORE


No output has been received in the last 10m0s, this potentially indicates a stalled build or something wrong with the build itself.
```
cc @headius 
I think this is actually a bug of the JVM, that after `Signal.handle(sig, SignalHandler.SIG_IGN)` all calls to `Signal.handle()` have no effect. I found the same issue in TruffleRuby and used this workaround: https://github.com/oracle/truffleruby/commit/a6e3d0b35db26816c21a5136d5f94220ea63946d

I'll skip that spec for now so you can decide how you want to handle this.

eregon:
Actually it seems this problem was already noticed for a very similar spec, and tagged in https://github.com/jruby/jruby/commit/c0d5a48947a2ad0327fc2cb70dc67af989fd6844#diff-e3cfc95cdb82e8b449320a583eae0c47dc5e9960327317f0cb203d1f3b49323d

headius:
@eregon Thanks for the SIG_IGN fix. Have you reported this to OpenJDK?

eregon:
No I did not, it was a long day and I figured I'd need a workaround for the foreseeable future anyway.
But it would still be good to report of course, would you mind to report it?

headius:
@eregon https://bugs.openjdk.java.net/browse/JDK-8262905

PR #6585, opened 2021-03-01
Kill all use of Arity+

enebo:
This branch was originally opened to solve some block problems but I quickly ran into use continuing to use Arity over Signature.

So nearly all of this is removing use of Arity from our code base (jruby-rack still uses a single Arity path in our codebase).

There are two block changes as well:

  1.  IRRuntimeHelpers.prepareProcArgs had a check in covertValueIntoArgArray asking whether it is NORMAL.  It is always a PROC so I just made that argument false.
  2. Block.prepareArgumentsForCall had extra logic to reduce the args array.  This is just extra work which later code gets no benefit from.

As for Signature code migration it is large but pretty straight-forward.  We calculate arity values for populators and we probably could do that differently in a post-Arity world but that is a topic for another day.  The perf aspect is worth repeating (from 9ccc321):

"On the performance front we will do less work (although I doubt we will
see much we can measure).  In startings up Rails server on a single
controller app we were doing 3k arity calcs and over 400k signature calcs.
Now we are doing no arity calcs and only 2k signature calcs.  With some
better mechanism for Signature creation we could get this down to a few
hundred.  The second aspect of this is reducing memory about 1Mb for the
same Rails app.  Signature is about 100k now but since it is immutable we
can reduce this down to close to 0 if we share all combos during creation."

The other thing which could/should happen post commit is to make a better mechanism for sharing the same signatures so we create less instances.  Right now a small Rails app makes 100k heap worth of Signatures but there are probably only a hundred or two hundred signatures which could exist (note: this was much larger number before this work).  That would make the heap size very small.  As it is this is about 0.1% of the heap for the Rails app.  Making it 0.01% is nice but not worth a lot of effort.


kares (approved):
:+1: nice to see the deprecated Arity param methods removed

enebo:
@kares honestly there are a ton which should be but jruby-rack popped out as a currently using Arity consumer and I wonder if there are others. So I removed a few which seemed unlikely to be used and got frightened of removing more.

We really need a nice way of auditing all the external java exts.

core/src/main/java/org/jruby/AbstractRubyMethod.java
```diff
@@ -1,4 +1,5 @@
-/***** BEGIN LICENSE BLOCK *****
+/*
+ **** BEGIN LICENSE BLOCK *****
```
headius, line 2: 😳

core/src/main/java/org/jruby/internal/runtime/methods/JavaMethod.java
```diff
@@ -1,4 +1,5 @@
-/***** BEGIN LICENSE BLOCK *****
+/*
+ **** BEGIN LICENSE BLOCK *****
```
headius, line 2: 😳

core/src/main/java/org/jruby/internal/runtime/methods/PartialDelegatingMethod.java
```diff
@@ -1,4 +1,5 @@
-/***** BEGIN LICENSE BLOCK *****
+/*
+ **** BEGIN LICENSE BLOCK *****
```
headius, line 2: 😳

headius (approved):
Nice cleanup. Make sure any unused Arity imports have been cleaned out.

PR #6586, opened 2021-03-01
Expand env-shebang to setup and pristine

headius:
The setup command is used when upgrading RubyGems, and pristine
is a clean reinstall. Neither honor the settings for "install" or
"update" so we need to force them too.

See rubygems/rubygems#4404

### DIFF
--- a/lib/ruby/stdlib/rubygems/defaults/jruby.rb
+++ b/lib/ruby/stdlib/rubygems/defaults/jruby.rb
@@ -20,7 +20,9 @@ module Gem
   def self.platform_defaults
     return {
         'install' => '--env-shebang',
-        'update' => '--env-shebang'
+        'update' => '--env-shebang',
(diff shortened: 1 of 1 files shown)


headius:
This is "probably safe" to add into 9.2.16, and fixes an open issue with RubyGems updating (rubygems/rubygems#4404) but not one that has been reported by an end user. Users running with the native executable will also not see this issue.

Ping @kares and @enebo to consider this. @deivid-rodriguez any reason you can think of this might introduce any risk?

deivid-rodriguez:
Not that I can think of, no. I'd say it's safe, specially given that you've been shipping this flag by default for `gem install` and `gem update` and there's been no issues, right?

headius:
> you've been shipping this flag by default for gem install and gem update and there's been no issues, right?

That is correct. Thanks for the input!

Issue #6589, opened 2021-03-01
RubyIO needs allocation-free read and write paths


headius:
Currently there is no way to read from or write to a RubyIO without creating a RubyString and accompanying ByteList. Meanwhile we have many places where we want to write a ByteList or even just a byte[] with a given encoding, such as IOOutputStream or the copy_stream logic for non-files.

We should build and use read and write paths that accept at least ByteList and ideally byte[], start, length, encoding to avoid allocation altogether.

This might be a good beginner to intermediate task for an external contributor.

headius:
After attempting this myself, I'd say this is more of an advanced item... even for skilled Java folks.

PR coming.

Issue #6588, opened 2021-03-01
RubyIO.getOutputStream incompatible with external encoding


headius:
The OutputStream implementation returned by RubyIO.getOutputStream always treats incoming bytes as being ASCII-8BIT encoding. This breaks writes when the underlying RubyIO has been initialized with an incompatible external encoding, like UTF-8.

https://github.com/jruby/jruby/blob/b79d640e7bfebf23a8c1a3463f589909f9dfc5bc/core/src/main/java/org/jruby/RubyIO.java#L347-L355

This led to one failure in the Psych tests as noted in https://github.com/ruby/psych/pull/481 and fixed by #6587.

I believe this stream should handle the encoding better using one of these options (perhaps?):

* Assume the bytes are encoded like the IO's internal encoding, which may be the default UTF-8 of Ruby. The bytes will be transcoded to whatever the external encoding is supposed to be.
* Assume the bytes are encoded like the IO's external encoding. They will be written directly without any transcoding (or any verification that they are in that encoding).

headius:
There's a third option: use syswrite, which does not care about bytes and will write directly to the underlying channel. This skips the question altogether and pushes the responsibility for encoding the bytes to the consumer of the OutputStream (as it would be in Java).

headius:
Punting this for exploration in 9.4. The buffered nature of these streams does appear to be necessary for some consumers. A different abstraction is necessary (possibly deprecating this one).

headius:
Unfortunately an additional change is needed in JRuby to fix this if we continue to use IOOutputStream.

IOOutputStream calls RubyIO.write and passes only the incoming bytes, offsets, and encoding:

https://github.com/jruby/jruby/blob/14ec399968bc9c1c2ff618f12be16b909d3e521a/core/src/main/java/org/jruby/util/IOOutputStream.java#L152-L163

This path was added as an optimization in https://github.com/jruby/jruby/commit/236f7ba25168af78b5b1635a828d6bc968c264d8 to avoid constructing string objects just to immediately unwrap them for IO writes.

Unfortunately it hits a bug eventually in EncodingUtils.rbByteEncode where two identical encodings will not no-op, but instead will trigger an encoding error because it rejects encoding from and to the same encoding.

The short-circuit checks here:

https://github.com/jruby/jruby/blob/844c1510f8b704925189b248873f54f49bdb569f/core/src/main/java/org/jruby/util/io/EncodingUtils.java#L983-L989

[...]

PR #6590, opened 2021-03-02
Provide direct byte[] read and write for RubyIO

headius:
This PR tracks work to address #6589 by adding allocation-free RubyIO read and write paths.

### DIFF
--- a/core/src/main/java/org/jruby/RubyIO.java
+++ b/core/src/main/java/org/jruby/RubyIO.java
@@ -1548,26 +1548,24 @@ public class RubyIO extends RubyObject implements IOEncodable, Closeable, Flusha
      * @return the count of bytes written
      */
     public int write(ThreadContext context, byte[] bytes, int start, int length, Encoding encoding, boolean nosync) {
-        Ruby runtime = context.runtime;
-        OpenFile fptr;
-        int n;
-
         if (length == 0) return 0;
 
-        fptr = getOpenFileChecked();
+        OpenFile fptr = getOpenFileChecked();
 
         boolean locked = fptr.lock();
         try {
             fptr = getOpenFileChecked();
             fptr.checkWritable(context);
 
-            n = fptr.fwrite(context, bytes, start, length, encoding, nosync);
-            if (n == -1) throw runtime.newErrnoFromErrno(fptr.errno(), fptr.getPath());
+            int n = fptr.fwrite(context, bytes, start, length, encoding, nosync);
+
+            if (n == -1) throw context.runtime.newErrnoFromErrno(fptr.errno(), fptr.getPath());
+
+            return n;
         } finally {
             if (locked) fptr.unlock();
         }
 
-        return n;
     }
 
     /**
--- a/core/src/main/java/org/jruby/RubyIO.java
+++ b/core/src/main/java/org/jruby/RubyIO.java
@@ -1573,6 +1573,64 @@ public class RubyIO extends RubyObject implements IOEncodable, Closeable, Flusha
         return n;
     }
 
+    /**
+     * Same as {@link #write(ThreadContext, byte[], int, int, Encoding, boolean)} but context will be retrieved.
+     *
+     * Heavy use should retrieve context once and call the original method to avoid thread-local overhead.
+     *
+     * @param bytes the bytes to write
+     * @param start start offset for writing
+     * @param length length to write
+     * @param encoding encoding of the bytes (will not be verified)
+     * @param nosync whether to write without syncing
+     * @return the count of bytes written
+     */
+    public int write(byte[] bytes, int start, int length, Encoding encoding, boolean nosync) {
+        return write(getRuntime().getCurrentContext(), bytes, start, length, encoding, nosync);
+    }
+
+    /**
+     * Same as {@link #write(ThreadContext, byte[], int, int, Encoding, boolean)} but context will be retrieved and
+     * nosync defaults to false.
+     *
+     * Heavy use should retrieve context once and call the original method to avoid thread-local overhead.
+     *
+     * @param bytes the bytes to write
+     * @param start start offset for writing
+     * @param length length to write
+     * @param encoding encoding of the bytes (will not be verified)
+     * @return the count of bytes written
+     */
+    public int write(byte[] bytes, int start, int length, Encoding encoding) {
+        return write(bytes, start, length, encoding, false);
+    }
+
+
+    /**
+     * Same as {@link #write(ThreadContext, byte[], int, int, Encoding, boolean)} but context will be retrieved nosync
+     * defaults to false, and the entire byte array will be written.
+     *
+     * Heavy use should retrieve context once and call the original method to avoid thread-local overhead.
+     *
+     * @param bytes the bytes to write
+     * @param encoding encoding of the bytes (will not be verified)
+     * @return the count of bytes written
+     */
+    public int write(byte[] bytes, Encoding encoding) {
+        return write(bytes, 0, bytes.length, encoding);
+    }
+
+    /**
+     * Same as {@link #write(ThreadContext, byte)} but context will be retrieved.
+     *
+     * Heavy use should retrieve context once and call the original method to avoid thread-local overhead.
+     *
+     * @param bite the byte to write, as an int
+     */
+    public void write(int bite) {
+        write(getRuntime().getCurrentContext(), (byte) bite);
+    }
+
     /** rb_io_addstr
      *
      */
--- a/core/src/main/java/org/jruby/RubyIO.java
+++ b/core/src/main/java/org/jruby/RubyIO.java
@@ -1346,7 +1346,7 @@ public class RubyIO extends RubyObject implements IOEncodable, Closeable, Flusha
             }
 
             ByteList strByteList = ((RubyString) str).getByteList();
-            n = OpenFile.writeInternal(context, fptr, fptr.fd(), strByteList.unsafeBytes(), strByteList.begin(), strByteList.getRealSize());
+            n = OpenFile.writeInternal(context, fptr, strByteList.unsafeBytes(), strByteList.begin(), strByteList.getRealSize());
 
             if (n == -1) throw runtime.newErrnoFromErrno(fptr.errno(), fptr.getPath());
         } finally {
--- a/core/src/main/java/org/jruby/RubyIO.java
+++ b/core/src/main/java/org/jruby/RubyIO.java
@@ -1422,14 +1422,23 @@ public class RubyIO extends RubyObject implements IOEncodable, Closeable, Flusha
         }
     }
 
-    /** io_write_m
+    /**
+     * Ruby method IO#write(str), equivalent to io_write_m.
      *
+     * @param context the current context
+     * @param str the string to write
      */
     @JRubyMethod(name = "write", required = 1)
     public IRubyObject write(ThreadContext context, IRubyObject str) {
         return write(context, str, false);
     }
 
+    /**
+     * Ruby method IO#write(str, ...), equivalent to io_write_m.
+     *
+     * @param context the current context
+     * @param args the strings to write
+     */
     @JRubyMethod(name = "write", rest = true)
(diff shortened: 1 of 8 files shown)


headius:
Note that due to the complexity of passing more than one "out" parameter, IO writes that require transcoding will still allocate a ByteList, but the initial wasted wrapper is not created in any case.

core/src/main/java/org/jruby/RubyIO.java
```diff
@@ -1441,12 +1450,42 @@ public IRubyObject write(ThreadContext context, IRubyObject[] args) {
...
 
-    final IRubyObject write(ThreadContext context, int ch) {
+    /**
+     * Write a single byte to this IO's write target.
+     *
+     * @param context the current context
+     * @param ch the byte to write, as an int
+     * @return the count of bytes written
+     */
+    public final IRubyObject write(ThreadContext context, int ch) {
```
kares, line 1460: :+1: would like to rather see a `public void write(int b)` (without the context for Java consumers)

core/src/main/java/org/jruby/RubyIO.java
```diff
@@ -1480,6 +1519,60 @@ public IRubyObject write(ThreadContext context, IRubyObject str, boolean nosync)
...
+     * IO).
+     *
+     * @param context the current context
+     * @param bytes the bytes to write
+     * @param start start offset for writing
+     * @param length length to write
+     * @param encoding encoding of the bytes (will not be verified)
+     * @return the count of bytes written
+     */
+    public int write(ThreadContext context,  byte[] bytes, int start, int length, Encoding encoding) {
```
kares, line 1535: do we really want context param passed around like for Ruby method impls?
we only need the runtime and from a Java ext perspective it would be nicer not to worry about context.
we could pass the runtime but I would love to also see a Java API like:
`public int write(byte[] bytes, int start, int length, Encoding encoding)`

kares (commented):
:1st_place_medal: these will be very useful.
have couple overload suggestions that would be nice to support as a Java API

core/src/main/java/org/jruby/RubyIO.java
enebo: Passing Ruby means anything which involves a call or a callsite (and this is just the ones I can think of readily) will have to getCurrentContext which would be great to never do.  Maybe still do it but only in places where we don't care about performance.  Obviously, the other case, is any TC state we want directly.

Historically we started with Runtime and it wasn't until later we realize TC made much more sense as our global handle.  I would like to talk to you about how we can change ext to TC is the default.   My actual goal is not about TC at all but to normalize and formalize most of the common accesses of a JRuby ext to static methods which can then be a dictionary of mappings to MRI cext names (or not when that doesn't make sense).

core/src/main/java/org/jruby/RubyIO.java
headius: This was actually just a protected method before, but I agree having non-context forms would be a good idea. In fact, I was thinking RubyIO could basically implement the same methods as Input/OutputStream if we decide on how to handle incoming bytes with no encoding (#6588).

core/src/main/java/org/jruby/RubyIO.java
headius: In this case the context ends up getting passed around mostly for access to the runtime, EncodingService, and metaclasses for raising errors at appropriate times, but the unavoidable use of context is to do an interruptible non-blocking write operation.

Any irreducible use of context means it should be passed in, since even after lots of work the context retrieval is a very heavy operation. See for example #5959 and https://github.com/eclipse/openj9/issues/7699.

This is not to say that I like passing it, because I totally agree it makes for a very cumbersome API, and the more we pass it, the more we avoid the threadlocal but the more we force people to pass it. I do agree that having non-context versions of these write (and eventually read) methods will be useful, and I will add them... but with lots of doco about the cost of ThreadContext retrieval and recommendations to acquire it once for hot read/write loops.

headius:
@kares Thank you for the review! Yeah I think there are more places we could be using these, including in extensions. @jonathanswenson commented on Matrix that he had attempted to modify Puma to use more direct IO calls but without this PR it could only get so far. This also came to my attention because Psych uses this stream abstraction in order to let SnakeYAML dump directly to an IO or IO-like, the former of which should be much more efficient now (and the latter may be worth special-casing common IO-likes like StringIO).

I will add the suggested API enhancements and work on adding read support today. Plus benchmarks.

headius:
With this latest push, IO implements context-free versions of the `write` methods (four overloads added with recommendations to call the context-receiving versions) and there should be zero allocation along the fast path write (with no transcoding).

headius:
Getting performance metrics on this from Ruby is difficult, because most of the key consumers (Psych, Marshal, GzipWriter) have more overhead of their own. However an allocation profile is interesting. Note the drastic reduction in byte[] and ByteList allocation overall.

Script:
```ruby
require 'tempfile'

io = Tempfile.new('bench_marshal_dump')

ary = ["foobar"] * 10000

n = 30
while n > 0
  Marshal.dump(ary, io)
  io.seek(IO::SEEK_SET, 0)
  n-=1
end
```

Before:
```
          percent          live          alloc'ed  stack class
 rank   self  accum     bytes objs     bytes  objs trace name
    1 28.47% 28.47%  14650408 366248  26851968 671287 300000 org.jruby.util.ByteList
    2 27.88% 56.35%  14346288 358645  24680088 616990 300000 org.jruby.RubyString
    3 18.23% 74.58%   9382496 367510  22223920 681015 300000 byte[]
    4  2.33% 76.91%   1198472 18274   8555048 117271 300000 char[]
    5  1.02% 77.93%    523600 16347    621648 19411 300000 java.util.concurrent.ConcurrentHashMap$Node
    6  0.85% 78.78%    437616 18213   1695936 70643 300000 java.lang.String
    7  0.67% 79.45%    344384 6956   1605896 31868 300000 java.lang.Object[]
    8  0.53% 79.98%    274680 1558   1248408 24813 300000 int[]
    9  0.44% 80.43%    227632 7098    234128  7301 300000 java.lang.ref.WeakReference
   10  0.39% 80.82%    201744 6289    501104 15644 300000 java.util.HashMap$Node
```

After:
```
          percent          live          alloc'ed  stack class
 rank   self  accum     bytes objs     bytes  objs trace name
    1  8.14%  8.14%   1192440 18217   8565008 117414 300000 char[]
    2  6.05% 14.19%    885136 13972   7830192 80777 300000 byte[]
    3  3.58% 17.77%    524016 16360    621808 19416 300000 java.util.concurrent.ConcurrentHashMap$Node
    4  3.47% 21.24%    508448 12699   2837568 70927 300000 org.jruby.util.ByteList
    5  2.99% 24.23%    437232 18197   1698336 70743 300000 java.lang.String
    6  2.32% 26.55%    339344 6936   1607816 31922 300000 java.lang.Object[]
    7  1.87% 28.42%    274440 1551   1249416 24832 300000 int[]
    8  1.55% 29.97%    227504 7094    234000  7297 300000 java.lang.ref.WeakReference
    9  1.40% 31.37%    204288 5095  24676488 616900 300000 org.jruby.RubyString
   10  1.37% 32.74%    201104 6269    500848 15636 300000 java.util.HashMap$Node
```

core/src/main/java/org/jruby/RubyIO.java
enebo: I just want to highlight another problem with using Ruby over ThreadContext since the three of us have written stuff.  I think it goes beyond whether you need it or not.  You may eventually need it.  For example, we end up adding a dyncall when originally we had not planned on having a fall back to a dyn call.  Whoops.  At this point we:

  1. use getCurrentContext()
  2. OR we deprecate current method and make a new one
  
So in or first outcome we are stuck with this extra overhead (admittedly sometimes it doesn't matter).  In the second, we end up with a method which we will keep for 5-10 years.  We have many deprecated methods we have never even removed.

I think we need to just accept TC needs to be passed by default and be very particular about when we don't.  So I think we need to find a way of liking passing TC around.

core/src/main/java/org/jruby/RubyThread.java
```diff
@@ -1606,37 +1608,78 @@ public void executeBlockingTask(BlockingTask task) throws InterruptedException {
...
+
     public void enterSleep() {
-        status.set(Status.SLEEP);
+        STATUS.set(this, Status.SLEEP);
     }
 
     public void exitSleep() {
-        if (getStatus() != Status.ABORTING) {
-            status.set(Status.RUN);
+        if (status != Status.ABORTING) {
```
enebo, line 1676: When I see the bare status field but then look down below it and see getStatus() has more logic it looks wrong at first glance.  I can see what you are doing here but I am wondering if a comment can help?

core/src/main/java/org/jruby/util/IOOutputStream.java
```diff
@@ -67,9 +67,22 @@
     public IOOutputStream(final IRubyObject io, Encoding encoding, boolean checkAppend, boolean verifyCanWrite) {
         this.io = io;
         this.runtime = io.getRuntime();
-        this.realIO = ( io instanceof RubyIO && !((RubyIO) io).isClosed() &&
-                ((RubyIO) io).isBuiltin("write") ) ?
-                    ((RubyIO) io) : null;
+
+        // If we can get a real IO from the object, we can do fast writes
+        RubyIO realIO = null;
```
enebo, line 72: Not saying do this but this section could be RubyIO realIO = determineRealIO(...); since the code block only calcs a single thing.

PS- I don't care one way or other :)

core/src/main/java/org/jruby/util/io/OpenFile.java
```diff
@@ -134,7 +136,7 @@ public void checkReopenSeek(ThreadContext context, Ruby runtime, long pos) {
 
     public EConv readconv;
     public EConv writeconv;
-    public IRubyObject writeconvAsciicompat;
+    public Encoding writeconvAsciicompat;
```
enebo, line 139: Seems very unlikely to me this will break anything but it is public.  It is for 9.3 too so I guess we can break.  Just bringing it up.

enebo (approved):
I did not compare the new methods you ported but I trust you nailed it :)

It is for 9.3 too so it does not have the same requirements as a backport (like changing a field type).

core/src/main/java/org/jruby/util/IOOutputStream.java
headius: I actually did that at one point but removed it to keep diff smaller. I think it is a fine change now, though.

core/src/main/java/org/jruby/util/io/OpenFile.java
headius: Yeah I played a little fast and loose with binary compat within OpenFile because nobody should be using it. Unfortunately it is in util.io package so these things have to be public to be accessible from RubyIO. Fingers crossed. 🤞

core/src/main/java/org/jruby/RubyThread.java
headius: Actually I think this was a mistake. I forgot getStatus is not just an accessor. Will revert.

core/src/main/java/org/jruby/RubyIO.java
```diff
@@ -1480,6 +1519,118 @@ public IRubyObject write(ThreadContext context, IRubyObject str, boolean nosync)
...
+     * @param bytes the bytes to write
+     * @param start start offset for writing
+     * @param length length to write
+     * @param encoding encoding of the bytes (will not be verified)
+     * @param nosync whether to write without syncing
+     * @return the count of bytes written
+     */
+    public int write(ThreadContext context, byte[] bytes, int start, int length, Encoding encoding, boolean nosync) {
+        Ruby runtime = context.runtime;
+        OpenFile fptr;
```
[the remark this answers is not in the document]
headius: Usually for new code I do Java style declare-at-use but some of this was just copied over from C ports. I will do a cleanup pass over some of these to get style more aligned with what we like.

PR #6592, opened 2021-03-03
Fixes to support Sorbet

headius:
See https://github.com/sorbet/sorbet/issues/1216

There will only be a handful of small fixes in here since the Java package module issues are rather tricky.


headius:
@kares See the linked bug for a minor issue in jossl (differing ancestors for one OSSL exception). You may also be interested in the linked issues surrounding our problematic package modules, which we have tried to replace without success.

headius:
I'm just going to merge this simple fix since the package module work is a project in itself, and the jossl issue should be filed there.

PR #6593, opened 2021-03-03
fix getsockopt failure

ahorek:
invastigate TypeError: size differ. expected as sizeof(int)=4 but 0 failure


headius:
Seems good to me. As discussed on Matrix there may be an underlying bug in jnr-ffi or jffi causing the IntBuffer code to fail to get the right value. We will have to investigate that separately.

ahorek:
it's a workaround, but it should at least fix the random failure. Thanks for the review @headius

Issue #4835, opened 2017-11-01
Java 9 module-info.java ruminations


raelik:
After Java 9 compatibility is made tentatively available in 9.1.14, discussions need to be had on how best to set up jRuby's module-info.java file(s) for the Java Platform Module System now available in Java 9.

headius:
We will also need to modularize the libraries we package, including but not limited to:

* All the JNR libraries
* bytelist/jcoding/joni
* utility and JVM shim libraries (invokebinder already done)

We also package several third-party libs and will want to see if they've started modularizing.

headius:
Tentatively marked for 9.2.

One big wrinkle here is that if we start including module-info in a project, we're going to need Java 9 tooling to build it. That's not a huge leap for some of the libraries, since we control them and are generally the only ones maintaining them, but JRuby itself is built by and for many different systems, some of which may not have Java 9 for a while.

raelik:
@headius Maybe just a separate build target for a modularized Java 9 deployment that includes the module-info, akin to how there is currently binary, Windows executable, source and complete jar packages of the jRuby release.

headius:
I'm trying to wrap up my day, but perhaps @raelik or someone else observing this issue can research whether there are any ways to build module-info.java without a Java 9 install present. I know we can make this produce a jar that works on Java 7 or 8, as I did in https://github.com/headius/invokedynamic.

raelik:
@headius as far as I can see, the syntax of the module-info.java file is only understood by Java 9's `javac`.

headius:
@raelik Yeah, I'm just wondering if there's a way to pull down OpenJDK 9's javac or an equivalent javac that understand module-info.java without running on 9.

headius:
We will put more focus into module support over the summer in preparation for the Java 11 LTS release.

headius:
Some of our deps have been modularized: jcodings, joni. The bytelist library has been rolled back into JRuby proper due to duplicated packages.

The jnr projects still need to be modularized.

I added an automatic module name in 681ab2210930854e7fbd8bb5d8f728f205363554.

More to come, ideally with full proper modularization in 9.2.2.

headius:
Much of this has been done. Fully modularizing JRuby and all dependencies will be a 9.3 task.

headius:
As of this writing, the following libraries either have no module info or we do not know their module status:

* org.ow2.asm:asm, -commons, -util, unknown status but recent releases most likely have module info
[...]

PR #6596, opened 2021-03-06
Remove BSF support

headius:
This removes all references to the Apache Bean Scripting Framework
except for the configurable BSF mode in our ScriptingContainer.

Relates to #4835

This has not been decided for 9.3 yet, but BSF has not had a release since 2006 and has largely been replaced by the JSR-223 APIs throughout the Java world. There may be some old legacy libraries or apps still using BSF but they are unlikely to work with newer JRuby anyway.


headius:
Windows failure seems unrelated.

headius:
BSF 3.x is still out there, but is now just a shim over javax.script. I think anyone currently using BSF with JRuby could migrate to 3.x without us shipping specific BSF support.

https://commons.apache.org/proper/commons-bsf/

enebo:
I confirm this bug

PR #6598, opened 2021-03-06
Modularize all the things

headius:
This PR will include changes necessary to fully JPMS modularize JRuby.


headius:
I have started to try to modularize JRuby itself by adding a proper module-info.java. The related commit spells out the changes, but there is one issue I have been unable to remedy...

There is a new compile-time error about the classes `Main` and `CallSite`, saying they conflict with a package of the same name. There is in each case a matching package name, but in lowercase, and this has never been an issue before.

https://gist.github.com/headius/14a50f446ab9d0f6976a231c8ec0f915

I will be unable to progress until this issue can be fixed. I have not been able to reproduce in a sample project and nobody so far has been able to explain why this breaks.

headius:
The issue that stopped me from proceeding has been reported as an OpenJDK bug. We will need to find a workaround until the patched versions have become commonplace.

headius:
https://bugs.openjdk.java.net/browse/JDK-8263432

headius:
This is very close. There are a few items that mean we have to punt the remaining work until 9.4.

* Due to [JDK-8263432](https://bugs.openjdk.java.net/browse/JDK-8263432), we have to build with Java 17. That is clearly not acceptable.
* A few libraries do not even have module names yet:
  * jzlib will have to be forked, see https://github.com/jruby/jzlib which has a WIP branch to add module name.
  * ant is an open issue. It has not been modularized by the Apache maintainers, and there does not appear to be any plan to do so.
* Most of the libraries that do have modules names are doing them automatically, which limits our ability to use tools like jlink. With more time we can get full module descriptors installed.

Of these, the first is the biggest issue. I have commented on the JDK issue asking for a workaround but have not year heard back.

headius:
Given the remaining issues and the challenge of fixing them while still supporting traditional non-modularized Java, I'm punting this to the next major release. That version will commit to a minimum Java of at least 11 and possibly 17, making the module changes both necessary and easier to deal with.

Issue #6504, opened 2020-12-22
SocketError when sending Unix socket datagrams


mohamedhafez:
when I run the simple script below, it works in MRI, but raises `SocketError (No message available)` in JRuby 9.2.14.0 on Ubuntu 18.04. It would be awesome to get this working so I could have my JRuby daemons communicate with systemd

```ruby
Addrinfo.unix("/run/systemd/notify", :DGRAM).connect do |s|
  s.close_on_exec = true
  s.write("READY=1")
end
```

here's the stack trace:
```
Traceback (most recent call last):
        9: from /home/web/.rvm/rubies/jruby-9.2.14.0/bin/irb:13:in `<main>'
        8: from org/jruby/RubyKernel.java:1189:in `catch'
        7: from org/jruby/RubyKernel.java:1189:in `catch'
        6: from org/jruby/RubyKernel.java:1442:in `loop'
        5: from org/jruby/RubyKernel.java:1048:in `eval'
        4: from (irb):2:in `evaluate'
        3: from /home/web/.rvm/rubies/jruby-9.2.14.0/lib/ruby/stdlib/socket.rb:139:in `connect'
        2: from /home/web/.rvm/rubies/jruby-9.2.14.0/lib/ruby/stdlib/socket.rb:66:in `connect_internal'
        1: from org/jruby/ext/socket/RubySocket.java:215:in `connect'
SocketError (No message available)
```

headius:
Could you come up with a reproduction that does not depend on systemd?

This may work better on master (9.3) which has better support for Addrinfo.

mohamedhafez:
I tried using master, but got the same error, gives this stack trace:

```
Traceback (most recent call last):
       11: from /home/web/.rvm/rubies/jruby-head/bin/irb:23:in `<main>'
       10: from org/jruby/RubyKernel.java:1049:in `load'
        9: from /home/web/.rvm/rubies/jruby-head/lib/ruby/gems/shared/gems/irb-1.0.0/exe/irb:11:in `<main>'
        8: from org/jruby/RubyKernel.java:1234:in `catch'
        7: from org/jruby/RubyKernel.java:1234:in `catch'
        6: from org/jruby/RubyKernel.java:1504:in `loop'
        5: from org/jruby/RubyKernel.java:1088:in `eval'
        4: from (irb):3:in `evaluate'
        3: from /home/web/.rvm/rubies/jruby-head/lib/ruby/stdlib/socket.rb:139:in `connect'
[...]

PR #6603, opened 2021-03-08
UNIX datagram socket support

headius:
Unix datagram sockets have never been properly wired up. In addition, datagram send and receive have not been hooked up for RubySocket, which makes it impossible to construct and use a UNIX DGRAM socket via the Socket class. This PR will fix all that for #6504.


headius:
More work required here than we are comfortable with in 9.2.17.0. Bumping to 9.3.

PR #6612, opened 2021-03-11
Return only inflated content from Zlib.inflate

### LINKED ISSUE #6606: zlib decompression produces wrong output
**Environment Information**
  
Provide at least:
* JRuby version (`jruby -v`) and command line: `jruby 9.2.16.0 (2.5.7) 2021-03-03 f82228dc32 OpenJDK 64-Bit Server VM 25.275-b01 on 1.8.0_275-b01 +jit [linux-x86_64]`
* Operating system and platform (e.g. `uname -a`): `Linux noblesse 5.4.80-gentoo-r1 #1 SMP Tue Dec 8 10:34:59 MSK 2020 x86_64 AMD Ryzen 7 3700X 8-Core Processor AuthenticAMD GNU/Linux`

**Way to reproduce**

1. Download attached [data.gz](https://github.com/jruby/jruby/files/6115781/data.gz)
2. Run `ruby -e "require 'zlib'; f = File.read('data.gz'); puts Zlib::Inflate.inflate(f).bytesize"` in the directory where you saved `data.gz` with different Ruby implementations

**Expected Behavior**

Uncompressed data size is **60992** bytes. This is what MRI says and what I actually expect.

`$ zlib-flate -uncompress < data.gz > data && ls -l data` agrees that uncompressed data is 60992 bytes.

`$ openssl zlib -d < data.gz > data && ls -l data` also thinks that uncompressed data is 60992 bytes.

`$ python3 -c "import zlib; print(len(zlib.decompress(open('data.gz', 'rb').read())))"` also thinks that uncompressed data is 60992 bytes.

Java also agrees data is 60992 bytes:

```java
  public static void main(String[] args) throws Exception {
    byte[] input = Files.readAllBytes(Path.of("data.gz"));
    Inflater inflater = new Inflater();
    inflater.setInput(input);

    ByteArrayOutputStream output = new ByteArrayOutputStream();
    byte[] buffer = new byte[1024];

    while (!inflater.finished()) {
      int len = inflater.inflate(buffer);
      output.write(buffer, 0, len);
    }
    System.out.println(output.toByteArray().length); // prints 60992
  }
```

**Actual Behavior**

Uncompressed data size is **60994** bytes on JRuby. I observe this on JRuby 9.2.13 and 9.2.16. I didn't test other JRuby versions.

headius:
Fixes #6606

Need to investigate the surrounding logic, since it no longer matches CRuby and this patch is a bit of a best guess at what is intended. The basic issue was that we were appending extra bytes on the input to the decompressed output. CRuby, on the other hand, just ensures the input buffer contains only those extra bytes, allowing you to pass the same input to inflate repeatedly to get multiple compressed chunks out of it.

We do not support this behavior, and the related test does not pass (this PR neither fixes nor breaks any CRuby zlib tests, as far as I could see).

### DIFF
--- a/core/src/main/java/org/jruby/ext/zlib/JZlibInflate.java
+++ b/core/src/main/java/org/jruby/ext/zlib/JZlibInflate.java
@@ -67,14 +67,12 @@ public class JZlibInflate extends ZStream {
         JZlibInflate inflate = (JZlibInflate) klass.allocate();
         inflate.init(JZlib.DEF_WBITS);
 
-        IRubyObject result;
         try {
-            inflate.append(string.convertToString().getByteList());
+            return inflate.inflate(context, string, Block.NULL_BLOCK);
         } finally {
-            result = inflate.finish(context, Block.NULL_BLOCK);
+            inflate.finish(context, Block.NULL_BLOCK);
             inflate.close();
         }
-        return result;
     }
 
     @JRubyMethod(name = "initialize", optional = 1, visibility = PRIVATE)
--- a/spec/tags/ruby/library/zlib/deflate/deflate_tags.txt
+++ b/spec/tags/ruby/library/zlib/deflate/deflate_tags.txt
@@ -1,7 +1,5 @@
 fails:Zlib::Deflate.deflate deflates chunked data
 fails:Zlib::Deflate#deflate without break deflates chunked data
 fails:Zlib::Deflate#deflate without break deflates chunked data with final chunk
-fails:Zlib::Deflate#deflate without break deflates chunked data without errors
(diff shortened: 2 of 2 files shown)


headius:
Need to dig deeper on this... several specs failed.

headius:
The new patch is simpler and just returns the inflated content. I believe this matches the MRI behavior better than before. The `finish` method should still return the post-deflate data from input, but for the class-level `Zlib.inflate` and `Zlib::Inflate.inflate` we only return the decompressed content.

slonopotamus:
Confirming that `jruby-9.2` branch properly passes initial testcase from this issue.

PR #6615, opened 2021-03-12
Detect modules more robustly

### LINKED ISSUE #6608: Java::JavaLang::NoClassDefFoundError (java/sql/Date)
**Environment Information**
  
Provide at least:
* JRuby version: 9.2.13.0 & 9.2.16.0
```
Picked up JAVA_TOOL_OPTIONS: -Djava.library.path=/nix/store/9l06v7fc38c1x3r2iydl15ksgz0ysb82-glibc-2.32/lib
jruby 9.2.13.0 (2.5.7) 2020-08-03 9a89c94bcc OpenJDK 64-Bit Server VM 11.0.9+0-adhoc..source on 11.0.9+0-adhoc..source +jit [linux-x86_64]
```
* Operating system and platform: `Linux fmzakari-glaptop 5.7.17-1rodete5-amd64 #1 SMP Debian 5.7.17-1rodete5 (2021-01-08) x86_64 GNU/Linux`

**Expected Behavior**

Attempting to call `to_java` on a _RubyTime_ object and have it return a _java.sql.Date_ object returns _NoClassDefFoundError_.

What is perculiar is that I can clearly create a  _java.sql.Date_, so it seems to be accessible via my classpath.

Here is a minimal reproduction with _irb_:
```
# let us prove first that java.sql.Date works!
irb(main):013:0> date = java.sql.Date.new(0)
=> #<Java::JavaSql::Date:0x2715644a>

# now fetch the class
irb(main):014:0> clazz = java.sql.Date.java_class
=> class java.sql.Date

# lets create a RubyTime object now
irb(main):015:0> t = Time.new(0)
=> 0000-01-01 00:00:00 -0752

# lets try to cast it to java.sql.Date
irb(main):016:0> t.to_java(clazz)
       16: from org.jruby.ir.interpreter.InterpreterEngine.processCall(InterpreterEngine.java:361)
       15: from org.jruby.ir.instructions.CallBase.interpret(CallBase.java:549)
       14: from org.jruby.runtime.callsite.CachingCallSite.call(CachingCallSite.java:84)
       13: from org.jruby.RubyKernel$INVOKER$s$0$3$eval.call(RubyKernel$INVOKER$s$0$3$eval.gen)
       12: from org.jruby.RubyKernel.eval(RubyKernel.java:1048)
       11: from org.jruby.RubyKernel.evalCommon(RubyKernel.java:1086)
       10: from org.jruby.ir.interpreter.Interpreter.evalWithBinding(Interpreter.java:182)
        9: from org.jruby.ir.interpreter.Interpreter.evalCommon(Interpreter.java:158)
        8: from org.jruby.ir.interpreter.Interpreter.INTERPRET_EVAL(Interpreter.java:106)
        7: from org.jruby.ir.interpreter.StartupInterpreterEngine.interpret(StartupInterpreterEngine.java:72)
        6: from org.jruby.ir.interpreter.InterpreterEngine.processCall(InterpreterEngine.java:316)
        5: from org.jruby.runtime.callsite.CachingCallSite.call(CachingCallSite.java:174)
        4: from org.jruby.runtime.callsite.CachingCallSite.cacheAndCall(CachingCallSite.java:375)
        3: from org.jruby.java.addons.KernelJavaAddons$INVOKER$s$to_java.call(KernelJavaAddons$INVOKER$s$to_java.gen)
        2: from org.jruby.java.addons.KernelJavaAddons.to_java(KernelJavaAddons.java:29)
        1: from org.jruby.RubyTime.toJava(RubyTime.java:1472)
Java::JavaLang::NoClassDefFoundError (java/sql/Date)
```

A bit of a headscratcher for me since right before the call to _to_java_, you can see it working.

**Actual Behavior**

I expect the cast to work.
If I change the JDK we are using to **JDK8**, I see that the results work as intended.
```
irb(main):001:0> clazz = java.sql.Date.java_class
=> class java.sql.Date
irb(main):002:0>  t = Time.new(0)
=> 0000-01-01 00:00:00 -0752
irb(main):003:0> t.to_java(clazz)
=> #<Java::JavaSql::Date:0x5223e5ee>
```

There is something funky here going on with module-loading.

headius:
Detecting module support based on the existence of the jmods
directory was fragile for at least two reasons:

* The jmods directory is only present in some builds of the JDK
  (not JRE) that can be used to jlink new runtime environments.
* Some distributions of the JDK may omit or relocate this
  directory, such as Nix's version (#6608).

Instead we detect modules using two other mechanisms that should
reliably work on all modularized Java runtimes:

* Check for JAVA_HOME/lib/modules, generated by the jlink process.
* Check for a JAVA_HOME/release file with a MODULES entry.

See https://stackoverflow.com/questions/66601929 for a summary of
detection mechanism.

Fixes #6608.

jruby-launcher will need a matching update.

### DIFF
--- a/bin/jruby.bash
+++ b/bin/jruby.bash
@@ -54,7 +54,7 @@ process_java_opts() {
 
     # On Java 9+, add an @argument for the given file.
     # On earlier versions the file contents will be read and expanded on the Java command line.
-    if [ "$is_java9" ]; then
+    if [ "$use_modules" ]; then
       java_opts_from_files="$java_opts_from_files @$java_opts_file"
     else
       java_opts_from_files="$java_opts_from_files $(cat "$java_opts_file")"
@@ -140,17 +140,17 @@ else
   fi
 fi
 
-# Detect Java 9+ by the presence of a jmods directory in JAVA_HOME
-if [ -d "$JAVA_HOME"/jmods ]; then
-  is_java9=1
+# Detect modularized Java
+if [ -f ${JAVA_HOME}/lib/modules ] || [ -f ${JAVA_HOME}/release ] && grep -q ^MODULES ${JAVA_HOME}/release; then
+  use_modules=1
 fi
 
 add_log "  JAVACMD: $JAVACMD"
 add_log "  JAVA_HOME: $JAVA_HOME"
 
-if [ "$is_java9" ]; then
+if [ "$use_modules" ]; then
   add_log
-  add_log "Detected Java modules at $JAVA_HOME/jmods"
+  add_log "Detected Java modules at $JAVA_HOME"
 fi
 
 # ----- Process .java_opts files ----------------------------------------------
@@ -384,7 +384,7 @@ fi
 
 # ----- Module and Class Data Sharing flags for Java 9+ -----------------------
 
-if [ "$is_java9" ]; then
+if [ "$use_modules" ]; then
   # Use module path instead of classpath for the jruby libs
   classpath_args=(--module-path "$JRUBY_CP" -classpath "$CP$CP_DELIMITER$CLASSPATH")
 


fzakaria:
Nix's distribution does have these two mechanisms present.
```
❯ ls -l /nix/store/2xv65447vkq09bpx1mfci1cx2vdvp153-openjdk-headless-11.0.9+11/lib/openjdk/lib/modules
.r--r--r-- 141M fmzakari 31 Dec  1969 /nix/store/2xv65447vkq09bpx1mfci1cx2vdvp153-openjdk-headless-11.0.9+11/lib/openjdk/lib/modules
❯ ls -l /nix/store/2xv65447vkq09bpx1mfci1cx2vdvp153-openjdk-headless-11.0.9+11/lib/openjdk/release
.r--r--r-- 1.2k fmzakari 31 Dec  1969 /nix/store/2xv65447vkq09bpx1mfci1cx2vdvp153-openjdk-headless-11.0.9+11/lib/openjdk/release
```

The challenge was that the JAVA_HOME set in the wrapped script was wrong :(
Java itself seems to know it's own home directory, which I feel like should have been used.
(is there any point then of even having JAVA_HOME set?)

```bash
❯ JAVA_HOME=~/ java -XshowSettings:properties -version 2>&1 | grep java.home

    java.home = /nix/store/2xv65447vkq09bpx1mfci1cx2vdvp153-openjdk-headless-11.0.9+11/lib/openjdk
```

fzakaria:
Here is a better example to demonstrate:
```bash
❯ echo $JAVA_HOME
/home/fmzakari/
❯ java -XshowSettings:properties -version 2>&1 | grep java.home

    java.home = /nix/store/2xv65447vkq09bpx1mfci1cx2vdvp153-openjdk-headless-11.0.9+11/lib/openjdk
```

Seems like having the user not even have to set JAVA_HOME is even better!

headius:
The problem is that we need to pass module flags to the `java` command, so we need to determine whether modules are present before we start up. If we have to run `java` twice – once to determine if it supports modules and once to launch – we add quite a lot of time to the startup process:

```
$ time java -XshowSettings:properties -version
Property settings:
    awt.toolkit = sun.lwawt.macosx.LWCToolkit
    file.encoding = UTF-8
...
    user.name = headius
    user.timezone = 

openjdk version "11.0.4" 2019-07-16
OpenJDK Runtime Environment AdoptOpenJDK (build 11.0.4+11)
OpenJDK 64-Bit Server VM AdoptOpenJDK (build 11.0.4+11, mixed mode)

real	0m0.115s
user	0m0.116s
sys	0m0.029s
```

We really need to be able to tell whether the current Java version supports modules without running any commands (or if we run anything it has to be very fast).

We have various mechanisms to figure out the JAVA_HOME if none is set, but they depend on the `java` command resolving to something inside a typical Java distribution (i.e. in a bin dir where ../lib/modules is present).

I see your PR for Nix was accepted...shouldn't that combined with #6615 resolve this?

fzakaria:
Oh totally, I mean it's resolved without this even since the Nix distribution does have the _jmod_ directory.
I was just enjoying the discussion and giving feedback! :)

I forgot that a main goal is to reduce startup time. :+1:

headius:
Modified the check to first confirm release file is there, to avoid grep error. I think this is ready to merge.

bin/jruby.bash
```diff
@@ -140,17 +140,17 @@ else
   fi
 fi
 
-# Detect Java 9+ by the presence of a jmods directory in JAVA_HOME
-if [ -d "$JAVA_HOME"/jmods ]; then
-  is_java9=1
+# Detect modularized Java
+if [ -f ${JAVA_HOME}/lib/modules ] || [ -f ${JAVA_HOME}/release ] && grep -q ^MODULES ${JAVA_HOME}/release; then
```
walro, line 144: Hi @headius, was the `"` drop around `$JAVA_HOME` intentional here? Asking, because I am seeing some errors in 9.2.17.0 which seems to be related to me having spaces in my `$JAVA_HOME`:

```
$ ruby --version
/Users/robin/.rubies/jruby-9.2.17.0/bin/ruby: line 144: [: /Library/Internet: binary operator expected
/Users/robin/.rubies/jruby-9.2.17.0/bin/ruby: line 144: [: /Library/Internet: binary operator expected
jruby 9.2.17.0 (2.5.8) 2021-03-29 84d363da97 Java HotSpot(TM) 64-Bit Server VM 25.281-b09 on 1.8.0_281-b09 +jit [darwin-x86_64]

$ unset JAVA_HOME

$ ruby --version 
jruby 9.2.17.0 (2.5.8) 2021-03-29 84d363da97 OpenJDK 64-Bit Server VM 15.0.1+9-18 on 15.0.1+9-18 +jit [darwin-x86_64]
```

headius: It was not intentional and I even thought at one point I need to restore those quotes. You can make the modification yourself locally, or use the native launcher from jruby-launcher gem which should not have this issue. I will open a PR to fix this in JRuby proper.

headius: FWIW we do recommend not having the JDK installed in a path with spaces for many other reasons, including the complexity of making sure re-launched Java/JRuby subprocesses use the proper quoting. Thank you for bringing this issue to my attention again, I have pushed https://github.com/jruby/jruby/pull/6638 to fix it.

headius: @walro Could you share more information about your system in the PR? For me, JDKs are installed in `/Library/Java/JavaVirtualMachines` so this `/Library/Internet Something` path may be a new thing we have to be prepared for.

walro: Absolutely, done so!

PR #6620, opened 2021-03-16
Remove generated pom.xml files

headius:
These files are both generated and versioned, which forces us to
always generate and commit them with pom.rb changes. Since
polyglot Maven has been supported fornearly a decade, we now
remove these unnecessary files.


headius:
There unknown risk in this and it serves no purpose other than reducing churn in the pom.xml files, so I have punted further down the road.

headius:
IntelliJ does not work well without the poms, but also gets confused when we have both. The path forward is not clear but simply deleting the poms does not appear to be the right way. I'm closing this and we'll try to figure out a modern approach (including the possibility of dropping polyglot Ruby poms).

PR #6621, opened 2021-03-16
Bind compiled methods using lambdas

headius:
This change modifies CompiledIRMethod to use a lambda-implemented
interface rather than calling through MethodHandle when the target
method handles are provided with a Lookup that can see the
compiled method class. This improves stack utilization and call
performance.

This mechanism currently only works for "def" that occurs in
compiled bytecode, since the compiled class is not visible to the
LambdaMetaFactory otherwise. Jitted methods will need to provide
their Lookup when being bound into a MixedModeIRMethod in order to
duplicate this logic.

See #3810 for the initial motivation for this work.


headius:
One way to support this mechanism in both CompiledIRMethod and MixedModeIRMethod (and similar work could be done for blocks) would be to have JIT-compiled method classes provide a factory method that can be used to construct the interface impl, keeping the lookup local to the class and uniformly constructing these interface impls using LambdaMetaFactory at a bytecode level.

headius:
I have not been able to show any substantial gains from this, plus getting LambdaMetaFactory to work for individual jitted methods appears to be rather tricky (the lookup passed in has to have private access to the method handle's class, which means we basically need to get a lookup out of the jitted method class) so I am ditching this effort for now. We will revisit this and other method binding performance issues in the next big release after 9.3.

Issue #6628, opened 2021-03-24
#dig slower with invokedynamic than without


chrisseaton:
This benchmark is _slower_ with indy. Seems to scale in the same way though, and still faster than MRI.

```ruby
require 'benchmark/ips'

puts RUBY_DESCRIPTION

array = [[[[[[[[[[[14]]]]]]]]]]]

Benchmark.ips do |x|
  x.report('dig-01') { array.dig(0) }
  x.report('dig-02') { array.dig(0, 0) }
  x.report('dig-03') { array.dig(0, 0, 0) }
  x.report('dig-04') { array.dig(0, 0, 0, 0) }
  x.report('dig-05') { array.dig(0, 0, 0, 0, 0) } # easy to find examples this long
  x.report('dig-06') { array.dig(0, 0, 0, 0, 0, 0) }
  x.report('dig-07') { array.dig(0, 0, 0, 0, 0, 0, 0) }
  x.report('dig-08') { array.dig(0, 0, 0, 0, 0, 0, 0, 0) } # possible to find examples this long
  x.report('dig-09') { array.dig(0, 0, 0, 0, 0, 0, 0, 0, 0) }
  x.report('dig-10') { array.dig(0, 0, 0, 0, 0, 0, 0, 0, 0, 0) }
  x.compare!
end
```

```
chrisseaton@Chris-Seatons-MacBook-Pro truffleruby % ruby dig-bench.rb        
jruby 9.2.16.0 (2.5.7) 2021-03-03 f82228dc32 OpenJDK 64-Bit Server VM 25.252-b14 on 1.8.0_252-b14 +jit [darwin-x86_64]
Warming up --------------------------------------
              dig-01   724.385k i/100ms
              dig-02   697.029k i/100ms
              dig-03   655.153k i/100ms
              dig-04   654.015k i/100ms
              dig-05   608.623k i/100ms
              dig-06   561.663k i/100ms
              dig-07   540.005k i/100ms
              dig-08   566.598k i/100ms
              dig-09   511.715k i/100ms
              dig-10   492.250k i/100ms
Calculating -------------------------------------
              dig-01     21.978M (±28.7%) i/s -     88.375M in   5.003148s
              dig-02     22.914M (±27.4%) i/s -     92.705M in   5.006835s
              dig-03     16.141M (±29.2%) i/s -     67.481M in   5.007251s
              dig-04     15.811M (±27.6%) i/s -     67.364M in   5.021521s
[...]

PR #6629, opened 2021-03-24
Unbreak CachingCallSite.isBuiltin

headius:
During the various deprecations and undeprecations of this method,
it was inadvertently broken due to switching to a cacheAndGet path
that does not set the builtinCache field. Since call sites used
for isBuiltin are usually *only* used for isBuiltin, I modified
this to just use the standard cache field.

Relates to #6628

### DIFF
--- a/core/src/main/java/org/jruby/runtime/callsite/CachingCallSite.java
+++ b/core/src/main/java/org/jruby/runtime/callsite/CachingCallSite.java
@@ -17,6 +17,7 @@ import static org.jruby.RubyBasicObject.getMetaClass;
 public abstract class CachingCallSite extends CallSite {
 
     protected CacheEntry cache = CacheEntry.NULL_CACHE;
+    @Deprecated
     protected CacheEntry builtinCache = CacheEntry.NULL_CACHE;
 
     public CachingCallSite(String methodName, CallType callType) {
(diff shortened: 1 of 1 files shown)


headius:
@kares Need your input here... you deprecated these some time ago, I believe when you were improving how we check isBuiltin for Fixnum and Float. Be my second pair of eyes and help me confirm this fix is ok and appropriate?

headius:
Note that `dig` does not use this version of isBuiltin, so I am not sure it is the cause of the caching issues seen in #6628.

headius:
This change likely has little to no effect on #6628 but is still valid. We need to clean up all these paths and either delete them or make them solid.

kares:
believe the deprecation was to prefer the public methods without the additional `String methodName` argument,
e.g. `retrieveCache(RubyClass selfType)` instead of `retrieveCache(RubyClass selfType, String methodName)`

a bit confused why 2 caches would be needed here, the 'new' `builtinCache` seems to be coming from 
https://github.com/jruby/jruby/commit/660b6d852abd6abf98d7a419a33c8a58db2e5c7e#diff-00eb72e55361b6fc7fdd5f21e190c18f9c43a66620abf58a1629ef219d78b7be 
... would rather drop it's usage completely since with this change it's no longer read

PR #6638, opened 2021-03-30
Restore quotes around module detection JAVA_HOME

headius:
See https://github.com/jruby/jruby/pull/6615#pullrequestreview-623982563

The quoting around `$JAVA_HOME` was inadvertently removed while fixing module detection in #6615. The review linked above points out that this causes problems when the JDK is installed in a path with spaces. This is not common on unixy environments, and the bash script is not used on Windows where it is more common, but the regression is real.

This PR restores the quoting around `$JAVA_HOME` to fix the issue.

### DIFF
--- a/bin/jruby.bash
+++ b/bin/jruby.bash
@@ -141,7 +141,7 @@ else
 fi
 
 # Detect modularized Java
-if [ -f ${JAVA_HOME}/lib/modules ] || [ -f ${JAVA_HOME}/release ] && grep -q ^MODULES ${JAVA_HOME}/release; then
+if [ -f "$JAVA_HOME"/lib/modules ] || [ -f "$JAVA_HOME"/release ] && grep -q ^MODULES "$JAVA_HOME"/release; then
   use_modules=1
 fi
 


walro:
Thanks!

I noticed this since I, for reasons I do not remember now, had my `$JAVA_HOME` pointing at `/Library/Internet Plug-Ins/JavaAppletPlugin.plugin/Contents/Home`. It's the "JRE path" of Oracle's Java installation, see for example:

* https://docs.oracle.com/javase/8/docs/technotes/guides/install/mac_jre.html#A1097025
* https://docs.oracle.com/javase/9/install/installation-jdk-and-jre-macos.htm#JSJIG-GUID-5F4A0659-BFC5-4CB9-9920-D2DEABF29894
* https://docs.oracle.com/javase/10/install/installed-directory-structure-jdk-and-jre.htm#JSJIG-GUID-F7178F2F-DC92-47E9-8062-CA6B2612D350

headius:
@walro Aha that explains it! We should run fine with a JRE so that is no problem, but I did not realize that the Oracle JRE gets installed in a path with spaces (and referencing the long-deprecated Applet feature, no less!)

I will look into adding a test for this situation, at least when on MacOS and such a path is available.

PR #6639, opened 2021-03-30
Resolve symlinked java command to find home

### LINKED ISSUE #6637: Fix issue with JAVA_HOME not being set
See https://github.com/jruby/jruby-launcher/issues/34
**Environment Information**
  
Provide at least:
* JDK 8
* Java located in /usr/bin/java
* CentOS with /usr/lib/modules in existence

**Expected Behavior**

No --module-path being used.

**Actual Behavior**

--module-path being used

Line in question: https://github.com/headius/jruby/blob/ebf92b39391f04f06ebd6e3c54878f5a4bba4632/bin/jruby.bash#L139

Comments from jRuby launcher: https://github.com/jruby/jruby-launcher/commit/88e3c9613b63f1d4ac1db2a1730f2887422f7b17#r48827667

headius:
This PR fixes #6637 by fully resolving any symlinks between the `java` command (found from `$JAVACMD` or `java` in PATH) before using that command's path to determine `$JAVA_HOME`.

The issue affects systems without `$JAVA_HOME` set, but that have a `java` command in PATH that resolves to a `$JAVA_HOME/bin/java` through a series of symlinks (e.g. CentOS).

This PR also exports `JRUBY_HOME`, `JAVACMD`, and `JAVA_HOME` so they will be seen by the JRuby process and subprocesses, avoiding recalculating this home or accidentally using a different JDK for additional JRuby launches.

I have added a test case for the symlink situation.

### DIFF
--- a/bin/jruby.bash
+++ b/bin/jruby.bash
@@ -84,7 +84,7 @@ resolve_symlinks() {
 BASE_DIR="$(cd -P -- "$(dirname -- "$BASH_SOURCE")" >/dev/null && pwd -P)"
 SELF_PATH="$(resolve_symlinks "$BASE_DIR/$(basename -- "$BASH_SOURCE")")"
 
-JRUBY_HOME="${SELF_PATH%/*/*}"
+export JRUBY_HOME="${SELF_PATH%/*/*}"
 
 # ----- File paths for various options and files we'll process later ----------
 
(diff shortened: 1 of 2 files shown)


headius:
First attempt at a test failed, indicating that the JDK installs on Ubuntu also symlink `<jdk home>/bin/java` to `<jdk home>/jre/bin/java` so fully resolving this path may potentially walk out of the JDK into a JRE root. I am unsure if this would be a problem in practice... we do not require the JDK and the JRE concept is going away.

headius:
Additional improvements thanks to @mrnoname1000:

* Use variable for return from function `resolve_symlinks` to avoid forking a subprocess.
* export vars separately from assignment, since that will mask failures if the variable's RHS command fails.

headius:
From de400b609115007f01cb21ce9a7a305d2d7ea1c3:

> Remove export of key discovered paths
> 
> I decided I am not comfortable with this change, since it would
> mean launching a subprocess JRuby with a different home or Java
> would require *clearing* these variables rather than just pointing
> at a new jruby or java executable.

PR #6644, opened 2021-03-31
Reduce backref usage in split

### LINKED ISSUE #6160: interface proxy generation is race-y
```ruby
class Runner
  def run; end
end

10.times do |i|
  runner = Runner.new

  3.times.map do
    Thread.start { p runner.to_java(java.lang.Runnable) }
  end.each(&:join)
end

```

the script will often fail, trying to generate the Runnable proxy impl : 
```
#<Java::OrgJrubyGen::InterfaceImpl1637004621:0x265773ef>
#<Java::OrgJrubyGen::InterfaceImpl1637004621:0x7fe0a0f9>
#<Java::OrgJrubyGen::InterfaceImpl1637004621:0x42a06b46>
#<Java::OrgJrubyGen::InterfaceImpl1637004621:0x59d5eadd>
warning: thread "Ruby-0-Thread-5: logger_race.rb:14" terminated with exception (report_on_exception is true):
#<Java::OrgJrubyGen::InterfaceImpl1637004621:0x59d5eadd>
NoMethodError: undefined method `__jcreate_meta!' for #<Runner:0x7b1a2825>
         to_java at org/jruby/java/addons/KernelJavaAddons.java:29
  race.rb at logger_race.rb:22
NoMethodError: undefined method `__jcreate_meta!' for #<Runner:0x7b1a2825>
         to_java at org/jruby/java/addons/KernelJavaAddons.java:29
  race.rb at logger_race.rb:22
```

headius:
String#split appears to do nothing with the context backref slot
except using it to receive the MatchData from a regexp match and
eventually clearing it to nil. This patch sets up a thread-local
IRubyObject[] holder to use for that out value and avoids using
the context backref at all except for clearing it at the end.

This fixes #6160 by avoiding all reads of the context backref. It
may still cause problems for other backref-reading functions that
run concurrently and see the clearing to nil unexpectedly.

### DIFF
--- a/core/src/main/java/org/jruby/RubyRegexp.java
+++ b/core/src/main/java/org/jruby/RubyRegexp.java
@@ -86,6 +86,8 @@ public class RubyRegexp extends RubyObject implements ReOptions, EncodingCapable
     private ByteList str = ByteList.EMPTY_BYTELIST;
     private RegexpOptions options;
 
+    private static final ThreadLocal<IRubyObject[]> TL_HOLDER = ThreadLocal.withInitial(() -> new IRubyObject[1]);
+
     public static final int ARG_ENCODING_FIXED     =   ReOptions.RE_FIXED;
     public static final int ARG_ENCODING_NONE      =   ReOptions.RE_NONE;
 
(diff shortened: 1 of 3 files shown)


headius:
There are likely other methods that claim to read or write backref but actually only use it as an extra return value from Regexp functions.

PR #6649, opened 2021-04-02
Compatibility fixes for open3 on Windows

headius:
This is a series of fixes to improve the behavior of the `open3` library when run on platforms where we cannot support native process launching (currently just Windows).

The bulk of changes here rewrite the non-native open3 shim to use java.lang.ProcessBuilder. Environment hash and options hash are also better supported.

PR #6650, opened 2021-04-06
Unsupported process waits warn and noop

headius:
When running without native support, we cannot do any of the pid-
based waits. However in many instances it may be more damaging to
raise a hard error for the unimplemented logic rather than just
warning and doing nothing. This commit changes the logic of these
methods to report not implemented and warn rather than failing
outright. The methods are also marked as not implemented in case
code checks this (though I have no such examples).


headius:
The commit in this PR was originally created in service of #6649, due to the tests attempting to use waitpid, but it turned out the tests had several other places they needed real native IO support. I push this as a PR here to discuss whether this would be better behavior for these methods when we cannot implement them.

PR #6653, opened 2021-04-07
Include overriding classes in backtrace mapping

### LINKED ISSUE #5200: Core method call missing in exception backtrace on JRuby 9.2.0.0
I noticed some very minor issue with the exception backtraces in JRuby 9.2.0.0: there is a core method call missing in the output of this script:

```ruby
puts RUBY_DESCRIPTION

def make_error
  1 / 0
end

begin
  make_error
rescue => e
  puts e
  puts e.backtrace
end
```

See _Actual Behavior_ below for the program output when running with JRuby 9.2.0.0.

Normal Ruby exceptions (via `raise`) show up correctly, and so do pure Java exceptions.

### Environment

`jruby -v`:
```
jruby 9.2.0.0 (2.5.0) 2018-05-24 81156a8 Java HotSpot(TM) 64-Bit Server VM 25.121-b13 on 1.8.0_121-b13 +jit [darwin-x86_64]
```

No `JRUBY_OPTS` set

OS:
```
Darwin MS-DOS6.local 16.7.0 Darwin Kernel Version 16.7.0: Thu Jan 11 22:59:40 PST 2018; root:xnu-3789.73.8~1/RELEASE_X86_64 x86_64
```

### Expected Behavior

Same as in JRuby 9.1.17.0 -- there should be a call to the `/` method somewhere in the backtrace:

```
jruby 9.1.17.0 (2.3.3) 2018-04-20 d8b1ff9 Java HotSpot(TM) 64-Bit Server VM 25.121-b13 on 1.8.0_121-b13 +jit [darwin-x86_64]
divided by 0
org/jruby/RubyFixnum.java:564:in `/'
./thing.rb:4:in `make_error'
./thing.rb:8:in `<main>'
```

For comparison, here's the same script run with MRI 2.5.0:

```
ruby 2.5.0p0 (2017-12-25 revision 61468) [x86_64-darwin16]
divided by 0
./thing.rb:4:in `/'
./thing.rb:4:in `make_error'
./thing.rb:8:in `<main>'
```

### Actual Behavior

```
jruby 9.2.0.0 (2.5.0) 2018-05-24 81156a8 Java HotSpot(TM) 64-Bit Server VM 25.121-b13 on 1.8.0_121-b13 +jit [darwin-x86_64]
divided by 0
./thing.rb:4:in `make_error'
./thing.rb:8:in `<main>'
```

headius:
This PR will fix #5200 by also adding subclasses' bound methods to the mapping table, in one way or another. The initial logic adds a JRubyMethod.implementers list of subclasses to include in the mapping table, but it is fairly naive and does not check if the method is actually overridden. The fix provided would be sufficient, but requires at least one annotation in a given native class to specify a list of implementers to include.

If other experiments don't pan out, we can go with this, but I will try to refine this to only include the actually-overridden methods, and ideally to not require an explicit list of classes.

### DIFF
--- a/core/src/main/java/org/jruby/AbstractRubyMethod.java
+++ b/core/src/main/java/org/jruby/AbstractRubyMethod.java
@@ -33,6 +33,7 @@
 
 package org.jruby;
 
+import org.jruby.anno.JRubyClass;
 import org.jruby.anno.JRubyMethod;
 import org.jruby.internal.runtime.methods.AliasMethod;
 import org.jruby.internal.runtime.methods.DynamicMethod;
@@ -48,6 +49,7 @@ import org.jruby.runtime.marshal.DataType;
  * @see RubyMethod
  * @see RubyUnboundMethod
  */
+@JRubyClass(name = {"Method", "UnboundMethod"}, overrides = {RubyMethod.class, RubyUnboundMethod.class})
 public abstract class AbstractRubyMethod extends RubyObject implements DataType {
     protected RubyModule implementationModule;
     protected String methodName;
--- a/core/src/main/java/org/jruby/Ruby.java
+++ b/core/src/main/java/org/jruby/Ruby.java
@@ -2953,19 +2953,18 @@ public final class Ruby implements Constantizable {
     }
 
     public void addBoundMethod(String className, String methodName, String rubyName) {
-        Map<String, String> javaToRuby = boundMethods.get(className);
-        if (javaToRuby == null) boundMethods.put(className, javaToRuby = new HashMap<>());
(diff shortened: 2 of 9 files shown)


headius:
@enebo This is green and done enough for now. It fixes the cases of Method/UnboundMethod, Integer/Fixnum/Bignum, and Array specialization overrides. It does not fix other "fast path" methods like RubyFixnum.op_plus_one but for that we would need to rethink how we annotate methods, probably by splitting binding and backtrace support into separate annotations.

This does fix the specific case called out in #5200.

PR #6660, opened 2021-04-22
[ji] set Ruby/Java cause on RaiseException

kares:
in Ruby we now get proper cause exception being hooked up.

if this (Ruby land) cause is a Ruby `Exception` or a Java `Throwable` we should also expose it on `RaiseException#getCause` as the proper cause ... thus a `re.printStackTrace` will properly print it's Ruby/Java cause throwable.

also helps logging using Java tools such as log4j.


headius:
Nice!

PR #6661, opened 2021-04-22
[refactor] move include_package into Java

kares:
also `java_import` was moved to not have these in 2 places (could ditch the whole JI *module.rb* ext file)

the motivation for the move is to have predictable include_package behavior e.g. on Java linking errors.
previously w `include_package`, all `NameError` where rescued and loop continued attempting to load a class with the next package -> effectively creating a weird set of errors where the real cause is unknown e.g. https://github.com/cheald/manticore/issues/87

the changes here make sure only `ClassNotFoundException` is a valid error to continue the loop, alll other `NameError` instances (e.g. caused by `LinkageError`) will be raised promptly.

### DIFF
--- a/core/src/main/java/org/jruby/RubyModule.java
+++ b/core/src/main/java/org/jruby/RubyModule.java
@@ -3712,7 +3712,7 @@ public class RubyModule extends RubyObject {
             RubySymbol sym = (RubySymbol) name;
 
             if (!sym.validConstantName()) {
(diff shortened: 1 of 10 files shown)


headius:
This looks good to me. Remove the commented-out load of module.rb when you decide you are done with it.

headius:
WIndows failure is #6678, disregard.

PR #6664, opened 2021-05-04
Copy 9.3 socket implementation to 9.2

headius:
This is a wholesale duplication of the 9.3 (master at time of
writing) socket implementation into 9.2 to pick up all recent
improvements in the socket library.

This should address #6541 and other socket issues fixed on master.

### DIFF
--- a/core/src/main/java/org/jruby/Ruby.java
+++ b/core/src/main/java/org/jruby/Ruby.java
@@ -4359,9 +4359,11 @@ public final class Ruby implements Constantizable {
 
     public <E extends Enum<E>> void loadConstantSet(RubyModule module, Class<E> enumClass) {
         for (E e : EnumSet.allOf(enumClass)) {
-            Constant c = (Constant) e;
-            if (c.defined() && Character.isUpperCase(c.name().charAt(0))) {
-                module.setConstant(c.name(), newFixnum(c.intValue()));
+            if (e instanceof Constant) {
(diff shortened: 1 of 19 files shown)


headius:
Ping @enebo and @ahorek to help evaluate this. No current schedule for 9.2.18 but this should be in the pipeline.

ahorek:
hi @headius !

just a few notes:
backport related tests? https://github.com/headius/jruby/pull/5
WDYT about https://github.com/jruby/jruby/pull/6401/files + your rework https://github.com/jruby/jruby/commit/1501bc7a0241b767ed5a0ccf1736bf00de5c024f ? it seems to be related to other changes.

ahorek (approved):
👍

headius:
@ahorek I merged in the tests. The one failure is a missing constant that I will investigate.

I will also look into the PR and commit you linked. Thanks!

headius:
> The one failure is a missing constant that I will investigate.

Actually it was a constant being defined on Linux that should not be, which is directly related to the other linked commits. I have cherrypicked the relevant changes to this branch.

ahorek:
```Socket.const_defined?``` is often used as a feature check like this https://github.com/puma/puma/blob/master/lib/puma/server.rb#L123 and definition of constants that aren't actually available might break such code. That's why I think this should be backported together.

anyway, thank your for fixing it @headius

PR #6666, opened 2021-05-09
Ruby 2.6 : Add :receiver option to NameError.new and NomethodError.new

k77ch7:
This PR adds :receiver option to to NameError.new and NomethodError.new.


kares (approved):
:+1: we could eliminate the Hash allocation from initialize 
(but it's less important than shipping the `receiver` feature)

k77ch7:
@kares Thanks for your review. I updated the commit.

PR #6671, opened 2021-05-12
Array max size auditing and fixes

headius:
This is a follow-up to #6670 that audits other locations we allocate arrays.

### DIFF
--- a/core/src/main/java/org/jruby/RubyArray.java
+++ b/core/src/main/java/org/jruby/RubyArray.java
@@ -87,9 +87,12 @@ import static org.jruby.RubyEnumerator.SizeFn;
 import static org.jruby.RubyEnumerator.enumeratorize;
 import static org.jruby.RubyEnumerator.enumeratorizeWithSize;
 import static org.jruby.RubyNumeric.checkInt;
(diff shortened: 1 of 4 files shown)


headius:
Hmm... so it turns out this effective max array size varies across VMs. As pointed out in some answers here, the OpenJDK ArrayList impl uses MAX_VALUE - 8 to be safe, but there does not appear to be any way to query for the actual effective maximum.

https://stackoverflow.com/questions/45196984/is-it-possible-to-query-the-java-vm-to-find-the-maximum-possible-array-size-how

We may want to make our max also MAX_VALUE - 8 since that would correspond to a 64-bit pointer alignment.

headius:
The array size maximum from AbstractCollection (mentioned in the SO answers above) appears to have been refactored into a utility similar to what we have in this PR:

https://github.com/openjdk/jdk/commit/218204b1a330f493e9eb50b4bb0098577077d502#diff-62a6afd2b76188557fb1b7ccabcfd8398d08eec42983af0c562ef72a588324bf

I think it is probably appropriate for us to use `MAX_VALUE - 8` since OpenJDK has normalized on that value.

Issue #6669, opened 2021-05-10
Intermittent NPE in interrupted read spec


headius:
The following failure was seen in one build for #6666 but appears to be intermittent:

```
IO#read raises IOError when stream is closed by another thread FAILED
Expected #<Java::JavaLang::NullPointerException: > (Java::JavaLang::NullPointerException)
to be kind of IOError
/home/travis/build/jruby/jruby/spec/ruby/core/io/read_spec.rb:322:in `block in <main>'
org/jruby/RubyBasicObject.java:2668:in `instance_exec'
org/jruby/RubyArray.java:4648:in `all?'
org/jruby/RubyArray.java:1852:in `each'
/home/travis/build/jruby/jruby/spec/ruby/core/io/read_spec.rb:178:in `<main>'
org/jruby/RubyKernel.java:1049:in `load'
org/jruby/RubyBasicObject.java:2668:in `instance_exec'
org/jruby/RubyArray.java:1852:in `each'
```

This spec tests whether a blocking read interrupted by the stream closing raises IOError. There are a number of possible explanations since this involves concurrent threads mucking about with the same IO object. I do not have a full stack trace and have not been able to reproduce this.

ntkme:
@headius I got similar intermittent `NullPointerException` when reading from a stream that is closed in another thread. In my case, I'm running a child process and polling stdout of child process. When the child process exits and closes the stdout stream, the polling thread may occasionally crash with `NullPointerException` instead of raising an `IOError`.

```
warning: thread "Ruby-0-Thread-3: /Users/runner/work/sass-embedded-host-ruby/sass-embedded-host-ruby/lib/sass/embedded/dispatcher.rb:17" terminated with exception (report_on_exception is true):
java.lang.NullPointerException
	at org.jruby.util.io.PosixShim.read(PosixShim.java:158)
	at org.jruby.util.io.OpenFile$2.run(OpenFile.java:1325)
	at org.jruby.util.io.OpenFile$2.run(OpenFile.java:1316)
	at org.jruby.RubyThread.executeReadWrite(RubyThread.java:1748)
	at org.jruby.util.io.OpenFile.readInternal(OpenFile.java:1384)
	at org.jruby.util.io.OpenFile.fillbuf(OpenFile.java:1295)
	at org.jruby.RubyIO.getByte(RubyIO.java:2882)
	at org.jruby.RubyIO.getbyte(RubyIO.java:2859)
[...]

PR #6672, opened 2021-05-13
Ruby 2.6 : Update ioOpenGeneric  for Pathname methods.

k77ch7:
This PR update _ioOpenGeneric_ for Pathname# methods.
see https://bugs.ruby-lang.org/issues/14245


headius:
Failure is same as #6669 and predates this PR so looks good to merge.

headius:
Thank you!

PR #6676, opened 2021-05-19
Update to psych 4.0

headius:


headius:
The specs at ruby/spec are unfortunately out of date, as mentioned here: https://github.com/ruby/spec/pull/837

They will get an update at the end of the month which will pass on both Psych 4 and Psych < 4.

Given this larger behavioral change (`safe_load` by default) I am punting this to JRuby 9.5 which may be the Ruby 3.1 release. I will do a separate PR to 9.3 for the latest equivalent Ruby 2.6 Psych.

headius:
This was handled long ago when we updated compatibility for JRuby 9.4 (which currently uses Psych 5.x).

PR #6677, opened 2021-05-19
Modify IO in and out streams to use sysread/write

### LINKED ISSUE #6588: RubyIO.getOutputStream incompatible with external encoding

headius:
This change addresses the remaining issue in #6588 by using
sysread and syswrite for the getInStream/getOutStream IO wrappers
that implement InputStream and OutputStream. This bypasses
encoding logic (as well as buffering), avoiding the question of
what encoding to treat incoming or outgoing bytes as.

I also added fast paths for sysread/syswrite that avoid allocating
any strings or byte arrays to accommodate this new logic.

Fixes #6588

### DIFF
--- a/core/src/main/java/org/jruby/RubyIO.java
+++ b/core/src/main/java/org/jruby/RubyIO.java
@@ -334,17 +334,17 @@ public class RubyIO extends RubyObject implements IOEncodable, Closeable, Flusha
 
             @Override
             public void write(int b) throws IOException {
-                RubyIO.this.write(runtime.getCurrentContext(), b);
+                RubyIO.this.syswrite(runtime.getCurrentContext(), b);
             }
 
             @Override
             public void write(byte[] b) throws IOException {
-                RubyIO.this.write(runtime.getCurrentContext(), b, 0, b.length, ASCIIEncoding.INSTANCE);
+                RubyIO.this.syswrite(runtime.getCurrentContext(), b, 0, b.length);
             }
 
             @Override
             public void write(byte[] b, int off, int len) throws IOException {
-                RubyIO.this.write(runtime.getCurrentContext(), b, off, len, ASCIIEncoding.INSTANCE);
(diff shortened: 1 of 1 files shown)


headius:
This appears to have caused some issues for consumers of the getInStream InputStream. Marshal and Psych both appear to expose these issues, ranging from warnings about buffered IO to simply not reading the right data (since sysread bypasses any buffered data). I don't think we can change the contract of these particular wrappers at this time.

headius:
Revisit for 9.4, perhaps as a separate path to getting these streams.

headius:
An alternative approach would be to use sysread/write but deal with any buffered data first... not sure if those two operations mesh when it comes to encodings, though.

headius:
Close this for now. Might be that we have to create a new class and deprecate this one. New class would require you give it an encoding to use (or some flag to just honor the target IO's internal encoding), and if you don't do that it will use syswrite. Require users to opt into both patterns, which we could do in Marshal and Psych to address the incompatibility there.

PR #6679, opened 2021-05-20
[ji] allow Java errors from Ruby threads to reach the handler

kares:


kares:
one potential concern for improvement, even though `java.lang.Error` reaches the uncaught Java handler, do we still want to keep the existing error around? (believe `Thread.status` will not report correctly in case the Error is actually considered okay)

headius:
Good point! I think we would want to still report that as the terminal exception for the thread.

kares:
hmm, seems like bubbling up Java errors is [revealing a bug with test/unit's parallel test runner](https://app.travis-ci.com/github/jruby/jruby/jobs/535643017#L1321-L1345):
```
java.lang.NullPointerException
	at org.jruby.util.io.PosixShim.read(PosixShim.java:158)
	at org.jruby.util.io.OpenFile$2.run(OpenFile.java:1325)
	at org.jruby.util.io.OpenFile$2.run(OpenFile.java:1316)
	at org.jruby.RubyThread.executeReadWrite(RubyThread.java:1747)
	at org.jruby.util.io.OpenFile.readInternal(OpenFile.java:1384)
	at org.jruby.RubyIO.getPartial(RubyIO.java:3093)
	at org.jruby.RubyIO.readpartial(RubyIO.java:3025)
```
not sure I guess we would need an uncaught Java handler during the MRI suite why this suddenly matters ...

spec/regression/GH-1106_fiber_must_propagate_java_exceptions_spec.rb
```diff
@@ -38,7 +38,7 @@
       
       Thread.pass while t.status
       
-      expect(t.status).to eq(false) # dead
+      expect(t.status).to eq(nil) # terminated with an exception
```
ahorek, line 41:
```
Thread.new { Thread.exit }.join.status => false
```
there's an internal exception but it shouldn't be propagated

```suggestion
      expect(t.status).to eq(false) # dead
```

core/src/main/java/org/jruby/RubyThread.java
```diff
@@ -1924,17 +1958,17 @@ protected void printReportExceptionWarning() {
...
-            Helpers.throwException(throwable);
-        }
+        assert isCurrent();
 
         final Ruby runtime = getRuntime();
 
-        assert isCurrent();
-
-        IRubyObject rubyException;
+        // if unrescuable (internal exceptions) just re-raise and let it be handled by thread handler
```
ahorek, line 1965:
```suggestion
        if (throwable instanceof Error || throwable instanceof MainExitException) {
            exitingException = throwable;
            Helpers.throwException(throwable);
            return;
        }

        // if unrescuable (internal exceptions) just re-raise and let it be handled by thread handler
        if (throwable instanceof Unrescuable) {
            Helpers.throwException(throwable);
            return;
        }
```

core/src/main/java/org/jruby/RubyThread.java
```diff
@@ -1924,17 +1958,17 @@ protected void printReportExceptionWarning() {
...
+        if (throwable instanceof Error || throwable instanceof Unrescuable) {
+            exitingException = throwable;
+            Helpers.throwException(throwable);
+        }
```
ahorek, line 1969:
```suggestion
        if (throwable instanceof Error || throwable instanceof MainExitException) {
            exitingException = throwable;
            Helpers.throwException(throwable);
            return;
        }

        // if unrescuable (internal exceptions) just re-raise and let it be handled by thread handler
        if (throwable instanceof Unrescuable) {
            Helpers.throwException(throwable);
            return;
        }
```

ahorek:
> 
> 
> hmm, seems like bubbling up Java errors is [revealing a bug with test/unit's parallel test runner](https://app.travis-ci.com/github/jruby/jruby/jobs/535643017#L1321-L1345):
> 
> ```
> java.lang.NullPointerException
> 	at org.jruby.util.io.PosixShim.read(PosixShim.java:158)
> 	at org.jruby.util.io.OpenFile$2.run(OpenFile.java:1325)
> 	at org.jruby.util.io.OpenFile$2.run(OpenFile.java:1316)
> 	at org.jruby.RubyThread.executeReadWrite(RubyThread.java:1747)
> 	at org.jruby.util.io.OpenFile.readInternal(OpenFile.java:1384)
> 	at org.jruby.RubyIO.getPartial(RubyIO.java:3093)
> 	at org.jruby.RubyIO.readpartial(RubyIO.java:3025)
> ```
> 
> not sure I guess we would need an uncaught Java handler during the MRI suite why this suddenly matters ...

test workers were dying because any ```ThreadKill``` exception killed the whole process. With my changes, all tests passed, @kares could you review the logic again?

PR #6680, opened 2021-05-20
Update to Psych 3.3.2

headius:
Another attempt to update Psych appropriate for a Ruby 2.6 release.


headius:
The failures here all appear to be due to a behavior change in SnakeYAML that causes it to defer some emits while waiting for comment events: https://bitbucket.org/asomov/snakeyaml/pull-requests/7

The tests in question can be fixed by adding `@stream.finish` but this behavior change in SnakeYAML may have other implications. I will exclude or modify the tests in this PR and file issues with Psych and SnakeYAML about it.

headius:
> The failures here all appear to be due to a behavior change in SnakeYAML

Correction: the empty output failures are due to this behavior change (https://github.com/ruby/psych/issues/499, https://bitbucket.org/asomov/snakeyaml/issues/503/comment-handling-causes-deferred-emit), while the emitter exceptions are due to ruby/psych#500.

Issue #2150, opened 2014-11-08
Dir#glob does not return files with proper cases


headius:
This appears to be a test for case-insensitive, case-preserving filesystems like HFS (non-case-sensitive mode), FAT32, NTFS and so on. We do not return the files with the proper case. This may be a simple fix, like absolutizing the filenames before we return them or something similar.

``` ruby
  def test_glob_cases
    feature5994 = "[ruby-core:42469] [Feature #5994]"
    feature5994 << "\nDir.glob should return the filename with actual cases on the filesystem"
    Dir.chdir(File.join(@root, "a")) do
      open("FileWithCases", "w") {}
      return unless File.exist?("filewithcases")
      assert_equal(%w"FileWithCases", Dir.glob("filewithcases"), feature5994)
    end
    Dir.chdir(File.join(@root, "c")) do
      open("FileWithCases", "w") {}
      mode = File.stat(".").mode
      begin
        File.chmod(mode & ~0444, ".")
        return if mode == File.stat(".").mode
        assert_equal(%w"filewithcases", Dir.glob("filewithcases"), feature5994)
      ensure
        File.chmod(mode, ".")
      end
    end
  end
```

Failure:

```
TestDir#test_glob_cases [/Users/headius/projects/jruby/test/mri/ruby/test_dir.rb:238]:
[ruby-core:42469] [Feature #5994]
Dir.glob should return the filename with actual cases on the filesystem.
<["FileWithCases"]> expected but was
<["filewithcases"]>.
```

Test can be run with `jruby test/mri/runner.rb ruby/test_dir.rb -n test_glob_cases`

groddeck:
@kelcecil - took me a minute to figure that out also, but check out the lines in each section beginning with ```return```.  These ensure that different parts of the test only apply to certain filesystem types.

In the first section, a file named ```FileWithCases``` is created, and then ```return unless File.exist?("filewithcases")``` will skip this test unless the filesystem says that, yes, a file with that name does exist.  If ```File.exist?``` returns true for the all-lowercase name, even though the file was created with camel-case, we have verified that we are on a case-insensitive, case-preserving filesystem type, and proceed.

[...]

PR #6681, opened 2021-05-23
Fix Dir.glob with cases

edipofederle:
Hi,

This PR updates the `addToResultIfExists` Java method in order to handle the correct file name in the file system. It should close this issue: https://github.com/jruby/jruby/issues/2150.

**note:** This is my attempt to achieve my first contribution to JRuby, so any feedback here would be great.

Thanks

### DIFF
--- a/core/src/main/java/org/jruby/util/Dir.java
+++ b/core/src/main/java/org/jruby/util/Dir.java
@@ -750,9 +750,34 @@ public class Dir {
             if (cwd == null) cwd = "C:";
             cwd = cwd + "/";
         }
+
         FileResource file = JRubyFile.createResource(runtime, cwd, fileName);
+
         if (file.exists()) {
-            return func.call(bytes, begin, end - begin, enc, arg);
+            byte[] newBytes;
+            
+            // get the real filename (case-sensitive)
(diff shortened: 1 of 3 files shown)


headius:
Looks like some valid problems in the travis run... are you able to run those suites locally to reproduce?

headius:
Converted to a draft while you try to work out issues. Thanks for digging into this!

Whenever we do merge I will probably have to squash your commits to get right of the "try again". If you can restructure the commits and do a force push to clean it up, then we can keep more of the history intact.

edipofederle:
> Converted to a draft while you try to work out issues. Thanks for digging into this!
> 
> Whenever we do merge I will probably have to squash your commits to get right of the "try again". If you can restructure the commits and do a force push to clean it up, then we can keep more of the history intact.

Thanks! Yes, sorry for the 'try again' commits, I think I can clean up once I finished it.

edipofederle:
> Looks like some valid problems in the travis run... are you able to run those suites locally to reproduce?

Hi, I'm enable to run, but with some errors, maybe I can share the issues I had locally ?

enebo:
@edipofederle Awesome!  Thanks for your first (of many) contribution(s)! :)

Issue #6006, opened 2019-12-21
Implement non-fiber generators for core classes


headius:
With #5672, we moved to a pure-Ruby Enumerator#next impl that uses Fibers. Along with that came the protocol originated by Rubinius whereby any class that implements `to_generator` could provide an external generator (iterator) that does not need a Fiber to maintain state.

We did not ever implement this method for any core classes.

Given that the vast majority of Enumerator#next usage is against Array, this would substantially reduce the number of live fibers in the system. I believe we should also do it for any other types that have their own `each` logic, such as Hash.

See #6004 for an example of how quickly our thread-based Fiber can blow up if it's leaking. If these were all external iterators, we might avoid the problem (but it should still be fixed).

headius:
Ok maybe I'm going crazy but I could find no reference to `to_generator` in Rubinius sources.

I made a quick attempt to implement a simple Array#each generator in the following patch, and it definitely works better than threads! However I'm not sure I have the protocol right and there's no example to follow.

```diff
diff --git a/core/src/main/ruby/jruby/kernel.rb b/core/src/main/ruby/jruby/kernel.rb
index 59c1af1fce..ce3aa7e8e2 100644
--- a/core/src/main/ruby/jruby/kernel.rb
+++ b/core/src/main/ruby/jruby/kernel.rb
@@ -25,4 +25,5 @@ load 'jruby/kernel/io.rb'
 load 'jruby/kernel/gc.rb'
 load 'jruby/kernel/range.rb'
 load 'jruby/kernel/file.rb'
+load 'jruby/kernel/array.rb'
 
diff --git a/core/src/main/ruby/jruby/kernel/array.rb b/core/src/main/ruby/jruby/kernel/array.rb
index e69de29bb2..917d632b73 100644
--- a/core/src/main/ruby/jruby/kernel/array.rb
+++ b/core/src/main/ruby/jruby/kernel/array.rb
@@ -0,0 +1,28 @@
+class Array
+  class Generator
+    def initialize(ary)
+      @ary = ary
+      @index = 0
+    end
+
+    def next?
+      @index < @ary.length - 1
+    end
+
+    def next
+      raise StopIteration unless next?
+      @result = @ary[@index]
+      @index += 1
+    end
+
+    def rewind
+      @index = 0
+    end
+  end
+
+  def to_generator(method, *args)
+    if method == :each && args == []
+      return Array::Generator.new(self)
+    end
+  end
+end
\ No newline at end of file
```

[...]

PR #6686, opened 2021-05-26
Initial support for lightweight generators

headius:
This adds a simple lightweight generator for Array#each and starts
the process of making a formal to_generator method that classes
can implement to do external iteration.

The use of Fiber for internal iteration allows all such cases to
work, but leads to a lot of thread creation that cannot easily
handle large numbers of enumerators. In addition, these fiber
threads frequently will get hard referenced and stuck alive,
leading to thread leaks such as in #6004. We can provide non-fiber
external iteration for many of the most common use cases, such as
Array#each, while also providing a protocol for users to add their
own generators returned by to_generator.

See #6006 for further discussion.

### DIFF
--- a/core/src/main/ruby/jruby/kernel.rb
+++ b/core/src/main/ruby/jruby/kernel.rb
@@ -14,6 +14,9 @@ rescue Exception # java.lang.ProcessBuilder not available
   # leave old backtick logic in place
 end unless JRuby::Util.native_posix? # native POSIX uses new logic for back-quote
 
+# Generators for Enumerator#next
+load 'jruby/kernel/jruby/generators.rb'
+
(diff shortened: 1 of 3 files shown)


headius:
This may become moot as we further integrate JDK19+ native fibers, but we should revisit this after 9.4.

headius:
This is largely unneeded given the availability of virtual threads on all supported JDKs for JRuby 10.0+. Other issues already track the overhead of spinning up new fibers on top of virtual threads, but that overhead is already far lower than when we were forced to use native threads.

This was an interesting experiment dating back to the Rubinius years. The path forward now is to simply make fibers as lightweight as possible.

Issue #6514, opened 2020-12-29
javax.annotation-api has been deprecated, please move to jakarta-annotation-api


baldurmen:
Hi!

It seems jruby still depends on `javax.annotation-api 1.3.2`, which has been archived, deprecated and replaced by `jakarta-annotation-api`.

The new source code can be found here: https://github.com/eclipse-ee4j/common-annotations-api

I'm planning to see if I can build jruby using `jakarta-annotation-api 2.0.0` in the next few days, but meanwhile I'm opening this issue to track the need to move from the old javax lib :)

headius:
@baldurmen Wow, we would love to have a PR to make this change! We only use the API at build time for our annotation processor, so it should be a pretty safe change to make. I am hoping, of course, that the Jakarta version of this API is basically a drop-in replacement.

PR #6694, opened 2021-06-01
Do not pollute CP with annotation-api

pzygielo:
https://github.com/jruby/jruby/issues/6514#issuecomment-752748414:
> We only use the API at build time for our annotation processor

Unfortunately, current `compile` scope used propagates to client project (it isn't _compile-time-only_) and is placed on project's classpath. I can't switch annotation api in the project, but I'd like to consider JRuby upgrade.

I've built and used jruby-core 9.3.0.0-SNAPSHOT to verify that this change would work as I expect it. (I used published snapshot in first place and learnt that it conflicts with the project.)

### DIFF
--- a/core/pom.rb
+++ b/core/pom.rb
@@ -75,7 +75,7 @@ project 'JRuby Base' do
 
   jar 'com.headius:backport9:1.10'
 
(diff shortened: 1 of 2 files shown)


headius:
Thank you... I noticed during the move to jakarta-annotation-api that this was being shaded and planned to fix it myself!

PR #6697, opened 2021-06-02
[ffi] update to 1.15.1

ahorek:


headius:
👍

headius:
We probably should try to do this for 9.2 as well, yes?

ahorek:
I don't think it's necessary and ffi wasn't gemified back in 9.2. So I would prefer to keep it for 9.3 unless there's some real issue that should be backported.

headius:
Ah yes, I agree.

PR #6698, opened 2021-06-02
webrick 1.7.0

ahorek:


headius:
This upgrade is okay since it is a gem but generally we try to match the standard library exactly.

ahorek:
that would be version 1.4.4 https://github.com/ruby/ruby/commit/8b49c3e4bc767bec8a66ac81cbda033330fb2703 so we already don't match it exactly

but the latest webrick claims to be compatible with ruby 2.4+ and I don't see a big compatibility risk by shipping a more recent version. Most changes to webrick are bugs and CVS backports anyway.

here's the changelog for the reference
https://github.com/ruby/webrick/compare/v1.6.1...v1.7.0

Issue #6523, opened 2021-01-11
Replacing timeout in Net::HTTP#connect with nonblocking io


mohamedhafez:
Years ago, I submitted a patch to MRI ruby to replace the timeout in Net::HTTP#connect with nonblocking io, however it was [shot down](https://bugs.ruby-lang.org/issues/12435) because there was no good way to time out the DNS call in `Socket.sockaddr_in` due to limitations in the C libraries available.

Are we subject to the same limitations in Java? I've been running a version of [this patch](https://github.com/jruby/jruby/compare/master...mohamedhafez:patch-1?diff=split) for years because I was terrified by @headius's [blog post from way back in 2008](http://blog.headius.com/2008/02/ruby-threadraise-threadkill-timeoutrb.html) about the dangers of ruby's `timeout`, lol.

So my questions are:

a) Is it possible to introduce a timeout for `Socket.sockaddr_in` in the Java implementation?

b) How big of a deal is this / is this worth the effort to address? With the use of `AtomicBoolean` and other synchronization methods and a thread pool executor in Jruby's `Timeout.timeout` implementation, are the concerns of that old blog post relevant to Net::HTTP#connect's use of `Timeout.timeout` in JRuby today?

mohamedhafez:
Actually, it looks like a way simpler patch would fix this: https://github.com/jruby/jruby/compare/master...mohamedhafez:patch-2?diff=split

While this still has the issue of not accounting for DNS timeouts, it looks like in Ruby 2.7 `Socket.tcp` does take a `:resolv_timeout` option, and then with just a 3 line patch we'd be able to do away with `Timeout.timeout`.

So I guess to revise my question, once JRuby is compatible with 2.7.0, what do you guys think about incorporating a patch like this, including a `:resolv_timeout`, into JRuby's version of net/http.rb? Or is the use of `Timeout.timeout` no longer an issue with the way JRuby implements it?

headius:
The big question here will be if there is a way to tell Java to timeout the DNS resolution, since we currently implement all the TCP sockets using Java's socket libraries. Maybe you can do some research on that?

mohamedhafez:
Hmm:/ So it looks like Java's InetAddress#getByName​, doesn't take a timeout option, and [just ends up calling getaddrinfo under the hood](https://medium.com/@maheshsenni/host-name-resolution-in-java-80301fea465a), at least in OpenJDK.

[...]

PR #6699, opened 2021-06-02
update [net] libraries

ahorek:
refs https://github.com/jruby/jruby/issues/6523

* get rid of ```Timeout.timeout```
* update to latest versions, there were many improvements since the last update

there're two main issues that prevent gemification

1/ missing OpenSSL APIs, mostly SNI support, see
https://github.com/jruby/jruby-openssl/issues/112 https://github.com/jruby/jruby-openssl/issues/103 https://github.com/jruby/jruby-openssl/issues/202
this should be really implemented, but it's skipped for now.

2/ https://github.com/ruby/digest
is a hard dependency and the gem has no JRuby version


headius:
Have these changes been merged into Ruby 2.6? We generally try to match their standard library exactly and not go beyond it.

ahorek:
changes like https://github.com/ruby/net-smtp/pull/4 https://github.com/ruby/net-smtp/pull/21 were already merged, but they weren't backported to 2.6.

here's a ruby ticket that tracks the promotion to bundled gems https://bugs.ruby-lang.org/issues/17873

> not go beyond it.

I have to admit there's a compatibility concern here, even if I didn't found any issue that could break puma or something like that. We should probably wait with this until JRuby 9.4, but the good part is we know what works, what's missing, and what needs to be done to make this happen.

> Could you open a few issues on JRuby and those libraries to get the process moving?

there are already open issues
1/ jOpenSSL APIs, https://github.com/jruby/jruby-openssl/issues/112
2/ Digest, https://github.com/ruby/digest/issues/20
3/ SSLSocket#session=, and there's also this change https://github.com/jruby/jruby/commit/12082fd045319a760de45842d4d637899edb8048

and release all gems (I think it won't happen before the Ruby 3.1 release, but maybe Ruby maintainers will be willing to do so sooner)

I'm really looking forward to finally address http://blog.headius.com/2008/02/ruby-threadraise-threadkill-timeoutrb.html after 13 years :)

headius:
Agreed on all points, thanks for the links to the key open issues. We will get 9.2.18 out today, then wrap up 9.3 by end of month, and then start the heavy lifting to bring 9.4 in line with Ruby 3.0 including these troublesome gems.

headius:
There is a mild update of net libraries in https://github.com/jruby/jruby/pull/6802, to align them with Ruby 2.6.8.

PR #6711, opened 2021-06-08
[launcher] basename for solaris

ahorek:
```
basename -- "$sym"
```
it fails on launch because "--" isn't a valid java command (it works on linux)

```
basename "$sym"
```
returns "java" on both platforms

### DIFF
--- a/bin/jruby.bash
+++ b/bin/jruby.bash
@@ -74,7 +74,7 @@ resolve_symlinks() {
     # 4) append the basename
     sym="$(readlink "$cur_path")"
     sym_base="$(cd -P -- "$(dirname -- "$cur_path")" >/dev/null && pwd -P)"
-    cur_path="$(cd "$sym_base" && cd "$(dirname -- "$sym")" && pwd -P)/$(basename -- "$sym")"
+    cur_path="$(cd "$sym_base" && cd "$(dirname -- "$sym")" && pwd -P)/$(basename "$sym")"
   done
   result="$cur_path"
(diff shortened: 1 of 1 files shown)


headius:
I will have @mrnoname1000 look at this since he has helped with this script recently. The original introduction of `basename --` appears to have been a long time ago in b1fec0dd242ee473094cbba32dc7a553f24dbb0b and I'm pretty sure I just got it off some SO post or blog entry somewhere.

mrnoname1000:
Looks like POSIX doesn't allow options to basename, so this is another GNU-ism we need to work around. Fortunately, basename operates purely on text, so it's pretty easy to implement in pure shell:
```sh
basename() {
    local base="$1"
    if [ -n "$base" ]; then
        # remove trailing slashes
        base=${base%"${base##*[!/]}"}
        # remove leading directories
        base=${base##*/}
        # if $1 is all slashes, base=/
        base=${base:-/}
    fi
    printf "%s\n" "$base"
}
```
The printf can be replaced with an alternate return method to avoid a subshell, such as `RET=$base`.

Also basename has a little-used feature to remove a suffix, this function doesn't implement that.

headius:
@mrnoname1000 Perhaps I should merge this as-is and you can push another PR with the pure-POSIX basename change?

mrnoname1000:
Actually I overlooked that stackexchange link, the implementation there looks better than mine and seems more efficient. I'll create a PR with that instead.

Issue #596, opened 2013-03-19
Adding a method to eigenclass doesn't work properly. Works in cruby


d1rewolf:
``` ruby
 class X
  def total_cost 
    99
  end
  alias_method :old_dup, :dup
  def dup
    c = self.old_dup
    class << c
      attr_reader :total_cost
      public
      def initialize_copy(orig)
        puts "Inside initialize_copy"
        @total_cost = orig.total_cost
      end
    end
    c.initialize_copy(self)
    return c
  end
end

 an_x = X.new
 an_x.dup
```

```
$ ruby /tmp/t.rb 
Inside initialize_copy
$ jruby /tmp/t.rb 
NoMethodError: private method `initialize_copy' called for #<X:0x300cd634>
     dup at /tmp/t.rb:16
  (root) at /tmp/t.rb:22
$ jruby -v
jruby 1.7.3 (1.9.3p385) 2013-02-21 dac429b on Java HotSpot(TM) 64-Bit Server VM 1.7.0_03-b04 [linux-amd64]
```

rjnienaber:
Still seems to be an issue with current rubies:

```
ruby 1.9.3p550 (2014-10-27) [x86_64-darwin14.5.0]
Inside initialize_copy

ruby 2.2.4p230 (2015-12-16 revision 53155) [x86_64-darwin14]
Inside initialize_copy

ruby 2.3.0p0 (2015-12-25 revision 53290) [x86_64-darwin14]
Inside initialize_copy

jruby 1.7.24 (1.9.3p551) 2016-01-20 bd68d85 on Java HotSpot(TM) 64-Bit Server VM 1.8.0_51-b16 +jit [darwin-x86_64]
NoMethodError: private method `initialize_copy' called for #<X:0x6f7fd0e6>
     dup at issues/596.rb:20
  (root) at issues/596.rb:26

jruby 9.0.5.0 (2.2.3) 2016-01-26 7bee00d Java HotSpot(TM) 64-Bit Server VM 25.51-b03 on 1.8.0_51-b16 +jit [darwin-x86_64]
NoMethodError: private method `initialize_copy' called for #<X:0x2b9627bc>
    dup at issues/596.rb:20
  <top> at issues/596.rb:26
```

kares:
confirmed (still) on 9.1.9.0 : 
```
NoMethodError: private method `initialize_copy' called for #<X:0x5bb21b69>
Did you mean?  initialize_clone
               initialize
               initialize_dup
[...]

PR #6712, opened 2021-06-09
Fix some specs regarding method visibility

edipofederle:
Untag some tests regarding method visibility. Also closes https://github.com/jruby/jruby/issues/596.

### DIFF
--- a/core/src/main/java/org/jruby/RubyModule.java
+++ b/core/src/main/java/org/jruby/RubyModule.java
@@ -2370,15 +2370,18 @@ public class RubyModule extends RubyObject {
                 IRMethod method = closure.convertToMethod(name.getBytes());
                 if (method != null) {
                     newMethod = new DefineMethodMethod(method, visibility, this, context.getFrameBlock());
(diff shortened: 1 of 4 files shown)


enebo (commented):
I think the functional part of this PR is that if clazz is a singleton it will return public which does seems to match vm_define_method in MRI.  This PR with only the singleton check of the if statement may be what you want (although we should be able to see some test/spec pass too as a result).

I am not sure what NO_SUPER_CLASS and NOT_IMPLEMENTED checks are for?  They are not doing what you expect but if you explain what you think they are doing I can maybe help you do the right thing.  NO_SUPER_CLASS.toString() -> "NO_SUPER_CLASS" which would only have an effect if your Ruby method was named that.

You can ping me or @headius on matrix to talk about this.  Deciphering our codebase and MRI's codebase can be tricky at times :)

headius (approved):
Looks ok to me. Concur with @enebo that the main change here is to use public visibility for singleton classes.

Issue #6714, opened 2021-06-11
getsockopt crashing with 9.2.18.0 and JRUBY_OPTS=--dev flag


gsouzab:
Hey all 👋 
After upgrading to the latest 9.2.18.0 we started experiencing this weird error. It happens only when using the JRUBY_OPTS="--dev" flag. Let me know if you need any extra details to help with the investigation.

**Environment Information**
* JRuby 9.2.18.0
* JRUBY_OPTS="--dev"

**Expected Behavior**
`getsockopt` works

**Actual Behavior**
When using JRUBY_OPTS=--dev `getsockopt` crashes with the following stacktrace

`Traceback (most recent call last):
       16: from org.jruby.runtime.callsite.CachingCallSite.call(CachingCallSite.java:85)
       15: from org.jruby.RubyKernel$INVOKER$s$0$3$eval.call(RubyKernel$INVOKER$s$0$3$eval.gen)
       14: from org.jruby.RubyKernel.eval(RubyKernel.java:1048)
       13: from org.jruby.RubyKernel.evalCommon(RubyKernel.java:1086)
       12: from org.jruby.ir.interpreter.Interpreter.evalWithBinding(Interpreter.java:182)
       11: from org.jruby.ir.interpreter.Interpreter.evalCommon(Interpreter.java:158)
       10: from org.jruby.ir.interpreter.Interpreter.INTERPRET_EVAL(Interpreter.java:106)
        9: from org.jruby.ir.interpreter.StartupInterpreterEngine.interpret(StartupInterpreterEngine.java:72)
        8: from org.jruby.ir.interpreter.InterpreterEngine.processCall(InterpreterEngine.java:325)
        7: from org.jruby.runtime.callsite.CachingCallSite.call(CachingCallSite.java:206)
        6: from org.jruby.runtime.callsite.CachingCallSite.cacheAndCall(CachingCallSite.java:397)
        5: from org.jruby.ext.socket.RubyBasicSocket$INVOKER$i$2$0$getsockopt.call(RubyBasicSocket$INVOKER$i$2$0$getsockopt.gen)
        4: from org.jruby.ext.socket.RubyBasicSocket.getsockopt(RubyBasicSocket.java:430)
        3: from com.sun.proxy.$Proxy30.getsockoptInt(Unknown Source)
        2: from jnr.ffi.provider.NativeInvocationHandler.invoke(NativeInvocationHandler.java:47)
[...]

PR #6716, opened 2021-06-12
[travis] add --dev stage, fix JRUBY_OPTS

ahorek:
1/ JRUBY_OPTS are being ignored, so half of the stages aren't tested as supposed
2/ add a --dev stage in order to catch https://github.com/jruby/jruby/issues/6714

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -46,17 +46,17 @@ matrix:
     - stage: test
       name: MRI core int
       script:
+        - export JRUBY_OPTS=-J-Xmx1G
         - mvn clean package -Pbootstrap
         - bundle install
         - jruby -S rake test:mri:core:int
-      env: JRUBY_OPTS='$JRUBY_OPTS -J-Xmx1G'
 
     - name: MRI core jit
       script:
+        - export JRUBY_OPTS=-J-Xmx1G
         - mvn clean package -Pbootstrap
         - bundle install
         - jruby -S rake test:mri:core:fullint
-      env: JRUBY_OPTS='$JRUBY_OPTS -J-Xmx1G'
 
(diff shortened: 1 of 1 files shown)


headius:
Why does simply adding this to `env` not work? That seems very counter-intuitive to me.

headius:
Looking into fixing this `getsockoptInt` thing. It appears that the non-generated logic is not using the right name for binding this method.

ahorek:
it's a problem with thit's a problem with the order

1/ env: JRUBY_OPTS=xxx
sets the variable at the very beginning
2/ before_script: **unset JRUBY_OPTS**
which is executed after clears them

so all env options like
https://github.com/jruby/jruby/blob/ad4e8a9d0678453d5855a7f2b06cc8abc88ee82a/.travis.yml#L143
are being ignored

however, there's a probably better way how to avoid the conflicte order

headius:
We may not need to clear it anymore. I think we started clearing it as some point because Travis was still enabling JRuby C extensions, which have not been supported since JRuby 1.6.

ahorek:
unfortunately, the problem is still there, I tried...

ahorek:
how about changing this
```
- name: JRuby tests jit indy jdk11
  script:
    - mvn clean package -Pbootstrap
  env: JRUBY_OPTS=-Xcompile.invokedynamic
```

to
```
- name: JRuby tests jit indy jdk11
  script:
    - export JRUBY_OPTS=-Xcompile.invokedynamic
    - mvn clean package -Pbootstrap
```
?

headius:
@ahorek That does seem a bit better than the double env var dance. I am annoyed that we can't just set this directly in env.

Issue #6682, opened 2021-05-24
Native libraries that need to be gemified


headius:
There are still several native libraries that CRuby has moved into gems and we have not. We need to deal with these so people can add those native gems to dependencies without issue. See #4875 for earlier discussion about gemification.

The remaining missing gems are summarized below:

bigdecimal - we still have our own version and it is not integrated into the gem in any way. The gem maintainer has expressed some reluctance to incorporating JRuby code in the past.
date - we have our own impl for 90% of this and it has not been incorporated into the gem.
dbm, gdbm, and sdbm have never been supported and likely never will.
digest - we have our own impl and it has not been incorporated into the gem.
etc - we have our own impl and it has not been incorporated into the gem. This could potentially be replaced by Ruby FFI code and made easier to add to the gem.
fcntl - we have our own impl and it has not been incorporated into the gem. This could also be replaced by FFI code.
fiddle - we have a pure-Ruby impl based on FFI that has not been incorporated into the gem. In the past we had talked with @chrisseaton about unifying on a shared fiddle impl that uses FFI, or else a shared FFI based on fiddle, but that has not happened.
openssl - we release this in gem form but our version of the library has not been incoporated into the standard gem.
stringio - we have our own impl and it has not been incorporated into the gem.
strscan - we have our own impl and it has not been incorporated into the gem (ruby/strscan#15).
zlib - we have our own impl and it has not been incorporated into the gem.

hsbt:
👍  I'll support this.

headius:
strscan gemification is started in https://github.com/ruby/strscan/pull/25.

headius:
See ruby/digest#20 / ruby/digest#21 for the gemification of Digest.

headius:
stringio started in https://github.com/ruby/stringio/pull/21

hsbt:
@headius Thanks for your work! Did we get the approval for relicensing JRuby libraries that are this work at https://github.com/ruby/digest/pull/21#issuecomment-861015401 ?

I'm not sure what we should get the approval for that each pull requests.

headius:
@hsbt I believe we got approval from everyone except @chrisseaton in that PR and updated my comment to be a checklist.

@chrisseaton You responded affirmatively to the PR, but could you add something along the lines of "I consent to having my contributions relicensed etc etc"?

headius:
[...]

PR #6719, opened 2021-06-12
draft: install digest from the gem

ahorek:
gemification, part of the https://github.com/jruby/jruby/issues/6682

depends on https://github.com/ruby/digest/issues/20 / https://github.com/ruby/digest/pull/21 that has to be released first


headius:
This has been done in my stdlib update for the incoming 3.0 branch. It's on pre2 now and awaiting the release of 3.1.0.

Thanks for helping to keep these gems moving!

Issue #2155, opened 2014-11-08
Method#super_method finds something when super is undef'ed


headius:
The last assertion in this test (from mri/ruby/test_method.rb) fails:

``` ruby
  def test_super_method_removed
    c1 = Class.new {private def foo; end}
    c2 = Class.new(c1) {public :foo}
    c3 = Class.new(c2) {def foo; end}
    c1.class_eval {undef foo}
    m = c3.instance_method(:foo)
    m = assert_nothing_raised(NameError, Feature9781) {break m.super_method}
    assert_nil(m, Feature9781)
  end
```

undef installs an UndefinedMethod, and Method#super_method checks for UndefinedMethod, so I'm not sure what's broken here. Here's the output

```
TestMethod#test_super_method_removed [/Users/headius/projects/jruby/test/mri/ruby/test_method.rb:857]:
[ruby-core:62202] [Feature #9781].
Expected #<UnboundMethod: #<Class:0x36c56a77>#foo> to be nil.
```

edipofederle:
@headius I investigate this one, and the issue occurs (`c1 = Class.new { def foo; end} `) when the `private` is used. In the description example, it works as expected without the `private`. I tried to debug it, but I still don't understand the issue. 

Maybe you have some place to look at? thx

PR #6728, opened 2021-06-25
Fix undef with private method

edipofederle:
Fixes: https://github.com/jruby/jruby/issues/2155


headius:
Looks good! I'll merge. Maybe there are specs that pass now too?

edipofederle:
> Looks good! I'll merge. Maybe there are specs that pass now too?

Thx! I will check if there some specs to untag.

PR #6743, opened 2021-07-05
[ci] do not rely on PATH and installed mvn

kares:
mostly, extracting CI bits from trying to setup ARM64 testing (which is smt I hope to follow up with)

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -13,7 +13,7 @@ cache:
 
 before_install:
   - unset _JAVA_OPTIONS
-  - rm ~/.m2/settings.xml
+  - rm -f ~/.m2/settings.xml
   - export MAVEN_SKIP_RC=true
   - export MAVEN_OPTS="-Xmn64M -Xmx512M -XX:CompressedClassSpaceSize=96M"
   - export JAVA_OPTS="$JAVA_OPTS -XX:CompressedClassSpaceSize=96M"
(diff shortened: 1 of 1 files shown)


kares:
interesting, this seems to be causing a regression in the regression spec suite:
```
Failures:

  1) #4122 Stdio streams in a simple non-native JRuby instance should have sane fileno

     Failure/Error: expect(filenos[0]).to eq 0

     

     NoMethodError:

       undefined method `[]' for nil:NilClass

     # ./spec/regression/GH-4122_stdio_fileno_sanity_spec.rb:10:in `block in <main>'


Finished in 1 minute 7.31 seconds (files took 2.49 seconds to load)

407 examples, 1 failure

Failed examples:

rspec ./spec/regression/GH-4122_stdio_fileno_sanity_spec.rb:8 # #4122 Stdio streams in a simple non-native JRuby instance should have sane fileno
```

headius:
omg I would love to get ant out of the Rake targets. Great progress!

headius (approved):

headius:
Looks good, but I am also confused about the failure. Seems like it is unable to launch JRuby.

headius:
I assume the failure has something to do with this line?

https://github.com/jruby/jruby/pull/6743/files#diff-6ac3f79fc25d95cd1e3d51da53a4b21b939437392578a35ae8cd6d5366ca5485L24

kares:
[that one last failure from Travis](https://github.com/jruby/jruby/pull/6743/checks?check_run_id=3181288460) is confusing as I do not see the job definition :open_mouth:

PR #6744, opened 2021-07-05
[ji] inspect for Thread/TimeZone/ZoneInfo

kares:
following up on https://github.com/jruby/jruby/pull/6493 with more `inspect` on Java types, namely:
- `java.lang.Thread`
- `java.util.TimeZone`
- `java.time.ZoneId`


headius (approved):
Looks good!

Issue #6745, opened 2021-07-06
exec on Windows doesn't find `.bat` files in PATH


deivid-rodriguez:
I _guess_ this might affect the whole `Process.spawn` family? Not really sure, I only tried `Kernel.exec`.

Rubygems installs `.bat` wrappers for gem executables on Windows to compensate for the lack of support for shebangs. They look like this:

```
@ECHO OFF
@"%~dp0ruby.exe" "%~dpn0" %*
```

On CRuby + Windows, `Kernel.exec` properly detects and uses these scripts:

```

PS V:\> ruby -e "Kernel.exec 'rake'"
-e:1:in `exec': No such file or directory - rake (Errno::ENOENT)
        from -e:1:in `<main>'

PS V:\> bundle
Fetching gem metadata from https://rubygems.org/
Fetching gem metadata from https://rubygems.org/.
Fetching rake 13.0.3
Installing rake 13.0.3
Using bundler 2.3.0.dev
Bundle complete! 1 Gemfile dependency, 2 gems now installed.
Use `bundle info [gemname]` to see where a bundled gem is installed.

PS V:\> ruby -e "Kernel.exec 'rake'"
rake aborted!
No Rakefile found (looking for: rakefile, Rakefile, rakefile.rb, Rakefile.rb)

(See full trace by running task with --trace)

PS V:\> rm C:\Ruby30-x64\bin\rake.bat

PS V:\> ruby -e "Kernel.exec 'rake'"
-e:1:in `exec': No such file or directory - rake (Errno::ENOENT)
        from -e:1:in `<main>'
```

However, on JRruby + Windows, they seem to be ignored:

```
PS V:\> jruby -e "Kernel.exec 'rake'"
Errno::ENOENT: No such file or directory - rake
  _exec_internal at org/jruby/RubyKernel.java:1767
            exec at uri:classloader:/jruby/kernel/kernel.rb:4
          <main> at -e:1

PS V:\> bundle
Fetching gem metadata from https://rubygems.org/
Fetching gem metadata from https://rubygems.org/.
Fetching rake 13.0.3
Installing rake 13.0.3
Using bundler 2.3.0.dev
Bundle complete! 1 Gemfile dependency, 2 gems now installed.
Use `bundle info [gemname]` to see where a bundled gem is installed.

PS V:\> jruby -e "Kernel.exec 'rake'"
Errno::ENOENT: No such file or directory - No message available
  _exec_internal at org/jruby/RubyKernel.java:1767
[...]

PR #6746, opened 2021-07-06
Fixes for Windows native exec

headius:
This PR contains various small fixes for the native `exec` support on Windows, mostly relating to finding the full paths to cmd.exe and the .exe/.com/.bat/.cmd executables that go with a given command.

See #6745.


headius:
I have added a small test for this behavior, but it will only run in the master/9.3 builds. The 9.2 branch has not received @kares fixes for running `test:jruby` on Windows.

I did confirm the test works on 9.2 branch.

Issue #6732, opened 2021-07-01
FFI issue loading libidn2 in macos 


HoneyryderChuck:
Hi,

Coming from [the ffi issue](https://github.com/ffi/ffi/issues/901), where maintainers suggested me to share with the Jruby community. Loading the `libidn2` lib on MacOS isn't working. It's working well in CRuby, whereas in JRuby, the library isn't found. I'm assuming there's not a complete overlap in the system lib paths.

Link to the failing CI build [here](https://github.com/HoneyryderChuck/idnx/runs/2849724922?check_suite_focus=true).

**Environment Information**
  
Provide at least:
* JRuby version: 9.2.18.0
* Operating system and platform: MacOS 10.15.7

headius:
Seems like some search logic for the library must be different? Can you give it a try on JRuby master (9.3)? We are closer to the FFI gem there (reusing all the .rb code as-is and passing all the same tests).

HoneyryderChuck:
I can confirm that the [suite passes against jruby-head](https://github.com/HoneyryderChuck/idnx/runs/2985525037?check_suite_focus=true). I'll close this then. Thank you for the tip!

headius:
I think it would be worth us backporting ffi for maintenance releases of 9.2. There weren't many changes and it would align is with master.

headius:
The backport of logic from master to use the FFI gem is complete in #6747.

PR #6747, opened 2021-07-07
Backport FFI from master

### LINKED ISSUE #6742: Backport FFI from master
For #6732 we should backport the FFI from master, which would let us align the .rb files.

headius:
This backports FFI from master so that we can source all Ruby
files from the gem rather than manually updating. This will fix
issue #6732 for JRuby 9.2.20.0 and higher.

Fixes #6742.


headius:
A bit of a snag here is that the ffi.rb from the gem is guarded to only be loadable on JRuby 9.3 or higher, due to 9.2 having an incompatible extension.

@enebo If we want to do this backport, we will need to get @larskanis to release a version of the ffi gem that lowers that guard to 9.2.20. Currently it is not possible to test this PR because ffi.rb does not load and it is needed by many libraries at boot.

The alternative would be to just update a minimal part of the ffi ruby code to satisfy #6732.

larskanis:
I can make a new ffi release with lowered JRuby check.

headius:
@larskanis Would you be willing to push a prerelease gem that lowers the check to 9.2.20? I am not yet sure whether we would make this move in that release or push out another lightweight fix release (for e.g. #6745) before making this larger change to FFI.

larskanis:
ffi-1.15.4.pre1 is pushed with the version guard lowered to 9.2.20.

larskanis:
@headius Do you still plan to backport changes to ffi to jruby-9.2.20? Or should I revert to 9.3.0 in the next ffi release?

headius:
@larskanis I will get back to this today or tomorrow, thank you for checking!

headius:
We'll see how CI goes with the prerelease FFI gem.

headius:
@larskanis This is looking pretty solid now, thank you for the help! I will merge this and update to the released 1.15.4 in some order. Just let me know when the full release is available.

PR #6753, opened 2021-07-12
add fast jit job

ahorek:
are these failures relevant?

https://app.travis-ci.com/github/jruby/jruby/jobs/523939968

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -92,6 +92,12 @@ matrix:
         - bundle install
         - jruby -S rake spec:ruby:fast
 
+    - name: spec/ruby fast jit
+      script:
+        - mvn clean package -Pbootstrap
+        - bundle install
+        - jruby -S rake spec:ruby:fast:jit
+
     - name: spec/ruby fast jdk11
       jdk: openjdk11
       script:


headius:
Ok back to work on this today.

headius:
First up:

```ruby
it "supports /o (once)" do
    2.times do |i|
      /#{i}/o.should == /0/
    end
  end
```

This may be an issue due to the transition from interpreter to JIT. I would expect that forced jit would compile this immediately, but perhaps it does not. Investigating further.

headius:
Confirmed:

```
$ jruby -Xjit.threshold=0 -Xjit.logging -e 'load "./blah.rb"'
2021-07-21T14:31:32.961-05:00 [main] INFO Ruby : done compiling target script: -e
2021-07-21T14:31:32.980-05:00 [main] INFO JITCompiler : block done jitting: Object _CLOSURE_1 at /Users/headius/projects/jruby/blah.rb:0
/0/
/1/
```

This is a known issue with block JIT. Because a jitted block needs to be called from a jitted outermost scope (or at least there is pre/post call logic that differs between interpreter and JIT), the actual compiled code does not get called until the *next* invocation. It shows up in the specs because the outermost scope is the toplevel of the file, which will never JIT on its own.

The problem is real but doesn't come up much in practice. Typically the `//o` regexp appears within a class body (e.g. a constant assignment) which will never transition to JIT code. There's still an issue for cases like this or method bodies under deferred JIT in that we cannot hand off the interpreter's already-compiled regexp to the jitted code.

I have one idea for improving this: when compiling a regexp with `//o` that has **already been executed once**, we should actually compile it as though it were a static regexp (i.e. compile in the cached expression rather than re-executing the dynamic regexp code again).

headius:
The fix I suggest is actually already in place! Looks like I did it over five years ago: https://github.com/jruby/jruby/commit/256e753aaa071fb2f8cb4fac6d2903744bdc36af

So the remaining issue is the jitted block transition. The above fix does not work under these specific conditions:

* JIT threshold set to 0
* Block executed from within a non-jitted scope

This is a fairly narrow situation that describes the spec. I will confer with @enebo about it.

headius:
I have pushed a fix in #6757, which pulls in your jit spec job. We will continue work there.

PR #6756, opened 2021-07-21
Fixes #6242. Sorbet srb init crashes

### LINKED ISSUE #6242: Sorbet `srb init` crashes
**Environment Information**
  
Provide at least:
* JRuby version: jruby 9.2.11.1 (2.5.7) 2020-03-25 b1f55b1a40 OpenJDK 64-Bit Server VM 25.252-b09 on 1.8.0_252-b09 +jit [darwin-x86_64]  
* Operating system and platform: Darwin MacBook-Pro.local 19.4.0 Darwin Kernel Version 19.4.0: Wed Mar  4 22:28:40 PST 2020; root:xnu-6153.101.6~15/RELEASE_X86_64 x86_64

* Installed or activated gems: sorbet, locked to `0.5.5675`
```
gem 'sorbet-runtime'
gem 'sorbet', :group => :development
```
* Application/framework: none
* Environment variables: nothing unusual here; RUBY_ROOT, GEM_ROOT, GEM_PATH...

**Expected Behavior**

* Running `bundle exec srb init` should set up sorbet's config and RBIs
* Followed the steps here: https://sorbet.org/docs/adopting

**Actual Behavior**

`srb init` consistently crashes with the following error:

```
Naming Object/Users/jakob/.gem/jruby/2.5.7/gems/sorbet-0.5.5675/lib/real_stdlib.rb:77: warning: constant ::NativeException is deprecated
Naming JRuby/Users/jakob/.gem/jruby/2.5.7/gems/sorbet-0.5.5675/lib/real_stdlib.rb:77: warning: constant JRuby::IR is deprecated
Generating /var/folders/j8/w10tktnj6514v501c6lh_kw80000gn/T/d20200526-75253-1t3issx/reflection.rbi with 700 modules and 26 aliases
TypeError: bind argument must be an instance of Class
             bind at org/jruby/RubyUnboundMethod.java:120
  real_superclass at /Users/jakob/.gem/jruby/2.5.7/gems/sorbet-0.5.5675/lib/real_stdlib.rb:62
  class_or_module at /Users/jakob/.gem/jruby/2.5.7/gems/sorbet-0.5.5675/lib/serialize.rb:50
   gen_source_rbi at /Users/jakob/.gem/jruby/2.5.7/gems/sorbet-0.5.5675/lib/hidden-definition-finder.rb:91
             each at org/jruby/RubyArray.java:1809
   gen_source_rbi at /Users/jakob/.gem/jruby/2.5.7/gems/sorbet-0.5.5675/lib/hidden-definition-finder.rb:90
   capture_stderr at /Users/jakob/.gem/jruby/2.5.7/gems/sorbet-0.5.5675/lib/hidden-definition-finder.rb:402
   gen_source_rbi at /Users/jakob/.gem/jruby/2.5.7/gems/sorbet-0.5.5675/lib/hidden-definition-finder.rb:89
             main at /Users/jakob/.gem/jruby/2.5.7/gems/sorbet-0.5.5675/lib/hidden-definition-finder.rb:45
             main at /Users/jakob/.gem/jruby/2.5.7/gems/sorbet-0.5.5675/lib/hidden-definition-finder.rb:38
        make_step at /Users/jakob/.gem/jruby/2.5.7/gems/sorbet-0.5.5675/bin/srb-rbi:232
             init at /Users/jakob/.gem/jruby/2.5.7/gems/sorbet-0.5.5675/bin/srb-rbi:121
             main at /Users/jakob/.gem/jruby/2.5.7/gems/sorbet-0.5.5675/bin/srb-rbi:196
           <main> at /Users/jakob/.gem/jruby/2.5.7/gems/sorbet-0.5.5675/bin/srb-rbi:237
```

enebo:
The fix here is to not register JavaPackage as a constant of Java.
This prevents srb from finding it via constants.  So the problem class
is never seen.

Note: I think in the scenario of walking ObjectSpace we can still run
across this but that may be another issue to deal with when we experience it.

The explanation of what is wrong is that JavaPackage is in fact a Class but
due to how we want it to behave we set its metaclass to Module.  When srb
retrieves an UnboundMethod off of Class that method cannot be bound to this
type because it says "nope I am a module" (even though it is actually a class).

### DIFF
--- a/core/src/main/java/org/jruby/javasupport/JavaSupport.java
+++ b/core/src/main/java/org/jruby/javasupport/JavaSupport.java
@@ -148,7 +148,7 @@ public abstract class JavaSupport {
     public abstract RubyClass getJavaClassClass();
 
     public abstract RubyClass getJavaPackageClass();
-    public abstract void setJavaPackageClass(RubyClass javaPackageClass);
+    abstract void setJavaPackageClass(RubyClass javaPackageClass);
 
     public abstract RubyModule getJavaInterfaceTemplate();
 
--- a/core/src/main/java/org/jruby/javasupport/JavaSupport.java
+++ b/core/src/main/java/org/jruby/javasupport/JavaSupport.java
@@ -147,7 +147,8 @@ public abstract class JavaSupport {
 
     public abstract RubyClass getJavaClassClass();
 
-    public abstract RubyClass getJavaPackageClass() ;
+    public abstract RubyClass getJavaPackageClass();
+    public abstract void setJavaPackageClass(RubyClass javaPackageClass);
 
     public abstract RubyModule getJavaInterfaceTemplate();
 
--- a/core/src/main/java/org/jruby/javasupport/Java.java
+++ b/core/src/main/java/org/jruby/javasupport/Java.java
@@ -108,7 +108,7 @@ public class Java implements Library {
     public void load(Ruby runtime, boolean wrap) {
         final RubyModule Java = createJavaModule(runtime);
 
-        JavaPackage.createJavaPackageClass(runtime, Java);
+        runtime.getJavaSupport().setJavaPackageClass(JavaPackage.createJavaPackageClass(runtime, Java));
 
         org.jruby.javasupport.ext.Kernel.define(runtime);
         org.jruby.javasupport.ext.Module.define(runtime);
--- a/core/src/main/java/org/jruby/javasupport/JavaPackage.java
+++ b/core/src/main/java/org/jruby/javasupport/JavaPackage.java
@@ -58,7 +58,7 @@ import static org.jruby.runtime.Visibility.PRIVATE;
 @JRubyClass(name="Java::JavaPackage", parent="Module")
 public class JavaPackage extends RubyModule {
 
-    static RubyModule createJavaPackageClass(final Ruby runtime, final RubyModule Java) {
+    static RubyClass createJavaPackageClass(final Ruby runtime, final RubyModule Java) {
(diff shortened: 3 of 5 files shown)


enebo:
@kares I think one big decision here is that Java::JavaPackage is no longer something we would care about seeing within Ruby code.  This makes some of my test changes in the higher level tests suspect.  It also makes JavaPackage even weirder because it thinks it exists but we have no simple way within Ruby to get an instance of it.

core/src/main/java/org/jruby/javasupport/JavaSupport.java
```diff
@@ -147,7 +147,8 @@ public synchronized RubyModule computeValue(Class<?> klass) {
 
     public abstract RubyClass getJavaClassClass();
 
-    public abstract RubyClass getJavaPackageClass() ;
+    public abstract RubyClass getJavaPackageClass();
+    public abstract void setJavaPackageClass(RubyClass javaPackageClass);
```
kares, line 151: should work if it's package private - not really public API
```suggestion
    abstract void setJavaPackageClass(RubyClass javaPackageClass);
```

kares (approved):
LGTM, tried to find anything on GH but seems JRuby was the only user of `Java::JavaPackage`.

if someone comes around we should consider having a blessed API to check for something being a Java package e.g. `JRuby.java_package?(some_type)` ... hopefully no one needs that :wink:

headius (approved):

headius:
Simple fix is best fix... ship it!

PR #6757, opened 2021-07-21
Fix failures in ruby/spec running with immediate jit mode

headius:
See #6753 for the PR that initiated this work, by @ahorek. That PR has been superseded by this one, which includes the added jit job and will eventually contain all fixes to get that job green.


headius:
Fix for the first failure from #6753 is already in this PR. The failure looked like this:

```
Regexps with modifiers supports /o (once) FAILED
Expected /1/ == /0/
to be truthy but was false
/home/travis/build/jruby/jruby/spec/ruby/language/regexp/modifiers_spec.rb:20:in `block in <main>'
```

The second failure looks like this:

```
Literal Regexps caches the Regexp object FAILED
Expected /foo/ to be identical to /foo/
/home/travis/build/jruby/jruby/spec/ruby/language/regexp_spec.rb:32:in `block in <main>'
```

This one may not be fixable currently due to the interpreted code and jitted code not sharing a cache of objects.

PR #6778, opened 2021-08-18
use optimized CGIEscape

ahorek:
builds on the top of https://github.com/jruby/jruby/pull/6777 that should fix the issue with prepends

see https://github.com/jruby/jruby/pull/5949


ahorek:
ok, confirmed locally and relevant tests passed. Even if it's still slower than the recent MRI, this change will improve escape/unescape performance which is important for ERB-like templates by 3 times.

I think we can include it to 9.3, once tests on https://github.com/jruby/jruby/pull/6777 are green. Thanks!

PR #6789, opened 2021-08-27
Fix detection of modularized java in mybinder environments

Ifiht:
When using JRUBY in a mybinder.org container you get the following error due to an unconventional $JAVA_HOME path (`/srv/conda/envs/notebook`):

`grep: /srv/conda/envs/notebook/release: No such file or directory`

I think I did a decent proof within a small mybinder environment here, the steps can be replicated by running the jupyter notebook but the gist is that I added parenthesis around the last two conditions to force order of evaluation. I also reversed order so bash would evaluate my new expression first, which it does fine. Seems to work from my tests:
[jruby-mybinder](https://github.com/Ifiht/jruby-mybinder)

### DIFF
--- a/bin/jruby.bash
+++ b/bin/jruby.bash
@@ -160,7 +160,7 @@ else
 fi
 
 # Detect modularized Java
-if [ -f "$JAVA_HOME"/lib/modules ] || [ -f "$JAVA_HOME"/release ] && grep -q ^MODULES "$JAVA_HOME"/release; then
+if ([ -f "$JAVA_HOME"/release ] && [ grep -q ^MODULES "$JAVA_HOME"/release ]) || [ -f "$JAVA_HOME"/lib/modules ]; then
   use_modules=1
 fi
 


headius:
There appears to be some syntax issues here:

```
/home/travis/build/jruby/jruby/bin/jruby: line 163: [: too many arguments
```

Is there a reason the `||` should be reversed?

headius:
What causes the grep error? Is the `release` file in your environment nonstandard in some way? Perhaps we just need to narrow our test here from being a file to being readable, or something?

Ifiht:
@headius thanks for the feedback, not sure why #163 is giving too many args since it works in my clean-install bash environment, as for the grep error it is due to non-standard $JAVA_HOME and many other env vars in the binder environment, which is just a basic container environment.

There isn't a need necessarily to reverse the statement, just to add the parens I think? Would this commit pass if I reverted it back to the original order?

headius:
> There isn't a need necessarily to reverse the statement, just to add the parens I think? Would this commit pass if I reverted it back to the original order?

I am not sure. I believe the error is due to the grep command being within `[]` which makes the `test` command used for it to think those are arguments. If you remove the brackets it may go back to erroring because the `grep` command will start running correctly.

The `grep` should only be running when the expected `release` file exists, due to the `[ -f ... ]` part, so either the `&&` is not short-circuiting (and the grep is running even if the `release` file does not exist) or the file it finds is peculiar in some way (triggering the `grep` error.

Could you provide the following information please?

* What is this `JAVA_HOME` set to?
* Show me the `ls -l` of that `JAVA_HOME`.

We'd love to get this fixed for future releases!

headius:
Calling @mrnoname1000! What are we doing wrong here?

headius:
Punting this until we have some time to reproduce in the prescribed environment. It would be a minor update to the bash script, in any case.

Ifiht:
@headius sorry for the delay, ironically it's been the compsci schoolwork keeping me busy...

Please see the image below, again if anyone on the team want's to test the environment for themselves I setup a sandbox here: [JRuby myBinder](https://hub.gke2.mybinder.org/user/ifiht-jruby-mybinder-necq3kdy/lab)

image

headius:
This got left behind a bit but we'll try to patch the launchers for 9.3.3.

headius:
Apologies for delays looking into this! We would still like to patch the launcher to work better for you.

Unfortunately it appears the mybinder link you provided has gone away so I cannot test in that environment. I am doing some testing locally to see if those conditions in the bash script are constructed properly.

headius:
I think I managed to reproduce this!

```
[] ~/projects/jruby-9.3 $ ls -lR blah
total 0
drwxr-xr-x  3 headius  staff  96 Jan 17 12:30 lib

blah/lib:
total 0
-rw-r--r--  1 headius  staff  0 Jan 17 12:29 modules

[] ~/projects/jruby-9.3 $ if [ -f blah/lib/modules ] || [ -f blah/release ] && grep -q ^MODULES blah/release; then   echo ok; fi
grep: blah/release: No such file or directory
```

I believe the problem is the order in which the conditionals are applied. In this case, there is a `blah/lib/modules` file but no `blah/release` file. The `||` still ends up true due to the `modules` file, so the && proceeds to try to grep the nonexistent `release` file and there we are.

If I group the `&&` conditionals together I think it will fix the problem.

headius:
Here's a stackexchange post that backs up my theory:

https://unix.stackexchange.com/questions/88850/precedence-of-the-shell-logical-operators

I have a patch on the way, so we will close this.

headius:
See #7004 for the fix we will release in 9.2.21 and 9.3.3.

PR #6791, opened 2021-08-31
restore api compatibility with nio4r

ahorek:
introduced by https://github.com/jruby/jruby/commit/84ece2bba5ee9824fefb9f805df6314298154764

blockingThreadPoll was a public API and nio4r / puma use the method at https://github.com/socketry/nio4r/blob/master/ext/nio4r/org/nio4r/Selector.java#L237


headius (approved):
D'oh!

headius:
Perhaps we should keep adding key libraries to our CI, so we can build up more confidence here... especially those with extensions.

PR #6794, opened 2021-09-01
Update to jruby-openssl 0.10.7

headius:
This updates 9.3 to jruby-openssl 0.10.7.

@kares Any reason we shouldn't do this?


kares (approved):
Should be good to go.

PR #6795, opened 2021-09-01
Update to 6.1.2.1 to match Ruby 2.6.8

headius:
This updates JRuby 9.3 to rdoc 6.1.2.1, which would match Ruby 2.6.8.

Unfortunately there is no release of rdoc 6.1.2.1, so this will initially fail to build.


headius:
Link to ruby/rdoc#835.

PR #6796, opened 2021-09-01
Update rexml to 3.1.9.1 to match ruby 2.6.8

headius:
This updates rexml to 3.1.9.1 which matches Ruby 2.6.8.

Unfortunately rexml 3.1.9.1 has not been released, so this will initially fail.


headius:
Link to https://github.com/ruby/rexml/issues/79.

headius:
rexml 3.1.9.1 has been released for ruby/rexml#79 and I have repushed to pick up the change.

Issue #6801, opened 2021-09-02
fileutils from Ruby 2.6.8 is out of sync with released 1.4.1


headius:
CRuby made a change to their copy of fileutils 1.4.1 and released that change in CRuby 2.6.8. However this change is not in the released fileutils gems until 1.5.0, which includes other changes.

I have filed https://github.com/ruby/fileutils/issues/59 on the fileutils repo. We may need to move to fileutils 1.5.0 to pick up this change if they are unwilling to release the exact version contained in Ruby 2.6.8 as a new gem version.

PR #6802, opened 2021-09-02
Update stdlib

headius:
This updates our non-gem stdlib files to match CRuby 2.6.8, other than two unreleased changes in fileutils and rdoc.

See https://github.com/jruby/jruby/issues/6801 and https://github.com/jruby/jruby/pull/6795.

I also updated the appropriate Ruby version values to reflect CRuby 2.6.8.


ahorek:
ftp tests sync should fix the failure https://github.com/ruby/net-ftp/commit/5709ece67cf57a94655e34532f8a7899b28d496a

headius:
@ahorek Good catch, I'll include that in this PR.

PR #6803, opened 2021-09-02
More agressively kill dead code

enebo:
LocalOptimizationPass will propagate values.  For example:

```ruby
_tmp = true
unless _tmp
 puts "Never"
end
```

In this case by the end of this pass we will see a BB end with a
b_false(true, some_label).  This clearly cannot happen.  If we eliminate
this instr then we should also be able to lop off the bb(s) representing
whatever is within the unless conditional.  In this snippet the entire
snippet should be dead code except _tmp = true.

To fix this I added simplifyBranch to BranchInstr.  This gives us the
ability to swap out a branch with a noop or a jump (in case where something
is unconditionally true).

After we replace the instr with a new one we need to reexamine the basicblock
it is in and correct any edge placement.

### DIFF
--- a/core/src/main/java/org/jruby/ir/IRManager.java
+++ b/core/src/main/java/org/jruby/ir/IRManager.java
@@ -62,7 +62,7 @@ public class IRManager {
 
     private int dummyMetaClassCount = 0;
     private final IRModuleBody object;
-    private final Nil nil = new Nil();
+    private final Nil nil = Nil.NIL;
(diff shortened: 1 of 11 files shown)


headius (approved):

kares:
nice, this could also work if arg1 is simply a `nil`, right?

enebo:
@kares I am in the process of simplifying all branch types.

PR #6805, opened 2021-09-03
Update tests from MRI 2.6.8

headius:
These are not all passing, due to a combination of unfixed features in JRuby and modifications to CRuby tests that require e.g. RubyVM features we do not support. This PR will track fixes, test modifications, or excludes to get green on the latest 2.6 suite.


headius:
Obsoleted by #7141.

PR #6814, opened 2021-09-08
Revert "update surefire plugin"

headius:
Reverts jruby/jruby#6788

I think this may have disabled tests, possibly due to this JUnit/Surefire conflict:

https://dzone.com/articles/why-your-junit-5-tests-are-not-running-under-maven

The latest JNR update (#6481) may have surfaced this issue by depending on a different version of JUnit and thereby letting tests run again.

The failures in that PR, then, might represent a valid regression that was hidden due to the tests not running.

Trying this reversion to see if we see tests running again on master, and if so whether they have the same failures as on the JNR update PR.

### DIFF
--- a/pom.rb
+++ b/pom.rb
@@ -120,7 +120,7 @@ project 'JRuby', 'https://github.com/jruby/jruby' do
 
     plugin :compiler, '3.8.1'
     plugin :shade, '3.2.4'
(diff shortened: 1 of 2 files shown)


headius:
This seems to be unrelated and the tests still do not run. I have localized it to the previous JNR update on Sept 1 (#6675).

Issue #6818, opened 2021-09-09
GC.start raises ArgumentError when passed named arguments


JasonLunn:
* JRuby version jruby-9.2.11.1
* Debian rodete 5.10.40-1rodete2 (2021-06-22) x86_64 GNU/Linux

**Expected Behavior**
Per the docs for `GC.start` and `GC.garbage_collect`, starting in version [2.1](https://ruby-doc.org/core-2.1.0/GC.html#method-c-start) these methods accept a dictionary of named arguments.

I understand that JRuby does not do anything with these values, however, calling these methods with named arguments currently raises an exception because neither are expecting any arguments.

**Actual Behavior**
`GC.start(full_mark: true, immediate_sweep: true)` raises an exception:

```ArgumentError: wrong number of arguments calling `start` (given 1, expected 0)```

headius:
Thanks for the help!

PR #6819, opened 2021-09-09
Add optional argument to start and garbage_collect

JasonLunn:
Proposed fix for #6818

### DIFF
--- a/core/src/main/java/org/jruby/RubyGC.java
+++ b/core/src/main/java/org/jruby/RubyGC.java
@@ -64,13 +64,13 @@ public class RubyGC {
         return result;        
     }
 
-    @JRubyMethod(module = true, visibility = PRIVATE)
-    public static IRubyObject start(ThreadContext context, IRubyObject recv) {
+    @JRubyMethod(module = true, visibility = PRIVATE, optional = 1)
+    public static IRubyObject start(ThreadContext context, IRubyObject recv, IRubyObject[] args) {
         return context.nil;
     }
 
-    @JRubyMethod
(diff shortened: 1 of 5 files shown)


ahorek:
could you also change it at https://github.com/jruby/jruby/blob/master/core/src/main/java/org/jruby/RubyObjectSpace.java#L194 and remove a tag https://github.com/jruby/jruby/blob/master/spec/tags/ruby/core/gc/start_tags.txt#L1 ?

JasonLunn:
@ahorek - thanks for the pointers - added your suggestions.
@headius - There was no spec tag to remove for the missing keyword argument support for ObjectSpace, so I assume that means there was no compliance test that was failing without it. Does one add those such a test directly in the JRuby repo, or does one need to contribute that upstream somewhere?

ahorek:
https://github.com/ruby/spec/blob/master/core/objectspace/garbage_collect_spec.rb#L3 is the right place

ahorek:
one more place https://github.com/jruby/jruby/blob/master/spec/jruby.mspec#L14

headius:
You can also add new specs to our repository directly, under `spec/ruby` which is equivalent to the https://github.com/ruby/spec repository.

FWIW, as you may have noticed GC.start is a no-op on JRuby, since the JVM GC is a very different beast than the CRuby GC and we do not really have direct control over it. I don't see us changing that any time soon.

JasonLunn:
Updated `jruby.mspec` and opened https://github.com/ruby/spec/pull/849

PR #6823, opened 2021-09-14
Fixes to run tests on Apple Silicon

headius:
We now have a self-hosted runner thanks to MacStadium's OSS program. Try to get a job running.


headius:
I've merged this with `mvn -Ptest` and `rake spec:ffi` running on our M1 instance. Other jobs will follow and I will keep the branch for ongoing M1 CI coverage improvements.

PR #6827, opened 2021-09-14
reduce some repeated method calls

ahorek:
just a small cleanup


headius:
Odd hang in rubyspec with JIT. I restarted it, we'll watch for it.

Issue #6831, opened 2021-09-15
require "json" outputs warnings


yamam:
**Environment Information**
  
```
jruby 9.3.0.0-SNAPSHOT (2.6.8) 2021-09-08 1420b569ca OpenJDK 64-Bit Server VM 11.0.11+9-Ubuntu-0ubuntu2.20.04 on 11.0.11+9-Ubuntu-0ubuntu2.20.04 +jit [linux-x86_64]
```

**Expected Behavior**

```#jruby -we "require 'json'"```

**Actual Behavior**

```
#jruby -we "require 'json'"
/jruby-head/lib/ruby/stdlib/ostruct.rb:446: warning: OpenStruct#define_singleton_method accesses caller method's state and should not be aliased
/jruby-head/lib/ruby/stdlib/ostruct.rb:446: warning: OpenStruct#!~ accesses caller method's state and should not be aliased
/jruby-head/lib/ruby/stdlib/ostruct.rb:446: warning: OpenStruct#=~ accesses caller method's state and should not be aliased
/jruby-head/lib/ruby/stdlib/ostruct.rb:446: warning: OpenStruct#instance_exec accesses caller method's state and should not be aliased
/jruby-head/lib/ruby/stdlib/ostruct.rb:446: warning: OpenStruct#instance_eval accesses caller method's state and should not be aliased
```

**Comment by headius** [comment]
This is likely due to a newer ostruct aliasing methods that access the caller's frame. I will look into this and see why they are doing it.

**Comment by headius** [comment]
We can work around this by reverting ostruct to 0.1.0, which is what Ruby 2.6.8 ships. With that change, the errors are gone.

```
$ jruby -we 'require "ostruct"; p OpenStruct.methods.grep("!")'
[]
```

headius:
I'm going to mark this as fix in 9.3.1.0 since reverting the gem to an earlier version just delays the problem. ruby/ostruct#30 will need to be addressed, or we will need a more permanent workaround for this aggressive aliasing.

headius:
This is still waiting on a verdict in ruby/ostruct#30 so we'll have to punt again.

I did merge #6842 into 9.3, which allows aliasing one frame-aware name to another frame-aware name (with equivalent framing requirements) but that does not fully resolve ruby/ostruct#30 without us hardcoding the special aliased names it creates for e.g. `instance_eval`/`instance_eval!`.

headius:
[...]

PR #6835, opened 2021-09-15
Revert ostruct tests/specs due to reverted gem

headius:
See https://github.com/ruby/ostruct/issues/30 for the ostruct issue that caused us to revert to an earlier gem. This PR reverts the test and tags the specs that do not pass on the older version of ostruct.

See also #6831.


headius:
Ok, there's yet another out of sync gem here. The failing test is due to CRuby shipping a custom ostruct that does not match the 0.1.0 release they say it is.

Issue #6832, opened 2021-09-15
"Syntax Error" occurs in "eval"


yamam:
**Environment Information**
  
```jruby 9.3.0.0-SNAPSHOT (2.6.8) 2021-09-08 1420b569ca OpenJDK 64-Bit Server VM 11.0.11+9-Ubuntu-0ubuntu2.20.04 on 11.0.11+9-Ubuntu-0ubuntu2.20.04 +jit [linux-x86_64]```

**Expected Behavior**
utf8.txt
```
"\u3042\u3044\u3046\u3048\u304A"
```

There are no errors with ruby 3.0.2 and jruby 9.2.19.0.
```
#ruby -ve 'eval(File.read("utf8.txt").force_encoding("Windows-31J"))'
ruby 3.0.2p107 (2021-07-07 revision 0db68f0233) [x86_64-linux]

#jruby -ve 'eval(File.read("utf8.txt").force_encoding("Windows-31J"))'
jruby 9.2.19.0 (2.5.8) 2021-06-15 55810c552b OpenJDK 64-Bit Server VM 11.0.11+9-Ubuntu-0ubuntu2.20.04 on 11.0.11+9-Ubuntu-0ubuntu2.20.04 +jit [linux-x86_64]
```
**Actual Behavior**

```
# jruby -ve 'eval(File.read("utf8.txt").force_encoding("Windows-31J"))'
jruby 9.3.0.0-SNAPSHOT (2.6.8) 2021-09-08 1420b569ca OpenJDK 64-Bit Server VM 11.0.11+9-Ubuntu-0ubuntu2.20.04 on 11.0.11+9-Ubuntu-0ubuntu2.20.04 +jit [linux-x86_64]
SyntaxError: (eval):1: UTF-8 mixed within Windows-31J source
"\u3042\u3044\u3046\u3048\u304...^
    eval at org/jruby/RubyKernel.java:1091
  <main> at -e:1
```

headius:
So this is reading in UTF-8 content, forcing it to have an encoding of Windows-31J, and then expects it to parse? Why should that work?

PR #6837, opened 2021-09-15
 Fix string literals in other encodings which should be utf-8 encoding

enebo:
We have the test of time just created a new string using the lexers
 encoding.  This worked until I fixed the last known issues in the
 parser reporting errors with mixed strings (like binary data + UTF-8
 escapes).  This is not a valid string and the parser should error out.

 Unfortunately, by setting the lex encoding right away we are not in a
 position of knowing whether the string should just have that encoding
 or whether we are in a new string where we have not yet figured out
 what encoding it should be.

 Case in point.  #6832.  It is evaling as a Window-31J string BUT the
 literal string is only utf-8 escapes.  It should resolve as being
 UTF-8 yet it errors because it sees the escapes and then errors out
 thinking it is already a windows-31J string.

 The solution is to emulate how MRI does this by indicating that we
 have not yet determined the encoding for the string and then if nothing
 special happens during the processing we just set it to the lex encoding.

### DIFF
--- a/core/src/main/java/org/jruby/lexer/yacc/HeredocTerm.java
+++ b/core/src/main/java/org/jruby/lexer/yacc/HeredocTerm.java
@@ -176,11 +176,12 @@ public class HeredocTerm extends StrTerm {
                 tok.append('#');
             }
 
+            boolean encodingDetermined[] = new boolean[] { false };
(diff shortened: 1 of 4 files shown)


enebo:
All failures seem to be related to picking up failures on master from openstruct.

PR #6839, opened 2021-09-16
load all socket AI_ constants

ahorek:
https://github.com/jnr/jnr-constants/blob/master/src/main/java/jnr/constants/platform/linux/AddressInfo.java


headius:
This has lingered for a while, and I'm of the opinion now we should only add the constants that we can actually support. If there are constants preventing third-party libraries or apps from running, we should try to implement those on a case-by-case basis. Adding the constants and then doing nothing with them will be difficult to diagnose and misleading for users.

PR #6840, opened 2021-09-16
load all socket IP_ constants

ahorek:
https://github.com/jnr/jnr-constants/blob/master/src/main/java/jnr/constants/platform/linux/IP.java

there's a risk that some of these constants don't have to be usable, so it could break feature checks like
```
if Socket.const_defined?(:Socket::IP_TTL)
  @sock.getsockopt(Socket::IPPROTO_IP, Socket::IP_TTL)
end
```

* to consider


headius:
Still not sure what to do with these. We can define the constants but if we don't handle them it feels misleading. I'm going to punt one more time to 9.4.1.0 and we'll make a decision then.

headius:
https://github.com/jruby/jruby/pull/6839#issuecomment-1418662542

PR #6866, opened 2021-09-29
Fix #6864. include? on an String Range always returns true.

### LINKED ISSUE #6864: Calling :include? on an String Range always returns true in 9.3.0.0
**Environment Information**
```Text
JRuby version:
jruby 9.3.0.0 (2.6.8) 2021-09-17 85c20e780f OpenJDK 64-Bit Server VM 11.0.11+9-Ubuntu-0ubuntu2.20.04 on 11.0.11+9-Ubuntu-0ubuntu2.20.04 +jit [linux-x86_64]
OS:
Ubuntu 20.04
Linux T480 5.4.0-86-generic #97-Ubuntu SMP Fri Sep 17 19:19:40 UTC 2021 x86_64 x86_64 x86_64 GNU/Linux
```
**Expected Behavior**
I expect the :include? method to have the same behavior as with JRuby 9.2.19 or MRI and return false.
```Ruby
("01".."10").include?("test")
=> false
("01".."10").include?("01")
=> true
```

**Actual Behavior**
It looks like :include? always returns true when called on an String Range. :include? still works when it is called on an array, so converting the String Range into an Array and then calling :include? is working fine.
```Ruby
# JRuby 9.3.0.0
("01".."10").include?("test")
=> true
("01".."10").to_a.include?("test")
=> false
# JRuby 9.2.19.0
("01".."10").include?("test")
=> false
# MRI 2.6.8
("01".."10").include?("test")
=> false
```

enebo:
Looks like during a refactoring strings >1 char passed beg as
test value instead of the actual test value.

### DIFF
--- a/c.diff
+++ b/c.diff
@@ -1,0 +1,20 @@
+diff --git a/core/src/main/java/org/jruby/ir/IREvalScript.java b/core/src/main/java/org/jruby/ir/IREvalScript.java
+index 3fbfc9e7ca1..6f7a926cb65 100644
+--- a/core/src/main/java/org/jruby/ir/IREvalScript.java
++++ b/core/src/main/java/org/jruby/ir/IREvalScript.java
+@@ -8,6 +8,15 @@ import org.jruby.util.ByteList;
+ public class IREvalScript extends IRClosure {
+     private String fileName;
+ 
++    public int getNextClosureId() {
++        nextClosureIndex++;
++
++        return nextClosureIndex;
++    }
++
++    // Index values to guarantee we don't assign same internal index twice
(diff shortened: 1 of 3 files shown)


c.diff
```diff
@@ -0,0 +1,20 @@
+diff --git a/core/src/main/java/org/jruby/ir/IREvalScript.java b/core/src/main/java/org/jruby/ir/IREvalScript.java
```
edipofederle, line 1: @enebo what is this diff file used for?

enebo: @edipofederle HAHAHA. thanks I don't know how I ended up committing that file.

PR #6875, opened 2021-10-05
Update to ruby/spec@ccf0d85

eregon:


eregon:
FYI one of the new specs triggers a NPE in Marshal.load: https://app.travis-ci.com/github/jruby/jruby/jobs/541498058

PR #6877, opened 2021-10-05
Fix raw and varargs backtrace elements

### LINKED ISSUE #6846: Unexpected frame in stack trace from `caller` method with create_method invocation
**Environment Information**
  
Provide at least:
`jruby 9.3.0.0-SNAPSHOT (2.6.8) 2021-09-18 ffffffffff OpenJDK 64-Bit Server VM 25.272-b10 on 1.8.0_272-b10 +jit [linux-x86_64]`

(built from source from most recent commit on master: 4727fcb3b037537ceecc67d771aa338bd5188af4 using a simple nix derivation) 
`Linux hostname 5.10.40-1rodete2-amd64 #1 SMP Debian 5.10.40-1rodete2 (2021-06-22) x86_64 GNU/Linux`

(glinux) 

**Expected Behavior**

we’re doing some dynamic validation detection of routes, and in doing so checking the call stack generated by the `caller` method. 

in 9.3.0.0 there appears to be an unexpected stack frame. 

This is not a repro, but a digestion of some code we hae. 

```
0  # file.rb
1  [:get, :patch].each do |verb| # we have some others here too
2    real_method = :"real_#{verb}"
3    alias_method real_method, verb
4    define_method(verb) do |path = nil, opts = {}, &block|
5      create_route(verb, real_method, path, opts, 1, &block)
6    end
7  end
8
9  def create_route(...)
10   # method call that calls caller
11 end
12
13 # caller_file.rb
14
15 get '/foo' do 
16   ...
17 end
18
19 patch '/bar' do 
20   ...
21 end
```

when patch is called, stack trace looks like
```
file.rb:10:in `create_route'
file.rb:9:in `create_route'
file.rb:5:in `block in patch'
caller_file.rb:38:in `<class:Engine>'
```

in this case, the frame that references file.rb:9 is unexpected. 

when get it called stack trace looks like: 
```
file.rb:10:in `create_route'
file.rb:5:in `block in get'
caller_file.rb:15:in `<class:Engine>'
```

and it doesn’t have this added frame (as expected).

headius:
Two fixes here:

* Fix the "raw" backtrace mode. Both "raw" and "full" were printing out all frames, but "raw" was improperly processing jitted frames into their Ruby method names like "full". I believe this was due to my unifying these two paths; "raw" passed in an empty set of interpreter frames, which prevented interpreter frames from being processed, but it still processed jitted method names. A new flag was added to indicate "raw" behavior, now skipping the jit frame rewriting.
* Fix the varargs method wrappers from JIT being included in exception traces. In #6771 I fixed the backtrace generation to include the JIT varargs wrapper, which previously had been ignored. However I messed up the guard that prevents those frames from being processed if the wrapped method also appears in the trace (varargs frame should only show up if it appears alone, as in arity errors). The fix is to remove the expectation of the "varargs" frame marker, since they still show up as "method" type. Fixes #6846.

### DIFF
--- a/core/src/main/java/org/jruby/runtime/ThreadContext.java
+++ b/core/src/main/java/org/jruby/runtime/ThreadContext.java
@@ -925,7 +925,7 @@ public final class ThreadContext {
     public static String createRawBacktraceStringFromThrowable(final Throwable ex, final boolean color) {
         return WALKER.walk(ex.getStackTrace(), stream ->
             TraceType.printBacktraceJRuby(null,
-                    new BacktraceData(stream, Stream.empty(), true, false, false).getBacktraceWithoutRuby(),
+                    new BacktraceData(stream, Stream.empty(), true, true, false, false).getBacktraceWithoutRuby(),
                     ex.getClass().getName(),
                     ex.getLocalizedMessage(),
                     color));
--- a/core/src/main/java/org/jruby/runtime/backtrace/BacktraceData.java
+++ b/core/src/main/java/org/jruby/runtime/backtrace/BacktraceData.java
@@ -96,8 +96,7 @@ public class BacktraceData implements Serializable {
 
                     if (decodedName != null) {
                         // skip varargs frames if we just handled the method's regular frame
-                        if (type == FrameType.VARARGS_WRAPPER &&
-                                previousElement != null &&
+                        if (previousElement != null &&
                                 element.getMethodName().equals(previousElement.getMethodName() + JavaNameMangler.VARARGS_MARKER)) {
 
                             previousElement = null;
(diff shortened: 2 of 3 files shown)


headius:
Second fix in action... both of these show just one "foo" frame even though they both pass through the varargs frame. The first triggers an arity error within the varargs wrapper, and the second raises within the real method body. The second case is taken from #6846.

```
$ jruby -e 'def foo(a); raise; end; def bar; a = [1,2]; foo(*a); end; bar'
ArgumentError: wrong number of arguments (given 2, expected 1)
     foo at -e:1
     bar at -e:1
  <main> at -e:1

$ jruby -e 'def foo(a); raise; end; def bar; a = [1]; foo(*a); end; bar'
RuntimeError: No current exception
     foo at -e:1
     bar at -e:1
  <main> at -e:1
```

Issue #6362, opened 2020-08-21
rbconfig and other built-in libraries don't load without stdlib


headius:
In #6340 I found that one of our tests tries to clear the `jruby.home` property and then `require "rbconfig"`. However we load the rbconfig extension via rbconfig.rb here: https://github.com/jruby/jruby/blob/master/lib/ruby/stdlib/rbconfig.rb

There may be other built-in libraries that depend on a single `load_ext` in stdlib.

We should consider moving such files into the base jruby.jar file, since they are frequently necessary for normal operation of Ruby, with or without a standard library present.

kares:
good find, makes sense and I am surprised the tests did not fail before.
as this was [changed](https://github.com/jruby/jruby/commit/fb3ba4b3543b72e735ba53b51e6b7ec8a9916e1b#diff-a40568b9da02963919222c8b1eb53774L1752) due bootsnap (in 9.2.6.0) to always have explicit *library.rb* files around compared to "virtual" .rb ones.

headius:
@kares Looks like you have this in hand, let me know if you want a review.

kares:
@headius not really, have only been adjusting tests to not assume (previous) built-in lazy lib behavior at https://github.com/jruby/jruby/pull/6340

headius:
@kares Ahh ok... so this is still an open task.

kares:
for (explicit) reference, here's the commit changing the tested behavior: https://github.com/jruby/jruby/commit/a6b428ba47b647da4336353b8a754913277d9028

headius:
On second thought this will drastically change in 9.4 so there's not much value in doing it now. Moving this and the PR to 9.4 to be reexamined once more libraries have moved out to gems.

headius:
I did a bit of spelunking and the original reason we moved to having mostly-empty .rb files in stdlib was to address [issues integrating with load-caching tools like bootsnap](https://github.com/Shopify/bootsnap/issues/162): https://github.com/jruby/jruby/commit/fb3ba4b3543b72e735ba53b51e6b7ec8a9916e1b

I'm not sure if there's a resolution that removes the .rb files from the filesystem without breaking the expectation of files-on-filesystem. The workaround, for now, is just using jruby-complete (or the jruby-stdlib jar combined with jruby core jars) so the files in question can be found without being on the filesystem, but I would assume running in this fashion probably isn't compatible with bootsnap.

[...]

PR #6884, opened 2021-10-08
Restructure built-in libs to load from jar

headius:
This attempts to address #6362 by doing the following:

* All ext-loading stubs are duplicated in the jar under the same path structure, so they can load from there or from the filesystem.
* Extra Ruby code used by these extensions is moved into the jar under `jruby/stdlib` so they can always be loaded.

The idea here is that these built-in extensions should be loadable with just the jar, so we duplicate the stub in both `lib/ruby/stdlib` and the jar resources and move everything else into the jar. This allows all of the following libraries to be loaded from a standalone jar file:

* cgi/escape
* digest
* digest/bubblebabble
* digest/md5
* digest/rmd160
* digest/sha1
* digest/sha2
* coverage
* date
* etc
* fcntl
* io/nonblock
* io/wait
* jruby
* jruby/core_ext/string
* jruby/serialization
* monitor
* nkf
* pathname
* rbconfig
* securerandom
* set
* socket
* stringio
* strscan
* tempfile
* timeout
* zlib

Note that three libraries with built-in extensions are not included, because they have additional Ruby files sourced from elsewhere:

* bigdecimal shares files with CRuby
* ffi uses the files from the gem
* ripper shares files with CRuby

These three probably should still be addressed. However, this starts to put too much of stdlib in the jar, and makes it hard or impossible to update these libraries from gems.

Given the move in 9.4 to having more stdlib sourced from gems, this may need to be reexamined and we will need to deal with these half-built-in libraries a different way.


headius:
This approach is no longer sufficient given that many of the related libraries now live as JRuby extension gems, with their own jar files maintained in a separate repository. We have also made modifications to JRuby's build to have a "base" jar that contains only the core of JRuby, so the structure here would have to be reexamined. And then there's Java modules to consider.

The best option for a standalone jar remains the jruby-complete artifact.

Going to close this and take a step back so we can try another approach after 9.4 release.

PR #6901, opened 2021-10-21
new_sprintf WIP: support for %c

PurityLake:
Fails two specs and one mri test

Modified:
core/src/main/java/org/jruby/util/SprintfParser.java


enebo:
I cherry picked the primary commits and added a tag for a failing taint() spec (since this is destined for Ruby 3 which has no more tainting).  Thanks @PurityLake!

PR #6922, opened 2021-10-29
Remove SortedSet

headius:
Users should install the sorted_set gem (https://rubygems.org/gems/sorted_set)
to regain this functionality.

This will require some additional work to allow the `sorted_set` implementation to work on JRuby, specifically implementing a Java version of the `rbtree` gem used to implement `SortedSet` efficiently.


headius:
The rbtree gem is now maintained at https://github.com/mame/rbtree by @mame. I have started adding support for JRuby in https://github.com/mame/rbtree/pull/4 but it won't be ready for 9.4.4.

headius:
Already done for JRuby 10.1 (Ruby 4.0) in https://github.com/jruby/jruby/pull/9066.

PR #6927, opened 2021-11-01
Suport sliced with Enumerator::ArithmeticSequence on Array

edipofederle:
WIP

### DIFF
--- a/core/src/main/java/org/jruby/RubyArray.java
+++ b/core/src/main/java/org/jruby/RubyArray.java
@@ -1283,6 +1283,107 @@ public class RubyArray<T extends IRubyObject> extends RubyObject implements List
         return subseq(getRuntime().getArray(), beg, len, true);
     }
 
+    /** rb_ary_subseq_step
+     *
+     */
+    public IRubyObject subseq_step(RubyArithmeticSequence arg0) {
+        long beg, len, step, end;
+
+        step = getStep(arg0);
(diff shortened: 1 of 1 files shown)


headius:
This feature is the only thing not passing on master's 3.0 specs. Let us know if you want some help getting it finished!

edipofederle:
> This feature is the only thing not passing on master's 3.0 specs. Let us know if you want some help getting it finished!

Hi @headius,

I'm planing back to work on it today or tomorrow, I let you know if I need something here.

core/src/main/java/org/jruby/RubyArray.java
```diff
@@ -1283,6 +1283,116 @@ public IRubyObject subseq(long beg, long len) {
...
+        if (end < 0) {  end += len; }
+
+        if (!aseqExcl) {  end++; }
+
+        len = end - beg;
+
+        if (len < 0) {  len = 0;  }
+
+        long alen = realLength;
+        if (beg > alen)  return null;
```
enebo, line 1332: Some null returns here.  I think these should be nil?

enebo (changes requested):
Also I mentioned using subseq in chat.

core/src/main/java/org/jruby/RubyArray.java
edipofederle: done

enebo:
Thanks @edipofederle!  Keep these PRs coming :)

PR #6930, opened 2021-11-03
Handle beginless ArithmeticSequence

edipofederle:
Makes  things like `(..0).step(1)` (beginless) returns an instance of `Enumerator::ArithmeticSequence`.


edipofederle:
@enebo mind to checking it? Thanks

PR #6931, opened 2021-11-09
Cache literal ranges constructed from immediate values

headius:
This adds logic in IR and JIT to cache the resulting Range object when a literal range is composed of other immediate literal values, like longs.

The first pass here is a simple introduction of logic to cache the resulting object, while still emitting code to create the input values. A more compact representation would not need to revisit the input value logic (even to check a cache) and would instead embed the values directly into the range, so it can be constructed and cached outside of IR/JIT output.

### DIFF
--- a/core/src/main/java/org/jruby/ir/operands/Range.java
+++ b/core/src/main/java/org/jruby/ir/operands/Range.java
@@ -1,0 +1,82 @@
+package org.jruby.ir.operands;
+
+import org.jruby.RubyRange;
+import org.jruby.RubyRational;
+import org.jruby.ir.IRVisitor;
+import org.jruby.ir.persistence.IRReaderDecoder;
+import org.jruby.ir.persistence.IRWriterEncoder;
+import org.jruby.runtime.ThreadContext;
+import org.jruby.runtime.builtin.IRubyObject;
+
+/**
+ * Literal Range with literal endpoints.
+ */
+public class Range extends ImmutableLiteral {
+    private final ImmutableLiteral begin;
+    private final ImmutableLiteral end;
+    private final boolean exclusive;
+
+    public Range(ImmutableLiteral begin, ImmutableLiteral end, boolean exclusive) {
+        super();
+
+        this.begin = begin;
+        this.end = end;
+        this.exclusive = exclusive;
+    }
+
+    @Override
+    public OperandType getOperandType() {
+        return OperandType.RANGE;
+    }
+
+    @Override
+    public Object createCacheObject(ThreadContext context) {
+        IRubyObject begin = (IRubyObject) this.begin.cachedObject(context);
+        IRubyObject end = (IRubyObject) this.end.cachedObject(context);
+
+        if (exclusive) {
+            return RubyRange.newExclusiveRange(context, begin, end);
+        } else {
+            return RubyRange.newInclusiveRange(context, begin, end);
+        }
+    }
+
+    @Override
+    public void encode(IRWriterEncoder e) {
+        super.encode(e);
+        begin.encode(e);
+        end.encode(e);
+        e.encode(exclusive);
+    }
+
+    public static Range decode(IRReaderDecoder d) {
+        return new Range((ImmutableLiteral) d.decodeOperand(), (ImmutableLiteral) d.decodeOperand(), d.decodeBoolean());
+    }
+
+    @Override
+    public String toString() {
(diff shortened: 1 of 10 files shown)


enebo:
SHIP IT

core/src/main/java/org/jruby/ir/operands/Range.java
```diff
@@ -0,0 +1,82 @@
...
+        super();
+
+        this.begin = begin;
+        this.end = end;
+        this.exclusive = exclusive;
+    }
+
+    @Override
+    public OperandType getOperandType() {
+        return OperandType.RANGE;
```
enebo, line 29: I don't see OperandType here?

enebo: Oh it is already there...what! :)

enebo: Ah ok this was originally an Operand and we realized since it could raise we had to make it buildRange.  I wonder why we never had the discussion of both since we can check to see if it will raise during IRBuild in cases?

enebo (approved):
I am not sure how I entered review mode but I had a question and answered it so I have no issues with anything here.  Seems like a reasonable base to build on top of too.

core/src/main/java/org/jruby/ir/targets/simple/NormalValueCompiler.java
```diff
@@ -125,6 +127,16 @@ public void pushByteList(final ByteList bl) {
         });
     }
 
+    public void pushRange(Runnable begin, Runnable end, boolean exclusive) {
+        cacheValuePermanentlyLoadContext("range", RubyRange.class, null, () -> {
+            compiler.loadContext();
+            begin.run();
+            end.run();
+            compiler.adapter.pushBoolean(exclusive);
+            compiler.adapter.invokestatic(p(RubyRange.class), "newRegexp", sig(RubyRange.class, ThreadContext.class, IRubyObject.class, IRubyObject.class, boolean.class));
```
kares, line 136: the RubyRange factory method name (`newRegexp`) seems like a copy pasta typo

headius: And here I was thinking "huh, maybe we need a better way to test the non-indy logic for constant values". Will fix and add a way to test.

headius: Adding an AOT mode to the compiler's `general_spec.rb` would have found this:

```
  10) JRuby's AOTSpecUtils assigns literal values to locals
      Failure/Error: Unable to infer file and line number from backtrace
      
      Java::JavaLang::NoSuchMethodError:
        'org.jruby.RubyRange org.jruby.RubyRange.newRegexp(org.jruby.runtime.ThreadContext, org.jruby.runtime.builtin.IRubyObject, org.jruby.runtime.builtin.IRubyObject, boolean)'
      # Users.headius.projects.jruby.spec.compiler.general_spec.range0(Users/headius/projects/jruby/spec/compiler//Users/headius/projects/jruby/spec/compiler/general_spec.rb)
      # Users.headius.projects.jruby.spec.compiler.general_spec.<main>(/Users/headius/projects/jruby/spec/compiler/general_spec.rb:170)
      # jdk.internal.reflect.GeneratedMethodAccessor22.invoke(jdk/internal/reflect/GeneratedMethodAccessor22)
      # jdk.internal.reflect.DelegatingMethodAccessorImpl.invoke(jdk/internal/reflect/DelegatingMethodAccessorImpl.java:43)
      # java.lang.reflect.Method.invoke(java/lang/reflect/Method.java:566)
      # RUBY.compile_run(/Users/headius/projects/jruby/spec/compiler/general_spec.rb:129)
      # RUBY.run(/Users/headius/projects/jruby/spec/compiler/general_spec.rb:74)
      # RUBY.<main>(/Users/headius/projects/jruby/spec/compiler/general_spec.rb:170)
```

headius:
Addressed all review comments and added a test run to ensure the non-indy side of this (and all other compilation logic) gets tested. Bonus: a couple fixes found by adding AOT to the compiler spec run.

headius:
Also note: a more generic solution would advise a way to construct large chains of constant literals, caching where appropriate without emitting IR or bytecode that only runs once. With the increasing interest in immutable objects, I expect we will see more complex ways to initialize immutable arrays, hashes, and structs.

Issue #6936, opened 2021-11-12
Migrating from Travis CI to Github Actions


fidothe:
I've ported over all the jobs in `.travis.yml` to the Github Actions `ci.yml`, except for the deploy job.

It all seems mostly okay - there are some failures, but that was true on Travis. Some triaging will be needed to figure out if a failure is because of the environment change or not.

One problem caused by the move is the failure of the Sequel CI run. That's because of wider environment changes - the shipping MySQL in the GHA runners doesn't allow the old `mysql -u root` approach, and isn't specially configured to allow passwordless access from the test-runner's user account.

There's scope for significant simplification of the repetitive jobs - there's a 'reusable workflows' feature in beta - and we could probably save a couple of minutes per job by figuring out how to cache the results of `mvn package` and `bundle install`.

We could also just pull the existing `ci-windows.yml` workflow into the main `ci.yml` or further split the `ci.yml` for size/complexity reasons.

I'll open a PR for the everything-but-deploy `ci.yml` and once we're happy I can add the deploy job too.

fidothe:
Using this to track unexpected failures

fidothe:
# Failing to find `zip` in a call to `

https://github.com/jruby/jruby/blob/a482e20f632625f89666517f56c850ced9b25a95/spec/java_integration/utilities/jar_glob_spec.rb#L111

Is failing because `/usr/bin` is not in `$PATH` anymore.

My guess is that the helper here: https://github.com/jruby/jruby/blob/f07a13a5dd13a03a86d19fa6836147d8d100461d/spec/java_integration/ant_spec_helper.rb#L121-L129 is doing the removal - it's being called before the spec failure, and it doesn't restore the PATH value, so we're probably looking either at a difference with where `ant` was located on Travis or a difference in ordering on the azul JDK in GHA.

fidothe:
I've tried making the helper into a block-accepting method that resets `$PATH` in an `ensure` clause after its passed block was called, but I don't see that happening - I see the `pre-hiding` line in the log, but not the `post-hiding` line:

https://github.com/fidothe/jruby/blob/50fbff90b0a2dd4c9c9de72b05a1d5f375e3a8f0/spec/java_integration/ant_spec_helper.rb#L121-L136

fidothe:
the `rake spec:ji` failures are addressed in #6944

headius:
[...]

PR #6937, opened 2021-11-12
Migrate Travis jobs to GHA

fidothe:
Part of the work for #6936. Once we're happy I'll squash this all down into one commit for merging.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -3,41 +3,36 @@ name: JRuby CI
 on: [push, pull_request]
 
 env:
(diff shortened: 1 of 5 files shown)


edipofederle:
@fidothe maybe worth adding [this step](https://github.com/marketplace/actions/cancel-workflow-action) as well?

fidothe:
> @fidothe maybe worth adding [this step](https://github.com/marketplace/actions/cancel-workflow-action) as well?

We don't want the whole workflow to stop if one job fails - we want to know if something works on JDK 8 but not 11 (or vice versa)...

PR #6940, opened 2021-11-15
WIP: Fix symbol arity

ahorek:


enebo:
This likely will break something else but may as well address the error for now since so much else is current broken on head

PR #6944, opened 2021-11-17
Fix spec:ji on GHA

fidothe:
A spec in ji/ant/ant_spec.rb was changing ENV['PATH'] under certain
circumstances, which resulted in the shells spawned by backtick being
unable to find common programs like `zip` or `which`. Fix that, and restore
`jar_glob_spec.rb`'s use of `zip -d`.


headius:
Nice! What a silly piece of code.

PR #6945, opened 2021-11-17
Socket fixes for GHA CI

headius:
We have a few failures with IPv6 that have escaped notice due to not running in Travis CI. When they failed locally we probably ignored those failures assuming we did not have IPv6 set up properly.

This PR will include some fixes and some tags for IPv6-related failures we now see when running CI on GitHub Actions.

* Set IPv6-related constants to nil when unavailable: CRuby has complicated logic for defining socket-related constants and query methods, but I believe this is a rough match for how they filter out IPv6-related constants when IPv6 is not available. This is checked by e.g. MRI tests and rubyspec to skip or alter IPv6-related tests.
* Preserve multibyte strings in UNIX address: The path was being round-tripped through Java incorrectly, mangling multibyte characters.
* Update Socket#listen tags for IPv6: none of these specs work because JRuby cannot support both client and server behaviors in Socket. JDK Socket APIs require that we choose server or client at creation time, and it cannot be changed later. This prevents us from supporting `listen` unless you instantiate ServerSocket. All the IPv4 specs were tagged but some IPv6 specs were untagged.
* Update BasicSocket tags for IPv6: Equivalent tags already existed for the IPv4 versions but we those did not run on Travis.

### DIFF
--- a/core/src/main/java/org/jruby/ext/socket/RubySocket.java
+++ b/core/src/main/java/org/jruby/ext/socket/RubySocket.java
@@ -63,6 +63,8 @@ import org.jruby.util.io.Sockaddr;
 
 import java.io.IOException;
 import java.net.DatagramSocket;
+import java.net.Inet6Address;
+import java.net.InetAddress;
 import java.net.InetSocketAddress;
 import java.net.InterfaceAddress;
 import java.net.NetworkInterface;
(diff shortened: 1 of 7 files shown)


ahorek:
it's a good idea, thanks for the fix

PR #6949, opened 2021-11-29
Update specs

eregon:


eregon:
Many specs failing but was already the case, so I guess expected to track remaining work on Ruby 3?

headius:
Yes, expected.

PR #6951, opened 2021-11-30
Limit Date.parse input length and make interruptible

headius:
Excessively long input to the various Date parsing methods can cause the calling thread to run for a long time, perhaps tying it up indefinitely. This PR incorporates to changes similar to those in https://github.com/ruby/date/commit/3959accef8da5c128f8a8e2fd54e932a4fb253b0.

* Add a `limit` option to the affected Date parse methods (see commit above for the list).
* Make the regexp matches used by these parse methods interruptible.
* Add the same test as CRuby under our JRuby test suite, since no spec was added by ruby-core and we might not see the CRuby test for a long time (and possibly never on our 9.2 branch).

The interruptibility patch here begs a question: should we just flip the switch and make all regexp matches interruptible all the time? We opted not to do so in https://github.com/jruby/joni/pull/6, but have since reduced the overhead of the interrupt check. It seems like CRuby has made all regexp interruptible all the time, and perhaps we should follow suit. This would eliminate the need for that part of the PR.

A partial workaround for users is to enable interruptible regexp matches with the JRuby property "regexp.interruptible=true" in .jruby_opts, as `-Xregexp.interruptible` on the command line, or by setting the JVM property `jruby.regexp.interruptible=true`, and then wrap the Date parsing methods in Timeout (or perhaps your framework will do this for you).

A more complete workaround would be to apply the Ruby changes here to your local JRuby environment. This will not make the matches interruptible, but the default 128 length limit will prevent such large strings from being parsed.

cc @jordansissel, @rsim


headius:
@enebo If we intend this to be in 9.2.19.1 we can make a 9.2.19 branch and I will rebase this.

Ditto for 9.3.1.1, and I will do a separate PR merging 9.2.19.1 over.

PR #6955, opened 2021-12-01
Use method location for includes/prepends

### LINKED ISSUE #6924: java.lang.Iterable object has Enumerable ancestor but map cannot be called (jruby-9.3.x)
The anomaly described below happens on jruby-9.3.x. On jruby-9.2.19.0 everything works as expected.

I have a case where an anonymous object implementing java.lang.Iterable is returned from java, but under some circumstances, the Enumerable methods cannot be called on it although Enumerable is one of the ancestors of that object.
Example:
```
require 'neo4j_ruby_driver'
node = Neo4j::Driver::GraphDatabase.driver('bolt://localhost:7687', Neo4j::Driver::AuthTokens.basic('neo4j', 'pass')) do |driver|
  driver.session do |session|
    session.run('CREATE (t:Test{a: ["a", "b"]}) RETURN t').first.first
  end
end
node.properties # 1st call
require 'active_graph'
node.properties # 2nd call
```

There is quite a bit of prepending and including of ruby modules into java classes going on in both gems. If the `1st call` is made everything looks good. But if the `1st call` is commented out the error is:
```
NoMethodError:
  undefined method `map' for #<Java::OrgNeo4jDriverInternalValue::ListValue::1:0x26f7b114>
  Did you mean?  tap
# /Users/heinrich/mck/neo4j-ruby-driver/jruby/neo4j/driver/ext/ruby_converter.rb:12:in `as_ruby_object'
# /Users/heinrich/mck/neo4j-ruby-driver/jruby/neo4j/driver/ext/map_converter.rb:9:in `to_h'
# ./lib/active_graph/core/entity.rb:7:in `properties'
# ./spec/spec_helper.rb:16:in `<main>'
```

the caller side looks like this:

```
values(&:itself).map(&:as_ruby_object)
```

`values(&:itself)` returns an object of an anonymous class implementing `java.lang.Iterable` which has `Enumerable` as ancestor.

You can call the `each` method from Iterable and then chain `map` onto it and this works but that's just a workaround.

I am sorry for this vague description but this is a close as I can get after multiple hours of investigation. 
I believe there are 2 problems:
1. Under some conditions the Enumerable methods are erased from anonymous Iterable
2. Some type of caching problem where the methods of the anonymous Iterable depend on the context in which the first instance of the anonymous class has been created.

I'm happy to do further debugging with your guidance, but right now I'm out of ideas. I was not able to create a minimal example to reproduce the issue.

headius:
When we want to include a given module, we should be retrieving
its method location. This avoids prepends, which we will mine out
during the normal process of walking all modules included or
prepended into the source module.

Fixes #6924

### DIFF
--- a/core/src/main/java/org/jruby/RubyModule.java
+++ b/core/src/main/java/org/jruby/RubyModule.java
@@ -3605,7 +3605,8 @@ public class RubyModule extends RubyObject {
         // In the current logic, if we getService here we know that module is not an
         // IncludedModuleWrapper, so there's no need to fish out the delegate. But just
         // in case the logic should change later, let's do it anyway
-        RubyClass wrapper = new IncludedModuleWrapper(getRuntime(), insertAbove.getSuperClass(), moduleToInclude);
+        RubyModule methodLocation = moduleToInclude.getMethodLocation();
+        RubyClass wrapper = new IncludedModuleWrapper(getRuntime(), insertAbove.getSuperClass(), methodLocation);
 
         // if the insertion point is a class, update subclass lists
         if (insertAbove instanceof RubyClass) {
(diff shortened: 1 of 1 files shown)


headius:
We went with #6954 for this instead.

PR #6970, opened 2021-12-15
Support for hash argument to `Hash#transform_keys!` and `Hash#transform_keys`

edipofederle:
…nsform_keys`

### DIFF
--- a/core/src/main/java/org/jruby/RubyHash.java
+++ b/core/src/main/java/org/jruby/RubyHash.java
@@ -62,7 +62,14 @@ import org.jruby.util.RecursiveComparator;
 import org.jruby.util.TypeConverter;
 
 import java.io.IOException;
-import java.util.*;
+import java.util.AbstractCollection;
+import java.util.AbstractSet;
+import java.util.ArrayList;
+import java.util.Collection;
+import java.util.Iterator;
+import java.util.Map;
+import java.util.NoSuchElementException;
+import java.util.Set;
 import java.util.concurrent.atomic.AtomicIntegerFieldUpdater;
 
 import static org.jruby.RubyEnumerator.enumeratorizeWithSize;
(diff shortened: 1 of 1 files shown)


core/src/main/java/org/jruby/RubyHash.java
```diff
@@ -1581,28 +1575,59 @@ public IRubyObject each_key(final ThreadContext context, final Block block) {
...
-            return result;
+    @JRubyMethod(name = "transform_keys", optional =  1, rest = true)
+    public IRubyObject transform_keys(final ThreadContext context, IRubyObject[] arg0, final Block block) {
+        RubyArray transformKeys = null;
+        RubyHash result = newHash(context.runtime);
+
+        if((arg0.length > 0) && (arg0 [0] != null))
+            transformKeys = ((RubyHash) arg0[0]).keys();
+
+        if(block.isGiven())
```
enebo, line 1586: All 'if(' should be 'if ('.  Also a couple of other style rules is if you have a single line if you can omit braces but only if it is on the same line.  Anything which also has else or you put on the next line should use braces.

core/src/main/java/org/jruby/RubyHash.java
```diff
@@ -1581,28 +1575,59 @@ public IRubyObject each_key(final ThreadContext context, final Block block) {
...
+            visitAll(context, new TransformKeysVisitor(block, transformKeys), result);
+        else
+            result = this;
+
+        if(transformKeys != null) {
+            for (int i = 0; i < transformKeys.size(); i++) {
+                RubySymbol orignalKey = (RubySymbol) transformKeys.get(i);
+                IRubyObject newKeySym = (IRubyObject) ((RubyHash) arg0[0]).get(orignalKey);
```
enebo, line 1594: I reread this a couple of times.  It feels like arg0[0] here should have a meaningful name like you did with arg0[0].keys().

core/src/main/java/org/jruby/RubyHash.java
```diff
@@ -1581,28 +1575,59 @@ public IRubyObject each_key(final ThreadContext context, final Block block) {
...
+            transformKeys = ((RubyHash) arg0[0]).keys();
+
+        if(block.isGiven())
+            visitAll(context, new TransformKeysVisitor(block, transformKeys), result);
+        else
+            result = this;
+
+        if(transformKeys != null) {
+            for (int i = 0; i < transformKeys.size(); i++) {
+                RubySymbol orignalKey = (RubySymbol) transformKeys.get(i);
```
enebo, line 1593: This should be fastARef() although I half wonder if there is a better mechanism for walking all key/value pairs than using this loop.  .get() is to satisfy Java Map functionality and will potentially convert the argument to Java.

core/src/main/java/org/jruby/RubyHash.java
```diff
@@ -1615,21 +1640,48 @@ public IRubyObject transform_values(final ThreadContext context, final Block blo
         return hashCopy(context).transform_values_bang(context, block);
     }
 
-    @JRubyMethod(name = "transform_keys!")
-    public IRubyObject transform_keys_bang(final ThreadContext context, final Block block) {
+    @JRubyMethod(name = "transform_keys!", optional =  1, rest = true)
+    public IRubyObject transform_keys_bang(final ThreadContext context, IRubyObject[] arg0, final Block block) {
+        RubyArray transformKeys = null;
+
+        if((arg0.length > 0) && (arg0 [0] != null))
```
enebo, line 1647: No space between 'arg0 [0]'.  Make it arg0[0].  We also name list of arguments args by convention.

core/src/main/java/org/jruby/RubyHash.java
```diff
@@ -1615,21 +1640,48 @@ public IRubyObject transform_values(final ThreadContext context, final Block blo
...
+
+                IRubyObject value = JavaUtil.convertJavaToRuby(getRuntime(), this.get(orignalKey));
+                if (value != null) {
+                    this.fastASet(newKeySym, value);
+                    this.remove(orignalKey);
+                }
+            }
+        }
+
+        if((transformKeys != null) || block.isGiven()) { return this; }
```
enebo, line 1683: This is an example of a one-liner.  In this case the braces are not needed.

core/src/main/java/org/jruby/RubyHash.java
```diff
@@ -1581,28 +1575,59 @@ public IRubyObject each_key(final ThreadContext context, final Block block) {
...
-        if (block.isGiven()) {
-            RubyHash result = newHash(context.runtime);
-            visitAll(context, new TransformKeysVisitor(block), result);
-            return result;
+    @JRubyMethod(name = "transform_keys", optional =  1, rest = true)
+    public IRubyObject transform_keys(final ThreadContext context, IRubyObject[] arg0, final Block block) {
+        RubyArray transformKeys = null;
+        RubyHash result = newHash(context.runtime);
+
+        if((arg0.length > 0) && (arg0 [0] != null))
```
enebo, line 1583: These two conditions do not need extra parens around them.

enebo (changes requested):
The .get() is really the main thing I noticed which has to change.

The rest are just style conventions we mostly use.  Since this is not your first PR I am asking for more changes to fit into that style.

edipofederle:
Thanks, @enebo ! 

I've applied some changes and I will look at the rest tomorrow. I let you know

core/src/main/java/org/jruby/RubyHash.java
```diff
@@ -62,13 +62,7 @@
...
 
 import java.io.IOException;
-import java.util.AbstractCollection;
-import java.util.AbstractSet;
-import java.util.Collection;
-import java.util.Iterator;
-import java.util.Map;
-import java.util.NoSuchElementException;
-import java.util.Set;
+import java.util.*;
```
enebo, line 65: Oh missed this one.  we do not star up packages.  This is probably a setting in your IDE doing this I bet.

core/src/main/java/org/jruby/RubyHash.java
edipofederle: @enebo  I don't know if I did it right now, so, please, let me know anything.

PR #6978, opened 2021-12-23
Avoid NullPointerException in Symbol operand

dee-see:
Hello team,

I've been doing some reverse engineering work on a JRuby application and while dumping the IR using some code I wrote that uses [`IRDumper`](https://github.com/jruby/jruby/blob/master/core/src/main/java/org/jruby/ir/persistence/IRDumper.java) I hit a few `NullPointerException`s.

I see `getString` is already fixed in 9.3.x but the rest was missing.

I wasn't sure what was the best way to contribute this PR. Should I open `master` first and then backport to `jruby-9.3` and `jruby-9.2`?

### DIFF
--- a/core/src/main/java/org/jruby/ir/operands/Symbol.java
+++ b/core/src/main/java/org/jruby/ir/operands/Symbol.java
@@ -1,6 +1,7 @@
 package org.jruby.ir.operands;
 
 import org.jcodings.Encoding;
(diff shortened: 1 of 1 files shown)


enebo:
@dee-see This was the right branch.   It is easy to see the problem and the risk is extremely low (not fixing it can only NPE).  Thanks!

PR #6999, opened 2022-01-15
fix ArithmeticSequence#inspect

k77ch7:
This commit makes the following tests pass.
 * spec/ruby/core/enumerator/arithmetic_sequence/inspect_spec.rb
 * test_num_step_inspect and test_range_step_inspect in test/mri/ruby/test_arithmetic_sequence.rb


headius:
Thank you!

PR #7000, opened 2022-01-15
fix ArithmeticSequence#first when begin param is nil.

k77ch7:
This commit makes the following test pass.
 * test_first in test/mri/ruby/test_arithmetic_sequence.rb


headius:
Thank you!

PR #7001, opened 2022-01-15
add ArithmeticSequence#eql?

k77ch7:
This commit makes the following test pass.
 * test_eql_p in test/mri/ruby/test_arithmetic_sequence.rb


headius:
Love these simple fixes. Thanks!

PR #7004, opened 2022-01-17
Module check mishandles bash operator precedence

headius:
This fixes the issue described in #6789 where a nonexistent `release` file gets grepped even though we do an existence check. The problem here is that in bash, unlike in C or Java, the `||` and `&&` operators have the same precedence, which means they will be evaluated from left to right. In the environment described by @lfiht in  #6789, where only the `modules` file exists, this leads to an error as the `grep` command tries to read the non-existent `release` file.

### DIFF
--- a/bin/jruby.bash
+++ b/bin/jruby.bash
@@ -213,8 +213,8 @@ else
   fi
 fi
 
-# Detect modularized Java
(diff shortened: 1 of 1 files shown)


headius:
Calling @mrnoname1000 for a review...

PR #7006, opened 2022-01-17
Fixes to run tests on Apple Silicon

headius:
This PR includes several fixes to JRuby and testing logic to run CI against our M1 instance.


headius:
I will merge this and we will add more jobs in the future as they are working.

PR #7007, opened 2022-01-17
Java class reopened to add a super call should dispatch normally

### LINKED ISSUE #6968: Calling super() in a reopened Java subclass method triggers NPE in IRRuntimeHelpers.javaProxySuper
Here's one more regression between JRuby 9.2.19.0 and 9.3.2.0 involving reopened Java classes (but this time no constructors):

```ruby
describe "JRuby Java inheritance with JRuby #{JRUBY_VERSION}" do
  class Java::Test::JRubyInheritanceTestThreeSubclass
    def foo(s)
      super(s + 'baz')
    end
  end

  it 'calls method defined in Java superclass of reopened class with super()' do
    obj = Java::Test::JRubyInheritanceTestThreeSubclass.new
    expect(obj.foo('bar')).to eq('foobarbaz')
  end
end
```

```java
package test;

public class JRubyInheritanceTestThree {
    public JRubyInheritanceTestThree() {}

    public String foo(String s) {
        return "foo" + s;
    }
}
```

```java
package test;

public class JRubyInheritanceTestThreeSubclass extends JRubyInheritanceTestThree {
}
```

#### RSpec output under JRuby 9.2.19.0:

```
JRuby Java inheritance with JRuby 9.2.19.0
  calls method defined in Java superclass of reopened class with super()

Finished in 0.02122 seconds (files took 0.36372 seconds to load)
1 example, 0 failures
```

#### RSpec output under JRuby 9.3.2.0:

```
JRuby Java inheritance with JRuby 9.3.2.0
  calls method defined in Java superclass of reopened class with super() (FAILED - 1)

Failures:

  1) JRuby Java inheritance with JRuby 9.3.2.0 calls method defined in Java superclass of reopened class with super()
     Failure/Error: Unable to find org.jruby.ir.runtime.IRRuntimeHelpers.javaProxySuper(IRRuntimeHelpers.java to read failed line
     Java::JavaLang::NullPointerException:
     # org.jruby.ir.runtime.IRRuntimeHelpers.javaProxySuper(IRRuntimeHelpers.java:1244)
     # org.jruby.ir.runtime.IRRuntimeHelpers.instanceSuper(IRRuntimeHelpers.java:1204)
     # org.jruby.ir.instructions.InstanceSuperInstr.interpret(InstanceSuperInstr.java:131)
     # org.jruby.ir.interpreter.InterpreterEngine.processCall(InterpreterEngine.java:361)
     # org.jruby.ir.interpreter.StartupInterpreterEngine.interpret(StartupInterpreterEngine.java:72)
...
     # org.jruby.Main.main(Main.java:207)

Finished in 0.02579 seconds (files took 0.36036 seconds to load)
1 example, 1 failure

Failed examples:

rspec  # JRuby Java inheritance with JRuby 9.3.2.0 calls method defined in Java superclass of reopened class with super()
```

headius:
The logic here was falling into the reified Ruby < Java subclass logic when trying to super out of a reopened normal Java class. When in such a situation, we should detect that we are not in a Ruby subclass of a Java class and not try to use that logic, which depends on additional structures not present on normal proxied Java classes

Fixes #6968.

### DIFF
--- a/spec/java_integration/fixtures/SubclassOfClassWithSimpleMethod.java
+++ b/spec/java_integration/fixtures/SubclassOfClassWithSimpleMethod.java
@@ -1,0 +1,12 @@
+package java_integration.fixtures;
+
+/**
+ * Reopened Java subclasses should super normally into the parent class, not using reified subclass logic.
+ *
+ * See jruby/jruby#6968
+ */
+public class SubclassOfClassWithSimpleMethod extends ClassWithSimpleMethod {
+    public String foo(String s) {
+        return "foo" + s;
+    }
+}
--- a/core/src/main/java/org/jruby/ir/runtime/IRRuntimeHelpers.java
+++ b/core/src/main/java/org/jruby/ir/runtime/IRRuntimeHelpers.java
@@ -57,6 +57,7 @@ import org.jruby.java.proxies.JavaProxy;
 import org.jruby.javasupport.JavaMethod;
 import org.jruby.javasupport.proxy.JavaProxyClass;
 import org.jruby.javasupport.proxy.JavaProxyMethod;
+import org.jruby.javasupport.proxy.ReifiedJavaProxy;
 import org.jruby.parser.StaticScope;
 import org.jruby.runtime.Arity;
 import org.jruby.runtime.Binding;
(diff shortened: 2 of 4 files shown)


spec/java_integration/fixtures/SubclassOfClassWithSimpleMethod.java
```diff
@@ -0,0 +1,12 @@
...
+
+/**
+ * Reopened Java subclasses should super normally into the parent class, not using reified subclass logic.
+ *
+ * See jruby/jruby#6968
+ */
+public class SubclassOfClassWithSimpleMethod extends ClassWithSimpleMethod {
+    public String foo(String s) {
+        return "foo" + s;
+    }
```
ikaronen-relex, line 11: Is this actually a valid test for the issue? My original test case for #6968 does not override `foo()` in the subclass.

headius: Aha, great catch! The fix does pass with this override removed, but I copy/pasted the wrong code into the body of the subclass. The subclass is fixed in c5bb622a3d.

PR #7011, opened 2022-01-18
WIP: sprintf functionality

PurityLake:
This is a WIP PR covering the new sprintf functionality.

Fixed failing tests regarding named specifiers. The problem was with a trying to use a numbered specifier along side a keyword specifier.


PurityLake:
%f specifier is pretty much done, fails 2 specs but I believe the specs may be wrong as in regular ruby the output matches what the new sprintf outputs

PR #7012, opened 2022-01-18
Allow reified classes to store instance variables in raw fields

byteit101:
This is principally for making Java integration produce more idomatic Ruby. This allows users to configure a class to store specific ivars in a field of a reified class. This is principally for JRubyFX at this point in time, as that framwork sets java fields on a provided class. 

Still a WIP. Notable things that need to be done:

- [x] Find a good configuration name
- [x] pick exception types
- [x] deduplicate some code
- [x] Write tests
- [ ] Document weird nonsense like `@foo.equals?(@foo)` being false for some types (ex: j.l.String)
- [ ] Document weird nonsense like `@foo=` being frozen, but `self.foo=` not
- [x] Deal with clones and parents

Quick working demo:

```rb
require 'java'
require 'jruby/core_ext'

class BackedClass
   java_field "java.lang.String testfield" # define the field to be reified
   headius_definitely_approved_this_ivar_configuration_method_name :@testfield # now mark it used for ivar storage

	def testread
		puts "Ruby R #{@testfield.inspect}"
	end
	def testwrite(v)
		puts "Ruby W #{v}"
		@testfield = v
	end
	def testjavaread
		puts "Java R #{self.class.java_class.get_field("testfield").get(self).inspect}"
	end
	def testjavawrite(v)
		self.class.java_class.get_field("testfield").set(self, v)
		puts "Java W #{v}"
	end
	become_java!
end

o = BackedClass.new
o.testjavaread
o.testread
o.testjavawrite("Java says hi")
o.testjavaread
o.testread
o.testwrite("Ruby strings are better!")
o.testjavaread
o.testread

```

Now outputs:
```txt
Java R nil
Ruby R nil
Java W Java says hi
Java R "Java says hi"
Ruby R "Java says hi"
Ruby W Ruby strings are better!
Java R "Ruby strings are better!"
Ruby R "Ruby strings are better!"
```

### DIFF
--- a/core/src/main/java/org/jruby/RubyClass.java
+++ b/core/src/main/java/org/jruby/RubyClass.java
@@ -71,6 +71,7 @@ import org.jruby.javasupport.JavaConstructor;
 import org.jruby.javasupport.proxy.JavaProxyClass;
 import org.jruby.javasupport.proxy.ReifiedJavaProxy;
 import org.jruby.javasupport.util.JavaClassConfiguration;
+import org.jruby.javasupport.util.JavaClassConfiguration.DirectFieldConfiguration;
 import org.jruby.lexer.yacc.SimpleSourcePosition;
 import org.jruby.parser.StaticScope;
 import org.jruby.runtime.*;
@@ -84,6 +85,7 @@ import org.jruby.runtime.ivars.VariableTableManager;
 import org.jruby.runtime.marshal.MarshalStream;
 import org.jruby.runtime.marshal.UnmarshalStream;
 import org.jruby.runtime.opto.Invalidator;
+import org.jruby.specialized.RubyObjectSpecializer;
 import org.jruby.util.ArraySupport;
 import org.jruby.util.ClassDefiningClassLoader;
 import org.jruby.util.CodegenUtils;
@@ -1378,12 +1380,16 @@ public class RubyClass extends RubyModule {
 
                 this.setInstanceVariable("@java_class", Java.wrapJavaObject(runtime, result));
                 JavaProxy.setJavaClass(this, result);
+                reifiedClassJava = Boolean.TRUE;
             } else {
                 setRubyClassAllocator(result);
+                reifiedClassJava = Boolean.FALSE;
             }
             reifiedClass = result;
             nearEnd = true;
             JavaProxyClass.ensureStaticIntConsumed();
+            if (javaClassConfiguration.requestedStorageVariables != null)
+                javaClassConfiguration.requestedStorageVariables.forEach(variableTableManager::requestFieldStorage);
             return; // success
         }
         catch (LinkageError error) { // fall through to failure path
@@ -2232,6 +2238,14 @@ public class RubyClass extends RubyModule {
             throw runtime.newTypeError("Attempted to get a Java class for a Ruby class");
         else
             return (Class<? extends ReifiedJavaProxy>) reifiedClass;
+    }
+
+    /**
+     * If the current class is reified (null otherwise) and if it extends java
+     * @return Null if not reified, true if a concrete extension class, and false if a ruby class
+     */
+    public Boolean getIsReifiedExtendedJavaClass() {
+        return reifiedClassJava;
     }
 
     public static Class<?> nearestReifiedClass(final RubyClass klass) {
--- a/core/src/main/java/org/jruby/javasupport/util/JavaClassConfiguration.java
+++ b/core/src/main/java/org/jruby/javasupport/util/JavaClassConfiguration.java
@@ -50,6 +50,7 @@ public class JavaClassConfiguration implements Cloneable {
     public Map<Class<?>, Map<String, Object>> classAnnotations;
     public Map<String, List<Class<?>[]>> methodSignatures;
     public Map<String, Class<?>> fieldSignatures;
+    public List<DirectFieldConfiguration> requestedStorageVariables; // Variable to map to fields if we are ever reified
 
     public boolean callInitialize = true;
     public boolean allMethods = true;
@@ -93,6 +94,7 @@ public class JavaClassConfiguration implements Cloneable {
         if (classAnnotations != null) other.classAnnotations = new HashMap<>(classAnnotations); // TOOD: deep clone
         if (methodSignatures != null) other.methodSignatures = new HashMap<>(methodSignatures); // TOOD: deep clone
         if (fieldSignatures != null) other.fieldSignatures = new HashMap<>(fieldSignatures); // TOOD: deep clone
+        if (requestedStorageVariables != null) other.requestedStorageVariables = new ArrayList<>(requestedStorageVariables);
 
         return other;
     }
@@ -123,5 +125,19 @@ public class JavaClassConfiguration implements Cloneable {
 
         included.add(name);
         excluded.remove(name);
+    }
+    
+    public static class DirectFieldConfiguration {
+        public String name;
+        public Class<?> fieldType;
+        public Boolean unwrap;
+        public Class<?> unwrapType;
+        public DirectFieldConfiguration(String name, Class<?> fieldType, Boolean unwrap, Class<?> unwrapType) {
+            this.name = name;
+            this.fieldType = fieldType;
+            this.unwrap = unwrap;
+            this.unwrapType = unwrapType;
+        }
+        
     }
 }
--- a/core/src/main/java/org/jruby/runtime/ivars/FieldVariableAccessor.java
+++ b/core/src/main/java/org/jruby/runtime/ivars/FieldVariableAccessor.java
@@ -58,11 +58,12 @@ public class FieldVariableAccessor extends VariableAccessor {
         super(realClass, name, index, classId);
 
         this.getter = getter;
+        this.setter = wrapSetter(setter);
+    }
 
+    protected MethodHandle wrapSetter(MethodHandle setter) {
         // mix frozen check into setter
-        setter = MethodHandles.foldArguments(setter, ENSURE_SETTABLE.asType(setter.type()));
-        
-        this.setter = setter;
+        return MethodHandles.foldArguments(setter, ENSURE_SETTABLE.asType(setter.type()));
     }
 
     public MethodHandle getGetter() {
--- a/core/src/main/java/org/jruby/runtime/ivars/RawFieldVariableAccessor.java
+++ b/core/src/main/java/org/jruby/runtime/ivars/RawFieldVariableAccessor.java
@@ -1,0 +1,198 @@
+/*
+ ***** BEGIN LICENSE BLOCK *****
+ * Version: EPL 2.0/GPL 2.0/LGPL 2.1
+ *
+ * The contents of this file are subject to the Eclipse Public
+ * License Version 2.0 (the "License"); you may not use this file
+ * except in compliance with the License. You may obtain a copy of
+ * the License at http://www.eclipse.org/legal/epl-v20.html
+ *
+ * Software distributed under the License is distributed on an "AS
+ * IS" basis, WITHOUT WARRANTY OF ANY KIND, either express or
+ * implied. See the License for the specific language governing
+ * rights and limitations under the License.
+ * 
+ * Alternatively, the contents of this file may be used under the terms of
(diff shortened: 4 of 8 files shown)


headius:
A couple general comments...

* Perhaps we should merge the new configuration method into `java_field` as an option or additional argument? Something like `java_field, "java.lang.String => testfield", variable: :@testfield` or perhaps `variable: true` to just use the same name for the instance variable (with @ of course).
* This also lets us punt on coming up with a good name 😀
* The lambdafication of the "variable accessor builder" is a good step. We may want to just formalize an interface, though, because there's no reason we'd want to mix and match, e.g. a different type of accessor for set and get.

Nicely done reusing the plumbing for FieldVariableAccessor! I can give this a run through JVM assembly dumping to confirm it is inlining the access to the field directly.

enebo:
Some thoughts from chat:

java_field, "java.lang.String => testfield", variable: :@testfield
I think :@testfield could possibly be variable: true if it is same name and variable: :test_field or whatnot if not
truthy eliminates needing to write the name twice
removing the @, to me at least, doesn't make it look like you can do other weird stuff like :$testfield

byteit101:
My current configuration thought is `java_field "java.lang.String foo", variable: true` for "basic" usage, and `java_field "java.lang.Object foo", variable: {unwrap: true, type: com.MyClass}` for "advanced" usage though I'm also debating `bind_variable` and `bind` instead of `variable`

kares:
> `variable: {unwrap: true, type: com.MyClass}` 

does it make sense to `unwrap: false, type: com.MyClass` ? if not than there should be just one option ...

headius:
> `java_field "java.lang.Object foo", variable: {unwrap: true, type: com.MyClass}`

A similar question to @kares: why would you declare the variable as `java.lang.Object` but then also as `com.MyClass`? The latter is just for unwrapping, but isn't it going to determine that type anyway when it unwraps?

byteit101:
> but isn't it going to determine that type anyway when it unwraps?
That's the argument to pass into toJava. The example above is kinda dumb, but short vs int vs long may be more useful.

I've updated it to be:
* Default: `java_field "java.lang.Object foo"` # bind_variable: false
* Auto Unwrap: `java_field "java.lang.Object foo", bind_variable: true`
* No Unwrap: `java_field "java.lang.Object foo", bind_variable: {to_java: false}`
* Unwrap with type: `java_field "java.lang.Object foo", bind_variable: {to_java: com.MyClass}`

I've also added some more tests for concrete vs ruby and unwrap vs no unwrap

As part of the concrete testing I had to expand the methodhandle stuff much more, but I think I'm ready for review now

headius:
A few general comments/questions as I start reviewing...

> Default: java_field "java.lang.Object foo" # bind_variable: false

So this is like current behavior on a Java proxy type, where the field will be presented as a `foo` method that converts Ruby/Java both ways, yes (specifically things like String and numerics)?

> Auto Unwrap: java_field "java.lang.Object foo", bind_variable: true

Something about this naming doesn't work for me. If we want to be explicit about it being an instance variable, why not "instance_variable" instead of "bind_variable"? This keyword is really about adjusting the behavior of the instance variable accesses, right?

If this is a more general purpose `bind` configuration then `bind: true` or `binding: true` or even `variable_bind: true` read better to me. Bikeshedding mostly.

byteit101:
> So this is like current behavior on a Java proxy type, where the field will be presented as a foo method that converts Ruby/Java both ways, yes (specifically things like String and numerics)?

Correct, and those methods are always present (the configuration doesn't change that)

core/src/main/java/org/jruby/RubyClass.java
```diff
@@ -307,6 +309,49 @@ public int getVariableTableSizeWithExtras() {
...
+            if (javaClassConfiguration.requestedStorageVariables == null)
+                javaClassConfiguration.requestedStorageVariables = new ArrayList<>();
+            javaClassConfiguration.requestedStorageVariables.add(config);
+        }
+    }
+
+    /**
+     * Actually requests field storage for the ivar once we are reified
+     */
+    private void requestFieldStorage(AdvancedFieldConfiguration config) {
```
headius, line 330: RubyClass seems to be accumulating a lot of clutter related to reification that isn't strictly part of the RubyClass behavior. Almost feels like this stuff should live in the VariableTableManager rather than here. Might at least be worth making the method above private  (if possible) until we decide how to reorg this logic.

core/src/main/java/org/jruby/RubyClass.java
```diff
@@ -307,6 +309,49 @@ public int getVariableTableSizeWithExtras() {
...
+            Class<?> unwrapType = config.unwrapType == null ? fieldType : config.unwrapType;
+            if (unwrap && !fieldType.isAssignableFrom(unwrapType))
+                throw runtime.newTypeError("java_field " + config.name + " has incorrectly specified unwrap type for @ivar mapping");
+            
+            variableTableManager.getVariableAccessorForJavaMappedVar("@" + config.name,
+                    unwrap,
+                    fieldType,
+                    unwrapType,
+                    RubyObjectSpecializer.LOOKUP.findGetter(getReifiedClass(), config.name, fieldType),
+                    RubyObjectSpecializer.LOOKUP.findSetter(getReifiedClass(), config.name, fieldType));
```
headius, line 347: There are many places where getReifiedClass is called repeatedly... these should use a variable and call it once.

core/src/main/java/org/jruby/javasupport/util/JavaClassConfiguration.java
```diff
@@ -124,4 +126,18 @@ public synchronized void include(String name) {
         included.add(name);
         excluded.remove(name);
     }
+    
+    public static class AdvancedFieldConfiguration {
```
headius, line 130: I don't like the vagueness of "Advanced" since we might have something more "advanced" in the future. Perhaps "DirectFieldConfiguration" would be more descriptive?

headius (approved):
Approved but we should chat through my very minor naming/encapsulation concerns.

core/src/main/java/org/jruby/RubyClass.java
byteit101: I've moved this to VariableTableManager, though I had to make it public so RubyClass could access it. The other method needs to be public as it's configured in class.rb

headius: Ok... we can go with that and think about refactoring in the future. Nobody should call this stuff outside of JRuby anyway.

byteit101:
> > Auto Unwrap: java_field "java.lang.Object foo", bind_variable: true
> 
> Something about this naming doesn't work for me. If we want to be explicit about it being an instance variable, why not "instance_variable" instead of "bind_variable"? This keyword is really about adjusting the behavior of the instance variable accesses, right?
> 
> If this is a more general purpose bind configuration then bind: true or binding: true or even variable_bind: true read better to me. Bikeshedding mostly.

I think we should get the naming right on the first go, so here is my reasoning:

- Should have something to indicate this is to do with Ruby instance variables
- Should indicate that this is for more advanced java integration of fields with ruby class syntax
  - We are constrained lexically by the positioning of this argument on a thing about java fields, whereas we want to tell the user what will happen to variables.
- Should indicate that this may change the semantics of fields (see the notes on the list above)
- Ergonomics should be reasonably short, though not too short. If the ergonomics are at an optimized level of clunky, they are enough to remember the basics, but not enough to remember the details, sending people to avoid unless more necessary, and sending people to the docs where we can warn them of the semantic issues above.
It's that last point that made me favor `bind`, but `link`, `map`, or any other similar modifier works for me. I like `instance_variable` but I think it may be too unwieldy in the name department when combined with a modifier, but I'm not sure.

A few I thought of:

- `expose_as_variable`
- `link_to_variable`
- `variable_link`
- `integrate_variable`
- `as_variable`
- `instance_variable_bind`
- `variable_bind` # I like this suggestion of yours better than my `bind_variable`

byteit101:
I do like the parity with `instance_variable_get` and `instance_variable_set`, but I don't think it necessarily reads as "and make it accessible" to me without one of the action words I proposed above (bind, link, expose, etc).

Because this is rather advanced usage and does weird things to semantics, I'd definitely like something a) easily understandable without googling in case a someone runs into it and hasn't seen this form before, and b) easily googleable. Due to the infrequency of use by end users expected, I think we should optimize understanding for the novice reader over the expert writer here. That's one reason  (in addition to those listed above) why I have preferred "verb-y" names thus far. 

Taking this into account, I definitely would prefer something like `instance_variable_bind: true` or `expose_via_instance_variable: true`

I notice you are preferring verbless names so far, what's the reasoning behind that preference?

As for the other options, `convert` is fine by me, as is inline (non-nested hash). Another thought I had was `store_as`, but I think `convert` is probably fine. My only concern is that now they may read as an independent option.

enebo:
I think adding a verb will become harder to remember than the word 'instance_variable'.    java_field is a declarative set of statements about the field.  One adjective here is it is also an instance_variable (or so it will appear so).

I cannot think of a better name that convert.  I would like to see a concrete example of why convert needs to be anything other than true or false.  I believe there is one but it would be nice to see a good example (all I am coming up with is field is a base type yet we decide to say it is a specific type with convert).  We need this example if for no other reason then we need to document the feature.

byteit101:
>  I would like to see a concrete example of why convert needs to be anything other than true or false. I believe there is one but it would be nice to see a good example (all I am coming up with is field is a base type yet we decide to say it is a specific type with convert). We need this example if for no other reason then we need to document the feature.

The need for a specific type comes from toJava requiring an argument, and me not knowing any better way to say "make this a java thing, whatever it is". Under the hood, convert is only used as self.my_field = value.to_java(convert_argument_here)

Examples:
```rb
java_field "java.lang.Object myfield", ..., convert: false # default for object, stores the IRubyObject in myfield
java_field "my.Class myfield", ..., convert: my.Class # default for non-Object, stores the raw java instance in myfield

java_field "java.lang.Object myfield", ..., convert: my.Class # stores the raw java instance in myfield
java_field "java.lang.Numeric myfield", ..., convert: j.l.Float # store a float in the number, not an integer, long, or short
```

Without convert, any type of Object either a) won't be able to store a java instance or b) won't be able to store a ruby instance. by adding convert, it's configurable.
As a more concrete example, imagine a poorly written java JSON library, that reads fields of type object, but expects only one of String or HashMap. To interface with that, you want to store a string, and thus have to use java_field "java.lang.Object myfield", ..., convert: j.l.String`

Another important note: the `convert` argument only really matters if java (or any jvm bits) will be reading this field after Ruby had written to it. It doesn't matter for writing to it from the JVM. JavaFX (via JRubyFX) only writes, never reads fields, so convert is unnecessary there. 

A summary of when `convert:` is useful:

- When Java is reading the field after Ruby writes, and
- The field type is `j.l.Object`, or
- The field type is the parent of a class hierarchy of differing-semantic classes (`j.l.Numeric`), or
- if the JRuby to_java itself has special logic (unsure)

I think that is all the reasons you would use `convert:`

As for the primary name, I'm not clear on which name you prefer? `instance_variable` plain? one of the verbed forms?

enebo:
I guess I see this as three possible forms:

No conversion (you are just passing around the natural Java type of the field:

```ruby
java_field "my.Address address", instance_variable: true, convert: false
```

Conversion (we are passing a Ruby type around which knows how to convert to the Java type [rubyToJava]):

```ruby
java_field "java.lang.String address", instance_variable: true, convert: true
```

Conversion but we want to specify what type we want to convert too (this seems like it could only be useful for a Ruby type which can be multiple Java classes/interfaces):

```ruby
java_field "java.swing.JComponent component", instance_variable: true, convert: java.swing.JButton
```

This last example is what I don't understand.  It is also possible I am missing something in the original proposal.  I largely just see a specified type to convert but I do not really understand why convert would ever be different than the type declaration.  If they are different then I am wondering what the behavior is if the Java access of this same type puts the base type but it will not match our convert type.

byteit101:
I have updated the syntax to "intance_variable:" and "to_java", as per the recent matrix chat. I have also squashed and rebased this to 9.3 as per the same matrix chat. That's all I have.

headius:
Merged!

PR #7014, opened 2022-01-18
Implement Thread.each_caller_location

headius:
This passes all tests in https://github.com/ruby/ruby/pull/5445 except for the enumerator test, which shows it is tricky to get the "top" of the stack to match up.

See https://github.com/ruby/ruby/pull/5445

See https://bugs.ruby-lang.org/issues/16663


headius:
Done separately for JRuby 10.0.0.0 in https://github.com/jruby/jruby/commit/e8996c6fef21b0d40b8918e95109d3c55542b986.

PR #7022, opened 2022-01-21
Bash cleanup

mrnoname1000:
The bash launcher was in dire need of a cleanup so I decided to go through and rework anything that could be easily improved.

PR #7024, opened 2022-01-21
Port bash launcher to sh

mrnoname1000:
The hardest thing to port is arrays, so I had the idea to store un-expanded variables in a string, then expand them with eval later. It's a little fragile to maintain but it works. Some documentation would probably be necessary to avoid unsafe evals.

I'm also not sure what to call the new script. I changed the shebang to `/bin/sh` for testing purposes but the filename still ends in .bash, and users may rely on that file being there.

Depends on #7022.


headius:
Note we do have a `jruby.sh` in `bin` but it is very old and doesn't include any launcher features added in the last decade or so. The new sh launcher would be named `jruby.sh` once we merge and that would then be the only script-based launcher in `bin` (copied to `bin/jruby` for release like the bash launcher).

headius:
Thank you! This has been merged to master. I will do some cleanup to move this to `jruby.sh`, with `jruby.bash` just invoking it, and modify the build to copy from `jruby.sh` to bin/jruby.

We look forward to further improvements!

Issue #7015, opened 2022-01-18
Ruby 3.1 support


headius:
See https://github.com/ruby/ruby/blob/v3_1_0/NEWS.md for the full 3.1 NEWS file.

This listing includes all features from the Ruby 3.1 news file that might be relevant to JRuby. I have removed sections specific to the Ruby C API or the static analysis and type-profiling libraries that are not currently supported on JRuby.

See also #6878 for Ruby 3.0 features and #6464 for Ruby 2.7, which will also be included. 

## Language changes

* [x]   The block argument can now be anonymous if the block will
    only be passed to another method. [[Feature #11256]] (#7044)

    ```ruby
    def foo(&)
      bar(&)
    end
    ```

* [x]   Pin operator now takes an expression. [[Feature #17411]]

    ```ruby
    Prime.each_cons(2).lazy.find_all{_1 in [n, ^(n + 2)]}.take(3).to_a
    #=> [[3, 5], [5, 7], [11, 13]]
    ```

* [x]   Pin operator now supports instance, class, and global variables.
    [[Feature #17724]]

    ```ruby
    @n = 5
    Prime.each_cons(2).lazy.find{_1 in [n, ^@n]}
    #=> [3, 5]
    ```

* [x]   One-line pattern matching is no longer experimental.

* [x]   Parentheses can be omitted in one-line pattern matching.
    [[Feature #16182]]

    ```ruby
    [0, 1] => _, x
    {y: 2} => y:
    x #=> 1
    y #=> 2
    ```

* [ ] #7440 Multiple assignment evaluation order has been made consistent with single assignment evaluation order.  With single assignment, Ruby uses a left-to-right evaluation order.  With this code:

    ```ruby
    foo[0] = bar
    ```

    The following evaluation order is used:

    1. `foo`
    2. `bar`
    3. `[]=` called on the result of `foo`

    In Ruby before 3.1.0, multiple assignment did not follow this
    evaluation order.  With this code:

    ```ruby
    foo[0], bar.baz = a, b
    ```

    Versions of Ruby before 3.1.0 would evaluate in the following
    order

    1. `a`
    2. `b`
    3. `foo`
    4. `[]=` called on the result of `foo`
    5. `bar`
    6. `baz=` called on the result of `bar`

    Starting in Ruby 3.1.0, the evaluation order is now consistent with
    single assignment, with the left-hand side being evaluated before
    the right-hand side:

    1. `foo`
    2. `bar`
    3. `a`
    4. `b`
[...]

PR #7027, opened 2022-01-22
Ruby 3.1: Add Array#intersect?

k77ch7:
For #7015

The current "test/mri/ruby/test_array.rb"  does not have "intersect?" test,  I get it from mri and run it localy.


enebo:
Keep 'em coming @k77ch7!

PR #7037, opened 2022-01-26
Fix Array#minmax with block

k77ch7:
This commit makes the following tests pass.
 * test_minmax in test/mri/ruby/test_array.rb
 * spec/ruby/core/array/minmax_spec.rb

### DIFF
--- a/core/src/main/java/org/jruby/RubyArray.java
+++ b/core/src/main/java/org/jruby/RubyArray.java
@@ -5543,7 +5543,7 @@ float_loop:
 
     @JRubyMethod
     public IRubyObject minmax(ThreadContext context, Block block) {
-        if (block.isGiven()) return Helpers.invokeSuper(context, this, block);
+        if (block.isGiven()) return RubyEnumerable.minmax(context, this, block);
 
         return RubyArray.newArray(context.runtime, callMethod("min"), callMethod("max"));
     }


enebo:
@k77ch7 This is interesting.  We had some methods where we have to call super.  I am not sure if 3.1 (or 3.0) decided to bypass calling super for minmax or whether this was wrong to begin with. @ahorek landed this when we were still targetting 3.0.

I am only commenting because this is essentially new code but fixed tests are fixed tests. :)

k77ch7:
@enebo 

I confirmed the your comment and Ruby 3.1 source. Ruby 3.1 uses calling super for minmax.
@ahorek 's commit is right, but some Array#minmax tests are failing on current master.

I would like to update my PR by this patch. Can I get your comment?
```diff
diff --git a/core/src/main/java/org/jruby/RubyArray.java b/core/src/main/java/org/jruby/RubyArray.java
index feb52aacbf..7e428d44e8 100644
--- a/core/src/main/java/org/jruby/RubyArray.java
+++ b/core/src/main/java/org/jruby/RubyArray.java
@@ -5543,7 +5543,7 @@ float_loop:
 
     @JRubyMethod
     public IRubyObject minmax(ThreadContext context, Block block) {
-        if (block.isGiven()) return Helpers.invokeSuper(context, this, block);
+        if (block.isGiven()) return Helpers.invokeSuper(context, this, context.runtime.getArray(), "minmax", NULL_ARRAY, block);
 
         return RubyArray.newArray(context.runtime, callMethod("min"), callMethod("max"));
     }
```

ahorek:
@k77ch7 nope, super override shouldn't be bypassed, your second patch is correct, thanks!

btw there's a bug with minmax, but it's not directly related to this change
```
[1,2,3].minmax {|a, b| 1 }
is [1, 3] but it should be [2, 3]
```

enebo:
@k77ch7 yeah submit a new PR with that change and I will apply it.  Thanks for figuring this out.

@ahorek yay!

PR #7043, opened 2022-01-27
Update Array#minmax with block

k77ch7:
See  #7037 . This reverts Array#minmax to use calling supser.


enebo:
@k77ch7 Thanks for the extra leg work on this!

Issue #6878, opened 2021-10-05
Ruby 3.0 Support


headius:
# Ruby 2.7 Note
This list, as mentioned below, is changed or added features since the Ruby **2.7.0** release. However, since JRuby skipped Ruby 2.7 (JRuby 9.3 was 2.6 compatible and 9.4 is 3.0 compatible), we will also need to audit that list of features and ensure that any changes exclusive to 2.7 are also implemented.

See https://github.com/jruby/jruby/issues/6464 for the checklist of Ruby 2.7 features.

# Ruby specs for 2.7 and 3.0

Many specs have already been added for Ruby 2.7 and Ruby 3.0 behaviors, but there are always gaps! Please help write a complete set of executable Ruby specifications for these releases. Your efforts will help Rubyists worldwide have a better understanding of Ruby and better tools to use it.

Ruby 2.7: https://github.com/ruby/spec/issues/745

Ruby 3.0: https://github.com/ruby/spec/issues/823

# Full list of changes for Ruby 3.0.0
NOTE: https://rubyreferences.github.io/rubychanges/3.0.html gives more details for many features and changes. From https://github.com/ruby/ruby/blob/master/doc/NEWS-3.0.0.md and [ruby/spec#823](https://github.com/ruby/spec/issues/823):

This document is a list of user visible feature changes since the **2.7.0** release, except for bug fixes.

Note that each entry is kept to a minimum, see links for details.

## Language changes
* [x]  Keyword arguments are now separated from positional arguments.
  Code that resulted in deprecation warnings in Ruby 2.7 will now
  result in ArgumentError or different behavior. [[Feature #14183](https://bugs.ruby-lang.org/issues/14183)]
* [x]  Procs accepting a single rest argument and keywords are no longer
  subject to autosplatting.  This now matches the behavior of Procs
  accepting a single rest argument and no keywords.
  [[Feature #16166](https://bugs.ruby-lang.org/issues/16166)]
  ```ruby
  pr = proc{|*a, **kw| [a, kw]}
  
  pr.call([1])
  # 2.7 => [[1], {}]
  # 3.0 => [[[1]], {}]
  
  pr.call([1, {a: 1}])
  # 2.7 => [[1], {:a=>1}] # and deprecation warning
  # 3.0 => [[[1, {:a=>1}]], {}]
  ```
* [x]  Arguments forwarding (`...`) now supports leading arguments.
  [[Feature #16378](https://bugs.ruby-lang.org/issues/16378)]
[...]

PR #7051, opened 2022-01-29
Fix Enumerable#grep and Enumerable#grep_v with Regexp

k77ch7:
Enumerable#grep(_v) with block handles non-string elements correctly.
The following tests pass.
 * spec/ruby/core/enumerable/grep_spec
 * spec/ruby/core/enumerable/grep_v_spec.rb

#6878


enebo:
You are on fire!  Thanks for the PRs.

PR #7054, opened 2022-01-31
Move everything to jruby.sh

headius:
* bin/jruby.bash sh script moved to bin/jruby.sh
* build copies bin/jruby.sh to bin/jruby if it is missing
* bin/jruby.bash warns of deprecation and launches bin/jruby.sh
* doco changes throughout

See https://github.com/jruby/jruby/pull/7024


headius:
Thumbs up from @mrnoname1000 so merging now.

PR #7055, opened 2022-01-31
Explicitly use public visibility for define_singleton_method

headius:
This PR makes public visibility the only visiblity for `define_singleton_method` methods, and adds a spec testing a simple case of this.

This is a bit of an open question, since in CRuby `define_singleton_method` ends up producing public methods nearly all of the time due to a quirk in how the frame's visibility is accessed.

There's at least one edge case where `define_singleton_method` **does** obey the frame visibility, but it requires calling `define_singleton_method` from within a scope rooted on the destination singleton class. I would argue this is a peculiar-enough edge case that it should be made to explicitly use public visibility in all cases.

See https://github.com/ruby/ostruct/issues/40 for a related issue, where we warn about aliasing `define_singleton_method` due to its rarely-used ability to source visibility from the caller's scope.

See https://bugs.ruby-lang.org/issues/18561 for my CRuby issue to codify this behavior and eliminate the edge cases.

### DIFF
--- a/core/src/main/java/org/jruby/RubyKernel.java
+++ b/core/src/main/java/org/jruby/RubyKernel.java
@@ -1453,7 +1453,7 @@ public class RubyKernel {
         return context.nil;
     }
 
-    @JRubyMethod(required = 1, optional = 1, reads = VISIBILITY)
+    @JRubyMethod(required = 1, optional = 1)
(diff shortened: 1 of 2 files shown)


headius:
This change appears to be getting some traction with the CRuby folks (https://github.com/ruby/ruby/pull/5636) so we will optimistically proceed with this change in 9.3.4.

PR #7056, opened 2022-01-31
Restore 9.2 collect signature and add more API

headius:
This addresses an issue found by jruby-openssl where we altered
the return value for a method in JRuby and broke backward compat
with JRuby 9.2 and earlier. Specifically we unintentionally caused
the RubyArray#collect(context, block) method to return IRubyObject
instead of RubyArray due to needing to return an enumerator for
some inputs.

As a result of this change, jruby-openssl compiled against the new
signature and became incompatible with JRuby 9.2.

This fix does not address the issue for JRuby 9.3.0 through 9.3.3,
which continue to have the bad signature. The restored signature
will remain through 9.3 and 9.4 maintenance at least, with two new
methods collectArray and collectEnum added to allow users to opt
into those specific paths. The Ruby-facing method is renamed to
rbCollect to indicate it is part of the Ruby interface for this
class.

See https://github.com/jruby/jruby-openssl/issues/245

### DIFF
--- a/core/src/main/java/org/jruby/RubyArray.java
+++ b/core/src/main/java/org/jruby/RubyArray.java
@@ -52,7 +52,6 @@ import java.util.List;
 import java.util.ListIterator;
 import java.util.RandomAccess;
 
-import org.jcodings.Encoding;
 import org.jcodings.specific.USASCIIEncoding;
 import org.jruby.anno.JRubyClass;
 import org.jruby.anno.JRubyMethod;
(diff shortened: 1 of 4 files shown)


headius:
For 9.3.0 through 9.3.3 the only feasible migration path would be to use `collect19`, which has the same signature in all 9.2 and 9.3 releases but supports the new behavior in 9.3. It is deprecated, so this is not a great answer.

Note also that `collectCommon` is removed as a method in this PR, after being added as a public method by mistake (the old public `collect` was simply renamed without changing visibility). This method was not intended to be exposed as a public RubyArray API.

PR #7057, opened 2022-02-01
Launcher fixes

mrnoname1000:
Some tests were failing because of an oversight in the argument parsing, and a couple other things caught my eye.


headius:
Looks like the shell issues are fixed in several of these builds, so I'm merging to investigate the other failures from my stdlib update.

Issue #7009, opened 2022-01-17
Break in block passed to Thread doesn't cause LocalJumpError


toy:
Following code doesn't cause a `LocalJumpError` in JRuby unlike in MRI:

```ruby
proc do |&block|
  Thread.new{ block.call }.join
end.call{ break }
```

Using docker to compare behaviour:

```shell
cat << 'RUBY' > test.rb
puts '#' * 60, "Using #{RUBY_DESCRIPTION}"
$stdout.flush

proc do |&block|
  Thread.new{ block.call }.join
end.call{ break }

puts "Still alive!"
RUBY

cat test.rb | docker run -i --rm ruby:2.7 ruby
cat test.rb | docker run -i --rm ruby:3.0 ruby
cat test.rb | docker run -i --rm ruby:3.1 ruby
cat test.rb | docker run -i --rm jruby:9.2 jruby
cat test.rb | docker run -i --rm jruby:9.3 jruby
```

Gives following output:

```
############################################################
Using ruby 2.7.2p137 (2020-10-01 revision 5445e04352) [x86_64-linux]
#<Thread:0x000055eb7e475698 -:5 run> terminated with exception (report_on_exception is true):
-:6:in `block in <main>': break from proc-closure (LocalJumpError)
	from -:5:in `block (2 levels) in <main>'
-:6:in `block in <main>': break from proc-closure (LocalJumpError)
	from -:5:in `block (2 levels) in <main>'
############################################################
Using ruby 3.0.3p157 (2021-11-24 revision 3fb7d2cadc) [x86_64-linux]
#<Thread:0x00005632073b3668 -:5 run> terminated with exception (report_on_exception is true):
-:6:in `block in <main>': break from proc-closure (LocalJumpError)
	from -:5:in `block (2 levels) in <main>'
-:6:in `block in <main>': break from proc-closure (LocalJumpError)
	from -:5:in `block (2 levels) in <main>'
############################################################
Using ruby 3.1.0p0 (2021-12-25 revision fb4df44d16) [x86_64-linux]
#<Thread:0x00007f5f06b234e8 -:5 run> terminated with exception (report_on_exception is true):
-:6:in `block in <main>': break from proc-closure (LocalJumpError)
	from -:5:in `block (2 levels) in <main>'
-:6:in `block in <main>': break from proc-closure (LocalJumpError)
	from -:5:in `block (2 levels) in <main>'
############################################################
[...]

PR #7061, opened 2022-02-01
Changes to how non-local break is propagated

headius:
This is an experiment to change how we propagate non-local breaks (i.e. within a closure or eval) to address #7009 and allow detecting that the target frame is not part of the current thread (and therefore should be a LocalJumpError right away).

The strategy is as follows:

* Switch non-local break to target a Frame, rather than a DynamicScope.
* Add a thread ID field to Frame that will be set at create time and used to identify the source thread.
* Compare the thread ID field at break-throw time to know whether the break has a valid target on the current thread.

### DIFF
--- a/core/src/main/java/org/jruby/RubyMethod.java
+++ b/core/src/main/java/org/jruby/RubyMethod.java
@@ -224,7 +224,7 @@ public class RubyMethod extends AbstractRubyMethod {
         int line = getLine(); // getLine adds 1 to 1-index but we need to reset to 0-index internally
         body = new MethodBlockBody(runtime.getStaticScopeFactory().getDummyScope(), signature, entry, argsDesc,
                 receiver, originModule, originName, getFilename(), line == -1 ? -1 : line - 1);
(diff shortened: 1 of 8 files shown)


headius:
This naive attempt does not work right and we will need to explore this further as in #7103. I believe this change will be too aggressive to do in 9.3 so retargeting for 9.4.

headius:
@toy Can you tell  us how critical this functionality is to your library? The changes needed to support this efficiently are too drastic to do in a 9.3.x release, and the stack search that CRuby does would add overhead to all non-local breaks (i.e. break in a block). If we can wait to fix this in 9.4 it would be the cleanest option.

toy:
I would not call it critical - just `break` is not supported when using `in_threads` with jruby, and tests are behaving properly (detecting if break doesn't work and skipping).

headius:
There's other work in progress to improve propagation of non-local break, such as #6643. Close this very old PR in favor of that PR and future work to simplify non-local flow control.

PR #7062, opened 2022-02-02
Update Enumerator#+ to fix #7058

### LINKED ISSUE #7058: OOM in TestEnumerator#test_lazy_chain (3.1)
Seems like something is going off the rails in the `test_lazy_chain` test in test/mri/ruby/test_enumerator.rb. Locally it hangs for longer than I care to wait, and on GHA it eventually OOMs with GC limit exceeded or heap space exceeded.

I will exclude but we should investigate before 9.4 release.

k77ch7:
Enumerator#+ returns Enumerator::Lazy instance when either of operands is Enumerator::Lazy instance.


headius:
Nice!

PR #7065, opened 2022-02-03
Ruby 3.1: Add Enumerable#compact

k77ch7:
Add Enumerable#compact

The following tests pass.
 * test_compact in test/mri/ruby/test_enum.rb
 
For #7015


headius:
You are in a roll! Thank you!

k77ch7:
I would make more PRs to improve the  compatibility of current master with 3.1.0 .

headius:
That would be wonderful. We are starting to move toward stabilizing 9.4, but we won't release until we feel confident about our compatibility level. Your help is greatly appreciated!

PR #7067, opened 2022-02-03
Fix Enumerable#tally for some edge cases

k77ch7:
The following tests pass.
 * test_tally in test/mri/ruby/test_enum.rb


headius:
🙇‍♂️

PR #7068, opened 2022-02-03
Update valid date range

edipofederle:


headius:
Hmm I merged this before I noticed that no CI jobs ran. Could you throw another PR together for 9.3, since it has a full complement of GHA jobs?

headius:
Or perhaps 9.3 was not affected?

edipofederle:
> Or perhaps 9.3 was not affected?

checking

edipofederle:
@headius 9.3 and 9.4 are affected by this as well.

PR #7073, opened 2022-02-05
Update Enumerable#grep for edge case

k77ch7:
In mri 3.1.0, enum_grep0 function in enum.c uses 'rb_method_basic_definition_p' macro to check === is defined as basic method.
As far as I see, the current master does not have any funtions equivalent to 'rb_method_basic_definition_p'.
Therefore I add checkMethodBasicDefinition to RubyClass.

This makes the following tests pass.
 * test_grep_optimization in test/mri/ruby/test_enum.rb

### DIFF
--- a/core/src/main/java/org/jruby/RubyClass.java
+++ b/core/src/main/java/org/jruby/RubyClass.java
@@ -820,6 +820,12 @@ public class RubyClass extends RubyModule {
         return method.call(context, self, entry.sourceModule, name, arg0, arg1, arg2);
     }
 
+    // MRI: rb_method_basic_definition_p
+    public boolean checkMethodBasicDefinition(String name) {
+        DynamicMethod method = searchMethod(name);
+        return method != null && method.isBuiltin();
+    }
+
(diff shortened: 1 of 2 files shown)


enebo:
Interesting.  I bet we have that same logic in multiple places.

k77ch7:
I agree with you. Searching by "rb_method_basic_definition_p", I found some files.
[https://github.com/ruby/ruby/search?l=C&p=2&q=rb_method_basic_definition_p]()

enebo:
Ah yeah!

k77ch7:
I guess this code in test/mri/ruby/test_range.rb is related to same logic. 
```ruby
  def test_last_with_redefine_each
    assert_in_out_err([], <<-'end;', ['true'], [])
      class Range
        remove_method :each
        def each(&b)
          [1, 2, 3, 4, 5].each(&b)
        end
      end
      puts [3, 4, 5] == (1..10).last(3)
    end;
  end
```
I will try digging into.

Issue #7070, opened 2022-02-04
Regression: 9.3 converts LoadError to NameError when autoloading repeatedly and loading fails and file being loaded is not in current directory


p-mongo:
In the Ruby MongoDB driver we have  a module that requires an external C library. When this module is loaded it attempts to load the library, which could fail if the library is missing. We then autoload this module so that the user can reference it again later and, if the library is put in place, the module would work.

I converted our logic to the following reproduce shell script which creates two files, a.rb and t/b.rb, sets up B to be autoloaded from t/b.rb, but in t/b.rb there is only a LoadError raised:

```
mkdir -p t

cat > t/b.rb <<EOT
# This line is needed for 9.2, which would not attempt to autoload again otherwise.
# MRI and 9.3 do not need it.
autoload :B, 'b'

raise LoadError, 'custom message'
EOT

cat >a.rb <<EOT
$: << 't'
autoload :B, 'b'
begin
p B
rescue Exception => e
p e
end
p B
p 'done'
EOT

ruby a.rb
```

Having b.rb not in the current directory is important.

When this is run on MRI, it produces:

```
root@1d66935478aa:/# ruby -v
ruby 2.7.0p0 (2019-12-25 revision 647ee6f091) [x86_64-linux-gnu]
root@1d66935478aa:/# mkdir -p t
root@1d66935478aa:/# 
root@1d66935478aa:/# cat > t/b.rb <<EOT
> autoload :B, 'b'
> raise LoadError, 'custom message'
> EOT
root@1d66935478aa:/# 
root@1d66935478aa:/# cat >a.rb <<EOT
> $: << 't'
> autoload :B, 'b'
> begin
> p B
> rescue Exception => e
> p e
> end
> p B
> p 'done'
> EOT
root@1d66935478aa:/# 
root@1d66935478aa:/# ruby a.rb
#<LoadError: custom message>
Traceback (most recent call last):
	3: from a.rb:8:in `<main>'
	2: from /usr/lib/ruby/2.7.0/rubygems/core_ext/kernel_require.rb:92:in `require'
	1: from /usr/lib/ruby/2.7.0/rubygems/core_ext/kernel_require.rb:92:in `require'
/t/b.rb:2:in `<top (required)>': custom message (LoadError)
```

LoadError with the custom message is raised twice.

Same result on 9.2:

```
root@1d66935478aa:/# ruby -v
[...]

PR #7074, opened 2022-02-07
Remove the require lock on LoadError

headius:
Leaving this lock in place indicates to later feature tests that
the file was never loaded, incorrectly causing autoload constant
lookups to fail to load the file a second time. Instead, we
always remove any lock set up for the require so the feature does
not look like it is still in the process of loading.

See #7070

This is a WIP and we will see how it handles tests. I believe I originally added the `destroyLock` logic to form fit behavior to some spec, but I cannot remember which spec behavior led me to doing this. It seems wrong now to leave the lock in place under any circumstances, since it will interfere with future attempts to load that file (the case in #7070).

### DIFF
--- a/core/src/main/java/org/jruby/runtime/load/LoadService.java
+++ b/core/src/main/java/org/jruby/runtime/load/LoadService.java
@@ -68,6 +68,7 @@ import org.jruby.RubyThread;
 import org.jruby.ast.executable.Script;
 import org.jruby.exceptions.CatchThrow;
 import org.jruby.exceptions.JumpException;
+import org.jruby.exceptions.LoadError;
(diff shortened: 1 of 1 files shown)


headius:
Specs did not like the first attempt here to clear locks no matter what!

When an error is raised from a loading file, we should allow any other concurrent require to proceed in its attempt to load the file. The LoadError case should clear the lock so that the file does not appear to have successfully loaded.

This logic around the lock may need some extra bits, like detecting if there are others waiting on this require. I believe the cases are as follows:

* LoadError raised when attempting to load the file, assume other threads would also error and remove the lock.
* Any other StandardEror, allow other threads to attempt the load with the same lock.

PR #7075, opened 2022-02-07
Prefer the native thread status

### LINKED ISSUE #7032: Ruby thread waiting for java.lang.Thread with join does not go to sleep state
**Environment Information**
  
jruby 9.3.3.0 (2.6.8) 2022-01-19 b26de1f5c5 OpenJDK 64-Bit Server VM 11.0.13+8-Ubuntu-0ubuntu1.20.04 on 11.0.13+8-Ubuntu-0ubuntu1.20.04 +jit [linux-x86_64]

**Expected Behavior**
When waiting for a ruby thread to join, the current thread goes into sleep state.

ruby-thread.rb
```ruby
ruby_thread = Thread.current
ruby_thread2 = Thread.new do 
    sleep 1
    p ruby_thread.status
end
ruby_thread2.join
```
$ jruby ruby-thread.rb
"sleep"

**Actual Behavior**
However, when waiting for a java thread to join, the current thread will not go into sleep state.

java-thread.rb
```ruby
ruby_thread = Thread.current
java_thread = java.lang.Thread.new do 
    sleep 1
    p ruby_thread.status
end
java_thread.start
java_thread.join
```
$ jruby java-thread.rb
"run"

headius:
The native status of a thread will be more accurate when the
thread is in a blocking or sleeping state, since we often cannot
detect these states before making calls into Java libraries. The
status we manage on our own is only set when calling through our
own blocking call wrappers, which are not present for blocking
calls made directly to Java.

The logic here uses the native status if it is anything other
than "new" or "run". Only if the native thread appears to be
runnable do we fall back on the Ruby thread status we maintain.

This appears to fix #7032 but it is unknown what other impact it
may have.

### DIFF
--- a/core/src/main/java/org/jruby/RubyThread.java
+++ b/core/src/main/java/org/jruby/RubyThread.java
@@ -1775,11 +1775,16 @@ public class RubyThread extends RubyObject implements ExecutionContext {
     }
 
     private Status getStatus() {
-        Status status = STATUS.get(this);
+        Status statusFromNative = nativeStatus();
+        Status statusFromRuby = Status.NATIVE;
 
-        if (status != Status.NATIVE) return status;
+        if (statusFromNative == Status.RUN) {
+            statusFromRuby = STATUS.get(this);
+        }
 
-        return nativeStatus();
+        if (statusFromRuby != Status.NATIVE) return statusFromRuby;
+
+        return statusFromNative;
     }
 
     private Status nativeStatus() {


headius:
This is a small but very visible change, so it may be better to merge into 9.4.

headius:
A few jobs seemed to hang so I'm rerunning the lot. Could be something about this is not quite right.

headius:
Moved to 9.4 as this is too high risk a change to make in 9.3.x.

PR #7080, opened 2022-02-09
Align StringIO compatibility with CRuby

headius:
Most of the fixes here are of the following types:

* Encoding handling in StringIO construction
* Proper checking for open/closed/read/write
* Style fixes to arity-split more methods

The three remaining excludes are as follows:

* Two that test overflow conditions that are different or not
  relevant on JRuby's JVM-based String
* One where we fail to reject trancoding UTF-8 to Windows-31J,
  somewhere in mixed-encoding negotiation logic

This is in preparation for moving StringIO out to the gem in https://github.com/ruby/stringio/pull/21.


headius:
This was superseded by work in ruby/stringio#21 and #7178.

PR #7081, opened 2022-02-09
Expand ported logic for encoding negotiation

headius:
This expands our port of encoding negotiation logic, with the goal of fixing the remaining encoding-related failure in #7080. Unfortunately it does not appear to fix that issue.

Pushing this as a PR until I can figure out what I'm missing.


headius:
Rebased on master. This is likely safe to merge even though it did not fix the intended issue.

PR #7082, opened 2022-02-09
Fix marshal

enebo:
More or less ended up being a rewrite.  Not finished as I still need to handle Marshal#dump.


enebo:
Marshal#dump still needs work on some 3+ features (ruby2keywords, arity failings) but it is already pretty broken on encoding symbols and ivars encodings so I think it can be done as a followup piece.  Also I want to clear this up as I look at some other problems for a while.

PR #7086, opened 2022-02-10
WIP: inherit encoding for interned symbols

ahorek:


ahorek:
heh ok, not so simple :) https://github.com/ruby/ruby/blob/fc988c0be60f348e46719944dd31b3e3e92da91e/symbol.c#L753

I guess this is the right place according to the comment https://github.com/jruby/jruby/blob/851be1daa8b5f95fcff6abf5a494772c30efb2b5/core/src/main/java/org/jruby/RubySymbol.java#L920

PR #7097, opened 2022-02-14
Update buildDot function to fix failing TestRange#test_evaluation_order

k77ch7:
This commit avoids to evaluating beginNode and endNode twice in buildDot function.
e.g.
 arr = [1,2]
 (arr.shift)..(arr.shift) should be (1)..(2) , not be (nil)..(nil) .

The following tests pass.
 * test_evaluation_order in test/mri/ruby/test_range.rb


enebo:
Good catch!

PR #7098, opened 2022-02-14
Rework framing for super and __callee__

headius:
This is a series of commits to make super less dynamic (or exactly as dynamic as it needs to be) in order for us to support the `__callee__` method which does need to be dynamic.


rdubya:
I think fixing `__callee__` would fix this failure from the jdbc adapter tests:
```
  1) Failure:
ExcludingTest#test_raises_on_record_from_different_class [~/rails/activerecord/test/cases/excluding_test.rb:53]:
--- expected
+++ actual
@@ -1 +1 @@
-"You must only pass a single or collection of Post objects to #without."
+"You must only pass a single or collection of Post objects to #excluding."
```

headius:
Yeah we've been kicking around a shorter-term option that would allow us to fix `__callee__` without a major overhaul of framing. For now we are working on getting 9.4.2 out but we'll circle back to this soon.

headius:
Following up: we did get `__callee__` functional in https://github.com/jruby/jruby/pull/7702 by encoding it and the real method name on the stack. There may be a future change that adds a separate stack entry for the callee name, but what we have works without any drastic changes.

cc @rdubya

PR #7105, opened 2022-02-18
Update Range#min to fix failing TestRange#test_min

k77ch7:
The following tests pass.
 * test_min in test/mri/ruby/test_range.rb


headius:
Thank you!

PR #7107, opened 2022-02-18
Update Range#first to fix failing TestRange#test_first_last

k77ch7:
The following tests pass.
 * test_first_last in test/mri/ruby/test_range.rb


headius:
Thank you!

PR #7108, opened 2022-02-18
Update SECURITY.md

Slim002:


Slim002:
Personal user interface

headius:
Not sure what you were hoping to do here... can you explain?

PR #7111, opened 2022-02-20
Fix #7104

### LINKED ISSUE #7104: Integer#to_s, Integer#to_json includes null character(\x00, \u0000) on JRuby 9.3.*
**Environment Information**
  
* This issue can be reproduced on either `jruby-9.3.3.0` or `jruby-9.3.2.0`
* This issue can be reproduced on either `linux-x86_64` or `arm64`(Apple Silicon/M1 processor) platform.

**Expected Behavior**

* Integer#to_s, Integer#to_json behave the same as CRuby and previous JRuby versions, for example, `jruby-9.2.11.1` does not have this issue.

**Actual Behavior**

* JRuby 9.3.* adds an unexpected leading null character to _**certain integers**_, but not to all.

```
# irb
jruby-9.3.3.0 :001 > 999_999_999_999_999.to_s # adds an unexpected null character
 => "\x00999999999999999"
jruby-9.3.3.0 :002 > require 'json'
 => true
jruby-9.3.3.0 :003 > 999_999_999_999_999.to_json # adds an unexpected null character
 => "\u0000999999999999999"
jruby-9.3.3.0 :004 > 123_456_789_012_345.to_s # no leading null character
 => "123456789012345" 
```

k77ch7:
This commit changes the way to get the digits of an integer.

In Java, the Math.log10 function treats a long as double type.
When an integer is large, a rounding error occurs and sizeWithDecimalString function gets the wrong digits.


k77ch7:
I think this PR be able to backport to 9.3.x

enebo:
@k77ch7 Can you rebase this against jruby-9.3?  I think it should go back to 9.3 as well.  If I change base to jruby-9.3 using PR it ends up being the worlds largest merge :)

k77ch7:
@enebo I finish rebasing against jruby-9.3.

enebo:
Thanks @k77ch7 ... In what is becoming a pattern ... I wonder if we have this rounding problem elsewhere?  :)

k77ch7:
@enebo I'm not sure about that... 
The current master still has same problem. Should I make same PR against the master?

enebo:
@k77ch7 All fixes on jruby-9.3 will merge to master periodically.  So we will get this fix there pretty soon.

k77ch7:
@enebo Thanks. I understand.

PR #7116, opened 2022-02-23
Update Range#to_a and #last

k77ch7:
1. This commit updates Range#to_a simplify to fix failing TestRange#test_last_with_redefine_each.
I wonder if the deleted code in the to_a function contributes to the optimization for some cases.
But TestRange#test_last_with_redefine_each does not work because of the redefine #each method is not evaluated.
I give priority to passing the Range tests.

2. This commit also updates Range#last. In mri 3.1.0, the range_last function in range.c does optimization.
When the begin and the end of a range are integer type and  #each method of a range is not redefined,
the range_last function returns correct values without making a unnecessary array instance.


enebo:
@k77ch7 yeah I am not sure if it was for an optimizations but passing tests is more important.  Especially on a development branch.

PR #7118, opened 2022-02-24
Ruby3.0: Remove Time#succ

k77ch7:
This commit removes Time#succ . And this commit updates the discreteObject
function in RubyRange to fix failing spec/ruby/core/range/each_spec.rb .
The discreteObject function does not need to check the type of arg because of Time#succ is deleted.

See https://bugs.ruby-lang.org/issues/17039


headius:
Gotta love removing things!

PR #7123, opened 2022-02-28
Ruby 3.1: Update Time#getlocal to support +HHMM and +HHMMSS format

k77ch7:
MRI 3.1.0 supports this format.
The following tests pass.
 * test_evaluation_order in test/mri/ruby/test_time.rb


headius:
Thank you!

PR #7124, opened 2022-03-02
Greening our suites

headius:
Fixes and test changes to get our JRuby-specific test suites running green and matching Ruby 3.1 behaviors.


headius:
The remaining failure in `test:jruby` is due to this commit, which pulled in updates to `pathname` that broke absolute pathing for URLs:

https://github.com/ruby/pathname/commit/b39c268cff0a28660f071f1a77d94bf962118476

This was pulled in with stdlib updates. We will need to come up with a mechanism for determining absolute path that supports URIs.

PR #7125, opened 2022-03-02
Tag work-in-process specs and add separate CI job for them

headius:
The `wip` tag will mark specs we are currently working on and hoping to support for the next release. The `spec:ruby:wip` task will run only these tags, or you can specify it with `mspec run -g wip`.

The goal here is to get the baseline spec CI runs back to green, so we can detect regressions.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -202,7 +202,7 @@ jobs:
 
     strategy:
       matrix:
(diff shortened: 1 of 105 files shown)


headius:
@ahorek @k77ch7 Hey folks! We have changed things up so that spec:ruby:fast is green in CI, so we have a baseline that should not regress. The failing 2.7/3.0/3.1 specs have been tagged with `wip` and can be run with `rake spec:ruby:wip`. When we remove the `wip` tag, they will go back into the main suite and we can keep them green.

Pass this on to anyone else you see contributing!

PR #7132, opened 2022-03-10
Architecture tweaks for JRuby on M1

headius:
In https://github.com/sass/sassc-ruby/issues/231 it was reported that JRuby builds an improper `ARCH_FLAG` on MacOS/M1, using `aarch64` when it should use `arm64`. This PR fixes that.

This may be an appropriate change for other `aarch64` platforms depending on what the compiler wants, but I do not currently have a working Linux/aarch64 system on which to test it.

I also noticed that the `-v` output on JRuby reverses the arch and OS compared to CRuby, so this PR fixes that as well.

Two spec tags are updated to the new `arm64` arch name on `darwin`.

### DIFF
--- a/core/src/main/java/org/jruby/ext/rbconfig/RbConfigLibrary.java
+++ b/core/src/main/java/org/jruby/ext/rbconfig/RbConfigLibrary.java
@@ -103,6 +103,7 @@ public class RbConfigLibrary implements Library {
         String architecture = Platform.ARCH;
         if (architecture == null) architecture = "unknown";
         if ("amd64".equals(architecture)) architecture = "x86_64";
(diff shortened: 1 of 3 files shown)


ahorek:
```arm64``` is specific to Darwin, so the platform name shouldn't be changed on other aarch64 platforms,

I have an aarch64/Linux VM up, so just let me know if you need to test something.

headius:
@ahorek Ok thanks for confirming this and your other fix for the `-m` flag. I'll merge this, but perhaps your other one can be retargeted at 9.3? We want to continue improving M1 support in the 9.3 line.

headius:
> retargeted at 9.3

Oh nevermind, I guess it was already targeted at 9.3. Merged!

PR #7138, opened 2022-03-14
Use non-range cover logic for eqq and include

### LINKED ISSUE #7137: `Range#===` should not accept range argument
**Environment Information**
  
`jruby 9.3.3.0 (2.6.8) 2022-01-19 b26de1f5c5 OpenJDK 64-Bit Server VM 11.0.13+8-Ubuntu-0ubuntu1.18.04 on 11.0.13+8-Ubuntu-0ubuntu1.18.04 +jit [linux-x86_64]`

**Expected Behavior**

Unlike `Range#cover?`, `Range#===` does not accept range arguments.

```rb
# MRI 2.6.9
p (1..10).cover?(2..3) #=> true
p (1..10) === (2..3) #=> false
```

**Actual Behavior**

`(1..10) === (2..3)` returns true.

headius:
This code has for years mistakenly dispatched to the full cover?
logic, which considers intersecting ranges as being covered. The
logic in CRuby, however, only uses a small part of the cover logic
that does not accept ranges. The naming of the methods (in CRuby
it's r_cover_p, and in JRuby that got mapped to the full cover_p
method instead of the equivalent rangeInclude method) probably
was confusing, or early in JRuby development CRuby's r_cover_p and
our cover_p were equivalent and that changed over time in CRuby.

This also eliminates a dynamic dispatch to `cover?` that does not
exist in CRuby.

Fixes #7137

### DIFF
--- a/core/src/main/java/org/jruby/RubyRange.java
+++ b/core/src/main/java/org/jruby/RubyRange.java
@@ -846,7 +846,7 @@ public class RubyRange extends RubyObject {
         if (iterable
                 || !TypeConverter.convertToTypeWithCheck(context, begin, runtime.getInteger(), to_int_checked).isNil()
                 || !TypeConverter.convertToTypeWithCheck(context, end, runtime.getInteger(), to_int_checked).isNil()) {
-            return cover_p(context, val);
+            return RubyBoolean.newBoolean(context, rangeIncludes(context, val));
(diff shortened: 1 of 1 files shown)


headius:
Ok something is clearly not right with this patch. Will continue to investigate.

headius:
Created the PR wrong again, which is probably why it failed. 🤦‍♂️

PR #7142, opened 2022-03-16
Update jcodings

headius:
This PR will update 9.3 to latest jcodings. Some test fixes are included due to the `&apos` changes in https://github.com/ruby/ruby/pull/3177.

The commit that updates the jcodings version should be rewritten to a release version before merging.


headius:
See https://bugs.ruby-lang.org/issues/16922 for my backport request for these test changes.

headius:
Needs joni update to match dependencies. I'm working on that now.

Issue #6976, opened 2021-12-19
[Better error/notification message] "calling java_import on a non-Module object is deprecated"


rubyFeedback:
I have some legacy jruby code, found on the www.

When I run it I get:

"calling java_import on a non-Module object is deprecated"

That's it.

I know ruby fairly well but I am still a semi java noob (but getting
better) and a semi jruby noob too.

My first gripe is ... I don't get the line of the error. I can figure
out indirectly, I suppose, by removing the faulty lines e. g.

    import javax.swing.JFrame
    import javax.swing.ImageIcon
    import java.awt.Color

But it would be nice if we could see the error message such as
"line 5" or something. I think we get this via -w too in ruby.
Would be nice to have this for jruby by default too (I did not
yet run jruby via -w ... not sure if that works or not).

The second gripe, though, is more important: the message
does not tell me much. Ok, it is deprecated ... but so what,
what shall I do instead? Would it be possible to add some
more information, e. g. what to change, to make jruby
not complain about this? Both is not that important I 
suppose, but it would be nice as it can save a bit developer
time when improving old legacy code in this regard. Right
now I don't know what else to do ... I may find out, but
then I may not add that information to the issue here since
I will forget, so it may be useful to solve that for other
people too in the long run.

headius:
I merged in a change to include the line number and have a better message about `import` vs `java_import`.

PR #7143, opened 2022-03-16
Better warnings for deprecated java_import forms

headius:
* import in any context: warn of deprecation, recommend
  java_import instead.
* java_import from a non-module object: removed deprecated
  definition since it was being overwritten by toplevel
  definition and lazy import is a good use case to support.
* java_import from toplevel: works without warning.

Fixes https://github.com/jruby/jruby/issues/6976

```
$ jruby -e 'java_import java.lang.System; p System'
Java::JavaLang::System

$ jruby -e 'import java.lang.System; p System'
-e:1: warning: import is deprecated; use java_import
Java::JavaLang::System

$ jruby -e 'class X; def foo; java_import java.lang.System; p System; end; end; X.new.foo'
Java::JavaLang::System
```

### DIFF
--- a/core/src/main/ruby/jruby/java/core_ext/object.rb
+++ b/core/src/main/ruby/jruby/java/core_ext/object.rb
@@ -1,13 +1,13 @@
 class Object
   # :nodoc:
   # @deprecated
-  def java_import(*import_classes, &block)
-    warn "calling java_import on a non-Module object is deprecated"
-    Object.send :java_import, *import_classes, &block
+  unless respond_to?(:import)
+    def import(*import_classes, &block)
+      warn "import is deprecated; use java_import", uplevel: 1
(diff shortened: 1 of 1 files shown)


headius:
This does introduce a visible change that might not be appropriate: the old logic overwrite the deprecated instance-method `java_import` form when defining the toplevel `java_import`, but now the instance-method form will warn. This may be a pattern we actually want to continue supporting, for lazily importing Java classes before first usage.

headius:
I modified this to undeprecate the instance method form of `java_import` since it was being overwritten by the non-deprecated toplevel `java_import` and we would like to continue supporting the lazy-import use case:

```ruby
def foo
  java_import some.class.we.need.Now
  do something with Now
end
```

PR #7148, opened 2022-03-17
allow symbols in date formatter

ahorek:
MRI allows symbols

```ruby
require 'date'
Date._rfc3339(:'2001-02-03T04:05:06Z')
```
```
/jruby/lib/ruby/stdlib/date/format.rb:337:in `_parse': no implicit conversion of Symbol into String (TypeError)
        from /jruby/lib/ruby/stdlib/date/format.rb:471:in `_rfc3339'
```

```
{:year=>2001, :mon=>2, :mday=>3, :hour=>4, :min=>5, :sec=>6, :zone=>"Z", :offset=>0}
```


ahorek:
hmm, maybe I ran into some optimization bug...

there's a new test failure
```
Date.parse([])
```
which locally fails as expected
```
TypeError: no implicit conversion of Array into String
  check_input at /jruby/lib/ruby/stdlib/date/format.rb:613
       _parse at /jruby/lib/ruby/stdlib/date/format.rb:330
        parse at /jruby/lib/ruby/stdlib/date.rb:487
```
but on GHA ```_parse``` method seems to be omitted.
```
Date#parse raises a TypeError trying to parse non-String-like object ERROR
Expected TypeError
but got: ArgumentError (invalid date)
/home/runner/work/jruby/jruby/lib/ruby/stdlib/date.rb:446:in `new_by_frags'
/home/runner/work/jruby/jruby/lib/ruby/stdlib/date.rb:488:in `parse'
```

headius:
Well that's odd. Most likely a difference between jit and interpreted mode but I don't know of an optimization offhand that would just drop a backtrace frame like this.

ahorek:
```Date.parse``` doesn't allow symbols as an argument
```Date._rfc3339``` but these methods do

this behavior isn't intentional and it'll be deprecated https://github.com/ruby/date/commit/d57818f3b3cf52e1ae6d52ae0873dc997ed2c6cb

is it worth restoring it? Real apps shouldn't depend on it, but there're some MRI tests that would fail otherwise...

PR #7150, opened 2022-03-17
Allow reified classes to store instance variables in raw fields (fixed)

byteit101:
See https://github.com/jruby/jruby/pull/7012 though now hopefully fixed

### DIFF
--- a/core/src/main/java/org/jruby/runtime/ivars/RawFieldVariableAccessor.java
+++ b/core/src/main/java/org/jruby/runtime/ivars/RawFieldVariableAccessor.java
@@ -1,0 +1,198 @@
+/*
+ ***** BEGIN LICENSE BLOCK *****
+ * Version: EPL 2.0/GPL 2.0/LGPL 2.1
+ *
+ * The contents of this file are subject to the Eclipse Public
+ * License Version 2.0 (the "License"); you may not use this file
+ * except in compliance with the License. You may obtain a copy of
+ * the License at http://www.eclipse.org/legal/epl-v20.html
+ *
+ * Software distributed under the License is distributed on an "AS
+ * IS" basis, WITHOUT WARRANTY OF ANY KIND, either express or
+ * implied. See the License for the specific language governing
+ * rights and limitations under the License.
+ * 
+ * Alternatively, the contents of this file may be used under the terms of
+ * either of the GNU General Public License Version 2 or later (the "GPL"),
+ * or the GNU Lesser General Public License Version 2.1 or later (the "LGPL"),
+ * in which case the provisions of the GPL or the LGPL are applicable instead
+ * of those above. If you wish to allow use of your version of this file only
+ * under the terms of either the GPL or the LGPL, and not to allow others to
+ * use your version of this file under the terms of the EPL, indicate your
+ * decision by deleting the provisions above and replace them with the notice
+ * and other provisions required by the GPL or the LGPL. If you do not delete
+ * the provisions above, a recipient may use your version of this file under
+ * the terms of any one of the EPL, the GPL or the LGPL.
+ ***** END LICENSE BLOCK *****/
+
+package org.jruby.runtime.ivars;
+
+import com.headius.invokebinder.Binder;
+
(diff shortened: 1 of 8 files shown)


core/src/main/java/org/jruby/runtime/ivars/RawFieldVariableAccessor.java
```diff
@@ -0,0 +1,198 @@
...
+                            .cast(void.class, realClass.getReifiedClass(), basetype)
+                            .invoke(setter);
+    
+                    MethodHandle temp = MethodHandles.foldArguments(wrapperSetter, wrapperSetConvert);
+                    return temp;
+                }
+            }
+            
+        } catch (NoSuchMethodException |IllegalAccessException e) {
+            RaiseException r = Ruby.getGlobalRuntime().newRuntimeError("JRuby bug detected. Please file an issue if not already filed. Raw Field Accessor Setter gave errors: " + e.getMessage());
```
kares, line 160: could we rather use the runtime from `realClass` here, this doesn't seem right?

headius: Yes we should definitely use the realClass.getClassRuntime() rather than the global one.

headius:
I have confirmed this will be green once merged with my Logger test fixes: https://github.com/jruby/jruby/actions/runs/2022952391

However I can't figure out why this example does not work... where is the field?

```
$ jruby -e 'class Foo < java.lang.Object; java_field "java.lang.String stringField"; end; f = Foo.new; puts f.class.java_class.declared_fields.to_a'
private final org.jruby.java.proxies.ConcreteJavaProxy rubyobj.Foo.this$rubyObject
private static final org.jruby.javasupport.proxy.JavaProxyClass rubyobj.Foo.this$rubyProxyClass
private static final org.jruby.javasupport.Java$JCtorCache rubyobj.Foo.this$rubyCtorCache
private static final org.jruby.Ruby rubyobj.Foo.ruby
private static final org.jruby.RubyClass rubyobj.Foo.rubyClass
```

This seems to mimic the (working) specs so I must be doing something wrong here. Once I understand what I'm doing wrong, this is ready to merge.

headius (approved):
Changes are approved, but I need an explanation for why my naive example is not working.

byteit101:
I always run into this too: any usage of the methods that are shared between `jrubyc` and `become_java!` for `become_java!` purposes needs to require `jruby/core_ext`

```rb
irb(main):001:0> class Foo < java.lang.Object; java_field "java.lang.String stringField"; end; f = Foo.new; puts f.class.java_class.declared_fields.to_a
private final org.jruby.java.proxies.ConcreteJavaProxy rubyobj.Foo.this$rubyObject
private static final org.jruby.javasupport.proxy.JavaProxyClass rubyobj.Foo.this$rubyProxyClass
private static final org.jruby.javasupport.Java$JCtorCache rubyobj.Foo.this$rubyCtorCache
private static final org.jruby.Ruby rubyobj.Foo.ruby
private static final org.jruby.RubyClass rubyobj.Foo.rubyClass
=> nil
irb(main):002:0> require 'jruby/core_ext'
=> true
irb(main):003:0> class Foo2 < java.lang.Object; java_field "java.lang.String stringField"; end; f = Foo2.new; puts f.class.java_class.declared_fields.to_a
private final org.jruby.java.proxies.ConcreteJavaProxy rubyobj.Foo2.this$rubyObject
private static final org.jruby.javasupport.proxy.JavaProxyClass rubyobj.Foo2.this$rubyProxyClass
private static final org.jruby.javasupport.Java$JCtorCache rubyobj.Foo2.this$rubyCtorCache
private static final org.jruby.Ruby rubyobj.Foo2.ruby
private static final org.jruby.RubyClass rubyobj.Foo2.rubyClass
public java.lang.String rubyobj.Foo2.stringField
=> nil
irb(main):004:0> 

```

headius:
@byteit101 Oh boy, that does fix it. Ok, we should do something better than that in the future, but that's not your bug.

Let's merge this!

PR #7162, opened 2022-03-29
Update and fix failures in ostruct tests

headius:
This updates ostruct tests and fixes one of the resulting failures.

The two existing excludes were resolved:

* test_method_missing depends on the native method_missing not appearing in NoMethodError backtrace. Fixed by `omit = true` in the JRuby method_missing binding, which is a new change.
* test_to_h_with_block was likely fixed by updating ostruct.

Two new failures are due to ruby/ostruct#41 and will be fixed by ruby/ostruct#45: test_each_pair, test_initialize


headius:
Whenever https://github.com/ruby/ostruct/pull/45 is fixed and released, this PR can update to latest ostruct and remove the excludes.

PR #7164, opened 2022-03-31
[ji] continued cleanup of JavaClass and friends

kares:
**WiP - proof of concept** (testing against 9.3 due CI - expected to re-target against master - 9.4)

---
The motivation here is to cleanup Java Integration, avoid the duality of wrapping Java objects as to reduce user confusion when e.g. facing `JavaObject` vs a Java proxy, also shipping new features becomes easier ...

This is a follow-up on the work done in 9.3 to eliminate `JavaClass`: https://github.com/jruby/jruby/pull/6513
It takes things few steps further, namely:
- eliminated `JavaObject` and `JavaArray` wrapper (they still exist and point to `JavaProxy` / `ArrayJavaProxy`)
  * these are now no longer used - and their Java proxies provide the same ruby API
  * the only concern is if a user (extension) does manual wrapping in Java using e.g. `JavaObject.wrap` but even that is going to function to some extent as a lot of the existing 'unwrapping' methods still hande the deprecated type
- as `JavaClass` and friends was already deprecated in 9.3 we can now remove the Ruby impl methods
  * note `JavaConstructor` is smt still use internally (had 2 responsibilities) - thus stayed non-deprecated
- `synchronized { }` was adjusted to yield the actual value we're synchronizing (previously yielded `nil`)
- `JavaSupport` vs `JavaSupportImpl` methods are now split between the 2 - 
  * JavaSupport is smt users could use `runtime.getJavaSupport()` in an ext so have clearly the methods user usable there
  * the internals (meant for JRuby use only) are moved `JavaSupportImpl` (the methods are deprecated in `JavaSupport`)
  * no API breaking changes except maybe for some very deep internals
- some of `JavaClass` API is now moved elsewhere (the class is now deprecated), we still use some of it's API internally but the most crucial ones have been moved and can be used by the user e.g. `JavaUtil.getJavaClass`
- some cleanup refactoring work that showed up along the way

### DIFF
--- a/core/src/main/java/org/jruby/javasupport/JavaMethod.java
+++ b/core/src/main/java/org/jruby/javasupport/JavaMethod.java
@@ -45,18 +45,11 @@ import java.lang.reflect.Modifier;
 import java.lang.reflect.Type;
 
 import org.jruby.Ruby;
-import org.jruby.RubyBoolean;
-import org.jruby.RubyClass;
 import org.jruby.RubyInstanceConfig;
-import org.jruby.RubyModule;
-import org.jruby.RubyString;
-import org.jruby.anno.JRubyClass;
-import org.jruby.anno.JRubyMethod;
 import org.jruby.exceptions.RaiseException;
 import org.jruby.javasupport.proxy.ReifiedJavaProxy;
 import org.jruby.javasupport.proxy.JavaProxyClass;
 import org.jruby.javasupport.proxy.JavaProxyMethod;
-import org.jruby.runtime.ObjectAllocator;
 import org.jruby.runtime.ThreadContext;
 import org.jruby.runtime.builtin.IRubyObject;
 
@@ -65,11 +58,7 @@ import static org.jruby.util.CodegenUtils.getBoxType;
 import static org.jruby.util.CodegenUtils.prettyParams;
 import static org.jruby.util.RubyStringBuilder.ids;
 
-// @JRubyClass(name="Java::JavaMethod")
 public class JavaMethod extends JavaCallable {
-
-    //private final static boolean USE_HANDLES = RubyInstanceConfig.USE_GENERATED_HANDLES;
-    //private final static boolean HANDLE_DEBUG = false;
 
     private final Method method;
     private final Class<?> boxedReturnType;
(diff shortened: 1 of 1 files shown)


core/src/main/java/org/jruby/javasupport/proxy/JavaProxyClass.java
```diff
@@ -549,16 +550,14 @@ public static JavaProxyClass setProxyClassReified(final Ruby runtime, final Ruby
...
-        if (objects != null) ensureStaticIntConsumed();
+        if (objects != null) ensureStaticInitConsumed();
         lookup.set(objects);
-        if (objects == null) return 0;
-        int val = objects.hashCode();
-        return val;
+        return Arrays.hashCode(objects); // 0 if null
     }
 
-    public static void ensureStaticIntConsumed() {
```
byteit101, line 561: Technically, this wasn't actually a typo, as were were dealing with a "static int" in the generated class that must have the corresponding lookup consumed, but fine to change

core/src/main/java/org/jruby/java/proxies/ConcreteJavaProxy.java
```diff
@@ -286,8 +283,8 @@ public IRubyObject call(ThreadContext context, IRubyObject self, RubyModule claz
             }
 
             if (ctor == null) {
-                JavaObject jo = (JavaObject) initialize.call(context, self, clazz, "new", args);
```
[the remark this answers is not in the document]
byteit101: This is possible to hit, but only with nonstandard and untested configurations. See the java class configurator, ruby ctor boolean.

core/src/main/java/org/jruby/javasupport/proxy/JavaProxyClass.java
kares: oh okay, thanks - did not realize the Int was intended.

core/src/main/java/org/jruby/javasupport/JavaMethod.java
```diff
@@ -196,92 +181,92 @@ public final int hashCode() {
...
-    public IRubyObject type_parameters(ThreadContext context) {
-        return Java.getInstance(context.runtime, method.getTypeParameters());
-    }
+//    @JRubyMethod
+//    @Override
+//    public RubyString name(ThreadContext context) {
+//        return context.runtime.newString(method.getName());
+//    }
+//
+//    @JRubyMethod(name = "public?")
```
enebo, line 190: Do you anticipate we may need this code still?  Is this some form of documentation of how we used to do things?  Otherwise you can delete this.

kares: oh man, good catch - thought I already deleted these ...

PR #7167, opened 2022-03-31
Dup strings being set into env

### LINKED ISSUE #7166: Encoding of string created from literal changes as side effect of ENV mutation
Hey hey! 👋 

As always, thanks for making JRuby awesome! I'm working on adding JRuby 9.3 to the [`ddtrace` gem CI](https://github.com/DataDog/dd-trace-rb) and ran into this really weird behavior 👀 

**Environment Information**
  
Provide at least:
* JRuby version: `jruby 9.3.4.0 (2.6.8) 2022-03-23 eff48c1ebf OpenJDK 64-Bit Server VM 11.0.14.1+1 on 11.0.14.1+1 +jit [x86_64-linux]`
* Operating system and platform: `Linux docker-desktop 5.10.104-linuxkit #1 SMP Wed Mar 9 19:05:23 UTC 2022 x86_64 GNU/Linux`

**Expected Behavior**

Testcase:

```ruby
puts RUBY_DESCRIPTION

utf8 = ''.encode(::Encoding::UTF_8).freeze
puts utf8.encoding

ENV['FOO'] = ''

puts utf8.encoding
```

When executed in the `jruby:9.3.4.0-jre11` docker image with `LANG=en_US.UTF-8`, e.g.:

```
$ docker run -ti -v `pwd`:/working jruby:9.3.4.0-jre11 /bin/bash
root# LANG=en_US.UTF-8 jruby working/encoding_bug.rb
jruby 9.3.4.0 (2.6.8) 2022-03-23 eff48c1ebf OpenJDK 64-Bit Server VM 11.0.14.1+1 on 11.0.14.1+1 +jit [x86_64-linux]
UTF-8
US-ASCII
```

This doesn't happen if don't change `LANG`. It also doesn't happen on 9.2, but seems to be there already in 9.3.0.0 so somewhere in the 9.3 work this bug came about.

headius:
These strings should not reference their old objects, since we
may need to negotiate a new encoding for them to fit the host
system.

Fixes #7166


headius:
Ok after reviewing it I am reminded that CRuby doesn't really have a separate store for these values, so we can do whatever we want. They will make a copy of the contents as required for setenv/putenv, but otherwise they retrieve from the actual environment every time.

So I think we can go ahead with this defensive copy and not worry about doing too much allocation (and hopefully people aren't cranking through hundreds of ENV entries in critical path code anyway).

PR #7188, opened 2022-05-03
Fixed test fail regarding `Array#reject!`

PurityLake:
The method for `Array#reject!` didn't check after each successive block yield if the array was frozen cause a failed test


enebo (changes requested):
This is not quite right:

```text
mri31 -e 'a = [1,2,3]; a.delete_if { |e| puts "A"; a.freeze; true }'
A
A
A
```

So the isTrue is not when this occurs but when the safeArraySet happens (and also I guess after the loop since all things will be true -- so where context.nil is as well).

PurityLake:
@enebo I believe I have made that change for you

enebo:
@PurityLake Thanks for the extra work on this.

PR #7190, opened 2022-05-03
Fix failing test Array#rindex

PurityLake:
Fixed issue with `Array#rindex`  test failing. Using `unpack` to deal with concurrency error and `Object#define_singleton_method` to replace `class_eval` which seemed to not be working

### DIFF
--- a/test/mri/ruby/test_array.rb
+++ b/test/mri/ruby/test_array.rb
@@ -2574,10 +2574,8 @@ class TestArray < Test::Unit::TestCase
     assert_raise(StopIteration) { e.next }
 
     o = Object.new
-    class << o; self; end.class_eval do
-      define_method(:==) {|x| a.clear; false }
-    end
     a = [nil, o]
+    o.define_singleton_method(:==) { |x| a.clear; false }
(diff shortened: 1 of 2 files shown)


test/mri/ruby/test_array.rb
```diff
@@ -2574,10 +2574,8 @@ def test_rindex2
     assert_raise(StopIteration) { e.next }
 
     o = Object.new
-    class << o; self; end.class_eval do
```
enebo, line 2577: Did you modify this or are we out of date on this file?  We normally do not change the test files we receive.

enebo, line 2577: Did you modify this or are we out of date on this file?  We normally do not change the test files we receive.

PurityLake: Yes I made a change to the test, I will change that, sorry

enebo: Cool.  Thanks

PurityLake:
@enebo apologies, I was unaware of not being allowed to change the tests, I have reverted the file and now it is back the way it was

PR #7201, opened 2022-05-07
Upate Integer#<< and Integer#>> for some edge cases

k77ch7:


core/src/main/java/org/jruby/RubyBignum.java
```diff
@@ -882,8 +882,21 @@ public IRubyObject op_lshift(ThreadContext context, IRubyObject other) {
...
+                        if (value.signum() >= 0) {
+                            return RubyFixnum.zero(context.runtime);
+                        } else {
+                            return RubyFixnum.minus_one(context.runtime);
+                        }
+                    }
+                }
                 break;
             } else if (other instanceof RubyBignum) {
+                if (value == BigInteger.ZERO) {
```
kares, line 896: think it's safer to not assume only one zero instance:
```suggestion
                if (value.signum() == 0) {
```

kares (commented):
:+1: nice work, one suggestion on treating zero.

k77ch7 (commented):
@kares Thank you for your comment.

PR #7206, opened 2022-05-10
Remove env shebang on Linux

headius:
As of a later Linux 2.6 release (2.6.27.something) the kernel now
recognizes when another script is used as the shebang command and
properly runs that script rather than the binary that script needs
to launch (e.g. putting a bash script in the shebang line will run
the script, rather than bash). Because of that improvement, we no
longer need to use "env shebangs" on Linux, since the bin/jruby
shell script will launch correctly.

This change was released in late 2008, so any 2.6 kernels past
that date work properly. Linux 2.4 was EOL in 2011, so all up-to-
date Linux installs after that date should be on a 2.6 or higher
Kernel, very likely after the fix was landed. And 2.6 itself was
EOL in 2016, so all up-to-date Linux installs after that date will
certainly have the fix. We believe it is now safe for us to drop
the requirement that RubyGems install bin scripts using the env
shebang on Linux.

Note that this is still necessary for most other *nix environments
including the BSDs, Solaris, AIX, HP/UX, and so on.

### DIFF
--- a/lib/ruby/stdlib/rubygems/defaults/jruby.rb
+++ b/lib/ruby/stdlib/rubygems/defaults/jruby.rb
@@ -18,13 +18,15 @@ module Gem
     end
   end
 
-  def self.platform_defaults
-    return {
-        'install' => '--env-shebang',
-        'update' => '--env-shebang',
-        'setup' => '--env-shebang',
-        'pristine' => '--env-shebang'
-    }
+  unless RbConfig::CONFIG['host_os'].match? /linux/
+    def self.platform_defaults
+      return {
+          'install' => '--env-shebang',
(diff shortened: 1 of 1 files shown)


headius:
cc @mrnoname1000 @enebo This is the shebang change we discussed. The obvious advantage here is that we no longer have to customize these defaults AT ALL on Linux. The additional benefit is that the bin scripts will now get absolute path shebangs and no longer need the `jruby` command in PATH to run properly (which should carry over to bin scripts installed in other places, like vendored bundles and such).

headius:
This looks good, with the same set of green test runs as before the change. Merging.

PR #7207, opened 2022-05-11
Fixed failing tests regarding `Array#[]=`

PurityLake:
`Array#[]=` was not raising the correct exceptions, `RubyRange` also needed ajusting


PurityLake:
Mistakenly didn't pass excludes to `runner.rb`

PR #7215, opened 2022-05-15
Ruby3.1: Add Queue.new with an Enumerable of initial values

k77ch7:
See #7015

### DIFF
--- a/core/src/main/java/org/jruby/ext/thread/Queue.java
+++ b/core/src/main/java/org/jruby/ext/thread/Queue.java
@@ -34,6 +34,7 @@ import java.util.concurrent.locks.Condition;
 import java.util.concurrent.locks.ReentrantLock;
 
 import org.jruby.Ruby;
+import org.jruby.RubyArray;
 import org.jruby.RubyBoolean;
 import org.jruby.RubyClass;
 import org.jruby.RubyMarshal;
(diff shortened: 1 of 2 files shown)


core/src/main/java/org/jruby/ext/thread/Queue.java
```diff
@@ -269,6 +271,23 @@ public IRubyObject initialize(ThreadContext context) {
         return this;
     }
 
+    @JRubyMethod(visibility = Visibility.PRIVATE)
+    public IRubyObject initialize(ThreadContext context, IRubyObject items) {
+        this.capacity = Integer.MAX_VALUE;
+        Ruby runtime = context.runtime;
+        IRubyObject tmp = TypeConverter.convertToTypeWithCheck(context, items, runtime.getArray(), context.sites.TypeConverter.to_a_checked);
```
enebo, line 278: @k77ch7 I think if to_a returns something which is not a RubyArray this will end up with a classcastexception.  We do not have this version of initialize at all so this is still commit worthy.

k77ch7: @enebo If the provieded argument does not respond to `#to_a`, the tmp variable becomes to be nil. And the following spec fails.

```
    it "raises if the provided Enumerable does not respond to #to_a" do
      enumerable = MockObject.new('mock-enumerable')
      -> { Queue.new(enumerable) }.should raise_error(TypeError, "can't convert MockObject nito Array")
    end
```

So, when tmp is nil, this logic throws TypeError to pass the above spec. Are there any other ways?

enebo: @k77ch7 My apologies.  You are right convertToTypeWithCheck will raise if it is anything other than RubyArray of RubyNil.  sorry.  Thanks for your comment and the PR.

PR #7230, opened 2022-05-29
Update IO.new, IO.open and IO.for_fd to raise ArgumentError for some edge cases

k77ch7:
For instance, IO.for_fd(IO.sysopen("tmp.rb"), "w", nil) shoud raise "`for_fd': wrong number of arguments (given 3, expected 1..2) (ArgumentError)".
This commit also decreases wip tag in rspecs.


k77ch7:
@enebo no problem

Issue #7225, opened 2022-05-23
JarCache still slow at startup in several Windows environments


DirkMahler:
**Environment Information**
  
* JRuby version: 9.2.20.1 (including latest 9.2.21.0-SNAPSHOT)
* Operating system: Windows 10

**Expected Behavior**

* Startup of jQAssistant/AsciidoctorJ which makes use of JRuby should be a matter of milliseconds or at most seconds

**Actual Behavior**

* The issue which had been fixed in #6730 re-appeared in virtualized, highly secured corporate Windows environments of a customer, startup again takes 5-8 minutes

**Possible Solution**

* The solution in #6730 was to reduce the amount of expensive File.lastModified calls in JarCache by introducing a grace-period of 500ms between checks (see https://github.com/jruby/jruby/blob/jruby-9.2/core/src/main/java/org/jruby/util/JarCache.java#L45). I verified that increasing the value to 750ms or 1s solves the issue again and I can provide an according PR. The question is if this value should be made configurable.

DirkMahler:
@enebo Shall I close this ticket as the PR has been merged?

enebo:
@DirkMahler I didn't realize it hadn't been.  Thanks for pinging.

PR #7232, opened 2022-05-29
#7225 make JarCache check interval configurable and raise default val…

DirkMahler:
- JarCache check interval is now configurable using the system property "jruby.jarCache.lastModifiedCheckInterval"
- The default value has been raised to 750ms

### DIFF
--- a/core/src/main/java/org/jruby/util/cli/Options.java
+++ b/core/src/main/java/org/jruby/util/cli/Options.java
@@ -186,6 +186,7 @@ public class Options {
     public static final Option<Boolean> USE_FIXNUM_CACHE = bool(MISCELLANEOUS, "fixnum.cache", true, "Use a cache of low-valued Fixnum objects.");
     public static final Option<Integer> FIXNUM_CACHE_RANGE = integer(MISCELLANEOUS, "fixnum.cache.size", 256, "Values to retrieve from Fixnum cache, in the range -X..(X-1).");
     public static final Option<Boolean> PACKED_ARRAYS = bool(MISCELLANEOUS, "packed.arrays", true, "Toggle whether to use \"packed\" arrays for small tuples.");
+    public static final Option<Integer> JAR_CACHE_TTL = integer(MISCELLANEOUS, "jarcache.ttl", 750, "The time (ms) between checks if a JAR file containing resources has been updated.");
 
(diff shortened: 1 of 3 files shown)


enebo:
@DirkMahler Thanks!  I can do this if you don't want to but we will put all our options through org.jruby.util.cli.Options.  This logic is nice because when you type 'jruby --properties' it shows all the properties we have with some documentation.  Then typically we will access that value via org.jruby.RubyInstanceConfig.  We do that so that it then become possible to change the config at runtime (I don't know if this is really important here but it is an optional idea -- is there any value in that?).  If you do have time to make that change it would be great.

DirkMahler:
Thanks for your feedback - I'll update the PR with the proposed changes.

core/src/main/java/org/jruby/util/cli/Options.java
```diff
@@ -186,6 +186,7 @@ public class Options {
     public static final Option<Boolean> USE_FIXNUM_CACHE = bool(MISCELLANEOUS, "fixnum.cache", true, "Use a cache of low-valued Fixnum objects.");
     public static final Option<Integer> FIXNUM_CACHE_RANGE = integer(MISCELLANEOUS, "fixnum.cache.size", 256, "Values to retrieve from Fixnum cache, in the range -X..(X-1).");
     public static final Option<Boolean> PACKED_ARRAYS = bool(MISCELLANEOUS, "packed.arrays", true, "Toggle whether to use \"packed\" arrays for small tuples.");
+    public static final Option<Integer> JAR_CACHE_UPDATE_CHECK_INTERVALL = integer(MISCELLANEOUS, "jarcache.updateCheck.intervall", 750, "The time (ms) between checks if a JAR file containing resources has been updated.");
```
kares, line 189: would have been nice to have some better naming e.g. `jar_cache.expiration.millis`
or at least consistency with casing (jarcache vs updateCheck) within one option and avoid typos (interval has one `l`)

enebo: @kares oh yeah the typo has to change anyways.  Since we have fixnum.cache I think jar.cache.expiration and the description tells us the units.

DirkMahler: I've created another PR, just containing the requested name change (that I'm absolutely comfortable with).

PR #7235, opened 2022-06-01
Implement most of Fiber#raise

headius:
This implements Fiber#raise by introducing a new transfer mechanism, FiberRequest, which can be used to do the raise safely on the Fiber side rather than trying to force it through thread mechanism. No specs regressed and nearly all Fiber#raise specs now pass, but I have not investigated the performance impact of this additional carrier object yet.

### DIFF
--- a/core/src/main/java/org/jruby/RubyKernel.java
+++ b/core/src/main/java/org/jruby/RubyKernel.java
@@ -873,6 +873,42 @@ public class RubyKernel {
     public static IRubyObject sprintf(IRubyObject recv, IRubyObject[] args) {
         return sprintf(recv.getRuntime().getCurrentContext(), recv, args);
     }
+    public static IRubyObject raise(ThreadContext context, IRubyObject self, IRubyObject arg0) {
+        final Ruby runtime = context.runtime;
+        int argc = 1;
+        boolean forceCause = false;
+
+        // semi extract_raise_opts :
+        IRubyObject cause;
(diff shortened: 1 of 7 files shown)


core/src/main/java/org/jruby/RubyKernel.java
```diff
@@ -873,6 +873,42 @@ public static IRubyObject sprintf(ThreadContext context, IRubyObject recv, IRuby
     public static IRubyObject sprintf(IRubyObject recv, IRubyObject[] args) {
         return sprintf(recv.getRuntime().getCurrentContext(), recv, args);
     }
+    public static IRubyObject raise(ThreadContext context, IRubyObject self, IRubyObject arg0) {
+        final Ruby runtime = context.runtime;
+        int argc = 1;
+        boolean forceCause = false;
+
+        // semi extract_raise_opts :
+        IRubyObject cause;
```
evaniainbrooks, line 882: These are not referenced until line 906, they could be moved down.

Also it looks like argc is assigned `1`, doesn't change, and is then compared with `0`

headius: Good catch, I'll look into it.

headius: Addressed in b61cf077d3. This `raise` is a 1-arg convenience method, and I just failed to remove all unused logic when I made it.

Issue #7234, opened 2022-06-01
Inconsistent result of Math.ldexp


TonyCTHsu:
**Environment Information**

jruby -v
```
jruby 9.3.4.0 (2.6.8) 2022-03-23 eff48c1ebf OpenJDK 64-Bit Server VM 25.292-b10 on 1.8.0_292-b10 [x86_64-darwin]
```
uname -a
```
Darwin COMP-JKGQTW6NN4 21.5.0 Darwin Kernel Version 21.5.0: Tue Apr 26 21:08:37 PDT 2022; root:xnu-8020.121.3~4/RELEASE_ARM64_T6000 arm64 arm Darwin
```

**Expected Behavior**

The result of the `Math.ldexp` should be identical.

**Actual Behavior**

Ruby 2.6.7
```
irb(main):001:0> RUBY_DESCRIPTION
=> "ruby 2.6.7p197 (2021-04-05 revision 67941) [arm64-darwin21]"
irb(main):002:0> Math.ldexp(0.5122058490966879, 1024)
=> 9.207889385574391e+307
```

JRuby 
```
irb(main):001:0> RUBY_DESCRIPTION
=> "jruby 9.3.4.0 (2.6.8) 2022-03-23 eff48c1ebf OpenJDK 64-Bit Server VM 25.292-b10 on 1.8.0_292-b10 [x86_64-darwin]"
irb(main):002:0> Math.ldexp(0.5122058490966879, 1024)
=> Infinity
```

k77ch7:
I think this can be closed. This works in JRuby 9.3.6.0.

PR #7236, opened 2022-06-02
Fix Math.ldexp for #7234

k77ch7:
When the exponent is greater than 1023, This commit avoids overflow by converting Math.pow(2.0, exponent) into Math.pow(2.0, exponent-1023) * Math.pow(2.0, 1023).
Because Math.pow(2.0, 1024) is evaluated to INFINITY(overflow).

#7234

### DIFF
--- a/spec/regression/GH-7234_math_ldexp_spec.rb
+++ b/spec/regression/GH-7234_math_ldexp_spec.rb
@@ -1,0 +1,8 @@
+describe "Math.ldexp" do
+
+  it "returns correct value that closes to the max value of double type" do
+    Math.ldexp(0.5122058490966879, 1024).should == 9.207889385574391e+307
+    Math.ldexp(0.9999999999999999, 1024).should == 1.7976931348623157e+308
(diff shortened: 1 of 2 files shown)


spec/regression/GH-7234_math_ldexp_spec.rb
```diff
@@ -0,0 +1,8 @@
+describe "Math.ldexp" do
+
+  it "returns correct value that closes to the max value of double type" do
```
ahorek, line 3: could you move these tests to https://github.com/ruby/spec/blob/master/core/math/ldexp_spec.rb ? truffleruby has the same issue

k77ch7: I will do it tomorrow.

k77ch7: @ahorek 
See https://github.com/ruby/spec/pull/928 .
Is this okay?

ahorek: great, thanks!

PR #7239, opened 2022-06-07
allow ObjectSpace._id2ref to work for objects with finalizer

kares:
this is pretty much a low hanging fruit that allows the following to work (with `-O` as well):
`bin/jruby -v -e "obj = [1]; ObjectSpace.define_finalizer(obj, lambda { |_| }); p obj.object_id; p ObjectSpace._id2ref obj.object_id"`

... since internally the `define_finalizer` call assigns the `object_id` to the object.

maybe it's not a good idea, but seemed useful as well as dangerous, allowing to potentially de-reference an object in the finalizer: 
`ObjectSpace.define_finalizer(obj, lambda { |id| obj = ObjectSpace._id2ref(id)  })`


headius:
I think this is fine but I would like to know if this came up in normal usage. Was there a finalization case that needed this ability? I am reluctant to provide new ways for `id2ref` to work, and even less likely to make it easier for users to resurrect objects that have been dereferenced in this way.

Issue #7219, opened 2022-05-17
ClassCastException when accessing java.util.concurrent.ConcurrentHashMap


keegangroth:
### Environment Information
  
Provide at least:
* JRuby version (`jruby -v`) and command line (flags, JRUBY_OPTS, etc)
```
$ jruby --version
jruby 9.3.4.0 (2.6.8) 2022-03-23 eff48c1ebf OpenJDK 64-Bit Server VM 25.292-b10 on 1.8.0_292-b10 +jit [x86_64-darwin]
```
* Operating system and platform (e.g. `uname -a`)
```
$ uname -a
Darwin bos-m-kroth02 21.4.0 Darwin Kernel Version 21.4.0: Fri Mar 18 00:45:05 PDT 2022; root:xnu-8020.101.4~15/RELEASE_X86_64 x86_64
```

### Reproduction
```
# frozen_string_literal: true

require "bundler/inline"

gemfile(true) do
  source "https://rubygems.org"

  gem "minitest"
end

require "java"
require "minitest/autorun"

class ThreadSafeHash < java.util.concurrent.ConcurrentHashMap
end

class BugTest < Minitest::Test
  def test_hash_access
    assert ThreadSafeHash.new['test'].nil?
  end
end
```

### Expected Behavior
When subclassing `java.util.concurrent.ConcurrentHashMap` I should be able to access stored values using the standard ruby hash access without errors.

```
$ jruby --version
jruby 9.2.20.1 (2.5.8) 2021-11-30 2a2962fbd1 OpenJDK 64-Bit Server VM 25.292-b10 on 1.8.0_292-b10 +jit [darwin-x86_64]

$ jruby jruby-map.rb
Fetching gem metadata from https://rubygems.org/.
Resolving dependencies...
Using bundler 2.3.13
Using minitest 5.15.0
Run options: --seed 63725

# Running:

.

Finished in 0.012380s, 80.7784 runs/s, 80.7784 assertions/s.

1 runs, 1 assertions, 0 failures, 0 errors, 0 skips
```

### Actual Behavior
A `Java::JavaLang::ClassCastException` exception is thrown.

```
$ jruby --version
jruby 9.3.4.0 (2.6.8) 2022-03-23 eff48c1ebf OpenJDK 64-Bit Server VM 25.292-b10 on 1.8.0_292-b10 +jit [x86_64-darwin]

$ jruby jruby-map.rb
Fetching gem metadata from https://rubygems.org/.
Resolving dependencies...
[...]

PR #7248, opened 2022-06-19
Avoid ClassCastException when super class is a Java Map for #7219

k77ch7:
This commit uses checking class's type in several parts to avoid ClassCastException.

#7219


kares (approved):
:+1: the reification introduced in 9.3 failed to account for the `MapJavaProxy` special case

headius (approved):

PR #7259, opened 2022-06-27
Renamed option for JAR cache epiration

DirkMahler:
Changed option name according to discussion under #7232


enebo:
@DirkMahler Perfect timing!  Thanks for the updated PR.

PR #7281, opened 2022-08-01
Use the wide versions of these functions always

### LINKED ISSUE #7106: Fiddle::DLError: cannot find the function: RegQueryInfoKey()
**Environment Information**
  
Provide at least:
* jruby-head (jruby 9.4.0.0-SNAPSHOT (3.1.0) 2022-02-08 3562498c80 OpenJDK 64-Bit Server VM 25.322-b06 on 1.8.0_322-b06 +jit [mswin32-x86_64])
* Microsoft Windows Server 2022 (10.0.20348)

This is a friendly heads-up about edge jruby on windows, which [idnx tests against](https://github.com/HoneyryderChuck/idnx/runs/5247913031?check_suite_focus=true), and started failing, due to some missing expected API (RegQueryInfoKey()) when requiring `win32ole`.

headius:
The two delete functions were modified to be wide in CRuby in
response to https://bugs.ruby-lang.org/issues/10820 seven years
ago. The query function was mentioned here but not updated to be
wide. I am unsure how CRuby is locating the appropriate `A` or `W`
function given the bare name, but since we have largely forked
this file and always want to use the wide version, this change
seems appropriate.

Fixes #7106

### DIFF
--- a/lib/ruby/stdlib/win32/registry.rb
+++ b/lib/ruby/stdlib/win32/registry.rb
@@ -235,11 +235,11 @@ For detail, see the MSDN[http://msdn.microsoft.com/library/en-us/sysinfo/base/pr
         %w/RegEnumKeyExW    LLPPLLLP     L/,
         %w/RegQueryValueExW LPLPPP       L/,
         %w/RegSetValueExW   LPLLPL       L/,
-        %w/RegDeleteValue   LP           L/,
-        %w/RegDeleteKey     LP           L/,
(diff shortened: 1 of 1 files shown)


headius:
I do not have an answer yet as to why the non-W version of RegQueryInfoKey works on CRuby, nor why they did not explicitly bind it to the wide version in registry.rb, but I believe this is the correct change to make.

PR #7282, opened 2022-08-01
Fixes #7218. Empty rescue block returns exception object instead of nil.

### LINKED ISSUE #7218: Empty rescue block returns exception object instead of nil
**Environment Information**
  
Provide at least:
* JRuby version: `jruby 9.3.4.0 (2.6.8) 2022-03-23 eff48c1ebf OpenJDK 64-Bit Server VM 11.0.14+9-LTS on 11.0.14+9-LTS [x86_64-darwin]`
* Operating system and platform: `Darwin ivoanjo-COMP-C02DT18AML87 20.6.0 Darwin Kernel Version 20.6.0: Tue Feb 22 21:10:41 PST 2022; root:xnu-7195.141.26~1/RELEASE_X86_64 x86_64 i386 Darwin`

**Expected Behavior**

Testcase:

```ruby
puts RUBY_DESCRIPTION

def handle_failure_and_return_nothing
  File.open('does-not-exist')
rescue StandardError => _e
  # do nothing
end

puts handle_failure_and_return_nothing.inspect
```

Here's what I get on MRI:

```
ruby 3.1.2p20 (2022-04-12 revision 4491bb740a) [x86_64-darwin20]
nil
````

**Actual Behavior**

Here's what I get on JRuby:

```
jruby 9.3.4.0 (2.6.8) 2022-03-23 eff48c1ebf OpenJDK 64-Bit Server VM 11.0.14+9-LTS on 11.0.14+9-LTS [x86_64-darwin]
#<Errno::ENOENT: No such file or directory - does-not-exist>
```

Removing the `=> _e` on the original code seems to do the trick, since it wasn't using it anyway, but yeah it took a bit to figure out what was going on :)

enebo:
The fix was an error in our parser.  When we get exception variables in
a rescue statement we will make effectively '_e = $!' and then append
whatever is actually in the body of the exception body.  In MRI's case
it ends up being 'nil'.  In our case it was 'null'.  Our impl thinks it
is smart and realizes nothing is happening and what should you append
if it is null anyways and just left the body as only '_e = $!'.  We
needed '_e = $!; nil'.

Trivial fix is to make sure any null return from compstmt in this grammar
production will always just be nil.  Impact from this change should only
effect the case when a rescue body is empty and in the precence of an
exception variable.  What we did was wrong.  I see no risk.

### DIFF
--- a/core/src/main/java/org/jruby/parser/RubyParser.java
+++ b/core/src/main/java/org/jruby/parser/RubyParser.java
@@ -3710,7 +3710,7 @@ states[456] = (ParserSupport support, RubyLexer lexer, Object yyVal, ProductionS
 states[459] = (ParserSupport support, RubyLexer lexer, Object yyVal, ProductionState[] yyVals, int yyTop, int count, int yychar) -> {
                     Node node;
                     if (((Node)yyVals[-3+yyTop].value) != null) {
-                        node = support.appendToBlock(node_assign(((Node)yyVals[-3+yyTop].value), new GlobalVarNode(((Integer)yyVals[-5+yyTop].value), support.symbolID(lexer.DOLLAR_BANG))), ((Node)yyVals[-1+yyTop].value));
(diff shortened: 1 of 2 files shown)


enebo:
HEH.  I always do this #7218 is fixed by this (subject never links the two???).

PR #7284, opened 2022-08-01
Fixed tests in RubyTime failing with to_a

PurityLake:
Fixes a few tests that were failing in master.

`Time#to_a` returns its timezone as the short name only. `Time#to_s` seems to have a different interpretation of this using something of the form `GMT-06:00` but the spec expected `CST`

To fix this I just changed how `Time#to_a` receives it's timezone.

### DIFF
--- a/core/src/main/java/org/jruby/RubyTime.java
+++ b/core/src/main/java/org/jruby/RubyTime.java
@@ -956,7 +956,11 @@ public class RubyTime extends RubyObject {
             sec(), min(), hour(),
             mday(), month(), year(),
             wday(), yday(), isdst(),
-            dt.getZone().getShortName(dt.getMillis()));
+            getTimezoneShortName());
+    }
+
+    private RubyString getTimezoneShortName() {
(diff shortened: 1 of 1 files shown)


core/src/main/java/org/jruby/RubyTime.java
```diff
@@ -951,7 +951,16 @@ public String toString() {
...
-        return RubyArray.newArrayNoCopy(context.runtime, sec(), min(), hour(), mday(), month(), year(), wday(), yday(), isdst(), zone());
+        return RubyArray.newArrayNoCopy(
+            context.runtime,
+            sec(), min(), hour(),
+            mday(), month(), year(),
+            wday(), yday(), isdst(),
+            getTimezoneShortName());
+    }
+
+    private RubyString getTimezoneShortName() {
```
headius, line 962: You should pass runtime in from `to_a` above to avoid retrieving it a second time.

headius (changes requested):
One minor change needed but otherwise looks good.

headius:
Thanks!

Issue #7262, opened 2022-06-30
jar-dependencies cannot be updated out-of-band from jruby


ccutrer:
I want to make feature improvements to jar-dependencies, but when I install the updated gem into my system, bundler will refuse to load it because the default gem has already been installed. I've determined that the problem is that the rubygems/defaults/jruby.rb requires a file from jar-dependencies extremely early in the boot process of _any_ jruby process in order to hook into the rubygems install process. I plan on resolving this and sending PRs for both jruby and jar-dependencies, but want a tracking ticket first since it spans both projects.

**Environment Information**
  
* `jruby 9.3.1.0 (2.6.8) 2021-10-13 2e01e7199d OpenJDK 64-Bit Server VM 11.0.15+10-Ubuntu-0ubuntu0.22.04.1 on 11.0.15+10-Ubuntu-0ubuntu0.22.04.1 [linux-x86_64]`
* `Linux opennuc 5.15.0-37-generic #39-Ubuntu SMP Wed Jun 1 19:16:45 UTC 2022 x86_64 x86_64 x86_64 GNU/Linux`

**Expected Behavior**

`Gemfile`:

```
gem "jar-dependencies", "~> 1.0" # installed manually locally; no actual code changes required other than the change in version number
```

Running `bundle console` should boot without error.

**Actual Behavior**

```
cody@opennuc:~/test$ bundle console
[DEPRECATED] bundle console will be replaced by `bin/console` generated by `bundle gem <name>`
[DEPRECATED] This Gemfile does not include an explicit global source. Not using an explicit global source may result in a different lockfile being generated depending on the gems you have installed locally before bundler is run. Instead, define a global source in your Gemfile like this: source "https://rubygems.org".
Resolving dependencies...
Gem::LoadError: You have already activated jar-dependencies 0.4.1, but your Gemfile requires jar-dependencies 1.0.0. Since jar-dependencies is a default gem, you can either remove your dependency on it or try updating to a newer version of bundler that supports jar-dependencies as a default gem.
  check_for_activated_spec! at /home/cody/.gem/jruby/2.6.8/gems/bundler-2.3.17/lib/bundler/runtime.rb:309
                      setup at /home/cody/.gem/jruby/2.6.8/gems/bundler-2.3.17/lib/bundler/runtime.rb:25
                       each at org/jruby/RubyArray.java:1865
                       each at /home/cody/.gem/jruby/2.6.8/gems/bundler-2.3.17/lib/bundler/spec_set.rb:140
[...]

PR #7291, opened 2022-08-05
update jar-dependencies 0.4.2 reference in lib/pom.rb

ccutrer:
refs #7262

### DIFF
--- a/lib/pom.rb
+++ b/lib/pom.rb
@@ -28,7 +28,7 @@ default_gems = [
     ['ipaddr', '1.2.2'],
     ['irb', '1.0.0'],
     ['io-console', '0.5.9'],
-    ['jar-dependencies', '0.4.1'],
+    ['jar-dependencies', '0.4.2'],
     ['jruby-readline', '1.3.7'],
     ['jruby-openssl', '0.13.0'],
     ['json', '2.5.1'],
--- a/lib/pom.xml
+++ b/lib/pom.xml
@@ -177,7 +177,7 @@ DO NOT MODIFIY - GENERATED CODE
     <dependency>
       <groupId>rubygems</groupId>
       <artifactId>jar-dependencies</artifactId>
-      <version>0.4.1</version>
+      <version>0.4.2</version>
       <type>gem</type>
       <scope>provided</scope>
       <exclusions>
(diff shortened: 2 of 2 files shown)


headius:
We are so close!

The next problem is that the build itself uses JRuby, via a 9.3.0.0 version of jruby-complete.jar (see maven/jruby-jars/Mavenfile). This is run using the same installed gems as the currently-building JRuby, so it sees jar-dependencies 0.4.2, tries to activate it, but sees the same issue that prompted this whole issue. See the failing builds in this PR for the same old activation error.

I have merged this, since it gets things closer to green. The remaining work is figuring out how we can get the jruby-complete jar to work without seeing jar-dependencies 0.4.2.

ccutrer:
I think we should yank jar-dependencies 0.4.2, and re-publish as 1.0.0. And ideally we could mark 1.0.0 as requiring jruby 9.3.7.0. That would allow other users to not "see" the changed version in most cases. But I don't know how to do that.  There's only the `required_ruby_version` directive, which is the CRuby version, and AFAICT is 2.6.8 for both jruby 9.3.6.0 and 9.3.7.0.

headius:
@ccutrer Ok, I am concerned that this won't help because gems with dependencies on jar-dependencies do not have narrow-enough version descriptors. Going to poke around a bit to see if this is the case.

ccutrer:
Yeah, I am too. That's why I was looking at the required_ruby_version, to give rubygems even more of an excuse to ignore the new version, I just don't know if we can change to 2.6.9 arbitrarily, since it seems like that's supposed to track CRuby.

headius:
@ccutrer Yeah there's no easy way to tell RubyGems to require a specific version of JRuby because (I believe) the available required_ruby_version is only for the CRuby version or equivalent compatibility level in JRuby. We could perhaps bump our Ruby compat level but that isn't going to happen for the 9.3.7.0 which we hoped to put out this week.

In addition, with at least Psych having an open-ended jar-dependencies requirement, the upgraded gem is going to impact a lot of people on JRuby 9.3.6 and earlier. Even if we pinned a new release of Psych to 0.4.1, we'd still have lots of users out there with existing Gemfiles that suddenly break.

I'm going to yank the gem and revert the changes to our build and we'll take our time figuring out the most compatible way to do this update.

ccutrer:
👍

PR #7297, opened 2022-08-09
Fix Integer#[]

k77ch7:
The following test will pass.
 * test_aref in test/mri/ruby/test_integer.rb

### DIFF
--- a/core/src/main/java/org/jruby/RubyBignum.java
+++ b/core/src/main/java/org/jruby/RubyBignum.java
@@ -973,7 +973,7 @@ public class RubyBignum extends RubyInteger {
      *
      */
     @Override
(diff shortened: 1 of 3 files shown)


k77ch7:
@enebo could you review this?

enebo:
@k77ch7 Sorry.  When I saw the signatures changed I thought I needed to spend time reviewing this and then forgot to.  I was worried that old Java code would stop working but upon reviewing the two changed methods will still be just as visible.  So no problems.

k77ch7:
@enebo  Thanks. All tests of `test/mri/ruby/test_integer.rb` now pass except for the irrelevant warnings.

Issue #7256, opened 2022-06-24
Undefined constants when including a module that has been prepended


tannalynn:
**Environment Information**
 - `jruby 9.3.4.0 (2.6.8) 2022-03-23 eff48c1ebf Java HotSpot(TM) 64-Bit Server VM 14.0.1+7 on 14.0.1+7 [x86_64-darwin]`
- `Darwin  21.5.0 Darwin Kernel Version 21.5.0: Tue Apr 26 21:08:22 PDT 2022; root:xnu-8020.121.3~4/RELEASE_X86_64 x86_64`

**Expected Behavior**
When a module defines a constant (and has been prepended) is included in a class, accessing that constant through the class should be able to grab the value of the constant.
I tested this behavior on several versions of ruby between 2.3.8 up to 3.1.1, and CRuby is consistently able to use the constant without errors.

Running this snippet of code reproduces the issue.

```
module PrependTest
  def foo
    nil
  end
end

module Foobar
  prepend PrependTest
  TESTCONSTANT = "bar"
  def foo
    nil
  end
end

class Bar
  include Foobar
  def test
    nil
  end
end

Foobar::TESTCONSTANT # no error
Bar::TESTCONSTANT # error
```

**Actual Behavior**
When a module that has a constant defined, and has been prepended, is included in a class, `NameError: uninitialized constant` is thrown when accessing that constant through the class.
Using jruby to run the code snippet above results in 
```
NameError: uninitialized constant Bar::TESTCONSTANT
  const_missing at org/jruby/RubyModule.java:3948
         <main> at ./ruby_test_const.rb:23
```

nonlandi:
We could also reproduce this issue with `jruby-9.2.20.1`

headius:
Good find, thank you. We will look into it.

kares:
already fixed in JRuby 9.3.7.0 but for now I marked this for 9.2.22.0 (as there's been some discussion this might be considered)
... assuming the backport would be easy.

headius:
JRuby 9.2.22.0 becomes less likely to happen with each passing day. We have stopped pushing new Docker images for 9.2.x and there's very little demand for additional maintenance releases along that line.

We might consider just pushing release bits for a JRuby 9.2.22.0, but only if JRuby users do the footwork to pull together new backport pull requests. Our focus will be on mainaining 9.3 and 9.4.

enebo:
[...]

PR #7299, opened 2022-08-11
Fix for issue 7256: Undefined constants when including a module that has been prepended

evaniainbrooks:
Fixes https://github.com/jruby/jruby/issues/7256

### DIFF
--- a/core/src/main/java/org/jruby/IncludedModuleWrapper.java
+++ b/core/src/main/java/org/jruby/IncludedModuleWrapper.java
@@ -150,6 +150,11 @@ public class IncludedModuleWrapper extends IncludedModule {
     }
 
     @Override
+    protected ConstantEntry constantEntryFetch(String name) {
+        return origin.constantEntryFetch(name);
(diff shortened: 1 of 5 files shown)


enebo:
I did kick it off against master anyways to also approve CI run since you are first time contributor.

enebo:
I added @headius for reviewing this.  I approve this but it is mildly scary stuff in possible impact so we should double up on the review.  I think the PR is intuitive that the wrappers we make for include/prepend don't contain their own constants.  So we should be asking something for it and I believe origin is correct.  The actual vocabulary for this has always been confusing.   Last year I did a bunch of work on fixing some issues in prepend and I already forgot.

@evaniainbrooks This is a pretty deep-level sort of fix.  We appreciate the PR!

evaniainbrooks:
Thank you for the feedback @enebo 

I re-created this PR against the jruby-9.3 branch here: https://github.com/jruby/jruby/pull/7302

PR #7302, opened 2022-08-11
Fix for issue 7256 (jruby-9.3)

evaniainbrooks:
Fixes https://github.com/jruby/jruby/issues/7256

Re-creating this original PR (https://github.com/jruby/jruby/pull/7299) with base set to jruby-9.3

### DIFF
--- a/core/src/main/java/org/jruby/IncludedModuleWrapper.java
+++ b/core/src/main/java/org/jruby/IncludedModuleWrapper.java
@@ -150,6 +150,11 @@ public class IncludedModuleWrapper extends IncludedModule {
     }
 
     @Override
+    protected ConstantEntry constantEntryFetch(String name) {
+        return origin.constantEntryFetch(name);
+    }
+
+    @Override
     protected IRubyObject constantTableRemove(String name) {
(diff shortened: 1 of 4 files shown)


enebo:
@evaniainbrooks Thanks.  There is a mechanism for switching branches for the PR opener but it is like an easter egg.

Added comment from older PR here (I should add my approval is obviously contingent on a green CI run):

I added @headius for reviewing this. I approve this but it is mildly scary stuff in possible impact so we should double up on the review. I think the PR is intuitive that the wrappers we make for include/prepend don't contain their own constants. So we should be asking something for it and I believe origin is correct. The actual vocabulary for this has always been confusing. Last year I did a bunch of work on fixing some issues in prepend and I already forgot.

headius (approved):
I went over this with @enebo so we could understand what wasn't working and I think we are comfortable with it now. Basically the issue is that prepend changes the `methodLocation` to point at a layer above where we want to look, and delegating back to origin fixes that. It also makes sense to delegate constant lookups to the actual module and never attempt to do it from the include/prepend wrapper. All cases we have come up with work correctly.

Issue #7316, opened 2022-08-17
Incompatible error when nested `LoadError`


koic:
**Environment Information**
  
Provide at least:
* JRuby version (`jruby -v`) and command line (flags, JRUBY_OPTS, etc)

```console
% ruby -v
jruby 9.3.7.0 (2.6.8) 2022-08-16 c79ef237e0 Java HotSpot(TM) 64-Bit Server VM 25.271-b09 on 1.8.0_271-b09 +jit [x86_64-darwin]
```

* Operating system and platform (e.g. `uname -a`)

```console
% uname -a
Darwin atelier.local 19.6.0 Darwin Kernel Version 19.6.0: Mon Aug 31 22:12:52 PDT 2020; root:xnu-6153.141.2~1/RELEASE_X86_64 x86_64
```

Other relevant info you may wish to add:

This is a reproduction case.

```ruby
% cat example.rb
begin
  require 'feature-foo'
rescue ::LoadError => e1
  begin
    require 'feature/foo'
  rescue ::LoadError => e2
    raise e1 # NOTE: Raise `LoadError` if `e2` instead of `e1`.
  end
end
```

**Expected Behavior**

Raise `LoadError`.

JRuby 9.3.1.0:

```console
% ruby example.rb
LoadError: no such file to load -- feature-foo
  require at org/jruby/RubyKernel.java:1017
  require at /Users/koic/.rbenv/versions/jruby-9.3.1.0/lib/ruby/stdlib/rubygems/core_ext/kernel_require.rb:85
   <main> at example.rb:2
```

Same `LoadError` in JRuby 9.3.0.0 and MRI.

**Actual Behavior**

Raise `RuntimeError`.

JRuby 9.3.4.0 or higher:

```console
% ruby example.rb
RuntimeError: circular causes
  <main> at example.rb:7
LoadError: no such file to load -- feature/foo
  require at org/jruby/RubyKernel.java:1017
  require at /Users/koic/.rbenv/versions/jruby-9.3.4.0/lib/ruby/stdlib/rubygems/core_ext/kernel_require.rb:85
   <main> at example.rb:5
LoadError: no such file to load -- feature-foo
  require at org/jruby/RubyKernel.java:1017
  require at /Users/koic/.rbenv/versions/jruby-9.3.4.0/lib/ruby/stdlib/rubygems/core_ext/kernel_require.rb:85
   <main> at example.rb:2
```

[...]

PR #7321, opened 2022-08-19
Fix issue #7316 runtime error when nested raise

evaniainbrooks:
Fix for https://github.com/jruby/jruby/issues/7316

Revert d2d594f8b76cb52758cc6dd498eb9ed0b61032bf and use similar approach from CRuby: https://github.com/ruby/ruby/commit/18492887224cd1aa1aa862ed0fa5237df31ead81

### DIFF
--- a/core/src/main/java/org/jruby/Main.java
+++ b/core/src/main/java/org/jruby/Main.java
@@ -68,6 +68,8 @@ import java.lang.reflect.Proxy;
 import java.util.ArrayList;
 import java.util.List;
 import java.util.Map;
+import java.util.Set;
+import java.util.HashSet;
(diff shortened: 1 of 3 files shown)


headius:
This fix looks good at a glance, but we should target the `jruby-9.3` branch. Can you rebase the PR and I will do a more complete review?

Thanks for the help on this!

headius:
Am I reading this right? It seems like CRuby allows the circular cause to be created, but avoids recursing forever when displaying it by using a set?

evaniainbrooks:
@headius Yes, that is the conclusion I came to after reviewing how this was solved in CRuby. I've rebased branch to 9.3, cheers!

PR #7328, opened 2022-08-26
Use vthreads when available

headius:
Virtual threads from Project Loom allow spinning up many more fibers than when using native threads. This is an early patch to add vthread support to JRuby's fibers when the vthread API is available (starting as a preview feature in JDK19). 

Enable the feature when running on JDK 19:

* Pass `--enable-preview` to the JVM
* Pass `-J--enable-preview` to JRuby
* Put `--enable-preview` in `JAVA_OPTS` env

Tested with 100k fibers spun up, and then later resumed, using the following script:

```ruby
5.times do
  t = Time.now

  # create 100k fibers
  ary = 100_000.times.map { Fiber.new { } }

  # resume and complete 100k fibers
  ary.each(&:resume)

  p Time.now - t
end
```

Without vthreads, this script runs for a few seconds and then blows up, either raising memory errors or crashing the JVM. With vthreads it can complete multiple iterations, warming up to around 0.7s per loop on my system.

### DIFF
--- a/core/src/main/java/org/jruby/ext/fiber/ThreadFiber.java
+++ b/core/src/main/java/org/jruby/ext/fiber/ThreadFiber.java
@@ -1,8 +1,5 @@
 package org.jruby.ext.fiber;
 
-import java.lang.ref.WeakReference;
-import java.util.Map;
-import java.util.concurrent.atomic.AtomicReference;
 import org.jruby.Ruby;
 import org.jruby.RubyArray;
 import org.jruby.RubyBoolean;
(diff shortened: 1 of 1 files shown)


headius:
Going to land this to start letting folks preview virtual threads once 9.3.8.0 is out. A more elaborate implementation, possibly using executors to schedule fibers within a given thread, can come in 9.4 or after.

eregon:
I've been thinking about this.
Is it correct to just use a VirtualThread instead of a Thread for Fibers?

My concern is AFAIK VirtualThreads automatically switch on blocking IO and blocking synchronization calls.
Given that the typical approach to avoid having two Fibers running at the same time (when implemented on top of threads) is to use locks or queues, it sounds like VirtualThreads might switch to another VirtualThread/Fiber unexpectedly.
Although maybe such Fibers would also be blocked anyway if all Fibers of a Ruby Thread wait on the same lock/queue?

Issue #7326, opened 2022-08-24
sum method - undefined method 'last' for 1:Integer


djberg96:
jruby 9.3.6.0 (2.6.8) 2022-06-27 OpenJDK 64-Bit Server VM 14.0.2+12-46 on 14.0.2+12-46 +jit [x86_64-darwin]
MacOS 12.5.1 (Monterey)
  
I'm seeing a failure with `each_slice` + `sum`:

```
irb(main):001:0> w0 = 1; w1 = 2; w2 = 3;
=> 3
irb(main):002:0> index = [1,2,3]
=> [1, 2, 3]
irb(main):003:0> needed_bytes = (w0 + w1 + w2) * index.each_slice(2).sum(&:last)
```
With MRI this returns 30. With JRuby I see `NoMethodError (undefined method 'last' for 1:Integer)`.

Initially spotted via https://github.com/gettalong/hexapdf/issues/190

PR #7333, opened 2022-08-27
fix Enumberable sum when elements are array

evaniainbrooks:
Closes Issue #7326

### DIFF
--- a/spec/ruby/core/array/sum_spec.rb
+++ b/spec/ruby/core/array/sum_spec.rb
@@ -5,8 +5,14 @@ describe "Array#sum" do
     [1, 2, 3].sum.should == 6
   end
 
-  it "applies a block to each element before adding if it's given" do
-    [1, 2, 3].sum { |i| i * 10 }.should == 60
+  context "with a block" do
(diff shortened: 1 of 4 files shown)


spec/ruby/core/array/sum_spec.rb
```diff
@@ -5,8 +5,14 @@
...
 
-  it "applies a block to each element before adding if it's given" do
-    [1, 2, 3].sum { |i| i * 10 }.should == 60
+  context "with a block" do
+    it "applies the block to each element" do
+      [1, 2, 3].sum { |i| i * 10 }.should == 60
+    end
+
+    it "does not destructure the element if it is an array" do
+      [[1, 2], [3]].sum(&:last).should == 5
```
ahorek, line 14: this won't fail without the fix, we need something like
```ruby
[[1, 2], [3]].to_enum.sum(&:last)
```

evaniainbrooks: Thanks, moved to spec/ruby/core/enumerable/sum_spec.rb

PR #7343, opened 2022-09-05
Fix for changing method visibility changes behavior for module-inherited methods

### LINKED ISSUE #7240: Changing method visibility changes behavior for module-inherited methods
**Environment Information**
  
- JRuby 9.3.4.0
- `jruby:9.3.4.0-jre11` docker image

**Expected Behavior**

Here is an example of how CRuby behaves (2.7.5):

Screen Shot 2022-06-09 at 11 26 51 AM

As we can see, method visibility changed, but behavior did not.

**Actual Behavior**

Here is what happens for the same example on JRuby:

Screen Shot 2022-06-09 at 11 23 27 AM

Here, the behavior of the method unexpectedly changed after it was made public, preferring the signature of the first module that was overridden by the signature of the second.

evaniainbrooks:
closes #7240


headius:
Same minor issue here as in #7356; don't commit global whitespace changes... only correct whitespace changes on lines that have functional changes in the patch.

Great work otherwise, thank you!

Issue #7349, opened 2022-09-07
New timeout impl leaves dangling threads


headius:
This is discussed in detail in https://github.com/ruby/timeout/pull/21. I will be disabling the `testTimeoutCleanup` test in our `ScriptingContainerTest` temporarily until we can get a fix merged and released.

headius:
The test is still affected despite #7351 due to differences in how aggressively we shut down threads in an embedded scenario. Previously, we registered a hook that would shut down known timeout threads, as part of embedded teardown.

justinstoller:
Where is this at?

Before I jump into attempting a local workaround, Is there a patch we take up or a preferred work around similar to https://github.com/puppetlabs/puppetserver/commit/c3760842a350a2ab6fe8d4e6f0b041e6032e3d79 that we could employ?

PR #7351, opened 2022-09-08
Avoid accidentally killing the teardown thread

headius:
If the main thread performing the teardown appears earlier in this map, we may skip the teardown of remaining threads. Instead, we avoid the kill+join dance for the current thread to ensure all other threads get a chance to terminate.


headius:
Relates to #7349 and ruby/timeout#21.

headius:
This is the primary cause of failures on master in the `ScriptingContainerTest.testTimeoutCleanup` due to the new timeout implementation using a background thread for all timeouts, which is left dangling until process termination. When the teardown thread is prematurely killed, this dangling thread does not get cleaned up and the test fails.

The timeout library should still be making a best effort to shut down its thread gracefully (rather than waiting for it to be forcibly killed) but this change should get the test green on master.

Issue #7340, opened 2022-09-02
Proposal: define `foo=` regardless of return type of `setFoo(1)`


byteit101:
The increased popularity of fluent and fluent-ish API's in Java libraries has meant that there are more setters that don't return void. I propose the following:

If a java method has a name of `set([A-Z]\w*)` and a single parameter, define a ruby-style setter proxy. As a ruby-oriented proxy, it should return the argument passed, ignoring the true return value of the underlying java method.

The benefits of this approach are:

1. More seamless Ruby-Java integration for those Java libraries that define fluent or fluent-ish APIs
2. No need to check the Javadoc of the library to determine if the ruby form `#foo=` is defined or merely the default `#set_foo`. Reading blog posts, irb tab completion, and other code listings is sufficient to translate java-style calls to ruby-style calls
3. Simpler to explain when alternate method names are defined

The drawbacks of this approach are:
1. No ability to get the return value with ruby-style calls, Must use java method name, or snake-cased version
2. People expecting `#foo=` to fail on java objects won't have that failure anymore

Thoughts? Any benefits or drawbacks I've missed?

kares:
> The increased popularity of fluent and fluent-ish API's in Java libraries 

Is there any libraries in particular that you had in mind?

Have no objections against skipping the return check, except that is how bean-style properties where defined. I remember there's also special care to only support `boolean isFoo` but not `Boolean isFoo`.

byteit101:
> Is there any libraries in particular that you had in mind?

No, though in the past ~6 months I have ran into this with 3 separate libraries

kares:
have run into such setters as well but mostly around the builder pattern where the return is usually `this`.
in those cases the ruby style setter is actually redundant e.g.

`obj.set_foo(:foo).set_bar('bar').build` vs `obj.foo=(:foo).bar=('bar').build` :red_circle: 

again I am not against it just would like to keep the discussion going for a bit, if you recall any concrete examples please share them.

byteit101:
Yes, if you chain it it is, but that's only part of the type of setters I've seen, and I wouldn't use the ruby-style setters for that case anyway, as it wouldn't work like that anymore[1]. I've mostly used that style for single property setters, and not in chains.

I've scanned my m2 library for various types of setters, and here is what I found:

## Standard fluent `return this` chainables

Maven and Aether are two big contributors in this category, and I ran into it repeatedly when trying to make the `maven-requre` library earlier this year. Here is actually one of my experimentation's:
[...]

PR #7352, opened 2022-09-08
non-void `setFoo` now generates ruby-style proxies #7340

byteit101:
See #7340 for discussion


headius:
Looks pretty complicated, but I guess we can merge it!

PR #7355, opened 2022-09-09
Fix kernel clone

enebo:
Rekajgger the freeze keyword handling.


enebo:
LOL. Additionally (not important enough to remove since this is just for CI blessing) I refactored Module#include to examine modules in the argument order they are provided to match MRI (which semantically is different but I do not think you could observe this since it does not print out which module is not really a module).

Issue #7344, opened 2022-09-05
Changing method visibility changes super_method for module-inherited methods


evaniainbrooks:
Follow-up to https://github.com/jruby/jruby/issues/7240

Environment Information

JRuby 9.3.4.0

```
module A
  private
  def derp message
    puts [message, ' from A'].join
  end
end

module B
  private
  def derp
    puts 'from B'
    super("superclass")
    puts "subclass"
  end
end

class C
  include A
  include B
end

pp C.instance_method(:derp).super_method
# <UnboundMethod: A#derp>
C.send(:public, :derp)
pp C.instance_method(:derp).super_method
# <UnboundMethod: B#derp>
```

Expected behavior: `C.instance_method(:derp).super_method` should not change from `A#derp` to `B#derp` after changing method visibility,

evaniainbrooks:
Similarly, after calling `alias_method`, the `super_method` is not reported correctly for module-inherited methods. It is actually returning `nil`:

```
module A
  private
  def derp
  end
end

module B
  private
  def derp
    super
  end
end

class C
  include A
  include B
end

C.alias_method :meow, :derp
pp C.instance_method(:meow).super_method
# nil
```

CRuby returns the expected `C(A)#derp`

headius:
Good examples, thanks!

Confirmed both examples are broken on 9.3 HEAD.

headius:
At a glance, I would say the first example fails because we now have a method entry in C (used to wrap the private `derp` with a public invoker) and rather than searching for a super method starting at B#derp it tries to do it from C#derp.

The second example seems like it may simply be a bug with `super_method` in that it is not considering the original name of the method when searching for a superclass method.

In both cases, though, we are losing the context for the super call when installing one of these wrapper methods (visibility or alias).

headius:
Thank you!

PR #7356, opened 2022-09-10
Fix for method visibility changes super_method for module-inherited methods

evaniainbrooks:
closes https://github.com/jruby/jruby/issues/7344


headius:
This looks great, thank you! Could you clean up the few rogue whitespace issues? We try to avoid committing small whitespace changes because they end up flipping back and forth as people use different formatting standards.

headius:
Put a bit differently... if there are no functional changes on a given line, there should be no whitespace changes. I see a few places where tabs got (correctly) replaced with spaces on or near lines that actually had behavior changes, and those would be fine to include.

headius:
Much better, thanks! This is merged, and after your other PR runs another CI (I had to manually merge the test changes) we'll be good there too!

PR #7360, opened 2022-09-13
Investigating slow and hanging MRI tests

headius:
Some tests seem much slower lately and some are outright hanging. This PR is to investigate them, and possibly move more slow tests to a slow run.


k77ch7:
wow! it's very interesting.

k77ch7:
Certenly, the latest mri tests(https://github.com/jruby/jruby/actions/runs/3113604144) finished in 20-30 min.

headius:
Fixed through contributions and other stuff.

PR #7371, opened 2022-09-21
Fixes #7370: yield in class body produces two SyntaxError outputs

### LINKED ISSUE #7370: yield in class body produces two SyntaxError outputs
It's working, but why do we get two reports?

```
[] ~/work/jruby $ jruby -e 'class Foo; yield; end'
SyntaxError: -e:1: Invalid yield
SyntaxError: -e:1: Invalid yield

[] ~/work/jruby $ jruby -e 'class Foo; class << self; yield; end; end'
SyntaxError: -e:1: Invalid yield
SyntaxError: -e:1: Invalid yield
```

enebo:
Fixes #7370.  We print out twice because main script will try and precompile and we will raise a syntaxerror (RaiseError) but our code will blanket catch Exception/Throwable and end up trying to run the same code through the interpreter.


headius:
Looks good!

PR #7377, opened 2022-09-23
Also include/prepend into existing include hiers.

headius:
When including or prepending a module Y into a module X, we must also visit hierarchies that include X and add Y to them so they reflect the change.


headius:
I fixed the only failures that were actually related to this, in our own test_method_cache where we expected future includes would not be honored by existing classes.

PR #7378, opened 2022-09-24
Fix Hash#hash to return a correct value when key and value are same

k77ch7:
If using xor of key's hash and value's hash, a hash value always become to be 0. Therefore, to avoid generating 0 value, this commit uses hash of 16 bytes array which contains key's hash and values's hash.

The following test will pass.
 * spec/ruby/core/hash/hash_spec.rb

### DIFF
--- a/core/src/main/java/org/jruby/RubyHash.java
+++ b/core/src/main/java/org/jruby/RubyHash.java
@@ -1332,11 +1332,13 @@ public class RubyHash extends RubyObject implements Map {
         return context.runtime.newFixnum(hval[0]);
     }
 
+    private static final ThreadLocal<byte []> hash16byte = ThreadLocal.withInitial(() -> {return new byte[16];});
+
     private static final VisitorWithState<long[]> CalculateHashVisitor = new VisitorWithState<long[]>() {
         @Override
         public void visit(ThreadContext context, RubyHash self, IRubyObject key, IRubyObject value, int index, long[] hval) {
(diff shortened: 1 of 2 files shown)


headius:
The fix looks good, but perhaps there's a way we could avoid the extra allocation? This version will now allocate a new 16-byte array for every element in the Hash, where it did not allocate anything before.

headius:
A possible workaround: allocate a thread-local byte[16] that can be reused for all calculations. Since it's just a thread-local holding a byte[] it should not leak classes or classloaders and never "leak" more than 16 bytes per live thread.

k77ch7:
@headius please check the updated.

core/src/main/java/org/jruby/RubyHash.java
```diff
@@ -1332,11 +1332,13 @@ public RubyFixnum hash(ThreadContext context) {
         return context.runtime.newFixnum(hval[0]);
     }
 
+    private static final ThreadLocal<byte []> hash16byte = ThreadLocal.withInitial(() -> {return new byte[16];});
```
headius, line 1335: Style issues:

* No space before array declaration: `byte[]`.
* static final constant should be all caps: `HASH16BYTE` or similar

headius (changes requested):
Looks good! Now there are just two minor code style issues.

core/src/main/java/org/jruby/RubyHash.java
k77ch7: @headius thanks

headius (approved):

headius:
Thank you!

PR #7379, opened 2022-09-24
Fix Hash#rehash when a hash is large and contains duplicate keys

k77ch7:
This commit adds additional logic to remove duplicate key.  
 The following test will pass.
   * spec/ruby/core/hash/rehash_spec.rb

Before applies this patch, the rehash spec sometimes fails.
```
  it "removes duplicate keys for large hashes" do
    a = [1,2]
    b = [1]

    h = {}
    h[a] = true
    h[b] = true
    100.times { |n| h[n] = true }
    b << 2
    h.size.should == 102
    h.keys.should.include? a
    h.keys.should.include? b
    h.rehash
    h.size.should == 101
    h.keys.should.include? a
    h.keys.should_not.include? [1]
  end
```

### DIFF
--- a/core/src/main/java/org/jruby/RubyHash.java
+++ b/core/src/main/java/org/jruby/RubyHash.java
@@ -67,6 +67,7 @@ import java.util.AbstractCollection;
 import java.util.AbstractSet;
 import java.util.ArrayList;
 import java.util.Collection;
+import java.util.HashSet;
 import java.util.Iterator;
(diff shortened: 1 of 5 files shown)


k77ch7:
ping @enebo

enebo:
@k77ch7 totally forgot about this.

k77ch7:
@enebo thank you for your review.

PR #7376, opened 2022-09-23
Fix Thread#inspect and #to_s behaviors to match mri

k77ch7:
The following tests will pass except encoding test.
 * spec/ruby/core/thread/inspect_spec.rb
 * spec/ruby/core/thread/to_s_spec.rb

PR #7381, opened 2022-09-27
Various fixes for Thread#to_s

headius:
Some of this was already fixed by @k77ch7 in #7376 but those changes were made to match our own `test_inspect_and_to_s`, which had several lines modified to pass on JRuby and not run on CRuby. I have unmasked those lines, modified them to pass on CRuby, and adjusted JRuby behavior accordingly.


headius:
Looking much better now, and only MRI tests, WIP specs, sequel and concurrent-ruby, and some specific platforms (Windows, MacOS M1) are failing.

Issue #7386, opened 2022-09-30
Update SnakeYaml to 1.32+ due to CVE-2022-38752


chadlwilson:
Currently latest JRuby comes bundled with Snakeyaml `1.31` that includes a recently reported DoS vulnerability CVE-2022-38752 discovered by OSS Fuzz. Sadly the fix for this missed the changes made with #7342 and Jruby `9.3.8.0`

While this appears to be relatively low risk despite the language used to raise the issue, tweaks appear to have been made in SnakeYAML to better detect another recursive key usage (disabled by default already) and fail gracefully rather than with a StackOverflow.

See
- https://bitbucket.org/snakeyaml/snakeyaml/issues/531/stackoverflow-oss-fuzz-47081
- [Full changelog since 1.31 here](https://bitbucket.org/snakeyaml/snakeyaml/wiki/Changes)
- https://github.com/ruby/psych/pull/585
- https://github.com/ruby/psych/pull/586
- https://github.com/ruby/psych/pull/587

**Expected Behavior**

* Latest JRuby 9.3 comes bundled with snakeyaml version which does not report vulnerabilities

**Actual Behavior**

* Latest JRuby `9.3.8.0` comes bundled with SnakeYAML `1.31` via Psych `3.3.3`

PR #7387, opened 2022-09-30
Update to Psych 4.0.6

chadlwilson:
Updates SnakeYAML to 1.33 to address CVE-2022-38752 per #7386


headius:
Yay!

chadlwilson:
Yikes - lots of failures 😬

The milestone might need updating on this one btw (currently set to `9.3.9.0` instead of `9.4.0.0`)

headius:
I blame jetlag.

We expect to see lots of failures on master because we don't want to just tag off all of the 3.0 and 3.1 features we haven't finished, so we'll have to compare with a recent master build and see if we regressed.

PR #7388, opened 2022-09-30
Update to Psych 3.3.4

chadlwilson:
Updates SnakeYAML to 1.33 to address CVE-2022-38752 per #7386


chadlwilson:
I think the failing one here might just need a re-run; looks like possibly a flake/timeout.

headius:
Yeah the MacOS AArch64 builds can be a little flaky. I'm not concerned.

PR #7389, opened 2022-09-30
non-void setFoo now generates ruby-style proxies for existing java classes #7340

byteit101:
Add tests and fix setters not being generated for some java objects #7340


headius:
Looks good!

PR #7390, opened 2022-09-30
Fix #7323 by excluding java packages from ObjectSpace Module iteration

### LINKED ISSUE #7323: IRB completion causes error
Newer version of IRB have an autocomplete feature (I.e. 1.4.1). Unfortunately, if you use this under JRuby, it quickly explodes, because they use ObjectSpace to find all Modules that are active, and look at their instance_methods. But Java::JavaPackage instances are modules, but they go to great pains to remove almost all methods to make java-like access to packages possible. Including instance_methods. I'm not sure of the best approach to fixing this. I've solved it in pure Ruby in my irbrc, but it would probably be best implemented in Java in JRuby proper -- which wouldn't have the disadvantage of some of the gymnastics I have to do to locate and call the correct module and methods. I even went so far as to look up known java packages so that the completions might actually be useful.

My code:

```ruby
# this callback will be called; but has been removed; don't explode it
java.class.define_singleton_method(:method_added) { |_| nil }

java.class.define_singleton_method(:instance_methods) { |_| [] }

# this is not the right module, but we still want to use it for namespacing
# purposes
module Java
  module JavaPackage
    symbol_methods = %w[? !].freeze
    comparison_methods = %w[< = >].freeze
    INHERITED_INSTANCE_METHODS = (%i[
      class
      superclass
      singleton_class
      object_id
      name
      const_set
      singleton_method_added
      singleton_method_undefined
      singleton_method_removed
      define_singleton_method
    ] + java.__methods__.select do |m|
          m = m.to_s
          symbol_methods.include?(m[-1]) || comparison_methods.include?(m[0]) || m.start_with?("__")
        end).freeze
    INSTANCE_METHODS = %i[
      package_name
      to_s
      inspect
      const_missing
      const_get
      respond_to?
      respond_to_missing?
      method_missing
      available?
      sealed?
    ].freeze

    # can't call class_eval directly, so rebind to call it on the correct instance
    Module.instance_method(:class_eval).bind(java.class).call do
      def instance_methods(*args)
        return method_missing(:instance_methods, *args) unless args.length == 1

        @instance_methods ||= (::Kernel.java.lang.Package.packages
            .map(&:name)
            .select { |p| p.start_with?("#{package_name}.") && p != package_name }
            .map { |p| p[(package_name.length + 1)..].split(".").first.to_sym }
            .uniq + INSTANCE_METHODS).freeze
        return @instance_methods if args.first == false

        @inherited_instance_methods ||= (@instance_methods + INHERITED_INSTANCE_METHODS).freeze # rubocop:disable Naming/MemoizedInstanceVariableName
      end
    end
  end
end
```

byteit101:
IRB and Pry find modules via ObjectSpace, and arguably packages aren't  modules.
This change removes java packages from iteration. 

Fixes #7323 

While this change and the IRB change have overlap, I think both should be implemented for completeness

### DIFF
--- a/core/src/main/java/org/jruby/RubyObjectSpace.java
+++ b/core/src/main/java/org/jruby/RubyObjectSpace.java
@@ -40,6 +40,7 @@ import java.util.Map;
 
 import org.jruby.anno.JRubyMethod;
 import org.jruby.anno.JRubyModule;
+import org.jruby.javasupport.JavaPackage;
 import org.jruby.runtime.Block;
 import org.jruby.runtime.ThreadContext;
 import static org.jruby.runtime.Visibility.*;
@@ -145,7 +146,7 @@ public class RubyObjectSpace {
             final ArrayList<IRubyObject> modules = new ArrayList<>(96);
             runtime.eachModule((module) -> {
                     if (rubyClass.isInstance(module)) {
-                        if (!(module instanceof IncludedModule || module instanceof PrependedModule)) {
+                        if (!(module instanceof IncludedModule || module instanceof PrependedModule || module == runtime.getJavaSupport().getJavaPackageClass() || module instanceof JavaPackage)) {
                             // do nothing for included wrappers or singleton classes
                             modules.add(module); // store the module to avoid concurrent modification exceptions
                         }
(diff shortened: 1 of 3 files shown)


enebo:
This seems like a reasonable solution but in the back of my mind I figure we had to have thought of this in the past and didn't implement it.  Who knows.  Let's try it.

byteit101:
Somehow things are leaking past this: `NoMethodError: undefined method `instance_methods' for #<Class:Java::JavaIo>`

PR #7393, opened 2022-10-02
Fix PTY.spawn using libfixposix

### LINKED ISSUE #6552: PTY.spawn does not set new controlling tty for child
As part of work to get our FFI-based io/console green (https://github.com/ruby/io-console/pull/22) I discovered that our PTY module seems to break the subprocess io/console tests. This appears to be because our current PTY.spawn implementation is unable to properly set the new controlling tty, and as a result the raw IO operations in the child are routed back to the parent process.

This leads to several tests hanging on JRuby, since output they expect to read never arrives.

The PTY implementation originally came from @Freaky and has not had many changes since: https://github.com/jruby/jruby/issues/3185

The code in CRuby for this logic performs a number of operations in the child before proceeding to exec, and I assume these are the missing pieces needed to grant the child its own tty: https://github.com/ruby/ruby/blob/3108ad7bf3dcae52054a1c29b86246cdb470000b/ext/pty/pty.c#L102-L165

At this point, I do not know how to fix this, so it is a known bug in our PTY module.

One solution might be to finally start shipping a minimal executable that we can use to fork + exec and do these various operations in the child, as a sort of super-posix_spawn. No plans to work on that currently.

byteit101:
Fixes #6552

This, combined with the new gems and https://github.com/ruby/io-console/pull/35 should allow PTY.spawn to work on all *nix systems.

TO-DO list:

- [x] Come up with better names of all gems & the monorepo with the APIs (current location: https://github.com/byteit101/subspawn) New name: SubSpawn
- [x] Clean up mid & high level API's
- [x] Transition PTY.spawn to high-level API
- [x] Figure out how to build & bundle libfixposix (an older version is packaged, but not installed by default on most distros)
- [x] Implement envp in mid-level API
- [x] Implement signals in mid-level API
- [x] Document
- [ ] Test thoroughly on multiple people's machines

This can also be used to fix older versions of JRuby as the fix is gemified


byteit101:
I've been using the change in Ruby.java to test the end-to-end path as there aren't any pty.spawn tests here, but I'm not sure about leaving it or if it should be that method of configuration for now

headius (approved):
Looks good. Talked through default gem setup and windows warning and I think the rest is fine and we can iterate if necessary.

Issue #7397, opened 2022-10-06
Update bundler in jruby-complete


apurtell:
jruby-complete 9.3.8.0 embeds a version of bundler at ./META-INF/jruby.home/lib/ruby/gems/shared/gems/bundler-2.2.29/ that is subject to [CVE-2021-43809](https://nvd.nist.gov/vuln/detail/CVE-2021-43809).

> In `bundler` versions before 2.2.33, when working with untrusted and apparently harmless `Gemfile`'s, it is not expected that they lead to execution of external code, unless that's explicit in the ruby code inside the `Gemfile` itself. ...

PR #7385, opened 2022-09-30
Updating jruby-openssl to 0.14.0

coheigea:
A fix for https://github.com/jruby/jruby/issues/7335

https://github.com/jruby/jruby-openssl/releases/tag/v0.14.0

PR #7401, opened 2022-10-07
[deps] update Bundler/Rubygems to latest patch

kares:
closing https://github.com/jruby/jruby/issues/7397

### DIFF
--- a/lib/pom.rb
+++ b/lib/pom.rb
@@ -17,8 +17,8 @@ end
 default_gems = [
     # treat RGs update special:
     # - we do not want bin/update_rubygems or bin/gem overrides
-    ['rubygems-update', '3.2.29', { bin: false }],
-    ['bundler', '2.2.29'],
+    ['rubygems-update', '3.2.33', { bin: false }],
+    ['bundler', '2.2.33'],
     ['cmath', '1.0.0'],
     ['csv', '3.2.5'],
     ['e2mmap', '0.1.0'],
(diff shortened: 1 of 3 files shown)


kares:
seems stdlib grows considerably (few hundred kb) when latest Bundler/Rubygems gets included:
`size of lib/target/jruby-stdlib-9.3.9.0-SNAPSHOT.jar expected smaller then 13*1024*1024 but got 14359221`

kares:
looking into this, stdlib was right at the edge of collapse already (before this change which adds a few kb):
```
  File: jruby-stdlib-9.3.9.0-SNAPSHOT.jar
  Size: 14319811
```
... at 13,984k

kares:
size increase is mainly caused by updating [jruby-openssl to 0.14](https://github.com/jruby/jruby/pull/7385)
the gem (due upgraded BC jars in 0.14.0) has grown from **7.79MB to 8.58MB**

chadlwilson:
Be afraid about BouncyCastle `1.72` coming jruby-openssl's way at some point in the future then - the provider jar went up over **5MB** on top of `1.71` due to addition of post-quantum crypto algorithms :'(

enebo:
@chadlwilson yeah that is unfortunate but we mostly are concerned enough to have a file size check because in the past we have accidentally included files we shouldn't.  So once we know the expected size we will just increase our check number and not fret too much about it.

headius:
All seems good to me. We will prioritize moving away from BC in 2023 but we don't have the resources to do it right now.

PR #7402, opened 2022-10-10
Fix Dir.glob and Dir.[] to respects the order of {} expressions

k77ch7:
This commit also removes 4 WIP tags.


enebo:
Awesome!  This is probably our roughest code in the code base.

PR #7407, opened 2022-10-14
Better match CRuby zone negotiation

### LINKED ISSUE #7406: Time.at with fractional seconds and :in doubles timezone offset
**Environment Information**

Appears to be OS-independent, as it is reproducible on both:  

`jruby 9.3.8.0 (2.6.8) 2022-09-13 98d69c9461 OpenJDK 64-Bit Server VM 25.342-b07 on 1.8.0_342-b07 +jit [x86_64-OpenBSD]`
`jruby 9.3.0.0 (2.6.8) 2021-09-17 85c20e780f Java HotSpot(TM) 64-Bit Server VM 25.131-b11 on 1.8.0_131-b11 +jit [mswin32-x86_64]`

**Expected Behavior**

With the following code:

```ruby
require 'tzinfo'
v = Time.now
tz = TZInfo::Timezone.get('America/New_York')
p Time.at(v.to_i, :in => tz)
p Time.at(v.to_i, v.nsec, :nsec, :in => tz)
```

I get the following output with CRuby 2.6, which is the output I would expect from JRuby:

```
2022-10-13 09:12:47 -0700
2022-10-13 09:12:47 -0700
```

**Actual Behavior**

I get the following output with JRuby:

```
2022-10-13 09:10:12 -0700
2022-10-13 02:10:12 -1400
```

Note the doubled timezone offset when fractional seconds are provided.

I hit this when adding fractional second support to Sequel's named_timezones extension.

headius:
* Don't normalize to UTC before setting time zone in the single- arg Time.at with a zone kwarg. timeZoneLocal does that before setting the passed-in zone, and if no zone is provided we should leave it local.
* When normalizing to UTC in zoneLocalTime, allow DateTime to also adjust the time, so we get the real equivalent time in UTC. This pairs with the above fix to only apply the zone offset once, and only apply it in the timeZoneLocal call tree.

Fixes #7406

### DIFF
--- a/core/src/main/java/org/jruby/RubyTime.java
+++ b/core/src/main/java/org/jruby/RubyTime.java
@@ -1385,7 +1385,7 @@ public class RubyTime extends RubyObject {
             time = time.gmtime();
 
             if (!zone.isNil()) {
-                timeZoneLocal(context, zone, time);
+                time = timeZoneLocal(context, zone, time);
             }
 
             return time;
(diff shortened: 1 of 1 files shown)


headius:
Note that I had to force push a smaller fix for this so the description here is a little out of date.

PR #7409, opened 2022-10-14
Update mri.stdlib.index to execute 'rake test:mri:stdlib'.

k77ch7:
'uninitialized constant IO::Buffer (NameError)' occurs because of IO::Buffer is not implmented.


k77ch7:
'rake test:mri:stdlib' seems very slow or hangs.

Issue #3399, opened 2015-10-16
Can't seek when file accessed using classpath: OR uri:classloader: schemes


r6p:
Error case followed by success case below. This issue breaks Sinatra AssetPack running on JRuby 9.0.1.0 when the Sinatra "root" directory is set using the classpath: scheme.
````
$ mkdir /tmp/classpath_seek
$ cd !$
$ rvm use jruby-9.0.1.0
$ touch foo.rb
$ irb
>> $CLASSPATH <<  Dir.pwd
=> ["file:/Users/me/.m2/repository/jline/jline/2.11/jline-2.11.jar", "file:/Users/me/.rvm/rubies/jruby-9.0.1.0/lib/ruby/stdlib/readline.jar", "file:/private/tmp/classpath_seek/"]
>> f = File.open('classpath:foo.rb')
=> #<File:classpath:foo.rb>
>> f.seek(0)
Errno::EPIPE: Broken pipe - classpath:foo.rb
	from org/jruby/RubyIO.java:1649:in `seek'
	from (irb):4:in `<eval>'
	from org/jruby/RubyKernel.java:978:in `eval'
	from org/jruby/RubyKernel.java:1291:in `loop'
	from org/jruby/RubyKernel.java:1098:in `catch'
	from org/jruby/RubyKernel.java:1098:in `catch'
	from /Users/me/.rvm/rubies/jruby-9.0.1.0/bin/irb:13:in `<top>'

>> f = File.open(File.join(Dir.pwd, 'foo.rb'))
=> #<File:/private/tmp/classpath_seek/foo.rb>
>> f.seek(0)
=> 0
````

r6p:
The same issue is true for the uri:classloader scheme.

````
>>  f = File.open('uri:classloader:foo.rb')
=> #<File:uri:classloader:foo.rb>
>> f.seek(0)
Errno::EPIPE: Broken pipe - uri:classloader:foo.rb
	from org/jruby/RubyIO.java:1649:in `seek'
	from (irb):6:in `<eval>'
	from org/jruby/RubyKernel.java:978:in `eval'
	from org/jruby/RubyKernel.java:1291:in `loop'
	from org/jruby/RubyKernel.java:1098:in `catch'
	from org/jruby/RubyKernel.java:1098:in `catch'
	from /Users/me/.rvm/rubies/jruby-9.0.1.0/bin/irb:13:in `<top>'
````

r6p:
Same issue via this route (foo.rb in a jar):

[...]

PR #7415, opened 2022-10-17
Raise EPIPE for all unseekable channels

headius:
This relates to the discussion in discourse/mini_mime#38 where the proposed patch must work around JRuby's silent seek failure for some types of resources (at least classloader resources).

The patch here will now raise EPIPE for all NIO channels that are not seekable through some means. It is rather aggressive given the old logic that only raised if a channel was not seekable AND not selectable (for reasons long lost).

### DIFF
--- a/core/src/main/java/org/jruby/util/io/PosixShim.java
+++ b/core/src/main/java/org/jruby/util/io/PosixShim.java
@@ -72,20 +72,8 @@ public class PosixShim {
             return ret;
         }
 
-        if (fd.chSelect != null) {
-            // For other channel types, we can't get at a native descriptor to lseek, and we can't use FileChannel
-            // .position, so we have to treat them as unseekable and raise EPIPE
-            //
-            // TODO: It's perhaps just a coincidence that all the channels for
-            // which we should raise are instanceof SelectableChannel, since
-            // stdio is not...so this bothers me slightly. -CON
-            //
-            // Original change made in 66b024fedbb2ee32316ccd9de8387931d07993ec
-            setErrno(Errno.EPIPE);
-            return -1;
-        }
-
-        return 0;
+        setErrno(Errno.EPIPE);
+        return -1;
     }
 
     public int write(ChannelFD fd, byte[] bytes, int offset, int length, boolean nonblock) {


ikaronen-relex:
The silent failure was originally introduced as a workaround for https://github.com/jruby/jruby/issues/3399. I'm all for getting rid of it, but we should probably make sure there's no regression for that issue.

ikaronen-relex:
FWIW, I did a bit of source digging: it looks like the error in https://github.com/jruby/jruby/issues/3399 was triggered by [this line](https://github.com/rack/rack/blob/1-5-stable/lib/rack/file.rb#L110) in `Rack::File#each` back in Rack 1.5. The code has since been moved and refactored, but a similar seek seems to still happen on [this line](https://github.com/rack/rack/blob/0a62f75eeee8fbf14e87a92d458a4428382e3504/lib/rack/files.rb#L172) in `Rack::Files::BaseIterator#each_range_part`.

The "fix" in https://github.com/jruby/jruby/commit/a4cc5eeffdf26f2975077b30ed81bd144ac69360 made the seek fail silently, which is probably fine in the common case where there's only one range that starts at position 0, but it also probably made actual [HTTP range requests](https://developer.mozilla.org/en-US/docs/Web/HTTP/Range_requests) with a non-zero start offset silently return incorrect data when the underlying resource is non-seekable. Or, rather, it probably allowed such requests to *continue* breaking silently, since I doubt they ever worked correctly for non-seekable resources in the first place.

It would probably be worth filing a Rack issue about this, but in the mean time we also probably shouldn't break the current Rack code until it's been fixed (in all supported Rack versions). I wonder if it would be possible to detect no-op seeks that don't actually change the current file position (e.g. by maintaining our own file position counter) and only silently ignoring those. 🤔

headius:
@ikaronen-relex Thanks for that research! I agree we should not go forward with the breaking change for now, but I'm unsure how to proceed.

It seems logical that classloader resources should not be seekable, since the only way to access them is through an InputStream that does not support positioning. However, we don't want to break rack applications that might expect the silent seek failure as you discovered.

PR #7416, opened 2022-10-17
Regexp new none option

headius:
🤦


enebo:
@headius You accidentally got a second change in this PR.

headius:
Oops.

PR #7417, opened 2022-10-18
Update test.rake option

k77ch7:
When test:mri:core:int or test:mri:core:jit run, many NoMethodError occurs.
```
NoMethodError: undefined method `formatter' for DidYouMean:Module
```
This commit removes --disable-gem option from test.rake file because test/mir/ruby/test_pattern_match.rb requires DidYouMean module.

### DIFF
--- a/rakelib/test.rake
+++ b/rakelib/test.rake
@@ -100,7 +100,7 @@ namespace :test do
 
     mri_suites = [:core, :extra, :stdlib]
     mri_suites = {
-      core: "--disable-gems -Xbacktrace.style=mri -Xdebug.fullTrace",
+      core: "-Xbacktrace.style=mri -Xdebug.fullTrace",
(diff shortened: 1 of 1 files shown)


k77ch7:
Noisy errors seem to decrease as expected. 

### Before
305 errors https://github.com/jruby/jruby/actions/runs/3267926714/jobs/5373739709

### After(this patch)
259 errors  https://github.com/jruby/jruby/actions/runs/3273875788/jobs/5386661746

enebo:
@k77ch7 yeah I only commented about time to run the test suite.  If time gets a lot worse then we just move the pattern matching test into test:mri:extra where I think we have stuck other slower tests due to needing to execute ruby a lot.

enebo:
Looks like 24.5 minutes to about 26 minutes....maybe?  I only looked at two jobs to compare but that is not too big of a change.

PR #7418, opened 2022-10-20
Fix Enumerator.new without a block

k77ch7:
Enumerator.new without a block should raise ArgumentError. 

 * spec/ruby/core/enumerator/new_spec.rb
 * test_initialize in test/mri/ruby/test_enumerator.rb

This PR also update `spec/compiler/general_spec.rb` to adapt Enuemurator.new behavior.
And this PR uses enumeratorize or enumeratorizeWithSize instead of `runtime.getEnumerator().callMethod("new", ...)` in som Java classes to prevent spec:compiler and spec:ji failures(https://github.com/jruby/jruby/actions/runs/3290468184).


enebo:
@k77ch7 Not sure what happened but look at spec:compiler failures.

k77ch7:
@enebo Thank you. I'll check it later.

k77ch7:
@enebo I added some patched and updated this description.

PR #7419, opened 2022-10-20
Make all extname pass.  May break something in windows?

enebo:


enebo:
Actually fixed the extname error in test:jruby on windows...so yay.

PR #7424, opened 2022-11-06
Fix String#to_r and Marshal#load for Rational Class

k77ch7:
This PR fixes String#to_r for edges case
 e.g. '5__5/33'.to_r => 5/1, '5/3.3_'.to_r => 50/33
This PR also fixes Marshal#load for Rational Class

 * test_parse in test/mri/ruby/test_rational.rb
 * test_marshal in test/mri/ruby/test_rational.rb


k77ch7:
@enebo Thank you for review

PR #7442, opened 2022-11-10
Revert "Expand ported logic for encoding negotiation"

headius:
Reverts jruby/jruby#7081


headius:
This was a fluke and the failures were not my fault.

PR #7443, opened 2022-11-11
Update Integer#fdiv to match mri behavior

k77ch7:
When operands are Integer type, operands should be divided by gcd.
After applying this patch, the following test will pass.
 * test_supp in test/mri/ruby/test_rational.rb


headius:
Thank you!

PR #7446, opened 2022-11-11
Bundled and jarified native library loading for ffi

byteit101:
After inspecting other options in the java ecosystem, like maven-lib-loader and others, I realized that all of them are for JNI and thus not applicable to JRuby. Further, doing this in JNR/JFFI would require duplicating code both there and in JRuby, so this seems to be the best spot for now.

so/dylib/dll files are now loadable from embedded resources by being extracted to the jruby temp dir before being sent to JFFI

### DIFF
--- a/core/src/main/java/org/jruby/ext/ffi/jffi/DynamicLibrary.java
+++ b/core/src/main/java/org/jruby/ext/ffi/jffi/DynamicLibrary.java
@@ -6,7 +6,6 @@ import com.kenai.jffi.Library;
 import java.io.File;
 import java.io.IOException;
 import java.io.InputStream;
-import java.nio.file.CopyOption;
 import java.nio.file.Files;
 import java.nio.file.Path;
 import java.nio.file.StandardCopyOption;
@@ -14,7 +13,6 @@ import java.util.HashMap;
 
 import org.jruby.Ruby;
 import org.jruby.RubyClass;
-import org.jruby.RubyFile;
 import org.jruby.RubyModule;
 import org.jruby.RubyNumeric;
 import org.jruby.RubyObject;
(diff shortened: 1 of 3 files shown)


core/src/main/java/org/jruby/ext/ffi/jffi/DynamicLibrary.java
```diff
@@ -60,8 +72,25 @@ final Library getNativeLibrary() {
     @JRubyMethod(name = {  "open" }, meta = true)
     public static final  IRubyObject open(ThreadContext context, IRubyObject recv, IRubyObject libraryName, IRubyObject libraryFlags) {
         final String libName = libraryName.isNil() ? null : libraryName.toString();
+        String loadName;
+        if (libName.contains(":") && !new File(libName).isAbsolute()) { // windows is C:\\ ..., but otherwise we may have an URI
+            // Use internal logic to parse the URI
+            FileResource resource = JRubyFile.createResource(context, libName);
+            if (JRubyFile.isResourceRegularFile(resource))
+            {
```
enebo, line 80: '{' at end of line and not on next line.

core/src/main/java/org/jruby/ext/ffi/jffi/DynamicLibrary.java
```diff
@@ -60,8 +72,25 @@ final Library getNativeLibrary() {
...
+                loadName = resource.absolutePath();
+            }
+            else try (InputStream internalStream = resource.openInputStream()){
```
enebo, line 83: 'else' on same line as '}'

core/src/main/java/org/jruby/ext/ffi/jffi/DynamicLibrary.java
```diff
@@ -60,8 +72,25 @@ final Library getNativeLibrary() {
...
+                loadName = extractLibrary(resource.path(), internalStream);
+            } catch (IOException e) {
+                // let it fail normally, file not found
+                loadName = resource.absolutePath();
+            }
+        }
+        else
```
enebo, line 90: no single line if or else without curly brackets unless they are on same line (e.g. if (foo) something;).  This will never apply to if/else cases.  If both then they need curly brackets.

core/src/main/java/org/jruby/ext/ffi/jffi/DynamicLibrary.java
enebo, line 83: Is the path() here guaranteed to always contain '\.'?

headius (approved):
Looks fine for now other than review comments from @enebo that should be fixed. We can iterate on this and improve it after release.

enebo (approved):

PR #7447, opened 2022-11-12
Use thread-local for load wrapper

### LINKED ISSUE #7439: Kernel#load now accepts a module as the second argument,and will load the file using the given module as thetop-level module. [Feature #6210]

headius:
Plumbing the wrapper all the way through the interfaces involved is more of a change than we are comfortable with at this stage, so use a thread-local to set up the fea cases that currently honor the wrapper.

Ruby feature https://bugs.ruby-lang.org/issues/6210

Part of Ruby 3.1 features (#7015).

Fixes #7439


headius:
Doesn't quite pass the related test... WIP.

enebo (approved):

PR #7455, opened 2022-11-16
Fixes for M1 CI runs of 9.4

headius:


headius:
The item fixed here turned out to be an old bash on the M1 CI machine, so the fix is not needed. FFI specs are still failing but will be dealt with separately.

PR #7461, opened 2022-11-17
Use old pty.rb on Windows

headius:
Something in the subspawn pty replacement causes stderr to be closed during testing. See https://github.com/jruby/jruby/actions/runs/3485667896/jobs/5831405466

This is a temporary fix until we can determine why subspawn's pty logic does this.


headius:
Looks like this fixes the failed suite on Windows. @byteit101 will look into the issue in subspawn soon. https://github.com/byteit101/subspawn/issues/1

PR #7462, opened 2022-11-17
Update EPL version in jruby-jars Gemspec

gtback:
JRuby was updated from EPL-1.0 to EPL-2.0 in https://github.com/jruby/jruby/commit/6fe0e5a1a93b006ee41764e2b7c1d7fc72514a90, but this metadata was missed.

If I've misunderstood, and the licensing of jruby-jars differs from jruby itself, I'm sorry. The [jruby-jars README](https://github.com/jruby/jruby/commit/6fe0e5a1a93b006ee41764e2b7c1d7fc72514a90#diff-7dce96ed26a75ec1de92a0f8111704c57e1ca510fc6ee8451428286b7a673186L835) was updated at the same time to use EPL-2.0


headius:
Nope you're right. We just missed this one. Thanks!

Issue #7459, opened 2022-11-16
Unfinished Ruby 3.0 features


headius:
This is based on the list in #6878 and includes only the Ruby 3.0 features that were not implemented in time for the 9.4 release.

## Language changes

* [ ] #7457 Magic comment shareable_constant_value added to freeze constants.See {Magic Comments}[rdoc-ref:doc/syntax/comments.rdoc@Magic+Comments] for more details.
  [[Feature #17273](https://bugs.ruby-lang.org/issues/17273)] (Parses in JRuby but currently ignored due to not supporting Ractor)
* [ ] #7437 When a class variable is overtaken by the same definition in an ancestor class/module, a RuntimeError is now raised (previously, it only issued a warning in verbose mode). Additionally, accessing a class variable from the toplevel scope is now a RuntimeError.
  [[Bug #14541](https://bugs.ruby-lang.org/issues/14541)]

## Command line options
* [ ] `--help` option
   [[Feature #16754](https://bugs.ruby-lang.org/issues/16754)] (https://github.com/jruby/jruby/issues/7458)

## Core classes updates

ConditionVariable

* [ ]  ConditionVariable#wait may now invoke the `block`/`unblock` scheduler
  hooks in a non-blocking context. [[Feature #16786](https://bugs.ruby-lang.org/issues/16786)]

ENV

* [ ]  Windows: Read ENV names and values as UTF-8 encoded Strings
  [[Feature #12650](https://bugs.ruby-lang.org/issues/12650)]

Encoding

* [ ]  Changed default for Encoding.default_external to UTF-8 on Windows
  [[Feature #16604](https://bugs.ruby-lang.org/issues/16604)]

IO

* [ ]  IO#nonblock? now defaults to `true`. [[Feature #16786](https://bugs.ruby-lang.org/issues/16786)]
* [ ]  IO#wait_readable, IO#wait_writable, IO#read, IO#write and other
  related methods (e.g. IO#puts, IO#gets) may invoke the scheduler hook
  `#io_wait(io, events, timeout)` in a non-blocking execution context.
  [[Feature #16786](https://bugs.ruby-lang.org/issues/16786)]

Kernel

* [ ]  Kernel.sleep invokes the scheduler hook `#kernel_sleep(...)` in a
  non-blocking execution context. [[Feature #16786](https://bugs.ruby-lang.org/issues/16786)]

Queue / SizedQueue

* [ ]  Queue#pop, SizedQueue#push and related methods may now invoke the
[...]

PR #7466, opened 2022-11-21
Default to nonblocking whenever possible

headius:
Ruby 3.0 began to set O_NONBLOCK by default for all IO, to allow more IO types to enlist in the new IO/fiber scheduling. JRuby has not previously defaulted to nonblock, but all IO operations against nonblock-compatible channels does use nonblocking semantics, so all we really need to align with CRuby is to just set them non-blocking at the NIO level on creation.

We can only do this when the underlying channel supports nonblocking semantics, but this commit attempts to set such channels up properly.

Part of scheduler work in #7459 and related issues.

### DIFF
--- a/core/src/main/java/org/jruby/RubyIO.java
+++ b/core/src/main/java/org/jruby/RubyIO.java
@@ -905,6 +905,15 @@ public class RubyIO extends RubyObject implements IOEncodable, Closeable, Flusha
                     }
                 }
             }
+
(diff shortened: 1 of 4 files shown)


headius:
This is less of a concern given our native threads but ideally we still should do it. Need to reexamine this for 9.4.2.

headius:
Do this whenever #7944 lands.

PR #7475, opened 2022-11-23
Update CGI to 0.1.0.2 for CVE-2021-33621

headius:
https://www.ruby-lang.org/en/news/2022/11/22/http-response-splitting-in-cgi-cve-2021-33621/

### DIFF
--- a/lib/ruby/stdlib/cgi.rb
+++ b/lib/ruby/stdlib/cgi.rb
@@ -253,7 +253,7 @@
 #           end
 #         end +
 #         cgi.pre do
-#           CGI::escapeHTML(
+#           CGI.escapeHTML(
 #             "params: #{cgi.params.inspect}\n" +
 #             "cookies: #{cgi.cookies.inspect}\n" +
 #             ENV.collect do |key, value|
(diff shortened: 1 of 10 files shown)


headius:
Note that this is just copied from the 0.1.0.2 tag on https://github.com/ruby/cgi, since there's no 0.1 branch or 0.1.0 branch and neither CRuby 2.7 nor 2.6 appear to have been patched yet.

headius:
Link https://github.com/ruby/cgi/issues/31 for clarity on where these old versions are being maintained.

headius:
It appears the v0.1.0.2 tag on the ruby/cgi repo does point at the correct, patched version. It does not exist in any branch, which we will deal with in ruby/cgi#31.

I have confirmed that the copied sources pass the new CVE-related tests also on the v0.1.0.2 branch. Puma appears to cause cgi.rb to be loaded, but does not appear to use the patched code.

headius:
The initial spec failures turned out to be bad cookie specs using bogus key names, which were rejected by the new stricter cookie logic. Those were patched for CRuby only in https://github.com/ruby/ruby/commit/36f297e62108072b9377d927321928b994f66a93 and are not yet incorporated into ruby/spec.

I have added those changes directly to this PR, since 9.3 does not get ruby/spec updates anymore (ruby/spec only maintains specs for supported CRuby versions).

@eregon Unfortunately I accidentally applied this spec patch to JRuby master, so hopefully that doesn't cause trouble when merging back and forth next time.

eregon:
It's not a problem to commit the diff in jruby too, as long as it's the same diff, that commit will just be skipped during the rebase to sync the specs.

Issue #7486, opened 2022-11-26
Regression in 9.4 (optimizer? JIT?)


byteit101:
**Environment Information**
  
`jruby 9.4.0.0 (3.1.0) 2022-11-23 95c0ec159f Java HotSpot(TM) 64-Bit Server VM 25.201-b09 on 1.8.0_201-b09 +jit [x86_64-linux]`

via RVM

**Actual Behavior**
```
jruby-9.4.0.0 :002 > str = "123456789" * 100 # 10 isn't enough
 =>  "..."
jruby-9.4.0.0 :003 > str.each_char.each_slice(2){|a, b| (a.ord ^ b.ord)}
org.jruby.ir.IRScope.cloneInstrs(IRScope.java:588): Java::JavaLang::NullPointerException
        from org.jruby.ir.IRScope.prepareForCompilation(IRScope.java:655)                                                       
        from org.jruby.runtime.MixedModeIRBlockBody.promoteToFullBuild(MixedModeIRBlockBody.java:159)
        from org.jruby.runtime.MixedModeIRBlockBody.commonYieldPath(MixedModeIRBlockBody.java:142)   
        from org.jruby.runtime.IRBlockBody.doYield(IRBlockBody.java:170)                             
        from org.jruby.runtime.BlockBody.yield(BlockBody.java:108)                                   
        from org.jruby.runtime.Block.yield(Block.java:188)                                           
        from org.jruby.RubyEnumerable.lambda$each_sliceCommon$5(RubyEnumerable.java:1262)            
        from org.jruby.runtime.BlockCallback.call(BlockCallback.java:54)                             
        from org.jruby.runtime.CallBlock.doYield(CallBlock.java:102)                                 
        from org.jruby.runtime.BlockBody.yield(BlockBody.java:116)                                   
        from org.jruby.runtime.Block.yieldValues(Block.java:234)                                     
        from org.jruby.RubyEnumerator$1.call(RubyEnumerator.java:406)                                
        from org.jruby.runtime.BlockCallback.call(BlockCallback.java:58)                             
        from org.jruby.runtime.CallBlock.doYield(CallBlock.java:97)                                  
	from org.jruby.runtime.BlockBody.yield(BlockBody.java:108)
	from org.jruby.runtime.Block.yield(Block.java:188)
	... 183 levels...

```

PR #7493, opened 2022-11-28
one-time-execute can sometimes contain a closure.  guard it.

enebo:
in irb we get a scenarion where we delete parent scope of a closure because at the time we delete it we do not know a closure will exist. The fix is a trivial bullet-proof test line.  Our actual code calling prepareForCompilation already is aware that null fic is a failure case.


enebo:
Note fixes #7486

enebo:
for whatever reason jit for java 11 is flaking out but all this PR could do is not cause JRuby to crash so it has to be unrelated (unless these tests never really run at all and crashing somehow is showing it as a pass).

PR #7503, opened 2022-12-01
[refactor] make Java embed utils generic

kares:
~~the main thing is to fix an issue with `java_alias` that did not work with interfaces (included test errors with a `ClassCastException` using JRuby 9.3.9.0).~~

ended up in a code base that uses `JavaEmbedUtils`'s `invokeMethod` and `rubyToJava`, and require and extra cast which seemed like a missed opportunity :wink:


kares:
hmm, wonder if I should target 9.3 branch, the iface fix would be nice to have there but still isn't super crucial ...

headius:
JRuby 9.3 will be maintained for the next year, so if you think this is valuable to have there we should retarget.

kares:
yeah, already did cherry-pick the fix commit for 9.3: https://github.com/jruby/jruby/pull/7514 (for the refactorings I have no strong opinions)

kares:
update the PR, since https://github.com/jruby/jruby/pull/7514 got merged over, now it's mostly just `JavaEmbedUtils` generic-ification

PR #7514, opened 2022-12-06
[fix] java_alias-ing for interface methods

kares:
previously doing a MyInterface.java_alias just errored:

`ClassCastException: class org.jruby.RubyModule cannot be cast to class org.jruby.RubyClass`

---

*extracted from https://github.com/jruby/jruby/pull/7503 after I bumped into this twice in 9.3.9.0*


headius:
Looks good. Trivial fix.

PR #7516, opened 2022-12-08
Improvements to instance variable shaping

headius:
This PR will incorporate a few improvements to how we "right-size" user-defined objects to pack instance variables.

* Better parent-driven layout of fields, so they can be shared across subclasses.
* Persistent shapes for stale objects; after we realize it needs to grow, existing objects can continue living with the smaller shape.
* Improved caching to avoid polymorphic thrashing when the fields are laid out the same.
* Analysis of implementation and exploration of other improvements.

Stretch goals might include playing with field-doubling to store numeric values without wrappers, but the primary goal is to do a better job of choosing the right shape or evolving into the right shape over time.


headius:
So far this is safe to do in 9.4, but we'll see how it goes.

headius:
Rebased to 9.5-dev.

PR #7523, opened 2022-12-14
Define IO::READABLE, WRITABLE, PRIORITY for IO#wait

headius:
These are to be used by the io-wait library, but there will be some lag before I can figure out how to make them honor these values without breaking backward compatibility with 9.4.0.0.


headius:
Do this whenever https://github.com/jruby/jruby/pull/7944 lands.

headius:
The missing magic was to also move the io-wait methods `wait`, `wait_readable`, and `wait_writable` into IO proper, since they are moving out of the io-wait gem in CRuby.

PR #7524, opened 2022-12-14
Add method! and public_send! for OpenStruct warnings

headius:
Due to these methods now accessing the caller's scope (for refinements) they now warn when being aliased to another name. We suppress the warning when the target name has a superset of client frame reads, so this patch adds `method!` and `public_send!` with SCOPE read.

This is not ideal, since any future method gaining frame fields will need to have the same treatment, and other similar methods are being skipped by OpenStruct (by our request). However this suppresses the warning for now and we can work on a long-term option.


headius:
Still not sure I like this solution, so bumping.

headius:
These warnings are now only displayed in verbose mode, so we can reevaluate whether this is still necessary in 9.5.

Issue #7521, opened 2022-12-12
FFI errno spec fails on JRuby on Windows


eregon:
See https://github.com/ffi/ffi/actions/runs/3673763975/jobs/6211192575
```
 Failures:

  1) FFI.errno FFI.errno contains errno from last function, FFI::LastError.winapi_error works differently per OS
     Failure/Error: expect(FFI::LastError.winapi_error).to eq(0x12345678)

       expected: 305419896
            got: 0

       (compared using ==)
     # ./spec/ffi/errno_spec.rb:28:in `block in <main>'
```

I'll skip it upstream (until fixed) to skip the FFI CI green.

larskanis:
I opened an issue on the ffi side as well: https://github.com/ffi/ffi/issues/984

PR #7527, opened 2022-12-15
alias winapi_error to lasterror

ahorek:
the current implementation isn't thread-safe, so the current errno context could be lost

errno is already implemented via LastError, so let's just alias it https://github.com/jnr/jffi/blob/master/jni/jffi/LastError.c#L72

fixes https://github.com/jruby/jruby/issues/7521


larskanis:
Looks good!

Issue #7529, opened 2022-12-16
Enumerator#inspect produces an ASCII-8BIT encoded String


cboos:
**Actual Behavior**

The `Enumerator#inspect` builds an ASCII-8 bit string, as opposed to what it does with other values:

```
# irb
irb(main):001:0> "Ä".inspect.encoding
=> #<Encoding:UTF-8>
irb(main):002:0> ["Ä"].inspect.encoding
=> #<Encoding:UTF-8>
irb(main):003:0> ["Ä"].each.inspect.encoding
=> #<Encoding:ASCII-8BIT>
```

This oddity in JRuby is quite inconvenient since it can easily cause an Encoding incompatibility when mixing this with other, UTF-8 encoded, strings (in custom #inspect methods, for instance):

```
irb(main):007:0> "#{["Ä"].each.inspect} / #{["Ä"].inspect}"
(irb):7:in `evaluate': incompatible encodings: ASCII-8BIT and UTF-8 (Encoding::CompatibilityError)
        from org/jruby/RubyKernel.java:1143:in `eval'
        from org/jruby/RubyKernel.java:1586:in `loop'
        from org/jruby/RubyKernel.java:1292:in `catch'
        from org/jruby/RubyKernel.java:1292:in `catch'
        from /opt/jruby/lib/ruby/gems/shared/gems/irb-1.4.1/exe/irb:11:in `<main>'
        from org/jruby/RubyKernel.java:1091:in `load'
        from /opt/jruby/bin/irb:23:in `<main>'
```

**Expected Behavior**

The above contrasts with CRuby (Linux or Windows, any version between 2.7 and 3.2):

```
irb(main):001:0> ["Ä"].each.inspect.encoding
=> #<Encoding:UTF-8>
irb(main):002:0> "#{["Ä"].each.inspect} / #{["Ä"].inspect}"
=> "#<Enumerator: [\"Ä\"]:each> / [\"Ä\"]"
```

**Environment Information**
  
* JRuby version
```
root@8c5e900b8cf0:/opt# jruby -v
jruby 9.4.0.0 (3.1.0) 2022-11-23 95c0ec159f OpenJDK 64-Bit Server VM 25.352-b08 on 1.8.0_352-b08 +jit [x86_64-linux]
```

(from the official docker image jruby:9.4)

No JRUBY_OPTS, .jrubyrc, etc.

* Operating system and platform

```
# uname -a
[...]

PR #7532, opened 2022-12-17
Enumerator inspect tests

cboos:
I wanted to contribute a regression test for issue #7529, as a follow-up to its fix by @enebo in #7531, but I noticed that there was *already* a corresponding test case in the MRI test suite.

Before the fix, with master 278aaac1cf29da19e245a3903aedbac6d6e34334:

```
$ jruby test/mri/runner.rb  test/mri/ruby/test_enumerator.rb
...
[48/63] TestEnumerator#test_inspect_encoding = 0.02 s
  7) Failure:
TestEnumerator#test_inspect_encoding [/home/cboos/workspace/jruby/test/mri/ruby/test_enumerator.rb:440]:
<#<Encoding:UTF-8>> expected but was
<#<Encoding:ASCII-8BIT>>.

Finished tests in 2.532488s, 24.8767 tests/s, 171.7678 assertions/s.
63 tests, 435 assertions, 6 failures, 1 errors, 0 skips

```

However, after the fix, with 1d9723771bb93d53acc4c6480e0097d65ba5154b, that test still failed!

With JRuby we got: `"#<Enumerator: #<#<Class:0x43312512>:0x1b868ef0>:?>"` instead of the expected `"#<Enumerator: #<#<Class:0x000002c3ad457ec8>:0x000002c3ad5dacc8>:あ>"` as with CRuby 3.0, for example. 

Reproducing with irb what the `TestEnumerator#test_inspect_encoding` test does gives the following:

```ruby
irb(main):001:0> c = Class.new{define_method("\u{3042}"){}}
=> #<Class:0x79cfc008>
irb(main):002:0> c.new.enum_for("\u{3042}")
=> #<Enumerator: #<#<Class:0x79cfc008>:0x54ebd3b4>:?>
c.new.enum_for("\u{3042}").instance_eval("@__method__")
=> :B
```

Even worse, trying to use that enumerator leads to:

```
irb(main):003:0> e.next
uri:classloader:/jruby/kernel/enumerator.rb:72:in `block in initialize': undefined method `B' for #<#<Class:0x4834e816>:0x3389c69d> (NoMethodError)
```

This pull request adds some follow-up tests and fixes for the above.

### DIFF
--- a/core/src/main/java/org/jruby/RubyEnumerator.java
+++ b/core/src/main/java/org/jruby/RubyEnumerator.java
@@ -450,7 +450,7 @@ public class RubyEnumerator extends RubyObject implements java.util.Iterator<Obj
         } else {
             result.cat19(RubyObject.inspect(context, object));
             result.cat(':');
-            result.catString(method);
+            result.cat19(getMethod().asString());
             if (methodArgs.length > 0) {
                 result.cat('(');
                 for (int i= 0; i < methodArgs.length; i++) {
@@ -617,6 +617,10 @@ public class RubyEnumerator extends RubyObject implements java.util.Iterator<Obj
 
     private IRubyObject getGenerator() {
         return getInstanceVariable(GENERATOR);
+    }
+
+    private IRubyObject getMethod() {
+        return getInstanceVariable(METHOD);
     }
 
     @Override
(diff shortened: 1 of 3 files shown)


enebo:
@cboos we have very slowly been finding places where we are not appropriately encoding m17n symbols into Strings.  Mostly in error strings but you found out we also have issues with some inspect()/to_s() (in reflective sorts of things).  Most people do not consume these things in a progmatic way so I think that partially explains why they have not been fixed.

This is great though.  Thanks for your help!

PR #7534, opened 2022-12-21
[feat] allow enabling warnings wout changing $VERBOSE

kares:
previously there was no way of "just enabling warnings", warnings were enabled with `setVerbose(IRubyObject verbose)`

to be able to consume warnings but not change `$VERBOSE` I've added a `setWarningsEnabled`.

also for Java consumers it's easier to deal with Java values thus `setVerbose(Boolean verbose)` overload is added ...

### DIFF
--- a/core/src/main/java/org/jruby/Ruby.java
+++ b/core/src/main/java/org/jruby/Ruby.java
@@ -2531,13 +2531,33 @@ public final class Ruby implements Constantizable {
         return warningsEnabled;
     }
 
-    /** Setter for property isVerbose.
-     * @param verbose New value of property isVerbose.
+    /**
+     * Setter that enables/disabled warnings (without changing verbose mode).
+     * @param warningsEnabled
+     * @see #setVerbose(IRubyObject)
      */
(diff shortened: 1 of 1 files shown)


headius:
This seems fine, but I wonder how much effect it will have after code has loaded and cached the current `$VERBOSE` value. We don't do much caching of global variables right now, but we probably will in the coming months. Also, there's a lot of code out there that makes verbosity decisions early on and never looks at this variable again.

kares:
> This seems fine, but I wonder how much effect it will have after code has loaded and cached the current `$VERBOSE` value.

The actual use-case, this came from, is actually setting up the runtime early. 
But yeah it's a valid point, given there's already a few places (within JRuby itself) doing a `setVerbose`.
Also while there's the `RubyInstanceConfig` the verbosity level cat only be controlled on the runtime, atm.

headius:
@kares Yeah that makes sense. I think it's a good change. If someone gets confused we can just explain that these values are frequently cached during post-boot startup by some libraries.

PR #7549, opened 2022-12-30
[ji] support converting RubyThread to a java.lang.Thread

kares:
stumbled upon `thread.to_java('org.jruby.RubyThread').native_thread` at least in 2 separate code-bases

believe 9.4 is a good opportunity to formalize `RubyThread#to_java` support to return a Java thread by default


headius:
Seems reasonable to me!

PR #7557, opened 2023-01-06
[test] un-pend failing test after Thread#to_java update

kares:
follow-up on be88ae16c7bd857c72d30527365dc998641068e9

Thread#to_java changes are from GH-7549 (bfa145196fa203efbffe)


enebo:
@kares  Is it possible to add a native_thread method to j.l.Thread?  I think it is possible someone will be doing that out in the wild and we were definitely depending on that method in the test.

headius:
I think we are all on board with this.

PR #7566, opened 2023-01-08
Fix Float#floor and Float#round for some edge cases

k77ch7:
This change is porting from mri(numeric.c and rational.c).

 * test_floor_ceil_ndigits in test/mri/ruby/test_numeric.rb
 * test_float_round_ndigits in test/mri/ruby/test_numeric.rb


enebo:
nice!  Love seeing rounding fixes.

PR #7567, opened 2023-01-08
Fix missing require in mri-core tests

jcharaoui:
Its absence causes an intermittent failure, likely caused by load ordering randomness.

```
TestEnumerable#test_sum:
NameError: uninitialized constant TestEnumerable::SimpleDelegator
    org/jruby/RubyModule.java:3947:in `const_missing'
        /tmp/autopkgtest-lxc.wadx2atf/downtmp/autopkgtest_tmp/test/mri/ruby/test_enum.rb:1051:in `test_sum'
```


headius:
Not surprising. We should probably fix this upstream in the MRI tests, too, since we sync from their repo.

PR #7601, opened 2023-01-26
Fixed launch jruby in MSYS2 environment

knovok:
I'm using MSYS2 environment.
jruby.exe is launched not from current directory but globally in system. Even if you run it from current directory, there will be no effect because jruby.exe does nothing but returns code 0.
In PR #37 was added launch jruby.exe of MSYS, but this is the first version of MSYS not MSYS2.

### DIFF
--- a/bin/jruby.sh
+++ b/bin/jruby.sh
@@ -120,8 +120,14 @@ cygwin=false
 case "$(uname)" in
     CYGWIN*) cygwin=true ;;
     MINGW*)
-        jruby.exe "$@"
-        exit $?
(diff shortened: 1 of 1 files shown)


headius:
Thank you! I have no way to evaluate this right now, but it appears to be low risk to anything other than cygwin/mingw environments, which @knovok is clearly using. I assume this patch still works on msys?

knovok:
I tried not to hook other environments other than MSYS2. Cygwin should not break, as well as MSYS. The fix checks the environment ID from the /etc/os-release file. Even if the file does not exist (which should not be https://0pointer.de/blog/projects/os-release.html), the implementation will run without the patch.

headius:
Ok we'll go for it! Thanks!

Issue #7604, opened 2023-01-31
Why is older fastercsv slower than newer fastercsv


enebo:
The csv in JRuby 1.7 is quite a bit faster than the one shipped in 9.3/9.4:

```text
system ~/work/jruby master 2052% jruby  -I../snippets ~/work/snippets/csv_bench.rb 
2.4891145080327988
2.0067069232463837
1.8442632555961609
1.718497270718217
1.716303520835936
1.8347881203517318
1.7011602241545916
1.701060843653977
1.7657083729282022
1.8220716370269656
2.243274923413992
```

With current csv:

```text
jruby  -I../snippets ~/work/snippets/csv_bench.rb 
4.355965277180076
3.160095911473036
2.868028214201331
2.769463988021016
3.0816614385694265
3.970997073687613
3.5069410810247064
3.3267764318734407
3.9052677769213915
3.463081202469766
3.283401306718588
3.3949629701673985
```

So almost 2x slower.  MRI 3.1 is maybe 10% slower with the newer versions. Benchmark:

```ruby
require 'csv'
#require 'old_csv'
require 'benchmark'

dir = "data"
Dir.mkdir(dir) unless Dir.exist?(dir)

input = "#{dir}/extract.csv"

unless File.exist?(input)
  Dir.chdir(dir) do
    zip = "sirene_2017111_E_Q.zip"
    system "wget", "-O", zip, "http://files.data.gouv.fr/sirene/#{zip}"
    system "unzip", zip
    system "head -10000 sirc* > extract.csv"
  end
end

30.times do
  p Benchmark.realtime {
    CSV.foreach(input, col_sep: ';', encoding: "ISO-8859-1:UTF-8") { |row|
    }
  }
end
```

headius:
A big difference between 1.7 and every JRuby version that followed is support for string encodings, which under some circumstances can be much more overhead than the old encodingless strings.

I will take a poke at this and see if anything obvious jumps out.

headius:
Indeed, when profiling the top items I see are mostly related to character transcoding.

Modifying your script from

```ruby
CSV.foreach(input, col_sep: ';', encoding: "ISO-8859-1:UTF-8")
```

to

```ruby
CSV.foreach(input, col_sep: ';', encoding: "ISO-8859-1")
```

Nearly doubles the performance on JRuby 9.4.1 and exceeds the performance of JRuby 1.7.

* JRuby 1.7 with "UTF-8": 0.93s
[...]

PR #7619, opened 2023-02-04
Small internal optimizations found profiling CSV

headius:
See #7604


core/src/main/java/org/jruby/util/io/OpenFile.java
```diff
@@ -154,6 +154,9 @@ public static class Buffer {
 
     protected volatile Set<RubyThread> blockingThreads;
 
+    private Ptr spPtr = new Ptr();
+    private Ptr dpPtr = new Ptr();
```
kares, line 158:
```suggestion
    private final Ptr spPtr = new Ptr();
    private final Ptr dpPtr = new Ptr();
```
... not sure how much (whether) it matters with C2 JIT these days

PR #7622, opened 2023-02-06
Investigating test_rubyoptions hang

headius:


headius:
It's definitely the test_rubyoptions hanging but not sure where. Reverting #7614 for now in #7624 and we'll revisit this after 9.4.1.0.

PR #7623, opened 2023-02-06
setup and fix issues reported by error-prone compiler

kares:
set up https://errorprone.info/ which provides additional checks and indeed discovered a bug or two in the code base.

the compiler plugin is currently hidden behind `-Perror-prone` as it slows down Java compilation considerably...
the profile is meant to activate on CI (with `ENV['CI']`).
also, error-prone plugin needs Java 11 (getting a Java 8 compatible version hooked up turned out to be quite difficult).

there are still warnings emitted but the blocking and most relevant stuff should be cleaned up here ...

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -30,7 +30,7 @@ jobs:
           java-version: ${{ matrix.java-version }}
           cache: 'maven'
       - name: bootstrap
-        run: mvn -Pbootstrap clean package
+        run: ./mvnw -Pbootstrap clean package
       - name: bundle install
         run: bin/jruby --dev -S bundle install
       - name: Save bootstrap build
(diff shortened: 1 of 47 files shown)


Pixisamess (approved):

Pixisamess (approved):

kares:
:red_circle: not sure why I get constant failures with the `bundle install` step:
```
Run bin/jruby --dev -S bundle install
Gem::Requirement::BadRequirementError: Illformed requirement [">= 0"]
```
... seems unrelated and I have a hunch it only reproduces on the kares/jruby fork (tried cleaning up cache)

headius (approved):
The reference equality warnings are a little annoying, but all the changes here seem like good ones (albeit a little pedantic at times).

The CollectionIncompatibleType warning is an interesting case... I'm not sure what the problem is but I'll leave it to you to investigate for now.

PR #7629, opened 2023-02-06
Split opts

enebo:


enebo:
@headius @lopex I am mostly looking for holes in RubyRegexp.isSimpleString.  The rest I think is just having more hotpaths and reorganizing to make selecting those hot paths easier to select.

The actual simple string logic should be very conservative in that lots of valid strings containing those chars are being rejected.  The two fundamental questions are:

  1. In single char strings is there something more than ```.```, ```^```, or ```$``` that should be rejected?
  2. in isExact am I missing a special character?

headius (approved):
The changes look fine, assuming discussions today about detecting simple regex have been considered or will be considered.

eregon:
Is there code out there using `String#split` with a single-char Regexp? Why don't they pass the String instead?
I could see it for `/ /` since `" "` is special (awk-style split), but for other values it doesn't seem of any utility vs just using the String. Or maybe I'm missing something?

enebo:
@eregon You may be missing that when people do things like this I am not there to tell them they could do it better :smiley: 

In the case I noticed this was someone submitted an issue where the regexp was not the specific issue but it still had:
```ruby
      arr << s.split(/,/).join('+')
```

I have seen people do stuff like ```/word/ ```in other cases as well.  It is uncommon and ill-advised but why not?

enebo:
@eregon Actually you just put me down a philosophical hole with this question.  Why should a valid regular expression which is just a simple string be expected to be slower than just using the string?  I admit I had the same reaction as you originally but part of me figured since it is correctable why not do it (note: I also did a hot path on ASCII string patterns at the same time so those also sped up)?

PR #7633, opened 2023-02-08
Do not return early from ThreadService#terminate

justinstoller:
Previously, if we encounter an adopted thread when terminating the ThreadService we return early and fail to finish killing threads we spawned.

This changes the `return` to `continue` which I believe was always the intent.

### DIFF
--- a/core/src/main/java/org/jruby/internal/runtime/ThreadService.java
+++ b/core/src/main/java/org/jruby/internal/runtime/ThreadService.java
@@ -151,7 +151,7 @@ public class ThreadService extends ThreadLocal<SoftReference<ThreadContext>> {
             // don't kill current thread that is doing teardown
             if (rth == getCurrentContext().getThread()) continue;
 
-            if (rth.isAdopted()) return;
+            if (rth.isAdopted()) continue;
 
             try {
                 rth.kill();


justinstoller:
Originally, I thought this was a continued problem with the Timeout module but I think it simply has the privileged distinction of being the only library thread we use that doesn't clean up after itself before termination.

kares:
Hey Justin, this small change indeed makes sense.
Did you consider targeting the [jruby-9.3](https://github.com/jruby/jruby/tree/jruby-9.3) branch, so it gets to next 9.3 (as well as 9.4) ?

kares:
> Also, is there an ETA on a version of 9.4 with this fix?

Since, 9.4.1 was just released 9.4.2 (and 9.3.11) might be several weeks away.
There were more than 2 months between 9.4.0 and 9.4.1, but with holidays period, so ~8 weeks might be a good guess.
Guess it depends on the severity of user reports coming in, which might move the release to a sooner date...

Would a nightly build be usable for you?

headius:
It won't be 8 weeks, but it will be more than two. Depends on how critical and how many issues we get in the next week or two.

Looking at this now.

headius:
Yeah this is the right fix. Old code, but I'm surprised this hasn't been reported previously.

I will merge this into 9.3 and forward to master and it will be in snapshots soon and releases a little less soon.

justinstoller:
Thanks! No need on the nightly build.

FWIW, we certainly have a fairly unique use case with Puppet Server. We also are still carrying a [work around](https://github.com/puppetlabs/puppetserver/commit/c3760842a350a2ab6fe8d4e6f0b041e6032e3d79) for #6127.

PR #7646, opened 2023-02-09
Add Integer#ceildiv

edipofederle:
Ticket: https://bugs.ruby-lang.org/issues/18809

I also push a PR for ruby spec: https://github.com/ruby/spec/pull/1013


edipofederle:
@headius can you take a look at this one? Thanks

headius:
Looks good!

PR #7651, opened 2023-02-11
Fix Array#fill for edge case

### LINKED ISSUE #7648: Array#fill not compliant with MRI
**Expected Behavior**
```
2.6.4 :001 > [:foo].fill(nil,2,0)
 => [:foo, nil]
```
```
3.1.3 :001 > [:foo].fill(nil,2,0)
 => [:foo, nil]
```

**Actual Behavior**
```
jruby-9.3.9.0 :001 > [:foo].fill(nil,2,0)
 => [:foo]
```
```
jruby-9.4.1.0 :006 > [:foo].fill(nil,2,0)
 => [:foo]
```

k77ch7:
If start >= array.size and length = 0, array#fill should extends self with nil.

Fixes #7648

### DIFF
--- a/core/src/main/java/org/jruby/RubyArray.java
+++ b/core/src/main/java/org/jruby/RubyArray.java
@@ -2350,8 +2350,7 @@ public class RubyArray<T extends IRubyObject> extends RubyObject implements List
         unpack();
         modify();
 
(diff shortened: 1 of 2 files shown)


headius:
What a weird condition! Looks good, thanks!

PR #7652, opened 2023-02-11
Fix Float#to_d of bigdecimal/util

### LINKED ISSUE #7650: BigDecimal arithmetic regression [jruby-9.4]
**Expected Behavior**

```
2.6.4 :001 > require 'bigdecimal/util'
 => true
2.6.4 :002 > 0.1.to_d
 => 0.1e0
 ```
```
3.1.3 :001 > require 'bigdecimal/util'
 => true
3.1.3 :002 > 0.1.to_d
 => 0.1e0
 ```
 ```
jruby-9.3.10.0 :001 > require 'bigdecimal/util'
 => true
jruby-9.3.10.0 :002 > 0.1.to_d
 => 0.1e0
```

**Actual Behavior**

```
jruby-9.4.1.0 :002 > require 'bigdecimal/util'
 => true
jruby-9.4.1.0 :003 > 0.1.to_d
 => 0.1000000000000000055511151231257827021181583404541015625e0
```

I understand that this is what you get with floating point arithmetic and that jruby-9.4 might be even more precise. But did this have to change in jruby-9.4? It leads to lower precision with basic calculations like `0.1.to_d - 0.1` which is not easy to understand for many users.

### PR DESCRIPTION
**Pull request by k77ch7**
This change uses BigDecimal(string) instead of BigDecimal(double) to generate correct BigDecimal value and to avoid precision error.

Fixes #7650

### DIFF
--- a/core/src/main/java/org/jruby/ext/bigdecimal/RubyBigDecimal.java
+++ b/core/src/main/java/org/jruby/ext/bigdecimal/RubyBigDecimal.java
@@ -545,7 +545,7 @@ public class RubyBigDecimal extends RubyNumeric {
         RubyBigDecimal res = newFloatSpecialCases(runtime, arg);
         if (res != null) return res;
 
-        return new RubyBigDecimal(runtime, (RubyClass) recv, new BigDecimal(arg.getDoubleValue(), mathContext));
+        return new RubyBigDecimal(runtime, (RubyClass) recv, new BigDecimal(arg.inspect().asJavaString(), mathContext));
     }
 
     private static RubyBigDecimal newFloatSpecialCases(Ruby runtime, RubyFloat val) {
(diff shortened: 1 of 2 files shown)

### CODE REVIEW

File: core/src/main/java/org/jruby/ext/bigdecimal/RubyBigDecimal.java [hunk 1]
```diff
@@ -545,7 +545,7 @@ private static RubyBigDecimal newInstance(Ruby runtime, IRubyObject recv, RubyFl
         RubyBigDecimal res = newFloatSpecialCases(runtime, arg);
         if (res != null) return res;
 
-        return new RubyBigDecimal(runtime, (RubyClass) recv, new BigDecimal(arg.getDoubleValue(), mathContext));
+        return new RubyBigDecimal(runtime, (RubyClass) recv, new BigDecimal(arg.inspect().asJavaString(), mathContext));
```
headius, line 548: I believe this can just be `arg.toString()`, avoiding the intermediate RubyString object.

headius (changes requested):

core/src/main/java/org/jruby/ext/bigdecimal/RubyBigDecimal.java
k77ch7: @headius thanks

headius:
I want there to be a better way to do this without the intermediate string, but this seems to be the best we can do with Java's BigDecimal for now. Merging. Thank you!

PR #7685, opened 2023-02-21
Add a warn form that takes a callback

headius:
Callback receives the warning's stack trace element and must return a formatted message. This allows reusing the same single trace element rather than generating it twice to build the message elsewhere.

### DIFF
--- a/core/src/main/java/org/jruby/common/RubyWarnings.java
+++ b/core/src/main/java/org/jruby/common/RubyWarnings.java
@@ -31,6 +31,8 @@ package org.jruby.common;
 
 import java.util.EnumSet;
 import java.util.Set;
+import java.util.function.Function;
+
 import org.joni.WarnCallback;
 import org.jruby.Ruby;
 import org.jruby.RubyHash;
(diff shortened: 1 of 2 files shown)


enebo (changes requested):
I think the ID should get removed.  The idea of it is nice and was originally useful for ruby-ide but we have long stopped using our main parser for that.  We do not use ID for anything internally and most new warns do what you just did and mark it miscellaneous.

 If we do ever have a need for a progmatic warn system I think we should consider completely rewriting our warn system.

I do really like the notion of making all warns lazy like this to reduce accidental overhead of forumlating a warn string to just not use it because warn is off.

headius:
@enebo This does avoid building the string until the last moment, but it introduces the construction of a lambda object. The callback form could be expanded to pass in a few key objects to avoid that allocation, like we do for Hash and lock callbacks elsewhere.

I'll see if I can cut the ID stuff out of the new code.

headius:
@enebo I cut out the ID path from this and actually fixed it to only generate a single backtrace frame (it was still doing two before).

I also pushed a change in 06f7b7d23c that makes this path zero-alloc for most common cases (adds two passthrough values for the lambda).

headius:
My commits got a little mixed up; see new hashes.

PR #7691, opened 2023-02-27
Updates irb to 1.4.2

### LINKED ISSUE #7690: Update `irb` to version `1.4.2+`
With the commit https://github.com/ruby/irb/commit/54c8df06ff9e161012f89d19a4e3aa2e0e37e1b0 `irb` project removed the inclusion of `rdoc` module, proposed with https://github.com/ruby/irb/pull/393.

The version that ship this commit is the [`1.4.2`](https://github.com/ruby/irb/compare/v1.4.1...v1.4.2).

andsel:
Update `irb` to version `1.4.2` because doesn't require `rdoc` module.

Fixes #7690

### DIFF
--- a/core/pom.xml
+++ b/core/pom.xml
@@ -81,8 +81,8 @@ DO NOT MODIFY - GENERATED CODE
       <version>1.2.0</version>
       <exclusions>
         <exclusion>
-          <groupId>com.github.jnr</groupId>
           <artifactId>jnr-ffi</artifactId>
+          <groupId>com.github.jnr</groupId>
         </exclusion>
       </exclusions>
     </dependency>
@@ -92,8 +92,8 @@ DO NOT MODIFY - GENERATED CODE
       <version>0.32.14</version>
       <exclusions>
         <exclusion>
-          <groupId>com.github.jnr</groupId>
           <artifactId>jnr-ffi</artifactId>
+          <groupId>com.github.jnr</groupId>
         </exclusion>
       </exclusions>
     </dependency>
@@ -103,8 +103,8 @@ DO NOT MODIFY - GENERATED CODE
       <version>0.38.19</version>
       <exclusions>
         <exclusion>
-          <groupId>com.github.jnr</groupId>
           <artifactId>jnr-ffi</artifactId>
+          <groupId>com.github.jnr</groupId>
         </exclusion>
       </exclusions>
     </dependency>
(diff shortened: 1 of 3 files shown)


kares (changes requested):
Hey Andrea, you updated the *pom.xml* but the *pom.rb* is really the source of truth (pom.xml gets generated from the .rb version).

I would suggest to update the [pom.rb line](https://github.com/jruby/jruby/blob/9.4.1.0/lib/pom.rb#L55) and run `mvn package` ... the xml should stay the same given you already have the changes.

andsel:
Hi @kares ! thanks for checking :-D
I've updated the `lib/pom.rb` to reflect the xml version.
If I run run `./mvnw package` on my local machine, it also updates `lib/pom.xml` and `core/pom.xml` just inverting of position the mvn's `groupid`, for example:

```diff
diff --git a/lib/pom.xml b/lib/pom.xml
index 63786be945..035ba65cbc 100644
--- a/lib/pom.xml
+++ b/lib/pom.xml
@@ -39,8 +39,8 @@ DO NOT MODIFY - GENERATED CODE
       <scope>provided</scope>
       <exclusions>
         <exclusion>
-          <groupId>rubygems</groupId>
           <artifactId>jar-dependencies</artifactId>
+          <groupId>rubygems</groupId>
         </exclusion>
       </exclusions>
     </dependency>
```

Is it correct? Or after the `mvn package` should I commit also the changed lines by maven?

enebo:
@andsel At some point maven changed something and ordering of processing a Map changes the order (Ruby stuff is put into a Map and Java side walks it).  A newer Maven will put it in the same order as we have it (we all made sure we were using newer maven at one point because we were playing ping pong with the generated xml files.

If it is not possible to update maven let me know and we will merge this as-is and I will correct that ordering manually as part of the merge

kares (approved):
:bow:

andsel:
@enebo I would have avoided playing ping pong on the xmls. However with commit 13560989a680579c32db0f62eda12228e0445a26 I've updated the `core/pom.xml` and `lib/pom.xml` with the new version of `irb` as generated by the maven wrapper.
Let me know if keep the commit or remove, before merge.

headius:
I merged this manually to avoid the reordering commit, and it is on master as of 4a1e6fcced.

Thanks for the help!

PR #7704, opened 2023-03-02
Remove --dev from tests that have own settings

headius:
Many of these tests were passing --dev to test runs that should be running our JIT or using indy. This effectively prevented those jobs from actually using the JIT. This is likely an mistake made when combining similar jobs to simplify the workflow.


headius:
Failures here are not critical items and were never reported by users. They've been documented (some with fixes) in #7707, #7708, #7711, and #7713.

PR #7720, opened 2023-03-14
More indy call optz

headius:
Various improvements to invokedynamic call binding.


headius:
Adding an indy test:jruby:jit job to CI uncovered #7726 which will have a fix from https://github.com/headius/invokebinder/pull/15 soon.

PR #7729, opened 2023-03-16
Sock fixes

enebo:
Various fixes related to ext/socket


enebo:
I did these long ago and there is some fallout that needs correcting before it lands.  bumping.

PR #7733, opened 2023-03-23
Pass keywords through call stack

headius:
This PR will attempt to pass enough information on the call stack that we can eliminate the thread-local `callInfo` state, which must be set before each call and cleared immediately upon entering the target body.


headius:
This is being closed in favor of the more comprehensive plans in #9219. Once that starts to land, there will be no point in propagating just `callInfo` since we'll have smart call sites that can wire up keyword arguments directly.

Issue #7736, opened 2023-03-28
Running JRuby 9.4 and Puma with SSL


treviateo:
Hi All,

I am trying to run JRuby 9.4.2.0 with ruby 3.1 and puma 5.6.5 with SSL. 

**System Configuration:**
JRuby: 9.4.2.0
Puma: 5.6.5
Open JDK: 1.8.0_322

I downloaded the sample keystore from _https://github.com/puma/puma/blob/429d17bca11f8d22dcc1434c9f6ac826a06fa836/examples/puma/keystore.jks_, password blahblah. And converted it to PKCS12 format.

The keystore looks ok as following:
<img width="530" alt="keystore" src="https://user-images.githubusercontent.com/129143940/228134030-c784608d-629a-49a7-bed8-2f9ced5d6b55.png">

When running JRuby with Puma and SSL binding as following:
**C:\jruby-9.4.2.0\bin\jruby.exe C:\[path_to_puma]\bin\puma -b "ssl://127.0.0.1:9292?keystore=C:\.ssl\keystore&keystore-pass=blahblah&verify_mode=none" -q -p 9191**

the server starts up ok as following:
<img width="332" alt="console1" src="https://user-images.githubusercontent.com/129143940/228134199-1afa6bd9-0557-4b70-88f5-6c60e7d4f20c.png">
<img width="412" alt="console2" src="https://user-images.githubusercontent.com/129143940/228134228-c81137e4-ace2-4db9-92ea-9243d755967b.png">

However when trying to curl (_curl https://localhost:9292 -v_), it failed as following:
<img width="308" alt="curl" src="https://user-images.githubusercontent.com/129143940/228134375-1d6f2de6-fed5-4c10-a955-34a08154aa94.png">

And on the server console, this error keep popping up everytime the curl command is entered:
**Error reached top of thread-pool: java.nio.ByteBuffer.clear()Ljava/nio/ByteBuffer; (Java::JavaLang::NoSuchMethodError)**
<img width="854" alt="error" src="https://user-images.githubusercontent.com/129143940/228141841-55408a8b-3fb0-49b8-81c5-4bb6f24a9763.png">

Does anyone have a hunch about this and could help with this issue? Thank you.

ahorek:
it's a known incompatibility with binaries compiled on a newer Java, see https://github.com/jruby/jruby/pull/5499

[...]

PR #7742, opened 2023-04-04
fix flipBuffer overloads

ahorek:
for binary compatibility with Java 8

fixes https://github.com/jruby/jruby/issues/7736


enebo:
@ahorek Thanks!

PR #7747, opened 2023-04-06
Ruby 3.1 : fix Bigdecimal

k77ch7:
This fixes some BigDecimal failures. I believe this PR will make it more compatible against  MRI, though not completely.

The following error may be caused by JSON gem
https://github.com/jruby/jruby/actions/runs/4638638467/jobs/8208637958#step:8:56

When this PR is merged, I will make a PR against JSON gem. 
```diff
diff --git a/java/src/json/ext/Parser.java b/java/src/json/ext/Parser.java
index ef28393..0a0995b 100644
--- a/java/src/json/ext/Parser.java
+++ b/java/src/json/ext/Parser.java
@@ -1195,7 +1195,7 @@ case 5:
             Ruby runtime = getRuntime();
             ByteList num = absSubSequence(p, new_p);
             IRubyObject numString = runtime.newString(num.toString());
-            return parser.decimalClass.callMethod(context, "new", numString);
+            return runtime.getKernel().callMethod(context, parser.decimalClass.getName(), numString);
         }

```


enebo:
@k77ch7 could you audit where we still use Bigdecimal.new in our internal suites.  This looks like lots of failures but I think it is the same test(s) failing in each run.  This is amazing too.  This fixed a lot of stuff!

core/src/main/java/org/jruby/ext/bigdecimal/RubyBigDecimal.java
```diff
@@ -1201,13 +1220,21 @@ public RubyBigDecimal op_pow(final ThreadContext context, IRubyObject exp) {
...
+    }
+
     @Deprecated
     public IRubyObject op_pow19(IRubyObject exp) {
         return op_pow(getRuntime().getCurrentContext(), exp);
     }
 
     @Deprecated
-    public RubyBigDecimal op_pow19(ThreadContext context, IRubyObject exp) {
+    public IRubyObject op_pow19(ThreadContext context, IRubyObject exp) {
```
kares, line 1237: we should either remove or not change signature for a deprecated method ...

core/src/main/java/org/jruby/ext/bigdecimal/RubyBigDecimal.java
```diff
@@ -483,7 +485,7 @@ private RubyBigDecimal getVpValueWithPrec(ThreadContext context, IRubyObject val
             }
 
             MathContext mathContext = new MathContext(RubyFloat.DIG + 1, getRoundingMode(context.runtime));
-            return new RubyBigDecimal(context.runtime, new BigDecimal(doubleValue, mathContext));
+            return new RubyBigDecimal(context.runtime, new BigDecimal(value.toString(), mathContext));
```
kares, line 488: this is interesting, we have the `double doubleValue` and check it's valid but still decide not to use it, in this branch.
would be nice to explain the reasoning (e.g. in a comment) of falling back to `value.to_s` here instead of using the value from RubyFloat

kares (commented):
:+1: nice to see so many tags now passing - great work.

core/src/main/java/org/jruby/ext/bigdecimal/RubyBigDecimal.java
k77ch7: i select to remove this deprecated method.

k77ch7:
@enebo i will work to audit and fix failures.

core/src/main/java/org/jruby/ext/bigdecimal/RubyBigDecimal.java
k77ch7: the above snippet is take from the test case.
```
  def test_mult_with_float
    assert_kind_of(BigDecimal, BigDecimal("3") * 1.5)
    assert_equal(BigDecimal("64.4"), BigDecimal(1) * 64.4)
  end
```

kares (commented):
:+1: should be down to one failure - due the JSON ext

but I believe this will need a PR on [JSON gem](https://github.com/flori/json)  (and a potential gem release) **before** this is merged because we would not want JRuby to ship with a broken JSON (default) gem.

the `decimal_class` logic should [try matching the C logic better](https://github.com/flori/json/blob/master/ext/json/ext/parser/parser.c#L1844-L1868) and either be backwards compatible with older JRuby or the `required_ruby_version` should be bumped ...

kares:
> I wonder if we could leave Bigdecimal#new until we fix JSON with an issue open to remove it once we get that addressed? I am open to either path as I don't know how much effort it will be to get that JSON fix and a release. Leaving new in for a while longer wouldn't break us any more than we are now

that would work as well, kind of depends on what JSON accepts and how long it takes (we had some issues accepting JRuby specific changes which ended up taking way too long to even merge, that shouldn't be the case these days) ...

I think on the JSON side having the Java ext sync-up with the C logic but in a way that keeps compatibility would be best, if possible.

k77ch7:
I agree with @enebo because of there are a lot of exclude BigDecimal tags. 
so I revert commits to leave Bigdecimal#new.
I'm Interested in improving BigDecimal compatibility.  
@kares  I would like you to look into JSON gem.

enebo:
Created  #7752 to track what needs to happen with JSON+JRuby.

PR #7751, opened 2023-04-11
Move arity-checking into variable-arity method bodies

headius:
During testing of indy-based compilation, it became obvious that we need to ensure variable-arity method bodies do their own arity-checking, so that direct calls from indy or Java do not cause array index errors. This PR adds such arity-checking to all such methods, following these rules:

Methods should do arity-checking manually if:

* they have a minimum argument count, or
* they have a maximum argument count
* and they access that argument list or pass it to normal Java methods that access that argument list

Methods do not need a manual check if they do not have minimums or maximums, or if they only pass their argument list to another variable-arity method that does a manual check.

This PR also removes arity-checking from the generated invokers, to avoid double-checking.

The main risk from this change lies in third-party extension code that has variable-arity methods but no manual checking. Passing incorrect numbers of arguments to these methods may trigger an array index error, as it would segfault in MRI (which also does not automatically check arity).

### DIFF
--- a/core/src/main/java/org/jruby/RubyArgsFile.java
+++ b/core/src/main/java/org/jruby/RubyArgsFile.java
@@ -48,6 +48,7 @@ import static org.jruby.runtime.Visibility.PRIVATE;
 import org.jruby.anno.JRubyMethod;
 import org.jruby.exceptions.RaiseException;
 import org.jruby.internal.runtime.GlobalVariable;
+import org.jruby.runtime.Arity;
 import org.jruby.runtime.Block;
 import org.jruby.runtime.CallSite;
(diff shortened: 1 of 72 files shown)


JasonLunn:
This PR introduced the behavior change in 9.4.3.0 observed in https://github.com/jruby/jruby/issues/7851. Given that this breaks 3rd party extensions that work correctly on 9.4.2.0  and below, I recommend rolling this back in 9.4.3.1 so that JRuby users using these extensions are not negatively impacted by an upgrade from an older version of JRuby.

If there is a long term need for this breaking change, it should be rescheduled and introduced in a major version bump.

PR #7762, opened 2023-04-20
[fix] avoid walking constructor instructions on every call

kares:
Since JRuby 9.3 Ruby sub-classes of Java classes are backed with a generated Java class,
the interpreter context for the constructor was always rebuild from scratch on every IR method call -> effectively looping through method's instructions.

---

**NOTE**: while this improves numbers by 5x it's still 4x slower than JRuby 9.2 (should probably implement a way of JIT-ing and avoid method lookup), the script from: https://github.com/jruby/jruby/issues/7763 ends up with the following numbers:

>```
> $ jruby slow_java_super.rb 
> jruby 9.2.20.0 (2.5.8) 2021-11-02 1a3255440b OpenJDK 64-Bit Server VM 11.0.14.1+1 on 11.0.14.1+1 +jit [linux-x86_64]
> 
> 1 RubyList.new(...) elapsed real-time: 3.4137503600068158
> 2 RubyList.new elapsed real-time: 2.6153695250031888
> ```

> ```
> $ jruby slow_java_super.rb 
> jruby 9.3.10.0 (2.6.8) 2023-02-01 107b2e6697 OpenJDK 64-Bit Server VM 11.0.14.1+1 on 11.0.14.1+1 +jit [x86_64-linux]
> 
> 1 RubyList.new(...) elapsed real-time: 54.96447179099778
> 2 RubyList.new elapsed real-time: 54.98580553600186
> ```

```
bin/jruby slow_java_super.rb 
jruby 9.3.11.0-SNAPSHOT (2.6.8) 2023-04-20 0a07e21a3b OpenJDK 64-Bit Server VM 25.322-b06 on 1.8.0_322-b06 +jit [x86_64-linux]

1 RubyList.new(...) elapsed real-time: 9.032593938994978
2 RubyList.new elapsed real-time: 7.724980847997358

```

### DIFF
--- a/core/src/main/java/org/jruby/internal/runtime/methods/MixedModeIRMethod.java
+++ b/core/src/main/java/org/jruby/internal/runtime/methods/MixedModeIRMethod.java
@@ -287,12 +287,11 @@ public class MixedModeIRMethod extends AbstractIRMethod implements Compilable<Dy
     public SplitSuperState<MethodSplitState> startSplitSuperCall(ThreadContext context, IRubyObject self,
             RubyModule clazz, String name, IRubyObject[] args, Block block) {
         // TODO: check if IR method, or is it guaranteed?
-        InterpreterContext ic = ((IRMethod) getIRScope()).builtInterperterContextForJavaConstructor();
-        if (!(ic instanceof ExitableInterpreterContext)) return null; // no super call/can't split this
+        // 2 -> IRMethod
+        ExitableInterpreterContext ic = ((IRMethod) getIRScope()).builtInterperterContextForJavaConstructor();
+        if (ic == null) return null; // no super call/can't split this
 
-        MethodSplitState state = new MethodSplitState(context, (ExitableInterpreterContext) ic, clazz, self, name);
-
-        if (IRRuntimeHelpers.isDebug()) doDebug(); // TODO?
+        MethodSplitState state = new MethodSplitState(context, ic, clazz, self, name);
 
         // TODO: JIT?
 
--- a/core/src/main/java/org/jruby/internal/runtime/methods/MixedModeIRMethod.java
+++ b/core/src/main/java/org/jruby/internal/runtime/methods/MixedModeIRMethod.java
@@ -288,7 +288,7 @@ public class MixedModeIRMethod extends AbstractIRMethod implements Compilable<Dy
             RubyModule clazz, String name, IRubyObject[] args, Block block) {
         // TODO: check if IR method, or is it guaranteed?
         // 2 -> IRMethod
-        ExitableInterpreterContext ic = ((IRMethod) getIRScope()).builtInterperterContextForJavaConstructor();
+        ExitableInterpreterContext ic = ((IRMethod) getIRScope()).builtInterpreterContextForJavaConstructor();
         if (ic == null) return null; // no super call/can't split this
 
         MethodSplitState state = new MethodSplitState(context, ic, clazz, self, name);
--- a/core/src/main/java/org/jruby/ir/IRMethod.java
+++ b/core/src/main/java/org/jruby/ir/IRMethod.java
@@ -126,6 +126,7 @@ public class IRMethod extends IRScope {
         return lazilyAcquireInterpreterContext();
     }
 
+    // 3 -> piling up here synchronized on the same method :
     /**
      * initialize methods in reified Java types will try and dispatch to the Java base classes
      * constructor when the Ruby in the initialize:
@@ -138,17 +139,33 @@ public class IRMethod extends IRScope {
      *
      * @return appropriate interpretercontext
      */
-    public synchronized InterpreterContext builtInterperterContextForJavaConstructor() {
-        InterpreterContext interpreterContext = builtInterpreterContext();
+    public ExitableInterpreterContext builtInterperterContextForJavaConstructor() {
+        ExitableInterpreterContext interpreterContextForJavaConstructor = this.interpreterContextForJavaConstructor;
+        if (interpreterContextForJavaConstructor == null) {
+            synchronized (this) {
+                interpreterContextForJavaConstructor = this.interpreterContextForJavaConstructor;
+                if (interpreterContextForJavaConstructor == null) {
+                    interpreterContextForJavaConstructor = doBuiltInterperterContextForJavaConstructor();
+                    this.interpreterContextForJavaConstructor = interpreterContextForJavaConstructor;
+                }
+            }
+        }
+        return interpreterContextForJavaConstructor == ExitableInterpreterContext.NULL ? null : interpreterContextForJavaConstructor;
+    }
 
-        if (usesSuper()) { // We know at least one super is in here somewhere
+    private volatile ExitableInterpreterContext interpreterContextForJavaConstructor;
+
+    private synchronized ExitableInterpreterContext doBuiltInterperterContextForJavaConstructor() {
+        final InterpreterContext interpreterContext = builtInterpreterContext();
+        if (usesSuper()) {
+            // We know at least one super is in here somewhere
             int ipc = 0;
             int superIPC = -1;
             CallBase superCall = null;
             Map<Label, Integer> labels = new HashMap<>();
             List<Label> earlyJumps = new ArrayList<>();
 
-            for(Instr instr: interpreterContext.getInstructions()) {
+            for (Instr instr: interpreterContext.getInstructions()) {
                 if (instr instanceof CallBase && ((CallBase) instr).getCallType() == CallType.SUPER) {
                     // We have already found one super call already.  No analysis yet to figure out if this is
                     // still ok or not so we will error.
@@ -186,7 +203,7 @@ public class IRMethod extends IRScope {
             }
         }
 
-        return interpreterContext;
+        return ExitableInterpreterContext.NULL;
     }
 
     final InterpreterContext lazilyAcquireInterpreterContext() {
--- a/core/src/main/java/org/jruby/ir/IRMethod.java
+++ b/core/src/main/java/org/jruby/ir/IRMethod.java
@@ -126,7 +126,6 @@ public class IRMethod extends IRScope {
         return lazilyAcquireInterpreterContext();
     }
 
-    // 3 -> piling up here synchronized on the same method :
     /**
      * initialize methods in reified Java types will try and dispatch to the Java base classes
      * constructor when the Ruby in the initialize:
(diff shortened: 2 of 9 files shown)


core/src/main/java/org/jruby/ir/IRMethod.java
```diff
@@ -138,17 +139,33 @@ public InterpreterContext builtInterpreterContext() {
...
+                if (interpreterContextForJavaConstructor == null) {
+                    interpreterContextForJavaConstructor = doBuiltInterperterContextForJavaConstructor();
+                    this.interpreterContextForJavaConstructor = interpreterContextForJavaConstructor;
+                }
+            }
+        }
+        return interpreterContextForJavaConstructor == ExitableInterpreterContext.NULL ? null : interpreterContextForJavaConstructor;
+    }
+
+    private ExitableInterpreterContext interpreterContextForJavaConstructor; // not volatile - worst case we built it more than once
```
ikaronen-relex, line 156: I think this _should_ be volatile, since (Exitable)InterpreterContext is not fully immutable (i.e. not all of its fields are final and contain only primitive values or other full immutable objects).

AIUI, using double-checked locking in Java without volatile risks other threads possibly accessing the object in a partially constructed state (i.e. the JVM is allowed to inline the constructor call and reorder the assignment of its return value to happen before the object is fully constructed). Using volatile will prevent this reordering.

(So would making all fields of the constructed object final, although the construction of any objects they reference could still be subject to reordering. Here, I see you've made the superCall and exitIPC fields of ExitableInterpreterContext final below, but the parent InterpreterContext class still seems to have non-final fields. In any case, even if we _could_ guarantee thread safety in this manner, I'd still much rather pay the fairly minor cost of volatile just to be sure.)

ikaronen-relex (commented):
LGTM, except that I do think interpreterContextForJavaConstructor should be volatile. (I also don't like the massive code duplication, but at least you haven't made it worse.)

enebo:
@kares not related to this PR but I wonder if adding a single instr to IR which can call Java super would be a longer term solution?  The tradeoff would be that all constructors would be immediately JIT compiled but we already are leaning on ASM here so I do not see much of a difference.  It would remove having the split interpreter.    This is a weirder suggestion than it may appear since IR JIT makes jruby methods but it probably can be repurposed to emit a valid Java constructor.

The other option is to make current JIT capable of compiling the two halves but this is more difficult.  temp vars in IR are java locals in JIT so it would require spilling locals into an array? to then pass back in for the second half (or guarantee no temps are accessible across the split).  I did attempt to make splitable methods a few years ago so we could split on BBs on huge methods but it was invasive (it did not spill locals but changed all local access to indirect through Object[] temp array).  That experiment only ran into large refactoring of JVMViisitor.  Conceptually it is a path to synthetic methods which this split logic fits.

kares:
> The other option is to make current JIT capable of compiling the two halves but this is more difficult.

I still do not understand all of the code but indeed approaching a way to JIT from an IR method seemed a bit impossible (without a massive refactoring of the Java super bits). 

Haven't thought about having an instruction for the Java super, sounds weird but at least the JITing would be much simpler (as you said byte-code is already generated in first pass anyways so shoudn't matter). Guess, we will have to attempt something on 9.4 at least. Basically anything to make the performance reasonable - comparable to 9.2.

core/src/main/java/org/jruby/ir/IRMethod.java
kares: yeah in theory I guess it makes sense but I also wonder if that is ever the case.

kares: yeah in theory I guess it makes sense but I also wonder if that is ever the case.

this is pretty much modeled after the (inherited) `protected InterpreterContext interpreterContext` which doesn't even [double check](https://github.com/jruby/jruby/blob/9.3.10.0/core/src/main/java/org/jruby/ir/IRMethod.java#L192-L198), upon lazy building, and relies upon the builder setting the (non-volatile) `interpreterContext` in a [synchronized method](https://github.com/jruby/jruby/blob/9.3.10.0/core/src/main/java/org/jruby/ir/IRMethod.java#L203). so if the re-ordering was an actual issue happening in practice we would have seen it with the `interpreterContext` (has been like this for years).

thing we could redesign the hierarchy a bit but was trying to keep this minimal (at least for 9.3) ...

ikaronen-relex: The IRMethod::hasBeenBuilt‎ method checks if `defNode == null`, where `defNode` is volatile and is set to null at the end of IRMethod::buildMethodImpl‎. I'm not 100% sure, but I _think_ that's enough under the Java memory model to prevent the problematic reordering. It still feels a bit sketchy (as in, someone could easily break it and introduce a very subtle concurrency bug without noticing), but I think it's technically safe. We don't have anything like that here, though.

ikaronen-relex: I took another look at this after the changes in bef9cce, but this still doesn't look 100% safe to me without volatile, since InterpreterContext still has non-final fields that should be initialized in the constructor for correct operation.

In particular, I believe it would be permissible for the JVM to reorder the `setInstructions` call in the InterpreterContext constructor to happen after the assignment to `interpreterContextForJavaConstructor`. If another thread then happened to grab the partially constructed object and call `getInstructions` or `getEngine` on it while `instructions` was still null, bad things could happen.

And this isn't something that can be easily fixed AFAICT, since deferring the call to `setInstructions` is apparently a valid use case (when using the other constructor that doesn't call it).

Honestly, I'd just bite the bullet and make this field volatile. Or at least benchmark it with and without volatile and see if there's even any measurable performance difference. At least, even with volatile, it should still be faster than it was before this PR.

kares: > I took another look at this after the changes in https://github.com/jruby/jruby/commit/bef9cce3577ec1ba8577128a62f11ef7d6999e68,

that one is just a low hanging fruit - making it all immutable is not possible due the lazy design of the instructions as you already noted.

> The IRMethod::hasBeenBuilt‎ method checks if `defNode == null`, where `defNode` is volatile

thanks I might have missed the volatile part there. will take another look at this, if we can do smt minimal, and if not probably just set the `volatile` as suggested ... given the normal `InterpreterContext` is behind a volatile check.

kares: updated [original commit](https://github.com/jruby/jruby/pull/7762/commits/d0710cc26875ff52c45e09ac2d5d9da73036ab90#diff-2d9019354e47c1489773d64b60d53afe974ef7676db097c83b23e46ca2fe70f3R156) so it's easy to cherry-pick and patch

kares:
moving this out of draft - have some more work locally to clean things up and at least remove duplication but to keep things minimal for 9.3 this at least avoids the contention issue on Java `super` calls.

core/src/main/java/org/jruby/ir/IRMethod.java
```diff
@@ -138,17 +139,33 @@ public InterpreterContext builtInterpreterContext() {
      *
      * @return appropriate interpretercontext
      */
-    public synchronized InterpreterContext builtInterperterContextForJavaConstructor() {
-        InterpreterContext interpreterContext = builtInterpreterContext();
+    public ExitableInterpreterContext builtInterperterContextForJavaConstructor() {
```
headius, line 142: This method name is misspelled... perhaps since this would be an API signature change, we deprecate the old spelling and add a new method with corrected spelling ("builtInterpreterContextForJavaConstructor") that has the more specific signature?

core/src/main/java/org/jruby/ir/IRMethod.java
```diff
@@ -138,17 +139,33 @@ public InterpreterContext builtInterpreterContext() {
...
+                }
+            }
+        }
+        return interpreterContextForJavaConstructor == ExitableInterpreterContext.NULL ? null : interpreterContextForJavaConstructor;
+    }
+
+    private volatile ExitableInterpreterContext interpreterContextForJavaConstructor;
 
-        if (usesSuper()) { // We know at least one super is in here somewhere
+    private synchronized ExitableInterpreterContext doBuiltInterperterContextForJavaConstructor() {
```
headius, line 158: This is new so we should update this spelling also.

core/src/main/java/org/jruby/internal/runtime/methods/MixedModeIRMethod.java
```diff
@@ -287,12 +287,11 @@ private IRubyObject INTERPRET_METHOD(ThreadContext context, InterpreterContext i
...
         // TODO: check if IR method, or is it guaranteed?
-        InterpreterContext ic = ((IRMethod) getIRScope()).builtInterperterContextForJavaConstructor();
-        if (!(ic instanceof ExitableInterpreterContext)) return null; // no super call/can't split this
+        // 2 -> IRMethod
+        ExitableInterpreterContext ic = ((IRMethod) getIRScope()).builtInterperterContextForJavaConstructor();
+        if (ic == null) return null; // no super call/can't split this
 
-        MethodSplitState state = new MethodSplitState(context, (ExitableInterpreterContext) ic, clazz, self, name);
-
-        if (IRRuntimeHelpers.isDebug()) doDebug(); // TODO?
```
headius, line 295: Don't want this?

headius (changes requested):
Just a couple naming things but otherwise looks good!

core/src/main/java/org/jruby/internal/runtime/methods/MixedModeIRMethod.java
kares: seems like copy-pasta

core/src/main/java/org/jruby/ir/IRMethod.java
kares: :+1: good call, missed the spelling issue.

core/src/main/java/org/jruby/ir/IRMethod.java
kares: :+1: fixed, also renamed to follow the `Impl` suffix convention as there are other private methods that use that here

kares:
this is now merged and expected to be in 9.3.11.0 and 9.4.4.0

the numbers are still not the same as in 9.2 but the contention issue is handled, for the record here are the numbers for 9.4 (using the script from https://github.com/jruby/jruby/issues/7763):

```
jruby 9.4.3.0 (3.1.4) 2023-06-07 3086960792 OpenJDK 64-Bit Server VM 11.0.14.1+1 on 11.0.14.1+1 +jit [x86_64-linux]

100x RubyList.new elapsed real-time: 40.68082665700058
100x RubyList2.new elapsed real-time: 5.673665171998437
100x ArrayList.new elapsed real-time: 0.16680395399453118

200x RubyList.new elapsed real-time: 88.13843810799881
200x RubyList2.new elapsed real-time: 12.741242688003695
200x ArrayList.new elapsed real-time: 0.18367314200440887

```

```
jruby 9.4.4.0-SNAPSHOT (3.1.4) 2023-08-14 cf956f1e53 OpenJDK 64-Bit Server VM 11.0.14.1+1 on 11.0.14.1+1 +jit [x86_64-linux]

100x RubyList.new elapsed real-time: 2.806557794989203
100x RubyList2.new elapsed real-time: 6.87243668200972
100x ArrayList.new elapsed real-time: 0.18653793400153518

200x RubyList.new elapsed real-time: 4.38154152900097
200x RubyList2.new elapsed real-time: 13.899855226991349
200x ArrayList.new elapsed real-time: 0.19196245599596296

```

PR #7768, opened 2023-04-23
Fix BigDecimal(0, NAN and INFINITY) to return always same object

k77ch7:
 * test_BigDecimal_with_float in  test/mri/bigdecimal/test_bigdecimal.rb


core/src/main/java/org/jruby/ext/bigdecimal/RubyBigDecimal.java
```diff
@@ -127,6 +127,12 @@ public class RubyBigDecimal extends RubyNumeric {
     private static final double SQRT_10 = 3.162277660168379332;
     private static final long NEGATIVE_ZERO_LONG_BITS = Double.doubleToLongBits(-0.0);
 
+    private static RubyBigDecimal POSITIVE_ZERO;
+    private static RubyBigDecimal NEGATIVE_ZERO;
+    private static RubyBigDecimal POSITIVE_INFINITY;
+    private static RubyBigDecimal NEGATIVE_INFINITY;
+    private static RubyBigDecimal NAN;
+
```
enebo, line 135: These cannot be static fields because:

  1. people with multiple Ruby runtimes in a single JVM will get values from other runtimes and bad things happen
  2. can potentially leak entire runtime since Runtime pins the entire system

enebo (changes requested):

core/src/main/java/org/jruby/ext/bigdecimal/RubyBigDecimal.java
k77ch7: @enebo thanks for your review. i updated my PR using setConstant and setConstantVisibility.

enebo: If we notice a performance issue with constant table lookup here we can promote these to ordinary fields in some per-runtime Java object but this looks fine.

PR #7773, opened 2023-04-26
FIXED #7710

### LINKED ISSUE #7710: Dead code in Enumerator
Due to the removal of 2- and 3-arity forms of Enumerator#initialize, we now have a lot of dead code in RubyEnumerator.

It appears these forms were deprecated in https://github.com/ruby/ruby/commit/c73b6bd7ebd01133538c645566944132dbde4d13 and eventually removed in https://github.com/ruby/ruby/commit/fa356a798aefc20725467d4fad02df8325d63e71.

There's a lot of logic here that can be deleted, spanning the Java and Ruby parts of Enumerator.

Discovered while working on the fix for #7708.

yykuk2003:
Some Ruby methods being removed in MRI. There are some dead codes in RubyEnumerator, so I removed it. It used to create Ruby for lazy evaluation and to create infinite sequences, but some of the code required modification. That modification brought dead code in Enumerator java code.

### DIFF
--- a/core/src/main/java/org/jruby/RubyEnumerator.java
+++ b/core/src/main/java/org/jruby/RubyEnumerator.java
@@ -127,7 +127,22 @@ public class RubyEnumerator extends RubyObject implements java.util.Iterator<Obj
             this.value = context.nil;
             return value;
         }
+    }
 
+    private RubyEnumerator(Ruby runtime, RubyClass type) {
+        super(runtime, type);
+        initialize(runtime, runtime.getNil(), RubyString.newEmptyString(runtime), NULL_ARRAY);
+    }
+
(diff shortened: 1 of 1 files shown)


enebo:
@yykuk2003 I do not think your PR compiles.  You should retry this but mvn clean just to make sure you are compiling all source.

yykuk2003:
> @yykuk2003 I do not think your PR compiles. You should retry this but mvn clean just to make sure you are compiling all source.

I fixed the compile error could you please recheck it for me?

enebo:
@yykuk2003 you can look at the errors from the CI run.  First thing I noticed was removing a form of Enumerator#new (which likely is triggered by missing #intialize method) which should still work but no longer does.  This is where the mspec commands I gave on channel can help you out (so you can just run specific specs that are failing.

PR #7778, opened 2023-04-29
Fix BigDecimal#to_s for edge case

k77ch7:
When a BigDecimal instance is special string case, #to_s method should return string with us ascii encoding. The following test will pass.
 * test_to_special_string in test/mri/bigdecimal/test_bigdecimal.rb


enebo:
@k77ch7 excellent work on hitting these corner cases

PR #7782, opened 2023-05-04
Fix defined?(super) on BasicObject from causing NullPointerException

snoglobe:
Related to https://github.com/ruby/ruby/pull/7758

Stops `defined?(super)` on `BasicObject` from causing a NullPointerException and instead makes it return `nil` like expected

```rb
class BasicObject
  def nullPtr
    defined?(super)
  end
end

nullPtr
```


enebo:
I will cherry-pick this back to 9.3 once it lands because this will have the same issue.

@snoglobe Thanks for the fix!

enebo:
Fixed with aedc818fc79 on jruby-9.3.

Issue #7588, opened 2023-01-20
Possible JIT optimizations


headius:
This is a list of optimizations I see could be done in the JIT but which require more work than just on the code that the JIT emits (i.e. specialized invokedynamic call sites or helper code).

* [ ] Calls to `block_given?` currently cause the method to deoptimize, since we need the method's frame to be able to retrieve the block from `RubyKernel#block_given_p`. This is unnecessary when we are in a normal method scope; we can just check the passed-in block directly, avoiding the deoptimization. This requires work in IR (use BlockGivenInstr rather than call when in a method scope, possibly with a guard for user-defined `block_given?`)
* [ ] BNEInstr logic used by JIT could be smarter about objects that are identical, if we do not need to actually call `op_equal`.
* [ ] BuildCompoundStringInstr emits a lot of code for each element being appended. This could be a single indy call with all inputs (if the order of evaluation and appending is not important) or N indy calls that do all of the coersion and appending in one shot.
* [ ]

PR #7789, opened 2023-05-12
Java call optimizations

headius:
This PR is for optimizations for calling Java methods from Ruby.


headius:
Merging as is for 9.4.3. Additional work is details in #7588.

PR #7794, opened 2023-05-17
[deps] update joda-time to latest 2.12.5

kares:
believe we should also release and update `tzdata.version` from **2019c** to **2023cgtz** (used in joda-time 2.12.2).

I wonder if the `org.jruby:joda-timezones` matters thought given the current **2019c** version was not aligned with `joda-time` 2.10.10 which is meant to be using (packages inside the joda-time.jar) **2021a**

### DIFF
--- a/core/pom.rb
+++ b/core/pom.rb
@@ -65,7 +65,7 @@ project 'JRuby Base' do
   jar 'org.osgi:org.osgi.core:5.0.0', :scope => 'provided'
 
   # joda timezone must be before joda-time to be packed correctly
-  jar 'org.jruby:joda-timezones:${tzdata.version}', :scope => '${tzdata.scope}'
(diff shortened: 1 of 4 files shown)


core/pom.rb
```diff
@@ -65,7 +65,7 @@
   jar 'org.osgi:org.osgi.core:5.0.0', :scope => 'provided'
 
   # joda timezone must be before joda-time to be packed correctly
-  jar 'org.jruby:joda-timezones:${tzdata.version}', :scope => '${tzdata.scope}'
+  # jar 'org.jruby:joda-timezones:${tzdata.version}', :scope => '${tzdata.scope}'
```
kares, line 68: I believe this is redundant given the joda-time includes the up-to-date tz-data and we want to make sure to use those

Issue #7805, opened 2023-05-23
High lock contention in Socket.select in JRuby 9.3


jsvd:
**Environment Information**
  
Provide at least:
* JRuby version (`jruby -v`) and command line (flags, JRUBY_OPTS, etc)
```
jruby 9.3.10.0 (2.6.8) 2023-02-01 107b2e6697 OpenJDK 64-Bit Server VM 17.0.6+10 on 17.0.6+10 +indy +jit [x86_64-linux]
```
```
-Xms8g -Xmx8g -Djava.awt.headless=true -Dfile.encoding=UTF-8 -Djruby.compile.invokedynamic=true -Djruby.jit.threshold=0 -Djruby.regexp.interruptible=true -XX:+HeapDumpOnOutOfMemoryError -Djava.security.egd=file:/dev/urandom -Dlog4j2.isThreadContextMapInheritable=true -Djdk.io.File.enableADS=true --add-exports=jdk.compiler/com.sun.tools.javac.api=ALL-UNNAMED --add-exports=jdk.compiler/com.sun.tools.javac.file=ALL-UNNAMED --add-exports=jdk.compiler/com.sun.tools.javac.parser=ALL-UNNAMED --add-exports=jdk.compiler/com.sun.tools.javac.tree=ALL-UNNAMED --add-exports=jdk.compiler/com.sun.tools.javac.util=ALL-UNNAMED --add-opens=java.base/java.security=ALL-UNNAMED --add-opens=java.base/java.io=ALL-UNNAMED --add-opens=java.base/java.nio.channels=ALL-UNNAMED --add-opens=java.base/sun.nio.ch=ALL-UNNAMED --add-opens=java.management/sun.management=ALL-UNNAMED
```
* Operating system and platform: observed in SLES/Ubuntu/Debian kernels 4.x and 5.x

Other relevant info you may wish to add:

* Code runs inside Logstash: https://github.com/logstash-plugins/logstash-input-syslog/blob/main/lib/logstash/inputs/syslog.rb#L235-L249

#### Problem description

Multiple endpoints pushing syslog data to a central TCP server.
Since the upgrade from JRuby 9.2 to 9.3.10.0, multiple users of Logstash are observing very high cpu usage.
I have not yet been able to reproduce this locally in macos or a debian vm on GCP.
What users observe is that multiple threads are in a blocked state.

If everything is going well, what we see is a runnable thread:

```
[...]

PR #7806, opened 2023-05-23
Reduce contention in heavy IO operations

headius:
This moves the creation of a new selector outside of the sync section, to avoid blocking threads that are just trying to put a selector back when we are waiting to open a new one.

This whole class could use some improvement using nonblocking collections, but this should help reduce contention for now.

Edit: This has expanded to remove additional areas of contention.

See jruby/jruby#7805.


headius:
I've pushed an additional commit that removes all locking from SelectorPool by using concurrent collections. If the contention is really due to locking in SelectorPool, this should definitely eliminate it.

headius:
Additional commit to remove synchronization from RubyBasicObject.dataGetStruct/dataWrapStruct because they should not need it (they utilize VariableTableManager which is concurrency-safe).

Issue #6607, opened 2021-03-10
Clean up javadoc for stricter Java 11


headius:
We have various places in our javadoc that are rejected by the strict javadoc generator in Java 11. We should be able to build javadoc on all released versions of Java, but at least on 11.

```
[ERROR] Failed to execute goal org.apache.maven.plugins:maven-javadoc-plugin:2.7:jar (attach-javadocs) on project jruby-core: MavenReportException: Error while creating archive:
[ERROR] Exit code: 1 - /Users/headius/projects/jruby-9.2/tmp/jruby-dist-build/core/src/main/java/org/jruby/RubyUnboundMethod.java:45: error: unknown tag: note
[ERROR]  * @note This was renamed from UnboundMethod.java
[ERROR]    ^
[ERROR] /Users/headius/projects/jruby-9.2/tmp/jruby-dist-build/core/src/main/java/org/jruby/embed/LocalVariableBehavior.java:34: error: malformed HTML
[ERROR]  * @author Yoko Harada <yokolet@gmail.com>
[ERROR]                        ^
[ERROR] /Users/headius/projects/jruby-9.2/tmp/jruby-dist-build/core/src/main/java/org/jruby/embed/LocalVariableBehavior.java:34: error: bad use of '>'
[ERROR]  * @author Yoko Harada <yokolet@gmail.com>
[ERROR]                                          ^
[ERROR] /Users/headius/projects/jruby-9.2/tmp/jruby-dist-build/core/src/main/java/org/jruby/embed/EvalFailedException.java:35: error: malformed HTML
[ERROR]  * @author Yoko Harada <yokolet@gmail.com>
[ERROR]                        ^
[ERROR] /Users/headius/projects/jruby-9.2/tmp/jruby-dist-build/core/src/main/java/org/jruby/embed/EvalFailedException.java:35: error: bad use of '>'
[ERROR]  * @author Yoko Harada <yokolet@gmail.com>
...
```

There are several pages of minor things to fix in the javadocs.

headius:
I started cleaning this up and every set of fixes just opened up more things that were wrong. The javadoc linter penalizes you for partial documentation but ignores methods with no documentation at all. I will be looking into disabling the linter in the short term and doing a larger intentional javadoc update effort after 9.3 is released.

headius:
@edipofederle I think I just abandoned the work when I discovered the JDK bug. We definitely want to revisit this now that the bug is fixed!

headius:
[...]

PR #7808, opened 2023-05-24
Update to backport9 1.13 for module javadoc fix

headius:
We still use an internal class, StackWalker8, which I exposed in backport9 1.13. This allows javadoc under Java 11 to generate successfully.

There's still more work to fully modularize.

Fixes the last part of #6607.


headius:
Cancelled workflow because the only change here just upgrades backport9 to a version that exposes more packages through JPMS.

PR #7812, opened 2023-05-24
Remove the CALL_KEYWORD_EMPTY masking

headius:
This mask can cause empty keyword flags from a previous call to propagate into the next call. If that call passes keywords itself, it will impact arity checking as the empty flag triggers kwarg negotiation to not subtract one from the incoming arg count.

I am not sure why this flag gets propagated in this way, so I am pushing this as a test PR.


headius:
This was experimental and showed that the flag is definitely in use. The actual issue was fixed by @enebo in #7816 and we'll reoptimize keywords later on.

PR #7821, opened 2023-05-30
enumerator product

gintama91:


enebo:
@gintama91 Thanks for contributing.  

How does this new code get called?  I do not see any annotation like @JRubyMethod.  I think this should be put on enumeratorProduct (proabably just renamed to "product").  I also believe this making explicit casts to RubyEnumerator may end up potentially causing classcastexceptions with improperly supplied arguments.  Test should help figure that out.

For testing I believe there should exist some specs under specs/ruby (although it is possible we have not updated those specs in a while but there should be a enumerator/product_spec.rb somewhere in there.

enebo:
Closing out old PRs.  If any progress is made then either reopen or make a new PR.

PR #7823, opened 2023-05-30
Reduce contention in heavy IO operations

headius:
This is the 9.4 version of #7806.


headius:
See #7805 for the original report.

headius:
Unneeded with @kares merging of 9.3. That merge also pulls over #7784 which fixed a build issue with the `write` arity-splitting I patched differently here.

Issue #7851, opened 2023-07-13
@JRubyMethod minimum argument not enforced in 9.4.3.0


JasonLunn:
* JRuby version: `jruby 9.4.3.0 (3.1.4) 2023-06-07 3086960792 OpenJDK 64-Bit Server VM 20.0.1+9 on 20.0.1+9 +jit [x86_64-darwin]`
* Operating system and platform: `Darwin jatl-macbookpro 22.5.0 Darwin Kernel Version 22.5.0: Thu Jun  8 22:22:22 PDT 2023; root:xnu-8796.121.3~7/RELEASE_X86_64 x86_64`

**Expected Behavior**

Given a Ruby class defined in Java via `@JRubyClass`, expect the `@JRubyMethod()` annotation to enforce the minimum number of arguments passed to a method consistently with previous JRuby versions.

**Actual Behavior**

JRuby 9.2 and 9.3 are observed to block invocations of underlying annotated Java method if the Ruby invocation does not have sufficient arguments, whereas 9.4 is observed to pass a 0 length array to the Java function instead.

* Describe or show the actual behavior.

`Foo.java`:

```java
import org.jruby.anno.JRubyClass;
import org.jruby.anno.JRubyMethod;
import org.jruby.Ruby;
import org.jruby.RubyClass;
import org.jruby.RubyModule;
import org.jruby.RubyObject;
import org.jruby.runtime.ObjectAllocator;
import org.jruby.runtime.ThreadContext;
import org.jruby.runtime.builtin.IRubyObject;

@JRubyClass(name = "Foo")
public class Foo extends RubyObject {
  public Foo(Ruby runtime, RubyClass klazz) {
    super(runtime, klazz);
  }

  @JRubyMethod(required = 1, optional = 2)
  public IRubyObject initialize(ThreadContext c, IRubyObject[] args) {
    System.out.println("Args has length " + args.length);
    return this;
  }

  public static void createBar(Ruby runtime) {
    RubyModule mFooBar = runtime.getClassFromPath("FuBar");
    RubyClass cBar =
        mFooBar.defineClassUnder(
            "Bar",
            runtime.getObject(),
            new ObjectAllocator() {
              @Override
              public IRubyObject allocate(Ruby runtime, RubyClass klazz) {
                return new Foo(runtime, klazz);
              }
            });
    cBar.defineAnnotatedMethods(Foo.class);
  }
}
```

`BarService.java`:
```
import java.io.IOException;
import org.jruby.Ruby;
[...]

PR #7860, opened 2023-07-17
Restore arity-checking with an opt-out flag

headius:
Removing automatic arity-checking for Java-based Ruby methods broke some users' test suites that expected ArgumentError to be thrown; without a manual arity-check, the target methods now tried to access the incoming vararg array blindly and raised Java index errors. This was reported as #7851.

This fixes the issue by restoring arity-checking by default and providing an opt-out flag that we and users can use to indicate that the Java code will check arity manually.

### DIFF
--- a/core/src/main/java/org/jruby/RubyArgsFile.java
+++ b/core/src/main/java/org/jruby/RubyArgsFile.java
@@ -405,7 +405,7 @@ public class RubyArgsFile extends RubyObject {
     /** Read a line.
      *
      */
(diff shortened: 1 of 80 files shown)


JasonLunn:
Thanks, @headius

headius:
Added a test for basic `checkArity` behavior. This is ready to go.

headius:
@JasonLunn If you could verify this fixes your case, we'd appreciate it!

PR #7867, opened 2023-07-23
[deps] update jruby-openssl to 0.14.2

chadlwilson:
See https://github.com/jruby/jruby-openssl/releases/tag/v0.14.2

Not sure from @kares if this is "ready", but there is some CVE noise (that doesn't affect JRuby) mentioned at https://github.com/jruby/jruby-openssl/pull/278 within the bundled bouncy castle version which would be nice to get rid of on the next JRuby `9.4.x` release.

### DIFF
--- a/lib/pom.rb
+++ b/lib/pom.rb
@@ -58,7 +58,7 @@ default_gems = [
     ['irb', '1.4.2'],
     ['jar-dependencies', '0.4.1'],
     ['jruby-readline', '1.3.7'],
-    ['jruby-openssl', '0.14.1'],
+    ['jruby-openssl', '0.14.2'],
     ['json', '2.6.1'],
     ['logger', '1.5.1'],
     ['mutex_m', '0.1.1'],
--- a/lib/pom.xml
+++ b/lib/pom.xml
@@ -385,7 +385,7 @@ DO NOT MODIFY - GENERATED CODE
     <dependency>
       <groupId>rubygems</groupId>
       <artifactId>jruby-openssl</artifactId>
-      <version>0.14.1</version>
+      <version>0.14.2</version>
       <type>gem</type>
       <scope>provided</scope>
       <exclusions>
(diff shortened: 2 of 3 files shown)


chadlwilson:
There seem to be a few flakes here, but imagine they might be OK after a rerun?

enebo:
@chadlwilson only one failure seems valid.  I am guessing newer openssl is larger than before so we exceed some size check.  This is not a real problem though unless we changed packaging of that gem and it has extra crud in it.

@kares is this something we should update yet?

chadlwilson:
@enebo Hmm, that seems odd. I'd actually expect it should **reduce** jar size by about 3 MB as they removed a whole lot of stuff from Bouncy Castle in this release which was quite a bit component within jruby-complete etc if I recall correctly.

Might require some digging to see why this is.

chadlwilson:
OK, so the increase does seem to be expected. While `1.74` used by jruby-openssl `0.14.2` **did** reduce `bcprov-jdk18on` (provider) jar size compared to `1.73`, it's still increased a lot compared to `1.71` which was the previous version in use by jruby-openssl `0.14.1`.

This is mainly due to addition of a lot of post quantum crypto support, even though [some algorithms have been removed](https://www.bouncycastle.org/releasenotes.html#r1rv74) in `1.74`.

https://rubygems.org/gems/jruby-openssl/versions/0.14.1-java 8.58 MB
https://rubygems.org/gems/jruby-openssl/versions/0.14.2-java 10.70 MB (+ 2.1 MB) 

bcprov-jdk18on raw jar sizes
`1.71` 5.8 MB
`1.72` 10.8 MB (+ 5.0 MB)
`1.73` 10.9 MB (+ 5.1 MB)
`1.74` 8.0 MB (+ 2.1 MB)

So while unfortunate, the size increase is expected 😬

enebo:
@chadlwilson Thanks for doing that sleuthing.  Now we will just see what @kares has to say.

kares (approved):
Thanks for taking care of this Chad, I do recall you mentioned code being removed from BC.

But indeed as I updated BC in JOSSL noticed a size increase, did not investigate why. 
I do assume it's legit - seems the bcprov jar had more than 2MB increase since 1.71.

kares:
for the record, also merged the changes for next 9.3: https://github.com/jruby/jruby/pull/7877

PR #7870, opened 2023-07-28
Launcher tweaks

mrnoname1000:
Forgot to PR some improvements to the launcher! Should be pretty self-explanatory and I made some notes going back over them:

Finally got rid of the realpath function from Stackoverflow and replaced it with some more self-explanatory functions. Should be faster now too since it runs as few external commands as possible. Hopefully it doesn't break things, if you want to test the readlink logic you can comment out lines 290-292 or remove realpath from your PATH.

Also I added a couple lines for compatibility with ksh93! I think every Bourne shell that's installable on modern operating systems can launch jruby now.

### DIFF
--- a/bin/jruby.sh
+++ b/bin/jruby.sh
@@ -340,13 +340,22 @@ else
     fi
 fi
 
-# Detect modularized Java if modules file is present or a MODULES line appears in release
-if [ -f "$JAVA_HOME"/lib/modules ] \
-    || {
-        [ -f "$JAVA_HOME"/release ] \
-        && grep -q ^MODULES "$JAVA_HOME"/release
-    }
-then
+# Detect modularized Java
+java_is_modular() {
+    # check that modules file is present
+    if [ -f "$JAVA_HOME"/lib/modules ]; then
+        return 0
+    fi
+
+    # check if a MODULES line appears in release
+    if [ -f "$JAVA_HOME"/release ] && grep -q ^MODULES "$JAVA_HOME"/release; then
+        return 0
+    fi
+
+    return 1
+}
+
+if java_is_modular; then
     use_modules=true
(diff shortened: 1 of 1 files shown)


enebo:
The carrot of this PR is really that we should stop getting "I can't run this script in my zorronexusobscure linux distro?".  The speed is always good but likely a drop in the bucket compared to actually boostrapping the JVM.  @headius will you review this?  I think we should land it quicker to find potential problems than later.

headius:
Sounds great! I will review and merge this weekend or early next week.

bin/jruby.sh
```diff
@@ -44,70 +52,66 @@ esceval()
...
 # to the named by LISTNAME1, preserving order.
 preextend() {
-    map prepend "$@"
+    eval "$1=\"\${$2} \${$1}\""
+}
+
+# echo [STRING...]
+#
+# Dumb echo, i.e. print arguments joined by spaces with no further processing
+echo() {
```
headius, line 113: Functions like this will only be visible within this script, correct?

headius:
Ok looks good to me with my level of shell knowledge! I will merge.

headius (approved):

bin/jruby.sh
headius: After chatting with @mrnoname1000 and trying some subprocess checks, it appears the functions defined in this launcher script do not propagate beyond the script:

```
$ jruby -e 'system "which esceval"'
$ jruby -e 'system "command -v esceval"'
$
```

mrnoname1000: Yeah, unless it's run in bash and exported with `export -f echo` or a preceding `set -a` it won't inherit via the environment.

PR #7872, opened 2023-08-01
Migrate to new mavengem

headius:
Mavengem has used the v1/dependencies API on rubygems.org to get gem information, but that API has now gone away (whitelisted for us until August 8). This change migrates to a newer version of mavengem that uses newer APIs to make up the same functionality.

See jruby/mavengem#9

### DIFF
--- a/.mvn/extensions.xml
+++ b/.mvn/extensions.xml
@@ -3,6 +3,6 @@
   <extension>
     <groupId>io.takari.polyglot</groupId>
     <artifactId>polyglot-ruby</artifactId>
-    <version>0.4.7</version>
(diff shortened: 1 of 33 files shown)


headius:
The RG.org folks did not actually timebox the whitelisting so we have as long as we need... but this is nearly complete. I have released mavengem 2.0.0, jruby-maven-plugins 3.0.0, and maven-tools 1.2.0 and we're just waiting for a 0.4.11 release of polyglot-maven due soon.

Once everything is released, we'll merge this and give it a few weeks to make sure everything is working properly.

headius:
All green! Ship it!

PR #7880, opened 2023-08-10
Migrate to new mavengem (9.3 version)

headius:
Mavengem has used the v1/dependencies API on rubygems.org to get gem information, but that API has now gone away (whitelisted for us until August 8). This change migrates to a newer version of mavengem that uses newer APIs to make up the same functionality.

See jruby/mavengem#9

### DIFF
--- a/.mvn/extensions.xml
+++ b/.mvn/extensions.xml
@@ -3,6 +3,6 @@
   <extension>
     <groupId>io.takari.polyglot</groupId>
     <artifactId>polyglot-ruby</artifactId>
-    <version>0.4.11-SNAPSHOT</version>
+    <version>0.4.11</version>
   </extension>
 </extensions>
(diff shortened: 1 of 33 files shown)


headius:
See #7872 for the master version. This will likely win and get merged forward.

.mvn/extensions.xml
```diff
@@ -3,6 +3,6 @@
   <extension>
     <groupId>io.takari.polyglot</groupId>
     <artifactId>polyglot-ruby</artifactId>
-    <version>0.4.7</version>
+    <version>0.4.11-SNAPSHOT</version>
```
kares, line 6: this is due awaiting a release from upstream, or?

headius: Correct. Hopefully this week!

headius:
Waiting on https://github.com/takari/polyglot-maven/pull/263 which won't be released until at least next week.

headius:
This is just about ready. The plugins need to be updated one more time and then we'll be on current versions of everything.

headius:
I think we're ready to go here.

PR #7882, opened 2023-08-11
[fix] duplicate hash key literal removal

kares:
This PR is a follow-up on https://github.com/jruby/jruby/pull/7623, namely the detected collection usage issue: https://github.com/jruby/jruby/commit/0224d5d2991abf6fd2492ae9e1bbf52a30bb181c

While fixing the collection removal (type) issue I discovered that the actual AST node removal on duplicate hash keys is problematic (there's a need to evaluate the RHS even despite it's end result being thrown away), so the actual removal is still NOT happening.

**The PR still fixes the duplicate line reporting - will now point to the correct (duplicate) location as does MRI! **

### DIFF
--- a/core/src/main/java/org/jruby/parser/RubyParserBase.java
+++ b/core/src/main/java/org/jruby/parser/RubyParserBase.java
@@ -1515,7 +1515,6 @@ public abstract class RubyParserBase {
 
     public Node remove_duplicate_keys(final HashNode hash) {
         final Map<Node, KeyValuePair<Node, Node>> encounteredKeys = new HashMap<>();
-        List<KeyValuePair<Node, Node>> pairsToRemove = null;
 
         for (KeyValuePair<Node,Node> pair: hash.getPairs()) {
             final Node key = pair.getKey();
@@ -1525,9 +1524,6 @@ public abstract class RubyParserBase {
                 IRubyObject value = ((LiteralValue) key).literalValue(runtime);
                 warning(ID.AMBIGUOUS_ARGUMENT, lexer.getFile(), hash.getLine(), str(runtime, "key ", value.inspect(),
                         " is duplicated and overwritten on line " + (key.getLine() + 1)));
-
-                if (pairsToRemove == null) pairsToRemove = new ArrayList<>(4);
-                pairsToRemove.add(encounteredKeys.get(key));
             }
             // even if the key was previously seen, we replace the value to properly remove multiple duplicates
             encounteredKeys.put(key, pair);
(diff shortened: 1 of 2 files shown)


core/src/main/java/org/jruby/parser/RubyParserBase.java
```diff
@@ -1511,28 +1513,30 @@ protected ArgsTailHolder new_args_tail(int line, ListNode keywordArg,
...
+            encounteredKeys.put(key, pair);
         }
 
-        for (Node key: encounteredKeys) {
-            hash.getPairs().remove(key);
-        }
+        // NOTE: we do not really remove the value part (RHS) as we still need to evaluate the expression - despite
+        // the value being dropped the side effects are desired and something MRI evaluates explicitly during removal
+        // if (pairsToRemove != null) hash.removeAll(pairsToRemove);
+
```
enebo, line 1539: I think you can just remove that removeAll comment and leave the text.  That code is not correct even if we do eventually address the comment.

kares: okay I considered removing but in that case I should do more cleanup - since we no longer need the `pairsToRemove` collection. having the comment without the collected nodes to be removed seemed weird, but I am pbly biased here :wink:

enebo: @kares I know think this should not be in parsing at all but in building.  Building is capable of emitting instrs for the rhs and also processes kwargs in its' own right.  So you could just write a comment that IR Build is responsible for modifying the hash too?

core/src/main/java/org/jruby/parser/RubyParserBase.java
```diff
@@ -1511,28 +1513,25 @@ protected ArgsTailHolder new_args_tail(int line, ListNode keywordArg,
...
             }
+            // even if the key was previously seen, we replace the value to properly remove multiple duplicates
+            encounteredKeys.put(key, pair);
         }
 
-        for (Node key: encounteredKeys) {
-            hash.getPairs().remove(key);
-        }
+        // NOTE: we do not really remove the value part (RHS) as we still need to evaluate the expression - despite
+        // the value being dropped the side effects are desired and something MRI evaluates explicitly during removal
```
kares, line 1533:
```suggestion
        // NOTE: we do not really remove the value part (RHS) as in that case we should evaluate the code - despite
        // the value being dropped the side effects are desired and something MRI evaluates explicitly during removal
        // with JRuby the modification of the hash should be done in the IR and not here (during the parsing phase)...
```

enebo: @kares yeah that is fine.

PR #7887, opened 2023-08-15
Fix BigDecimal#power with negative integer

k77ch7:
This commits improves precision of power algorithm.
This commits also updates getPrec() of private method to match MRI behavior.

The following test will pass.
 * test_power_without_prec in test/mri/bigdecimal/test_bigdecimal.rb


k77ch7:
These snippets match exactly with mri 3.1.3
`BigDecimal("987654.3210987654321") ** -987`
`BigDecimal("0.00070809010203040506") ** -123`

enebo:
Only 13 excludes left.  Thanks @k77ch7.

PR #7884, opened 2023-08-14
Update joni to 2.2.1 to fix regex match regression for 9.3

skunkworker:
9.3.10.0 is currently broken with https://github.com/jruby/jruby/issues/7830. 9.3.9.0 is healthy. And `joni` should be updated in order to have a healthy `9.3.11.0` release in the future.

PR #7889, opened 2023-08-15
Update joni and strscan for recent fixes

headius:
* joni updated to 2.2.1 with the new Region factory and shapes.
* strscan updated to a modified version of the gem 3.0.6 that can compile against joni 2.2 (which hides Region fields).

Note that these changes mean any users depending on access to joni Region fields, or its constructor, will see some breakage, but these changes have been in the wild in 9.4.x for sevreal months without issues.

Relates to #7884


skunkworker:
Thanks for making this @headius

PR #7897, opened 2023-08-20
Update test_bigdecimal.rb and fix a few failures

k77ch7:
This PR updates test_bigdecimal.rb from bigdecimal v3.1.4.
And this PR fixes a few failures.
 * BigDecimal#remainder with Infinity
 * The exception message in BigDecimal#round with invalid round mode

I believe that this implementation enhances JRuby's compatibility. However, please note that an exclusion tag has been added.


enebo:
@k77ch7 I think that is ok but it is always hard to know what people actually hit in these libraries.

k77ch7:
@enebo I feel the same.

By the way, do you have any plans to gemify JRuby's BigDecimal implementation?
If you do, I think it would be preferable to reduce the number of failure/exclusion tags.
I suppose that fixing the remaining failures could be a challenging problem.

enebo:
@k77ch7 I just going to :+1: this but you are the one who needs thanking.  Bigdecimal would not be where it is today without your work.

PR #7907, opened 2023-08-28
Improvements for vthread fibers

headius:
* Use method handles to launch
* Guard behind an option so they can be disabled


headius:
I'll do the Java 21 CI updates in a separate PR.

PR #7919, opened 2023-09-03
Cleanup emacs-related temp files

jcharaoui:


headius:
Hah, those buggers keep sneaking back in.

PR #7926, opened 2023-09-05
Proc.new with no args no longer supported in 3.1

enebo:
Also noticed IR persist would not restore this state :smile:


headius:
Wow, it's about time they removed this from Ruby. Cassandra effect all over again; We knew this feature would eventually go away.

Issue #7902, opened 2023-08-24
Marshal.load error for Time with ivar


segiddins:
**Environment Information**
  
Provide at least:
`jruby 9.4.3.0 (3.1.4) 2023-06-07 3086960792 OpenJDK 64-Bit Server VM 19.0.1+10-21 on 19.0.1+10-21 +jit [arm64-darwin]`

**Expected Behavior**

The following test passes on MRI 3.2.2:

```ruby
require 'rubygems'
require 'minitest/autorun'

class T < Minitest::Test
  def test_a
    a = Marshal.load("\x04\bIu:\tTime\r\x01\xDF\x1E\x80eDI}\b:\n@type:\fruntime:\voffseti\xFE\x90\x9D:\tzoneI\"\bPDT\x06:\x06EF")
    assert_instance_of Time, a
    assert_equal :runtime, a.instance_variable_get(:@type)
    
  end
  
  def test_b
    b = Marshal.load("\x04\bIu:\tTime\r\x01\xDF\x1E\x80\x16o\x17}\t:\rsubmicro\"\x06\x00:\voffseti\xFE\x90\x9D:\tzoneI\"\bPDT\x06:\x06ET:\n@type:\fruntime")
    assert_instance_of Time, b
    assert_equal :runtime, b.instance_variable_get(:@type)
  end
  
  def test_c
    c = Marshal.load("\x04\bIu:\tTime\r\x01\xDF\x1E\x80qX\x8B/\t:\voffseti\xFE\x90\x9D:\tzoneI\"\bPDT\x06:\x06ET:\n@type:\fruntime:\rsubmicro\"\x06\x00")
    assert_instance_of Time, c
    assert_equal :runtime, c.instance_variable_get(:@type)
  end
end
```

**Actual Behavior**

On jruby:

```
jruby 9.4.3.0 (3.1.4) 2023-06-07 3086960792 OpenJDK 64-Bit Server VM 19.0.1+10-21 on 19.0.1+10-21 +jit [arm64-darwin]
Run options: --seed 65435

# Running:

EEE

Finished in 0.031763s, 94.4484 runs/s, 0.0000 assertions/s.

  1) Error:
T#test_a:
ArgumentError: dump format error for symbol(0xb)
    org/jruby/RubyMarshal.java:151:in `load'
    Untitled 17.rb:10:in `test_a'

  2) Error:
T#test_b:
ArgumentError: marshal data too short
    org/jruby/RubyMarshal.java:153:in `load'
[...]

PR #7930, opened 2023-09-09
Improve marshaling of Time

headius:
Relates to #7902. These fixes should eventually get those examples passing.


headius:
I obviously broke something along the way. Bumping.

headius:
Did not have time to get back to this. It might slip to 10, where @enebo is working on Time improvements separately.

PR #7940, opened 2023-09-21
More testing with Java 21

kares:
The PR adds CI testing with Java 21 (release), also added some variations on the Java flavor used (with temurin).

The embedded (jruby-complete) test targets now also work on Java > 8 and are actively tested against later versions.

A few minor (Java) deprecations are addressed, most of the remaining ones are around SecurityManager and the related access exception being used... which are likely going to be tricky to get rid of.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -16,7 +16,7 @@ jobs:
     strategy:
       matrix:
         target: ['test:jruby:int', 'spec:ruby:fast', 'spec:ji', 'spec:ffi']
-        java-version: ['8', '17']
+        java-version: ['8', '17', '21']
       fail-fast: false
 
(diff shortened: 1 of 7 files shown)


headius (approved):
The changes look good to me overall.

I am a bit concerned with the number of additional jobs added, however. I count something like 28 additional jobs. Some of these are quick, but several are very heavy and will add time to CI (e.g. spec:ruby:fast variations).

We can discuss this after merging (since the missing 21-ea is breaking builds right now) but the short story is that I made some suites only test on 8 and whatever is latest, assuming that passing both would mean the intermediate versions work too. We need to balance how long CI runs with how much we need to reasonably test.

headius:
I'm going ahead with merging this since master is broken due to "21-ea" not being an installable Java anymore.

PR #7944, opened 2023-09-29
Support for fiber scheduler API

headius:
Work in progress. Mostly modeling this after logic on CRuby HEAD.

Optimistically marking this for 9.4.5 but probably would be better to drop in the next major release.


headius:
We will ship this, disabled, in 9.4.5.0. Set property `jruby.experimental.fiber.scheduler` to enable.

PR #7941, opened 2023-09-25
Update ffi to 1.16.1

ntkme:
Closes #7843.

PR #7945, opened 2023-09-30
Update ffi to 1.16.2

chadlwilson:
Follow-up to #7941, looks like there's a minor regression in 1.16.1 (https://github.com/ffi/ffi/issues/1050)


headius:
Thanks!

FYI, when you make changes to any of the pom.rb files, you should run a build so that the pom.xml gets updated too. Several of them are generated but committed, so if we don't commit the change they look dirty for everyone.

chadlwilson:
Ahh ok, my bad. Find and replace fail and copying original PR too closely. 😅

headius:
We forget about them frequently! No worries.

Issue #7942, opened 2023-09-27
fnmatch not matching glob syntax in 9.3.8 but it did in 9.2.20.1


shivanesabharwal:
**Environment Information**
  
Provide at least:
* JRuby version (`jruby -v`) and command line (flags, JRUBY_OPTS, etc):
```
jruby 9.3.8.0 (2.6.8) 2022-09-28 ffffffffff OpenJDK 64-Bit Server VM 11.0.15+0-adhoc..source on 11.0.15+0-adhoc..source +jit [x86_64-linux]

```
* Operating system and platform (e.g. `uname -a`)

```
Linux 6.3.11-1rodete2-amd64 #1 SMP PREEMPT_DYNAMIC Debian 6.3.11-1rodete2 (2023-08-24) x86_64 GNU/Linux
```

**Expected Behavior**

On Jruby 9.2.20.1 in an irb session:
```
File.fnmatch?('employees2*.view*l', 'employees2.view.lkml', File::FNM_PATHNAME)
=> true
```

**Actual Behavior**
I expected this to continue returning true. However after upgrade to 9.3.8.0 we see:
```
File.fnmatch?('employees2*.view*l', 'employees2.view.lkml', File::FNM_PATHNAME)
=> false
```

headius:
I see a few minor changes in `fnmatch` recently, but nothing that should cause this sort of change.

Perhaps you can try to bisect it?

shivanesabharwal:
Interestingly, when I check out the tag associated with JRuby 9.3.8.0, I am unable to reproduce the issue.

The only difference I can see is the value for `RUBY_VERSION` in the `irb` session.

in the JRuby repo:
```
shivane@shivanes:~/jruby$ irb
irb(main):001:0> RUBY_VERSION
=> "3.1.2"
```

In my project:
```
irb(main):002:0> RUBY_VERSION
=> "2.6.8"
```

I'm not sure I understand the relationship between the JRuby version and the Ruby version. Do you have any pointers here?

shivanesabharwal:
I thought the `jruby` package fixes the ruby version used but that doesn't seem to be the case. Also running into issues building the jruby jar locally, which is preventing me from git bisecting on my clone of this repo. I don't have much else in the way of contributing.

ahorek:
it's this change https://github.com/jruby/jruby/pull/6231 + a dirty fix https://github.com/jruby/jruby/pull/7947 :)

[...]

PR #7947, opened 2023-10-05
fnmatch not matching glob syntax

ahorek:
closes https://github.com/jruby/jruby/issues/7942


headius:
@ahorek This does look like the right fix. The failed path can be followed normally from the various `break` in the switch, but it needs an additional trigger at the top of the loop.

I'll merge and add some comments about why this code is duplicated.

PR #7951, opened 2023-10-09
backport: fnmatch not matching glob syntax

ahorek:
backport https://github.com/jruby/jruby/issues/7942 to 9.3.x

### DIFF
--- a/core/src/main/java/org/jruby/util/Dir.java
+++ b/core/src/main/java/org/jruby/util/Dir.java
@@ -122,7 +122,18 @@ public class Dir {
                 while (true) {
                     // in place of expected C null char check that falls into default in switch below
                     if (p >= pend) {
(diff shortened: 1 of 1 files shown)


shivanesabharwal:
@headius Is there a timeline for when new patch release builds make it into Maven Central repository?

headius:
@shivanesabharwal No not really. If there's significant community need we will run a release. We would recommend upgrading to 9.4 if you are stuck on a 9.3 issue.

shivanesabharwal:
@headius unfortunately, we are running into issues upgrading the minor version in our build and would appreciate a patch release for 9.3 branch if possible. Is that feasible?

headius:
@shivanesabharwal We would like to help resolve any issues preventing you from upgrading to 9.4, but I will chat with @enebo to see when we can get a 9.3 update released.

PR #7957, opened 2023-10-10
Use new GHA M1 runner

headius:
We would not need our self-hosted runner if this works out.


headius:
Doesn't appear to work in the free tier yet. ☹️

PR #7958, opened 2023-10-10
Prepend internal marker to internal sources

### LINKED ISSUE #6430: Using <internal: for core library methods defined in Ruby
Core library methods' `source_location` and in backtraces defined with Ruby code [look like](https://gist.github.com/eregon/912e6359e83781c5fa1c638d3768c526) this:
* `<internal:kernel>:89` on CRuby
* `<internal:core> core/kernel.rb:533` on TruffleRuby
* `uri:classloader:/jruby/kernel/enumerator.rb:243` on JRuby

I would suggest to use a `<internal:` prefix since this is becoming the de facto standard.
That will also provide a consistent way to filter out core library methods defined in Ruby when desirable.

Also note that both CRuby 3 and TruffleRuby ignore entries starting with `<internal:` for `Kernel#warn(msg, uplevel: n)` (details in https://github.com/rubygems/rubygems/pull/3987#issuecomment-704334889).

headius:
This allows tools to omit any of our internal Ruby-based kernel sources when processing backtraces and source_locations. The filtering here is only done in source_location and backtrace generation currently, with some extra logic during backtrace mining to skip internal frames for e.g. Kernel#warn.

I only did this filtering at the highest levels, when the filename is actually about to be used and returned to a Ruby consumer, to avoid any internal usage of these filenames breaking due to the new prefix.

Fixes #6430

### DIFF
--- a/core/src/main/java/org/jruby/runtime/backtrace/TraceType.java
+++ b/core/src/main/java/org/jruby/runtime/backtrace/TraceType.java
@@ -603,16 +603,23 @@ public class TraceType {
     }
 
     public static String maskInternalFiles(String filename) {
-        if (isInternal(filename)) {
+        if (isInternalFile(filename)) {
             return "<internal:" + filename + ">";
         }
 
         return filename;
     }
 
-    public static boolean isInternal(String filename) {
-        return filename.startsWith("uri:classloader:/jruby/kernel/") ||
-                filename.startsWith("<internal:");
+    public static boolean isInternalFile(String filename) {
+        return filename.startsWith("uri:classloader:/jruby/kernel/");
+    }
+
+    public static boolean hasInternalMarker(String filename) {
+        return filename.startsWith("<internal:");
+    }
+
+    public static boolean isExcludedInternal(String filename) {
+        return TraceType.isInternalFile(filename) || TraceType.hasInternalMarker(filename);
     }
 
     @Deprecated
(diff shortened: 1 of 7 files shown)


headius:
Needs bake time before release, so I bumped it to 9.4.5.0.

core/src/main/java/org/jruby/runtime/backtrace/TraceType.java
```diff
@@ -581,6 +602,19 @@ public static IRubyObject generateMRIBacktrace(Ruby runtime, RubyStackTraceEleme
...
+        if (isInternal(filename)) {
+            return "<internal:" + filename + ">";
+        }
+
+        return filename;
+    }
+
+    public static boolean isInternal(String filename) {
+        return filename.startsWith("uri:classloader:/jruby/kernel/") ||
+                filename.startsWith("<internal:");
```
kares, line 615: seems like the above could than theoretically double mask as `"<internal:<internal:" + ...` if the check is present in `maskInternalFiles` and a masked `filename` gets passed...

headius: You're right... I think this code evolved as I went and became general "is this source file internal" check. I will look into it and figure out why I did this.

headius: @kares I believe my last commit fixes and clarifies this logic.

Issue #6346, opened 2020-07-28
TCPSocket hangs when RST is received on Windows


yamam:
**Environment Information**

jruby 9.2.12.0 (2.5.7) 2020-07-01 db01a49ba6 Java HotSpot(TM) Client VM 25.161-b12 on 1.8.0_161-b12 +jit [mswin32-i386]

**Expected Behavior**

server.rb
```ruby
require 'socket'
sock = TCPServer.new(nil, 12345).accept
while line = sock.gets
    p line
end
```

client.rb
```ruby
require 'socket'
s = TCPSocket.open('localhost', 12345)
s.puts("hello")
sleep 1
# send RST on close
linger = [1,0].pack('ii')
s.setsockopt(Socket::SOL_SOCKET, Socket::SO_LINGER, linger)
s.close
```

$ ruby server.rb
"hello\n"
Traceback (most recent call last):
        1: from server.rb:3:in `<main>'
server.rb:3:in `gets': An existing connection was forcibly closed by the remote host. @ io_fillbuf - fd:4  (Errno::ECONNRESET)

$ ruby client.rb

**Actual Behavior**

$ c:\jruby-9.2.12.0\bin\jruby.exe server.rb
"hello\n"

The program hangs and CPU utilization is high.

headius:
Raises the error on Darwin but message could be improved:

```
"hello\n"
Errno::ECONNRESET: Connection reset by peer - channel: org.jruby.util.io.ChannelFD@38364841
    gets at org/jruby/RubyIO.java:2318
  <main> at server.rb:3
```

The code in this area does not differ significantly on Windows, so this might be a behavioral difference at the JDK level.

rdp:
Seems like it's generating exception internally jconsole main thread:

```

Stack trace: 
java.base@20.0.2/java.lang.Throwable.fillInStackTrace(Native Method)
java.base@20.0.2/java.lang.Throwable.fillInStackTrace(Throwable.java:820)
   - locked java.net.SocketException@5687ca66
java.base@20.0.2/java.lang.Throwable.<init>(Throwable.java:273)
java.base@20.0.2/java.lang.Exception.<init>(Exception.java:67)
java.base@20.0.2/java.io.IOException.<init>(IOException.java:57)
java.base@20.0.2/java.net.SocketException.<init>(SocketException.java:47)
java.base@20.0.2/sun.nio.ch.SocketChannelImpl.throwConnectionReset(SocketChannelImpl.java:401)
[...]

PR #7967, opened 2023-10-18
Propagate read errors in IO backend

### LINKED ISSUE #7961: Socket.each loops after receiving RST pegging CPU at 100%
**Environment Information**
  
JRuby versions tested:

* jruby 9.2.20.1 (2.5.8) 2021-11-30 2a2962fbd1 OpenJDK 64-Bit Server VM 17.0.4.1+1 on 17.0.4.1+1 +jit [darwin-aarch64]
* jruby 9.3.11.0 (2.6.8) 2023-09-11 cf0bf1a786 OpenJDK 64-Bit Server VM 17.0.4.1+1 on 17.0.4.1+1 +jit [arm64-darwin]
* jruby 9.4.4.0-SNAPSHOT (3.1.4) 2023-10-11 1ae2b687b5 OpenJDK 64-Bit Server VM 17.0.4.1+1 on 17.0.4.1+1 +jit [arm64-darwin]

* Operating system and platform 

Darwin arkham.local 22.6.0 Darwin Kernel Version 22.6.0: Fri Sep 15 13:41:28 PDT 2023; root:xnu-8796.141.3.700.8~1/RELEASE_ARM64_T6000 arm64

**Expected Behavior**

* Describe your expectation of how JRuby should behave:

MRI will raise Errno::ECONNRESET for each connection that is closed. Cpu stays at ~0%.

* Provide an executable Ruby script: 

```ruby
require 'socket'

HOST = 'localhost'
PORT = 5555

def connect_and_close
  socket = TCPSocket.new(HOST, PORT)
  linger = [1,0].pack('ii')
  socket.setsockopt(Socket::SOL_SOCKET, Socket::SO_LINGER, linger)
  socket.close
end

def tcp_receiver(socket)
  socket.each { |line| puts line }
rescue Errno::ECONNRESET
  puts "connection reset"
end

server_thread = Thread.new do
  server_socket = TCPServer.new(HOST, PORT)
  loop do
    socket = server_socket.accept

    Thread.new(socket) do |socket|
      tcp_receiver(socket)
    end
  end
end

sleep 1
10.times { connect_and_close }

server_thread.join
```

**Actual Behavior**

* Describe or show the actual behavior

All JRuby versions that I tested will loop, raising CPU usage by about 8-10% per thread handling a connection. With 10 connections CPU stays pegged at nearly 100% on my machine.

I thought this would be related to https://github.com/jruby/jruby/issues/7805, but I've tested with a recent 9.4.4.0 snapshot and 9.3.11.0, and the issue still happens.

headius:
The JRuby IO subsystem simulates a POSIX environment by capturing Java exceptions and translating them to their equivalent errno values. When a read produces an error, we return -1 and set the local shim's errno. This can then be used by POSIX-like consumers of the API to process the error without exception-handling.

However in the case of IO read operations, we were not appropriately checking the errno result for a failed (i.e. -1) read. Instead we assumed it was always an EAGAIN-like result and proceeded to select for read on the socket. Under normal circumstances, like the socket being fully closed on both ends, that select would raise an error anyway so no errors were lost.

In the example from #7961 (and likely the case in #6346), only one side of the socket was closed using SO_LINGER=0, causing an RST packet to be returned to future reads, manifested by the `read` operation as a SocketException("Connection reset"). The logic described above then proceeded directly to a `select` operation, which did not in this case raise any error. Instead, `select` reported that the socket was still readable, so we loop back to the `read` ad infinatum.

The changes here make three changes:

* Properly handle IOExceptions with a message of "Connection reset" by producing an ECONNRESET errno. The translation logic from Java exceptions to errnos was also cleaned up a bit.
* When `read` returns -1, only proceed to selection if the errno result was `EAGAIN`/`EWOULDBLOCK`, otherwise raising the appropriate error.
* An unrelated change to avoid using an iterator to check `select` results.

Fixes #7961 and #6346.

### DIFF
--- a/core/src/main/java/org/jruby/util/io/OpenFile.java
+++ b/core/src/main/java/org/jruby/util/io/OpenFile.java
@@ -1298,10 +1298,12 @@ public class OpenFile implements Finalizable {
                     r = readInternal(context, this, fd, rbuf.ptr, 0, rbuf.capa);
 
                     if (r < 0) {
-                        if (waitReadable(context, fd)) {
+                        Errno errno = posix.getErrno();
+                        if (errno == Errno.EAGAIN || errno == Errno.EWOULDBLOCK
+                                && waitReadable(context, fd)) {
                             continue retry;
                         }
-                        throw context.runtime.newErrnoFromErrno(posix.getErrno(), "channel: " + fd + (pathv != null ? " " + pathv : ""));
+                        throw context.runtime.newErrnoFromErrno(errno, "channel: " + fd + (pathv != null ? " " + pathv : ""));
                     }
                     break;
                 }
--- a/core/src/main/java/org/jruby/RubyThread.java
+++ b/core/src/main/java/org/jruby/RubyThread.java
@@ -2222,7 +2222,7 @@ public class RubyThread extends RubyObject implements ExecutionContext {
                         if (result == 1) {
                             Set<SelectionKey> keySet = currentSelector.selectedKeys();
 
-                            if (keySet.iterator().next() == key) {
+                            if (keySet.contains(key) && key.isValid()) {
                                 return true;
                             }
                         }
--- a/core/src/main/java/org/jruby/runtime/Helpers.java
+++ b/core/src/main/java/org/jruby/runtime/Helpers.java
@@ -10,6 +10,7 @@ import java.lang.reflect.Array;
 
 import java.net.BindException;
 import java.net.PortUnreachableException;
+import java.net.SocketException;
 import java.nio.channels.ClosedChannelException;
 import java.nio.channels.NonReadableChannelException;
 import java.nio.channels.NonWritableChannelException;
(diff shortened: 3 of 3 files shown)


headius:
Risks from this PR:

* We will only proceed to `waitReadable` if errno is EAGAIN/EWOULDBLOCK, which may not cover other cases that could be selected and read again. I do not know of such a case offhand.
* I could have screwed up the errno translation, but most of the refactor was done by IntelliJ.

Other notes:

* I included a small change here to use Set.contains rather than create an Iterator to get the sole contained value. We can further improve this by moving to Java 11 and using Selector.select(Consumer<SelectionKey>) which will eliminate the Set as well.

core/src/main/java/org/jruby/util/io/OpenFile.java
```diff
@@ -1298,7 +1298,8 @@ public int fillbuf(ThreadContext context) {
                     r = readInternal(context, this, fd, rbuf.ptr, 0, rbuf.capa);
 
                     if (r < 0) {
-                        if (waitReadable(context, fd)) {
+                        if (posix.getErrno() == Errno.EAGAIN || posix.getErrno() == Errno.EWOULDBLOCK
+                                && waitReadable(context, fd)) {
```
enebo, line 1302: will this do two foreign calls for errno value?

headius: In this case it's not a foreign call; `posix` is a shim class that hides native vs JDK but simulates the posix style. It should be retrieved once, though.

Issue #7968, opened 2023-10-18
RbConfig::CONFIG["SOEXT"] is missing on JRuby


eregon:
**Environment Information**
  
Provide at least:
* JRuby version (`jruby -v`) and command line (flags, JRUBY_OPTS, etc)
`jruby 9.4.3.0 (3.1.4) 2023-06-07 3086960792 OpenJDK 64-Bit Server VM 17.0.8+7 on 17.0.8+7 +jit [x86_64-linux]`
* Operating system and platform (e.g. `uname -a`)

**Expected Behavior**

It should work like on CRuby and TruffleRuby e.g. by returning `so` on Linux and `dylib` on Darwin and `dll` on Windows.

SOEXT is relevant e.g. for FFI.

**Actual Behavior**

It's nil:
```
$ jruby -e 'p RbConfig::CONFIG["SOEXT"]'
nil
```

What is very surprising to me is https://github.com/ruby/prism/blob/bb35e38000cfec301061e1f6742e257eff238c64/lib/prism/ffi.rb#L17 seems to work on JRuby in CI, but does not for me locally.
Maybe FFI/dlopen() accepts to load `"/path/to/librubyparser."` and appends the right extension?

eregon:
Ah it "works" in Prism's CI because `bundle exec rake` on JRuby creates a file named `librubyparser.` since SOEXT is nil.
That's not really how it's supposed to work though.

enebo:
Unfortunately, we are in process of releasing right now so this will wait but I suppose even if weird it at least works.  Next version people would need to reinstall to get it working again but I am not sure we have any prism users yet so we may fix this before that point.

ahorek:
you can alternatively use `FFI::Platform::LIBSUFFIX` which would be my preferred option here:
`ffi_lib File.expand_path("../../build/librubyparser.#{RbConfig::CONFIG["SOEXT"]}", __dir__)`

eregon:
@ahorek I thought about that, but it doesn't help for other places in prism which need it, including the Makefile.

PR #7969, opened 2023-10-18
add RbConfig::CONFIG["SOEXT"]

ahorek:
fixes https://github.com/jruby/jruby/issues/7968


enebo:
I will chance it and land it before CI finishes since I think it is fairly obvious nothing covers this that we run.  Thanks @ahorek !  :rocket:

PR #7973, opened 2023-10-20
Use gem for tempfile

headius:
It would be one less library to maintain if we just use the standard tempfile gem, but we would lose the benefit of our own Tempfile that does not use delegate.rb to wrap an IO (ours extends IO itself). This PR is to try it out and discuss.

See https://github.com/ruby/tempfile/issues/7.


headius:
Remaining failure is due to a fix for https://bugs.ruby-lang.org/issues/11015 applied to CRuby in ruby/ruby@efbfd9027282e6f82f41edd32a1732d4a07e90a3. We'll need something similar in `IO.copy_stream`.

headius:
Rebased to 9.5-dev since this will involve a fairly major change to a standard library, plus the recent logic in `IO.copy_stream` requires changes from 3.2.0 (https://github.com/ruby/ruby/pull/6867).

PR #7974, opened 2023-10-22
[ji] support java_alias with constructor

kares:
the use-case, as demonstrated in the spec, is to be able to bypass constructor lookup with:

`java_alias :_java_ctor_alias, :'<init>'` the reasoning is similar to using [`java_alias` with overloads](https://github.com/jruby/jruby/wiki/ImprovingJavaIntegrationPerformance)

---

also, there's a bunch of refactoring/cleanup commits:  
- avoid constructor lookup when using `setClassAllocator(Class)`
- set-up Java causes for (Java) allocator failing exceptions


headius:
We want to push out 9.4.5.0 this week so I think this should move to 9.4.6.0 so it has time to bake on master.

PR #7977, opened 2023-10-24
Prototype of fast block_given? logic using indy

headius:
The basic idea here is to avoid pushing a frame for block_given? when it is known to be our core implementation. Instead, we use indy to test the block the current method was called with and return a result directly. When the block_given? is not our core impl, it falls back on a normal indy invocation. This allows us to remove the frame requirement from block_given? callers, and the cost of the block_given? call largely disappears.

Caveats:

* This is very much hacked in place, with duplicated code in multiple places.
* In order to pass through the block we were given, we add a Block parameter to the call site. If someone defines a block_given? in Ruby, it would receive our given block.


headius:
Superseded by #8170.

PR #7985, opened 2023-10-26
Make sure gems path is regex-safe

### LINKED ISSUE #7983: Build, test failures when build directory contains special characters
On Debian, the build directory names often contain the package name with the version number appended. That version number may contain a `+` character, which has special meaning in a regular expression or URI.

When building JRuby in a directory whose tree contains a plus sign (`+`), bad things happen:

- The rubygem bin stubs are not copied in the right directory:

```
--- gem racc-1.6.0 ---                                                                                    
copy gem content to /jruby+ds/lib/ruby/stdlib                                                             
 copying: /jruby+ds/lib/target/rubygems/gems/racc-1.6.0-java/lib/racc to /jruby+ds/lib/ruby/stdlib        
 copying: /jruby+ds/lib/target/rubygems/gems/racc-1.6.0-java/lib/racc.rb to /jruby+ds/lib/ruby/stdlib                                                                                                                
copy executable /jruby+ds/lib/target/rubygems/bin/racc to /jruby+ds/bin                                                                                                                                              
copy racc to /jruby+ds/lib/ruby/gems/shared/gems/jruby+ds/lib/target/rubygems/gems/racc-1.6.0-java/bin/racc                                                                                                          
copy to specifications/default: racc-1.6.0-java.gemspec
```
This can be fixed in `lib/pom.rb` by replacing `source.sub( /#{gems}/, '')` with `source.sub( gems, '')`

- `JarResourceTest#testCreateJarResourceWithSpaceCharInPath` fails

This can be worked around by replacing `+` with `%2b` in the resource path.

headius:
Fixes #7983

### DIFF
--- a/lib/pom.rb
+++ b/lib/pom.rb
@@ -339,7 +339,7 @@ project 'JRuby Lib Setup' do
         spec.executables.each do |f|
           bin = Dir.glob(File.join( gems, "#{gem_name}*", spec.bindir ))[0]
           source = File.join( bin, f )
-          target = File.join( bin_stubs, source.sub( /#{gems}/, '' ) )
+          target = File.join( bin_stubs, source.sub( gems, '' ) )
           log "copy #{f} to #{target}"
           FileUtils.mkdir_p( File.dirname( target ) )
           FileUtils.cp_r( source, target )


jcharaoui:
I've tested this and it doesn't work:

```
--- gem racc-1.6.0 ---                                                                                    
copy gem content to /<<PKGBUILDDIR>>/lib/ruby/stdlib                                                      
 copying: /<<PKGBUILDDIR>>/lib/target/rubygems/gems/racc-1.6.0-java/lib/racc to /<<PKGBUILDDIR>>/lib/ruby/stdlib
 copying: /<<PKGBUILDDIR>>/lib/target/rubygems/gems/racc-1.6.0-java/lib/racc.rb to /<<PKGBUILDDIR>>/lib/ruby/stdlib
copy executable /<<PKGBUILDDIR>>/lib/target/rubygems/bin/racc to /<<PKGBUILDDIR>>/bin          
copy racc to /<<PKGBUILDDIR>>/lib/ruby/gems/shared/gems/<<PKGBUILDDIR>>/lib/target/rubygems/gems/racc-1.6.0-java/bin/racc
copy to specifications/default: racc-1.6.0-java.gemspec
```

We shouldn't see the `<<PKGBUILDDIR>>/lib/target/rubygems` component in the binstub path.

enebo:
Seems like it is quoting more than we bargained for

headius:
After another look at this I think the original proposed change makes more sense anyway: replace the regex with the `gems` string directly. If we are essentially trying to quote the entire contents of a regex, there's no reason to use a regex. I'll push your proposed fix instead.

headius:
@jcharaoui I assume since this was your proposed fix that this PR is now working for you, yes?

jcharaoui:
Right, this works for me!

PR #7990, opened 2023-10-27
[fix] infinite loop calling Java super from Ruby

kares:
introduced in 9.4.4.0 while doing a variant of: https://github.com/jruby/jruby/commit/bd9d40c02a1a693b46aa14e5feac7d64b5442246#diff-d269af154b613900dfe27e6c544a8c6748bf52af91e74395df49c70764db1478R1841

managed to miss copy-ing the loop last step (and I was pretty confident that is a tested code path)


jimtng:
Thanks for the fix! I've tested it and can confirm that this solved the problem I was having.

headius:
Looks good to me!

kares:
unfortunately, this also made it to 9.3.11.0 - thus got back-ported at https://github.com/jruby/jruby/pull/7991

PR #8007, opened 2023-11-06
Fix intermittent fails in CI

headius:
This will contain small fixes to get our CI more reliable, by fixing the causes of intermittent failures.

* `IO#read raises IOError when stream is closed by another thread` intermittent due to different ways of handling a nulled-out "fd". All cases where "fd" is null when it was previously confirmed to be non-null should indicate parallel close occurred. (Better state management here would be helpful, to ensure we **know** it only could have been closed in parallel.

### DIFF
--- a/core/src/main/java/org/jruby/Ruby.java
+++ b/core/src/main/java/org/jruby/Ruby.java
@@ -4209,6 +4209,16 @@ public final class Ruby implements Constantizable {
         return RubySystemExit.newInstance(this, status, message).toThrowable();
     }
 
+    /**
+     * Prepare a throwable IOError with the given message.
+     *
+     * This constructor should not be used to create a RubyException object to be raised on a different thread.
+     *
+     * @see RaiseException#from(Ruby, RubyClass, String)
+     *
+     * @param message the message for the new IOError
+     * @return a fully-prepared IOError throwable
+     */
     public RaiseException newIOError(String message) {
         return newRaiseException(getIOError(), message);
     }
@@ -4312,10 +4322,13 @@ public final class Ruby implements Constantizable {
      *
      * There are additional forms of this construction logic in {@link RaiseException#from}.
      *
+     * This constructor should not be used to create a RubyException object to be raised on a different thread.
+     *
+     * @see RaiseException#from(Ruby, RubyClass, String)
+     *
      * @param exceptionClass the exception class from which to construct the exception object
      * @param message a simple message for the exception
      * @return a new RaiseException wrapping a new Ruby exception
-     * @see RaiseException#from(Ruby, RubyClass, String)
      */
     public RaiseException newRaiseException(RubyClass exceptionClass, String message) {
         IRubyObject cause = getCurrentContext().getErrorInfo();
--- a/core/src/main/java/org/jruby/RubyIOError.java
+++ b/core/src/main/java/org/jruby/RubyIOError.java
@@ -50,4 +50,8 @@ public class RubyIOError extends RubyStandardError {
     protected RaiseException constructThrowable(String message) {
         return new IOError(message, this);
     }
+
+    public static RubyIOError newIOError(Ruby runtime, String message) {
+        return ((RubyIOError) runtime.getIOError().newInstance(runtime.getCurrentContext(), RubyString.newString(runtime, message)));
+    }
 }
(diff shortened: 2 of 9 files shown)


headius:
```
TestIO#test_read_unlocktmp_ensure:
Errno::EBADF: Bad file descriptor - No message available
    org/jruby/RubyIO.java:2196:in `close'
    org/jruby/RubyIO.java:[45](https://github.com/jruby/jruby/actions/runs/6748586512/job/18405119095#step:4:46)04:in `pipe'
    /home/runner/work/jruby/jruby/test/mri/ruby/test_io.rb:3592:in `test_read_unlocktmp_ensure'
```

headius:
```
Failure: test_cleanup(TestTempfileCleanup)
D:/a/jruby/jruby/test/jruby/test_tempfile_cleanup.rb:38:in `test_cleanup'
     35: 
     36:     tmp_files = Dir["#{@tmpdir}/*"]
     37:     # test that the files are gone
  => 38:     assert_equal 0, tmp_files.size, "Files were not cleaned up: (#{tmp_files.size}) #{tmp_files}"
     39:   end
     40: end
```

headius:
```
TestIO#test_readpartial_buffer_error:
Errno::EBADF: Bad file descriptor - No message available
    org/jruby/RubyIO.java:2196:in `close'
    /home/runner/work/jruby/jruby/test/mri/ruby/test_io.rb:65:in `with_pipe'
    /home/runner/work/jruby/jruby/test/mri/ruby/test_io.rb:1505:in `test_readpartial_buffer_error'
```

Similar to `test_read_unlocktmp_ensure` in that a pipe IO appears to be closed already by the time we try to close it (either manually or via `IO.pipe`'s block form

headius:
```
Example took longer than the configured timeout of 120.0s:
Signal.trap accepts a proc
```

This one appears to be getting stuck, perhaps losing a signal somewhere and failing to trigger. When it fails, mspec dumps all thread stack traces which usually includes many unfinalized threads and fibers, making this difficult to find in the output.

headius:
Most recent commit [8722fd2](https://github.com/jruby/jruby/pull/8007/commits/8722fd2ff6db6cf699dd8b1dfbe17bcb55cb7e3a) modifies the read-select logic in IO to that if select fails (returns false, possibly a sporadic wake-up or error), we re-try it rather than ignoring the result and proceeding into a blocking read. If we do not do this, we may end up in an NIO blocking read that closes the stream on interrupt. I suspect this is the cause of these unexplained EBADF; we close an in-progress NIO read, which causes the stream to be closed, and then subsequent IO operations see ChannelClosedException and raise it as EBADF.

headius:
The signal handler issue seems to be due to us re-setting the handler to the same handler, as recommended in many guides on using unix signal handling. Unfortunately, we set it in a `finally` block in the handler logic, which meant that it might overwrite any new handler update that happens before this handler has finished executing responding to the signal. The fix is to move the re-set to the top of the handler, to avoid stepping on any code that might change the handler while we execute.

See https://github.com/jruby/jruby/pull/8007#issuecomment-1799939394 for the problem discussion and example script.

headius:
See https://www.cs.kent.edu/~ruttan/sysprog/lectures/signals.html for a discussion about signal handling at the C level, which shows the re-set of the handler as the first thing the handler does.

headius:
This `ErlangActor` sporadic failure appears to also have been noticed on TruffleRuby and they added skips for it.

```
  1) Concurrent ErlangActor on pool behaves like erlang actor asking timing out
     Failure/Error: raise NoActor.new(@Pid) if @Terminated.resolved?

     Concurrent::ErlangActor::NoActor:
       #<Concurrent::ErlangActor::Pid:0x7f58d87f terminated normally with true>
     Shared Example Group: "erlang actor" called from ./spec/concurrent/edge/erlang_actor_spec.rb:975
     # ./lib/concurrent-ruby-edge/concurrent/edge/erlang_actor.rb:706:in `ask'
     # ./lib/concurrent-ruby-edge/concurrent/edge/erlang_actor.rb:82:in `ask'
     # ./spec/concurrent/edge/erlang_actor_spec.rb:934:in `block in <main>'
     # org/jruby/RubyBasicObject.java:2614:in `instance_exec'
```

https://github.com/ruby-concurrency/concurrent-ruby/pull/943/files#diff-80d88e9530950f5a6b2198da6c47ed17d0cf98e088412359936ce86bc2998890R920

I am adding the same skip and some improvements to another sporadic failure in https://github.com/ruby-concurrency/concurrent-ruby/pull/1012.

PR #8009, opened 2023-11-10
Expand WeakMap to allow floats and fixnums

### LINKED ISSUE #7862: Ruby 2.7 conformance: ObjectSpace::WeakMap is broken for FixNum keys
* JRuby version: Reproducible on 9.4.3.0 release

**Expected Behavior**

Per https://github.com/rubyreferences/rubychanges/blob/master/2.7.md#objectspaceweakmap-now-accepts-non-gc-able-objects as of Ruby 2.7, ObjectSpace::WeakMap should be able to accept numeric keys.

**Actual Behavior**

Accessor functions like `#key?` and `#[]` behave as though a numeric key is not present when queried, except if the instance of the key that appears in `#keys` is used.

```ruby
foo = Object.new
m=ObjectSpace::WeakMap.new
original_key = 105553124374448
m[original_key] = foo
puts "m.key?(105553124374448) #{m.key? 105553124374448}"
puts m.inspect
puts m.keys.inspect
puts m.values.inspect
puts "m.key?(m.keys.first) #{m.key? m.keys.first}"
puts "m.key?(original_key) #{m.key? original_key}"
```

Outputs:
```
m.key?(105553124374448) false
#<ObjectSpace::WeakMap:0x799f354a: 105553124374448 => #<Object:0x45e140ae>>
Keys [105553124374448]
Values [#<Object:0x45e140ae>]
m.key?(m.keys.first) true
m.key?(original_key) false
```

After investigation, the root cause seems to be that the underlying backing map is `java.util.IdentityHashMap`, which uses `System.identityHashCode(Object))` to compare keys. `System.identityHashCode(Object))` does not return identical values for two `org.jruby.RubyFixnum` instances that have identical values for `IRubyObject.id()` or `Object.hashCode()`.

headius:
They will use value comparisons rather than identity comparisons to mimic CRuby's behavior for immediate values.

Fixes #7862

### DIFF
--- a/core/src/main/java/org/jruby/RubyObjectSpace.java
+++ b/core/src/main/java/org/jruby/RubyObjectSpace.java
@@ -37,6 +37,7 @@ import static org.jruby.RubyEnumerator.enumeratorize;
 
 import java.util.Iterator;
 import java.util.Map;
+import java.util.stream.Stream;
 
 import org.jruby.anno.JRubyMethod;
 import org.jruby.anno.JRubyModule;
@@ -51,6 +52,7 @@ import org.jruby.runtime.builtin.IRubyObject;
 import org.jruby.util.Inspector;
 import org.jruby.util.Numeric;
 import org.jruby.util.collections.WeakValuedIdentityMap;
+import org.jruby.util.collections.WeakValuedMap;
 
 @JRubyModule(name="ObjectSpace")
 public class RubyObjectSpace {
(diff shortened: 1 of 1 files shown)


enebo (approved):

JasonLunn:
Thank you for circling back to this issue. Is the right way to enhance the spec to add it the spec in this PR, or should a separate PR go to http://github.com/ruby/spec ?

headius:
@JasonLunn either way works! Commits to the specs in the JRuby repository eventually merge back to the main spec repository and vice versa. If you want to submit a pull request to my repository for this branch, I can merge it in and we can confirm it works in CI.

core/src/main/java/org/jruby/RubyObjectSpace.java
```diff
@@ -231,59 +233,74 @@ public WeakMap(Ruby runtime, RubyClass cls) {
...
         public IRubyObject op_aref(ThreadContext context, IRubyObject key) {
-            IRubyObject value = map.get(key);
+            Map<IRubyObject, IRubyObject> weakMap = getWeakMapFor(key);
+            IRubyObject value = weakMap.get(key);
             if (value != null) return value;
             return context.nil;
         }
 
+        private Map<IRubyObject, IRubyObject> getWeakMapFor(IRubyObject key) {
+            if (key instanceof RubyFixnum || key instanceof RubyFloat) {
```
kares, line 243: is it on purpose that we do not handle `RubyInteger`  but only `RubyFixnum` values?

kares (approved):

core/src/main/java/org/jruby/RubyObjectSpace.java
headius: Even in CRuby, integer values outside Fixnum range are normal objects.

Issue #5933, opened 2019-10-22
early return from a block is slow


ahorek:
### Environment
jruby 9.2.9.0-SNAPSHOT (2.5.7) 2019-10-22 c3171f5 Java HotSpot(TM) 64-Bit Server VM 11.0.2+9-LTS on 11.0.2+9-LTS +jit [linux-x86_64]

refs https://github.com/jruby/jruby/pull/5928

I've investigated why is String#unpack1 so slow
https://github.com/jruby/jruby/blob/d51c0ca7a3f46691ae27e2c00c2411b8894b0da9/core/src/main/ruby/jruby/kernel/string.rb#L2

using an early return inside a block is a very expensive operation on jruby. (13x! slower on graal)

```
jruby (9.2.9.0 dev)
 unpack early return    700.842k (± 7.8%) i/s -      3.485M in   5.012230s
        unpack block      2.289M (± 9.2%) i/s -     11.253M in   4.992273s
        unpack first      2.853M (± 5.8%) i/s -     14.189M in   4.994309s

jruby + indy (9.2.9.0 dev)
 unpack early return    923.728k (±10.2%) i/s -      4.545M in   4.987390s
        unpack block      2.914M (± 9.3%) i/s -     14.266M in   4.985981s
        unpack first      3.770M (± 7.0%) i/s -     18.699M in   4.991746s

jruby + indy + graal (9.2.9.0 dev)
 unpack early return    348.480k (±10.8%) i/s -      1.720M in   5.013014s
        unpack block      2.973M (±10.6%) i/s -     14.613M in   4.995835s
        unpack first      4.495M (± 7.3%) i/s -     22.248M in   4.990273s

ruby (2.5.3)
 unpack early return      2.212M (± 9.2%) i/s -     10.946M in   5.000249s
        unpack block      2.688M (± 5.4%) i/s -     13.407M in   5.007471s
        unpack first      2.868M (± 4.1%) i/s -     14.313M in   5.001607s

ruby (2.7 dev)
 unpack early return      2.698M (± 7.2%) i/s -     13.406M in   5.008788s
        unpack block      3.253M (± 5.6%) i/s -     16.222M in   5.010958s
        unpack first      2.974M (± 5.1%) i/s -     14.827M in   5.007237s

ruby + jit (2.7 dev)
 unpack early return      2.783M (± 5.3%) i/s -     13.902M in   5.014952s
        unpack block      3.941M (± 9.2%) i/s -     19.537M in   5.007289s
[...]

Issue #1980, opened 2014-09-17
IR return issue with method-level ensure block


nirvdrum:
I've been having difficulties reducing this any further, so my reproduction case requires having redis installed and the `connection_pool` and `redis` gems installed.  When I run:

```
require 'connection_pool'
require 'redis'

POOL = ConnectionPool.new { Redis.new }

def blah
  return [:whatever].each { return 3 } 
ensure
  POOL.with { }
end

x = blah
puts x
```

I get:

```
IRReturnJump.java:25:in `create': IRReturnJump:<Static Type[1913113425]: local [timeout=5,deadline=2014-09-16 20:42:45 -0400,%block=]:#<Redis:0x6892cc6f>>
    from IRRuntimeHelpers.java:97:in `initiateNonLocalReturn'
    from Interpreter.java:406:in `processReturnOp'
    from Interpreter.java:567:in `interpret'
    from Interpreter.java:659:in `INTERPRET_BLOCK'
    from InterpretedIRBlockBody.java:47:in `commonYieldPath'
    from IRBlockBody.java:137:in `doYield'
    from BlockBody.java:82:in `yield'
    from Block.java:147:in `yield'
    from RubyArray.java:1546:in `each'
    from RubyArray$INVOKER$i$0$0$each.gen:-1:in `call'
    from JavaMethod.java:474:in `call'
    from CachingCallSite.java:286:in `cacheAndCall'
    from CachingCallSite.java:81:in `callBlock'
    from CachingCallSite.java:85:in `call'
    from CallBase.java:413:in `interpret'
    from Interpreter.java:339:in `processCall'
    from Interpreter.java:564:in `interpret'
    from Interpreter.java:673:in `INTERPRET_METHOD'
    from InterpretedIRMethod.java:99:in `call'
    from DynamicMethod.java:202:in `call'
    from DynamicMethod.java:198:in `call'
    from CachingCallSite.java:306:in `cacheAndCall'
    from CachingCallSite.java:136:in `call'
    from Interpreter.java:323:in `processCall'
    from Interpreter.java:564:in `interpret'
    from Interpreter.java:639:in `INTERPRET_ROOT'
    from IRScriptBody.java:112:in `interpret'
    from Interpreter.java:133:in `execute'
    from Interpreter.java:29:in `execute'
    from IRTranslator.java:43:in `execute'
    from Ruby.java:886:in `runInterpreter'
    from Ruby.java:893:in `runInterpreter'
[...]

PR #8031, opened 2023-12-05
Reuse thread-local jump exceptions to avoid construction

headius:
Thread-local branching requires us to raise an exception to unroll the stack past multiple frames. These exceptions should never escape a given thread, and yet we always construct a new object for them. We could improve performance of non-local flow control by using a thread-local exception object (for cases that have state) or a global singleton exception object (for cases that do not).

This PR will attempt to do that.

Relates to performance challenges like #5933

### DIFF
--- a/core/src/main/java/org/jruby/Ruby.java
+++ b/core/src/main/java/org/jruby/Ruby.java
@@ -1302,8 +1302,8 @@ public final class Ruby implements Constantizable {
              * This will capture any return which says it should return to a script scope as the reasonable
              * exit point.  We still raise when jump off point is anything else since that is a bug.
              */
-            if (!ex.methodToReturnFrom.getStaticScope().getIRScope().isScriptScope()) {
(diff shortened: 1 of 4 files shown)


headius:
At least one of the failures shows a problem with this approach: there may be more than one IRReturnJump in flight at a time.

From #1980:

https://github.com/jruby/jruby/blob/6d497bd585c6ce535907205d8285f644235d07c1/spec/regression/GH-1980_multiple_nonlocal_returns_in_flight_spec.rb

In order to avoid reallocating this object, then, we'd need to have some sort of pooling mechanism, or a flag to indicate it is currently in use.

headius:
This was a good experiment but ultimately the overhead of the thread-local appears to match or exceed the overhead of just creating a new object every time. The example in #5933 is no faster, and possibly a bit slower, using the code from this PR. The real culprit in that case is the construction of a useless DynamicScope.

Moving this into a simple ThreadContext field might reduce the overhead of managing a one-element pool, but it would require passing ThreadContext into all the IRReturnJump handling methods, a larger change than I had planned and possibly again more overhead managing the pool than just creating a new object.

PR #8040, opened 2023-12-09
Match `Exception#full_message` with MRI 3.2

ntkme:
In my very limited test, this should matches MRI 3.2's formatting at byte level as long as the backtrace is exactly the same. E.g. rescue then re-raise a different error creates an additional line in backtrace for MRI, but not in jruby - the output will be slightly different in such case.

Why match MRI 3.2 not 3.1? The only difference between 3.1 and 3.2 is how `order:` works. In 3.2, `order` default to `:top`, which is the same as the current jruby behavior. Therefore, matching 3.2 is actually a smaller change than matching 3.1.

### DIFF
--- a/core/src/main/java/org/jruby/Main.java
+++ b/core/src/main/java/org/jruby/Main.java
@@ -558,19 +558,6 @@ public class Main {
         boolean isatty = runtime.getPosix().isatty(FileDescriptor.err);
 
         System.err.print(traceType.printBacktrace(raisedException, isatty));
-
-        Set<Object> shownCauses = new HashSet<Object>();
-        shownCauses.add(raisedException);
-
-        for (
-                Object cause = raisedException.getCause();
-                cause != null && cause instanceof RubyException && !shownCauses.contains(cause);
-                cause = ((RubyException) cause).getCause()) {
-
-            System.err.print(traceType.printBacktrace((RubyException) cause, isatty));
-            shownCauses.add(cause);
-        }
-
         return 1;
     }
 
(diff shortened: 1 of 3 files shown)


ntkme:
This is the limit test I did:

``` sh
diff <(ruby test.rb) <(./bin/jruby test.rb)
diff <(ruby test.rb 2>/dev/null) <(./bin/jruby test.rb 2>/dev/null)
```

Test script:

``` ruby
def recurse(depth, limit)
  if depth < limit
    recurse(depth + 1, limit)
  else
    raise 'limit reached'
  end
end

begin
  recurse(0, 10)
rescue => e
  puts e.full_message
  puts '---'
  puts e.full_message(highlight: nil)
  puts '---'
  puts e.full_message(order: nil)
  puts '---'
  puts e.full_message(highlight: nil, order: nil)
  puts '---'
  puts e.full_message(highlight: nil, order: :top)
  puts '---'
  puts e.full_message(highlight: nil, order: :bottom)
  puts '---'
  puts e.full_message(highlight: false, order: nil)
  puts '---'
  puts e.full_message(highlight: false, order: :top)
  puts '---'
  puts e.full_message(highlight: false, order: :bottom)
  puts '---'
  puts e.full_message(highlight: true, order: nil)
  puts '---'
  puts e.full_message(highlight: true, order: :top)
  puts '---'
  puts e.full_message(highlight: true, order: :bottom)
end
```

enebo:
@ntkme cool.  If someone opens an issue about order being wrong then we can reconsider whether we should change it but the fact it has changed in 3.2 means people should not depend on order default anyways.

PR #8045, opened 2023-12-18
Try to find class resource to avoid exception

headius:
The exception thrown here when we cannot find a class can be very expensive at boot time. For most third-party extensions, this exception will be raised twice for every bound method, due to double-checking the classloader under lock a second time. By first looking for a .class resource, we can avoid the expensive exception if it is unlikely to succeed.


headius:
This is complete. It does not appear to affect a single command-line launch of JRuby significantly, but for repeat launches in the same VM the results improve: for `gem list`, from 0.581s to 0.539s per instance (my M1 system with a few hundred gems).

This number will increase with the number of gem-based extension methods bound. In this case it is primarily from booting stringio and strscan, I believe.

Issue #7845, opened 2023-07-05
Including a module again after prepending another module to it isn't idempotent


janko:
**Environment Information**

```sh
$ jruby -v
jruby 9.4.3.0 (3.1.4) 2023-06-07 3086960792 OpenJDK 64-Bit Server VM 17.0.2+0 on 17.0.2+0 +jit [arm64-darwin]

$ uname -a
Darwin Jankos-MacBook-Air.local 22.5.0 Darwin Kernel Version 22.5.0: Thu Jun  8 22:22:19 PDT 2023; root:xnu-8796.121.3~7/RELEASE_ARM64_T8103 arm64
```

**Expected Behavior**

In [Shrine](https://github.com/shrinerb/shrine), I have specific situation where I can prepend a module twice, but prepend a second module to it in between.

```rb
module Subextension
  def foo
    puts "FIRST"
    super
  end
end

module Extension
  def foo
    puts "SECOND"
    super
  end
end

module Base
  def foo
    puts "THIRD"
  end
end

class Foo
  include Base
end

subclass = Class.new(Foo)
subclass.include Extension

Extension.prepend Subextension
subclass.include Extension

p subclass.ancestors

subclass.new.foo
```

In CRuby, this is still idempotent:

```
[#<Class:0x0000000100703860>, Subextension, Extension, Foo, Base, Object, Kernel, BasicObject]
FIRST
SECOND
THIRD
```

**Actual Behavior**

In JRuby, however, it seems that an extra module gets added into the ancestor change, duplicating calls in the inheritance hierarchy:

```
[#<Class:0x53feeac9>, Subextension, Extension, #<Module:0x70091872>, Foo, Base, Object, Kernel, BasicObject]
FIRST
SECOND
SECOND
THIRD
```

kares:
very interesting, thought it's not meant to be idempotent (or maybe it just hasn't been in older Ruby versions). 
wonder if there are any cases where it's not, probably none ... maybe I am confusing this with singleton class extend.

anyhow thanks for the report, if you discover some non-idempotent cases in CRuby please let us know :wink:

janko:
Yeah, I admit it's a very specific use case, I only try to report any differences I find through CI failures of my gems. FWIW, older CRuby versions also seem to have this behavior, because Shrine's test suite is passing on 2.3+, it's only failing for JRuby in this one place.

PR #8055, opened 2023-12-29
override getDelegate in PrependedModule to return origin

evaniainbrooks:
issue #7845 "Including a module again after prepending another module to it isn't idempotent"


enebo:
@evaniainbrooks outstanding!  Thanks for the PR. It is hilarious that was marked as beginner.  This was not a simple part of the code :smile:

PR #8057, opened 2023-12-30
recognize RubySymbol as assignable to String

### LINKED ISSUE #8015: Java method lookup for arity-1 method with inexact argument fails in the presence of arity-2 method with same name
**Test case:**

Create the following Java class:

```java
package foobar;

public class JRubyArityTest {
    public JRubyArityTest() {}

    public String foo(String s) {
        return "foo(" + s + ")";
    }

    public String bar(String s) {
        return "bar(" + s + ")";
    }

    public String bar(String s, boolean b) {
        return "NOT EXPECTED TO BE CALLED";
    }
}
```

and run the following JRuby code:

```ruby
obj = Java::Foobar::JRubyArityTest.new
puts obj.foo('string')
puts obj.foo(:symbol)
puts obj.bar('string')
puts obj.bar(:symbol)
```

**Expected output:**

```
foo(string)
foo(symbol)
bar(string)
bar(symbol)
```

**Observed output:**

```
foo(string)
foo(symbol)
bar(string)
NameError: no method 'bar' for arguments (org.jruby.RubySymbol) on Java::Foobar::JRubyArityTest
```

Tested on JRuby 9.3.11.0 and 9.4.5.0. However, this appears to be a longstanding issue, since our codebase apparently has old workarounds for this bug dating from 2018.

evaniainbrooks:
resolves #8015

### DIFF
--- a/core/src/main/java/org/jruby/RubySymbol.java
+++ b/core/src/main/java/org/jruby/RubySymbol.java
@@ -1431,6 +1431,11 @@ public class RubySymbol extends RubyObject implements MarshalEncoding, EncodingC
     }
 
     @Override
+    public Class getJavaClass() {
+        return String.class;
+    }
+
+    @Override
     public boolean shouldMarshalEncoding() {
         Encoding enc = getMarshalEncoding();
 
--- a/core/src/main/java/org/jruby/java/dispatch/CallableSelector.java
+++ b/core/src/main/java/org/jruby/java/dispatch/CallableSelector.java
@@ -16,6 +16,7 @@ import org.jruby.RubyFloat;
 import org.jruby.RubyInteger;
 import org.jruby.RubyProc;
 import org.jruby.RubyString;
+import org.jruby.RubySymbol;
 import org.jruby.java.invokers.RubyToJavaInvoker;
 import org.jruby.java.util.ClassUtils;
 import org.jruby.javasupport.JavaCallable;
@@ -700,7 +701,7 @@ public class CallableSelector {
                 return 1;
             }
         }
-        else if ( arg instanceof RubyString ) {
+        else if ( arg instanceof RubyString || arg instanceof RubySymbol ) {
             if ( type == String.class ) return 10;
             if ( type == byte[].class ) return 8;
             if ( CharSequence.class.isAssignableFrom(type) ) return 7;
--- a/core/src/main/java/org/jruby/java/dispatch/CallableSelector.java
+++ b/core/src/main/java/org/jruby/java/dispatch/CallableSelector.java
@@ -701,11 +701,15 @@ public class CallableSelector {
                 return 1;
             }
         }
-        else if ( arg instanceof RubyString || arg instanceof RubySymbol ) {
+        else if ( arg instanceof RubyString ) {
             if ( type == String.class ) return 10;
             if ( type == byte[].class ) return 8;
             if ( CharSequence.class.isAssignableFrom(type) ) return 7;
             if ( type == Character.class ) return 1;
+        }
+        else if ( arg instanceof RubySymbol ) {
+            if ( type == String.class ) return 10;
+            if ( CharSequence.class.isAssignableFrom(type) ) return 7;
         }
         else if ( arg instanceof RubyBoolean ) {
             if ( type == Boolean.class ) return 10;
--- a/test/jruby/test_java_extension.rb
+++ b/test/jruby/test_java_extension.rb
@@ -348,6 +348,17 @@ class TestJavaExtension < Test::Unit::TestCase
     assert ! wrapped1.equal?(wrapped2)
   end
 
+  def test_calling_overridden_method_arity1_with_symbol
+    result = org.jruby.test.TestHelper.testOverriddenMethod(:symbol)
+
+    assert result == "String"
+  end
+
+  def test_calling_overridden_method_arity2_with_symbol
+    result = org.jruby.test.TestHelper.testOverriddenMethod(:symbol, true)
+
+    assert result == "StringBoolean"
+  end
 
   import org.jruby.test.Runner
 
--- a/test/jruby/test_java_extension.rb
+++ b/test/jruby/test_java_extension.rb
@@ -351,13 +351,31 @@ class TestJavaExtension < Test::Unit::TestCase
   def test_calling_overridden_method_arity1_with_symbol
     result = org.jruby.test.TestHelper.testOverriddenMethod(:symbol)
 
-    assert result == "String"
+    assert_equal result, "String"
   end
 
   def test_calling_overridden_method_arity2_with_symbol
     result = org.jruby.test.TestHelper.testOverriddenMethod(:symbol, true)
 
-    assert result == "StringBoolean"
+    assert_equal result, "StringBoolean"
+  end
+
+  def test_calling_overridden_method_arity3_with_symbol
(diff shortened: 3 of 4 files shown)


enebo:
I think this looks good (this is suggestion number 2 in #8015) but I think we want @kares and @headius to consider this change to make sure classifying Symbol as j.l.String has no unexpected consequences. The single arity example show it already set Symbol as String so I am partially confused how it arrives at this behavior without your fix.  There maybe is followup cleanup to remove some code which was somehow doing that?  Not sure.

@evaniainbrooks Not sure if this PR will get looked at until after new years so it may be a few days.

core/src/main/java/org/jruby/java/dispatch/CallableSelector.java
```diff
@@ -700,7 +701,7 @@ private static int calcTypePreference(Class<?> type, final IRubyObject arg) {
                 return 1;
             }
         }
-        else if ( arg instanceof RubyString ) {
+        else if ( arg instanceof RubyString || arg instanceof RubySymbol ) {
```
kares, line 704: This is not a good idea, we should make a separate if case and only handle Java String or maybe CharSequence for RubySymbol.

we do not want to treat it exactly as a RubyString which could convert to a byte sequence.

core/src/main/java/org/jruby/RubySymbol.java
```diff
@@ -1430,6 +1430,11 @@ public static int javaStringHashCode(ByteList iso8859) {
         return h;
     }
 
+    @Override
+    public Class getJavaClass() {
```
kares, line 1434: considered this in the past but always backed-off since it won't auto-convert on a round-trip and IMHO leads to more "inconsistency".

e.g. having a Ruby string key in a hash can be looked up with a Java string and also a Java string value gets converted to RubyString. we can't have the same parity with a symbol so it seemed better to just fail than eventually end up w RubySymbol -> java.lang.String -> RubyString

kares: I think the fix for https://github.com/jruby/jruby/issues/8015 does not need this change, or does it?

if it does we should add more tests covering the changed behavior as the (default) Ruby -> Java (`JavaUtil.convertRubyToJava`) mechanism will end up converting `RubySymbol` -> `java.lang.String` e.g. when seeing a generic `java.lang.Object` (param) type or when doing a generic `call_from_java_a_method_returning_symbol`.

the current behavior is that it passed down `org.jruby.RubySymbol` as is, this could be considered a bug as well as a breaking change (I am fine to consider it a bug atm but more tests on what is changing could reveal concerns).

sorry I haven't looked in detail, hopefully my comments are somehow useful :wink:

evaniainbrooks: Thank you for taking the time to review this. It will take me some time to dig around and digest everything as this is completely new territory for me.

Actually, I think the other change included here is not necessary for this fix. That code path is only hit when the resolution is ambiguous (ambiguous = true in CallableSelector.findMatchingCallableForArgs).

I will have a go at making the other suggested change in #8015 if that is preferred. It seems that may have fewer unintentional side effects.

kares: You were on the spot with the changes, it's just that the JI coverage isn't that good opposed to what's out there in terms of usage. I would have been fine shipping the previous version if we added some more tests, as well.

The new approach seems fine with me, although probably worth revisit later as that special case handling seems a bit "redundant".

We can revisit the approach in another PR with the test included... all in all :+1: job!

core/src/main/java/org/jruby/java/invokers/RubyToJavaInvoker.java
```diff
@@ -451,14 +457,21 @@ protected final T findCallableArityZero(IRubyObject self, String name) {
     protected final T findCallableArityOne(IRubyObject self, String name, IRubyObject arg0) {
         T callable = this.javaCallable;
         if ( callable == null ) {
+            if ( javaCallables[1].length == 1 ) {
+              callable = javaCallables[1][0];
+              if (!callable.isVarArgs()) checkCallableArity(callable, 1);
+              return callable;
+            }
```
kares, line 464: think the logic should be moved, once we're sure `javaCallables[1]` is valid

probably having an `else` statement bellow the 
`if ( javaCallables.length <= 1 || (callablesForArity = javaCallables[1]) == null ) {` check

kares (changes requested):

core/src/main/java/org/jruby/RubySymbol.java
enebo: Wow just a passing comment but it sucks that to generalize all these arities it would force boxing the arguments.  I do see that boxing occurring in deeper calls but generally only on first hit to find an invoker.  I am curious if there is some way of moving logic back into findCallAble* and then reintroducing static methods with more attention to not passing multiple arity signatures?  (Note: definitely not in this PR just for future work).  Not sure if that would lead to simpler/smaller or complicated/larger but it looks like an enticing puzzle.

evaniainbrooks: @kares ok, I am going back to the original implementation. Could you let me know what sort of tests you are looking for, and I will take a crack at them. Am I on the right track with what I have written? Do you want to see something more direct (eg. writing some unit tests for CallableSelector in `core/src/test/java/org/jruby/java/dispatch/CallableSelectorTest.java`?)

Thanks again & sorry for the churn

kares: actually, I am sorry, apparently the `getJavaClass` is not sufficient to cause the behavior I assumed:

```ruby
pp ({ 'foo' => :bar }).to_java.get('foo')
```
I expected your changes to cause this to print a `"bar"` string instead of symbol (current behavior).

... for the record, the test I had in mind would have been smt like:
```diff
diff --git spec/java_integration/types/coercion_spec.rb spec/java_integration/types/coercion_spec.rb
index c15c0df903..bee9e30d6c 100644
--- spec/java_integration/types/coercion_spec.rb
+++ spec/java_integration/types/coercion_spec.rb
@@ -9,7 +9,7 @@ java_import "java_integration.fixtures.PackageConstructor"
 java_import "java_integration.fixtures.ProtectedConstructor"
 java_import "java_integration.fixtures.PrivateConstructor"
 
-describe "Java String and primitive-typed methods" do
+describe "Java coercion" do
   it "should coerce to Ruby types when returned" do
     expect(CoreTypeMethods.getString).to be_kind_of(String)
     expect(CoreTypeMethods.getString).to eq("foo")
@@ -207,6 +207,15 @@ describe "Java Object-typed methods" do
     expect(CoreTypeMethods.getObjectType(1 << 128)).to eq("class java.math.BigInteger")
   end
 
+  it "converts symbols to Java string" do
+    expect(CoreTypeMethods.getObjectType(:foo)).to eq("class java.lang.String")
+  end
+
   it "passes coerced to_java values - keeping the Java type" do
     expect(CoreTypeMethods.getObjectType('foo'.to_java)).to eq "class java.lang.String"
     expect(CoreTypeMethods.getObjectType(0.to_java)).to eq "class java.lang.Long"
```
current behavior is: `"class org.jruby.RubySymbol"`  and that behavior does not change with your changes in the PR. sorry again for the confusion...

kares (approved):
LGTM, well done!

Issue #8029, opened 2023-12-04
Ruby 3.3 support


headius:
# NEWS for Ruby 3.3.0

This document is a list of user-visible feature changes
since the **3.2.0** (see #7517 for 3.2 support) release, except for bug fixes.

Note that each entry is kept to a minimum, see links for details.

## Command line options

* [x] A new `performance` warning category was introduced.
  They are not displayed by default even in verbose mode.
  Turn them on with `-W:performance` or `Warning[:performance] = true`. [[Feature #19538]]

* [ ] A new `RUBY_CRASH_REPORT` environment variable was introduced to allow
  redirecting Ruby crash reports to a file or sub command. See the `BUG REPORT ENVIRONMENT`
  section of the ruby manpage for further details. [[Feature #19790]]

## Core classes updates

Note: We're only listing outstanding class updates.

* Array

    * [x] Array#pack now raises ArgumentError for unknown directives. [[Bug #19150]]

* Dir

    * [ ] Dir.for_fd added for returning a Dir object for the directory specified
      by the provided directory file descriptor. [[Feature #19347]]
    * [ ] Dir.fchdir added for changing the directory to the directory specified
      by the provided directory file descriptor. [[Feature #19347]]
    * [x] Dir#chdir added for changing the directory to the directory specified by
      the provided `Dir` object. [[Feature #19347]]

* MatchData

    * [x] MatchData#named_captures now accepts optional `symbolize_names`
      keyword. [[Feature #19591]]

* Module

    * [x] Module#set_temporary_name added for setting a temporary name for a
      module. [[Feature #19521]]

* ObjectSpace::WeakKeyMap

    * [x] New core class to build collections with weak references.
      The class use equality semantic to lookup keys like a regular hash,
      but it doesn't hold strong references on the keys. [[Feature #18498]]

* ObjectSpace::WeakMap

    * [x] `ObjectSpace::WeakMap#delete` was added to eagerly clear weak map
      entries. [[Feature #19561]]

* Proc
    * [x] Now Proc#dup and Proc#clone call `#initialize_dup` and `#initialize_clone`
      hooks respectively.  [[Feature #19362]]

* Process

    * [x] New `Process.warmup` method that notify the Ruby virtual machine that the boot sequence is finished,
      and that now is a good time to optimize the application. This is useful
      for long-running applications. The actual optimizations performed are entirely
      implementation-specific and may change in the future without notice. [[Feature #18885]]

* Process::Status

[...]

PR #8058, opened 2024-01-01
Ruby 3.3 support: Queue#freeze now raises TypeError

evaniainbrooks:
issue #8029

### DIFF
--- a/core/src/main/java/org/jruby/ext/thread/Queue.java
+++ b/core/src/main/java/org/jruby/ext/thread/Queue.java
@@ -12,7 +12,7 @@
  * rights and limitations under the License.
  *
  * Copyright (C) 2006 MenTaLguY <mental@rydia.net>
- * 
+ *
  * Alternatively, the contents of this file may be used under the terms of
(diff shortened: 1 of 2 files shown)


enebo:
@evaniainbrooks Can you apply this PR against the 9.5-dev branch?  9.4 is 3.1.x Ruby and 9.5 will be 3.3.  Also if this is new 3.3 behavior you should add a version guard like:

```ruby
  ruby_version_is "3.2" do
    it "raises a TypeError if passed a Hash with an unknown key" do
      -> { Struct.new(:animal, { name: 'chris' }) }.should raise_error(TypeError)
    end
  end
```

It also accepts a range so you can do things like ```..3.2```.  At some future point when this becomes our main branch we get our spec changes merged back to ruby/spec so this will save that project some bother of having to add the version guard.  At any time if you feel like adding new specs you can also PR directly to ruby/spec but it will still get there from our PRs at some point.

enebo:
Looks like we have some weird problems on 9.5-dev still but it is relatively new.  Thanks for the fix.  I marked this item as finished in https://github.com/jruby/jruby/issues/8029.

PR #8059, opened 2024-01-02
Ruby 3.3 support: add symbolize_names kwarg to MatchData#named_captures

evaniainbrooks:
issue #8029 

I am unsure if my usage of keywords/checkArity/optional/checkArgumentCount is correct. I couldn't find another example exactly like this in the codebase. It seems I can still call `named_captures(:bogus)` without any complaints so perhaps someone can point me in the right direction.

### DIFF
--- a/core/src/main/java/org/jruby/RubyMatchData.java
+++ b/core/src/main/java/org/jruby/RubyMatchData.java
@@ -857,19 +857,18 @@ public class RubyMatchData extends RubyObject {
         RubyHash hash = RubyHash.newHash(runtime);
         if (regexp == context.nil) return hash;
 
-        IRubyObject symbolizeNamesValue = runtime.getNil();
-
+        boolean symbolizeNames = false;
         if ( argc == 1 ) {
             final IRubyObject opts = args[0];
             if (opts instanceof RubyHash) {
-                symbolizeNamesValue = ArgsUtil.extractKeywordArg(runtime.getCurrentContext(), (RubyHash) opts, "symbolize_names");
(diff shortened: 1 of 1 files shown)


core/src/main/java/org/jruby/RubyMatchData.java
```diff
@@ -846,27 +848,47 @@ public RubyFixnum hash() {
...
+    @JRubyMethod(keywords = true, checkArity = false, optional = 1)
+    public RubyHash named_captures(ThreadContext context, IRubyObject[] args) {
         check();
+
         Ruby runtime = context.runtime;
+        int argc = Arity.checkArgumentCount(runtime, args.length, 0, 0, true);
         RubyHash hash = RubyHash.newHash(runtime);
         if (regexp == context.nil) return hash;
 
+        IRubyObject symbolizeNamesValue = runtime.getNil();
```
enebo, line 860: I think this is probably better as a boolean and lower perform the isTrue when processing the option.

enebo (changes requested):
The argument processing stuff has always been a bit tedious in JRuby.  The code itself is not that complicated but it feels like we could be doing this better.  Keywords in particular really need a rethink for native methods.

Just change the two things and this is in.

enebo:
@evaniainbrooks I do not see my second message about argumenterror if optional arg is not a hash.  Do you see that comment?  I do not seem to be able to see the actual review,

evaniainbrooks:
> @evaniainbrooks I do not see my second message about argumenterror if optional arg is not a hash.  Do you see that comment?  I do not seem to be able to see the actual review,

I did not see it either, but no worries, I will add this now.

enebo:
Thanks!   CI is pretty weird atm.  I am trying to update the parser now too so we can at least pass more.

PR #8067, opened 2024-01-19
Remove the nkf extension in preparation for the gem

headius:
JRuby's implementation is moving to the gem:

https://github.com/ruby/nkf/issues/13

This won't happen until 3.4 compatibility and this should be rebased on our 3.4 branch when available.


headius:
This is already largely done except for moving the gem to bundled, which I'm about to do on the branch.

PR #8071, opened 2024-01-19
Update some warnings to deprecated

headius:
This also marks syscall as not implemented (warns of pending deletion in CRuby) and modifies all RubyWarnings.warning* forms to no-op on verbose, as they used to (unsure when this changed).


headius:
See https://github.com/ruby/ruby/pull/3917 for the same changes in CRuby.

headius:
Changing warning() to only fire in verbose mode exposed several places where we call warning() instead of warn() as in CRuby. I'll fix those cases as well.

PR #8072, opened 2024-01-20
Added Module#refinements, Refinement#target and Refinement#refined_class

moste00:
This PR adds methods to expose all the `Refinement` objects that a `Module` have (if any), and 2 aliases for a method returning the refined class of a given `Refinement` module (one is deprecated and ear-marked for removal in MRI 3.4).

### DIFF
--- a/core/src/main/java/org/jruby/RubyModule.java
+++ b/core/src/main/java/org/jruby/RubyModule.java
@@ -117,6 +117,7 @@ import org.jruby.runtime.marshal.UnmarshalStream;
 import org.jruby.runtime.opto.Invalidator;
 import org.jruby.runtime.opto.OptoFactory;
 import org.jruby.runtime.profile.MethodEnhancer;
+import org.jruby.util.RubyStringBuilder;
 import org.jruby.util.ByteList;
 import org.jruby.util.ClassProvider;
 import org.jruby.util.CommonByteLists;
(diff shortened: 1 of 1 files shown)


core/src/main/java/org/jruby/RubyModule.java
```diff
@@ -4510,6 +4510,35 @@ public final void setConstantVisibility(Ruby runtime, String name, boolean hidde
...
+            refinementModules.append(value);
+        });
+        return refinementModules;
+    }
+
+    @JRubyMethod(name = "target")
+    public IRubyObject target(ThreadContext context) {
+        return getRefinedClassOrThrow(context, true);
+    }
+    @JRubyMethod(name = "refined_class")
```
headius, line 4527: Newline here... we like a little extra whitespace.

core/src/main/java/org/jruby/RubyModule.java
```diff
@@ -4510,6 +4510,35 @@ public final void setConstantVisibility(Ruby runtime, String name, boolean hidde
...
+    public IRubyObject refined_class(ThreadContext context) {
+        return getRefinedClassOrThrow(context, false);
+    }
+
+    private IRubyObject getRefinedClassOrThrow(ThreadContext context, boolean nameIsTarget) {
+        if (!isRefinement()) {
+            String name = getSimpleName();
+            String clsName = getMetaClass().getSimpleName();
+            String methodName = nameIsTarget? "target":"refined_class";
```
headius, line 4536: Spaces around `?` and `:`.

core/src/main/java/org/jruby/RubyModule.java
```diff
@@ -4510,6 +4510,35 @@ public final void setConstantVisibility(Ruby runtime, String name, boolean hidde
...
+            String errMsg = "undefined method `"+ methodName +"' for "+ name +":"+ clsName;
```
headius, line 4537: @enebo might want to comment here; do we need to do anything extra here for encoding? Class names etc.

headius (commented):
Looks good with minor nitpicks. This will need to be rebased on 9.5-dev where we are doing 3.2 and 3.3 work.

core/src/main/java/org/jruby/RubyModule.java
enebo: @moste00 @headius yeah something like (static RubyStringBuilder.str):

```
String errMsg = str(runtime, "undefined method `" + methodName + "' for ", rubyBaseName(), ":", getMetaClass());
```

`methodName` is going to be a simple ascii name so we can just j.l.String concat but module name and metaclass name can be multibyte encoded and might not print out properly.  rubyBaseName() is like simpleName but will properly generate an encoded RubyString.  passing in metaclass should have it properly print out the type of the metaclass which if it is within singleton or other patterns will properly print out the string (see RubyStringBuilder.types to see the other processing it will do to types it encounters).'

It is also possible you do not want rubyBaseName and getmetaclass but could use just `this`.   not sure.

headius:
Everything looks good! I'll let latest CI run complete and then merge.

PR #8076, opened 2024-01-22
Revert "Update strscan to 3.0.9"

headius:
Reverts jruby/jruby#8075

Several tainting tests fail because we've removed tainting in the gem, and for some reason the new string copying logic leads to boundary errors. Reverting the update for now and will evaluate changes individually.


headius:
Boundary errors are very likely the same issue as #8083. Still need to get 9.3 strscan updated.

PR #8080, opened 2024-01-25
[fix] avoid Class#subclasses deprecation with class.rb ext

kares:
When doing a `require 'jruby/core_ext/class.rb'` the `Class#subclasses` method gets re-defined and cause a deprecation warning. Rails 7.x relies on (the Ruby 3.1 feature) `Class#subclasses` causing a lot of warning under JRuby 9.4.

The javadoc should provide an explanation (history of `Class#subclasses`), here's the demonstration of the issue:

```
bin/jruby -v -rjruby/core_ext/class -e 'puts String.subclasses'
jruby 9.4.6.0-SNAPSHOT (3.1.4) 2024-01-25 98ca4a1a53 OpenJDK 64-Bit Server VM 11.0.14.1+1 on 11.0.14.1+1 +jit [x86_64-linux]
DidYouMean::ClassNameChecker::ClassName
```
:green_circle: no warning

```
bin/jruby -v -rjruby/core_ext/class -e 'puts String.subclasses'
jruby 9.4.6.0-SNAPSHOT (3.1.4) 2024-01-25 98ca4a1a53 OpenJDK 64-Bit Server VM 11.0.14.1+1 on 11.0.14.1+1 +jit [x86_64-linux]
-e:1: warning: klass.subclasses is deprecated, use JRuby.subclasses(klass) instead
DidYouMean::ClassNameChecker::ClassName
```
:red_circle: 

---

AFTER applying this patch:

```
bin/jruby -v -rjruby/core_ext/class -e 'puts String.subclasses'
jruby 9.4.6.0-SNAPSHOT (3.1.4) 2024-01-25 98ca4a1a53 OpenJDK 64-Bit Server VM 11.0.14.1+1 on 11.0.14.1+1 +jit [x86_64-linux]
DidYouMean::ClassNameChecker::ClassName
```

```
bin/jruby -v -rjruby/core_ext/class -e 'puts String.subclasses(all: false)'
jruby 9.4.6.0-SNAPSHOT (3.1.4) 2024-01-25 98ca4a1a53 OpenJDK 64-Bit Server VM 11.0.14.1+1 on 11.0.14.1+1 +jit [x86_64-linux]
-e:1: warning: Class#subclasses(opts) is deprecated, please use the supported Class#subclasses() version
DidYouMean::ClassNameChecker::ClassName
```

The only down-side is that `String.subclasses(one_arg)` will now be available even without doing a `require 'jruby/core_ext/class'` but than doing that "properly" seemed like a lot of extra work, lmkwyt.

### DIFF
--- a/core/src/main/java/org/jruby/RubyClass.java
+++ b/core/src/main/java/org/jruby/RubyClass.java
@@ -57,6 +57,7 @@ import java.util.stream.Collectors;
 
 import org.jruby.anno.JRubyClass;
 import org.jruby.anno.JRubyMethod;
+import org.jruby.common.IRubyWarnings;
 import org.jruby.compiler.impl.SkinnyMethodAdapter;
 import org.jruby.exceptions.RaiseException;
 import org.jruby.internal.runtime.methods.DynamicMethod;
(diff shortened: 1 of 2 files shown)


headius:
Seems good to me. Should also get a PR in place to simply remove it in 9.5.

For that matter, we should consider putting together a PR for 9.5 that removes a whole bunch more deprecated stuff, since most of it has been there for years and years.

cc @enebo

enebo:
@kares @headius yeah I think it looks good.  We have quite a lot of warns when running Rails 7+ so anything to reduce noise is good.

kares:
Will mérge and open a follow up PR for 9.5.
The red CI is legit, will take a look at failures first.

PR #8084, opened 2024-02-03
remove (JRuby-only) deprecated Class#subclasses(arg)

kares:
follow-up on https://github.com/jruby/jruby/pull/8080 (merged **9.5-dev** with master first)


kares:
reviewed :red_circle: - failures (e.g. in `rake spec:ji`) existed prior and I do not see any new ones introduced

PR #8093, opened 2024-02-07
Rework visibility checks for Java integration

### LINKED ISSUE #8061: `enumerable.map` raise wrong number of arguments only when Ruby script is executed from Gradle using JDK 21
**Environment Information**
  
Provide at least:
* JRuby version (`jruby -v`) and command line (flags, JRUBY_OPTS, etc): `9.4.5.0`
* Operating system and platform (e.g. `uname -a`): <
```
Darwin andreas-MacBook-Pro-2.local 23.2.0 Darwin Kernel Version 23.2.0: Wed Nov 15 21:55:06 PST 2023; root:xnu-10002.61.3~2/RELEASE_ARM64_T6020 arm64
```

* Installed or activated gems
* Environment variables

**Expected Behavior**

When running the simple script using JDK 21
```ruby
hash_set = Java::JavaUtil::LinkedHashSet.new
hash_set << 1
hash_set << 2
hash_set << 3
res = hash_set.map {|v| v % 2}
```
within JRuby's `ScriptingContainer.runScriptlet` should create a new enumerable without throwing error when it's instantiated inside an embedding Java application and also when it's used from Gradle.

https://github.com/andsel/test_jdk21_jruby

**Actual Behavior**

*With JDK 21 the above test runs as expected  in an 
[embedding Java application](https://github.com/andsel/test_jdk21_jruby/blob/6fd243f53a39775b368af498fd0f8985bb87ce09/src/main/java/org/dna/App.java#L20) but fails with 
```
(ArgumentError) wrong number of arguments (given 1, expected 0)
``` 
if the same script is executed form a `ScriptingContainer` container [instantiated inside a Gradle script](https://github.com/andsel/test_jdk21_jruby/blob/6fd243f53a39775b368af498fd0f8985bb87ce09/build.gradle#L48).

headius:
In #8061 we found that Gradle is opening several core JDK packages to support the needs of Groovy. Unfortunately this causes some normally-inaccessible methods to be accessible to JRuby's Java integration layer. In the case of HashSet, a package-private `map` method gets bound that conflicts with the Ruby version, leading to the bug reported in #8061.

Gradle's behavior is unsanitary, but ultimately the fix we have settled on here is to never try to set package-private methods accessible since they are truly not intended for consumption outside the package. That fixes this issue without addressing the larger problem of Gradle opening up core JDK packages.

Fixes #8061

### DIFF
--- a/core/src/main/java/org/jruby/javasupport/binding/MethodGatherer.java
+++ b/core/src/main/java/org/jruby/javasupport/binding/MethodGatherer.java
@@ -8,6 +8,7 @@ import org.jruby.java.invokers.ConstructorInvoker;
 import org.jruby.java.util.ClassUtils;
 import org.jruby.javasupport.Java;
 import org.jruby.javasupport.JavaSupportImpl;
+import org.jruby.platform.Platform;
 import org.jruby.runtime.Block;
 import org.jruby.runtime.ThreadContext;
 import org.jruby.runtime.builtin.IRubyObject;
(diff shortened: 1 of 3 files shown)


core/src/main/java/org/jruby/javasupport/binding/MethodGatherer.java
```diff
@@ -717,17 +718,28 @@ private static class PartitionedMethods {
...
+                return true;
+            } else if (Modifier.isPrivate(mod)) {
+                // Exclude private
+                return false;
+            } else if (Modifier.isProtected(mod)) {
+                // Include protected if they can be set accessible
+                return Java.trySetAccessible(method);
+            } else {
+                // We split this on 17 because at that level unopened non-public functions are inaccessible
+                if (Platform.JAVA_VERSION >= 17) {
```
enebo, line 735: Is this backwards?

headius: It was indeed! I fixed it.

PR #8126, opened 2024-02-27
UnboundMethod#clone of frozen UnboundMethod should stay frozen

enebo:
UnboundMethod#clone of frozen UnboundMethod should stay frozen


headius:
The PR in #8218 and its commit accidentally referenced this bug instead of #8216.

PR #8140, opened 2024-03-05

kares:
a quick attempt to support partial removing of a reified class hierarchy... (https://github.com/jruby/jruby/issues/8141)

if this seems fishy I am okay to drop and leave this use-case as not supported

### DIFF
--- a/core/src/main/java/org/jruby/RubyClass.java
+++ b/core/src/main/java/org/jruby/RubyClass.java
@@ -1404,8 +1404,7 @@ public class RubyClass extends RubyModule {
         final boolean concreteExt = java_box[0]; 
 
         // calculate an appropriate name, for anonymous using inspect like format e.g. "Class:0x628fad4a"
-        final String name = getBaseName() != null ? getName() :
-                ( "Class_0x" + Integer.toHexString(System.identityHashCode(this)) );
+        final String name = getBaseName() != null ? getName() : ( "Class_0x" + Integer.toHexString(System.identityHashCode(this)) );
 
         final String javaName = "rubyobj." + StringSupport.replaceAll(name, "::", ".");
         final String javaPath = "rubyobj/" + StringSupport.replaceAll(name, "::", "/");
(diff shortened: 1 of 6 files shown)


core/src/main/java/org/jruby/RubyClass.java
```diff
@@ -1427,24 +1426,29 @@ public synchronized void reify(String classDumpDir, boolean useChildLoader) {
...
 
-        final ClassDefiningClassLoader parentCL;
+        ClassDefiningClassLoader classLoader; // usually parent's class-loader
         if (parentReified.getClassLoader() instanceof OneShotClassLoader) {
-            parentCL = (OneShotClassLoader) parentReified.getClassLoader();
+            classLoader = (OneShotClassLoader) parentReified.getClassLoader();
+
+            if (classLoader.hasDefinedClass(javaName)) { // class removed after being reified but parent got kept around
+                // while this seems like a leak the parent Java class might be GCable just hasn't been collected, yet
+                classLoader = new OneShotClassLoader((OneShotClassLoader) classLoader);
```
kares, line 1435: this extra wrapping (which under normal condition won't happen) is the fix, 
not super happy about it but I wonder if there are any concerns?

kares: alternatively we could append some suffix (counter) to the class-name but having the same name seemed nice.

this also reminds me that using `rubyobj.` package prefix with modularization might not be "ideal".

if we would to consider having some kind of package prefix than we might want to leverage `org.jruby.gen`:
- "`org.jruby.gen.reified_#{ruby_class.id}.SubList"`

kares: alternative proposal: https://github.com/jruby/jruby/pull/8147

kares: alternative alternative proposal: https://github.com/jruby/jruby/pull/8185

Issue #8135, opened 2024-02-29
Cannot install latest `rubocop`


boris-petrov:
```gradle
plugins {
    id 'com.github.jruby-gradle.base' version '2.0.2'
}

jruby.jrubyVersion = '9.4.6.0'

repositories {
    mavenCentral()
    ruby.gems()
}

configurations {
    rubocop
}

dependencies {
    rubocop 'rubygems:rubocop:1.61.0'
}

import com.github.jrubygradle.JRubyExec

def inputFiles = fileTree('.').include('**/*.rb').exclude('build').files.toList()

task rubocop(type: JRubyExec) {
    inputs.files(inputFiles + ['.rubocop.yml'])
    outputs.upToDateWhen { true }

    jruby.gemConfiguration 'rubocop'
    jrubyArgs '-S'
    script 'rubocop'
    scriptArgs(*inputFiles)
}
```

Run `gradle rubocop` on this (using Gradle 7.6.4 and JDK < 20 - e.g. 17). It will blow up:

```
> Task :jrubyPrepareRubocop
Building native extensions. This could take a while...
ERROR:  Error installing /home/boris/.gradle/caches/modules-2/files-2.1/rubygems/prism/0.24.0/e8e4979d1520a501a8e03c08cf18ad7cc74d6222/prism-0.24.0.gem:
        ERROR: Failed to build gem native extension.

    current directory: /home/boris/code/test/build/.gems-rubocop/gems/prism-0.24.0/ext/prism
/usr/lib/jvm/java-17-openjdk/bin/java -cp :/home/boris/.gradle/caches/modules-2/files-2.1/org.jruby/jruby-complete/9.4.6.0/7bb92a666a65520ff691fb4b804876c7965717bb/jruby-complete-9.4.6.0.jar org.jruby.Main -I uri:classloader:/META-INF/jruby.home/lib/ruby/stdlib extconf.rb
extconf failedBad file descriptor - /bin/sh
```

Happens because of the new version of `rubocop-ast` - 1.31.0 - which depends on `prism` - which can't be installed for some reason.

**Environment Information**
  
Provide at least:
* JRuby version - 9.4.6.0
* Operating system and platform - Arch Linux

**Expected Behavior**

Gem to be installed.

**Actual Behavior**

The above exception.

enebo:
[...]

PR #8142, opened 2024-03-05
Add-Opens for complete -jar runs as expected in later Java's

enebo:
jruby-complete.jar needs some module opens to work in Java 11+

This is another fix for #8135.  This still may not fully solve the original issue as that executes through a gradle plugin but it definitely fixes calling complete via `java -jar jruby-complete.jar`.  So there is a chance it will address the gradle plugin problem as well.


kares:
nice, did not know you can add these in the manifest...

Issue #1229, opened 2013-11-15
duplicate class definition when reifying classes


nirvdrum:
Since JRuby 1.7.8 had a reification fix, I figured I'd try running my project specs with that enabled again and ran into a new problem.  I'm not sure if the duplicate class definition is due to class re-opening or what.  There's a couple gems that patch Net::HTTP in play.

```
     Failure/Error: Unable to find matching line from backtrace
     Java::JavaLang::LinkageError:
       loader (instance of  org/jruby/util/ClassCache$OneShotClassLoader): attempted  duplicate class definition for name: "rubyobj/Net/HTTP"
     # java.lang.ClassLoader.defineClass1(Native Method)
     # java.lang.ClassLoader.defineClass(ClassLoader.java:800)
     # org.jruby.util.ClassCache$OneShotClassLoader.defineClass(ClassCache.java:74)
     # org.jruby.RubyClass.reify(RubyClass.java:1477)
     # org.jruby.RubyClass.reifyWithAncestors(RubyClass.java:1185)
     # org.jruby.RubyClass.reifyWithAncestors(RubyClass.java:1160)
     # org.jruby.RubyObject$13.allocate(RubyObject.java:268)
     # org.jruby.RubyClass.allocate(RubyClass.java:224)
     # org.jruby.RubyClass.newInstance(RubyClass.java:810)
     # org.jruby.RubyClass$INVOKER$i$newInstance.call(RubyClass$INVOKER$i$newInstance.gen)
     # org.jruby.internal.runtime.methods.JavaMethod$JavaMethodZeroOrOneOrTwoOrNBlock.call(JavaMethod.java:314)
     # org.jruby.internal.runtime.methods.AliasMethod.call(AliasMethod.java:66)
     # org.jruby.runtime.callsite.CachingCallSite.call(CachingCallSite.java:202)
     # org.jruby.ast.CallTwoArgNode.interpret(CallTwoArgNode.java:59)
     # org.jruby.ast.NewlineNode.interpret(NewlineNode.java:105)
     # org.jruby.evaluator.ASTInterpreter.INTERPRET_METHOD(ASTInterpreter.java:74)
     # org.jruby.internal.runtime.methods.InterpretedMethod.call(InterpretedMethod.java:225)
     # org.jruby.internal.runtime.methods.DefaultMethod.call(DefaultMethod.java:214)
     # org.jruby.runtime.callsite.CachingCallSite.call(CachingCallSite.java:202)
[...]

PR #8147, opened 2024-03-07

kares:
this is an alternative to https://github.com/jruby/jruby/pull/8140

the difference is no class-loader magic on re-created reified classes but it changes the Java class naming:

```ruby
module Bar
  class Foo < java.lang.Object
  end
end
```
when reified this class would be named:

- current `rubyobj.Bar.Foo`
- after the change in this MR: `rubyobj.23e5fef5.Bar.Foo`

to support the case of potential dropping and re-creating a named Ruby class

---

:warning: the fix would need to revert https://github.com/jruby/jruby/issues/1229 which I do not think has a good test [assuming 2 classes with the same name](https://github.com/jruby/jruby/blob/c50fb9b/spec/regression/GH-1229_reified_parent_and_child_have_same_name_spec.rb#L11) would share the same Java class

### DIFF
--- a/core/src/main/java/org/jruby/RubyClass.java
+++ b/core/src/main/java/org/jruby/RubyClass.java
@@ -1401,13 +1401,10 @@ public class RubyClass extends RubyModule {
         boolean[] java_box = { false };
         // re-check reifiable in case another reify call has jumped in ahead of us
         if (!isReifiable(java_box)) return;
-        final boolean concreteExt = java_box[0]; 
+        final boolean concreteExt = java_box[0];
 
-        // calculate an appropriate name, for anonymous using inspect like format e.g. "Class:0x628fad4a"
-        final String name = getBaseName() != null ? getName() : ( "Class_0x" + Integer.toHexString(System.identityHashCode(this)) );
-
-        final String javaName = "rubyobj." + StringSupport.replaceAll(name, "::", ".");
-        final String javaPath = "rubyobj/" + StringSupport.replaceAll(name, "::", "/");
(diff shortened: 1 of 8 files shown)


core/src/main/java/org/jruby/RubyClass.java
```diff
@@ -1496,6 +1492,18 @@ public synchronized void reify(String classDumpDir, boolean useChildLoader) {
...
+    private String getReifiedJavaClassName() {
+        final String basePackagePrefix = "rubyobj.";
+        if (getBaseName() == null) { // anonymous Class instance: rubyobj.Class$0x1234abcd
+            return basePackagePrefix + anonymousMetaNameWithIdentifier().replace(':', '$');
+        }
+        final String identityHash = Integer.toHexString(System.identityHashCode(this));
+        final CharSequence name = StringSupport.replaceAll(getName(), "::", ".");
+        // we need to include a Class identifier in the Java class name, since a Ruby class might be dropped
+        // (using remove_const) and re-created in which case using the same name would cause a conflict...
+        return basePackagePrefix + identityHash + '.' + name; // TheFoo::Bar -> rubyobj.320efdf5.TheFoo.Bar
```
kares, line 1504: and while this works "better" (well except for 'breaking' the `rubyobj.` naming convention) it might still be better to just add a (counter) suffix. given patching of a reified class should re-generate the Java class which means there needs to be more classloader-fu or we simply add a `v2` suffix and keep the existing loader (previous class is no longer used so should GC away, eventually)

PR #8149, opened 2024-03-12
Remove remaining "19" methods

headius:
These methods were not deprecated before JRuby 10, so they were left in place for one additional "major" release. This deletes them.

They are very low risk of causing compile or execution problems, but this safer approach gives any remaining users a full version cycle to switch to the long-available normal versions.

See jruby/jruby#8143 for the original pass that removed already- deprecated methods.


headius:
A wrinkle here is the use of `RubyString#cat19` from StringIO (https://github.com/ruby/stringio/issues/83). Those were originally added to 9.5/10 in #8143 and backported to 9.4 in 4114fc5a539d92f1d5f692cd1b2974a29f44429e.
For purposes of this PR, we will not remove the methods. Instead, we will track this migration in https://github.com/ruby/stringio/issues/83.

headius:
I modified the patch in ruby/stringio#84 to work on both old and new methods, so once that is released we can update this PR to use the new stringio version (or just update 9.4 to that version and merge it forward).

headius:
This has been rebased on a merge of master to 9.5-dev and goes back to removing all these methods. The stringio issue will be resolved once a release contains ruby/stringio#84 and ruby/stringio#88, which lets it support both old and new names. We will ship 9.4 with both old and new names, and 9.5 with new names only.

headius:
A few more method references were found in json: https://github.com/flori/json/pull/576

Once released, that PR will move to versions that exist in JRuby 9.4 so no special gymnastics are required (but we'll need that json released before we can make this change safely).

Issue #8141, opened 2024-03-05
can not re-reify sub-class of a Java sub-class


kares:
This is a bit of a stretch, usually not something that happens in production but still Ruby class removing is not that uncommon.

```ruby
class SubList < java.util.ArrayList
  def initialize(arg1, arg2)
    puts "initialize #{self}"
    super(arg1 + arg2)
  end
end

class SubSubList < SubList
end

p SubSubList.new(1, 2)

Object.send(:remove_const, :SubSubList) 
# ... while parent SubList stays around (if it's removed as well it's not an issue)

class SubSubList < SubList
end

p SubSubList.new(1, 2)

```

**Environment Information**
  
Provide at least:
* JRuby 9.3.0.0 - 9.4.6.0

**Expected Behavior**

* Script works, reification is able to deal with the removed sub-class while parent was kept

**Actual Behavior**

* Second `SubSubList.new` will fail:
```
initialize 
#<SubSubList: []>
ArgumentError: wrong number of arguments for constructor
  <main> at repro_java_sub_class.rb:26
```

due (`-XreifylogErrors`):
```
2024-03-05T15:10:02.637+01:00 [main] ERROR RubyClass : failed to reify class SubSubList due to: 
java.lang.LinkageError: loader org.jruby.util.OneShotClassLoader @1f172892 attempted duplicate class definition for rubyobj.SubSubList. (rubyobj.SubSubList is in unnamed module of loader org.jruby.util.OneShotClassLoader @1f172892, parent loader org.jruby.java.codegen.MultiClassLoader @6ba6557e)
	at java.base/java.lang.ClassLoader.defineClass1(Native Method)
	at java.base/java.lang.ClassLoader.defineClass(ClassLoader.java:1017)
	at org.jruby.dist/org.jruby.util.OneShotClassLoader.defineClass(OneShotClassLoader.java:20)
	at org.jruby.dist/org.jruby.RubyClass.reify(RubyClass.java:1447)

```

PR #8185, opened 2024-04-02

kares:
an attempt at fixing proper re-reification of Java (sub-)class hierarchy... (https://github.com/jruby/jruby/issues/8141)

this version behaves the same 99% of production uses (Java class name is kept); but when a Java class gets re-reified it will start using a versioned suffix  (`@v1`) for the Java class name.

2 previous attempts:
- https://github.com/jruby/jruby/pull/8140
- https://github.com/jruby/jruby/pull/8147

### DIFF
--- a/core/src/main/java/org/jruby/RubyClass.java
+++ b/core/src/main/java/org/jruby/RubyClass.java
@@ -1403,25 +1403,10 @@ public class RubyClass extends RubyModule {
         if (!isReifiable(java_box)) return;
         final boolean concreteExt = java_box[0];
 
-        final String javaName = getReifiedJavaClassName();
-        final String javaPath = javaName.replace('.', '/');
-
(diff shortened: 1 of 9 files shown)


core/src/main/java/org/jruby/RubyClass.java
```diff
@@ -1496,6 +1502,15 @@ public synchronized void reify(String classDumpDir, boolean useChildLoader) {
         }
     }
 
+    private String getReifiedJavaClassName() {
+        final String basePackagePrefix = "rubyobj.";
```
headius, line 1506: This might be a good time to select a new package name that can fit in with JPMS, or perhaps to provide a way for users to configure the package name.

headius (commented):
Looks good, only note is about us making package name configurable in preparation for modularity work in JRuby 10.

PR #8172, opened 2024-03-28
Optimized __method__ and __callee__

headius:
This PR optimizes the `__method__` and `__callee__` methods by doing the following:

* Introduce the FrameNameInstr instruction that performs a lightweight, frameless operation when the target `__method__` or `__callee__` is the built-in version (similar to `block_given?` in #8170).
* Introduce a new mapping from AliasMethod compound names to the pair of symbols they would produce for `__method__` and `__callee__`, avoiding extra allocation of strings while parsing that compound string.

As with #8170, if either of the target methods have been replaced by a user, we fall back to a normal invocation. If either method is called via metaprogramming paths, they will force a frame and use it as before.

A benchmark is included, and shows that both forms are now much faster (due to them no longer needing the caller's frame), and neither slow down when being used inside an aliased call:

```
[] jruby $ (chruby jruby-9.4.5.0 ; jruby bench/bench_frame_name_methods.rb)
Warming up --------------------------------------
     __method__ same   522.657k i/100ms
__method__ different   252.464k i/100ms
     __callee__ same   523.748k i/100ms
__callee__ different   264.330k i/100ms
Calculating -------------------------------------
     __method__ same      5.213M (± 1.8%) i/s -     26.133M in   5.014736s
__method__ different      2.565M (± 4.0%) i/s -     12.876M in   5.027256s
     __callee__ same      5.475M (± 1.8%) i/s -     27.759M in   5.071986s
__callee__ different      2.793M (± 1.9%) i/s -     14.009M in   5.017930s
[] jruby $ jruby bench/bench_frame_name_methods.rb                        
Warming up --------------------------------------
     __method__ same   885.744k i/100ms
__method__ different   936.830k i/100ms
     __callee__ same   926.174k i/100ms
__callee__ different   938.103k i/100ms
Calculating -------------------------------------
     __method__ same      9.313M (± 0.4%) i/s -     46.944M in   5.040932s
__method__ different      9.119M (± 0.5%) i/s -     45.905M in   5.034226s
     __callee__ same      9.187M (± 1.1%) i/s -     46.309M in   5.041351s
__callee__ different      9.056M (± 0.5%) i/s -     45.967M in   5.075968s
```

PR #8190, opened 2024-04-05
Fixes for regressions from recent optimizations

headius:
There are a few regressions that snuck in from recent optimization work (#8170 #8172 #8176 #8177 #8180). This PR will fix them.


headius:
https://github.com/jruby/jruby/pull/8227

Issue #8199, opened 2024-04-12
Enumerator#next_values behavior does not match MRI or docs


ryannevell:
**Environment Information**
jruby 9.4.6.0 (3.1.4) 2024-02-20 576fab2c51 OpenJDK 64-Bit Server VM 11.0.15+10-LTS on 11.0.15+10-LTS +jit [arm64-darwin]
MacOS Sonoma 14.2.1, M3 Max

**Expected Behavior**
https://ruby-doc.org/3.1.4/Enumerator.html#method-i-next_values describes the behavior for `Enumerator#next_values` with this table:

```
## yield args       next_values      next
#  yield            []               nil
#  yield 1          [1]              1
#  yield 1, 2       [1, 2]           [1, 2]
#  yield nil        [nil]            nil
#  yield [1, 2]     [[1, 2]]         [1, 2]
```

CRuby/MRI behaves as above for all versions tested:
- ruby 2.6.10p210 (2022-04-12 revision 67958) [universal.arm64e-darwin23]
- ruby 2.7.5p203 (2021-11-24 revision f69aeb8314) [arm64-darwin23]
- ruby 3.1.4p223 (2023-03-30 revision 957bb7cb81) [arm64-darwin23]
- ruby 3.2.2 (2023-03-30 revision e51014f9c0) [arm64-darwin23]

**Actual Behavior**
JRuby's behavior differs for the last 2 rows:
```
## yield args       next_values
#  yield nil        []     (expecting [nil])
#  yield [1, 2]     [1, 2] (expecting [[1, 2]])
```

**Test Code**
```
require 'minitest/autorun'
describe Enumerable do
  describe 'next_values' do
    it 'yield nil' do
      o = Object.new
      def o.each
        yield nil
      end
      _(o.to_enum.next_values).must_equal [nil]
    end

    it 'yield [1, 2]' do
      o = Object.new
      def o.each
        yield [1, 2]
      end
      _(o.to_enum.next_values).must_equal [[1, 2]]
    end
  end
end
```
**Output**
```
jruby 9.4.6.0 (3.1.4) 2024-02-20 576fab2c51 OpenJDK 64-Bit Server VM 11.0.15+10-LTS on 11.0.15+10-LTS +jit [arm64-darwin]
Run options: --seed 58630

# Running:

FF

Finished in 0.007662s, 261.0441 runs/s, 261.0441 assertions/s.

  1) Failure:
Enumerable::next_values#test_0001_yield nil [test.rb:13]:
[...]

PR #8201, opened 2024-04-17
Fix issue#8199

ryannevell:
This PR fixes the issue described in https://github.com/jruby/jruby/issues/8199, adds 2 new test cases, and removes tags for 2 existing, disabled test cases

### DIFF
--- a/core/src/main/ruby/jruby/kernel/enumerator.rb
+++ b/core/src/main/ruby/jruby/kernel/enumerator.rb
@@ -1,6 +1,11 @@
 class Enumerator
 
   def next
+    vals = next_values
+    vals.size <= 1 ? vals[0] : vals
+  end
+
+  def next_values
     return @__lookahead__.shift unless @__lookahead__.empty?
 
     unless @__generator__
@@ -14,7 +19,7 @@ class Enumerator
     end
 
     begin
-      return @__generator__.next if @__generator__.next?
+      return @__generator__.next_values if @__generator__.next?
     rescue StopIteration
       nil # the enumerator could change between next? and next leading to StopIteration
     end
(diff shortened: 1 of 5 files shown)


headius:
Very nice!

headius (approved):
Clever fix. If I'm reading it right, you avoid the confusion between passing nil and passing no arguments by adding an extra always-ignored nil into the Fiber.yield. That allows us to detect whether nil was passed explicitly or implicitly and the rest of the logic remains fairly simple even with the fix in place.

I have a minor concern that all Enumerator#next will now produce an array internally, usually immediately unwrapping it, but that's an optimization issue for another day.

ryannevell:
The additional, always-ignored `nil` into `Fiber.yield` is because `Fiber.yield` has very similar behavior to `next`, and implicitly unwraps arrays of size 1.

As for the performance concern, I actually don't think there's any additional array unwrapping. In the previous code, there was:
```
@result = object.__send__ method, *args do |*vals|
ret = Fiber.yield(*vals)
```
so that splat into `Fiber.yield(*vals)` was previously doing the array unwrapping. In the new code, the `vals` array is passed straight through `Fiber.yield(vals, nil)` and returned from `next_values` to finally be unwrapped in `next`. So it should be the same number of allocations and unwraps.

I could imagine some performance degradation due to the additional call, tho, since `next` now calls `next_values`. I did not attempt to benchmark this change.

headius:
> I did not attempt to benchmark this change.

You are right... there's already a lot of argument array-juggling so if anything has changed it's probably a lateral move. I have not benchmarked this code much *at all* so it's an open project to investigate that (and likely the hottest pieces would need to move to Java code to improve performance).

Thanks for the fix!

ryannevell:
I'm thrilled to give back. Thanks for all your hard work on JRuby, it's a fantastic project.

PR #8204, opened 2024-04-22
bump rdoc dependency to 6.4.1.1

jsvd:
The current recommendation for Ruby 3.1 users is to update to rdoc 6.4.1.1 due to known issues.


headius:
Seems fine. Thanks!

PR #8206, opened 2024-04-22
Mark frame-aware method names with bang suffix for ostruct

### LINKED ISSUE #8200: Warnings emitted by requiring 'ostruct'
:wave: Hello my friends! Long time no see, some might say the not seeing level is over 9000 ;)

## Environment

```
tobi@qiqi:~/github/simplecov(ci-madness)$ jruby -v
jruby 9.4.6.0 (3.1.4) 2024-02-20 576fab2c51 OpenJDK 64-Bit Server VM 11.0.22+7-post-Ubuntu-0ubuntu222.04.1 on 11.0.22+7-post-Ubuntu-0ubuntu222.04.1 +jit [x86_64-linux]
tobi@qiqi:~/github/simplecov(ci-madness)$ uname -a
Linux qiqi 6.5.0-27-generic #28~22.04.1-Ubuntu SMP PREEMPT_DYNAMIC Fri Mar 15 10:51:06 UTC 2 x86_64 x86_64 x86_64 GNU/Linux
```

## Behaviour

Just requiring open struct yields the following warnings with warnings turned on

```
tobi@qiqi:~/github/simplecov(ci-madness)$ cat wtwas.rb 
require 'ostruct'
tobi@qiqi:~/github/simplecov(ci-madness)$ jruby -w wtwas.rb 
/home/tobi/.asdf/installs/ruby/jruby-9.4.6.0/lib/ruby/stdlib/ostruct.rb:466: warning: OpenStruct#public_send accesses caller method's state and should not be aliased
/home/tobi/.asdf/installs/ruby/jruby-9.4.6.0/lib/ruby/stdlib/ostruct.rb:466: warning: OpenStruct#method accesses caller method's state and should not be aliased
```

**Expected Behavior**

No warnings :)

## More context

FWIW the simplecov CI is failing because of this, I'll work around it for now somehow. We have a test that just makes sure we don't emit warnings.

Cheers, and thanks for all your work :green_heart:

headius:
Also flag frame names with "!" for ostruct

The ostruct library aliases all methods from Object/Kernel with a bang suffix ("!") to allow them to be callable even though the OpenStruct should support them as field names. Because we use record these names in a table of known frame-aware methods, we have typically warned when aliasing them. This rarely comes up, since aliasing usually is accompanied by wrapping, which breaks their behavior on all implementations, but this aliasing in ostruct is unusual and pervasive, leading to issues like #8200.

This patch assumes we're going to have issues with ostruct forever and eagerly adds the "!" names alongside the regular names for all method names that match /[a-z_]/, so taht existing methods and future methods will behavior properly and not warn when aliased.

It adds a small amount to startup, since these method names must be added twice, but there are not many such names in the system.

Fixes #8200

Replaces #7524

### DIFF
--- a/core/src/main/java/org/jruby/anno/AnnotationHelper.java
+++ b/core/src/main/java/org/jruby/anno/AnnotationHelper.java
@@ -6,6 +6,7 @@ import java.util.HashSet;
 import java.util.List;
 import java.util.Map;
 import java.util.Set;
+import java.util.StringJoiner;
 import java.util.function.BiConsumer;
 import java.util.stream.Collectors;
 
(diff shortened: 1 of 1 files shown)


headius:
A total of 47 additional names are added to the list of frame-aware methods:

```
__callee__!
__dir__!
__method__!
attr!
attr_accessor!
attr_reader!
attr_writer!
autoload!
binding!
chomp!
chop!
class_eval!
class_exec!
define_method!
eval!
extend_proxy!
foreach!
gets!
gsub!
index!
instance_eval!
instance_exec!
instance_method!
last_match!
local_variables!
match!
method!
module_eval!
module_exec!
module_function!
nesting!
partition!
print!
private!
protected!
public!
public_send!
readline!
refine!
require_relative!
rindex!
rpartition!
scan!
slice!
sub!
used_modules!
using!
```

This translates to 69 additional adds during `-e 1` startup due to some methods being defined in multiple classes.

PR #8208, opened 2024-04-24
Only set up Java:: constants when accessed directly

headius:
Previously whenever we see a new type of Java object, we both set up the proxy class AND set the constant. If that object's class is not from the typical Java integration classloader hierarchy, it will pollute the Java:: module with a normally inaccessible entry.

This change only sets up the constant when it is accessed through the normal package reference, rather than when first seen.


headius:
This is now guarded by an option so we can ship it in 9.4.7.0.

PR #8212, opened 2024-04-25
Correct fiddle values so things like size_t work as expected

### LINKED ISSUE #7883: NoMethodError when using Fiddle struct with size_t field
**Environment Information**
  
Provide at least:
* JRuby version (`jruby -v`): `jruby 9.4.3.0 (3.1.4) 2023-06-07 3086960792 OpenJDK 64-Bit Server VM 17.0.7+7 on 17.0.7+7 +jit [x86_64-linux]`

```
$ ruby -rfiddle/import -e 'module M; extend Fiddle::Importer; Buffer = struct ["char *value", "size_t length"]; end'
NoMethodError: undefined method `>' for nil:NilClass
    alignment at /home/eregon/.rubies/jruby-9.4.3.0/lib/ruby/stdlib/fiddle/struct.rb:285
         each at org/jruby/RubyArray.java:1989
    alignment at /home/eregon/.rubies/jruby-9.4.3.0/lib/ruby/stdlib/fiddle/struct.rb:279
       create at /home/eregon/.rubies/jruby-9.4.3.0/lib/ruby/stdlib/fiddle/struct.rb:253
  module_exec at org/jruby/RubyModule.java:3728
   initialize at org/jruby/RubyModule.java:2943
   initialize at org/jruby/RubyClass.java:957
          new at org/jruby/RubyClass.java:904
       create at /home/eregon/.rubies/jruby-9.4.3.0/lib/ruby/stdlib/fiddle/struct.rb:216
       struct at /home/eregon/.rubies/jruby-9.4.3.0/lib/ruby/stdlib/fiddle/import.rb:222
    <class:M> at -e:1
       <main> at -e:1
```

But this works fine in CRuby.
Pretty much blocks using Fiddle for YARP in https://github.com/ruby/yarp/pull/1112

enebo:
Hmm a lot of wrong values and some extra ones MRI does not have.  Maybe fixes #7883

### DIFF
--- a/lib/ruby/stdlib/fiddle/jruby.rb
+++ b/lib/ruby/stdlib/fiddle/jruby.rb
@@ -18,25 +18,36 @@ module Fiddle
     Pointer.to_ptr(val)
   end
 
-  TYPE_VOID         = 0
-  TYPE_VOIDP        = 1
-  TYPE_CHAR         = 2
+
+  TYPE_VOID=0
+  TYPE_VOIDP=1
+  TYPE_CHAR=2
+  TYPE_SHORT=3
+  TYPE_INT=4
+  TYPE_LONG=5
+  TYPE_LONG_LONG=6
+  TYPE_FLOAT=7
+  TYPE_DOUBLE=8
+  TYPE_SIZE_T=-5
+  TYPE_SSIZE_T=5
+  TYPE_CONST_STRING=10
+  TYPE_PTRDIFF_T=5
+  TYPE_INTPTR_T=5
+  TYPE_UINTPTR_T=-5
+
+  # Types defined in MRI
+  TYPE_INT8_T=2
+  TYPE_INT16_T=3
+  TYPE_INT32_T=4
+  TYPE_INT64_T=5
+  TYPE_VARIADIC=9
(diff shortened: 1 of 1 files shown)


ahorek:
yeah, some of these values are wrong. Also, they're platform-specific like 32 vs 64bit, Windows vs Linux... etc. we should generate them like TR does https://github.com/oracle/truffleruby/blob/master/lib/truffle/truffle/fiddle_backend.rb

ahorek:
fiddle tests are currently disabled https://github.com/jruby/jruby/blob/master/test/mri.stdlib.index#L90 IIRC there were many issues with the pointer logic...

enebo:
@ahorek yeah generation sounds totally right.  The FFI pointer thing I noticed was strcpy in libc would return an FFI::Pointer and not the String it represents.  Amusingly I could never the get pointer to print out the actual value (although the actual argument which also was represented by the pointer did become a valid String so I just didn't figure out the magic).

enebo:
This PR was just seeing how quickly I could get around a problem.  I think we need a broader effort to fix this (and that is not this).  Closing.

Issue #8079, opened 2024-01-25
Inconsistency between MRI and JRuby source location.


ioquatix:
```
samuel@aiko ~/P/i/bake (modernize-gem)> ruby --version
jruby 9.4.5.0 (3.1.4) 2023-11-02 1abae2700f OpenJDK 64-Bit Server VM 21.0.2+13 on 21.0.2+13 +jit [x86_64-linux]

samuel@aiko ~/P/i/bake (modernize-gem)> cat test.rb
class Thing
	attr :name
end

puts Thing.instance_method(:name)

samuel@aiko ~/P/i/bake (modernize-gem)> chruby jruby
samuel@aiko ~/P/i/bake (modernize-gem)> ruby test.rb
#<UnboundMethod: Thing#name(@name)()>
samuel@aiko ~/P/i/bake (modernize-gem)> chruby 3.3
samuel@aiko ~/P/i/bake (modernize-gem)> ruby test.rb
#<UnboundMethod: Thing#name() test.rb:2>
```

I believe `attr` is not having the source location set.

headius:
Probably not, because it would require us to walk the stack to get it. Maybe can do it in 9.5 when we can rely on the lighter weight StackWalker API in JVM.

We could do other tricks too like making a special call site for it, but I dunno if it's worth all that.

ioquatix:
I worked around the problem, but I just thought I'd mention the inconsistency.

rubyFeedback:
> We could do other tricks too like making a special call site for it, but I dunno if it's worth all that.

Ideally 100% compatibility should be attained, but specifically for the "attr", I actually never
used it. I used attr_reader, attr_writer and attr_accessor in the past. However had, I even stopped
doing that some years ago, largely because I always want to have something like:

    def foobar?
      @foobar
    end; alias foobar foobar?

I like the trailing '?'.

These days, when a class has too many @foo variables, I just store them in a single
hash, which I usually call @internal_hash, and handle via "def reset" (which is the
general method I typically use for resetting towards the initial state).

Evidently many other ruby users may not do something like this, but I just wanted
to add that for me, personally, attr is the least important from among the attr-family,
and even that whole family is not so useful to me personally (though I totally understand
the advantage of being able to quickly and programmatically define accessors, of
course; I am not disputing that part at all).

headius:
> Ideally 100% compatibility should be attained

[...]

PR #8215, opened 2024-04-25
Make attrs position-aware

headius:
This is a proof-of-concept for adding the definition location (from source where `attr`, `attr_reader`, `attr_accessor` or `attr_writer` was called) into all `attr`-style methods, to address the missing information reported in #8079.

A benchmark is included, and shows how severely this impacts the creation of attrs, due to the stack walking involved:

```
[] jruby $ jruby bench/bench_attr_reader.rb 
jruby 10.0.0.0-SNAPSHOT (3.4.0) 2024-04-25 feb5f4c851 OpenJDK 64-Bit Server VM 17.0.5+8 on 17.0.5+8 +jit [arm64-darwin]
         attr_reader    11.853k i/100ms
         attr_reader    128.672k (± 0.5%) i/s -    651.915k in   5.066593s
[] jruby $ (chruby jruby-9.4.6.0 ; jruby bench/bench_attr_reader.rb)
         attr_reader   960.880k i/100ms
         attr_reader      9.701M (± 0.4%) i/s -     49.005M in   5.051463s
```

Using a specialized call site would make this largely free, similar to how #8170 avoids frame access for `block_given?` but this example demonstrates one (slow) way to solve the bug.

### DIFF
--- a/bench/bench_attr_reader.rb
+++ b/bench/bench_attr_reader.rb
@@ -1,0 +1,12 @@
+require 'benchmark/ips'
+
+Benchmark.ips do |bm|
+  bm.report("attr_reader") do |i|
+    Class.new do
+      while i > 0
+        i-=1
+        attr_reader :foo
+      end
+    end
+  end
+end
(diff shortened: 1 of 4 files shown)


headius:
This is clearly not the way we would want to do this, given the impact to attr definition. It should just be a special call site, and we have other places where we need `__FILE__`-aware calls like for `__dir__` and `require_relative`. Added note to #7588.

PR #8231, opened 2024-05-09
Update ruby/specs

andrykonchin:
Import recent changes in ruby/spec.

The previous PR #8182


andrykonchin:
There are failing new specs. Wondering if there is a script/command to tag them easily?

enebo:
Thanks @andrykonchin!

PR #8239, opened 2024-05-15
Fix Dir.home for empty dir name

edipofederle:


enebo:
Thanks @edipofederle .  Pretty wild semantics.

PR #8250, opened 2024-05-21
[ji] avoid already initialized constant warnings

kares:
similar to how `java_import` works we should check whether the constant was already set and not set it again - this effectively avoids emitting a lot of *"already initialized constant"* in a multi threaded environment. 

on top of that using synchronization we avoid *"already initialized constant"* even in rare cases (which could have still happened with Java proxy class initialization in general).

reproducer for the warnings is easy to spot esp. with `include_package`:

`jruby -v -e "mod = Module.new { include_package 'java.lang.reflect' }; threads = (0..10).map { Thread.start { sleep(0.01); mod::Array } }.each { |t| t.join }"`

```
jruby 9.4.6.0 (3.1.4) 2024-02-20 576fab2c51 OpenJDK 64-Bit Server VM 11.0.14.1+1 on 11.0.14.1+1 +jit [x86_64-linux]
-e:1: warning: already initialized constant #<Module:0x5957245>::Array
-e:1: warning: already initialized constant #<Module:0x5957245>::Array
-e:1: warning: already initialized constant #<Module:0x5957245>::Array
-e:1: warning: already initialized constant #<Module:0x5957245>::Array
-e:1: warning: already initialized constant #<Module:0x5957245>::Array
-e:1: warning: already initialized constant #<Module:0x5957245>::Array
-e:1: warning: already initialized constant #<Module:0x5957245>::Array
```

### DIFF
--- a/test/jruby/test_higher_javasupport.rb
+++ b/test/jruby/test_higher_javasupport.rb
@@ -1773,16 +1773,29 @@ CLASSDEF
       include_package 'java.math'
       include_package 'java.net'
     end
+
+    thread_count = 50
+
     output = with_stderr_captured do
-      threads = (0..100).map do
+      latch = java.util.concurrent.CountDownLatch.new(1)
+      waiting_thread_count = java.util.concurrent.atomic.AtomicInteger.new
+
+      threads = (0...thread_count).map do
         Thread.start do
-          sleep 0.01; mod::URLEncoder
-          sleep 0.01; mod::Field
-          sleep 0.01; mod::ThreadInfo
-          sleep 0.01; mod::LockInfo
-          sleep 0.01; mod::Array
+          waiting_thread_count.incrementAndGet
+          latch.await
+
+          mod::URLEncoder
+          mod::Field
+          mod::ThreadInfo
+          mod::LockInfo
+          mod::Array
         end
       end
+
+      Thread.pass while (waiting_thread_count.get < thread_count)
+      latch.countDown
+
       threads.each { |thread| thread.join }
     end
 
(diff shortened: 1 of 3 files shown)


test/jruby/test_higher_javasupport.rb
```diff
@@ -1760,7 +1760,40 @@ def test_no_warnings_on_concurrent_class_const_initialization
...
+      include_package 'java.net'
+    end
+    output = with_stderr_captured do
+      threads = (0..100).map do
+        Thread.start do
+          sleep 0.01; mod::URLEncoder
+          sleep 0.01; mod::Field
+          sleep 0.01; mod::ThreadInfo
+          sleep 0.01; mod::LockInfo
+          sleep 0.01; mod::Array
```
ikaronen-relex, line 1783: A better concurrency test could be to have each spawned thread wait on something like a CountDownLatch before trying to access each imported Java class, and have the main thread release the latches one at a time.

kares: yeah, that's a good point. although this actually does reproduce both types of warnings fairly reliably...

test/jruby/test_higher_javasupport.rb
```diff
@@ -1773,16 +1773,29 @@ def test_no_warnings_on_concurrent_class_loading_with_include_package
       include_package 'java.math'
       include_package 'java.net'
     end
+
+    thread_count = 50
+
     output = with_stderr_captured do
-      threads = (0..100).map do
+      latch = java.util.concurrent.CountDownLatch.new(1)
+      waiting_thread_count = java.util.concurrent.atomic.AtomicInteger.new
```
ikaronen-relex, line 1781: What about making this one a CountDownLatch too? Waiting for something to happen _N_ times is what they're designed for, even if I do often end up using them with _N_ = 1 😄

PR #8251, opened 2024-05-21
Dir.chdir conflicting warning and error

edipofederle:


core/src/main/java/org/jruby/RubyDir.java
```diff
@@ -84,6 +84,8 @@ public class RubyDir extends RubyObject implements Closeable {
     private int pos;               // current position in directory
     private boolean isOpen = true;
     private Encoding encoding;
+    private static final ReentrantLock lockChdir = new ReentrantLock();
+    private static RubyThread currentThread;
```
enebo, line 88: We cannot store any Ruby objects into a static field in Java because we can have multiple Ruby runtimes existing at the same time.  This needs to be saved in the individual instance of runtime somewhere.

enebo (changes requested):
You can probably put this as a field on Ruby (with a more specific name than currentThread) and otherwise I think this looks good.

The lone failure in sequel is unrelated to this PR.

enebo:
@edipofederle @headius Actually I have other concerns with this now that I think about it.  If you have an n-thread server which all was calling out to a process with the same directory this would lock the entire server to becoming synchronous.  If we allowed all threads which called processes in the same directory to run it would alleviate some of this concern but I now question if this is a good idea or not.

enebo:
Just some scenarios I thought about from the above comment:

 1. n Dir.chdir to the SAME dir means none of those other threads can execute at the same time
 2. one Dir.chdir which never executes would block any other Dir.chdir from ever exectuting.

I am not sure what the right thing to do here is but the second problem seems like it will generate a different issue.  I half wonder if there is some solution where every thread externally tracks chdir and if anything using chdir (like process launch) it somehow gets used instead.  This I can already see would not fix call outs to external Java libraries doing processes but I feel like that problem is not solvable.

enebo:
@edipofederle This is the issue to follow for this https://bugs.ruby-lang.org/issues/15661

So I think you must have read this issue based on how the PR was written.  I think just moving currentThread to Ruby so it is not a staticfield and I think the lock should be removed.

In particular https://bugs.ruby-lang.org/issues/15661#note-16 should work (or at least the comments make me think it should work).  With the PR as it is I think it will warn then dead lock.

enebo:
@edipofederle thanks for bearing with me on that.  We got there in the end.  Thanks for the contribution!

PR #8258, opened 2024-05-23
Check for override method on eql?

edipofederle:


core/src/main/java/org/jruby/RubyHash.java
```diff
@@ -1238,6 +1238,10 @@ public RubyBoolean compare(final ThreadContext context, VisitorWithState<RubyHas
             if (!sites(context).respond_to_to_hash.respondsTo(context, other, other)) {
                 return context.fals;
             }
+            if(other.getMetaClass().getMethods().containsKey("eql?")){
+                return (RubyBoolean) other.callMethod(context, "eql?", this);
+            }
+
             return Helpers.rbEqual(context, other, this);
```
enebo, line 1245: I think this is "probably" wrong.  rbEqual is calling `==` and I am wondering here why would hash be checking both `eql?` and `==`.  I suspect it should only be calling `eql?`.  You can try removing your if and switch from `rbEqual` to `rbEql` and see if that fixes this case and nothing else breaks.

enebo (changes requested):

core/src/main/java/org/jruby/RubyHash.java
edipofederle: just using the` rbEql `make more tests to fail. I will investigate in more tails soon.

headius: I could not find logic in CRuby that dispatches to `==` but that code is pretty involved. We should sort out exactly what the sequence of checks and dispatches is supposed to be and fix this once.

edipofederle: @enebo @headius I pushed a change that, from what I see, match more what is present in CRuby. Maybe can take a look when possible? Thanks

enebo: @edipofederle This looks pretty good to me but I feel like since you hit these other types this would have to fix those types in the same way it fixed the test you were able to un-exclude.  Any chance you can make some tests to show that this fixes similar problems in the other types?

enebo: @headius You should definitely look this over too.  It looks good to me because it looks consistent and clean but I honestly don't know if it is actually correct.  If @edipofederle shows some other tests that should build more confidence.

headius: Looks good to me! I initially missed that this is a fix for `Hash#==` and `Hash#eql?` but having reviewed that logic in CRuby it appears to match the new logic by @edipofederle.

headius:
All issues have been resolved so I'm going to go ahead with the merge. Thanks again @edipofederle!

Issue #6844, opened 2021-09-18
Feature Request: Improve polyglot debugging


JasonLunn:
Enhance the debugger to support stepping from JRuby to Java and back again without resorting to tricks like attaching a [separate debugger for JRuby and Java](https://intellij-support.jetbrains.com/hc/en-us/community/posts/206709905-Debug-Java-and-Ruby-in-same-application-).

Extra points if the debugger works from within standard graphical tooling (read: RubyMine/Intellij) or can be made to do so with supplemental community-contributed updates.

headius:
So the short story here is that it should be possible to step-debug Ruby and Java code together by setting JRuby to force JIT compilation of code it runs. That means it will all be bytecode and steppable, except for cases we can't compile to bytecode (usually just really large methods).

The long story is that because Ruby sources dynamically, we have not figured out the magic sauce for getting e.g. IntelliJ to recognize the Ruby sources as they are loaded. This is also complicated by how we embed the Ruby source file names into the generated bytecode, which currently uses an absolute path. With some work we should be able to propagate the relative paths through, which would allow adding an individual gem's sources to IntelliJ's debugger.

So the pieces are here for this to work, but it needs some finessing to get them in the right places for traditional JVM debuggers to work properly.

headius:
I started doing a lot of work on this in #8259:

* More readable stack trace, still up for debate on format
* Local variables show up as locals in the JVM debugger, sans internal variables a user shouldn't need to see (pass -Xjit.debug to see all of them)
* Exploring options for better rendering in IntelliJ
* Exploring options for passing through a relative path if that would help debug tools find our sources: https://gist.github.com/headius/0bf47e90d121b48293d64e7a513bdc99

I am actively working on making JRuby+JVM debugging better. Hoping that we can get some help from @hurricup or others at JetBrains to make IDEA debugging support for JRuby better.

PR #8259, opened 2024-05-24
Improved representation of JRuby for JVM tools

headius:
This PR will encompass several improvements to JRuby's representation of code and objects, in order to make JVM tooling more useful.

Among the ideas explored here:

* Cleaner representation of methods and class for stack traces and profiling.
* Always-on class reification to show real class names in JVM heap inspection tools.
* Better JIT configuration to allow JVM debugging tools to step through Ruby code and set breakpoints and such.
  * This may include some changes to force larger scopes to JIT so they can be debugged, even if this means the JVM cannot native JIT them.
* Enhancements for JVM debuggers to better see into dynamic scopes, Ruby objects with instance variables, and core Ruby types with complicated internals.

### DIFF
--- a/core/src/main/java/org/jruby/RubyBasicObject.java
+++ b/core/src/main/java/org/jruby/RubyBasicObject.java
@@ -43,7 +43,6 @@ import java.io.IOException;
 import java.io.ObjectInputStream;
 import java.io.ObjectOutputStream;
 import java.io.Serializable;
-import java.util.AbstractMap;
 import java.util.ArrayList;
 import java.util.Collections;
 import java.util.HashSet;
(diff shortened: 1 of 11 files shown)


headius:
This piece may not make sense to introduce the dependency, but it's super cool... IntelliJ annotations for better rendering of objects in the debug view:

https://gist.github.com/headius/5c7e90fe501a66b0159a7bfecabf9907

With these changes, DynamicScope will show a simple name to value mapping for local variables, and RubyBasicObject will show instance variable names mapped to their current values.

This was discovered while investigating why the IntelliJ debugger can't currently render DynamicScope... it turns out they added a renderer for us years ago but it's based on an old object layout: https://youtrack.jetbrains.com/issue/IDEA-353990/JRuby-DynamicScope-renderer-is-broken

headius:
An additional change here makes a new class for not-given blocks, but might be risky since up til now we have not had any subclasses of Block. Note the change in the JIT necessary to make sure that NotGiven is cast to Block for indy purposes:

https://gist.github.com/headius/59b074be588004ade9e32d68843aff54

headius:
Going to stop here but I've filed https://github.com/jruby/jruby/issues/8265 for propagating the relative path through to the JIT bytecode source attribute, if that turns out to be useful for debugging tools (see #6844 and some discussion in https://youtrack.jetbrains.com/issue/RUBY-32948/JRuby-DynamicScope-renderer-is-broken).

JasonLunn:
Amazing! Is there anything you need contributed for this effort?

headius:
@JasonLunn Hey I missed this but yeah there's probably a ton of ways to help! What JVM tools are you familiar with?

PR #8285, opened 2024-06-07
Update to ruby/spec@517f06f

andrykonchin:


headius:
Thanks @andrykonchin!

Going forward, I think we should either add you as a committer (so you can merge these) or we should take over the update ourselves. Thoughts?

andrykonchin:
Yeah, it would be great if I could merge such PRs on my own.

headius:
@andrykonchin I have sent an invite to add you to a writable team!

andrykonchin:
Thank you!

PR #8294, opened 2024-06-20
Fix IPSocket.getaddress to handle special case IPv6 addresses

### LINKED ISSUE #8248: JRuby parses "mixed" IPv6 address as IPv4
**Environment Information**
  
* JRuby 9.4.7.0 (jruby 9.4.7.0 (3.1.4) 2024-04-29 597ff08ac1 OpenJDK 64-Bit Server VM 25.412-b08 on 1.8.0_412-b08 +jit [arm64-darwin])
* macOS 14.4.1

I've observed the same behavior on Ubuntu 22.04.4 with the same version of JRuby.

**Expected Behavior**

IPv6 addresses can mix the formatting of IPv6 and IPv4 in one address (e.g. ::ffff:192.0.2.128), as defined in [RFC4291:](https://www.rfc-editor.org/rfc/rfc4291#section-2.2)

>  An alternative form that is sometimes more convenient when dealing
>  with a mixed environment of IPv4 and IPv6 nodes is
>  x:x:x:x:x:x:d.d.d.d, where the 'x's are the hexadecimal values of
>  the six high-order 16-bit pieces of the address, and the 'd's are
>  the decimal values of the four low-order 8-bit pieces of the
>  address (standard IPv4 representation).  Examples:
> 
>   0:0:0:0:0:0:13.1.68.3
> 
>   0:0:0:0:0:FFFF:129.144.52.38
> 
>  or in compressed form:
> 
>  ::13.1.68.3
> 
>  ::FFFF:129.144.52.38

MRI 3.1.5 correctly parses these types of IPv6 addresses:

```
$ ruby -v
ruby 3.1.5p252 (2024-04-23 revision 1945f8dc0e) [arm64-darwin23]
$ irb -r socket
irb(main):001:0> IPSocket.getaddress('::ffff:192.0.2.128')
=> "::ffff:192.0.2.128"
```

**Actual Behavior**

JRuby 9.4.7.0 incorrectly parses the same address as an IPv4 address:

```
$ ruby -v
jruby 9.4.7.0 (3.1.4) 2024-04-29 597ff08ac1 OpenJDK 64-Bit Server VM 25.412-b08 on 1.8.0_412-b08 +jit [arm64-darwin]
$ irb -r socket
irb(main):001:0> IPSocket.getaddress('::ffff:192.0.2.128')
=> "192.0.2.128"
```

Note that when JRuby drops the hex characters and `::` syntax at the start of the address.

This can have knock-on effects for things like `Addrinfo`:

```
irb(main):002:0> Addrinfo.new(['AF_INET6', 0, nil, '::ffff:192.0.2.128'], :INET6)
org/jruby/ext/socket/Addrinfo.java:139:in `initialize': getaddrinfo: Address family for hostname not supported (SocketError)
```

mullermp:
Fixes #8248

When given an IPv4 mapped IPv6 address, Java returns an Inet4Address object that returns an IPv4 address. This is different than what Ruby's IPSocket.getaddress will return. Since IPv4 mapped IPv6 addresses are a special case and have an expected format, we can just use this string directly.

This change also fixes the special case of IPv6 localhost.


core/src/main/java/org/jruby/ext/socket/SocketUtils.java
```diff
@@ -502,7 +501,25 @@ private static InetAddress specialAddress(String addressString) throws UnknownHo
 
     public static IRubyObject getaddress(ThreadContext context, IRubyObject hostname) {
         try {
-            return RubyString.newInternalFromJavaExternal(context.runtime, InetAddress.getByName(hostname.convertToString().toString()).getHostAddress());
+            String hostnameString = hostname.convertToString().toString();
+            InetAddress address = InetAddress.getByName(hostnameString);
+
+            String ipv4MappedAddressPrefix = "::ffff:";
```
enebo, line 507: Can you make these two string static finals?

enebo (changes requested):
Looks good.  If you do not have time I can make the changes I asked for after landing.  Just let me know.  I will approve so we can see it pass in CI.

core/src/main/java/org/jruby/ext/socket/SocketUtils.java
mullermp: I had to move them outside of the method to make them static.

enebo: @mullermp yeah this was just to make those strings once.  Thanks!

Issue #799, opened 2013-06-11
Make linkage errors from accessing Java classes more consistent.


headius:
When there's a linkage error loading a Java class, Java reports an exception that indicates the class it could not find (usually because a required jar file was not in classpath). For java_import, we scrape this class name out of the message for our error.

Example:

```
ext-jruby-local ~/projects/jruby $ jar tf blah.jar
META-INF/
META-INF/MANIFEST.MF
Blah.class

ext-jruby-local ~/projects/jruby $ jar tf foo.jar
META-INF/
META-INF/MANIFEST.MF
Foo.class

ext-jruby-local ~/projects/jruby $ cat Blah.java
public class Blah {
  static {
    Foo.foo();
  }
}

ext-jruby-local ~/projects/jruby $ cat Foo.java
public class Foo {
  public static void foo() {
    System.out.println("foo");
  }
}
```

If we java_import the class by name, the error reports that it's probably the Foo class missing:

```
ext-jruby-local ~/projects/jruby $ jruby -e "require 'blah.jar'; java_import 'Blah'"
NameError: cannot link Java class Blah, probable missing dependency: Foo
         for_name at org/jruby/javasupport/JavaClass.java:1242
  get_proxy_class at org/jruby/javasupport/JavaUtilities.java:34
      java_import at file:/Users/headius/projects/jruby/lib/jruby.jar!/jruby/java/core_ext/object.rb:26
              map at org/jruby/RubyArray.java:2412
      java_import at file:/Users/headius/projects/jruby/lib/jruby.jar!/jruby/java/core_ext/object.rb:22
           (root) at -e:1
```

However, if we access it using the Java namespace (as a lazy constant), we do not include this information:

```
ext-jruby-local ~/projects/jruby $ jruby -e "require 'blah.jar'; java_import Java::Blah"
NameError: cannot link Java class Blah
  get_top_level_proxy_or_package at org/jruby/javasupport/JavaUtilities.java:49
                   const_missing at file:/Users/headius/projects/jruby/lib/jruby.jar!/jruby/java/java_module.rb:4
                          (root) at -e:1
```

The error logic for both paths should be unified so we have consistent errors.

headius:
Changes over the years have improved this in some ways and degraded it in others.

The second form now includes information about the `NoClassDefFound` preventing `Blah` from linking:

```
$ jruby -e 'Java::Blah'
NameError: cannot link Java class Blah (java.lang.NoClassDefFoundError: Foo)
[...]

PR #8304, opened 2024-06-26
Eager initialize java_import'ed classes

### LINKED ISSUE #8299: Linkage errors from java_import are lazy
As of JRuby 9.3.0.0, linkage errors that would occur during `java_import` have become lazy, happening only after the import upon the first use of the class.

Using a modified version of the case from #799, the `blah.Blah` class can be `java_import`ed without error, but eventually fails due to the missing `blah.Foo` class:

```
$ (chruby jruby-9.3.14.0 ; jruby -e 'java_import "blah.Blah"')
$ (chruby jruby-9.3.14.0 ; jruby -e 'java_import "blah.Blah"; Blah.new')
Unhandled Java exception: java.lang.NoClassDefFoundError: blah/Foo
java.lang.NoClassDefFoundError: blah/Foo
               <clinit> at blah/Blah.java:5
```

I think this was probably a simple mistake in the porting of `java_import` from Ruby to Java in #6661 at the following line, which defers class initialization for this path:

https://github.com/jruby/jruby/commit/5a8ef9fea5333abae43c9917676b1df75e7aa22a#diff-cabfd925021b0b615287200ad6cf2469755af74bf4dfd560ea31161b22fda59eR116

We probably should fix this to eagerly initialize `java_import` classes so that linkage errors show up sooner.

headius:
This was originally done on purpose while working toward the possibility of lazy imports, but it's better for now to just let the class initialize and fail fast if there are linkage errors.

Fixes #8299

Resolves regression discussed in [#799](https://github.com/jruby/jruby/issues/799#issuecomment-2187125505) that prevented consistent error messages for unlinkable classes.

### DIFF
--- a/core/src/main/java/org/jruby/javasupport/ext/Module.java
+++ b/core/src/main/java/org/jruby/javasupport/ext/Module.java
@@ -113,7 +113,7 @@ public class Module {
             if (className.contains("::")) {
                 throw runtime.newArgumentError("must use Java style name: " + className);
             }
-            javaClass = Java.getJavaClass(runtime, className, false); // raises NameError if not found
+            javaClass = Java.getJavaClass(runtime, className); // raises NameError if not found
         } else if (klass instanceof JavaPackage) {
             throw runtime.newArgumentError("java_import does not work for Java packages (try include_package instead)");
         } else if (klass instanceof RubyModule) {


headius:
New output for unlinkable `java_import` shows consistency with the `method_missing` and `const_missing` forms:

```
$ jruby -e 'java_import "blah.Blah"'
NameError: cannot link Java class blah.Blah (java.lang.NoClassDefFoundError: blah/Foo)
  java_import at org/jruby/javasupport/ext/Module.java:99
  java_import at uri:classloader:/jruby/java/core_ext/object.rb:3
       <main> at -e:1
$ jruby -e 'Java::blah.Blah'
NameError: cannot link Java class blah.Blah (java.lang.NoClassDefFoundError: blah/Foo)
  method_missing at org/jruby/javasupport/JavaPackage.java:253
          <main> at -e:1
$ jruby -e 'Java::blah::Blah'
NameError: cannot link Java class blah.Blah (java.lang.NoClassDefFoundError: blah/Foo)
  const_missing at org/jruby/javasupport/JavaPackage.java:126
         <main> at -e:1
```

PR #8306, opened 2024-06-27
Properly fall back if can't setAccessible

headius:
Changes in e8dbe9c26 broke this fallback by silencing errors from the old setAccessible (as part of becoming more module-friendly). This patch checks if the setAccessible was successful and returns null for any fields that could not be made accessible. This in turn allows the USABLE fallback to properly reflect actual usability.


headius:
This PR was made against 9.3 (since it broke at 9.3.0.0) but may not see a release there for some time. Setting milestone to 9.4.8.0 because that's where most users will end up seeing it.

PR #8326, opened 2024-07-31
Speed up incremental builds

headius:
This is a series of patches to make a fast-as-possible incremental build profile.

## Reducing the build

Currently when doing a default build of JRuby, the three modules "core" (jruby-base, the actual JRuby sources), "shaded" (jruby-core, the combined JRuby and dependencies jar), and "lib" (jruby-stdlib) are always build. The "lib" module is expensive to build because it launches JRuby to conditionally fetch and unpack or install gems to populate the standard library. This can become tedious when iterating on source-only changes to JRuby.

## Making jar builds faster

The core and shaded jar files are always built with full compression, which is unnecessary and slow for local incremental builds. The shaded jar also builds a shaded sources jar which is useless for local development and produces warnings about duplicate files. We can make these jars faster to build and omit unnecessary bits.

### DIFF
--- a/BUILDING.md
+++ b/BUILDING.md
@@ -31,6 +31,18 @@ This will run the default "install" goal (`mvn install`) and will do all of the 
 
 The environment is now suitable for running Ruby applications.
 
+If you have Maven installed in your PATH, you can just use `mvn` instead of `./mvnw`.
+
+Incremental Builds
+------------------
+
+When working on JRuby sources, it is helpful to incrementally rebuild only the `lib/jruby.jar` file rather than also
+re-assembling the standard library. You can add `-Dcore` to the `mvn` command line to speed up incremental builds:
+
+```
+./mvnw -Dcore
+```
+
 Running JRuby
 -------------
 
--- a/pom.rb
+++ b/pom.rb
@@ -160,10 +160,17 @@ project 'JRuby', 'https://github.com/jruby/jruby' do
                    :phase => 'site-deploy' )
   end
 
-  modules [ 'shaded', 'core', 'lib' ]
-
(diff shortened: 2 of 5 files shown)


headius:
The best way to conditionally active the "lib" module I have found is to only add it to the active modules when a property "core" is unset. This allows a developer to pass "-Dcore" to `mvn` to avoid adding and building the "lib" module.

For me this reduces the zero-change rebuild time from 12-13s down to 6.5-7.5s.

I believe this does not impact any other build configurations since the "core" property will always be unset by default.

headius:
@mkristian If you have any suggestions here I would appreciate the help! I am having some difficulty figuring out how to disable jar compression; nothing I do seems to turn it off.

Also, it seems I can only disable the build of the shaded sources jar at the top-level configuration. I've pushed a patch for you to review. Obviously we don't want to break maven central pushing, so we need the sources jar to be generated and deployed during release.

cc @enebo for any concerns or ideas

enebo:
I remember a few years ago `mvn -pl core` worked then stopped at some point.  I think it still compressed the jars though?

headius:
@enebo `mvn -pl core` still "works", but doesn't fully rebuild `lib/jruby.jar`. When we moved the shaded jar build to `./shaded` we broke the ability to just rebuild the `./core` module. The former is now `jruby-core` and produces the shaded `lib/jruby.jar` file and the latter is now `jruby-base` unshaded; therefore in order to get the equivalent of "just rebuild the lib/jruby.jar" you must do `mvn -pl core -pl shaded`. The change here basically makes that into `mvn -Dcore`.

PR #8332, opened 2024-08-01
Allow nanosecond precision for Kernel#sleep

jpcamara:
* Previously sleep would only convert into milliseconds, so something like 0.0001 would get rounded to 0. 0 was a no-op in RubyKernel#sleep.

* Now it works in nanoseconds internally. RubyThread#sleep has also been expanded to have an overloaded signature which takes both milliseconds and nanoseconds. This supports existing calls expecting milliseconds only, while supporting nanosecond precision for sleep.

### DIFF
--- a/core/src/main/java/org/jruby/RubyKernel.java
+++ b/core/src/main/java/org/jruby/RubyKernel.java
@@ -48,6 +48,7 @@ import java.math.BigInteger;
 import java.util.ArrayList;
 import java.util.Arrays;
 import java.util.Map;
+import java.util.concurrent.TimeUnit;
 
 import jnr.constants.platform.Errno;
 import jnr.posix.POSIX;
@@ -837,27 +838,30 @@ public class RubyKernel {
 
     @JRubyMethod(module = true, visibility = PRIVATE)
     public static IRubyObject sleep(ThreadContext context, IRubyObject recv, IRubyObject timeout) {
-        long milliseconds = (long) (RubyTime.convertTimeInterval(context, timeout) * 1000);
+        long nanoseconds = (long) (RubyTime.convertTimeInterval(context, timeout) * 1_000_000_000);
         // Explicit zero in MRI returns immediately
-        if (milliseconds == 0) return RubyFixnum.zero(context.runtime);
+        if (nanoseconds == 0) return RubyFixnum.zero(context.runtime);
 
-        return sleepCommon(context, milliseconds);
+        return sleepCommon(context, nanoseconds);
     }
 
-    private static RubyFixnum sleepCommon(ThreadContext context, long milliseconds) {
-        final long startTime = System.currentTimeMillis();
+    private static RubyFixnum sleepCommon(ThreadContext context, long nanoseconds) {
+        long milliseconds = TimeUnit.NANOSECONDS.toMillis(nanoseconds);
+        long remainingNanos = nanoseconds - TimeUnit.MILLISECONDS.toNanos(milliseconds);
+
+        final long startTime = System.nanoTime();
         final RubyThread rubyThread = context.getThread();
 
         boolean interrupted = false;
         try {
             // Spurious wakeup-loop
             do {
-                long loopStartTime = System.currentTimeMillis();
+                long loopStartTime = System.nanoTime();
 
-                if (!rubyThread.sleep(milliseconds)) break;
+                if (!rubyThread.sleep(milliseconds, remainingNanos)) break;
 
-                milliseconds -= (System.currentTimeMillis() - loopStartTime);
-            } while (milliseconds > 0);
+                nanoseconds -= (System.nanoTime() - loopStartTime);
+            } while (nanoseconds > 0);
         } catch (InterruptedException ie) {
             // ignore; sleep gets interrupted
             interrupted = true;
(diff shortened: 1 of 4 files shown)


headius (changes requested):
Looks great, and thank you for preserving the public interface of sleep(millis)! I requested one change: move the test into rubyspec so it will be available for all implementations.

headius:
There's one failure in the JRuby suite. I'm looking into it.

headius:
The failure is happening in a fairly simple test: a thread sleeps and is woken up one it appears to be asleep. This works fine for the "sleep forever" case. In the "sleep for 100s" case, it doesn't wake up. I do not see anything wrong with your PR that would cause this.

headius:
Ok I think the math is wrong on resuming the sleep on spurious wake-up.

In RubyKernel#sleepCommon, it checks that the amount of time slept was >= the amount requested. It sleeps, subtracts the total amount of time slept, and then checks the result. If the requested time has not elapsed (and the thread was not forcibly interrupted, which skips this) it goes back into the sleep for the remaining time.

With your change, the total number of nanoseconds gets decremented, but it goes back into the sleep with exactly the same amount of requested time. In the failing test, after the first wake-up, it goes back into sleep again for another 100s. The test only allows this to continue for 10s, so it fails.

Please review the logic for decrementing the remaining time and sleeping for that time. 😀

headius:
Now here's the really shocking part: only this one JRuby-specific test caught the bug, and I think most folks would agree this is REALLY big bug. 🤦

@jpcamara Could you remove the JRuby test and turn it into a rubyspec as well?

jpcamara:
facepalm

It'll probably take me a few days to get back to this, but I will fix that and move the spec.

jpcamara:
> move the test into rubyspec so it will be available for all implementations.

Where do I move it to make it part of rubyspec?

headius:
@jpcamara Hmm I am not sure if it is more a sleep spec or a wakeup spec. 🤔 Probably wakeup so spec/ruby/thread/wakeup_spec.rb.

headius:
Oh FYI the spec/ruby dir is regularly synced with the ruby/spec repo.

headius:
@jpcamara Do you think you'll be able to revisit this soon? It would be nice to get it landed.

jpcamara:
Sorry - i've been away on vacation and I've had trouble getting JRuby setup on my personal laptop.

```
[INFO] JRuby Lib Setup .................................... FAILURE [ 12.328 s]
[INFO] ------------------------------------------------------------------------
[INFO] BUILD FAILURE
[INFO] ------------------------------------------------------------------------
[INFO] Total time:  19.428 s
[INFO] Finished at: 2024-08-27T22:04:47Z
[INFO] ------------------------------------------------------------------------
[ERROR] Failed to execute goal on project jruby-stdlib: Could not resolve dependencies for project org.jruby:jruby-stdlib:jar:9.4.9.0-SNAPSHOT: Could not find artifact rubygems:jruby-openssl:gem:0.15.0 in mavengems (mavengem:https://rubygems.org) -> [Help 1]
```

Haven't had a chance to figure out what's causing it, so I'll have to wait till i'm back to finish it out.

headius:
Could it simply be that you were offline when it tried to download a dependency? That gem appears to be published just fine: https://rubygems.org/gems/jruby-openssl/versions/0.15.0-java

jpcamara:
@headius sorry it took so long! I think this was the primary issue https://github.com/jruby/jruby/commit/49ba0dfbc42978ff5a133da2762c500f7a8edb61

I'm gonna see if there are any additional specs I could write to verify the behavior

headius:
I've approved the CI run and we'll see how it looks. Thanks for following up!

jpcamara:
Ugh, wrong syntax for the spec in `spec/ruby/core/thread/wakeup_spec.rb`. I couldn't get those specs to run locally at first but I think I missed the docs on running specs here https://github.com/jruby/jruby/blob/master/BUILDING.md#setup-testing. I'll give those a try and get it fixed.

The wakeup spec that had been breaking is fixed with this, at least.

jpcamara:
Ok, updated - `jruby spec/mspec/bin/mspec ci spec/ruby/core/thread/wakeup_spec.rb` is passing locally

jpcamara:
I also didn't know what would be a valuable extra spec to add. The `wakeup` spec handles the problem nicely. Still isn't ideal that it's the one spec that caught the early wakeup issue though

jpcamara:
@headius I see a few failures, but I don't think they're related to my code. You'd know best. If somehow I broke those, I just don't know at this point 🥴

headius:
I will take a look at the failures but I assume they are unrelated. If they were your fault I would expect to see a lot more tests failing. Thanks for your efforts on this, we'll get it merged in soon!

headius:
I'm re-running them. The tests that failed are across multiple JDK versions and suites, combinations that passed elsewhere, so I suspect something transient broke them. If they fail again I'll dig deeper.

headius:
Two failed again with errors that don't seem related to your change at all. I'm going to go ahead and merge and we'll sort it out on master if it's a real failure.

headius:
Thanks for sticking with it!

ioquatix:
Thanks!

Issue #8327, opened 2024-07-31
Clean up and Rubify all pom.rb files


headius:
We have many pom.rb files with peculiar code styles, mostly stemming from changing coding conventions over time and improvements in Ruby like the new keyword syntax. There's also unusual indentation and inconsistent use of symbols versus strings and colon-hashes versus hashrockets. This makes maintaining the build scripts annoying and if we start using newer conventions in part of a file it looks inconsistent with the rest of the file.

I propose that we do a sweep for JRuby 10 to clean up Ruby syntax in these files and consistently apply modern Ruby coding conventions.

This will necessarily split the history of these files, but it's a band-aid that needs to be ripped off eventually. Doing this in JRuby 10 makes sense because there will be other large-scale changes for Java 17+ and modularity that will also make forward-merging difficult or impossible.

headius:
@mkristian @enebo Another effort I think would be worth it for JRuby 10. This would make the build more approachable both for us and for new users getting familiar with JRuby development. I think we've all been frustrated from time to time with the inconsistent coding conventions throughout our build scripts.

enebo:
Stylistically I don't have any opinion.  Making it one style will not hurt though.  My suggestion is I forget which pom.rb contains things like core has java daps and lib has ruby gem deps.  Not sure if we could consolidate these into a single top-file but it might make it simpler if we could what I consider config data (e.g. things we routinely update) in one place.

headius:
That may be possible but I'm not sure of the best approach. Maven really wants submodules in their own dirs with their own poms, so this isn't as much a faster of our polyglot ruby scripts as it is just the way Maven structures things.

headius:
Ideas from Matrix discussion:

* Move some content out to requireable files, like under `build/core-dependencies.rb` or similar. These can be structured like configuration to make it easier to update dependencies, default gems, etc.
* Move programmatic model manipulation out to requireable files. This includes a lot of logic from `lib/pom.rb` for unpacking and installing default/bundled gems.
* Aim for pom.rb files to contain only Maven structure and limited programmatic model building loaded in from `build/*` files.

headius:
Most of this happened by using Rubocop to clean up pom.rb files in #8975 (and a similar PR for 9.4.x in #8974).

Any additional cleanup can be done as one-off fixes.

PR #8333, opened 2024-08-01
Clean up and Rubify all pom.rb files

woodie:
Do a sweep for JRuby 10 to clean up Ruby syntax in these files and consistently apply modern Ruby coding conventions. See https://github.com/jruby/jruby/issues/8327 for details


woodie:
How do we feel about linter rules? @headius

headius:
The changes made are generally what I had in mind, but there are some peculiarities that may be worth discussing:

* I question the value of replacing double-quoted strings with single-quoted strings but that is perhaps a minor issue. This one rule does comprise the bulk of changes, though (excluding whitespace).
* Most of the indentation changes seem to be in keeping with "standard" Ruby formatting, but those that reformat a multi-line call seem less readable to me:
  ```
    plugin(:compiler,
      encoding: "utf-8",
      verbose: "false",
      ...
  ```
  I would not say the original formatting, with all arguments aligned (possibly not on 2-space offsets) is the best way either, though. Interestingly, when formatting a collection, it does adhere to the same-line formatting:
  ```
      execute_goals("attach-artifact",
        id: "attach-checksums",
        phase: :package,
        artifacts: [{file: "${project.build.directory}/jruby-dist-${project.version}-bin.tar.gz.sha256",
                     classifier: :bin,
                     type: "tar.gz.sha256"},
  ```
* standardrb (or perhaps Rubocop) does not appear to have a strong opinion about `do...end` vs `{...}` for blocks. As the reformatted files are 100% a Maven DSL, I have wondered if they might be better served with the latter:
  ```
  profile "jruby.sh" do
    activation do
      file(missing: "../bin/jruby")
    end
  end
  ```
  vs
  ```
  profile("jruby.sh") {
    activation {
      file(missing: "../bin/jruby")
    }
  }
  ```
  If we are to make this change globally I think it's worth debating changes like this.
* We have some inconsistency to address ourselves in where we use Symbols versus Strings. Strings will work in all cases, but Symbols frequently require additional quoting, reducing their utility. It would seem that Strings are the better choice, and this would need to be done manually:
  ```
  phase :clean do
    plugin :dependency do
      execute_goals(*jni_config)
    end
  end

  profile "native" do
    activation do
  ```
  vs
  ```
  phase("clean") {
    plugin("dependency") {
      execute_goals(*jni_config)
    }
  }

  profile("native") {
    activation {
  ```

Obviously the elimination of hashrockets (`=>`) is very nice, and the inconsistency there is one of the things that drove me to finally make this push.

Much of what this did automatically seems benign and basically good, but as evidenced by the CI result it was apparently not zero-sum.

I would like some input from @enebo, @kares, and anyone else who might be tracking this.

woodie:
Basically, I ran `standardrb` (which is rubocop config that likes double quotes) and then `rubocop` with the default config. On top of that I did manually update the hash-rocket syntax, but everything else is just the linters.

Depending on which linter we run, we'll get different results for spacing and single/double quotes. I hope we can pick some existing standard to impose now and avoid crafting our own rubocop rules.

> those that reformat a multi-line call seem less readable to me. I would not say the original formatting, with all arguments aligned (possibly not on 2-space offsets) is the best way either, though. Interestingly, when formatting a collection, it does adhere to the same-line formatting

IMHO adding  spaces to create alignment ASCII art can be a bad habit. The `standardrb` approach can be less "fun" but should be best for a large codebase with many users.

> to have a strong opinion about do...end vs {...} for blocks.

I prefer `do/end` except for on one-liners, but we can certainly construct our own rubocop config to enforce whatever we want.

> We have some inconsistency to address ourselves in where we use Symbols versus Strings.

We still end up with a symbol when when the keys have special characters, but it can look messy.
```rb
:foo => :foo 
:"foo-bar" => :"foo-bar" 
```

> Much of what this did automatically seems benign and basically good, but as evidenced by the CI result it was apparently not zero-sum.

I hope we can nail down the hash-key format we like (which I assume is the issue), and then make sure we have a working build.

enebo:
I think omakase might be a reasonable option as it is a tiny rule set and
will likely help to kill people using rubocop default cops

woodie:
Agreed that rubocop defaults are bad, but not having a linter is more bad.

bench/pom.rb
```diff
@@ -1,55 +1,55 @@
-version = ENV['JRUBY_VERSION'] ||
-  File.read(File.join(basedir, '..', 'VERSION')).strip
+version = ENV["JRUBY_VERSION"] ||
+          File.read(File.join(basedir, "..", "VERSION")).strip
```
kares, line 2: really rubocop default configuration uses `'` ?
been operating for years based on (GH's?) Ruby guide to prefer `'` unless interpolation is used.
why would they diverge from a default Ruby style guide that has been around prior... :unamused:

woodie:
Let's focus on upgrading the has rocket syntax (possibly converting some string keys to symbols) and determine a linter strategy in a follow up PR.

headius:
I have started a new attempt at this that is pure RuboCop here: #8974.

PR #8335, opened 2024-08-01
Lint pom files with standardrb

woodie:
Apply safe lint with standardrb.
`find . -type f | grep pom.rb | xargs standardrb --fix`


woodie:
The first commit passed all checks, the the second commit (adding some whitespace) failed the first four checks. @headius

PR #8337, opened 2024-08-02
upgrade rexml to 3.3.4

jsvd:
This upgrade addresses recent security issues in rexml. the issues are not critical, so this is not urgent.


headius:
Failures don't appear to be related. Thank you!

Issue #8346, opened 2024-09-10
Compiled jsr223 scripts caused an ArrayIndexOutOfBounds exception


jimtng:
In an OSGi environment, jruby is invoked through `new org.jruby.embed.jsr223.JRubyEngineFactory()`. When the script was first compiled and then some context added to the scriptengine, calling `eval()` on the compiled script caused an AIOBE.

This occurred on jruby 9.4.8.0 but I suspect this issue exists in prior versions too. It just happened to be noticed now because we've just started using compiled scripts.

The exception:
```
javax.script.ScriptException: java.lang.ArrayIndexOutOfBoundsException: Index 2 out of bounds for length 2
	at org.jruby.embed.jsr223.JRubyEngine.wrapException(JRubyEngine.java:263) ~[?:?]
	at org.jruby.embed.jsr223.JRubyEngine.doEval(JRubyEngine.java:104) ~[?:?]
	at org.jruby.embed.jsr223.JRubyCompiledScript.eval(JRubyCompiledScript.java:78) ~[?:?]
	at org.openhab.core.automation.module.script.internal.handler.AbstractScriptModuleHandler.eval(AbstractScriptModuleHandler.java:214) ~[?:?]
	at org.openhab.core.automation.module.script.internal.handler.ScriptActionHandler.lambda$0(ScriptActionHandler.java:80) ~[?:?]
	at java.util.Optional.ifPresent(Optional.java:178) [?:?]
	at org.openhab.core.automation.module.script.internal.handler.ScriptActionHandler.execute(ScriptActionHandler.java:78) [bundleFile:?]
	at org.openhab.core.automation.internal.RuleEngineImpl.executeActions(RuleEngineImpl.java:1299) [bundleFile:?]
	at org.openhab.core.automation.internal.RuleEngineImpl.runRule(RuleEngineImpl.java:1057) [bundleFile:?]
	at org.openhab.core.automation.internal.TriggerHandlerCallbackImpl$TriggerData.run(TriggerHandlerCallbackImpl.java:86) [bundleFile:?]
	at java.util.concurrent.Executors$RunnableAdapter.call(Executors.java:539) [?:?]
	at java.util.concurrent.FutureTask.run(FutureTask.java:264) [?:?]
	at java.util.concurrent.ScheduledThreadPoolExecutor$ScheduledFutureTask.run(ScheduledThreadPoolExecutor.java:304) [?:?]
[...]

PR #8347, opened 2024-09-10
Properly set up binding for compiled JSR-223 scripts

headius:
The `preEval` calls are run later in `doEval` immediately before executing. Doing it here appears to have caused some issue in either the parse or the set-up of the scoping used by the parsed result (see jruby/jruby#8346). Removing these pre-parse `preEval` calls fixes the issue.

Fixes jruby/jruby#8346.

Edit: It turns out the `preEval` is actually necessary for the parse to pick up values properly, so instead we do a full `preEval` and `postEval` around the parse and again around the eval.

### DIFF
--- a/core/src/main/java/org/jruby/embed/jsr223/JRubyCompiledScript.java
+++ b/core/src/main/java/org/jruby/embed/jsr223/JRubyCompiledScript.java
@@ -57,7 +57,11 @@ public class JRubyCompiledScript extends CompiledScript {
         this.container = container;
         this.engine = engine;
         Utils.preEval(container, engine.getContext());
-        unit = container.parse(script);
+        try {
+            unit = container.parse(script);
+        } finally {
+            Utils.postEval(container, engine.getContext());
+        }
     }
 
     JRubyCompiledScript(ScriptingContainer container, JRubyEngine engine, Reader reader) throws RaiseException {
@@ -65,7 +69,11 @@ public class JRubyCompiledScript extends CompiledScript {
         this.engine = engine;
         String filename = System.getProperty(ScriptEngine.FILENAME);
         Utils.preEval(container, engine.getContext());
-        unit = container.parse(reader, filename, Utils.getLineNumber(engine.getContext()));
+        try {
+            unit = container.parse(reader, filename, Utils.getLineNumber(engine.getContext()));
+        } finally {
+            Utils.postEval(container, engine.getContext());
+        }
     }
 
     @Override


jimtng:
I've tested it with the preEval + postEval in the constructor and that works too, thank you!

jimtng:
I managed to create a test that failed when compiled, but not when eval'ed. However, this test didn't pass on this PR:

```java
package org.jruby.test;

import static org.junit.jupiter.api.Assertions.assertEquals;
import org.junit.jupiter.api.Test;

public class JRubyCompileTest {
    private String theTest(String input, boolean compile) throws javax.script.ScriptException {
        var script = """
                var = "#{ctx["input"]}"
                """;
        
        var engine = new org.jruby.embed.jsr223.JRubyEngineFactory().getScriptEngine();
        javax.script.CompiledScript compiled = null;
        if (compile) {
            compiled = ((javax.script.Compilable) engine).compile(script);
        }

        var context = engine.getContext();
        var contextNew = new java.util.HashMap<>();
        contextNew.put("input", input);
        context.setAttribute("ctx", contextNew, javax.script.ScriptContext.ENGINE_SCOPE);
        context.setAttribute("input", input, javax.script.ScriptContext.ENGINE_SCOPE);

        return (String)(compile ? compiled.eval() : engine.eval(script));
    }

    @Test
    public void setAttributeOnCompiledScript() throws javax.script.ScriptException {
        assertEquals("compiled test", theTest("compiled test", true));
    }

    @Test
    public void setAttributeOnNonCompiledScript() throws javax.script.ScriptException {
        assertEquals("not compiled", theTest("not compiled", false));
    }
}
```

Note that this PR did solve my actual issue though. Strange.

headius:
I dug a bit deeper into the code and I think this is probably a current limitation of how compiled scripts work.

One important piece that comes out of the parser is the various scaope layout objects, StaticScope instances and their runtime instantiation, DynamicScope. For an execution context in the JRuby 223 engine, it's a MultiVarDynamicScope that can grow over time if needed.

During interpretation, some of this structure is prepared for the eval via `preEval` which sets up the binding and propagates values into it. Variable names without any leading sigil are treated as persistent local variables, and set into the ManyVarDynamicScope for the upcoming script to use. During compilation, this setup is done twice: once in the constructor around the parse, and once again before executing the pre-parsed code.

I suspect in the compiled case, the local variable bindings are getting messed up in some way, probably because it's a new scope and not the one we parsed with. Such bindings may have worked with the old code (unpatched) because it never `postEval`ed after the parse, so the variables were not cleared out, but this led to your other issues when we did this double initialization.

Given that there's certain details the parser will want from the top-level ManyVarsDynamicScope, it may be difficult to get compilation to work for cases that want to pass in local variables. If we don't set up the scope, the parser will treat them all as new variables or as method calls. If we set up the scope both times, the parser will know about them but then not be able to access them from a different scope.

This code is a bit of a dark corner for me... contributed many years ago and always a bit of a mystery.

headius:
I believe @jimtng has worked around the limitation with top-level variables not translating correctly between the precompiled source and the eventual eval, which as I describe above may simply be a limitation of precompiling. I think we should merge this for now and if we want to do something more exotic with the compilation later on we can address it then.

jimtng:
Thanks, @headius

Issue #8156, opened 2024-03-19
JRuby adds Java proxy classes to the Java module even if they are not from JRubyClassLoader


barrkel:
JRuby version 9.3.13.0
Platform 6.5.13-1rodete2

Expected behavior:
- There is a class `org.example.C` which is in a jar which is not on the class path, call this `C_jar`
- There is a class `org.example.C` which is in a jar which is on the class path, call this `C_classpath`
- `C_jar` is loaded via `URLClassLoader` and instantiated in Java code. This is done to isolate the different versions of `C`.
- Use Ruby to evaluate an object expression which results in an instance of an object with the class `C_jar`
- Use the `Java` module to reference the class `C`, expecting to get `C_classpath` (i.e. `Java::OrgExample::C`)

Actual behavior:
- Using the `Java` module syntax to access `C` results in `C_jar` 

Workaround:
- Use the `Java` module syntax to access `C` before any Ruby expression which evaluates to an object of class `C_jar`

In [Java.java](https://github.com/jruby/jruby/blob/4fdce77fa6e1f0fbba6c4771b4d1e9fc7e92ee3a/core/src/main/java/org/jruby/javasupport/Java.java#L508), `generateClassProxy` calls `addToJavaPackageModule` without checking the class loader of the class it generated a proxy for. Effectively, manipulating values from non-`JRubyClassLoader` classes pollutes the `Java` module with incompatible classes.

headius:
Interesting! Do you think you can come up with a small reproduction for us?

barrkel:
[java-module-pollution.zip](https://github.com/jruby/jruby/files/14758799/java-module-pollution.zip)

Steps:

1. Unzip the attached zip and enter the directory.
2. Run `./build.sh` to compile the Java classes and build the jar; `javac` and `jar` need to be on the path.
3. Run `./repro.rb broken` with `jruby` on the path and observe:
```
Expecting 'isolated in jar': isolated in jar
Expecting 'packaged with app': isolated in jar
```
4. Run `./repro.rb workaround` and observe:
```
Expecting 'isolated in jar': isolated in jar
Expecting 'packaged with app': packaged with app
```

barrkel:
Having thought about it a little bit more:

- There's a potential backward compatibility issue; it's possible some people have relied on `Java::` access to classes they loaded from a jar similar to how the repro works. This may have worked OK if they didn't have duplicate classes on the app class path.
[...]

PR #8350, opened 2024-09-16
do not expose Java proxy generated classes in Ruby land

### LINKED ISSUE #8349: internal proxy class is stored in Ruby land and prints warning
There's a (minor) Java integration regression since 9.4.7.0 caused by https://github.com/jruby/jruby/commit/31f1ed6f7458acc42a545a032424580f5b22d746 (fix for https://github.com/jruby/jruby/issues/8156)

**Environment Information**

* JRuby version **9.4.8.0**

Consider the sample script trying to utilize the "duck-typing" Java auto-interface impl feature:

```ruby
class Run1
  class << self
    def run; puts self end
  end
end

class Run2 < Run1; end
class Run3 < Run1; end

#java.lang.Thread.new(Run1).start
java.lang.Thread.new(Run2).start
java.lang.Thread.new(Run3).start
```
The script will print a Ruby constant initialization warning (will get printed the amount of times there are sub-classes since the impl proxy will be cached once per RubyClass):
```
impl_repro.rb:12: warning: already initialized constant org.jruby.gen::InterfaceImpl1353309827
```

The internal (interface implementation) proxy class ends up being stored in Ruby land, this did not happen previously.

p.s. There might be other Java proxy path worth checking, whether they end up eagerly setting their class proxy wrappers in Ruby land, effectively preventing garbage collection of such classes...

kares:
this PR changes some of the bits from: https://github.com/jruby/jruby/commit/31f1ed6f7458acc42a545a032424580f5b22d746

e.g. it effectively makes sure `getProxyClass` (which might be used in user ext) behaves as before - not setting a Java proxy class constant upon every invocation...

fixes #8349

### DIFF
--- a/core/src/main/java/org/jruby/java/proxies/JavaInterfaceTemplate.java
+++ b/core/src/main/java/org/jruby/java/proxies/JavaInterfaceTemplate.java
@@ -306,9 +306,12 @@ public class JavaInterfaceTemplate {
 
     private static IRubyObject newInterfaceProxy(final IRubyObject self) {
         final RubyClass current = self.getMetaClass();
+        final Ruby runtime = current.getRuntime();
         // construct the new interface impl and set it into the object
         Object impl = Java.newInterfaceImpl(self, Java.getInterfacesFromRubyClass(current));
-        IRubyObject implWrapper = Java.getInstance(self.getRuntime(), impl);
+        RubyClass proxyClass = (RubyClass) Java.getProxyClass(runtime, impl.getClass(), false);
+        // we do not want the (InterfaceImpl) proxy-class in this case to be set on the Ruby side
+        IRubyObject implWrapper = Java.getInstanceInternal(runtime, impl, proxyClass, false);
         JavaUtilities.set_java_object(self, self, implWrapper); // self.dataWrapStruct(newObject);
         return implWrapper;
     }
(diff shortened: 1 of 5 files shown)


kares:
right, so this is turning out to be a mess - current approach is not sufficient and we basically restore more of https://github.com/jruby/jruby/commit/31f1ed6f7458acc42a545a032424580f5b22d746 to get proxy class initialization ~ in 9.4.6.0

first, the https://github.com/jruby/jruby/pull/8208 causes the proxy class to be assigned multiple times (it's just that the warnings do not surface in CI much and JRuby also has checks to ignore them due potential concurrent loading e.g. https://github.com/jruby/jruby/pull/8250)

second, I am not sure the fix for https://github.com/jruby/jruby/issues/8156 is legit, internally both proxy classes will be present it's just a question which one will "surface" first.

headius:
@kares The main problem here is just the proxy classes getting into the constant table, right? Can't we just avoid that aspect and leave the rest the way it is? Store it some other way?

kares:
the thing I do not like is we're now constantly doing the checking of the proxy constant on every `getProxyClass` path, previously the setting of the constant only happened on paths that mostly executed once (e.g. on `method_missing`).

the current way I could think of (wout bigger refactoring) would be to revert https://github.com/jruby/jruby/commit/31f1ed6f7458acc42a545a032424580f5b22d746 and try implementing it without changing `getProxyClass`... unless you have a better idea I would try that (the current changes here do not work).

headius:
> unless you have a better idea I would try that

@kares Yeah go ahead and try something else. I was hopeful that that simple fix would be safe, but it seems there's this side effect we definitely don't want.

The original issue was that a random generated version of a class was added into that global namespace, masking the one that lives in JRuby's classloader hierarchy. If we can solve that in a more targeted way (comparing classloader hierarchy? some fast way to check if this class was loaded from our classloaders?) I'd be thrilled.

kares:
second attempt at https://github.com/jruby/jruby/pull/8368

PR #8364, opened 2024-10-08
Revert "Allow nanosecond precision for Kernel#sleep"

headius:
Reverts jruby/jruby#8332

Testing to see if this somehow caused flaky `getsockopt` tests we have seen recently.


headius:
Did not help.

PR #8368, opened 2024-10-10
re-implement the option to lazy bind Java:: constants

kares:
this is a second attempt at resolving (https://github.com/jruby/jruby/issues/8349) the slight regression introduced at https://github.com/jruby/jruby/commit/31f1ed6f7458acc42a545a032424580f5b22d746

the commit is reverted and the `-Xji.eager.constants` feature is re-implemented more conservatively (without effecting all `getProxyClass` callers).

also rename and cleaned up the method setting the actual proxy class constant, the tests are a bit futile but should prevent doing a similar `getProxyClass` refactoring, which would have undesired side effects.


headius:
@kares Are you happy with this one? Definitely a good idea to get this fixed in 9.4.9.0.

kares:
seems to work fine. the pre 9.4.7.0 `getProxyClass` behavior is restored and `-Xji.eager.constants=false` should have the same effect (in terms of setting the constant) as the original implementation...

headius (approved):

PR #8369, opened 2024-10-10
[feat] support clearing of thread local state

kares:
this is to support cases where e.g. a Java thread pool is running Ruby code.

if the ruby code uses any fiber/thread local state e.g. `Thread.current[:foo] = :bar` than there isn't an easy way to cleanup accumulated state in RubyThread's locals (and due the way foreign Java threads are adopted the state will stick around for when the Java thread is re-cycled).

### DIFF
--- a/core/src/main/java/org/jruby/RubyThread.java
+++ b/core/src/main/java/org/jruby/RubyThread.java
@@ -940,9 +940,18 @@ public class RubyThread extends RubyObject implements ExecutionContext {
         return RubyString.newString(runtime, rubyName);
     }
 
+    /**
+     * Returns the current Ruby thread.
+     * @param runtime
+     * @return current ruby thread
+     */
+    public static RubyThread current(final Ruby runtime) {
+        return runtime.getCurrentContext().getThread();
+    }
+
     @JRubyMethod(meta = true)
     public static RubyThread current(IRubyObject recv) {
-        return recv.getRuntime().getCurrentContext().getThread();
+        return current(recv.getRuntime());
     }
 
     @JRubyMethod(meta = true)
@@ -1027,6 +1036,30 @@ public class RubyThread extends RubyObject implements ExecutionContext {
             }
         }
         return locals;
+    }
+
+    /**
+     * Clear the fiber local variable storage for this thread.
+     * Meant for Java consumers when reusing threads (e.g. during thread pooling).
+     * @see #clearThreadLocals()
+     */
+    public void clearFiberLocals() {
+        final Map<IRubyObject, IRubyObject> locals = getFiberLocals();
+        synchronized (locals) {
+            locals.clear();
+        }
+    }
+
+    /**
+     * Clear the thread local variable storage for this thread.
+     * Meant for Java consumers when reusing threads (e.g. during thread pooling).
+     * @see #clearFiberLocals()
+     */
+    public void clearThreadLocals() {
+        final Map<IRubyObject, IRubyObject> locals = getThreadLocals();
+        synchronized (locals) {
+            locals.clear();
+        }
     }
 
     @Override
--- a/core/src/main/java/org/jruby/RubyThread.java
+++ b/core/src/main/java/org/jruby/RubyThread.java
@@ -940,18 +940,9 @@ public class RubyThread extends RubyObject implements ExecutionContext {
         return RubyString.newString(runtime, rubyName);
     }
 
-    /**
-     * Returns the current Ruby thread.
-     * @param runtime
-     * @return current ruby thread
-     */
-    public static RubyThread current(final Ruby runtime) {
-        return runtime.getCurrentContext().getThread();
-    }
-
     @JRubyMethod(meta = true)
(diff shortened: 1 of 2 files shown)


core/src/main/java/org/jruby/RubyThread.java
```diff
@@ -1029,6 +1029,15 @@ private Map<IRubyObject, IRubyObject> getThreadLocals0() {
         return locals;
     }
 
+    /**
+     * Clear the local variable storage for this thread.
+     * Meant for Java consumers when reusing threads (e.g. dur thread pooling).
+     */
+    public synchronized void clearLocalVariables() {
```
headius, line 1036: I have a problem with this name 😀

Perhaps something like `clearConcurrentLocals` or some better name that indicates these are variables local to one of the two concurrency constructs?

headius: `clearVariableStores`?

kares: I am fine with whatever name you choose from the two, originally started with simply `clearLocals`.

I also can be convinced that having 2 `clearFiberLocals` and `clearThreadLocals` would be better...

headius: > I also can be convinced that having 2 clearFiberLocals and clearThreadLocals would be better...

I do like that idea. There are bound to be use cases for clearing one and not the other, e.g. Fiber-heavy framework wants to clear just Fiber locals after processing some chunk of work.

kares: updated, please let me know what you think...

also added a convenience static `RubyThread.current(runtime)` so it's easier to retrieve the current Ruby thread on the Java side.

core/src/main/java/org/jruby/RubyThread.java
```diff
@@ -940,9 +940,18 @@ public IRubyObject getName() {
         return RubyString.newString(runtime, rubyName);
     }
 
+    /**
```
enebo, line 943: I think I would be interested in seeing this fall into org.jruby.api somewhere.  I have been trying to get all methods people would use in native extensions into this package space.  I have also been dogfooding our internals to use these same APIs when it makes sense.

org.jruby.api.Runtime#currentContext(Ruby) or something like that?  Not sure on either the class name or the method name but just putting that out there.

The other broad point is to kill off any external consumers from hitting Ruby methods at all.   I would love to never see us use Ruby as a reference anywhere some day but I think needing to get current ThreadContext will be with us for a long time.

enebo (changes requested):

core/src/main/java/org/jruby/RubyThread.java
enebo: Just to add here that 9.4 has none of the new .api packages but the plan is to backport those files so we have a migration path for 9.4 and possibly 9.3 for native extensions.

core/src/main/java/org/jruby/RubyThread.java
headius: LGTM!

core/src/main/java/org/jruby/RubyThread.java
kares: > I would love to never see us use Ruby as a reference anywhere some day

it's a multi-tenant runtime so I am not sure how that would end up being achieved...

what I can tell, as I've been on multiple bigger projects consuming and mixing Ruby and Java is that the runtime is usually something that is kept externally or queried frequently `Ruby.getGlobalRuntime` (even if it's a single runtime use-case).

> seeing this fall into `org.jruby.api` somewhere

would be nice, but it's a big chunk IMHO, the only way to cleanup the distinction would be to have interfaces on objects that Ruby users from Java encounter e.g. `RubyArray implements IRubyArray`, `RubyHash`, `RubyThread` etc. having just the one `IRubyThread` added here, with a static method, doesn't get us far...

I can remove the extra `current(final Ruby runtime)` in this PR and can continue consuming `current(IRubyObject recv)`, it's mostly for convenience

enebo: @kares yeah for embedding with multiple runtimes you have to use runtime.  I meant for internal use and java native extension gems.  Even then I know we will still need something to get TC from runtime.

Also I am only talking about this single method `current` so we don't have another way of getting TC.  If you just move this to a new class with a static method:  org.jruby.api.VM.currentThread(Ruby runtime)  then I think it is fine.

kares: I've reverted the change for now (the work-around with `RubyThread.curren(IRubyObject)` will do, it's likely a minor use-case anyways).
I think I get what you mean but I believe it deserves a proper API design, just moving a static method somewhere while still returning `RubyThread` sounds a bit half-baked and I can not commit to the full thingy atm.

kares:
marked as 9.4.9.0 would be nice to have it but feel free to post-pone...

headius:
I'm +1 to merge.

PR #8372, opened 2024-10-10
Update to ruby-maven 3.9.0

headius:
This should fix issues with the polyglot extension not loading pom.rb-formatted files at the root of the filesystem (or anywhere else).


headius:
Unfortunately, due to issues getting jar-dependencies updated (https://github.com/mkristian/jar-dependencies/issues/84) I think we need to punt this to the next release. jar-dependencies has a lot of old version dependencies and has not run integration tests in some time. I'd like to get it right when we update it.

headius:
Superseded by the jar-dependencies update in #8477.

PR #8397, opened 2024-10-30
Always set parent and name even if not eager

headius:
Setting a constant has the side effect of also setting the name when a class has none. By making JI constants lazy, we also skip the constant set, so the class remains anonymous.

$ jruby -Xji.eager.constants=false -e 'p java.lang.Runnable.name' nil

This patch restores parent module and name-setting to the last phase of Java proxy class initialization, so that the proxy class gets its real name.

### DIFF
--- a/core/src/main/java/org/jruby/RubyModule.java
+++ b/core/src/main/java/org/jruby/RubyModule.java
@@ -4995,7 +4995,7 @@ public class RubyModule extends RubyObject {
         return value;
     }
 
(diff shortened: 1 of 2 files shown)


kares (approved):
good catch, seems I should have done more testing. thanks

headius:
This is moot now as we have reverted the related PRs and will make a completely new attempt. See #8399 and linked issues and PRs.

PR #8413, opened 2024-11-06
Update specs

eregon:


eregon:
@headius Could you add tags for failing specs as documented in https://github.com/ruby/spec/wiki/Merging-specs-from-JRuby-and-other-sources#jruby in a new commit on this branch (you can push to it) and merge this PR?

headius:
I already did that with several fixes in #8411.

PR #8414, opened 2024-11-07
Pre-generate specialized object shapes

headius:
This patch generates all the specialized widths of RubyObject that we support, up to the configured maximum. This avoids racing to to generate these classes and potentially speeds up early boot stages.


headius:
Marked as draft... something's not right.

headius:
Closed because of 9.5-dev branch deletion, but will be merged into #8433.

PR #8415, opened 2024-11-09
Skip extension builds for default gems

headius:
Seems to be issues launching these subprocesses via Maven. While we investigate that, temporarily disable extension building for default gems.

The problem manifests during the default gem install of the fiddle gem, which generates a dummy Makefile on JRuby because it uses the FFI version instead. The launch of extconf.rb exhibits problems in various environments:

* In CI, it ends up running with the system default Java rather than the one configured in JAVA_HOME.
* On local Linux environments, it fails to launch due to a "bad file descriptor: /bin/sh" error.

It seems that the subprocess launch is losing environment in some broken way.

### DIFF
--- a/lib/pom.rb
+++ b/lib/pom.rb
@@ -252,6 +252,18 @@ project 'JRuby Lib Setup' do
       File.join(global_bin, "jruby#{RbConfig::CONFIG['EXEEXT']}")
     end
 
+    # Disable extension build for gems (none of ours require a build)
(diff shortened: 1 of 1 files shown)


headius:
This appears to work fine but @enebo pointed out we will eventually need this to start building exts again for the prism gem. For now, this will allow 9.4 (master) and 10 to build without running into the env issues.

PR #8425, opened 2024-11-14
Don't clear the ThreadGroup when Thread terminates

headius:
CRuby does not clear the ThreadGroup set in the Thread when the thread terminates. We do, which leads to some unpredictability in specs that check the group against a very short-lived thread. This patch removes the call to clear the thread's group.

### DIFF
--- a/core/src/main/java/org/jruby/RubyThreadGroup.java
+++ b/core/src/main/java/org/jruby/RubyThreadGroup.java
@@ -120,7 +120,6 @@ public class RubyThreadGroup extends RubyObject {
     
     public void remove(RubyThread rubyThread) {
         synchronized (rubyThread) {
-            rubyThread.setThreadGroup(null);
             rubyThreadList.remove(rubyThread);
         }
     }


headius:
The failure is intermittent but looks like this:

```
1)
Thread#group returns the thread group explicitly set for this thread FAILED
Expected nil == #<ThreadGroup:0x29d8d547>
to be truthy but was false
/home/runner/work/jruby/jruby/spec/ruby/core/thread/group_spec.rb:12:in `block in <main>'
org/jruby/RubyBasicObject.java:2655:in `instance_exec'
org/jruby/RubyArray.java:4910:in `all?'
org/jruby/RubyArray.java:1981:in `each'
/home/runner/work/jruby/jruby/spec/ruby/core/thread/group_spec.rb:3:in `<main>'
org/jruby/RubyKernel.java:1223:in `load'
org/jruby/RubyBasicObject.java:2655:in `instance_exec'
org/jruby/RubyArray.java:1981:in `each'
```

It's because in this code, sometimes the group check happens before the thread terminates, and sometimes it happens after. If it happens after, we will have cleared the group.

```ruby
it "returns the thread group explicitly set for this thread" do
  thread = Thread.new { nil }
  thread_group = ThreadGroup.new
  thread_group.add(thread)
  thread.group.should == thread_group
ensure
  thread.join if thread
end
```

PR #8447, opened 2024-11-22
Eliminate dynscope for leaf blocks with break/return

headius:
This is a patch to eliminate the need to push a new DynamicScope for leaf blocks that have non-local flow control (break or return) by using the self block's binding's DynamicScope to find the jump target.

See https://github.com/jruby/jruby/pull/8330 for original PR closed by 9.5-dev branch deletion.


headius:
Still experimental.

PR #8453, opened 2024-11-23
Fix Float#ceil and Float#floor behavior for edge cases

k77ch7:
- Ensures the following tests now pass:
  * spec/ruby/core/float/ceil_spec.rb
  * spec/ruby/core/float/ceil_float.rb


enebo:
@k77ch7 you probably saw but this regressed some Integer specs for the same methods.

headius:
Punting for extra time to get this green.

PR #8458, opened 2024-11-25
[deps] bump jruby-openssl to 0.15.2

kares:


kares:
yeah, noticed they are kind of "expected" due the checks the suite has: `if openssl?(3, 0, 0)`

JOSSL isn't really OpenSSL 3.0 compatible, but the `OPENSSL_VERSION_NUMBER` was set in a way that got the `if openssl?(3, 0, 0)` to return `true`. now it's set to return a version < 3.0.0 as this actually matters with libraries out there (tested with https://github.com/jwt/ruby-jwt).

so, it's a bit of a mess, guess we could wait a while before having this one included in JRuby for user reports or I will eventually find time to address those failures (so that at least OpenSSL 1.x compatibility is more reliable).

headius:
@kares This is ready to go then yes?

kares:
Feel free to postpone, another release is almost ready with more stuff that has been requested (and also resolves most of the "new" test failures here).l, but it might take me some time to release...

headius:
@kares okay thanks for the update. Let me know if there's any way I can help.

headius:
Based on conversation on Matrix I believe this is ready to merge, and 9.4.10.0 is out so we'll have this in 9.4.11.0.

Issue #8457, opened 2024-11-24
Class#subclasses slows down with larger sets


headius:
When Class#subclasses returns a longer list, we appear to slow down more relative to CRuby.

With the four classes from Numeric:

```
$ ruby -rbenchmark -e '5.times { puts Benchmark.measure { 10_000_000.times { Numeric.subclasses } } }'            
  0.840000   0.040000   0.880000 (  0.719652)
  0.610000   0.000000   0.610000 (  0.597529)
  0.610000   0.010000   0.620000 (  0.598173)
  0.600000   0.000000   0.600000 (  0.600692)
  0.650000   0.010000   0.660000 (  0.600609)
$ chruby-exec 3.3 -- ruby -rbenchmark -e '5.times { puts Benchmark.measure { 10_000_000.times { Numeric.subclasses } } }'
  0.892543   0.003522   0.896065 (  0.896127)
  0.880852   0.002772   0.883624 (  0.883620)
  0.902136   0.003467   0.905603 (  0.905592)
  0.902449   0.003406   0.905855 (  0.905844)
  0.888070   0.003159   0.891229 (  0.891218)
```

And the 83 subclasses (82 in CRuby) of Object (note reduced iterations in benchmark):

```
$ ruby -rbenchmark -e '5.times { puts Benchmark.measure { 1_000_000.times { Object.subclasses } } }'         
  2.090000   0.020000   2.110000 (  1.885031)
  1.780000   0.020000   1.800000 (  1.813361)
  1.800000   0.010000   1.810000 (  1.821201)
  1.760000   0.010000   1.770000 (  1.754643)
  1.810000   0.010000   1.820000 (  1.777603)
$ chruby-exec 3.3 -- ruby -rbenchmark -e '5.times { puts Benchmark.measure { 1_000_000.times { Object.subclasses } } }'
  1.006108   0.018142   1.024250 (  1.024307)
  0.989379   0.014869   1.004248 (  1.004562)
  1.006066   0.018692   1.024758 (  1.024845)
  1.022598   0.022539   1.045137 (  1.091699)
  0.985969   0.019927   1.005896 (  1.006522)
```

Ignoring the overhead of the benchmark, it appears CRuby slows down about 11x and we slow down 30x for an increased class count of 20x.

headius:
The fix for this turned out to be fairly non-invasive so I've retargeted JRuby 9.4.10.0.

headius:
[...]

PR #8462, opened 2024-11-25
Optimize Class#subclasses

headius:
There were several issues with the old implementation:

* Iterating by creating a Set and an Iterator adds significant allocation overhead. Better to use Map#forEach.
* ConcurrentWeakHashMap is a very old implementation that has several inefficiencies as well as no custom implementation of Map#forEach (it falls back on the default impl which creates a Set and Iterator). Instead we use a synchronized WeakHashMap.
* The initial array was created with a default size of 4 elements. For many classes this will be insufficient, leading to excessive array allocation and copying as it grows. Instead we use the current class's "subclasses" size as an initial estimate.
* When no subclasses appear to be present, bail out early with an empty array that does minimal allocation.

Performance is significantly better for zero, small, and large subclass lists.

1M * Object.subclasses, with 83 elements:

Before:

  2.020000   0.040000   2.060000 (  1.823539)
  1.980000   0.030000   2.010000 (  1.737808)
  1.830000   0.010000   1.840000 (  1.738772)
  1.730000   0.020000   1.750000 (  1.740675)
  1.720000   0.000000   1.720000 (  1.707344)

After:

  1.340000   0.020000   1.360000 (  1.034745)
  0.930000   0.000000   0.930000 (  0.918107)
  0.920000   0.010000   0.930000 (  0.914828)
  0.920000   0.000000   0.920000 (  0.922137)
  0.920000   0.000000   0.920000 (  0.915440)

10M * Numeric.subclasses, with 4 elements:

Before:

  0.930000   0.030000   0.960000 (  0.789997)
  0.640000   0.010000   0.650000 (  0.621404)
  0.620000   0.010000   0.630000 (  0.614404)
  0.630000   0.000000   0.630000 (  0.629492)
  0.630000   0.010000   0.640000 (  0.608538)

After:

  0.720000   0.010000   0.730000 (  0.559470)
  0.460000   0.000000   0.460000 (  0.454176)
  0.510000   0.000000   0.510000 (  0.429875)
  0.430000   0.010000   0.440000 (  0.434487)
  0.430000   0.000000   0.430000 (  0.427971)

Fixes https://github.com/jruby/jruby/issues/8457

### DIFF
--- a/core/src/main/java/org/jruby/RubyArray.java
+++ b/core/src/main/java/org/jruby/RubyArray.java
@@ -178,6 +178,10 @@ public class RubyArray<T extends IRubyObject> extends RubyObject implements List
     }
 
     public static final RubyArray newArray(final Ruby runtime, final int len) {
+        if (len == 0) {
+            return newEmptyArray(runtime);
+        }
+
         IRubyObject[] values = IRubyObject.array(validateBufferLength(runtime, len));
         Helpers.fillNil(values, 0, len, runtime);
         return new RubyArray(runtime, values, 0, 0);
--- a/core/src/main/java/org/jruby/RubyClass.java
+++ b/core/src/main/java/org/jruby/RubyClass.java
@@ -111,6 +111,7 @@ import org.objectweb.asm.commons.GeneratorAdapter;
 public class RubyClass extends RubyModule {
 
     private static final Logger LOG = LoggerFactory.getLogger(RubyClass.class);
+    private static final double SUBCLASSES_CLEAN_FACTOR = 0.25;
 
     public static void createClassClass(Ruby runtime, RubyClass classClass) {
         classClass.setClassIndex(ClassIndex.CLASS);
@@ -1011,11 +1012,9 @@ public class RubyClass extends RubyModule {
 
     @JRubyMethod
     public IRubyObject subclasses(ThreadContext context) {
-        int subclassEstimate = this.subclassEstimate;
-
-        RubyArray<RubyClass> subs = RubyArray.newArray(context.runtime, subclassEstimate == -1 ? 4 : subclassEstimate);
-
-        concreteSubclasses(subs);
+        RubyArray<RubyClass> subs = newConcreteSubclassesArray(context);
+        int clearedCount = concreteSubclasses(subs);
+        finishConcreteSubclasses(subs, clearedCount);
 
         return subs;
     }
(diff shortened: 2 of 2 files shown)


headius:
@mohamedhafez Could you give this branch a try and see how the performance looks for you?

mohamedhafez:
Sure, can I just cherry-pick this commit onto a branch based on 9.4.9.0 so I'm sure everything else is compatible with my app?

headius:
@mohamedhafez Yes, should apply just fine to 9.4.9.0.

headius:
Reporting back from @mohamedhafez showed that the hard locking here is a big bottleneck for a highly-concurrent app, so I'm looking at lighter-weight locking options.

This patch was an attempt to use a ReentrantReadWriteLock, but it has overhead of its own and did not help much: https://gist.github.com/headius/30578ea0e26c9566462e70d963b2cd29

I'll attempt a lock-free version.

headius:
A second attempt with a volatile field and atomic updates improves over the synchronized version but uses a lot of background CPU spinning (a better impl would park the thread but then we have to track waiters, etc).

In the same gist as above: https://gist.github.com/headius/30578ea0e26c9566462e70d963b2cd29

headius:
A third attempt makes the `subclasses` field a read-only collection, and copy-on-write whenever it needs to be updated. This has very fast performance for `Class#subclasses` but very high cost for modifying the subclass list. It also is potentially not *quite* thread-safe, since WeakHashMap.get may mutate the list if there's a vacated reference.

headius:
I've pushed a new implementation of the subclasses collection that is just a linked list of weak references. The list is immutable so the only synchronization required is an atomic update of the head reference. Adding a class is O(1) and just attaches a new head. Removal, replacement, and traversal are O(n) linked list walks. Periodically (currently hardcoded to 25% evacuated references) the list gets rebuilt at the end of traversal, removing any dead links but preserving previous links with the same weak references; this allows the remove operation to simply `clear` weak references knowing that will be visible on any thread's copy of the list.

This is the fastest option so far for `Object.subclasses` but it is occasionally, strangely slower for the smaller `Numeric.subclasses` case. Benchmark results are below for 9.4.9.0, 9.4.10.0 (master), and CRuby 3.3 (benchmark will be committed soon):

9.4.9.0:
```
1 thread Numeric.subclasses
                          1.292k (±14.6%) i/s -      6.240k in   5.011785s
1 thread Object.subclasses
                         56.428  (± 3.5%) i/s -    285.000  in   5.056858s
5 thread Numeric.subclasses
                          1.906k (±18.1%) i/s -      9.265k in   5.000260s
5 thread Object.subclasses
                        157.154  (± 4.5%) i/s -    795.000  in   5.068326s
10 thread Numeric.subclasses
                          1.130k (± 6.6%) i/s -      5.700k in   5.064841s
10 thread Object.subclasses
                        170.303  (± 5.9%) i/s -    850.000  in   5.009226s
50 thread Numeric.subclasses
                        323.931  (± 4.3%) i/s -      1.632k in   5.047326s
50 thread Object.subclasses
                        139.149  (± 3.6%) i/s -    705.000  in   5.072644s
```

9.4.10.0
```
1 thread Numeric.subclasses
                          2.025k (±17.0%) i/s -      9.869k in   5.031959s
1 thread Object.subclasses
                        198.861 (±10.1%) i/s -    988.000 in   5.094656s
5 thread Numeric.subclasses
                          1.744k (±27.6%) i/s -      8.190k in   5.021846s
5 thread Object.subclasses
                        367.245 (±10.1%) i/s -      1.820k in   5.015953s
10 thread Numeric.subclasses
                        990.057 (±11.4%) i/s -      4.898k in   5.024722s
10 thread Object.subclasses
                        372.548 (±11.0%) i/s -      1.855k in   5.040239s
50 thread Numeric.subclasses
                        323.640 (± 5.3%) i/s -      1.632k in   5.059211s
50 thread Object.subclasses
                        216.725 (± 5.1%) i/s -      1.092k in   5.051484s
```

CRuby 3.3
```
1 thread Numeric.subclasses
                        999.470 (± 2.4%) i/s    (1.00 ms/i) -      5.047k in   5.052801s
1 thread Object.subclasses
                         86.706 (± 1.2%) i/s   (11.53 ms/i) -    440.000 in   5.076068s
5 thread Numeric.subclasses
                        796.605 (± 7.4%) i/s    (1.26 ms/i) -      3.956k in   5.011233s
5 thread Object.subclasses
                         84.310 (± 3.6%) i/s   (11.86 ms/i) -    424.000 in   5.035321s
10 thread Numeric.subclasses
                        656.311 (± 9.8%) i/s    (1.52 ms/i) -      3.237k in   5.012322s
10 thread Object.subclasses
                         83.197 (± 2.4%) i/s   (12.02 ms/i) -    416.000 in   5.002811s
50 thread Numeric.subclasses
                        363.659 (± 6.3%) i/s    (2.75 ms/i) -      1.833k in   5.067702s
50 thread Object.subclasses
                         72.960 (± 2.7%) i/s   (13.71 ms/i) -    371.000 in   5.089632s
```

headius:
I've pushed a refactoring that gets all benchmarks faster than 9.4.9.0, although sometimes just barely. The larger Object.subclasses cases are even faster:

9.4.9.0:
```
1 thread Numeric.subclasses
                          1.281k (±10.4%) i/s -      6.348k in   5.015677s
1 thread Object.subclasses
                         57.555  (± 1.7%) i/s -    290.000  in   5.042078s
5 thread Numeric.subclasses
                          1.866k (±12.8%) i/s -      9.243k in   5.026434s
5 thread Object.subclasses
                        156.215  (± 3.8%) i/s -    780.000  in   5.000381s
10 thread Numeric.subclasses
                          1.151k (±11.8%) i/s -      5.700k in   5.027341s
10 thread Object.subclasses
                        182.022  (±15.9%) i/s -    864.000  in   5.076040s
50 thread Numeric.subclasses
                        276.258  (±33.3%) i/s -      1.122k in   5.030777s
50 thread Object.subclasses
                        149.436  (±16.1%) i/s -    705.000  in   5.005946s
```

New impl:
```
1 thread Numeric.subclasses
                          1.871k (± 9.9%) i/s -      9.313k in   5.028326s
1 thread Object.subclasses
                        194.344 (± 2.6%) i/s -    988.000 in   5.087741s
5 thread Numeric.subclasses
                          1.966k (±12.9%) i/s -      9.750k in   5.044797s
5 thread Object.subclasses
                        475.950 (± 8.4%) i/s -      2.392k in   5.065777s
10 thread Numeric.subclasses
                          1.170k (±16.9%) i/s -      5.720k in   5.238812s
10 thread Object.subclasses
                        356.624 (±41.5%) i/s -      1.470k in   5.065979s
50 thread Numeric.subclasses
                        338.594 (±13.9%) i/s -      1.665k in   5.075536s
50 thread Object.subclasses
                        251.153 (±10.4%) i/s -      1.250k in   5.025052s
```

headius:
This is ready to merge any time. This is about the best I think we can do right now, all things considered.

We discussed on Matrix a bit about how Rails really should not be generating these lists over and over again, but that is a larger optimization to explore and implement. If they use this same logic on CRuby and JRuby, we should at least not be slower at it.

PR #8478, opened 2024-12-03
Avoid re-polling while reporting a Thread#raise

### LINKED ISSUE #8479: Thread interrupt requests can overwrite each other
For the spec:

> Thread#report_on_exception= when set to true prints the backtrace even if the thread was killed just after Thread#raise

it is expected that the exception asynchronously raised on the thread with `Thread#raise` will continue to propagate even if that thread is killed before the exception bubbles out. As the exception propagates, however, there will be additional event polls that my detect and propagate the new kill request. This leads to sporadic failures of the spec above, and does not reflect the behavior of CRuby.

An attempt to fix this by clearing the queue is in #8478 but it fails many test suites.

I am tagging the spec for now but we'll want to figure out how CRuby avoids interrupting the propagation of the `Thread#raise` exception.

headius:
Avoid polling during propagation and report of a `Thread#raise` exception. If we dispatch dynamically to the configured stderr, we may hit additional polls that trigger new interrupts, wiping out the raise interrupt currently being handled.

Fixes #8479

### DIFF
--- a/core/src/main/java/org/jruby/Ruby.java
+++ b/core/src/main/java/org/jruby/Ruby.java
@@ -172,10 +172,15 @@ import java.io.IOException;
 import java.io.InputStream;
 import java.io.PrintStream;
 import java.io.PrintWriter;
+import java.io.Writer;
 import java.lang.invoke.MethodHandle;
 import java.lang.ref.WeakReference;
 import java.net.BindException;
+import java.nio.ByteBuffer;
+import java.nio.channels.Channels;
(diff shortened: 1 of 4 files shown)


headius:
Initial naive attempt clearly doesn't pass.

headius:
Replaced original attempt (clear interrupt queue when raising) with a new version that simply avoids polling during IO for the raise report. This should fix the original issue without substantially changing the way interrupts are handled.

PR #8482, opened 2024-12-04
specify bin/jruby in BUILDING doc

mohamedhafez:
With rbenv installed on my system, the commands listed only work if I specify bin/jruby in all the commands given in BUILDING.md

### DIFF
--- a/BUILDING.md
+++ b/BUILDING.md
@@ -117,14 +117,14 @@ For normal day-to-day testing, we recommend running the Ruby specs. We have set 
 "fast" grouping that takes only a couple minutes to run:
 
 ```
-jruby -S rake spec:ruby:fast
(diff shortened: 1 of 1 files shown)


enebo (changes requested):
@mohamedhafez Sorry I missed this.  Could you just change `bin/jruby` to `./bin/jruby`?  Your suggestion requires either you are in project home AND you have '.' in your path (or project root).  The extra dot removes having PATH set a particular way.

I personally use ./bin/jruby all the time for JRuby 10 work and it is a force of habit to not depend on PATH.

mohamedhafez:
done!

enebo:
@mohamedhafez thanks for the doc update!

Issue #8422, opened 2024-11-12
ConcurrentLocalContextProvider leaks memory per thread


matthias-fratz-bsz:
**Environment Information**
  
Provide at least:
* JRuby version: commit 6f9df83e (pom.xml says 9.4.10.0-SNAPSHOT)
* Operating system: Linux 6.1.0-26-amd64 #1 SMP PREEMPT_DYNAMIC Debian 6.1.112-1 (2024-09-30) x86_64 GNU/Linux

**Test Case**

[ThreadMemoryLeak.java](https://github.com/user-attachments/files/17716880/ThreadMemoryLeak.java.txt)

This code creates 10k threads, each of them executing a trivial amount of JRuby code, to simulate a heavily multi-threaded server application churning through threads. It then drops a heap dump for analysis.

Note: This simulates a use case seen in a production Cantaloupe server, which for some reason seems to create and terminate ~4 threads per minute in our case (I don#t understand Jetty's automatic thread pool management there). Because it keeps running for weeks at a time, it churns through tens of thousands of threads, accumulating tens of thousands of ~100kB `LocalContext` objects waiting to be cleaned up in `terminate()`. It eventually slows to a crawl due to memory starvation, or runs out of memory entirely and crashes.

**Expected Behavior**

JRuby creates a `LocalContext` for each thread, holds it in a `ThreadLocal`, and cleans it up when that particular thread has terminated.
The heap dump is a few MB and its size does not scale with the number of terminated threads.

The synthetic example doesn't actually have any concurrent threads; they all are more or less sequential. Thus it should be capable of running with a very small heap, no matter how high you crank the total number of threads that will be created and destroyed (line 36).

**Actual Behavior**

JRuby doesn't remove the `LocalContext` for each thread until the ScriptingContainer itself is disposed. In a long-running server application, that would be "never". In the example, it doesn't happen before the heap dump is written, at which point the JVM is essentially terminating.

The heap dump is hundreds of MB and scales with the number of threads that have been terminated. It is completely dominated by thousands of stale local contexts held by the ScriptingContainer's `ConcurrentLocalContextProvider`.

![biggest-object](https://github.com/user-attachments/assets/ae34c7a7-55d1-4a1e-a46b-492bf213ead2)

![accumulated-objects](https://github.com/user-attachments/assets/ace8d9eb-1434-4a15-b52d-92fa7c094e71)

Setting the number of threads to a high value causes massive heap consumption, and even the synthetic example will eventually run out of memory.

[...]

PR #8483, opened 2024-12-04
Fix memory leak in Concurrent/ThreadSafeLocalContextProvider, using a Thread for JRuby 9

matthias-fratz-bsz:
Uses the Cleaner API to run LocalContext.remove() when a Thread has terminated, but also eagerly calls all LocalContext.remove() on terminate().

This gets rid of the memory leak demonstrated by the test case in #8422 .

### DIFF
--- a/core/src/main/java/org/jruby/embed/internal/ConcurrentLocalContextProvider.java
+++ b/core/src/main/java/org/jruby/embed/internal/ConcurrentLocalContextProvider.java
@@ -44,33 +44,7 @@ import org.jruby.embed.LocalVariableBehavior;
  * @author Yoko Harada &lt;<a href="mailto:yokolet@gmail.com">yokolet@gmail.com</a>&gt;
  */
 public class ConcurrentLocalContextProvider extends AbstractLocalContextProvider {
-    private volatile ConcurrentLinkedQueue<AtomicReference<LocalContext>> contextRefs =
-        new ConcurrentLinkedQueue<AtomicReference<LocalContext>>();
-
-    private final ThreadLocal<AtomicReference<LocalContext>> contextHolder =
-            new ThreadLocal<AtomicReference<LocalContext>>() {
-                @Override
-                public AtomicReference<LocalContext> initialValue() {
-                    AtomicReference<LocalContext> contextRef = null;
-
-                    try {
-                        contextRef = new AtomicReference<LocalContext>(getInstance());
-                        contextRefs.add(contextRef);
-                        return contextRef;
-                    } catch (NullPointerException npe) {
-                        if (contextRefs == null) {
-                            // contextRefs became null, we've been terminated
-                            if (contextRef != null) {
-                                contextRef.get().remove();
-                            }
-
-                            return null;
-                        } else {
-                            throw npe;
-                        }
-                    }
-                }
-            };
+    private final ThreadLocalContext contextHolder = new ThreadLocalContext(this::getInstance);
 
     public ConcurrentLocalContextProvider(LocalVariableBehavior behavior) {
         super( getGlobalRuntimeConfigOrNew(), behavior );
@@ -93,12 +67,12 @@ public class ConcurrentLocalContextProvider extends AbstractLocalContextProvider
 
     @Override
     public BiVariableMap getVarMap() {
-        return contextHolder.get().get().getVarMap(this);
+        return contextHolder.get().getVarMap(this);
     }
 
     @Override
     public Map getAttributeMap() {
-        return contextHolder.get().get().getAttributeMap();
+        return contextHolder.get().getAttributeMap();
     }
 
     @Override
@@ -108,20 +82,7 @@ public class ConcurrentLocalContextProvider extends AbstractLocalContextProvider
 
     @Override
     public void terminate() {
-        ConcurrentLinkedQueue<AtomicReference<LocalContext>> terminated = contextRefs;
-        contextRefs = null;
-
-        if (terminated != null) {
-            for (AtomicReference<LocalContext> contextRef : terminated) {
-                contextRef.get().remove();
-                contextRef.lazySet(null);
-            }
-
-            terminated.clear();
-        }
-
-        contextHolder.remove();
-        contextHolder.set(null);
+    	contextHolder.terminate();
     }
 
 }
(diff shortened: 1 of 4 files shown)


matthias-fratz-bsz:
So... oops, `Cleaner` is Java 9+. `BUILDING.md` says JDK 8 or up, so my PR should probably be targeting that.

I'm planning to reimplement it directly on top of the `Reference` API some time after the christmas break. It'll probably take a while because the `Cleaner` API makes the boundary cases much easier to handle... but obviously that is of no help where it isn't available.

headius:
@matthias-fratz-bsz Do the PR against the `10-dev` branch for now... that's our JRuby 10 branch and it will require Java 21 minimum.

We'll have to evaluate options for 9.4 since that must continue to support 8 for at least a year.

headius:
> Do the PR against the 10-dev branch for now.

... using the Cleaner API.

Bonus points if you want to do a separate PR for 9.4 that just emulates Cleaner with threads and references!

matthias-fratz-bsz:
@headius I ported the Cleaner API version to 10-dev as PR #8561 (if "git cherry-pick" can be considered porting), and added a simple implementation here that just uses a cleanup thread per `LocalContextProvider`.

For simplicity, that replacement code isn't as sophisticated as `Cleaner`: If you forget to call `terminate()`, the thread sticks around forever. `Cleaner` does some really neat trickery to terminate its thread once the `Cleaner` instance itself has been garbage collected... but I don't think this is necessary here. It's a daemon thread that doesn't keep the JVM from exiting, and `terminate()` does always seem to be called eventually.

headius:
Ok this is going to need more review and some discussion with @enebo @kares because it introduces another "cleaner" thread.

To clarify for anyone catching up... this PR is a 9.4 version of #8561, but with a small hand-made `Cleaner` since we can't use the Java 9+ version. This may be fine to merge, but introducing another thread demands some extra scrutiny.

headius:
@matthias-fratz-bsz How critical is this fix to you? We are hoping to get another quick release out (9.4.11.0) with some critical fixes, but I still have some concerns about the lifecycle of this new cleaning thread. I'd like to postpone until 9.4.12.0.

I will look it over again today and see if I can assuage my concerns.

matthias-fratz-bsz:
@headius Go ahead; this is not a time-critical issue for us. The memory leak in Cantaloupe is slow enough that daily restarts, while ugly, are a serviceable workaround. Also, https://github.com/cantaloupe-project/cantaloupe/pull/715 (essentially: turning off Variable Sharing) should reduce the amount of memory leaked to a point where it is essentially irrelevant.

headius (approved):
After review I feel pretty good about this impl. I summarize it here for myself and others who might review.

* LocalContext are set up on a per-thread basis, mirroring ThreadContext in the deeper JRuby runtime.
* The atomic reference holding the LocalContext is held in a ThreadLocal as before, but also in a PhantomReference.
* When the LocalContextProvider (Concurrent or ThreadSafe) is terminated, the cleaner cleans out all LocalContext references. This would be the normal situation.
* If terminate is not called, but the thread terminates and is successfully GCed, then its internal ThreadLocal table should go away, dereferencing the AtomicReference. Once GC clears that, the PhantomReference will be enqueued and a "Cleaner" daemon thread will eventually clear out remaining references to the LocalContext.

As for the extra thread... we do still spin up threads (in executors) for JIT, asynchronous IO select operations, and fibers. These executors get terminated when the JRuby runtime is terminated, which happens along the way when a ScriptingContainer is terminated. ScriptingContainer termination also calls terminate on the LocalContextProvider.

Possible concerns:

* ScriptingContainers that do not get terminated will not clean up the cleaner thread. Of course they will also not clean up the runtime, which will not clean up the executors, leaving potentially dozens of orphaned threads.
* There are few tests for this behavior to confirm that things are being cleaned up.

@matthias-fratz-bsz Perhaps you could write up a quick test of this embedding that, for example, spins up and terminates many threads and verifies they get cleaned up?

I would be ok merging this, since it's "just another thread" alongside the executors, it's specific to the embedding API, and users of that API must already be calling `terminate` if they expect resources to get cleaned up.

matthias-fratz-bsz:
I also analyzed the thread's lifecycle a bit more closely:

- It gets started essentially from the constructor. This isn't exactly best practice. Maybe it would be better to delay its start to, say, the first time a `LocalContext` is actually requested? Might be worth the slightly more complicated code.
- It is only terminated from `LocalContextProvider.termiante()`. So if the `LocalContextProvider` were to be garbage-collected without calling `termiante()`, the thread would absolutely stick around forever (well, until JVM termination). However, `LocalContextProvider.terminate()` is called from `ScriptingContainer.finalize()` (via `ScriptingContainer.terminate()`), so there should be no situations where `LocalContextProvider` gets garbage-collected without having `terminate()` called first.

matthias-fratz-bsz:
@headius I'll try to make such a test case and add it here. Current idea is to hold a `PhantomReference` to the thread's LocalContext we want to monitor, then keep poking the GC until the reference gets enqueued proving that the `LocalContext` was cleaned. I haven't tried it but I think it'll work.

headius:
@matthias-fratz-bsz that sounds like a good plan.

matthias-fratz-bsz:
@headius So... third time's the charm, I guess. This testcase now checks that:

- When the test thread quits, `LocalContext.remove()` is called and the `LocalContext` is garbage-collected.
- When `LocalContextProvider.terminate()` is called, so is `LocalContext.remove()`. Once the test thread then terminates, its `LocalContext` is also garbage-collected.

The test for `LocalContext` being garbage-collected uses a `PhantomReference` as discussed. For checking that `remove()` was called, I found it easier to just override the method in a subclass. I'm checking both `LocalContext.remove()` and garbage-collection because `LocalContext.remove()` is presumably important to clean up resources, and garbage-collection avoids the memory leaks that got me here.

headius:
@matthias-fratz-bsz Thank you for the test! Looks like about as good as we can get it considering we're trying to test GC effects.

We may need to do another "quick fix" 9.2.12.0 release that doesn't include this (trying to minimize changes again), but I think it's ready to merge into master after that.

@enebo This could use some review but I believe it is ready.

headius:
I've re-pushed this as https://github.com/jruby/jruby/pull/8969 to see if the issues clear up.

headius:
The re-push in #8969 looks good so we'll merge that one for 9.4.14.0.

PR #8497, opened 2024-12-08
Nil can get NilClass withoutretrieving it through runtime. False and …

enebo:
…True start frozen already


enebo:
I already did this a second time in a second PR :)  The approach ended up similiarly.

PR #8503, opened 2024-12-10
[ji] do not expose InterfaceImpl classes in Ruby land

kares:
Yet another attempt at resolving https://github.com/jruby/jruby/issues/8349, for good.

Since the reverts done in JRuby 9.4.9.0 still generate the warnings.

Seems like there's more (related) stuff that changed (since 9.4.6.0) given the checks, changed in this PR, [were there](https://github.com/jruby/jruby/blob/9.4.6.0/core/src/main/java/org/jruby/javasupport/Java.java#L507-L509) but did not cause warnings :shrug:. 

Instead of chasing around I took an approach of marking `InterfaceImpl` generated classes synthetic and excluded those from being set in Ruby land...

### DIFF
--- a/core/src/main/java/org/jruby/javasupport/Java.java
+++ b/core/src/main/java/org/jruby/javasupport/Java.java
@@ -521,7 +521,9 @@ public class Java implements Library {
             for ( int i = interfaces.length; --i >= 0; ) {
                 proxy.includeModule(getInterfaceModule(runtime, interfaces[i]));
             }
-            if ( Modifier.isPublic(clazz.getModifiers()) ) {
+            // we don't want to expose any synthetic classes into Ruby constants,
+            // JRuby generated classes such as interface impls (org.jruby.gen.InterfaceImpl) are marked synthetic
+            if (Modifier.isPublic(clazz.getModifiers()) && !clazz.isSynthetic()) {
                 addToJavaPackageModule(proxy);
             }
         }
--- a/test/jruby/test_higher_javasupport.rb
+++ b/test/jruby/test_higher_javasupport.rb
@@ -1737,6 +1737,29 @@ CLASSDEF
     end
   end
 
+  def test_no_warnings_on_interface_impls_being_set_as_constants
+    runner_base = Class.new do
+      class << self
+        def run; self != self end
+      end
+    end
+
+    runner_impl1 = Class.new(runner_base)
+    runner_impl2 = Class.new(runner_base)
+    runner_impl3 = Class.new(runner_base)
+
+    output = with_stderr_captured do
+      threads = []
+      threads << java.lang.Thread.new(runner_impl1)
+      threads << java.lang.Thread.new(runner_impl2)
+      threads << java.lang.Thread.new(runner_impl3)
+      threads.each(&:start)
+      threads.each(&:join)
+    end
+    # expect no warning: already initialized constant org.jruby.gen::InterfaceImpl1353309827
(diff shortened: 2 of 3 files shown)


core/src/main/java/org/jruby/javasupport/Java.java
```diff
@@ -521,7 +521,8 @@ else if ( clazz == Object.class ) {
             for ( int i = interfaces.length; --i >= 0; ) {
                 proxy.includeModule(getInterfaceModule(runtime, interfaces[i]));
             }
-            if ( Modifier.isPublic(clazz.getModifiers()) ) {
+            // we don't want to expose JRuby internal generated classes such as org.jruby.gen.InterfaceImpl
+            if (Modifier.isPublic(clazz.getModifiers()) && !clazz.isSynthetic() && !Proxy.isProxyClass(clazz)) {
```
headius, line 525: I think the comment should be more general... we don't want to expose *any* synthetic classes.

test/jruby/test_higher_javasupport.rb
```diff
@@ -1737,6 +1737,29 @@ def test_indexed_bean_style_accessors_are_not_aliased
     end
   end
 
+  def test_no_warnings_on_interface_impls_being_set_as_constants
```
headius, line 1740: Just to confirm, you were able to make this fail? I would have thought a more aggressive looping test would be necessary.

headius (approved):
This is genius in its simplicity... *of course* we wouldn't want to bind synthetic classes, and hidden classes we generate should *obviously* be marked as synthetic.

test/jruby/test_higher_javasupport.rb
kares: yes, it does reproduce fairly reliably, at least for me locally:
```
Failure: test_no_warnings_on_interface_impls_being_set_as_constants(TestHigherJavasupport):
  /home/kares/workspace/oss/jruby/test/jruby/test_higher_javasupport.rb:1754: warning: already initialized constant org.jruby.gen::InterfaceImpl776208565
  /home/kares/workspace/oss/jruby/test/jruby/test_higher_javasupport.rb:1755: warning: already initialized constant org.jruby.gen::InterfaceImpl776208565
  .
  <85> is neither nil or false.
/home/kares/workspace/oss/jruby/test/jruby/test_higher_javasupport.rb:1760:in `test_no_warnings_on_interface_impls_being_set_as_constants'
     1757:       threads.each(&:join)
     1758:     end
     1759:     # expect no warning: already initialized constant org.jruby.gen::InterfaceImpl1353309827
  => 1760:     refute output.index('already initialized constant'), output

```

core/src/main/java/org/jruby/javasupport/Java.java
kares: changed it to: https://github.com/jruby/jruby/pull/8503/files#diff-41d5e3a18474c6013321ccd78b4084effc7f6504032f053413df31c227f76720R524-R525

also removed the `isProxyClass` check, that felt risky in terms of causing regressions. 
was originally motivated by the fact that there's `-Xinterfaces.useProxy` to switch to using `java.lang.reflect.Proxy` for interface impls but I do not believe it's used much (actually seems broken and should maybe be removed in 10).

headius: Ok sounds good to me.

headius:
@kares All good on my end, merge whenever you are ready!

PR #8506, opened 2024-12-10
Test new version of resolv

headius:
This is a version of ruby/resolv lightly modified to fix the Windows platform check. We can merge this or wait for a release, but this will help confirm that the new version runs ok.


headius:
See https://github.com/ruby/resolv/pull/75.

Issue #8507, opened 2024-12-10
Eliminate use of unsafe in StringSupport


headius:
The following error shows up now when running with JDK 25 builds:

```
WARNING: A terminally deprecated method in sun.misc.Unsafe has been called
WARNING: sun.misc.Unsafe::arrayBaseOffset has been called by org.jruby.util.StringSupport (file:/home/headius/work/jruby/lib/jruby.jar)
WARNING: Please consider reporting this to the maintainers of class org.jruby.util.StringSupport
WARNING: sun.misc.Unsafe::arrayBaseOffset will be removed in a future releas
```

We should replace this usage with something like VarHandle in JRuby 10, and consider options for avoiding this warning on 9.4 maintenance releases.

PR #8508, opened 2024-12-11
Replace Unsafe byte[] long-walking with VarHandle

headius:
Fixes jruby/jruby#8507


headius:
Also using this PR to remove other uses of Unsafe. The varTable management was simple enough. Two uses remain that require a bit more work:

* NonBlockingHashMapLong uses Unsafe to directly access arrays and fields.
* SipHashInline uses Unsafe for wide-stride byte[] walking.

headius:
SipHashInline is done.

PR #8519, opened 2024-12-12
inspect() -> inspect(ThreadContext)

enebo:
This started as a PR to fix all TC-less methods in RubyIO and RubyFile but the main problem was inspect().  This lead to bubbling through all the inspect definers and consumers.  The reason for review is:

  1. new default impl on IRubyObject.  This should never get hit normally but it exists for implements IRubyObject types.  Not sure this will ever happen but it is weird enough to need more eyes.
  2. RubyString RubyString#inspect() caused problems from str_scan extension and perhaps others.  It is sticking around with deprecation but this made me wonder if perhaps there are other fallout from extensions assuming something from inspect() I did not anticipate.


headius (approved):
PR looks about like I'd expect. Lots of new propagation of ThreadContext, and a few new getCurrentContext that I assume are just pushing the line further up the stack (to be replaced with a passed-in argument eventually).

enebo:
> PR looks about like I'd expect. Lots of new propagation of ThreadContext, and a few new getCurrentContext that I assume are just pushing the line further up the stack (to be replaced with a passed-in argument eventually).

Yep.  Slowly pushing context down further and further but in order to do it all callers of same lower-level method needs to have a context at hand.

Issue #8516, opened 2024-12-11
Strange performance difference?


jzakiya:
In testing this code against 4 Ruby VMs there is a significant performance anomaly in JRuby (via `rvm`).

```
jruby 9.4.9.0 (3.1.4) 2024-11-04 547c6b150e Java HotSpot(TM) 64-Bit Server VM 15.0.2+7-27 on 15.0.2+7-27 +jit [x86_64-linux]
```

Test Code
```
module Primes
  module Utils

    # Miller-Rabin prime test in Ruby
    # From: http://en.wikipedia.org/wiki/Miller-Rabin_primality_test
    # Ruby Rosetta Code: http://rosettacode.org/wiki/Miller-Rabin_primality_test
    # I modified the Rosetta Code, as shown below

    def primemr2?(k=5)  # increase k for more reliability
      return self >= 2 if self <= 3 
      return false unless 6.gcd(self) == 1
      neg_one_mod = n = d = self - 1      # these are even as self is always odd
      s = 0
      (d >>= 1; s += 1) while d.even?
      k.times do
        a = 2 + rand(self-4)
        y = a.pow(d,self)                 # y = (a**d) mod self
        next if y == 1 or y == neg_one_mod
        (s-1).times do
          y = y.pow(2, self)              # y = (y**2) mod self
          return false if y == 1
          break if y == neg_one_mod
        end
        return false unless y == neg_one_mod
      end
      true                                # n is prime (with high probability)
    end
    
    def primemr3?(k=5)  # increase k for more reliability
      return self >= 2 if self <= 3
      return false unless 6.gcd(self) == 1
      neg_one_mod = n = d = self - 1      # these are even as self is always odd
      d >>= 2 while (d & 0b11) == 0; d >>= (d & 1)^1  # make d odd number
      k.times do
        a = 2 + rand(self-4)
        y = a.pow(d, self)                # y = (a**d) mod self
        next if y == 1 or y == self-1
        until y == 1 || y == neg_one_mod || d == n
	      y = y.pow(2, self)              # y = (y**2) mod self
          d <<= 1
        end
        return false unless y == neg_one_mod
      end
      true                                # n is prime (with high probability)
    end

    # Returns true if +self+ is a prime number, else returns false.
    def primemr?(k = 5)
      wits = WITNESS_RANGES.find { |range, wits| range > self }
      witnesses = wits ? wits[1] : k.times.map{ rand(self - 4) + 2 }
[...]

PR #8523, opened 2024-12-13
Revert multiply to use non-intrinsic exactness checks

headius:
The call to Math.multiplyExact works very well when the result does not overflow, skipping range checks and using the host CPU's support for overflow checks. However when we are dancing on the range boundary and multiplyExact fails repeatedly, the cost of raising all those exceptions massively outweighs the gains.

This is the primary cause of performance issues reported in jruby/jruby#8516.

This patch reverts to a version of the old range-checking logic. Performance on the bug report's benchmark is an order of magnitude faster for the two cases that seemed to overflow a lot, and the other cases do not seem to degrade.

Fixes jruby/jruby#8516

### DIFF
--- a/core/src/main/java/org/jruby/RubyFixnum.java
+++ b/core/src/main/java/org/jruby/RubyFixnum.java
@@ -40,7 +40,6 @@ package org.jruby;
 import java.math.BigInteger;
 
 import org.jcodings.specific.USASCIIEncoding;
-import org.jruby.RubyEnumerator.SizeFn;
 import org.jruby.compiler.Constantizable;
 import org.jruby.runtime.Block;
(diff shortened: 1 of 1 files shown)


headius:
Note that JRuby 10 could use a different patch proposed in #8516 that uses `Math.multiplyHigh` as an overflow check before doing the exact operation. That may be no faster than this, but it has the potential to reduce to two intrinsic Math calls rather than always using our hand-written logic.

headius:
Ok clearly I botched this. Will circle back.

headius:
New patch looks clean enough. I'm only doing this one targeted fix since we have alternatives in JRuby 10 using Java 21.

PR #8533, opened 2024-12-17
Fixes and impls for Ruby 3.4 specs

headius:
Trying to whittle down the number of failures in 3.4 specs on JRuby 10.


headius:
Merged this after knocking down about 100 failures and getting us well below 100 remaining. The rest will be trickier and probably should be done a PR at a time for each class or method.

Issue #8384, opened 2024-10-24
OutOfMemoryError while constructing a NameError in the context of a large object


mrckzgl:
Causing a NameError within the scope of a large Object results in an OutOfMemoryError instead of the NameError, hiding the actual user space code location of the NameError and potentially crashing the process.

**Environment Information**
  

* jruby 9.4.8.0 (3.1.4) 2024-07-02 4d41e55a67 OpenJDK 64-Bit Server VM 17.0.12+7-Ubuntu-1ubuntu222.04 on 17.0.12+7-Ubuntu-1ubuntu222.04 +jit [x86_64-linux]
* `Linux 6.8.0-47-generic #47~22.04.1-Ubuntu SMP PREEMPT_DYNAMIC Wed Oct  2 16:16:55 UTC 2 x86_64 x86_64 x86_64 GNU/Linux`

**Expected Behavior**

The NameError should be shown.

**Actual Behavior**

All the memory of the process gets filled while collecting an inspect string of the large object. If not enough memory is available, an OutOfMemoryError is shown, hiding the actual NameError location.

**Minimal Example**

```
class Test
  def initialize()
    @obj = {}
    for t in 1..800000 do
      @obj[t.to_s] = Random.rand(10000000).to_s
    end
  end
  def error
    nameerror
  end
end

puts("generating large obj")
test = Test.new
puts("done")
test.error
```
Running with `jruby -w -J-Xmx250M test2.rb` results in:
```
generating large obj
done
Error: Your application used more memory than the safety cap of 250M.
Specify -J-Xmx####M to increase it (#### = cap size in MB).
java.lang.OutOfMemoryError: Java heap space
	at org.jruby.dist/org.jruby.util.ByteList.ensure(ByteList.java:362)
	at org.jruby.dist/org.jruby.util.io.EncodingUtils.strBufCat(EncodingUtils.java:1883)
	at org.jruby.dist/org.jruby.util.io.EncodingUtils.strBufCat(EncodingUtils.java:1868)
	at org.jruby.dist/org.jruby.util.io.EncodingUtils.encCrStrBufCat(EncodingUtils.java:1991)
	at org.jruby.dist/org.jruby.util.io.EncodingUtils.encStrBufCat(EncodingUtils.java:1894)
	at org.jruby.dist/org.jruby.RubyBasicObject.inspectObj(RubyBasicObject.java:1145)
[...]

PR #8538, opened 2024-12-20
Backport no-inspect NameError logic

headius:
This is a backport of the no-inspect NameError logic added to JRuby 10 in #8533, guarded by an option for users on JRuby 9.4.

On 9.4, the inspect logic is enabled by default, but can be disabled by with the option `-XnameError.inspect.object=false`. Because the object inspect is a specified behavior of Ruby 3.1, we can't disable it by default, but this allows users to opt out of the behavior.

JRuby 10 will ignore this logic and avoid inspecting the object attached to the NameError. The option will have no effect.

Using the example from #8384 with this new option:

```
$ jruby -w -J-Xmx250M -XnameError.inspect.object=false blah.rb
generating large obj
done
NameError: undefined local variable or method `nameerror' for an instance of Test
   error at blah.rb:9
  <main> at blah.rb:16
```

Interestingly, I discovered that the original code does not OOM on Java 21, because it never seems to call the expected inspect method. I will open an issue to investigate that.

### DIFF
--- a/core/src/main/java/org/jruby/Ruby.java
+++ b/core/src/main/java/org/jruby/Ruby.java
@@ -389,7 +389,9 @@ public final class Ruby implements Constantizable {
         // Initialize all the core classes
         comparableModule = RubyComparable.createComparable(this);
         enumerableModule = RubyEnumerable.createEnumerableModule(this);
+
         stringClass = RubyString.createStringClass(this);
+        emptyFrozenString = freezeAndDedupString(RubyString.newEmptyString(this));
 
         falseString = newString(FALSE_BYTES);
         falseString.setFrozen(true);
@@ -3529,6 +3531,10 @@ public final class Ruby implements Constantizable {
         return emptyFrozenArray;
     }
 
+    public RubyString getEmptyFrozenString() {
+        return emptyFrozenString;
+    }
+
     public RubyBoolean newBoolean(boolean value) {
         return value ? trueObject : falseObject;
     }
@@ -5880,6 +5886,7 @@ public final class Ruby implements Constantizable {
     }
 
     private final RubyArray emptyFrozenArray;
+    private final RubyString emptyFrozenString;
 
     /**
      * A map from Ruby string data to a pre-frozen global version of that string.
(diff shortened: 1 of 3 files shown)


enebo:
> This is a backport of the no-inspect NameError logic added to JRuby 10 in #8533, guarded by an option for users on JRuby 9.4.
> 
> On 9.4, the inspect logic is enabled by default, but can be disabled by with the option `-XnameError.inspect.object=false`. Because the object inspect is a specified behavior of Ruby 3.1, we can't disable it by default, but this allows users to opt out of the behavior.
> 
> JRuby 10 will ignore this logic and avoid inspecting the object attached to the NameError. The option will have no effect.
> 
> Using the example from #8384 with this new option:
> 
> ```
> $ jruby -w -J-Xmx250M -XnameError.inspect.object=false blah.rb
> generating large obj
> done
> NameError: undefined local variable or method `nameerror' for an instance of Test
>    error at blah.rb:9
>   <main> at blah.rb:16
> ```
> 
> Interestingly, I discovered that the original code does not OOM on Java 21, because it never seems to call the expected inspect method. I will open an issue to investigate that.

I realized that when I reproduced this it was Java 21 BUT also it was JRuby 10.  I also cannot reproduce with Java 21 on 9.4.  If I had to guess my TC refactoring for inspect() "fixed" it.

```
./bin/jruby -vw -J-Xmx250M ../snippets/name1.rb 
jruby 10.0.0.0-SNAPSHOT (3.4.0) 2024-12-19 7554ac66ab OpenJDK 64-Bit Server VM 21.0.5+11-LTS on 21.0.5+11-LTS +indy +jit [x86_64-linux]
uri:classloader:/jruby/kernel/data.rb:64: warning: ambiguity between regexp and two divisions: wrap regexp in parentheses or add a space after `/' operator
generating large obj
done
Error: Your application used more memory than the safety cap of 250M.
Specify -J-Xmx####M to increase it (#### = cap size in MB).
java.lang.OutOfMemoryError: Java heap space

```

headius:
Ok I think it's just that 21 is a bit more memory efficient, so it's able to handle this case in 250MB. I lowered it to 225MB and it fails like Java 8.

```
$ jruby -v -J-Xmx225M blah.rb
jruby 9.4.10.0-SNAPSHOT (3.1.4) 2025-01-15 2f915ca63d OpenJDK 64-Bit Server VM 21.0.5+11-LTS on 21.0.5+11-LTS +jit [arm64-darwin]
generating large obj
done
Error: Your application used more memory than the safety cap of 225M.
Specify -J-Xmx####M to increase it (#### = cap size in MB).
java.lang.OutOfMemoryError: Java heap space
```

Good to merge for 9.4.10.

PR #8542, opened 2024-12-21
Fix Dir.glob ../ and cases

### LINKED ISSUE #8426: Dir[] vs "../../"
**Environment Information**
  
* jruby 9.4.9.0 (3.1.4) 2024-11-04 547c6b150e OpenJDK 64-Bit Server VM 21.0.4+7-1 on 21.0.4+7-1 +jit [x86_64-OpenBSD]
* OpenBSD armata 7.6 GENERIC.MP#338 amd64

and

*  Windows Server 2019 10.0 amd64 JVM 11.0.23+9-LTS Azul Systems, Inc.

**Expected Behavior**

```ruby
pp Dir[File.join('../../t', '**', 'site.css')]

# ruby 3.3.5 (2024-09-03 revision ef084cc8f4) [x86_64-openbsd]
#
# [ "../../t/sg6/public/sheets/site.css",
#   "../../t/sg6/war_20240407_1734/sheets/site.css" ]
```

**Actual Behavior**

```ruby

pp Dir[File.join('../../t', '**', 'site.css')]

# jruby 9.4.9.0 (3.1.4) 2024-11-04 547c6b150e OpenJDK 64-Bit Server
#   VM 21.0.4+7-1 on 21.0.4+7-1 +jit [x86_64-OpenBSD]
#
# [ "./aux/t/sg6/public/sheets/site.css",
#   "./aux/t/sg6/war_20240407_1734/sheets/site.css"]
```

danini-the-panini:
Reverts #6681 to resolve #8426 

TODO: find new fix for #2150

### DIFF
--- a/core/src/main/java/org/jruby/util/Dir.java
+++ b/core/src/main/java/org/jruby/util/Dir.java
@@ -776,30 +776,7 @@ public class Dir {
         FileResource file = JRubyFile.createResource(runtime, cwd, fileName);
 
         if (file.exists()) {
-            byte[] newBytes;
-            
-            // get the real filename (case-sensitive)
-            if(file.isFile() && cwd != null) {
-                String path = file.isSymLink() ? file.absolutePath() : file.canonicalPath();
-
-                // compensate for missing slash
-                if (fileName.endsWith("/")) {
-                    path += "/";
-                }
-
-                if(fileName.contains("./")) {
-                    newBytes = ("./" + path.substring((path.length() - end + 2))).getBytes();
-                } else {
-                    int tempBegin = path.length() - fileName.length();
-                    newBytes = path.substring(tempBegin).getBytes();
-                }
-
-                end = newBytes.length;
-            } else {
-                newBytes = bytes;
-            }
-
-            return func.call(newBytes, begin, end - begin, enc, arg);
+            return func.call(bytes, begin, end - begin, enc, arg);
         }
 
         return 0;


headius:
I am reluctant to revert that fix without some input from @edipofederle on what exactly it fixed. Punting to 9.4.11.0 so we can discuss further.

danini-the-panini:
Looking at the logic they added, it just queried the filesystem for the actual path and replaced the glob result with that, taking into account paths starting with `./`. However, this does not take into account any `..` in the path, which causes massive problems when glob patterns include it. I'm very pro reverting it until a better fix comes around, since I think preserving case in a case-insensitive filesystem is less important than globbing actually working.

The idea wasn't to merely just revert the change, though. I'm also still trying to find a way to do it on the fly, however it's proving to be quite tricky since the glob logic is so convoluted. I may still just try rework the original fix to better take into account `..` and `.` in the glob pattern.

danini-the-panini:
@headius @enebo how are we feeling about this?

core/src/main/java/org/jruby/util/Dir.java
```diff
@@ -776,50 +810,39 @@ private static int addToResultIfExists(Ruby runtime, String cwd, byte[] bytes, i
...
+        final byte[] bytes = path.getUnsafeBytes();
+        final byte[] scheme = extractScheme(bytes, begin, end);
+        if (scheme != null) begin += scheme.length;
+        return glob_helper(runtime, cwd, scheme, bytes, begin, end, enc, sub, flags, func, arg);
+    }
+
+    private static int glob_helper(Ruby runtime, String cwd, byte[] scheme, ByteList path, int sub, int flags, GlobFunc<GlobArgs> func, GlobArgs arg) {
         final int begin = path.getBegin();
-        final int end = begin + path.getRealSize();
+        final int end = begin + path.length();
```
danini-the-panini, line 831: i'm not sure if this changes anything. looks like length just returns realSize :/

enebo: @danini-the-panini length() is there to satisfy Java CharSequence.  It is fine to leave it since it will always match byte length but it is a weird method since it implies it should return char length.

enebo (approved):
I think it is impossible to trace through this and know you are prepending scheme in all cases but that is the nature of this code.  It is very complicated.

I see no new tests but I do wonder if you noticed any cases which we are lacking coverage?

danini-the-panini:
@enebo there aren't really any new test cases, the case I sensitive one was enabled by the previous fix which I reverted. The only other thing to test would be the ../ case but it's quite niche.

core/src/main/java/org/jruby/util/Dir.java
```diff
@@ -776,50 +810,39 @@ private static int addToResultIfExists(Ruby runtime, String cwd, byte[] bytes, i
...
         byte[] path, int begin, int end, Encoding enc, int sub,
         int flags, GlobFunc<GlobArgs> func, GlobArgs arg) {
         int status = 0;
 
         int ptr = sub != -1 ? sub : begin;
 
-        if ( ! has_magic(path, ptr, end, flags) ) {
+        final GlobMagic nonMagic = CASEFOLD_FILESYSTEM ? GlobMagic.PLAIN : GlobMagic.ALPHA;
+
+        if ( has_magic(path, ptr, end, flags).compareTo(nonMagic) <= 0 ) {
```
headius, line 845: This change seems to mean we'll want to ensure we preserve enum order. Hopefully we're not likely to add any more? Probably should have some documentation to indicate that the order of that enum is important, or else add a method to the enum that more clearly spells out how the "priority" of the magic is determined.

headius (approved):
Complicated logic being patched here but I like the changes and the move to an enum for the three-mode magic glob behavior. Minor recommendations to address but otherwise this is good to go.

core/src/main/java/org/jruby/util/Dir.java
```diff
@@ -53,7 +53,7 @@
  */
 public class Dir {
     public final static boolean DOSISH = Platform.IS_WINDOWS;
-    public final static boolean CASEFOLD_FILESYSTEM = DOSISH;
+    public final static boolean CASEFOLD_FILESYSTEM = DOSISH || Platform.IS_MAC;
```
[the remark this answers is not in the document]
danini-the-panini: If we could somehow call (volumeSupportsCaseSensitiveNames)[https://developer.apple.com/documentation/foundation/urlresourcevalues/1780299-volumesupportscasesensitivenames]

core/src/main/java/org/jruby/util/Dir.java
danini-the-panini: MRI has more, but it handles braces and brackets differently

danini-the-panini: I've added a comment stating order is important

core/src/main/java/org/jruby/util/Dir.java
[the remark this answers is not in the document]
headius: Good find! I'll do some research today, but this is mostly a "wish" so it shouldn't prevent merging. The vast majority of Windows and MacOS systems will be running on case-preserving filesystems.

headius:
I approve of merging this for 9.4.11.0, or punting to 9.4.12.0.

PR #8562, opened 2025-01-08
Use same logic as IR for kwarg handling in IO#write

### LINKED ISSUE #8398: ruby2_keywords + forwarding to native does not properly check arity
When combining `ruby2_keywords` with `alias_method` or `alias`, arguments are not handled properly.  This appears to require a built in function, since I'm not able to reproduce this with a normal function which I alias and redefine.  The same issue occurs with both `alias` and `alias_method`.

Basic Repro:
```
class File
  class << self
    alias original_write write

    ruby2_keywords def write(*args, &block)
      puts "traced #{__method__}"
      original_write(*args, &block)
    end

  end
end


File.write('scratch', "stuff", 0, mode: 'a+')
```

On MRI:
```
$ /usr/bin/ruby --version
ruby 3.1.2p20 (2022-04-12 revision 4491bb740a) [x86_64-linux-gnu]
$ /usr/bin/ruby test.rb
traced write
$ 
```

On JRuby:
```
$ jruby --version
jruby 9.4.8.0 (3.1.4) 2024-07-02 4d41e55a67 OpenJDK 64-Bit Server VM 11.0.15+0-adhoc..source on 11.0.15+0-adhoc..source +jit [x86_64-linux]
$ jruby test.rb
traced write
ArgumentError: wrong number of arguments (given 4, expected 2..3)
   write at org/jruby/RubyIO.java:4290
   write at temp.rb:7
  <main> at temp.rb:13
$ 
```

headius:
When passing kwargs from a ruby2_keywords method to a core "real keywords" method, we lose the keywordiness of the incoming hash and raise an argument error for the arity mismatch. By using the same logic as IR here, we properly handle the incoming r2k hash.

This same patch could be applied generally to all core methods that accept keywords, but given the rarity of r2k to core "real keywords" methods we have chosen to only do targeted fixes. This reduces the potential impact (versus a new, more general solution) and the required work (9.4.10 is eagerly awaited). We will explore more reliable, general improvements to core method kwarg handling in JRuby 10.

Fixes #8398.

### DIFF
--- a/core/src/main/java/org/jruby/RubyFile.java
+++ b/core/src/main/java/org/jruby/RubyFile.java
@@ -45,6 +45,7 @@ import org.jcodings.specific.UTF8Encoding;
 import org.jruby.anno.JRubyClass;
 import org.jruby.anno.JRubyMethod;
 import org.jruby.exceptions.NotImplementedError;
+import org.jruby.ir.runtime.IRRuntimeHelpers;
 import org.jruby.runtime.*;
 import org.jruby.runtime.JavaSites.FileSites;
 import org.jruby.runtime.builtin.IRubyObject;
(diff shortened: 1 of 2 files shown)


headius:
@enebo I'm going to merge this. Let me know if you see any issues.

Issue #8406, opened 2024-11-04
Timeout::timeout leads to OOM with multiple threads


jsvd:
This is a follow up to https://github.com/jruby/jruby/issues/8403 where #8404 fixed the single threaded crash but we're observing an OOM when multiple threads are involved.
The OOM happened less frequently in our tests before #8404 since the TimeoutException "leak" would typically happen first.

**Environment Information**
  
jruby 9.4.9.0-SNAPSHOT (3.1.4) 2024-11-04 746b84aa02 OpenJDK 64-Bit Server VM 21.0.2+13-LTS on 21.0.2+13-LTS +jit [arm64-darwin]

**Expected Behavior**

1 thread seems to work fine and we can see the 10 million iterations finishing after 55 seconds:

```
❯ time JRUBY_OPTS="-J-Xmx128m -Xreify.classes=true" bin/jruby -rtimeout -e "timeout = 0.001; 1.times.map { Thread.new { 10_000_000.times { Timeout::timeout(timeout) { } rescue nil; } } }.each(&:join)"
JRUBY_OPTS="-J-Xmx128m -Xreify.classes=true" bin/jruby -rtimeout -e   67.62s user 11.53s system 143% cpu 55.135 total
```

It'd be expected that multiple threads would also complete in a similar amount of time.

**Actual Behavior**

With 2 threads the process OOMs eventually:

```
/tmp/jruby-9.4.9.0-SNAPSHOT
❯ time JRUBY_OPTS="-J-Xmx128m -Xreify.classes=true" bin/jruby -rtimeout -e "timeout = 0.001; 2.times.map { Thread.new { 10_000_000.times { Timeout::timeout(timeout) { } rescue nil; } } }.each(&:join)"

Exception: java.lang.OutOfMemoryError thrown from the UncaughtExceptionHandler in thread "Ruby-0-Thread-1: -e:1"
Exception: java.lang.OutOfMemoryError thrown from the UncaughtExceptionHandler in thread "Ruby-0-Thread-2: -e:1"
Exception: java.lang.OutOfMemoryError thrown from the UncaughtExceptionHandler in thread "main"

JRUBY_OPTS="-J-Xmx128m -Xreify.classes=true" bin/jruby -rtimeout -e   332.54s user 26.38s system 137% cpu 4:20.93 total
```

same but much faster with 10 threads:

```
/tmp/jruby-9.4.9.0-SNAPSHOT
[...]

PR #8576, opened 2025-01-17
Force a Thread's block to optimize eagerly

headius:
The block of code passed to Thread.new is frequently only executed once during a program's runtime, and in such cases there's often heavy lifting performed within the block such as an IO or Queue loop. Because we do not support any form of on-stack replacement, these blocks may never fully optimize and will remain in whatever execution mode JRuby starts them in (usually the slowest "startup" interpreter).

This patch adds logic from Thread.new into the block body to force completion of builds steps that would optimize the body. This is akin to the behavior of the command-line target script, which we also eagerly compile due to the great number of single-script utilities and benchmarks in the Ruby ecosystem.

This patch also plumbs that force-ability through the other "buildable" execution units: methods and block-based methods like define_method. This may become useful in the future to have more direct control over when a target block or method optimizes.

### DIFF
--- a/core/src/main/java/org/jruby/RubyThread.java
+++ b/core/src/main/java/org/jruby/RubyThread.java
@@ -67,6 +67,7 @@ import org.joni.Matcher;
 import org.jruby.anno.JRubyClass;
 import org.jruby.anno.JRubyMethod;
 import org.jruby.api.Create;
+import org.jruby.compiler.Compilable;
 import org.jruby.exceptions.MainExitException;
 import org.jruby.exceptions.RaiseException;
 import org.jruby.exceptions.ThreadKill;
@@ -658,6 +659,12 @@ public class RubyThread extends RubyObject implements ExecutionContext {
         if (threadImpl != ThreadLike.DUMMY) throw context.runtime.newThreadError("already initialized thread");
 
         BlockBody body = block.getBody();
+
+        // Force top-level body for thread to compile, since it may be executed only once and never get to JIT
+        if (body instanceof Compilable compilable) {
+            compilable.forceBuild(context);
+        }
+
         startThread(context, new RubyRunnable(this, context, args, block, callInfo), body.getFile(), body.getLine());
 
         return context.nil;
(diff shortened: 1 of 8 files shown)


headius:
Performance improvement based on the first attempt here:

COMMAND
```
$ jruby -Xjit.logging -e 'loop { th = eval("def x; Thread.new { sleep; i = 0; while i < 10_000_000; i+=1; end }; end; x"); sleep 1; t = Time.now; th.wakeup; th.join; puts Time.now - t }'
```

BEFORE (note no JIT output except the failure to compile the -e script)
```
[] jruby 
2025-01-17T13:43:42.664-06:00 [main] ERROR Ruby : failed to compile target script: -e - AOT not supported for scripts containing eval.
0.721603
0.665112
0.6637540000000001
0.649576
0.663361
0.6614
0.650284
0.64456
0.660865
0.6648860000000001
```

AFTER (note JIT output and significant improvement in the performance of each thread's execution)
```
2025-01-17T13:44:20.562-06:00 [main] ERROR Ruby : failed to compile target script: -e - AOT not supported for scripts containing eval.
2025-01-17T13:44:20.695-06:00 [main] INFO JITCompiler : block done jitting: Object new &||1 at (eval at -e:1):0
0.19823400000000002
2025-01-17T13:44:21.925-06:00 [main] INFO JITCompiler : block done jitting: Object new &||1 at (eval at -e:1):0
0.15478
2025-01-17T13:44:23.089-06:00 [main] INFO JITCompiler : block done jitting: Object new &||1 at (eval at -e:1):0
0.153646
2025-01-17T13:44:24.256-06:00 [main] INFO JITCompiler : block done jitting: Object new &||1 at (eval at -e:1):0
0.080538
2025-01-17T13:44:25.346-06:00 [main] INFO JITCompiler : block done jitting: Object new &||1 at (eval at -e:1):0
0.07915900000000001
2025-01-17T13:44:26.432-06:00 [main] INFO JITCompiler : block done jitting: Object new &||1 at (eval at -e:1):0
0.07181399999999999
2025-01-17T13:44:27.510-06:00 [main] INFO JITCompiler : block done jitting: Object new &||1 at (eval at -e:1):0
0.062474
```

headius:
Note idea came out of #8406, where a one-shot, long-running background thread services all timeouts. If that thread's code immediately optimized, it may be able to keep up with the onslaught of tiny timeout jobs.

headius:
I experimented with this optimization and the timeout issue in #8406, and it does help, but the many client threads still swamp the one timeout thread. See ruby/timeout#58 for the ongoing issue.

Issue #8602, opened 2025-01-30
ConcurrentModificationException in subclasses map


headius:
Discovered by @kalenp while attempting to deploy 9.4.11.0 to their test environments.

It appears at least one call path into the new subclasses collections (Map.remove) was not properly synchronized, which can lead to ConcurrentModificationException under concurrent load and modification of the class hierarchy.

The following test is sufficient to reproduce the exception most of the time. There are many different types of class hierarchy modification here and several attempts to force traversal of all subclasses, so I'm unsure which operations specifically can trigger this error. The test is intended to try everything under reasonably high concurrent load.

```ruby
require 'test/unit'

class TestClassSubclasses < Test::Unit::TestCase
  def test_concurrent_modification
    c = Class.new
    c2 = Class.new(c)
    go = false

    thread_count = 100
    threads = thread_count.times.map {
      Thread.new {
        Thread.pass until go
        c3 = Class.new(c2)
        o = c3.new
        c.class_eval {
          def foo; end
        }
        class << o
          extend Enumerable
          prepend Comparable
        end
        c.class_eval {
          def foo; end
        }
        c3.class_eval {
          include Enumerable
          prepend Comparable
        }
        c.class_eval {
          def foo; end
        }
        :success
      }
    }

    go = true
    assert_equal(threads.map(&:value), [:success] * thread_count)
  end
end
```

A high percentage of runs of this test will produce output similar to the following:

```
$ jruby test/jruby/test_class_subclasses.rb
Loaded suite test/jruby/test_class_subclasses
Started
java.util.ConcurrentModificationException
	at java.base/java.util.WeakHashMap.forEach(WeakHashMap.java:1037)
	at org.jruby.dist/org.jruby.RubyClass$SubclassesWeakMap.forEach(RubyClass.java:1191)
	at org.jruby.dist/org.jruby.RubyClass.invalidateCacheDescendants(RubyClass.java:1255)
	at org.jruby.dist/org.jruby.RubyClass.lambda$invalidateCacheDescendants$5(RubyClass.java:1255)
	at java.base/java.util.WeakHashMap.forEach(WeakHashMap.java:java.util.ConcurrentModificationException
	at java.base/java.util.WeakHashMap.forEach(WeakHashMap.java:1037)
	at org.jruby.dist/org.jruby.RubyClass$SubclassesWeakMap.forEach(RubyClass.java:1191)
[...]

PR #8603, opened 2025-01-30
Ensure subclass set prevents concurrent modification

headius:
The new SubclassesWeakMap properly included read/write locking for the `put` and `forEach` methods, but unfortunately the `remove` and `isEmpty` methods were not overridden with locking. This went unnoticed until after the 9.4.11.0 release, leading to jruby/jruby#8602.

This patch makes two modifications to ensure proper locking:

* Instead of working with the full Map interface, and possibly missing other method uses in the future, this narrows the subclasses map interface to a RubyClassSet with only unique methods that we use. This guarantees only these methods can be called, and we can ensure they are implemented properly.
* Add the missing locking to the `remove` method, now named `removeClass`.

Fixes jruby/jruby#8602.

### DIFF
--- a/core/src/main/java/org/jruby/RubyClass.java
+++ b/core/src/main/java/org/jruby/RubyClass.java
@@ -54,7 +54,7 @@ import java.lang.reflect.Method;
 import java.lang.reflect.Modifier;
 import java.util.*;
 import java.util.concurrent.atomic.AtomicReferenceFieldUpdater;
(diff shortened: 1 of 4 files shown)


headius:
An additional change may be needed here due to several read operations mutating the collection to clean out stale entries. We may need to return to a fully locked implementation while using this implementation of WeakHashMap.

PR #8607, opened 2025-01-30
Update specs to ruby/spec@affef93

andrykonchin:


headius:
Thanks @andrykonchin!

Issue #8611, opened 2025-02-05
Setting an instance var on a Java object no longer warns


headius:
In 1.7 we stopped caching the wrappers for Java objects that enter Ruby, and introduced warnings for two cases that could be affected:

* creating a singleton object from one of these wrappers
* setting instance variables on one of these wrappers

The singleton case still warns today (JRuby 9.4.12.0-SNAPSHOT on master), but the instance variable case does not:

```
$ jruby -e 'class << java.lang.Object.new; end'
-e:1: warning: singleton on non-persistent Java type Java::JavaLang::Object (https://github.com/jruby/jruby/wiki/Persistence)
$ jruby -e 'java.lang.Object.new.instance_variable_set(:@foo, 1)'
$ jruby -X-C -e 'java.lang.Object.new.instance_variable_set(:@foo, 1)'
$
```

We must restore this warning and add tests to ensure it is not broken again.

Relates to #8224.

headius:
Looking at the code, this is likely due to us unifying all ivar set/get operations through the VariableTableManager and VariableTableAccessor instances. Specifically in this case, the StampedVariableAccessor is used [here](https://github.com/jruby/jruby/blob/807900a0478fd4d9a5509eb4ad1172a1a89a2c1e/core/src/main/java/org/jruby/runtime/ivars/StampedVariableAccessor.java#L62) ultimately bottoming out in atomic logic [here](https://github.com/jruby/jruby/blob/807900a0478fd4d9a5509eb4ad1172a1a89a2c1e/core/src/main/java/org/jruby/runtime/ivars/StampedVariableAccessor.java#L89-L106) that skips the overridden `getVariable`/`setVariable` ([here](https://github.com/jruby/jruby/blob/89709aa77eda223f95fb2d2464760af5d474355a/core/src/main/java/org/jruby/java/proxies/JavaProxy.java#L566-L576)) and does not check if we are attempting to set variables on a proxy object.

PR #8613, opened 2025-02-05
Warn when creating a new ivar table on JavaProxy wrapper

headius:
This warning was lost when we optimized variable updates to skip RubyBasicObject's `setVariable` as reported in https://github.com/jruby/jruby/issues/8611. The changes here restore the warning:

* Move common logic for initializing the variable table to the VariableAccessor superclass.
* Ensure all places where a new variable table is created check for JavaProxy and warn appropriately.

Note that field-based variable access does not have this same proxy check because JavaProxy will never use field-based variables.


headius:
Do not merge until tests have been added.

PR #8616, opened 2025-02-09
Eiminate last uses of UnsafeHolder

headius:
This removes the last usage of UnsafeHolder and removes it.

This also migrates NonBlockingHashMapLong to use VarHandle instead of either Unsafe or Atomic* logic. This change also restores atomicity to all logic.


headius:
@kares You did a conversion long ago to support NonBlockingHashMapLong without Unsafe, could you review this to make sure I removed everything correctly?

PR #8617, opened 2025-02-09
Reduce RubyModule/RubyClass size and creation overhead

headius:
We have too much allocation and overhead in the construction logic for RubyModule and all subtypes. This increase overall memory footprint, slows startup, and slows down runtime class creation (especially for singleton objects). This PR will reduce that overhead by:

* eliminating fields and structures that are no longer necessary,
* reducing the overhead of remaining fields by initializing them lazily or using more efficient designs,
* avoid initialization overhead for features that will never be used, like method caches on include/prepend wrappers.

The goal is to have only the minimum overhead and state needed at creation time for each type of "module", so lightweight classes can be created and discarded quickly.


headius:
@kares @enebo If you get a chance to look this over I would appreciate it, but I'm going to move forward and merge. Biggest area I want more eyes on is the lazy volatile inits and the refinementStore object.

Issue #8610, opened 2025-02-05
Automatic AppCDS usage in JRuby launcher


headius:
Recent versions of the JDK have continued to show great improvement in startup from using AppCDS (Class Data Store). Previously, there were two ways to generate this archive for a given JRuby install: manually passing the AppCDS flags to the JVM; or using the `jruby-startup` gem's `generate-appcds` command. Once generated, the existing `jruby` launcher script detects the archive and uses it to reduce startup:

https://gist.github.com/headius/94cb33aa2acbbaf433c2ad0fbfefcfb5

These improvements apply to larger commands as well, with diminishing returns as the overall run-time of the command increases.

We would like to make the generation and use of AppCDS archives entirely automatic, built into the launcher script, so that even users with no knowledge of this extra startup trick can benefit.

The new feature would have the following requirements:

* Lazily generate an AppCDS archive whenever an appropriate one cannot be found.
* Store the AppCDS archive somewhere local to the user, rather than in the JRuby dist directory, to ensure it can be written and controlled by the local user.
* The AppCDS archive store should be structured or keyed in such a way that if the installed JRuby changes or if the user switches to a different JDK it will not conflict with a now-stale archive.
* A given version of JRuby on a given version of JDK (where version is either numeric release or vendor) should always use AppCDS and produce no warnings.
* Optional: some mechanism for cleaning out old AppCDS archives that have not been used in some time.

The structure of the AppCDS storage could be based on real path to the JDK, which might produce false positives if upgrading a given JDK means overwriting old files, or based on actual JDK version, which may not be queriable without running at least one command line.

The bottom-line goal is to automatically produce and use AppCDS archives that are specific (enough) to the current JRuby and JDK to avoid any warnings or errors (from either JRuby or the JDK itself), and to prevent those archives from becoming stale (causing JDK to warn) or downright old (potentially just booting wrong).

headius:
@mrnoname1000 This is the feature request I chatted about with you at FOSDEM.

mrnoname1000:
On the Java side it looks like we can get all we need from the `$JAVA_HOME/release` file, but finding the version of JRuby looks like it'll be more difficult. The JRuby JAR has some metadata we can extract with an unzipping program, but the standard tarballs don't seem to have any version info stored in a machine-readable file. I do see a VERSION file in the JRuby repo though, how much work would it be to include that in the tarballs?

headius:
[...]

PR #8625, opened 2025-02-12
Add --cache flag to regenerate AppCDS

headius:
This experimental flag can be passed to JRuby, with or without Ruby arguments, to regenerate the AppCDS archive jruby.jsa under the lib dir of the current JRuby install. This file is automatically used if present by the launcher to improve startup time.


headius:
Relates to #8610.

bin/jruby.sh
```diff
@@ -628,6 +633,11 @@ fi
 # The rest of the arguments are for ruby
 append ruby_args "$@"
 
+# If regenerating the JSA archive but no Ruby arguments were passed, do -e 1
+if $regenerate_jsa_file & [ \'"$ruby_args"\' = "' '" ]; then
+    assign ruby_args '-e' '1'
+fi
```
mrnoname1000, line 639: That looks like a pretty fragile way to check that the `ruby_args` array is empty, I would prefer to either parse it and check `$#` or we could get clever and match against `*[![:space:]]*`. Also quotes are unnecessary on line 638

mrnoname1000: Oh and there's a bug here, `&` forks off a subshell so if `[ "$ruby_args" = " " ]` then this will always execute

PR #8626, opened 2025-02-12
Fix default arguments for generating AppCDS flag

mrnoname1000:
Fixes a bug in #8625 and relates to features discussed in #8610


headius:
Thank you! I had several iterations of this trying to get it right.

PR #8635, opened 2025-02-18
Fix `tr!` and `reverse!` on ByteList view with non-zero begin index

### LINKED ISSUE #8624: Inconsistent String behavior between MRI and JRuby
Yesterday I noticed an inconsistent behavior related to Strings using GitHub Actions and jruby-head.
Today I was able to obtain the following minimal reproducible example using Linux and RVM:
```ruby
s = '!test'
s.delete_prefix!('!')
p s
s.tr!('-','_')
p s
```

The expected behavior is to output ``"test"`` twice.
Output of  MRI 3.4.1:
```
"test"
"test"
```

Output of JRuby head (CI environment) and JRuby 9.4.10.0 (RVM locally):
```
"test"
"est\u0000"
```

ntkme:
- Fixes #8624

### DIFF
--- a/core/src/main/java/org/jruby/RubyString.java
+++ b/core/src/main/java/org/jruby/RubyString.java
@@ -1606,6 +1606,7 @@ public class RubyString extends RubyObject implements CharSequence, EncodingCapa
                     p += cl;
                 }
                 value.setUnsafeBytes(obytes);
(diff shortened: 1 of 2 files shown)


enebo:
@ntkme Thanks for the analysis and fix!

PR #8643, opened 2025-02-19
Fixes to get stringio tests and specs green

headius:
After working on getting the CRuby stringio tests green in ruby/stringio#116 I turned my attention to stringio specs. These fixes are intended to get the specs green as well.


headius:
These will need ruby/stringio#116 to be merged and released before being green.

PR #8644, opened 2025-02-20
Check if tmpStream is null to avoid a NullPointerException.

nabice:
In JDK versions higher than 9, tmpStream always returns null. Reducing the use of exceptions can decrease resource consumption.


headius:
Good catch!

headius:
Could you rebase this against the `jruby-9.4` branch please? The main branch has moved on to JRuby 10 work now and we should fix this in 9.4 as well.

PR #8648, opened 2025-02-21
Handle errors whe looking for Java executable

mrnoname1000:
Should fix https://github.com/jruby/jruby-launcher/issues/44. Doesn't explicitly check if `java` is a directory but `command -v` will  fail because it's not a file that can be executed.

### DIFF
--- a/bin/jruby.sh
+++ b/bin/jruby.sh
@@ -375,23 +375,29 @@ if [ -z "$JAVACMD" ]; then
             && [ ! -d "$java_home_command" ]
         then
             # use java_home command when none is set (on MacOS)
-            JAVA_HOME="$("$java_home_command")"
+            JAVA_HOME="$("$java_home_command")" &&
             JAVACMD="$JAVA_HOME"/bin/java
         else
             # Linux and others have a chain of symlinks
-            resolve "$(command -v java)"
+            JAVACMD="$(command -v java)" &&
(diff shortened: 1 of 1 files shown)


headius:
This should be rebased against master.

If I'm reading right, these are launching a subshell to check the `java` executable, yes? Can't we just check if it is a normal executable file more cheaply?

enebo:
If java must get launched I think we entertain ```java -Xinternalversion```,  This would not only test validity but could also be used for proper version handling.  This is only 0.004s on my machine (I guess MacOS is higher).

mrnoname1000:
> If I'm reading right, these are launching a subshell to check the java executable, yes?

It is launching a subshell but it was always doing that because it's necessary to capture the output of `command -v` (which is a pure-shell alternative to `which`). It's not actually running any external programs, it's just checking that there's a file called `java` in `$PATH` that's executable. We could check that it succeeds before running it in a subshell but I don't think it's really necessary to save one fork on a code path that's going to error anyway.

headius:
@mrnoname1000 Ok that seems find then. We can merge this once you rebase it on master.

headius:
We can merge this, as it is a good change, but it doesn't actually fix jruby/jruby-launcher#44. That issue is actually about the **native** launcher, not the shell script.

PR #8653, opened 2025-02-22
Align 9.4 launcher with 10

headius:
Features like --logcache and --nocache go hand in hand with --cache and automatic use of the jsa file. The rest of the changes largely support using jsa.


headius:
Rebased on master with #8648 incorporated.

PR #8654, opened 2025-02-23
fix glob dir on windows

### LINKED ISSUE #8623: Installing sassc broke on jruby-head on Windows
See https://github.com/ruby/setup-ruby/actions/runs/13279868644/job/37076053011?pr=705
It also fails on windows-2022 and windows-2025, so all jruby-head + windows.
It looks like a regression because AFAIK this is the first time it fails.

Output:
```
gem install sassc
Building native extensions. This could take a while...
ERROR:  Error installing sassc:
	ERROR: Failed to build gem native extension.

    current directory: D:/jruby-head/lib/ruby/gems/shared/gems/sassc-2.4.0/ext
D:/jruby-head/bin/jruby.exe extconf.rb
creating Makefile

current directory: D:/jruby-head/lib/ruby/gems/shared/gems/sassc-2.4.0/ext
make DESTDIR\= sitearchdir\=./.gem.20250212-744-b2t36v sitelibdir\=./.gem.20250212-744-b2t36v clean

current directory: D:/jruby-head/lib/ruby/gems/shared/gems/sassc-2.4.0/ext
make DESTDIR\= sitearchdir\=./.gem.20250212-744-b2t36v sitelibdir\=./.gem.20250212-744-b2t36v
compiling ./libsass/src/ast.cpp
compiling ./libsass/src/ast2c.cpp
compiling ./libsass/src/ast_fwd_decl.cpp
compiling ./libsass/src/ast_sel_cmp.cpp
compiling ./libsass/src/ast_sel_super.cpp
compiling ./libsass/src/ast_sel_unify.cpp
compiling ./libsass/src/ast_sel_weave.cpp
compiling ./libsass/src/ast_selectors.cpp
compiling ./libsass/src/ast_supports.cpp
compiling ./libsass/src/ast_values.cpp
compiling ./libsass/src/backtrace.cpp
compiling ./libsass/src/base64vlq.cpp
compiling ./libsass/src/bind.cpp
compiling ./libsass/src/c2ast.cpp
compiling ./libsass/src/c99func.c
compiling ./libsass/src/cencode.c
compiling ./libsass/src/check_nesting.cpp
compiling ./libsass/src/color_maps.cpp
compiling ./libsass/src/constants.cpp
compiling ./libsass/src/context.cpp
compiling ./libsass/src/cssize.cpp
compiling ./libsass/src/emitter.cpp
compiling ./libsass/src/environment.cpp
compiling ./libsass/src/error_handling.cpp
compiling ./libsass/src/eval.cpp
compiling ./libsass/src/eval_selectors.cpp
compiling ./libsass/src/expand.cpp
compiling ./libsass/src/extender.cpp
compiling ./libsass/src/extension.cpp
compiling ./libsass/src/file.cpp
compiling ./libsass/src/fn_colors.cpp
compiling ./libsass/src/fn_lists.cpp
compiling ./libsass/src/fn_maps.cpp
compiling ./libsass/src/fn_miscs.cpp
compiling ./libsass/src/fn_numbers.cpp
compiling ./libsass/src/fn_selectors.cpp
compiling ./libsass/src/fn_strings.cpp
compiling ./libsass/src/fn_utils.cpp
compiling ./libsass/src/inspect.cpp
compiling ./libsass/src/json.cpp
compiling ./libsass/src/lexer.cpp
compiling ./libsass/src/listize.cpp
compiling ./libsass/src/operators.cpp
compiling ./libsass/src/output.cpp
compiling ./libsass/src/parser.cpp
compiling ./libsass/src/parser_selectors.cpp
compiling ./libsass/src/plugins.cpp
compiling ./libsass/src/position.cpp
compiling ./libsass/src/prelexer.cpp
compiling ./libsass/src/remove_placeholders.cpp
compiling ./libsass/src/sass.cpp
compiling ./libsass/src/sass2scss.cpp
compiling ./libsass/src/sass_context.cpp
compiling ./libsass/src/sass_functions.cpp
compiling ./libsass/src/sass_values.cpp
compiling ./libsass/src/source.cpp
compiling ./libsass/src/source_map.cpp
compiling ./libsass/src/stylesheet.cpp
compiling ./libsass/src/to_value.cpp
compiling ./libsass/src/units.cpp
compiling ./libsass/src/utf8_string.cpp
compiling ./libsass/src/util.cpp
compiling ./libsass/src/util_string.cpp
compiling ./libsass/src/values.cpp
linking shared-object sassc/libsass.dll
C:/msys64/mingw64/bin/../lib/gcc/x86_64-w64-mingw32/14.2.0/../../../../x86_64-w64-mingw32/bin/ld.exe: cannot find allocator.o: No such file or directory
C:/msys64/mingw64/bin/../lib/gcc/x86_64-w64-mingw32/14.2.0/../../../../x86_64-w64-mingw32/bin/ld.exe: cannot find shared_ptr.o: No such file or directory
collect2.exe: error: ld returned 1 exit status
make: *** [Makefile:221: libsass.dll] Error 1

make failed, exit code 2

Gem files will remain installed in D:/jruby-head/lib/ruby/gems/shared/gems/sassc-2.4.0 for inspection.
Results logged to D:/jruby-head/lib/ruby/gems/shared/extensions/universal-java-1.8/3.1.0/sassc-2.4.0/gem_make.out
Error: Process completed with exit code 1.
```

danini-the-panini:
Resolve #8623

### DIFF
--- a/core/src/main/java/org/jruby/util/Dir.java
+++ b/core/src/main/java/org/jruby/util/Dir.java
@@ -976,7 +976,7 @@ public class Dir {
                             buf.append(base);
                             buf.append( isRoot(base) ? EMPTY : SLASH );
                             buf.append( getBytesInUTF8(file) );
-                            resource = JRubyFile.createResource(runtime, cwd, RubyString.byteListToString(buf));
+                            resource = JRubyFile.createResource(runtime, cwd, new String(buf.unsafeBytes(), buf.begin(), buf.length(), enc.getCharset()));
                             boolean dirMatch = SLASH_INDEX == end-1 && resource.isDirectory();
                             if ( dirMatch || SLASH_INDEX == -1 ) {
                                 if (scheme != null) {
--- a/core/src/main/java/org/jruby/util/Dir.java
+++ b/core/src/main/java/org/jruby/util/Dir.java
@@ -976,8 +976,11 @@ public class Dir {
                             buf.append(base);
                             buf.append( isRoot(base) ? EMPTY : SLASH );
                             buf.append( getBytesInUTF8(file) );
-                            resource = JRubyFile.createResource(runtime, cwd, new String(buf.unsafeBytes(), buf.begin(), buf.length(), enc.getCharset()));
-                            boolean dirMatch = SLASH_INDEX == end-1 && resource.isDirectory();
+                            boolean dirMatch = false;
+                            if (SLASH_INDEX == end - 1) {
+                                resource = JRubyFile.createResource(runtime, cwd, new String(buf.unsafeBytes(), buf.begin(), buf.length(), enc.getCharset()));
+                                dirMatch = resource.isDirectory();
+                            }
                             if ( dirMatch || SLASH_INDEX == -1 ) {
                                 if (scheme != null) {
                                     byte[] bufBytes = buf.bytes();
@@ -990,7 +993,7 @@ public class Dir {
                                 if ( status != 0 ) break;
                                 continue;
                             }
-                            if (resource.isDirectory()) links.add(new DirGlobber(buf));
+                            links.add(new DirGlobber(buf));
                             buf = new ByteList(20);
                             buf.setEncoding(enc);
                         }
@@ -1001,11 +1004,14 @@ public class Dir {
                     for ( DirGlobber globber : links ) {
                         final ByteList link = globber.link;
                         if ( status == 0 ) {
-                            final int len = link.getRealSize();
-                            buf.length(0);
-                            buf.append(link);
-                            buf.append(path, SLASH_INDEX, end - SLASH_INDEX);
-                            status = glob_helper(runtime, cwd, scheme, buf, buf.getBegin() + len, flags, func, arg);
+                            resource = JRubyFile.createResource(runtime, cwd, new String(link.unsafeBytes(), link.begin(), link.length(), enc.getCharset()));
+                            if ( resource.isDirectory() ) {
+                                final int len = link.getRealSize();
+                                buf.length(0);
+                                buf.append(link);
+                                buf.append(path, SLASH_INDEX, end - SLASH_INDEX);
+                                status = glob_helper(runtime, cwd, scheme, buf, buf.getBegin() + len, flags, func, arg);
+                            }
                         }
                     }
                     break mainLoop;
--- a/core/src/main/java/org/jruby/util/Dir.java
+++ b/core/src/main/java/org/jruby/util/Dir.java
@@ -973,16 +973,24 @@ public class Dir {
                         }
                         if ( fnmatch(magic, 0, magic.length, fileBytes, 0, fileBytes.length, flags) == 0 ) {
                             buf.length(0);
-                            if (scheme != null && SLASH_INDEX == -1) buf.append(scheme);
                             buf.append(base);
                             buf.append( isRoot(base) ? EMPTY : SLASH );
                             buf.append( getBytesInUTF8(file) );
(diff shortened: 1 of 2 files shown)


core/src/main/java/org/jruby/util/Dir.java
```diff
@@ -973,16 +973,24 @@ private static int glob_helper(Ruby runtime, String cwd, byte[] scheme,
                         }
                         if ( fnmatch(magic, 0, magic.length, fileBytes, 0, fileBytes.length, flags) == 0 ) {
                             buf.length(0);
-                            if (scheme != null && SLASH_INDEX == -1) buf.append(scheme);
                             buf.append(base);
                             buf.append( isRoot(base) ? EMPTY : SLASH );
                             buf.append( getBytesInUTF8(file) );
-                            if ( SLASH_INDEX == -1 ) {
+                            resource = JRubyFile.createResource(runtime, cwd, RubyString.byteListToString(buf));
```
enebo, line 979: Up above we see the same resource creation but it uses the encoding for charset:

```java
resource = JRubyFile.createResource(runtime, cwd, new String(buf.unsafeBytes(), buf.begin(), buf.length(), enc.getCharset()));
```

This enc comes from the original pattern and that pattern is fed through RubyFile#get_path/filePathConvert which has special logic for transcoding if NOT on windows to match filesystem encoding.   I think the use of byteListToString (used both here and below) is wrong.  It will make an ISO-8859-1 string where each byte is a char vs the line above which properly decodes the bytes given with the charset provided.

I think the reason we do not ever see problems is almost all files are single-byte ASCII.

enebo (changes requested):
I believe we should be using the String construction listed above in my comment for your createResource line.  I think links.size() > 0 section also uses the same call and it should also be changed.

I am going to ask @headius on channel to review this since glob is always too complicated.

core/src/main/java/org/jruby/util/Dir.java
headius: Yes I think this risks losing encoding and possibly dealing with mangled characters. If we cannot fix it by working with the possibly-multibyte characters directly, then we should use the functions in RubyEncoding to decode them to String properly.

core/src/main/java/org/jruby/util/Dir.java
```diff
@@ -993,14 +1001,11 @@ private static int glob_helper(Ruby runtime, String cwd, byte[] scheme,
...
-                                buf.length(0);
-                                buf.append(link);
-                                buf.append(path, SLASH_INDEX, end - SLASH_INDEX);
-                                status = glob_helper(runtime, cwd, scheme, buf, buf.getBegin() + len, flags, func, arg);
-                            }
+                            final int len = link.getRealSize();
+                            buf.length(0);
+                            buf.append(link);
+                            buf.append(path, SLASH_INDEX, end - SLASH_INDEX);
+                            status = glob_helper(runtime, cwd, scheme, buf, buf.getBegin() + len, flags, func, arg);
```
danini-the-panini, line 1008: so i moved the `isDirectory` check to where we append to `links` rather than the loop. This _might_ cause an issue  where links contains only non-directories, it should still call `break mainLoop`, although I haven't seen any issues. I don't know how I would cause this to happen.

core/src/main/java/org/jruby/util/Dir.java
danini-the-panini: sorted

core/src/main/java/org/jruby/util/Dir.java
danini-the-panini: I've reverted this change

PR #8657, opened 2025-02-25
Add EditorConfig file

mrnoname1000:
I finally got tired of accidentally indenting the launcher with tabs 😅

This PR adds an [EditorConfig](https://editorconfig.org) file for standardized code style settings across editors. It comes included recent versions of Vim/Neovim, and IntelliJ since about 2014. Not sure what the desired style settings would be for `*.rb` or `*.java` but I can add those if desired.

### DIFF
--- a/.editorconfig
+++ b/.editorconfig
@@ -1,0 +1,9 @@
+root = true
+
+[*]
+end_of_line = lf
+insert_final_newline = true
+
+[*.{sh,bash}]
+indent_style = space
+indent_size = 4


headius:
I would also like to commit some code standards to the repo (we still get folks occasionally submitting patches with 2-space or 1-tab indents). However this format seems sort of dead? I went to the repository for it and there's hardly any commits in the last five years.

mrnoname1000:
Which repo did you go to? The [specification repo](https://github.com/editorconfig/specification) has had several updates throughout the past year. Check out [their other repositories](https://github.com/orgs/editorconfig/repositories) too, the library and plugin repos seem well maintained. They even have a [maven plugin](https://github.com/editorconfig/editorconfig-maven-plugin) which JRuby could use if all the source was already conformant.

headius:
Ok that is more promising. I just searched for "editorconfig" and eventually ended up at https://github.com/editorconfig/editorconfig.

@enebo I have no objections to including this. Even IntelliJ recognizes it.

PR #8662, opened 2025-02-26
Start implementing IO#timeout

headius:
This only implements it for a couple forms of read and for Socket#connect. Other places still need to be wired up.


headius:
This may not be the right approach. The select methods called are frequently called in a loop until they produce an error or return a true result. In the cast of `Socket#connect`, if a timeout is exceeded, it will always return false without raising an error.

We probably need to enhance the RubyThread.select logic to optionally raise a timeout error rather than just return false, since there's no way outside of these methods to determine that the select timed out versus simply selecting nothing.

Issue #8661, opened 2025-02-26
Update Unicode to 15.0


headius:
For JRuby 10 we will not update Unicode, but this should happen in an update release.

Failures from ruby/spec:

```
50)
RbConfig::CONFIG['UNICODE_EMOJI_VERSION'] is 15.0 FAILED
Expected "13.1" == "15.0"
to be truthy but was false
/Users/headius/work/jruby/spec/ruby/library/rbconfig/unicode_emoji_version_spec.rb:20:in 'block in <main>'
org/jruby/RubyBasicObject.java:2665:in 'instance_exec'
org/jruby/RubyArray.java:4919:in 'all?'
org/jruby/RubyArray.java:2123:in 'each'
/Users/headius/work/jruby/spec/ruby/library/rbconfig/unicode_emoji_version_spec.rb:4:in '<main>'
org/jruby/RubyKernel.java:1179:in 'load'
org/jruby/RubyBasicObject.java:2665:in 'instance_exec'
org/jruby/RubyArray.java:2123:in 'each'

51)
RbConfig::CONFIG['UNICODE_VERSION'] is 15.0.0 FAILED
Expected "13.0.0" == "15.0.0"
to be truthy but was false
/Users/headius/work/jruby/spec/ruby/library/rbconfig/unicode_version_spec.rb:20:in 'block in <main>'
org/jruby/RubyBasicObject.java:2665:in 'instance_exec'
org/jruby/RubyArray.java:4919:in 'all?'
org/jruby/RubyArray.java:2123:in 'each'
/Users/headius/work/jruby/spec/ruby/library/rbconfig/unicode_version_spec.rb:4:in '<main>'
org/jruby/RubyKernel.java:1179:in 'load'
org/jruby/RubyBasicObject.java:2665:in 'instance_exec'
org/jruby/RubyArray.java:2123:in 'each'
```

ahorek:
updated in https://github.com/jruby/jcodings/pull/66

ahorek:
@headius could you update joni before the release to satisfy the dependency check? thanks

headius:
jcodings 1.0.62 has been released as well as joni 2.2.4 with the updated jcodings!

headius:
I have pushed #8667 to update jcodings and joni, but we may not want to do that in a 9.4 update. In any case, this will be fixed by #8667.

ahorek:
technically, the change belongs to 10.0, but it would block backporting future updates in joni and jcodings to version 9.4.

[...]

PR #8664, opened 2025-02-28
Unicode 15.0

ahorek:
https://github.com/jruby/jruby/issues/8661


enebo:
Thanks @ahorek!  We will still need an update to joni for new jcodings but I think @headius said he would do that this week.

PR #8667, opened 2025-03-03
Update jcodings to 1.0.62 and joni to 2.2.4

headius:
jcodings 1.0.62:

* update unicode to 15.0 (jruby/jcodings#66)

joni 2.2.4:

* update jcodings to 1.0.62
* fix: char class casefold for certain chars (jruby/joni#85)

### DIFF
--- a/core/pom.rb
+++ b/core/pom.rb
@@ -51,8 +51,8 @@ project 'JRuby Base' do
   jar 'com.github.jnr:jffi:${jffi.version}'
   jar 'com.github.jnr:jffi:${jffi.version}:native'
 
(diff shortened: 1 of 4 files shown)


headius:
@enebo This hasn't been an issue in the past, but should we rebase this to update 10 only? Ruby 3.1 still reports Unicode version 13, so this will probably break a few minor tests on 9.4.

headius:
As expected there's failures checking the Unicode version:

https://github.com/jruby/jruby/actions/runs/13638643350/job/38123298554?pr=8667

Tagging these would be fine though, I think.

PR #8671, opened 2025-03-04
fix a regression in KDDI transcoding

ahorek:
fixes https://github.com/jruby/jcodings/pull/64
depends on https://github.com/jruby/jcodings/pull/67

### DIFF
--- a/core/pom.rb
+++ b/core/pom.rb
@@ -51,8 +51,8 @@ project 'JRuby Base' do
   jar 'com.github.jnr:jffi:${jffi.version}'
   jar 'com.github.jnr:jffi:${jffi.version}:native'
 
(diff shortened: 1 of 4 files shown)


headius:
Some commits snuck into master but not jruby-9.4 so I rebased the latter. Shouldn't affect this if you rebase the PR.

Once we have a repaired jcodings I will merge that to 9.4 and you can merge or rebase onto the PR.

headius:
@ahorek We're close! Looks like two failures here after the updated versions of jcodings and joni are available. I get the same failures locally without the jcodings/joni updates, so it's not a regression from anything you did. I'm going to go ahead and merge this and exclude from the jruby-9.4 branch. If you feel so inclined to keep digging into those last two, go for it!

PR #8685, opened 2025-03-10
Additional improvements to new marshal logic

### LINKED ISSUE #8683: New marshal could break ext authors using old one
From revapi work:

```text
type: java.method.addedToInterface - Method was added to an interface.
   
    method void org.jruby.runtime.ObjectMarshal<T>::marshalTo(T, RubyClass, org.jruby.runtime.marshal.NewMarshal, ThreadContext, org.jruby.runtime.marshal.NewMarshal.RubyOutputStream)
    semantic: potentially_breaking, source: breaking, binary: non_breaking
```

This new method will break any existing implements of ObjectMarshal (if there are any).

headius:
This PR is intended to address the concerns mentioned in #8683.

Fixes #8683.

Goals:

* [x] Mitigate new interface impl requirements (primary concern from #8683).
* [x] Propagate ThreadContext consistently and never re-acquire it (mentioned here: https://github.com/jruby/jruby/issues/8683#issuecomment-2707464039).


headius:
@enebo I have implemented the context-propagation part of this, which did find a few places that were forced to re-acquire the context.

However I now feel it is excessive to have a special RubyOutputStream that can propagate the context solely for the purpose of raising an error. I'd like your thoughts on how to reduce or eliminate RubyOutputStream, since all it does is raise Ruby errors and drop `throws IOException` from the standard OutputStream write methods.

enebo:
@headius yeah let's chat about this when we are both around.

PR #8691, opened 2025-03-11
Implement AutoCloseable on JRubyEngine

### LINKED ISSUE #8690: Using JSR223 ScriptEngine interface has no way to clean up resources
**Expected Behavior**

I embed JRuby into openHAB, using the JSR223 interface (since that's how openHAB handles several different scripting languages). A new `ScriptEngine` is created for each script, including if it's reloaded because I've changed it. openHAB's core code that manages scripts will check if the `ScriptEngine` implements `AutoCloseable` when it's unloading a script, and call close() on it if it does. But `JRubyEngine` doesn't implement `AutoCloseable`, nor does it have any way to access the ScriptingContainer in order for me to call `terminate` myself. I would expect it to implement `AutoCloseable`, and calling `close()` on the engine will just call `container.terminate()`.

**Actual Behavior**

When I reload a lot of scripts, my memory usage goes up, and eventually I get to the point where GC is using 100% CPU because it's having such a hard time freeing up resources from old engines.

ccutrer:
Closes #8690

### DIFF
--- a/core/src/main/java/org/jruby/embed/jsr223/JRubyEngine.java
+++ b/core/src/main/java/org/jruby/embed/jsr223/JRubyEngine.java
@@ -57,7 +57,7 @@ import org.jruby.runtime.builtin.IRubyObject;
  *
  * @author Yoko Harada &lt;<a href="mailto:yokolet@gmail.com">yokolet@gmail.com</a>&gt;
  */
-public class JRubyEngine implements Compilable, Invocable, ScriptEngine {
+public class JRubyEngine implements AutoCloseable, Compilable, Invocable, ScriptEngine {
 
     final ScriptingContainer container;
     private final JRubyEngineFactory factory;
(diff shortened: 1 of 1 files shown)


headius:
Could you rebase this on the jruby-9.4 branch? We should fix it there as well.

ccutrer:
I'm not sure the current state of things with 9.4 and 10. Do you want a separate PR for 9.4, or just re-target this one to 9.4?

headius:
We always merge forward so changes to 9.4 will get into master/10.

Are the .project changes necessary for the AutoCloseable? If not, what are they for? Probably should be a separate PR if it's unrelated.

ccutrer:
> Are the .project changes necessary for the AutoCloseable?

Nope, that was probably just VSCode doing something weird, and I didn't notice them when I did `git commit -a`.

ccutrer:
Okay, this PR now targets 9.4, and I removed the spurious project changes.

enebo:
@ccutrer Thanks for putting this together.  Our CI did not rerun on jruby-9.4 for unknown reasons but this should be fine on jruby-9.4 (10 is still not quite fully green yet).

PR #8702, opened 2025-03-17
was recursively warning on itself

enebo:
A chilled string was warning on itself in a StringIO which was calling itself somehow (inspectin value?) and it would recursively go back to warning about chill on itself.  The scenario why is not 100% solid for me but the reason is.


headius:
Perhaps it was dyncalling a Ruby `warn` and that triggered further chilled string warnings?

Issue #8696, opened 2025-03-13
New restriction warnings on Java 24


headius:
There are two new warnings for a basic `-e 1` on Java 24:

1. Restriction on native library loading that must be enabled in the future:

```
WARNING: A restricted method in java.lang.System has been called
WARNING: java.lang.System::load has been called by com.kenai.jffi.internal.StubLoader in module org.jruby.dist (file:/Users/headius/work/jruby/lib/jruby.jar)
WARNING: Use --enable-native-access=org.jruby.dist to avoid a warning for callers in this module
WARNING: Restricted methods will be blocked in a future release unless native access is enabled

WARNING: A terminally deprecated method in sun.misc.Unsafe has been called
WARNING: sun.misc.Unsafe::putLong has been called by com.kenai.jffi.UnsafeMemoryIO$UnsafeMemoryIO64 (file:/Users/headius/work/jruby/lib/jruby.jar)
WARNING: Please consider reporting this to the maintainers of class com.kenai.jffi.UnsafeMemoryIO$UnsafeMemoryIO64
WARNING: sun.misc.Unsafe::putLong will be removed in a future release
```

2. Usage of Unsafe is now warning more noisily:

```
WARNING: A restricted method in java.lang.System has been called
WARNING: java.lang.System::load has been called by com.kenai.jffi.internal.StubLoader in module org.jruby.dist (file:/Users/headius/work/jruby/lib/jruby.jar)
WARNING: Use --enable-native-access=org.jruby.dist to avoid a warning for callers in this module
WARNING: Restricted methods will be blocked in a future release unless native access is enabled

WARNING: A terminally deprecated method in sun.misc.Unsafe has been called
WARNING: sun.misc.Unsafe::putLong has been called by com.kenai.jffi.UnsafeMemoryIO$UnsafeMemoryIO64 (file:/Users/headius/work/jruby/lib/jruby.jar)
WARNING: Please consider reporting this to the maintainers of class com.kenai.jffi.UnsafeMemoryIO$UnsafeMemoryIO64
WARNING: sun.misc.Unsafe::putLong will be removed in a future release
```

These are both in JFFI but I file it here because it is very visible for JRuby users.

See https://github.com/jnr/jffi/issues/165 for part of this (Unsafe stuff).

boris-petrov:
I also see (in addition to the ones mentioned in the first post):

```
WARNING: A terminally deprecated method in sun.misc.Unsafe has been called
[...]

PR #8708, opened 2025-03-20
Refactor launcher AppCDS logic

mrnoname1000:
The launcher now detects the Java version and uses the appropriate AppCDS flags. If I researched correctly, AppCDS is supported on Java 10+ and the AutoCreateSharedArchive flag on 19+. The modification time of the JRuby JAR is used to determine if the AppCDS archive should be explicitly regenerated.

The Java version currently does not factor into the regeneration calculation. The easiest solution would be to have an archive for every Java version, which would only involve sticking the relevant version in the filename, but could get unwieldy without a way to clean up old archives. We could also store the last-used version in a sidecar file, which also wouldn't be too bad since we already need access to the JRuby folder. I could use some input on what strategy we would prefer.

### DIFF
--- a/bin/jruby.sh
+++ b/bin/jruby.sh
@@ -750,13 +750,13 @@ if $use_jsa_file; then
         jruby_jsa_file="$JRUBY_JSA"
     fi
 
+    # Default to no-op script when explicitly generating
     if $regenerate_jsa_file; then
-        # Run no-op script if AppCDS caching was requested
         case $ruby_args in
-            (*[![:space:]]*) append ruby_args -e 1 ;;
+            (*[![:space:]]*) ;;
+            (*) append ruby_args -e 1 ;;
         esac
     fi
-
 
     # Archive should be regenerated manually if requested or it's outdated relative to JRuby
     if ! $regenerate_jsa_file && is_newer "$jruby_jar" "$jruby_jsa_file"; then
--- a/bin/jruby.sh
+++ b/bin/jruby.sh
@@ -145,6 +145,18 @@ a_isempty() {
     return 0
 }
 
+# exists [FILE...]
+#
+# Returns 0 if all FILEs exist or none provided, otherwise returns 1
+exists() {
+    while [ "$#" -gt 0 ]; do
+        [ -e "$1" ] || return
+        shift
+    done
+
+    return 0
+}
+
 # is_newer FILE OTHER...
 #
 # Returns 0 if FILE is newer than all OTHER files. If FILE doesn't exist,
@@ -500,7 +512,7 @@ add_log "Detected Java version: $java_version"
 java_major=${java_version%%.*}
 
 # AppCDS support
-if [ "$java_major" -ge 13 ]; then
+if [ "$java_major" -ge 13 ] && exists "$JAVA_HOME"/lib/server/*.jsa; then
     java_has_appcds=true
 else
     java_has_appcds=false
--- a/bin/jruby.sh
+++ b/bin/jruby.sh
@@ -477,10 +477,8 @@ java_is_modular() {
 }
 
 if java_is_modular; then
-    use_jsa_file=true
     use_modules=true
 else
-    use_jsa_file=false
     use_modules=false
 fi
 readonly use_modules
@@ -507,6 +505,9 @@ else
     java_has_appcds=false
 fi
 readonly java_has_appcds
+
+# Default to using AppCDS if available
+use_jsa_file="$java_has_appcds"
 
 # AppCDS autogeneration
 if [ "$java_major" -ge 19 ]; then
--- a/bin/jruby.sh
+++ b/bin/jruby.sh
@@ -706,7 +706,7 @@ do
         --restore) append java_args -XX:CRaCRestoreFrom=.jruby.checkpoint ;;
         --cache)
             if $java_has_appcds; then
-                use_jsa_file=true
+                #use_jsa_file=true  # Currently this defaults to true
                 regenerate_jsa_file=true
             else
                 echo "Warning: Java $java_major doesn't support automatic AppCDS, ignoring --cache" >&2
--- a/bin/jruby.sh
+++ b/bin/jruby.sh
@@ -135,6 +135,19 @@ preextend() {
     eval "$1=\"\${$2} \${$1}\""
 }
 
+# is_newer FILE OTHER...
+#
+# Returns 0 if FILE is newer than all OTHER files. If FILE doesn't exist,
+# return error. If OTHER files don't exist, pretend they're older than FILE.
+is_newer() {
+    local output master
+    master="$1"
+    shift
+
+    # Find any other files that are newer, negate outside find in case any don't exist
+    [ -e "$master" ] && ! find "$@" -newer "$master" 2>/dev/null | read -r _
+}
+
 # echo [STRING...]
 #
 # Dumb echo, i.e. print arguments joined by spaces with no further processing
--- a/bin/jruby.sh
+++ b/bin/jruby.sh
@@ -703,34 +703,6 @@ JFFI_OPTS="-Djffi.boot.library.path=$JRUBY_HOME/lib/jni"
 
 CLASSPATH="${CP-}${CP_DELIMITER}${CLASSPATH-}"
 
-# ----- Tweak console environment for cygwin ----------------------------------
-
-if $cygwin; then
-    use_exec=false
-    JRUBY_HOME="$(cygpath --mixed "$JRUBY_HOME")"
-    JRUBY_SHELL="$(cygpath --mixed "$JRUBY_SHELL")"
-
-    eval set -- "$ruby_args"
-
-    case $1 in
-        /*)
-            if [ -f "$1" ] || [ -d "$1" ]; then
-                # replace first element of ruby_args with cygwin form
-                win_arg="$(cygpath -w "$1")"
-                shift
-                set -- "$win_arg" "$@"
-                assign ruby_args "$@"
-            fi
-            ;;
-    esac
-
-    # fix JLine to use UnixTerminal
-    if stty -icanon min 1 -echo > /dev/null 2>&1; then
-        JAVA_OPTS="$JAVA_OPTS -Djline.terminal=jline.UnixTerminal"
-    fi
-
-fi
-
 # ----- Module and Class Data Sharing flags for Java 9+ -----------------------
 
 if $use_modules; then
@@ -757,6 +729,34 @@ if $use_modules; then
             append java_args -Xlog:cds=info:file="$JRUBY_JSA".log -Xlog:cds+dynamic=info:file="$JRUBY_JSA".log
         fi
     fi
+fi
+
+# ----- Tweak console environment for cygwin ----------------------------------
+
+if $cygwin; then
+    use_exec=false
+    JRUBY_HOME="$(cygpath --mixed "$JRUBY_HOME")"
+    JRUBY_SHELL="$(cygpath --mixed "$JRUBY_SHELL")"
+
+    eval set -- "$ruby_args"
+
+    case $1 in
+        /*)
+            if [ -f "$1" ] || [ -d "$1" ]; then
+                # replace first element of ruby_args with cygwin form
+                win_arg="$(cygpath -w "$1")"
+                shift
+                set -- "$win_arg" "$@"
+                assign ruby_args "$@"
+            fi
+            ;;
+    esac
+
+    # fix JLine to use UnixTerminal
+    if stty -icanon min 1 -echo > /dev/null 2>&1; then
+        JAVA_OPTS="$JAVA_OPTS -Djline.terminal=jline.UnixTerminal"
+    fi
+
 fi
 
 # ----- Final prepration of the Java command line -----------------------------
--- a/bin/jruby.sh
+++ b/bin/jruby.sh
@@ -744,7 +744,6 @@ readonly jruby_jsa_file="$JRUBY_HOME/lib/jruby-java$java_version.jsa"
 assign jruby_jsa_files "$JRUBY_HOME"/lib/jruby-java*.jsa
 readonly jruby_jsa_files
 
-
 if $use_jsa_file; then
     # Allow overriding default JSA file location
     if [ -n "${JRUBY_JSA-}" ]; then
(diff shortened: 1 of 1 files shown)


headius:
Very nice! High priority to get this landed for the upcoming 10 release. I'll review.

I chatted with @enebo as well and we agree that it would be good to include the Java version in the JSA filename, so every Java version gets its own file. Same would go for the log output when that is enabled. On my system, JRuby 10 + Java 21 JSA file is only 5.4MB so I don't think it's a big deal to leave a few lying around.

We also chatted about **where** we should store these files. I think it's fine to go with JRuby's `lib` dir for now but we may want to change that to the `bin` dir or potentially even a user location like `~/.jruby`.

headius:
@mrnoname1000 You still have this marked as draft... is there more work coming?

mrnoname1000:
It's still a draft because I haven't implemented version in the JSA filename yet, but I'm home from work now so I can add that last feature, This PR should be working though if you want the existing work merged sooner.

headius:
Go ahead and add that and then I'll review.

mrnoname1000:
I added the version string to the JSA filename as well as a flag to remove all the JSA files we may have generated. If we find out JSA files have compatibility guarantees between point releases, we can trim off parts of the version string to allow sharing between those releases.

bin/jruby.sh
```diff
@@ -689,20 +721,87 @@ fi
...
+if $use_jsa_file; then
+    # Allow overriding default JSA file location
+    if [ -n "${JRUBY_JSA-}" ]; then
+        jruby_jsa_file="$JRUBY_JSA"
+    fi
+
+    if $regenerate_jsa_file; then
+        # Run no-op script if AppCDS caching was requested
+        case $ruby_args in
+            (*[![:space:]]*) append ruby_args -e 1 ;;
```
headius, line 757: That's some serious shell magic there!

headius: Hmm perhaps it's not working? Running `jruby --cache` just hangs waiting for input now.

bin/jruby.sh
```diff
@@ -805,6 +876,15 @@ if $print_environment_log; then
...
 fi
 
+# ----- Perform final mutations after logging ---------------------------------
+if $remove_jsa_files; then
+    # Delete All AppCDS files if requested
+    eval rm -f -- "$jruby_jsa_files"
+elif $regenerate_jsa_file && [ -e "$jruby_jsa_file" ]; then
+    # Delete AppCDS file if requested or if it's outdated
+    rm -f -- "$jruby_jsa_file"
+fi
```
headius, line 886: I'm not sure this section is doing what we want:

* My thought was that --rmcache would run nothing but clear those files. As it is now if you don't pass a Ruby script or `-e` it will wait on stdin for code.
* Doesn't this `$regenerate_jsa_file` line cause it to immediately wipe out the archive it just regenerated? I thought it would but can't reproduce thiat with `jruby --cache -e 1`

headius (changes requested):
There's some minor issues and some regressed behavior (`--cache`).

bin/jruby.sh
mrnoname1000: Oops I badly moved that from somewhere else in the code, fixing

mrnoname1000: The shell magic isn't doing much but I agree it's obtuse, I put it in its own function

bin/jruby.sh
mrnoname1000: - Oh yeah it doesn't make sense to proceed there if you can already use the `--cache` option
- No, the launcher doesn't touch the archive except for here. That's why I'm checking if it exists first, because otherwise we don't need to delete it and it'll get generated by Java

headius:
We have some command line tests under test/jruby that could be expanded for this new functionality.

headius:
I'll retest locally with latest changes. 👍

headius:
So far everything is working well on JRuby 10!

A wild question appears: should `--rmcache` delete logs too? Doesn't have to be answered for this one and I don't have a strong opinion one way or another.

headius:
Issue with 9.4 under Java 11:

```
$ jruby --logcache -e 1
Unrecognized VM option 'ArchiveClassesAtExit=/Users/headius/work/jruby94/lib/jruby-java11.0.17.jsa'
Error: Could not create the Java Virtual Machine.
Error: A fatal exception has occurred. Program will exit.
```

I'm not sure this particular flag existed until after 11. Earlier CDS support required a couple flags and in some cases, multiple steps.

I'm thinking maybe we don't include the automatic CDS feature in the launcher unless we're on a Java that supports one-shot "at exit" dumping. There's still the `jruby-startup` gem and its `generate-appcds` that can handle the old style flags.

headius:
See here for the Java 11 equivalent, which does appear to require a two step process (dump class list, then generate CDS archive): https://docs.oracle.com/en/java/javase/11/tools/java.html#GUID-31503FCE-93D0-4175-9B4F-F6A738B2F4C4

This page documents the feature for Java 10, but it was still a "commercial" feature at that point: https://docs.oracle.com/javase/10/tools/java.htm#JSWOR-GUID-31503FCE-93D0-4175-9B4F-F6A738B2F4C4

I believe Java 13 is the first release to support the "one shot" `ArchiveClassesAtExit` flag: https://docs.oracle.com/en/java/javase/13/docs/specs/man/java.html#application-class-data-sharing

And you appear to be right that 19 introduces the automatic flag: https://docs.oracle.com/en/java/javase/19/docs/specs/man/java.html#creating-dynamic-cds-archive-file-with--xxautocreatesharedarchive

So I say we should disable the automatic dumping as well as the `--cache` flag when running Java 12 and earlier. It's just too much hassle right nowto support the feature with the two-step dance. 

@mrnoname1000 I can make this change after merging, or you can make it if you want. I'm keen to get this landed so I can do some other version-specific flags (e.g. for #8696).

headius:
Hmm, my change doesn't seem to be enough. It's still trying to pass the Java 13+ flag:

```
[] jruby94 $ jruby -e 1
Unrecognized VM option 'ArchiveClassesAtExit=/Users/headius/work/jruby94/lib/jruby-java11.0.17.jsa'
Error: Could not create the Java Virtual Machine.
Error: A fatal exception has occurred. Program will exit.
[] jruby94 $ jruby --environment -e 1
JRuby Environment
=================

JRuby executable:
  /Users/headius/work/jruby94/bin/jruby
JRuby command line options:
  --environment -e 1
Current directory:
  /Users/headius/work/jruby94

Environment:
  JRUBY_HOME: /Users/headius/work/jruby94
  JRUBY_OPTS: 
  JAVA_OPTS: 
  JAVACMD: /Library/Java/JavaVirtualMachines/temurin-11.jdk/Contents/Home/bin/java
  JAVA_HOME: /Library/Java/JavaVirtualMachines/temurin-11.jdk/Contents/Home

Detected Java modules at /Library/Java/JavaVirtualMachines/temurin-11.jdk/Contents/Home

Adding Java options from: /Users/headius/work/jruby94/bin/.jruby.java_opts

Adding Java options from: /Users/headius/work/jruby94/bin/.jruby.module_opts

Regenerating CDS archive at:
  /Users/headius/work/jruby94/lib/jruby-java11.0.17.jsa

Java command line:
  /Library/Java/JavaVirtualMachines/temurin-11.jdk/Contents/Home/bin/java @/Users/headius/work/jruby94/bin/.jruby.java_opts @/Users/headius/work/jruby94/bin/.jruby.module_opts -Xss2048k -Djffi.boot.library.path=/Users/headius/work/jruby94/lib/jni -Djava.security.egd=file:/dev/urandom -XX:ArchiveClassesAtExit=/Users/headius/work/jruby94/lib/jruby-java11.0.17.jsa -Xlog:cds=off -Xlog:cds+dynamic=off --module-path /Users/headius/work/jruby94/lib/jruby.jar -classpath : -Djruby.home=/Users/headius/work/jruby94 -Djruby.lib=/Users/headius/work/jruby94/lib -Djruby.script=jruby -Djruby.shell=/bin/sh org.jruby.Main -e 1
```

mrnoname1000:
Yeah definitely don't want to do that two step dance. The change you made should have bumped the required version up to 13, I can investigate in a few minutes here and I think we should add some version into to the debug logs.

mrnoname1000:
The AppCDS logic was linked with the modular Java logic, so any modular Java was coming up as wanting AppCDS enabled. I changed it to default to true/false depending on if it's supported, then allowing reassignment later.

mrnoname1000:
> We have some command line tests under test/jruby that could be expanded for this new functionality.

How would we test the new functionality that's highly dependent on the current version of Java?

headius:
A new unfortunate wrinkle: Eclipse Temurin builds of OpenJDK on macos on Apple Silicon do not include the base JDK CDS dump, so you get cryptic errors: https://github.com/adoptium/adoptium-support/issues/937

```
[] jruby94 $ pickjdk 17
/Library/Java/JavaVirtualMachines/temurin-17.jdk/Contents/Home
[] jruby94 $ jruby --logcache -e 1
Error occurred during initialization of VM
DynamicDumpSharedSpaces is unsupported when base CDS archive is not loaded
[] jruby94 $ pickjdk 18
/Library/Java/JavaVirtualMachines/temurin-18.jdk/Contents/Home
[] jruby94 $ jruby -e 1              
Error occurred during initialization of VM
-XX:ArchiveClassesAtExit is unsupported when base CDS archive is not loaded. Run with -Xlog:cds for more info.
[] jruby94 $ pickjdk 19
/Library/Java/JavaVirtualMachines/temurin-19.jdk/Contents/Home
[] jruby94 $ jruby -e 1
OpenJDK 64-Bit Server VM warning: -XX:ArchiveClassesAtExit is unsupported when base CDS archive is not loaded. Run with -Xlog:cds for more info.
[] jruby94 $ pickjdk 20
/Library/Java/JavaVirtualMachines/temurin-20.jdk/Contents/Home
[] jruby94 $ jruby -e 1
OpenJDK 64-Bit Server VM warning: -XX:ArchiveClassesAtExit is unsupported when base CDS archive is not loaded. Run with -Xlog:cds for more info.
```

Switching to a Zulu build works correctly with all versions I tried. I expressed my discontent on the Adoptium issue above... hopefully they will have an answer or we'll need to also detect a CDS file in the JDK before starting. 😡

headius:
On my Zulu 21 install the built-in jsa is here:

`/Library/Java/JavaVirtualMachines/zulu-21.jdk/Contents/Home/lib/server/classes.jsa`

(`.../Contents/Home` is `$JAVA_HOME`)

I suppose we could just check if there's any .jsa file in the `$JAVA_HOME/lib/server` dir? What a pain.

mrnoname1000:
> I suppose we could just check if there's any .jsa file in the $JAVA_HOME/lib/server dir? What a pain.

All we need to do is `find "$JAVA_HOME" -iname "*.jsa" | read -r _` but yeah recursing through JAVA_HOME would be a bit of an expensive check. Looks like it completely errors if the base CDS files aren't present, what do the `release` files look like for those builds?

mrnoname1000:
This is what the patch would look like, I can push it if you want:

```diff
diff --git a/bin/jruby.sh b/bin/jruby.sh
index 8cd0aec5e3..66e080bac0 100755
--- a/bin/jruby.sh
+++ b/bin/jruby.sh
@@ -145,6 +145,18 @@ a_isempty() {
     return 0
 }
 
+# exists [FILE...]
+#
+# Returns 0 if all FILEs exist or none provided, otherwise returns 1
+exists() {
+    while [ "$#" -gt 0 ]; do
+        [ -e "$1" ] || return
+        shift
+    done
+
+    return 0
+}
+
 # is_newer FILE OTHER...
 #
 # Returns 0 if FILE is newer than all OTHER files. If FILE doesn't exist,
@@ -500,7 +512,7 @@ add_log "Detected Java version: $java_version"
 java_major=${java_version%%.*}
 
 # AppCDS support
-if [ "$java_major" -ge 13 ]; then
+if [ "$java_major" -ge 13 ] && exists "$JAVA_HOME"/lib/server/*.jsa; then
     java_has_appcds=true
 else
     java_has_appcds=false
```

headius:
Let's hold off and do this additional detection separately. I heard back from the Temurin folks and they say builds since January should include the missing files. Maybe this problem will just go away, but if it doesn't we can add the detection in.

mrnoname1000:
My worry is just that JRuby will suddenly break for anyone with these JDKs unless they pass the as-yet undocumented `--nocache` flag. How does the inclusion of the base JSA file work? I assume it's just a build flag, so anyone who wanted a faster build could exclude that step.

headius:
@mrnoname1000 That's a fair point. Let's go with the patch you have and we can refine it after we land this PR.

headius:
Also, let's make `--cache` just exit with error when CDS is not easily available.

mrnoname1000:
Added the discussed features:
- If no `*.jsa` files exist in `$JAVA_HOME/lib/server`, AppCDS is disabled implicitly
- `--cache` exits if used on an unsupported JVM
- If the AppCDS archive exists and is not writable, or if it doesn't exist and its parent directory is not writable, a warning is printed and AppCDS is disabled

mrnoname1000:
Actually the JSA file is created without write permissions for some reason so I changed it to just check the parent directory

headius:
> Actually the JSA file is created without write permissions for some reason so I changed it to just check the parent directory

Yeah I think this is a holdover from the JDK-level archive that they don't want anyone to overwrite, but it's definitely a little weird.

headius:
I've been running with this latest version since you pushed it and everything works great on Java 21 and 23 with JRuby master. I'm going to merge this and we can make additional tweaks there.

Excellent work, @mrnoname1000, thank you! I see a few next steps:

* Tests!
* Document new flags and behavior. This means `--help` output and probably a good wiki page with details and links to CDS/Leyden info.
* Trying this out with JRuby 9.4 on Java 8-12 and seeing how it behaves. Ideally we merge it back and 9.4 users will get some benefit from CDS if they use the right JDKs.

I also still like the idea of making this a separate project (module, repository, whatever) that gets installed on build. If it's a gem, it would mean users can upgrade the launcher any time. But either way having it sourced from one place would mean we're always including the latest cool launcher script features.

And of course we still need to explore what it might take to build a native executable with this and busybox, or else help @enebo finish his Rust launcher.

PR #8718, opened 2025-03-26
Apply Java version-specific flags in launcher script

headius:
This adds flags for native access and unsafe memory functions and cleans up some of the launcher logic around Java version detection.

Relates to #8696.

### DIFF
--- a/bin/jruby.sh
+++ b/bin/jruby.sh
@@ -530,6 +530,22 @@ else
 fi
 readonly java_has_appcds_autogenerate
 
+# Native access
+if [ "$java_major" -ge 22 ]; then
+    enable_native_access=true
+else
+    enable_native_access=false
+fi
+readonly enable_native_access
+
+# Unsafe memory access
+if [ "$java_major" -ge 23 ]; then
+    enable_unsafe_memory=true
+else
+    enable_unsafe_memory=false
+fi
+readonly enable_unsafe_memory
+
 # ----- Process .java_opts files ----------------------------------------------
 
 # We include options on the java command line in the following order:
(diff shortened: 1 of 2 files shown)


bin/jruby.sh
```diff
@@ -511,25 +511,52 @@ add_log "Detected Java version: $java_version"
...
+
+    if [ "$java_major" -ge 19 ]; then
+        # AppCDS autogeneration
+        java_has_appcds_autogenerate=true
+
+        if [ " $java_major" -ge 22 ]; then
+            # Native access
+            enable_native_access=true
+
+            if [ " $java_major" -ge 23 ]; then
```
mrnoname1000, line 531: You've got spaces in your quoted variables

mrnoname1000 (commented):
I would avoid nesting so many conditionals, it's much more verbose and comparisons are cheap. We could use some other structure if we wanted to consolidate, but no matter what every check is going to happen on >=23 or <13 depending on which we check first

bin/jruby.sh
```diff
@@ -832,6 +848,18 @@ if $use_jsa_file; then
...
+if $enable_native_access; then
+    append java_args --enable-native-access=org.jruby.dist
+fi
+
+# Enable access to Unsafe memory functions
+if $enable_unsafe_memory; then
+    append java_args --sun-misc-unsafe-memory-access=allow
+fi
+
+# Enable access to Unsafe memory functions
```
mrnoname1000, line 872: Duplicate comment

headius:
Perhaps it's just better to drop that commit then. It's not critical, and I don't like it as much as I thought.

Your point about doing all the checks anyway depending on how we search also makes sense.

mrnoname1000 (approved):
Aside from that duplicate comment looks good to me

bin/jruby.sh
headius: Fixed

PR #8731, opened 2025-03-31
[ji] avoid false duplicate method warning

kares:
The test case is mirrored after a real-world Map impl from Guava:

```ruby
map = com.google.common.cache.CacheBuilder.new_builder.build.as_map 
# map is an instance of: Java::ComGoogleCommonCache::LocalCache
map.compute(:foo) { |_, value| 42 }
```
leading to a false-positive warning:
```
(RuntimeError) multiple Java methods found, dumping backtrace and choosing compute(java.lang.Object,java.util.function.BiFunction)
```
upon inspection of the `CallableSelector` there were 2 methods:
- `public default java.lang.Object java.util.concurrent.ConcurrentMap.compute(java.lang.Object,java.util.function.BiFunction)` 
- `public default java.lang.Object java.util.Map.compute(java.lang.Object,java.util.function.BiFunction)]`

### DIFF
--- a/core/src/main/java/org/jruby/javasupport/binding/MethodGatherer.java
+++ b/core/src/main/java/org/jruby/javasupport/binding/MethodGatherer.java
@@ -215,6 +215,11 @@ public class MethodGatherer {
     }
 
     private static boolean methodsAreEquivalent(Method child, Method parent) {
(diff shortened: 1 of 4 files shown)


headius:
Nice one!

headius (approved):

Issue #8728, opened 2025-03-29
Incorrect warning: given block not used


notEthan:
Working with a subclass of Set, JRuby throws a warning from #initialize - minimal reproduction:

```ruby
require 'set'
class ASet < Set
  def initialize(&block)
    super()
    block[]
  end
end
ASet.new { }
```

`test.rb:4: warning: given block not used`. Doesn't seem to happen with a different superclass or method other than #initialize. 

**Environment Information**
  
* jruby 9.4.12.0
* macos & ubuntu

headius:
I'm guessing it's using the new method to detect this rather than the initialize method. Thanks for the report!

headius:
The IR appears to be getting compiled such that the incoming block gets passed on in the super:

```
block #2 (out: exit): LBL_1:-1
  0: %v_3 := load_implicit_closure
  1: %v_0 := recv_kw(hasRestArg: false, acceptsKeywords: false)
  2:         check_arity(%v_0, required: 0, opt: 0, rest: false, restKey: -1)
  3:         line_num(lineNumber: 0, coverage: false, oneshot: false)
  4: %v_2 := instance_super(mod<0>, %v_3, isLiteralBlock: false, callType: SUPER, name: initialize, potentiallyRefined: false, flags: 0)
  5:         return(%v_2)
```

It does this even when we don't capture the block (`def initialize`):

```
block #2 (out: exit): LBL_1:-1
  0: %v_2 := load_implicit_closure
  1: %v_0 := recv_kw(hasRestArg: false, acceptsKeywords: false)
  2:         check_arity(%v_0, required: 0, opt: 0, rest: false, restKey: -1)
  3:         line_num(lineNumber: 0, coverage: false, oneshot: false)
  4: %v_1 := instance_super(mod<0>, %v_2, isLiteralBlock: false, callType: SUPER, name: initialize, potentiallyRefined: false, flags: 0)
  5:         return(%v_1)
```

headius:
Fix is in #8732 and will be in JRuby 9.4.13 and JRuby 10.0.

headius:
Ok I was wrong, it's actually specified behavior to pass the block through implicitly even if it is not explicitly passed to a non-zsuper super.

You can work around the extra warning by explicitly passing a nil block like `super(&nil)`.

Unfortunately #8732 is not the right fix. I'm trying to figure out what CRuby does for such cases.

Orthogonal to this... in the default `set` library, `Set#initialize` does consume the block:

[...]

PR #8732, opened 2025-03-31
Don't attempt to forward block to non-zsuper

headius:
Not sure how this has not been noticed up to now, but we should not be forwarding the passed-in block to super when it is a normal non-z super. This patch removes that logic and only passes along any block that was directly provided to the super call.

Fixes jruby/jruby#8728

This should be fixed for 9.4.13.0 but JRuby 10 will be released first... so I'm not sure how best to mark this milestone.


headius:
Ok I was wrong, it's actually specified behavior to pass the block through implicitly even if it is not explicitly passed to a non-zsuper super.

PR #8751, opened 2025-04-07
Update URI ge mto 1.0.3

andsel:


lib/pom.xml
```diff
@@ -1254,7 +1254,7 @@ DO NOT MODIFY - GENERATED CODE
           <include>gems/timeout-0.4.3*/**/*</include>
           <include>gems/tsort-0.2.0*/**/*</include>
           <include>gems/un-0.3.0*/**/*</include>
-          <include>gems/uri-1.0.2*/**/*</include>
+          <include>gems/uri-1.0.3*/**/*</include>
```
kaisecheng, line 1257: The header of pom.xml says "DO NOT MODIFY - GENERATED CODE". I think this file is not supposed to update.

andsel: Thanks @kaisecheng , fixed

headius:
Actually we do commit changes to that file but it should not be edited *by hand*. It gets regenerated by the build, so you make a change to the pom.rb, run one build, and commit both files.

Thanks for the PR! A few things for you to consider:

* JRuby master is JRuby 10. If this change is only intended for JRuby 10, that's fine... otherwise it should be retargeted against the `jruby-9.4` branch.
* Master should be green now, modulo some intermittent failures we are still catching in CI. If you rebase, it should provide a better view of pass/fail.
* You can go ahead and restore the changes to the pom.xml.

andsel:
Thanks @headius for checking. This PR is targeted for JRuby 10, while the one for 9.4 branch is #8755

headius:
Thank you!

PR #8753, opened 2025-04-08
Java version detection improvements

headius:
We want to be able to use this script with older versions of JRuby that will have different minimum Java versions, so use a new file bin/.java-version as the minimum.

Once this is approved we should be able to backport the script to JRuby 9.4 and know that it won't cause problems if it is copied or backported manually to versions older than that.


headius:
A few additional tweaks:

* Check for existence of `$JAVA_HOME/release` file.
* Detect "1.8" and treat it as "8".

bin/jruby.sh
```diff
@@ -505,7 +505,21 @@ fi
...
+    java_version=$(. "$JAVA_HOME/release" && echo "${JAVA_VERSION-}")
+    if echo "${java_version}" | grep -q "1.8.\\.*"; then
+        java_version=8
+    fi
+else
+    java_version=8
+fi
+# shellcheck source=/dev/null
+if [ -f "$JRUBY_HOME/bin/.java-version" ]; then
+    minimum_java_version=$(. "$JRUBY_HOME/bin/.java-version" && echo "${JRUBY_MINIMUM_JAVA_VERSION-}")
```
mrnoname1000, line 524: I think we can just source this file directly since it's from a trusted source
```sh
if [ -f "$JRUBY_HOME/bin/.java-version" ] && . "$JRUBY_HOME/bin/.java-version" && [ "${JRUBY_MINIMUM_JAVA_VERSION-}" ]; then
    minimum_java_version=$JRUBY_MINIMUM_JAVA_VERSION
else
    # Only 9.4.12.0 and earlier will have shipped without a .java-version file, so fall back on minimum of 8
    minimum_java_version=8
fi
```

bin/jruby.sh
```diff
@@ -505,11 +505,37 @@ fi
...
 # ----- Detect Java version and determine available features ------------------
 # shellcheck source=/dev/null
-java_version=$(. "$JAVA_HOME/release" && echo "${JAVA_VERSION-}")
+if [ -f "$JAVA_HOME/release" ]; then
+    java_version=$(. "$JAVA_HOME/release" && echo "${JAVA_VERSION-}")
+
+    # treat 1.8 as 8
+    case $java_version in
+        1.8 | 1.8.*) java_major=8 ;;
+        *)  # Split version out for integer comparisons
```
mrnoname1000, line 514: Totally personal preference but I would put this comment above the case statement to group related code closer together

bin/jruby.sh
mrnoname1000: We could also stick this filename in a variable somewhere to avoid repeating ourselves but not a big deal

bin/jruby.sh
headius: Fixed this in a subsequent change, one comment line above the case now.

bin/jruby.sh
headius: Fixed in subsequent commit.

mrnoname1000:
LGTM

PR #8755, opened 2025-04-08
Update uri to version 0.12.3 in JRuby 9.4

andsel:


headius:
Looks fine, thank you!

Issue #6913, opened 2021-10-27
Reimplement sprintf


enebo:
Working on bugs involving sprintf are difficult.  We have a massive switch statement and a lot of complicated state toggling.  We also have an implementation where parsing is done at the same time as outputting the result string.  If parsing was in a separate phase we could save the parsing result to a cache (potentially even callsite caching it).  We could also consider bytecode compiling it.

This is a significant addition to JRuby 9.4.  It will minimally cache parsed format strings (if frozen) and we will add support for %a/%A.  We also plan on fixing all tagged specs and MRI test suite failures.

enebo:
new_sprintf has integer printf largely working (and working quite a bit more compatibly than our current implementation).  floating point formatting is not implemented but there is a start on that (based on paper: https://www.cs.tufts.edu/~nr/cs257/archive/florian-loitsch/printf.pdf) using grisu2.

The plan on deploying this will show up in a point release and we will have an option which allows us to use legacy support if any problems arise.  The next major JRuby version will remove the legacy version.

PR #8769, opened 2025-04-15
Fix `sprinf` with explicit zero precision and zero value

Earlopain:
In that case, nothing is written. It's documented: https://docs.ruby-lang.org/en/3.4/format_specifications_rdoc.html#:~:text=if%20the%20precision%20is%200%20and%20the%20value%20is%200%2C%20nothing%20is%20written

NOTE: I found this code quite hard to follow and there was at least one other place I contemplated implementing a fix. Not sure if what I chose is in fact the best place for a fix.

### DIFF
--- a/core/src/main/java/org/jruby/util/Sprintf.java
+++ b/core/src/main/java/org/jruby/util/Sprintf.java
@@ -874,7 +874,16 @@ public class Sprintf {
                             if (leadChar != 0) len++;
                         }
                     }
+
                     int numlen = bytes.length - first;
+                    switch (fchar) {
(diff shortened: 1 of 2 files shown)


enebo:
I have an ancient branch (#6913) which rewrites much of this but I got stuck on dtoa impl.  We know this is horrifically messy and your fix looks fine even if it "maybe" could be somewhere else.  As it stands this fixes the issue with new specs.  LGTM.  Also you may be passing some more tests in test/mri/ruby/test_sprintf_comb.rb (another fun piece of code).

PR #8774, opened 2025-04-17
Run scheduled workflows on main repository only

Earlopain:
They currently also run on forks, where they will fail during publishing


headius:
This seems like a legitimate change and I'll do it myself, but why did you close it?

headius:
46678ce513

Earlopain:
I should have bumped this instead, sorry.

PR #8792, opened 2025-04-22
Update jruby-openssl to 0.15.4.pre1 for testing

headius:


headius:
@kares This line seems like the most likely source of the null, but I'm not sure what would cause this to return a null public key.

https://github.com/jruby/jruby-openssl/blob/f176706dc0a4b4865c28d8b09c41eb889c8e4577/src/main/java/org/jruby/ext/openssl/PKey.java#L140

You are right, this does not appear to be directly related to Java 21+ but I don't think we can rule that out. There shouldn't be any changes in JRuby 10 that would cause this line to start returning a null public key. It could be something elsewhere, though.

For the purposes of this immediate jruby-openssl update, however, I don't think it should be considered blocking.

headius:
No longer needed.

PR #8802, opened 2025-04-28
Fixes to improve DRb tests

headius:
* Fix missing class error during path search from Marshal#load.


headius:
See also #8804 which fixes a DRb error expectation by registering RubyNoMethodError.args as an internal variable of NoMethodError.

headius:
Only one fix made it into this PR based on @kou's comments from ruby/drb#36 but we can go ahead with it.

PR #8804, opened 2025-04-28
Allow binding instance/internal vars to fields

headius:
This is an extension of the existing mechanism for reification of Ruby classes, allowing us to use normal Java fields for either instance or internal variables.

The logic here is largely unchanged from the original reification logic except for enhancements to allow actively requesting a specific binding:

* Allow specifying a Ruby name to be used in the variable table, so that both instance vars (@var) and internal vars (no @) can be tied to a specific field.
* Allow the direct field configuration to aggregate a Lookup object so that private fields can bound.

The example usage here binds the "args" internal variable on NoMethodError to the "args" field on RubyNoMethodError, allowing it to be seen by marshal dumping and fixing one of the DRb issues in ruby/drb#36. The reification change was not strictly necessary to fix that issue, but the alternative converts it into a slow-path internal variable lookup which is less than ideal.

Assuming this change is safe, all other instance and internal variables used by concrete JRuby core classes can be converted with the following process:

* Set the reified class (the real concrete class) at definition time.
* Request the direct field with a Java name and either the @var or bare var name, providing an appropriate Lookup object if the field shoud be private.

Enhancements to this feature could include:

* Allow binding read-only variables. Currently both a read and write path are expected.
* Enhance conversion logic to allow variables to be a more natural Java type. The logic for this exists but has not been heavily tested. See Java logic for treating fields as instance variables.

### DIFF
--- a/core/src/main/java/org/jruby/RubyClass.java
+++ b/core/src/main/java/org/jruby/RubyClass.java
@@ -54,7 +54,9 @@ import static org.objectweb.asm.Opcodes.ACC_SYNTHETIC;
 import static org.objectweb.asm.Opcodes.ACC_VARARGS;
 
 import java.io.IOException;
+import java.lang.invoke.MethodHandles;
 import java.lang.reflect.Constructor;
+import java.lang.reflect.Field;
 import java.lang.reflect.InvocationTargetException;
 import java.lang.reflect.Method;
 import java.lang.reflect.Modifier;
@@ -65,6 +67,7 @@ import java.util.function.BiConsumer;
 import java.util.stream.Collectors;
 
 import org.jruby.anno.JRubyClass;
+import org.jruby.anno.JRubyField;
 import org.jruby.anno.JRubyMethod;
 import org.jruby.api.JRubyAPI;
 import org.jruby.compiler.impl.SkinnyMethodAdapter;
@@ -2649,6 +2652,46 @@ public class RubyClass extends RubyModule {
         }
         while ( current != null );
         return null;
+    }
+
+    @Override
+    public RubyClass defineMethods(ThreadContext context, Class... methodSources) {
+        return super.defineMethods(context, methodSources);
+    }
+
+    /**
+     * Define instance and internal variables based on Java fields annotated with {@link JRubyField}.
+     *
+     * @param context
+     * @param constantSource class containing the field annotations
+     * @return itself for composable API
+     */
+    @JRubyAPI
+    public <T extends RubyModule> T defineFields(ThreadContext context, Class constantSource, MethodHandles.Lookup lookup) {
+        for (Field field : constantSource.getDeclaredFields()) {
+            if (Modifier.isStatic(field.getModifiers())) continue;
+            defineAnnotatedField(context, field, lookup);
+        }
+        return (T) this;
+    }
+
+    private boolean defineAnnotatedField(ThreadContext context, Field field, MethodHandles.Lookup lookup) {
+        JRubyField jrubyField = field.getAnnotation(JRubyField.class);
+        if (jrubyField == null) return false;
+
+        Class tp = field.getType();
+        Class reifiedType = getReifiedRubyClass();
+        if (reifiedType == null) {
+            throw new RuntimeException("Class does not have a reified form: " + this);
+        }
+
+        String javaName = field.getName();
+        String rubyName = jrubyField.value();
+        if (rubyName.isBlank()) rubyName = javaName;
+
+        getVariableTableManager().addDirectField(javaName, rubyName, tp, false, null, lookup);
+
+        return true;
     }
 
     public Map<String, List<Map<Class<?>, Map<String,Object>>>> getParameterAnnotations() {
--- a/core/src/main/java/org/jruby/RubyModule.java
+++ b/core/src/main/java/org/jruby/RubyModule.java
@@ -68,6 +68,7 @@ import org.jruby.anno.AnnotationHelper;
 import org.jruby.anno.FrameField;
 import org.jruby.anno.JRubyClass;
 import org.jruby.anno.JRubyConstant;
+import org.jruby.anno.JRubyField;
 import org.jruby.anno.JRubyMethod;
 import org.jruby.anno.JavaMethodDescriptor;
 import org.jruby.anno.TypePopulator;
--- a/core/src/main/java/org/jruby/RubyNoMethodError.java
+++ b/core/src/main/java/org/jruby/RubyNoMethodError.java
@@ -26,6 +26,7 @@
 
 package org.jruby;
 
+import org.jruby.anno.JRubyField;
 import org.jruby.anno.JRubyMethod;
 import org.jruby.anno.JRubyClass;
 import org.jruby.exceptions.NoMethodError;
@@ -47,15 +48,17 @@ import static org.jruby.api.Define.defineClass;
  */
 @JRubyClass(name="NoMethodError", parent="NameError")
 public class RubyNoMethodError extends RubyNameError implements Reified {
+    @JRubyField
     private IRubyObject args;
 
     private static final ObjectAllocator ALLOCATOR = RubyNoMethodError::new;
 
     static RubyClass define(ThreadContext context, RubyClass NameError) {
         RubyClass noMethodError = defineClass(context, "NoMethodError", NameError, ALLOCATOR).
-                defineMethods(context, RubyNoMethodError.class);
-        noMethodError.reifiedClass(RubyNoMethodError.class);
-        noMethodError.getVariableTableManager().requestFieldStorage("args", "args", IRubyObject.class, false, null, MethodHandles.lookup());
+                reifiedClass(RubyNoMethodError.class).
+                defineMethods(context, RubyNoMethodError.class).
+                defineFields(context, RubyNoMethodError.class, MethodHandles.lookup());
+
         return noMethodError;
     }
 
--- a/core/src/main/java/org/jruby/RubyNoMethodError.java
+++ b/core/src/main/java/org/jruby/RubyNoMethodError.java
@@ -30,10 +30,13 @@ import org.jruby.anno.JRubyMethod;
 import org.jruby.anno.JRubyClass;
 import org.jruby.exceptions.NoMethodError;
 import org.jruby.exceptions.RaiseException;
+import org.jruby.java.codegen.Reified;
 import org.jruby.runtime.Block;
 import org.jruby.runtime.ObjectAllocator;
 import org.jruby.runtime.ThreadContext;
 import org.jruby.runtime.builtin.IRubyObject;
+
+import java.lang.invoke.MethodHandles;
 
 import static org.jruby.api.Define.defineClass;
 
@@ -43,14 +46,17 @@ import static org.jruby.api.Define.defineClass;
  * @see NoMethodError
  */
 @JRubyClass(name="NoMethodError", parent="NameError")
-public class RubyNoMethodError extends RubyNameError {
+public class RubyNoMethodError extends RubyNameError implements Reified {
     private IRubyObject args;
 
     private static final ObjectAllocator ALLOCATOR = RubyNoMethodError::new;
 
     static RubyClass define(ThreadContext context, RubyClass NameError) {
-        return defineClass(context, "NoMethodError", NameError, ALLOCATOR).
+        RubyClass noMethodError = defineClass(context, "NoMethodError", NameError, ALLOCATOR).
                 defineMethods(context, RubyNoMethodError.class);
+        noMethodError.reifiedClass(RubyNoMethodError.class);
+        noMethodError.getVariableTableManager().requestFieldStorage("args", "args", IRubyObject.class, false, null, MethodHandles.lookup());
+        return noMethodError;
     }
 
     protected RubyNoMethodError(Ruby runtime, RubyClass exceptionClass) {
--- a/core/src/main/java/org/jruby/anno/JRubyField.java
+++ b/core/src/main/java/org/jruby/anno/JRubyField.java
@@ -1,0 +1,5 @@
+package org.jruby.anno;
+
+public @interface JRubyField {
+    String value() default "";
+}
--- a/core/src/main/java/org/jruby/javasupport/util/JavaClassConfiguration.java
+++ b/core/src/main/java/org/jruby/javasupport/util/JavaClassConfiguration.java
@@ -26,6 +26,9 @@
 
 package org.jruby.javasupport.util;
 
+import org.jruby.specialized.RubyObjectSpecializer;
+
+import java.lang.invoke.MethodHandles;
 import java.util.ArrayList;
 import java.util.Arrays;
 import java.util.Collections;
@@ -129,15 +132,22 @@ public class JavaClassConfiguration implements Cloneable {
     
     public static class DirectFieldConfiguration {
         public String name;
+        public String rubyName;
         public Class<?> fieldType;
         public Boolean unwrap;
         public Class<?> unwrapType;
+        public MethodHandles.Lookup lookup;
         public DirectFieldConfiguration(String name, Class<?> fieldType, Boolean unwrap, Class<?> unwrapType) {
+            this(name, '@' + name, fieldType, unwrap, unwrapType, RubyObjectSpecializer.LOOKUP);
+        }
+
+        public DirectFieldConfiguration(String name, String rubyName, Class<?> fieldType, Boolean unwrap, Class<?> unwrapType, MethodHandles.Lookup lookup) {
             this.name = name;
+            this.rubyName = rubyName;
             this.fieldType = fieldType;
             this.unwrap = unwrap;
             this.unwrapType = unwrapType;
+            this.lookup = lookup;
         }
-        
     }
 }
--- a/core/src/main/java/org/jruby/runtime/ivars/VariableTableManager.java
+++ b/core/src/main/java/org/jruby/runtime/ivars/VariableTableManager.java
@@ -31,6 +31,7 @@ import java.io.IOException;
 import java.io.ObjectInputStream;
 import java.io.ObjectOutputStream;
 import java.lang.invoke.MethodHandle;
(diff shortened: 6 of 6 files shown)


headius:
This feature is still a good idea, but the implementation here suffers from the same problems as the new Data implementation: eagerly setting up the fields on the current class means they will be incorrectly managed on subclasses.

I believe the correct path forward will be to have a hierarchical VariableTableManager that makes its shape decisions based on a current view of both the current class and its superclasses. Future revisions would also be able to maintain several live shapes in a given runtime for a given class, all with a chain of evolving variable table managers to eventually provide the optimal shapes for any class and its subclasses without eagerly pinning the shape.

I'll close this as it cannot land in this condition and a much larger rework of the variable table manager must come first.

headius:
Complete patch preserved here.

```diff
From e9e0cb98c8984d3bf35ca6e11e97c567493453ff Mon Sep 17 00:00:00 2001
From: Charles Oliver Nutter <headius@headius.com>
Date: Mon, 28 Apr 2025 10:34:42 -0500
Subject: [PATCH 1/2] Allow binding instance/internal vars to fields

This is an extension of the existing mechanism for reification of
Ruby classes, allowing us to use normal Java fields for either
instance or internal variables.

The logic here is largely unchanged from the original reification
logic except for enhancements to allow actively requesting a
specific binding:

* Allow specifying a Ruby name to be used in the variable table, so
  that both instance vars (@var) and internal vars (no @) can be
  tied to a specific field.
* Allow the direct field configuration to aggregate a Lookup object
  so that private fields can bound.

The example usage here binds the "args" internal variable on
NoMethodError to the "args" field on RubyNoMethodError, allowing
it to be seen by marshal dumping and fixing one of the DRb issues
in ruby/drb#36. The reification change was not strictly necessary
to fix that issue, but the alternative converts it into a slow-path
internal variable lookup which is less than ideal.

Assuming this change is safe, all other instance and internal
variables used by concrete JRuby core classes can be converted with
the following process:

* Set the reified class (the real concrete class) at definition
  time.
* Request the direct field with a Java name and either the @var or
  bar var name, providing an appropriate Lookup object if the field
  shoud be private.

Enhancements to this feature could include:

* Allow binding read-only variables. Currently both a read and
  write path are expected.
* Enhance conversion logic to allow variables to be a more natural
  Java type. The logic for this exists but has not been heavily
  tested. See Java logic for treating fields as instance variables.
---
 .../java/org/jruby/RubyNoMethodError.java     | 10 +++++++--
 .../util/JavaClassConfiguration.java          | 12 ++++++++++-
 .../runtime/ivars/VariableTableManager.java   | 21 +++++++++++++------
 3 files changed, 34 insertions(+), 9 deletions(-)

diff --git a/core/src/main/java/org/jruby/RubyNoMethodError.java b/core/src/main/java/org/jruby/RubyNoMethodError.java
index d746ab33d0d..b87c5605b1c 100644
--- a/core/src/main/java/org/jruby/RubyNoMethodError.java
+++ b/core/src/main/java/org/jruby/RubyNoMethodError.java
@@ -30,11 +30,14 @@
 import org.jruby.anno.JRubyClass;
 import org.jruby.exceptions.NoMethodError;
 import org.jruby.exceptions.RaiseException;
+import org.jruby.java.codegen.Reified;
 import org.jruby.runtime.Block;
 import org.jruby.runtime.ObjectAllocator;
 import org.jruby.runtime.ThreadContext;
 import org.jruby.runtime.builtin.IRubyObject;
 
+import java.lang.invoke.MethodHandles;
+
 import static org.jruby.api.Define.defineClass;
 
 /**
@@ -43,14 +46,17 @@
  * @see NoMethodError
  */
 @JRubyClass(name="NoMethodError", parent="NameError")
-public class RubyNoMethodError extends RubyNameError {
+public class RubyNoMethodError extends RubyNameError implements Reified {
     private IRubyObject args;
 
     private static final ObjectAllocator ALLOCATOR = RubyNoMethodError::new;
 
     static RubyClass define(ThreadContext context, RubyClass NameError) {
-        return defineClass(context, "NoMethodError", NameError, ALLOCATOR).
+        RubyClass noMethodError = defineClass(context, "NoMethodError", NameError, ALLOCATOR).
                 defineMethods(context, RubyNoMethodError.class);
+        noMethodError.reifiedClass(RubyNoMethodError.class);
+        noMethodError.getVariableTableManager().requestFieldStorage("args", "args", IRubyObject.class, false, null, MethodHandles.lookup());
+        return noMethodError;
     }
 
     protected RubyNoMethodError(Ruby runtime, RubyClass exceptionClass) {
diff --git a/core/src/main/java/org/jruby/javasupport/util/JavaClassConfiguration.java b/core/src/main/java/org/jruby/javasupport/util/JavaClassConfiguration.java
index c257533fd98..2d5e7224bf0 100644
--- a/core/src/main/java/org/jruby/javasupport/util/JavaClassConfiguration.java
+++ b/core/src/main/java/org/jruby/javasupport/util/JavaClassConfiguration.java
@@ -26,6 +26,9 @@
 
 package org.jruby.javasupport.util;
 
+import org.jruby.specialized.RubyObjectSpecializer;
+
+import java.lang.invoke.MethodHandles;
 import java.util.ArrayList;
 import java.util.Arrays;
 import java.util.Collections;
@@ -129,15 +132,22 @@ public synchronized void include(String name) {
     
     public static class DirectFieldConfiguration {
         public String name;
+        public String rubyName;
         public Class<?> fieldType;
         public Boolean unwrap;
         public Class<?> unwrapType;
+        public MethodHandles.Lookup lookup;
         public DirectFieldConfiguration(String name, Class<?> fieldType, Boolean unwrap, Class<?> unwrapType) {
+            this(name, '@' + name, fieldType, unwrap, unwrapType, RubyObjectSpecializer.LOOKUP);
+        }
+
+        public DirectFieldConfiguration(String name, String rubyName, Class<?> fieldType, Boolean unwrap, Class<?> unwrapType, MethodHandles.Lookup lookup) {
             this.name = name;
+            this.rubyName = rubyName;
             this.fieldType = fieldType;
             this.unwrap = unwrap;
             this.unwrapType = unwrapType;
+            this.lookup = lookup;
         }
-        
     }
 }
diff --git a/core/src/main/java/org/jruby/runtime/ivars/VariableTableManager.java b/core/src/main/java/org/jruby/runtime/ivars/VariableTableManager.java
index 1ab578351c1..5c211b56b6b 100644
--- a/core/src/main/java/org/jruby/runtime/ivars/VariableTableManager.java
+++ b/core/src/main/java/org/jruby/runtime/ivars/VariableTableManager.java
@@ -31,6 +31,7 @@
 import java.io.ObjectInputStream;
 import java.io.ObjectOutputStream;
 import java.lang.invoke.MethodHandle;
+import java.lang.invoke.MethodHandles;
 import java.util.ArrayList;
 import java.util.Arrays;
 import java.util.Collections;
@@ -194,8 +195,16 @@ public static void setVariableInternal(RubyClass realClass, RubyBasicObject self
      * This is internal API, don't call this directly if you aren't in the JRuby codebase, it may change
      * Request that the listed ivars (no @ in name) have field storage when we are reified
      */
-    public synchronized void requestFieldStorage(String name, Class<?> fieldType,  Boolean unwrap, Class<?> toType) {
-        DirectFieldConfiguration config = new JavaClassConfiguration.DirectFieldConfiguration(name, fieldType, unwrap, toType);
+    public synchronized void requestFieldStorage(String name, Class<?> fieldType, Boolean unwrap, Class<?> toType) {
+        requestFieldStorage(name, '@' + name, fieldType, unwrap, toType, RubyObjectSpecializer.LOOKUP);
+    }
+
+    /**
+     * This is internal API, don't call this directly if you aren't in the JRuby codebase, it may change
+     * Request that the listed ivars (no @ in name) have field storage when we are reified
+     */
+    public synchronized void requestFieldStorage(String name, String rubyName, Class<?> fieldType, Boolean unwrap, Class<?> toType, MethodHandles.Lookup lookup) {
+        DirectFieldConfiguration config = new JavaClassConfiguration.DirectFieldConfiguration(name, rubyName, fieldType, unwrap, toType, lookup);
         if (realClass.reifiedClass() != null)
             requestFieldStorage(config);
         else {
@@ -212,7 +221,7 @@ public synchronized void requestFieldStorage(String name, Class<?> fieldType,  B
     public void requestFieldStorage(DirectFieldConfiguration config) {
         try {
             Class<? extends Reified> reifiedClass = realClass.reifiedClass();
-            Class<?> fieldType = reifiedClass.getField(config.name).getType();
+            Class<?> fieldType = reifiedClass.getDeclaredField(config.name).getType();
             if (fieldType != config.fieldType) {
                 throw typeError(realClass.getClassRuntime().getCurrentContext(), "java_field " + config.name + " has incorrectly specified types for @ivar mapping");
             }
@@ -224,12 +233,12 @@ public void requestFieldStorage(DirectFieldConfiguration config) {
                 throw typeError(realClass.getClassRuntime().getCurrentContext(), config.name + " has incorrectly specified unwrap type for @ivar mapping: type is incompatible");
             }
             
-            getVariableAccessorForJavaMappedVar("@" + config.name,
+            getVariableAccessorForJavaMappedVar(config.rubyName,
                     unwrap,
                     fieldType,
                     unwrapType,
-                    RubyObjectSpecializer.LOOKUP.findGetter(reifiedClass, config.name, fieldType),
-                    RubyObjectSpecializer.LOOKUP.findSetter(reifiedClass, config.name, fieldType));
+                    config.lookup.findGetter(reifiedClass, config.name, fieldType),
+                    config.lookup.findSetter(reifiedClass, config.name, fieldType));
         } catch (NoSuchFieldException e) {
             throw nameError(realClass.getClassRuntime().getCurrentContext(), "java_field " + config.name + " was marked for @ivar mapping, but wasn't found (was the class reifed already?)", config.name);
         } catch (IllegalAccessException e) {

From e8f387cb350dec18f91e711fc9e15efcff012cd9 Mon Sep 17 00:00:00 2001
From: Charles Oliver Nutter <headius@headius.com>
Date: Thu, 1 May 2025 13:53:44 -0500
Subject: [PATCH 2/2] Add JRubyField annotation for binding Java fields

This hides the plumbing of defining an internal or instance
variable accessor that binds directly to a Java instance field. In
NoMethodError, it is used to bind the "args" field so that it is
seen by Marshal.dump. In other classes, it can be used to avoid
the overhead of dynamically retrieving instance variables when a
field would fit the code better.
---
 core/src/main/java/org/jruby/RubyClass.java   | 43 +++++++++++++++++++
 core/src/main/java/org/jruby/RubyModule.java  |  1 +
 .../java/org/jruby/RubyNoMethodError.java     |  9 ++--
 .../main/java/org/jruby/anno/JRubyField.java  |  5 +++
 .../runtime/ivars/VariableTableManager.java   |  5 +--
 5 files changed, 57 insertions(+), 6 deletions(-)
 create mode 100644 core/src/main/java/org/jruby/anno/JRubyField.java

diff --git a/core/src/main/java/org/jruby/RubyClass.java b/core/src/main/java/org/jruby/RubyClass.java
index 2726736be08..d42fed9474b 100644
--- a/core/src/main/java/org/jruby/RubyClass.java
+++ b/core/src/main/java/org/jruby/RubyClass.java
@@ -54,7 +54,9 @@
 import static org.objectweb.asm.Opcodes.ACC_VARARGS;
 
 import java.io.IOException;
+import java.lang.invoke.MethodHandles;
 import java.lang.reflect.Constructor;
+import java.lang.reflect.Field;
 import java.lang.reflect.InvocationTargetException;
 import java.lang.reflect.Method;
 import java.lang.reflect.Modifier;
@@ -65,6 +67,7 @@
 import java.util.stream.Collectors;
 
 import org.jruby.anno.JRubyClass;
+import org.jruby.anno.JRubyField;
 import org.jruby.anno.JRubyMethod;
 import org.jruby.api.JRubyAPI;
 import org.jruby.compiler.impl.SkinnyMethodAdapter;
@@ -2651,6 +2654,46 @@ public static Class<?> nearestReifiedClass(final RubyClass klass) {
         return null;
     }
 
+    @Override
+    public RubyClass defineMethods(ThreadContext context, Class... methodSources) {
+        return super.defineMethods(context, methodSources);
+    }
+
+    /**
+     * Define instance and internal variables based on Java fields annotated with {@link JRubyField}.
+     *
+     * @param context
+     * @param constantSource class containing the field annotations
+     * @return itself for composable API
+     */
+    @JRubyAPI
+    public <T extends RubyModule> T defineFields(ThreadContext context, Class constantSource, MethodHandles.Lookup lookup) {
+        for (Field field : constantSource.getDeclaredFields()) {
+            if (Modifier.isStatic(field.getModifiers())) continue;
+            defineAnnotatedField(context, field, lookup);
+        }
+        return (T) this;
+    }
+
+    private boolean defineAnnotatedField(ThreadContext context, Field field, MethodHandles.Lookup lookup) {
+        JRubyField jrubyField = field.getAnnotation(JRubyField.class);
+        if (jrubyField == null) return false;
+
+        Class tp = field.getType();
+        Class reifiedType = getReifiedRubyClass();
+        if (reifiedType == null) {
+            throw new RuntimeException("Class does not have a reified form: " + this);
+        }
+
+        String javaName = field.getName();
+        String rubyName = jrubyField.value();
+        if (rubyName.isBlank()) rubyName = javaName;
+
+        getVariableTableManager().addDirectField(javaName, rubyName, tp, false, null, lookup);
+
+        return true;
+    }
+
     public Map<String, List<Map<Class<?>, Map<String,Object>>>> getParameterAnnotations() {
         if (javaClassConfiguration == null || getClassConfig().parameterAnnotations == null) return Collections.EMPTY_MAP;
         return javaClassConfiguration.parameterAnnotations;
diff --git a/core/src/main/java/org/jruby/RubyModule.java b/core/src/main/java/org/jruby/RubyModule.java
index ea99ad95eb1..4245f8de2d3 100644
--- a/core/src/main/java/org/jruby/RubyModule.java
+++ b/core/src/main/java/org/jruby/RubyModule.java
@@ -68,6 +68,7 @@
 import org.jruby.anno.FrameField;
 import org.jruby.anno.JRubyClass;
 import org.jruby.anno.JRubyConstant;
+import org.jruby.anno.JRubyField;
 import org.jruby.anno.JRubyMethod;
 import org.jruby.anno.JavaMethodDescriptor;
 import org.jruby.anno.TypePopulator;
diff --git a/core/src/main/java/org/jruby/RubyNoMethodError.java b/core/src/main/java/org/jruby/RubyNoMethodError.java
index b87c5605b1c..80ad0f7bdd4 100644
--- a/core/src/main/java/org/jruby/RubyNoMethodError.java
+++ b/core/src/main/java/org/jruby/RubyNoMethodError.java
@@ -26,6 +26,7 @@
 
 package org.jruby;
 
+import org.jruby.anno.JRubyField;
 import org.jruby.anno.JRubyMethod;
 import org.jruby.anno.JRubyClass;
 import org.jruby.exceptions.NoMethodError;
@@ -47,15 +48,17 @@
  */
 @JRubyClass(name="NoMethodError", parent="NameError")
 public class RubyNoMethodError extends RubyNameError implements Reified {
+    @JRubyField
     private IRubyObject args;
 
     private static final ObjectAllocator ALLOCATOR = RubyNoMethodError::new;
 
     static RubyClass define(ThreadContext context, RubyClass NameError) {
         RubyClass noMethodError = defineClass(context, "NoMethodError", NameError, ALLOCATOR).
-                defineMethods(context, RubyNoMethodError.class);
-        noMethodError.reifiedClass(RubyNoMethodError.class);
-        noMethodError.getVariableTableManager().requestFieldStorage("args", "args", IRubyObject.class, false, null, MethodHandles.lookup());
+                reifiedClass(RubyNoMethodError.class).
+                defineMethods(context, RubyNoMethodError.class).
+                defineFields(context, RubyNoMethodError.class, MethodHandles.lookup());
+
         return noMethodError;
     }
 
diff --git a/core/src/main/java/org/jruby/anno/JRubyField.java b/core/src/main/java/org/jruby/anno/JRubyField.java
new file mode 100644
index 00000000000..b1ddff5b2c6
--- /dev/null
+++ b/core/src/main/java/org/jruby/anno/JRubyField.java
@@ -0,0 +1,5 @@
+package org.jruby.anno;
+
+public @interface JRubyField {
+    String value() default "";
+}
diff --git a/core/src/main/java/org/jruby/runtime/ivars/VariableTableManager.java b/core/src/main/java/org/jruby/runtime/ivars/VariableTableManager.java
index 5c211b56b6b..9f7f12d603e 100644
--- a/core/src/main/java/org/jruby/runtime/ivars/VariableTableManager.java
+++ b/core/src/main/java/org/jruby/runtime/ivars/VariableTableManager.java
@@ -33,7 +33,6 @@
 import java.lang.invoke.MethodHandle;
 import java.lang.invoke.MethodHandles;
 import java.util.ArrayList;
-import java.util.Arrays;
 import java.util.Collections;
 import java.util.LinkedHashMap;
 import java.util.Map;
@@ -196,14 +195,14 @@ public static void setVariableInternal(RubyClass realClass, RubyBasicObject self
      * Request that the listed ivars (no @ in name) have field storage when we are reified
      */
     public synchronized void requestFieldStorage(String name, Class<?> fieldType, Boolean unwrap, Class<?> toType) {
-        requestFieldStorage(name, '@' + name, fieldType, unwrap, toType, RubyObjectSpecializer.LOOKUP);
+        addDirectField(name, '@' + name, fieldType, unwrap, toType, RubyObjectSpecializer.LOOKUP);
     }
 
     /**
      * This is internal API, don't call this directly if you aren't in the JRuby codebase, it may change
      * Request that the listed ivars (no @ in name) have field storage when we are reified
      */
-    public synchronized void requestFieldStorage(String name, String rubyName, Class<?> fieldType, Boolean unwrap, Class<?> toType, MethodHandles.Lookup lookup) {
+    public synchronized void addDirectField(String name, String rubyName, Class<?> fieldType, Boolean unwrap, Class<?> toType, MethodHandles.Lookup lookup) {
         DirectFieldConfiguration config = new JavaClassConfiguration.DirectFieldConfiguration(name, rubyName, fieldType, unwrap, toType, lookup);
         if (realClass.reifiedClass() != null)
             requestFieldStorage(config);
```

PR #8813, opened 2025-04-29
Fix possible NPE in UnboxableOpsAnalysisNode:getUnboxedVar()

### LINKED ISSUE #8812: Possible NPE in UnboxableOpsAnalysisNode:getUnboxedVar() found by SAST
Our SAST shows possible NullPointerException in https://github.com/jruby/jruby/blob/master/core/src/main/java/org/jruby/ir/dataflow/analyses/UnboxableOpsAnalysisNode.java#L362

```
    private TemporaryLocalVariable getUnboxedVar(Class reqdType, Map<Variable, TemporaryLocalVariable> unboxMap, Variable v, boolean createNew) {
        TemporaryLocalVariable unboxedVar = unboxMap.get(v);
        // FIXME: This is a bit broken -- SSA will eliminate this need for type verification
        if ((unboxedVar == null && createNew) || !matchingTypes(reqdType, unboxedVar.getType())) {
            unboxedVar = problem.getFIC().getNewUnboxedVariable(reqdType);
            unboxMap.put(v, unboxedVar);
        } else if (unboxedVar == null) {
            // FIXME: throw an exception here
            System.out.println("ERROR: No unboxed var for : " + v);
        }
        return unboxedVar;
    }
```

If `unboxedVar` is null and `createNew` is false, then `unboxedVar.getType()` will cause NPE. And null `unboxedVar` is expectable condition, but appears to be unreachable.

**Environment Information**
  
We are analyzing versions 9.4.x (8-12), but this problem is still in master

**Expected Behavior**

* No NPE at all. But it seems nobody catch it through the years, so i don't know if this condition exists at all

**Actual Behavior**

* have no tests to show this NPE, can't figure out how to make one. This is simply code analysis, that showed possible NPE

ngr-ilmarh:
Fix #8812

### DIFF
--- a/core/src/main/java/org/jruby/ir/dataflow/analyses/UnboxableOpsAnalysisNode.java
+++ b/core/src/main/java/org/jruby/ir/dataflow/analyses/UnboxableOpsAnalysisNode.java
@@ -359,7 +359,7 @@ public class UnboxableOpsAnalysisNode extends FlowGraphNode<UnboxableOpsAnalysis
     private TemporaryLocalVariable getUnboxedVar(Class reqdType, Map<Variable, TemporaryLocalVariable> unboxMap, Variable v, boolean createNew) {
         TemporaryLocalVariable unboxedVar = unboxMap.get(v);
         // FIXME: This is a bit broken -- SSA will eliminate this need for type verification
-        if ((unboxedVar == null && createNew) || !matchingTypes(reqdType, unboxedVar.getType())) {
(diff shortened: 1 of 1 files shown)


headius:
The unboxed logic has not been active for many years, and was experimental even then. I would like to see some tracing to see if `unboxedVar` is ever actually null, and if it is not we could just guarantee that in other ways and eliminate the checks.

PR #8820, opened 2025-04-30
Fix recent potential NPE reports

headius:
Fixes for the stack of "potential NPE" reports that have come in recently.


headius:
Most of the remaining NPE reports either have PRs in progress or it's not clear how or if they should be fixed.

Issue #8634, opened 2025-02-17
9.4.12.0 unable to build itself


jcharaoui:
I'm currently working to update the packaging for JRuby in Debian, which currently ships 9.4.8.0. I'm able to successfully build 9.4.12.0, but I'm hitting an issue when using 9.4.12.0 in the JRuby build environment.

```
[INFO] --- polyglot:0.4.11:execute (install_gems) @ jruby-stdlib ---
using jruby 9.4.12.0
install gems unless already installed
/build/reproducible-path/jruby-9.4.12.0+ds/debian/maven-repo/rubygems/rubygems-update/3.5.0/rubygems-update-3.5.0.gem
Exception while verifying /build/reproducible-path/jruby-9.4.12.0+ds/debian/maven-repo/rubygems/rubygems-update/3.5.0/rubygems-update-3.5.0.gem
[INFO] ------------------------------------------------------------------------
[INFO] Reactor Summary for JRuby 9.4.12.0:
[INFO]          
[INFO] JRuby .............................................. SUCCESS [  0.104 s]
[INFO] JRuby Base ......................................... SUCCESS [ 19.878 s]
[INFO] JRuby Core ......................................... SUCCESS [  6.332 s]
[INFO] JRuby Lib Setup .................................... FAILURE [  0.365 s]
[INFO] JRuby Artifacts .................................... SKIPPED
[INFO] JRuby Main Maven Artifact .......................... SKIPPED
[INFO] JRuby Complete ..................................... SKIPPED
[INFO] ------------------------------------------------------------------------
[INFO] BUILD FAILURE                                                                                      
[INFO] ------------------------------------------------------------------------
[INFO] Total time:  29.303 s             
[INFO] Finished at: 2025-02-17T18:01:26Z                                                                  
[INFO] ------------------------------------------------------------------------
[ERROR] Failed to execute goal io.takari.polyglot:polyglot-maven-plugin:0.4.11:execute (install_gems) on project jruby-stdlib: org.jruby.exceptions.NameError: (NameError) uninitialized constant Gem::Security::Open
SSL -> [Help 1]
```

I'm tracked it down the problem to the `rubygems-update` upgrade to 3.6.3 in c4e369d42c8f70b838af81e8bd89251f63b65eb3. If I revert the gem to 3.3.26, then it builds with itself just fine.

[...]

PR #8822, opened 2025-05-02
Build updates

cstamas:
Changes:
* Polyglot to 0.8.0 (JRuby 9.4.12.0)
* maven wrapper to latest
* Maven to 3.9.9
* Added RRF

Out of these Polyglot update may be there interesting one.


cstamas:
Locally fails in very same (similar?) manner as https://github.com/jruby/jruby/issues/8634

https://gist.github.com/cstamas/8ed13f20404b9a9bdf42957f26692516

PR #8829, opened 2025-05-05
Data fixes and improvements

headius:
The fixes here were driven by net-imap tests (https://github.com/ruby/net-imap/issues/454) that depended on these features to even start up the library and tests. The fixes include:

* Improvements for Data subclasses, which were not fully supported at JRuby 10 release time.
* Allow Data to have no fields.

There will need to be additional work to handle the variable table of data subclasses, rather than searching up the hierarchy, but this is enough to get the net-imap tests running.


eregon:
Let's merge this?
I think it is blocking https://github.com/ruby/net-imap/pull/470

Issue #5023, opened 2018-01-31
excessive loading of same InterfaceImpl


MartinNowak:
### Environment

Provide at least:
* jruby 9.1.15.0 (2.3.3) 2017-12-07 929fde8 Java HotSpot(TM) 64-Bit Server VM 25.144-b01 on 1.8.0_144-b01 +jit [linux-x86_64]
  `jruby -J-Xms1g -J-Xmx1g -J-XX:+UseG1GC -J-XX:MetaspaceSize=192M -Xcompile.invokedynamic=true`
* Linux 4.4.0-109-generic #132-Ubuntu SMP Tue Jan 9 19:52:39 UTC 2018 x86_64 x86_64 x86_64 GNU/Linux

Other relevant info you may wish to add:
* jbundler, pry, and rake
* vert.x-3.5.0, cassandra-driver, kafka, and rxjava (all java libs)

### Expected Behavior

* Identical
* Describe your expectation of how JRuby should behave, perhaps by showing how CRuby/MRI behaves.
* Provide an executable Ruby script or a link to an example repository.

### Actual Behavior

* Describe or show the actual behavior.
* Provide text or screen capture showing the behavior.

Verbose class loading log, digged into this when seeing 2+M loaded classes in JConsole, most being collected again, ~50K classes resident.
```
Jan 31 11:49:32 bundle[18113]: [Loaded org.jruby.gen.InterfaceImpl702540685 from __JVM_DefineClass__]
Jan 31 11:49:32 bundle[18113]: [Loaded org.jruby.gen.InterfaceImpl702540685 from __JVM_DefineClass__]
Jan 31 11:49:32 bundle[18113]: [Loaded org.jruby.gen.InterfaceImpl702540685 from __JVM_DefineClass__]
```
[iface_impl.log](https://github.com/jruby/jruby/files/1681566/iface_impl.log)

With `-Xinterfaces.useProxy=true` I'm only seeing ~13K resident classes and ~750MiB less memory being used (guess that's Metaspace).

kares:
believe they're cached to some extend.
what would help to confirm a bug is a reproducer, do you know where they're created from?
would you care to attempt a minimal reproducer - where you expect the same Impl class to be reused and can confirm it isn't.

enebo:
I half wondered if vert.x isolates via classpaths and perhaps confuses a cache as needing to regenerate something?

MartinNowak:
> believe they're cached to some extend.
[...]

PR #8843, opened 2025-05-20
Don't bind synthetic classes to Ruby

### LINKED ISSUE #8842: Metaspace leak in 9.4
**Environment Information**
  
```
$ ruby --version
jruby 9.4.12.0 (3.1.4) 2025-02-11 f4ab75096a OpenJDK 64-Bit Server VM 11.0.15+0-adhoc..source on 11.0.15+0-adhoc..source +jit [x86_64-linux]

$ java -version
openjdk version "11.0.15" 2022-04-19
OpenJDK Runtime Environment (build 11.0.15+0-adhoc..source)
OpenJDK 64-Bit Server VM (build 11.0.15+0-adhoc..source, mixed mode)
```

**Behavior**

Since upgrading from JRuby 9.3 to 9.4, we are seeing what appears to be unbounded Metaspace usage.  This eventually leads to an OOM: Metaspace exception and process exit.  

As an example, we can see this behavior in our production monitoring.  When running 9.4, Total Classes and Loaded Classes closely track each other and tend to rise over time.  The rate of increase tracks the daily usage of the instance.  When running 9.3, Total Classes also rises, but Loaded Classes plateaus.  This corresponds to a plateau in the total Metaspace usage, as classes appear to be GCed.
Image

Running locally with class loader logging enabled (`-Xlog:class+load=debug:loadedClasses-%t-%p.txt -Xlog:class+path=info:loadedClasses-%t-%p.txt -Xlog:class+unload=trace:loadedClasses-%t-%p.txt`), I can see in 9.3 that there are both class,load and class,unload log lines.

```
[282.268s][info ][class,load  ] org.apache.calcite.linq4j.tree.Expressions$FluentList source: file:/u
sr/local/google/home/kalenp/src/helltool/ruby-gems/ht_java/lib/jar/helltool-java-all.jar
[282.268s][debug][class,load  ]  klass: 0x00000008427f6728 super: 0x0000000800004460 interfaces: 0x00
000008002ec6d8 loader: [loader data: 0x00007fa79c96db40 for instance a 'org/jruby/util/JRubyClassLoad
er'{0x00000000a1c2c8b8}] bytes: 1188 checksum: fa4eb98b
[282.268s][info ][class,load  ] org.apache.calcite.linq4j.tree.TernaryExpression source: file:/usr/lo
cal/google/home/kalenp/src/helltool/ruby-gems/ht_java/lib/jar/helltool-java-all.jar
[282.268s][debug][class,load  ]  klass: 0x00000008427f6928 super: 0x00000008425afa28 loader: [loader
data: 0x00007fa79c96db40 for instance a 'org/jruby/util/JRubyClassLoader'{0x00000000a1c2c8b8}] bytes:
 3861 checksum: 750abf23
[282.269s][info ][class,load  ] usr.local.google.home.kalenp.src.helltool.lib.helltool.models.encrypt
ed.db_connection source: __JVM_DefineClass__
[282.269s][debug][class,load  ]  klass: 0x00000008427da040 super: 0x0000000800004460 loader: [loader
data: 0x00007fa6bd586500 for instance a 'org/jruby/util/OneShotClassLoader'{0x0000000153874500}] byte
s: 4017 checksum: abf1c398
...
[282.274s][info ][class,unload] unloading class java.lang.invoke.LambdaForm$DMH/0x0000000841d97c40 0x
0000000841d97c40
[282.274s][info ][class,unload] unloading class java.lang.invoke.LambdaForm$MH/0x0000000841d95840 0x0
000000841d95840
[282.274s][info ][class,unload] unloading class java.lang.invoke.LambdaForm$MH/0x0000000841d35c40 0x0
000000841d35c40
[282.274s][info ][class,unload] unloading class java.lang.invoke.LambdaForm$MH/0x0000000841d32c40 0x0
000000841d32c40
[282.274s][info ][class,unload] unloading class java.lang.invoke.LambdaForm$DMH/0x0000000841cca840 0x
0000000841cca840
[282.274s][info ][class,unload] unloading class java.lang.invoke.LambdaForm$MH/0x0000000841ca9440 0x0
000000841ca9440
[282.274s][info ][class,unload] unloading class java.lang.invoke.LambdaForm$MH/0x0000000841ca7440 0x0
000000841ca7440
```

In 9.4, I see a similar pattern of class,load lines.  However, for class,unload, I only see a few at startup and then a large number at application shutdown time.  I don't see the same pattern of unloading at shutdown time in 9.3.

headius:
Synthetic classes, such as our own "InterfaceImpl" proxies generated on-the-fly to adapt Ruby objects to Java interfaces, are generally not supposed to be visible to users. A change in JRuby 9.4 to reduce wrapping overhead for InterfaceImpl proxy objects led to them also being bound as Ruby proxy classes, which had the effect of causing these frequently-generated and frequently- abandoned classes to be pinned in memory by the proxy class's method bindings. This led to a class and metaspace leak reported in #8842.

The fix here explicitly avoids binding any synthetic classes, instead skipping such classes and walking up to the nearest non- synthetic class before beginning the binding process. This avoids our InterfaceImpl classes from being bound as Ruby proxies and appears to eliminiate at least the leak reproduced in #8842.

This may also fix other undiscovered or un-reproduced cases of generated classes leaking into the Ruby proxy class hierarchy, so long as all of those classes are being generated as synthetic.

This may also be a more direct and appropriate fix for the eager binding and logging of InterfaceImpl classes reported in #8349, which was previously fixed by avoiding synthetic classes when adding proxies to our Java package namespace structures (#8503). If that is the case, we may want to consider turning that check into an error, since it will probably always be incorrect to have synthetic classes get bound into a public namespace. Additional checks elsewhere in our Java Integration subsystem could also help prevent similar issues in the future, and we should audit all generated classes to ensure they are marked as synthetic.

Fixes #8842.

May be a better fix for #8349 than #8503 but that fix's check should perhaps remain or become a hard error.

Needs tests for as many JI class generation cases as possible.

### DIFF
--- a/core/src/main/java/org/jruby/javasupport/Java.java
+++ b/core/src/main/java/org/jruby/javasupport/Java.java
@@ -430,7 +430,11 @@ public class Java implements Library {
     }
 
     @SuppressWarnings("deprecation")
-    public static RubyModule getProxyClass(final Ruby runtime, final Class<?> clazz) {
+    public static RubyModule getProxyClass(final Ruby runtime, Class<?> clazz) {
+        // skip synthetic classes we should not see, like InterfaceImpl*
+        while (clazz.isSynthetic()) {
+            clazz = clazz.getSuperclass();
+        }
         RubyModule proxy = runtime.getJavaSupport().getUnfinishedProxy(clazz);
         if (proxy != null) return proxy;
         return runtime.getJavaSupport().getProxyClassFromCache(clazz);


headius:
Note also that this would not have been as serious an issue if we were reusing InterfaceImpl classes that represented the same superclass and interfaces. Unfortunately, as reported in other issues like #5023. We should also fix that, but due to the more invasive nature of such a fix it might be safest to do it only on the active-development JRuby 10 line.

headius:
The tests indicate that synthetic is too broad a net to cast. Lambda expressions result in interface implementation classes that are synthetic, but they do not have a superclass with the requisite methods to be invoked. We probably need to mark our generated classes with an additional interface and use that to indicate that they should not be bound as proxy classes.

This issue could also affect the fix in #8503 but at least lambda expressions don't have a named class we can use directly and would not go into the package structure.

headius:
My experiments trying to ignore interface impl classes led to a few discoveries:

* As found with the per-class ignore logic, we could not properly call methods on synthetic classes from the JDK like lambda expressions.
* No level of ignorance allowed all such classes to remain functional, but allowing them to construct Ruby proxy classes and methods resulted in rooting of the generated impls.
* The rooting appeared to be primarily due to the implementation of per-class values in our own ClassValue mimic. The default implementation held neither classes nor values weak. The Java 7 version holds only classes weak, but does it in a way that still seems to leak.
* Reverting to just using plain old JDK ClassValue behavior appears to eliminate the leaks and produce a stable memory profile even on a more aggressive test.

See #8844.

core/src/main/java/org/jruby/javasupport/Java.java
```diff
@@ -430,7 +430,11 @@ public static RubyModule getProxyClass(Ruby runtime, JavaClass javaClass) {
     }
 
     @SuppressWarnings("deprecation")
-    public static RubyModule getProxyClass(final Ruby runtime, final Class<?> clazz) {
+    public static RubyModule getProxyClass(final Ruby runtime, Class<?> clazz) {
+        // skip synthetic classes we should not see, like InterfaceImpl*
+        while (clazz.isSynthetic()) {
```
kares, line 435: isn't there a better place to put to check, with regards to the interface impl that are problematic?

feels a bit confusing, and most callers do not need to worry about having a synthetic class here.

maybe add another `getClosestProxyClass` as an alternative? 
that way we would know which calls are internal and which are from user supplied classes...

headius: It turned out this is too aggressive; lambda expressions are synthetic and won't properly bind methods, and because we bind interface methods on the concrete type (interfaces don't really map to modules right) we always need that lowest concrete type to hold the bindings. But #8844 seems to fix it, even though it's much more invasive. We can hide it behind an option for users that run into this, and we should also finally fix the re-use of interface impl classes.

headius:
Wrong approach; we do actually need to bind these classes so there's a place for their interface methods to live, and lambda expressions are synthetic and invoking them from Ruby breaks with this change. See #8844 for a better fix.

PR #8844, opened 2025-05-21
Use a normal ClassValue for all such cases

headius:
JRuby includes two versions of ClassValue: one with a hard- referenced Map from Class to value, and one based on a normal ClassValue but with the added oddity that it hard-references its values while storing them in the ClassValue slot weakly. This was originally intended to allow the values to collect properly when the ClassValue collected, but on recent testing it seems to just leak in a different way.

This change siwtches everything to normal ClassValue. Given this change, the example cases in #8842 no longer leak, nor does a more aggressive torture test that also spins up and tears down JRuby runtimes as well as interface implementations. A 30-minute run of that test uses a lot of memory and GCs heavily, but does level off in both heap and metaspace. Forced GCs reduce both nearly back to startup levels.

Note this is a work in progress; we will likely not want to change the default behavior for 9.4.x, so this will need to be guarded behind an option relating to the existing invokedynamic.class.values toggle. It may be safe enough as-is for JRuby 10 as that is still stabilizing.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -37,6 +37,36 @@ jobs:
       - name: rake ${{ matrix.target }}
         run: bin/jruby -S rake ${{ matrix.target }}
 
+  rake-test-ji-stable-class-values:
+    runs-on: ubuntu-latest
+
+    strategy:
+      matrix:
+        target: ['spec:ji']
+        java-version: ['8']
+      fail-fast: false
+
+    name: rake ${{ matrix.target }} (Java ${{ matrix.java-version }})
+
+    env:
+      JRUBY_OPTS: "-Xji.class.values=STABLE"
+
+    steps:
+      - name: checkout
+        uses: actions/checkout@v3
+      - name: set up java ${{ matrix.java-version }}
+        uses: actions/setup-java@v3
+        with:
+          distribution: 'zulu'
+          java-version: ${{ matrix.java-version }}
+          cache: 'maven'
+      - name: bootstrap
+        run: mvn -ntp -Pbootstrap clean package
+      - name: bundle install
+        run: bin/jruby --dev -S bundle install
+      - name: rake ${{ matrix.target }}
+        run: bin/jruby -S rake ${{ matrix.target }}
+
   rake-test-8:
     runs-on: ubuntu-latest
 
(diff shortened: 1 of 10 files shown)


headius:
Additional tweak to synchronize the class value calculators didn't seem to solve the related test failures, so there's a bit more work required here. And of course guarding the new logic behind an option.

headius:
I realized a problem with the new torture test: because the ClassValue objects rooting these classes go away with the JRuby instance, they don't leak in any of the modes. It does show, however, that classes are not leaking across instances in other the old mode or the new mode.

Using the original single-runtime test, I can confirm that the old mode leaks very rapidly and chokes, while the new mode levels off and continues running just fine.

Old (shorter because it started to hit metaspace limit):
old_classvalue_single_runtime

New:
classvalue_fix_single_runtime

I will push an update shortly that maintains compatibility with the JRuby 9.4 ClassValue implementation and provides a new option for configuring the fixed mode. This mode will be the default in JRuby 10.0.1.

headius:
From the most recent commit, which should align this work with JRuby 9.4 behavior and allow configuring the new STABLE option to fix #8842:

```
In order to adopt the JDK ClassValue for our own ClassValue API,
and maintain compatibility with JRuby 9.4, this patch includes the
following changes:

* Add a StableClassValue that uses Java's ClassValue but with a
  once-only computation guarantee. This is the mode that fixes
  https://github.com/jruby/jruby/issues/8842.
* Deprecate the old invokedynamic.class.values option, but allow
  it to still select the Java7ClassValue implementation.
* Add a new ji.class.values option based on ClassValue.Type enum
  and the following values:
  * STABLE, the stable java.lang.ClassValue implementation
  * HARD_MAP, the default hard-referenced Map implementation
  * HARD_VALUES, the Java7 version
  JRuby 9.4 will continue to default to the HARD_MAP version, to
  avoid introducing the new behavior where it is not expected.
  JRuby 10 will merge this change and default to STABLE starting
  with 10.0.1.

See discussion of ClassValue and stability on the mlvm-dev mailing
list here:
```

headius:
Tests appear to be green other than an oddity in sequel. The functional parts of this PR can be reviewed now. I still need to figure out some tests.

headius:
The sequel failure was due to a new MariaDB test mode being run, so I've merged in the appropriate configs and repushed.

headius:
A 12h graph of heap and metaspace stability with stable class values:

classvalue_fix_single_runtime_12h

headius:
I have pushed a regression test that confirms both modes' behaviors:

* STABLE (the new mode) should unload the implementation class.
* HARD_MAP (the old mode) should not unload the implementation class.

@kares @enebo check my work here. This took too long to figure out a test for, and might be fragile, but it's very difficult to detect class unloading on JVM.

kares (approved):
This looks great, well done!

Issue #8736, opened 2025-04-01
Time.new timezone changes for 3.2-3.4


headius:
There appears to be some new specs for Time.new handling of timezones, possibly new behavior for Ruby 3.2-3.4. We will address these after the release of JRuby 10.

headius:
Likely more of the same here in CRuby's test/ruby/test_time_tz.rb:

```
TestTimeTZ#test_utc_names [/Users/headius/work/jruby/test/mri/ruby/test_time_tz.rb:267]:
TestTimeTZ::DummyTZ#test_dst?@America/Los_Angeles [/Users/headius/work/jruby/test/mri/ruby/test_time_tz.rb:739]:
TestTimeTZ::DummyTZ#test_fractional_second@America/Los_Angeles [/Users/headius/work/jruby/test/mri/ruby/test_time_tz.rb:702]:
TestTimeTZ::DummyTZ#test_marshal@Africa/Ndjamena [/Users/headius/work/jruby/test/mri/ruby/test_time_tz.rb:694]:
TestTimeTZ::DummyTZ#test_marshal@America/Los_Angeles [/Users/headius/work/jruby/test/mri/ruby/test_time_tz.rb:694]:
TestTimeTZ::DummyTZ#test_marshal@Asia/Tokyo [/Users/headius/work/jruby/test/mri/ruby/test_time_tz.rb:694]:
TestTimeTZ::DummyTZ#test_marshal@Etc/UTC [/Users/headius/work/jruby/test/mri/ruby/test_time_tz.rb:694]:
TestTimeTZ::DummyTZ#test_new@Africa/Ndjamena [/Users/headius/work/jruby/test/mri/ruby/test_time_tz.rb:620]:
TestTimeTZ::DummyTZ#test_new@America/Los_Angeles [/Users/headius/work/jruby/test/mri/ruby/test_time_tz.rb:620]:
TestTimeTZ::DummyTZ#test_new@Asia/Tokyo [/Users/headius/work/jruby/test/mri/ruby/test_time_tz.rb:620]:
TestTimeTZ::DummyTZ#test_new@Etc/UTC [/Users/headius/work/jruby/test/mri/ruby/test_time_tz.rb:620]:
TestTimeTZ::DummyTZ#test_strftime@Africa/Ndjamena [/Users/headius/work/jruby/test/mri/ruby/test_time_tz.rb:652]:
[...]

PR #8858, opened 2025-06-05
Fix `Time.new` when using (empty) keywords (2 issues)

PChambino:
Fixes #8854 and #8860.

```
#8854
irb> Time.new(2025, 4, 3, **{})
org.jruby.dist/org.jruby.RubyTime.initialize(RubyTime.java:1758):
Cannot load from object array because "opts" is null (Java::JavaLang::NullPointerException)

#8860
irb> Time.new(2025, 4, 3, in: 'UTC')
=> 2025-04-01 00:00:00 +0000
# The day is incorrect. It always loses the last arg.
```

The [`travel_to` test helper from ActiveSupport](https://github.com/rails/rails/blob/v7.1.5.1/activesupport/lib/active_support/testing/time_helpers.rb#L185) triggers the empty keywords issue.

As a result of fixing the NullPointerException and writing a regression test, it actually triggered the other issue of losing the last argument when using keywords.

I realised that if keywords was empty, `hasKeywords` still returns true and `extractKeywordArgs` returns null, and `args` doesn't have the Hash object, it is just missing. So that is why we can only subtract 1 from argc after `extractKeywordArgs` and confirming it is not null. Also, since we are already subtracting one, then we don't need the keywords condition again in the argc switch.

I added a test for it since it was a regression from JRuby v9.

PR #8890, opened 2025-07-04
Fix Time.new parsing (time/new_tags.txt)

PChambino:
The idea was to help towards this #8736 but most of the fixes here are specifically around parsing the time string, not timezone related.

@enebo I noticed you did recent changes around Time.new parsing as well, in case you want to review.

I'll add some comments in the PR to highlight which changes fix which parts, but here is the TLDR:

- Time.new parsing should ignore "whitespace" chars, not only space chars

- precision keyword argument should only be used by Time.new parsing and ignored otherwise (see Ruby docs), and it should default to 9

- RubyTimeParser.parseInt now returns nil if no digits (so missing year can be detected) and only to a set number of digits (for subsec precision)

- correctly convert subsec to microseconds in RubyTimeParser when more than 9 digits

### DIFF
--- a/spec/tags/ruby/core/time/new_tags.txt
+++ b/spec/tags/ruby/core/time/new_tags.txt
@@ -1,7 +1,6 @@
 fails:Time.new with a utc_offset argument returns a Time with a UTC offset of the specified number of Rational seconds
 fails:Time.new with a utc_offset argument raises ArgumentError if the String argument is not in an ASCII-compatible encoding
 fails:Time.new with a utc_offset argument with an argument that responds to #to_r coerces using #to_r
-fails:Time.new with a utc_offset argument raises ArgumentError if the month is greater than 12
 fails(not implemented, jruby/jruby#6161):Time.new with a timezone argument the #abbr method is used by '%Z' in #strftime
 fails(not implemented, jruby/jruby#6161):Time.new with a timezone argument Time-like argument of #utc_to_local and #local_to_utc methods has attribute values the same as a Time object in UTC
 fails(not implemented, jruby/jruby#6161):Time.new with a timezone argument #name method uses the optional #name method for marshaling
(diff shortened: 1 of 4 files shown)


spec/tags/ruby/core/time/new_tags.txt
```diff
@@ -1,20 +1,13 @@
...
 fails(not implemented, jruby/jruby#6161):Time.new with a timezone argument Time-like argument of #utc_to_local and #local_to_utc methods has attribute values the same as a Time object in UTC
 fails(not implemented, jruby/jruby#6161):Time.new with a timezone argument #name method uses the optional #name method for marshaling
 fails(not implemented, jruby/jruby#6161):Time.new with a timezone argument #name method cannot marshal Time if #name method isn't implemented
 fails(not implemented, jruby/jruby#6161):Time.new with a timezone argument subject's class implements .find_timezone method calls .find_timezone to build a time object at loading marshaled data
 fails(only during full spec run):Time.new with a utc_offset argument raises ArgumentError if the String argument is not of the form (+|-)HH:MM
 fails:Time.new with a timezone argument returned value by #utc_to_local and #local_to_utc methods cannot have arbitrary #utc_offset if it is an instance of Time
-fails(https://github.com/jruby/jruby/issues/8736):Time.new with a timezone argument :in keyword argument allows omitting minor arguments
```
PChambino, line 12: This was already passing, I think I might have fixed it in my previous PR: https://github.com/jruby/jruby/pull/8858
But I wasn't aware of these tag files.

spec/tags/ruby/core/time/new_tags.txt
```diff
@@ -1,20 +1,13 @@
...
 fails(https://github.com/jruby/jruby/issues/8736):Time.new with a timezone argument Time.new with a String argument accepts precision keyword argument and truncates specified digits of sub-second part
```
PChambino, line 11: RubyTime only has precision up to 9 (nanoseconds) so it is not possible to remove this tag, but the microseconds conversion was wrong ( `+ 3` vs `- 3`) when there were more than 9 digits. So it now truncates correctly.

spec/tags/ruby/core/time/new_tags.txt
```diff
@@ -1,20 +1,13 @@
...
-fails(https://github.com/jruby/jruby/issues/8736):Time.new with a timezone argument Time.new with a String argument converts precision keyword argument into Integer if is not nil
 fails(https://github.com/jruby/jruby/issues/8736):Time.new with a timezone argument Time.new with a String argument raises ArgumentError if date/time parts values are not valid
 fails(https://github.com/jruby/jruby/issues/8736):Time.new with a timezone argument Time.new with a String argument raises ArgumentError if utc offset parts are not valid
-fails(https://github.com/jruby/jruby/issues/8736):Time.new with a timezone argument Time.new with a String argument raises ArgumentError if string doesn't start with year
-fails(https://github.com/jruby/jruby/issues/8736):Time.new with a timezone argument Time.new with a String argument raises ArgumentError when there are leading space characters
-fails(https://github.com/jruby/jruby/issues/8736):Time.new with a timezone argument Time.new with a String argument raises ArgumentError when there are trailing whitespaces
```
PChambino, line 19: These two were just about ignoring more "whitespace" chars, not just spaces.

spec/tags/ruby/core/time/new_tags.txt
```diff
@@ -1,20 +1,13 @@
...
-fails:Time.new with a timezone argument Time.new with a String argument returns Time with correct subseconds when given seconds fraction is longer than 9 digits
```
PChambino, line 20: This was fixed by adding support for "precision" in RubyTimeParser.parseInt.

spec/tags/ruby/core/time/new_tags.txt
```diff
@@ -1,20 +1,13 @@
 fails:Time.new with a utc_offset argument returns a Time with a UTC offset of the specified number of Rational seconds
 fails:Time.new with a utc_offset argument raises ArgumentError if the String argument is not in an ASCII-compatible encoding
 fails:Time.new with a utc_offset argument with an argument that responds to #to_r coerces using #to_r
-fails:Time.new with a utc_offset argument raises ArgumentError if the month is greater than 12
```
PChambino, line 4: This was already passing, so I just removed it.

enebo:
@PChambino cool.  The -3 vs +3 makes sense and explains a confusion I had when working on the other recent PR I did :smile:.

I am mildly confused about the already passing specs since I thought I had did a mspec-tag to untag passing things but I must have done that manually?

PR #8898, opened 2025-07-08
Update maven plugins to latest

headius:
This fixes issues with downloads (of poms, gems, etc) sometimes garbling the output and including null bytes (jruby/jruby-maven-plugins@fea0a42502af8bb94e9dc6c9e4a2d7aa394a08cf) as well as updating to latest JRuby 9.4 and a few other small fixes.


headius:
Failures are due to new strscan, which I deal with in #8900.

PR #8901, opened 2025-07-09
Update to ruby/spec@ed254ba

andrykonchin:


headius:
I don't know what's up with those string scanner tests but I'll deal with it on Master if it still fails. Thanks @andrykonchin!

PR #8907, opened 2025-07-10
Symbol table tweaks

headius:
A few small tweaks to the symbol table.


headius:
These were discovered while attempting to get `test_marshal_with_ruby2_keywords_hash` passing, since it depends on comparing an incoming UTF-7 "K" symbol with the default US-ASCII version already in the table. Fixing the reset ordering resolved an issue finding the existing symbol (it was being created and then wiped out) but because we key our symbol table off the hash of the raw bytes without the encoding, we can't ultimately distinguish between these otherwise identical symbols. There's no way to pass the test with the symbol table not considering encoding at this point (see some explanation of that limitation at https://github.com/jruby/jruby/pull/6341#issuecomment-664588352).

PR #8912, opened 2025-07-12
Don't include Object for pathed constant searches

headius:
The fix here is a little bit form-fitting but largely avoids inheriting constants from Object when doing String-based constant lookups and skips the additional search of Object if we are doing a path search.


headius:
I did this fix against 9.4 but it will be released in 10.0.1.0 first. Stupid github still doesn't allow multiple milestones.

Issue #8920, opened 2025-07-18
KWargs issue with mustermann gem preventing JRuby 10.0.1.0 upgrade


NC-piercej:
**Environment Information**

JRuby: 10.0.1.0 (via `jruby:10.0.1.0-jdk21` docker image)

Using Sinatra 4.1.1 with the `mustermann` gem at `3.0.3`

**Expected Behavior**

We expect no regressions from JRuby 10.0.0.1.

**Actual Behavior**

When running unit tests for the controller using `Rack::Test::Methods`:

```
     Failure/Error:
       post "services/#{service_name}/#{version}", body, {
         "Content-Type": "application/octet-stream"
       }
     
     ArgumentError:
       wrong number of arguments (given 2, expected 1)
     Shared Example Group: "endpoint configuration route" called from ./spec/controllers/services_controller_spec.rb:63
     # /usr/local/bundle/gems/mustermann-3.0.3/lib/mustermann/ast/parser.rb:17:in 'parse'
     # /usr/local/bundle/gems/mustermann-3.0.3/lib/mustermann/ast/pattern.rb:91:in 'block in to_ast'
     # /usr/local/bundle/gems/mustermann-3.0.3/lib/mustermann/equality_map.rb:43:in 'fetch'
     # /usr/local/bundle/gems/mustermann-3.0.3/lib/mustermann/ast/pattern.rb:90:in 'to_ast'
     # /usr/local/bundle/gems/mustermann-3.0.3/lib/mustermann/ast/pattern.rb:130:in 'param_converters'
     # /usr/local/bundle/gems/mustermann-3.0.3/lib/mustermann/ast/pattern.rb:124:in 'map_param'
     # /usr/local/bundle/gems/mustermann-3.0.3/lib/mustermann/pattern.rb:207:in 'block in params'
     # /usr/local/bundle/gems/mustermann-3.0.3/lib/mustermann/pattern.rb:207:in 'block in params'
     # /usr/local/bundle/gems/mustermann-3.0.3/lib/mustermann/pattern.rb:206:in 'params'
     # /usr/local/bundle/gems/sinatra-4.1.1/lib/sinatra/base.rb:1105:in 'process_route'
     # /usr/local/bundle/gems/sinatra-4.1.1/lib/sinatra/base.rb:1072:in 'block in route!'
     # /usr/local/bundle/gems/sinatra-4.1.1/lib/sinatra/base.rb:1069:in 'route!'
[...]

PR #8922, opened 2025-07-21
Don't clear callInfo for Ruby methods

headius:
In https://github.com/jruby/jruby/issues/8918 we attempted to fix an issue where the keyword arguments `callInfo` flags were getting stuck passing through a native method that did not formally accept keyword arguments. Unfortunately that change incorrectly included Ruby-based `ruby2_keywords` methods in the set of methods considered to not accept formal keywords and require `callInfo` clearing. This led to #8920 where a `ruby2_keywords` method fails to properly forward incoming keyword arguments to a non-`ruby2_keywords` method.

The fix here narrows the `callInfo` clearing to only native methods that do not formally accept keywords, preventing such flags from propagating to other calls but avoiding the `callInfo` clearing previously done for `ruby2_keywords` Ruby-based methods.

### DIFF
--- a/core/src/main/java/org/jruby/ir/targets/indy/InvokeSite.java
+++ b/core/src/main/java/org/jruby/ir/targets/indy/InvokeSite.java
@@ -832,12 +832,12 @@ public abstract class InvokeSite extends MutableCallSite {
         SmartBinder baseBinder = SmartBinder.from(signature.changeReturn(void.class)).permute("context");
 
         // if target method takes keywords and we are passing them, set callInfo
-        boolean acceptsKeywords = true;
+        boolean acceptsKeywords;
         DynamicMethod method = entry.method;
 
-        if (method instanceof AbstractIRMethod irMethod && irMethod.getRuby2Keywords()) {
-            // Ruby methods with ruby2_keywords don't use formal keywords
-            acceptsKeywords = false;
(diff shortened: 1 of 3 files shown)


floehopper:
I haven't had time to look into it, but I've seen a possible regression in v10.0.2.0 which sounds like it might be related to the  change in this PR. See [this Mocha CI build failure](https://app.circleci.com/pipelines/github/freerange/mocha/886/workflows/5086ccc0-7856-4031-ac38-2e11b3b94fe7/jobs/13337).

floehopper:
@headius I've done a bit of investigation in https://github.com/freerange/mocha/issues/760. It seems as if the problem(s) may have existed since v10.0.1.0, but may only be surfaced when the tests are run in a particular order, so the changes in this PR might be a red herring. Obviously it's possible there is a flaw in the Mocha tests, but I note that the same test failures do not occur when running against CRuby v3.4.5 or v3.5.0preview1.

headius:
@floehopper If we do not behave the same way as CRuby 3.4, then it's a bug in JRuby... so I'm glad you have reported this and done some extra homework!

Please open a new JRuby issue with everything you know so far and try to provide some way we can reproduce it.

floehopper:
@headius I’ve created https://github.com/jruby/jruby/issues/8976

Issue #8933, opened 2025-07-31
Output buffer size calculation for unpack('m') and unpack('u') can overflow for long inputs


ikaronen-relex:
Evaluate any of the following expressions:

1. `('X' * (2**31 / 3 + 98)).unpack('m').first.length`
2. `('X' * (2**31 / 3 + 98)).unpack('u').first.length`
3. `('X' * (2**32 / 3 + 99)).unpack('m').first.length`
4. `('X' * (2**32 / 3 + 99)).unpack('u').first.length`

**Expected Behavior**

Expressions 1 and 2 evaluate to 536870985 and expressions 3 and 4 to 1073741898 respectively.

**Actual Behavior**

Expression 1 crashes with:

```
org.jruby.util.Pack.unpack_m(Pack.java:1258): -536870839 (Java::JavaLang::NegativeArraySizeException)
	from org.jruby.util.Pack.unpackInternal(Pack.java:1083)
	from org.jruby.util.Pack.unpackWithBlock(Pack.java:870)
	from org.jruby.RubyString.unpack(RubyString.java:6605)
	from org.jruby.RubyString$INVOKER$i$unpack.call(RubyString$INVOKER$i$unpack.gen)
	from org.jruby.internal.runtime.methods.JavaMethod$JavaMethodOneOrNBlock.call(JavaMethod.java:423)
        ...
```

Expression 2 crashes with:

```
org.jruby.util.Pack.unpack_u(Pack.java:1392): -536870839 (Java::JavaLang::NegativeArraySizeException)
	from org.jruby.util.Pack.unpackInternal(Pack.java:1080)
	from org.jruby.util.Pack.unpackWithBlock(Pack.java:870)
	from org.jruby.RubyString.unpack(RubyString.java:6605)
	from org.jruby.RubyString$INVOKER$i$unpack.call(RubyString$INVOKER$i$unpack.gen)
	from org.jruby.internal.runtime.methods.JavaMethod$JavaMethodOneOrNBlock.call(JavaMethod.java:423)
```

Expression 3 crashes with:

```
org.jruby.util.Pack.unpack_m_nonzeroOccurrences(Pack.java:1320): Index 74 out of bounds for length 74 (Java::JavaLang::ArrayIndexOutOfBoundsException)
	from org.jruby.util.Pack.unpack_m(Pack.java:1266)
	from org.jruby.util.Pack.unpackInternal(Pack.java:1083)
[...]

PR #8936, opened 2025-08-01
Fix integer overflow in buffer length calculation for unpack('m') and unpack('u')

ikaronen-relex:
This fixes https://github.com/jruby/jruby/issues/8933

### DIFF
--- a/core/src/main/java/org/jruby/util/Pack.java
+++ b/core/src/main/java/org/jruby/util/Pack.java
@@ -75,7 +75,7 @@ public class Pack {
      **/
     private static final String NATIVE_CODES = "sSiIlLjJ";
     private static final String MAPPED_CODES = "sSiIqQjJ";
-    
+
     private static final char BE = '>' + 127; // 189, bumped up to avoid collisions with LE
     private static final char LE = '<'; // 60
     private static final String ENDIANESS_CODES = new String(new char[] {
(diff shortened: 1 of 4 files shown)


enebo:
@ikaronen-relex This is still marked WIP and it targets master which is JRuby 10.x.  If you are not finished then please consider retargeting to jruby-9.4 (and we/I will merge it to 10 as well).  If you are finished let me know and I will merge.  I can also cherry-pick this back so just let me know when you are done.

ikaronen-relex:
@enebo I marked it as draft since I was getting some test failures due to OOM errors. Looks like the tests pass now, so I've marked it as ready.

In the process, I also fixed what appears to be an unnecessary array copy in Pack.unpackBase46Strict. That didn't fix the OOMs (but apparently splitting the specs into several smaller ones did), but I left it in since I think it's still a reasonable (and safe, AFAICT!) optimization.

core/src/main/java/org/jruby/util/Pack.java
```diff
@@ -935,7 +935,7 @@ private static RubyString unpackBase46Strict(ThreadContext context, ByteList inp
                 out[index++] = (byte)((b << 4 | c >> 2) & 255);
             }
         }
-        return newString(context, new ByteList(out, 0, index));
+        return newString(context, new ByteList(out, 0, index, false));
```
ikaronen-relex, line 938: Passing `false` here avoids an unnecessary array copy in the ByteList constructor. The `out` array isn't otherwise exposed outside this method, so avoiding the copy seems safe to me.

enebo:
@ikaronen-relex Yeah I noticed you caught the extra arraycopy.  Good find.

PR #8940, opened 2025-08-05
Verify to_s is string or fall back on anyToString

headius:
Logic for "AsString" in a dynamic string must verify the result of calling to_s is a String, or else use anyToString on the original object. Previous logic here only called to_s and did not verify the result, proceeding to append non-string values like Integer or Float rather than falling back. The patch here adds an additional guard around the to_s call to perform this check and fallback.

This only affects the rare case where someone has implemented a to_s method that does not return a String, and is only observable once the dstring has been optimized and invoked twice.

As this is a bit invasive, I'm marking it for 10.0.3.0 so we can create more specs and do more verification that it works correctly before release.

### DIFF
--- a/core/src/main/java/org/jruby/RubyBasicObject.java
+++ b/core/src/main/java/org/jruby/RubyBasicObject.java
@@ -30,6 +30,7 @@ package org.jruby;
 
 import org.jcodings.Encoding;
 import org.jruby.anno.JRubyMethod;
+import org.jruby.api.Convert;
 import org.jruby.api.Create;
 import org.jruby.api.JRubyAPI;
 import org.jruby.ast.util.ArgsUtil;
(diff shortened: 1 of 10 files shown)


headius:
There's no spec specifically for this case, but without the patch it fails a hand-written example designed to lazily JIT:

```
$ jruby -Xjit.threshold=0 -e 'eval "1"; obj = Object.new; class << obj; def to_s; 42; end; private :to_s; end; 3.times { puts "#{obj}" }'
#<Object:0x640c216b>
42
42
```

The closest relevant spec is here, but it only ensures the result is a string:

https://github.com/jruby/jruby/blob/5465a89ecd684a705478a7f7a718c002ff6a5134/spec/ruby/language/string_spec.rb#L136-L147

I will add a more specific spec that ensures non-string to_s results are not included in the dynamic string.

headius:
@eregon The existing spec for this hardly verified anything, so I replaced it with one that confirms a non-String to_s result is discarded when building a dynamic string. Let me know if you have thoughts.

headius:
@eregon A thought: perhaps better to `!~ /42/` rather than match the hashy inspect output? I am generally reluctant to specify "not match" behaviors because they can be easily defeated by other bugs.

eregon:
The changes to spec/ruby/language/string_spec.rb look fine. I agree, it's clearer and safer to test that it looks like a usual inspect rather than check something is not there (which would pass e.g. if empty).

PR #8947, opened 2025-08-07
New check for System.console tty on JDK 22+

### LINKED ISSUE #8934: JRuby under Java 22+ always thinks it's running under a TTY
**Environment Information**
  
* JRuby version (`jruby -v`):
`jruby 10.0.0.0-SNAPSHOT (3.4.2) 2025-07-31 55a0013630 OpenJDK 64-Bit Server VM 24.0.1 on 24.0.1 +indy +jit [arm64-darwin]`
* Operating system: `Darwin`

**Bug description**

JRuby uses `System.console() != null` internally to determine whether it's printing to a TTY. On Java 22+, `System.console()` always returns an object, so `System.console().isTerminal()` has to be used instead. As it is now, this causes JRuby to print ANSI color codes to non-terminals, e.g. when using `-Xir.print`.

headius:
This uses the new `isTerminal` method when running on JDK 22+ due to that version now always returning a non-null `Console` from `System.console()`.

I may modify the handling of these color-configuring properties such that "true" means "use color if we are outputting to a terminal", defaulting to true.

Fixes #8934.

### DIFF
--- a/core/src/main/java/org/jruby/util/cli/Options.java
+++ b/core/src/main/java/org/jruby/util/cli/Options.java
@@ -29,6 +29,8 @@
 
 package org.jruby.util.cli;
 
+import java.io.Console;
+import java.lang.reflect.InvocationTargetException;
 import java.util.ArrayList;
 import java.util.Collection;
 import java.util.Collections;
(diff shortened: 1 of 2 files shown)


headius:
I added a very basic test that confirms a backtick subprocess does not default to using color. I'm not sure how to guarantee a test for a real TTY will trigger color... probably needs some pty magic I don't know.

We'll go with the simple negative test for now.

PR #8949, opened 2025-08-07
Broaden closure to method coercion with included blocks

enebo:
This allows define_method to be converted to a method with nested blocks so long as those nested blocks do not get captures from outside define_method.

As it stands this hurts performance:
https://gist.github.com/enebo/7256fd51d3f0ebb49d433f2bc3e77d02

It appears that something is not optimizing correctly for these converted blocks.  `nested capture def` should be the performance for `nested capture` and for sure should be faster than just invoking `nested capture` with the optimization disabled.


headius:
Seems like this needs more analysis to not be a performance hit. Bumping.

PR #8954, opened 2025-08-08
Secuity: Bump cgi to 0.4.2. CVE-2025-27220 CVE-2025-27219

ylecuyer:
Bump cgi version to fix CVE-2025-27220 and CVE-2025-27219

Screenshot_2025-08-08_11-17-42

### DIFF
--- a/BUILDING.md
+++ b/BUILDING.md
@@ -27,7 +27,7 @@ This will run the default "install" goal (`mvn install`) and will do all of the 
 
 * Compile JRuby
 * Build `lib/jruby.jar`, needed for running at command line
-* It will install the default gems specifications `lib/ruby/gems/shared/specifications/default/` and the ruby files of those gems in `lib/ruby/stdlib/`.
+* It will install the default gems specifications `lib/pom.rb` and the ruby files of those gems in `lib/ruby/stdlib/`.
 
 The environment is now suitable for running Ruby applications.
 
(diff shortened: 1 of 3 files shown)


headius:
Thank you! I'll cherry-pick this to 9.4 as well.

ylecuyer:
Hello, I see jruby 9.4.14.0 has been released with the cgi bump, but I can't see the 10.x release with the fix. What is the expected timeline for the 10.x branch ?

headius:
@ylecuyer For CVEs that are part of the standard library, you can simply upgrade your local copy of that gem (`gem install cgi` or add newer version to `Gemfile`, etc). We have not typically done special releases just for stdlib CVEs because of the upgrade path.

If you have a need for a "clean" release of 10.x on a shorter timeframe, get in touch with me directly and we can talk about options: headius@headius.com.

ylecuyer:
You know how security scans work, they check the whole container and flag if they find the vulnerable version (be it used or not) and cgi being a stdlib you can’t uninstall the vulnerable version afaict :(

headius:
@ylecuyer Ahh yes, I understand the problem. We will try to prioritize a JRuby 10 update release soon.

Meanwhile, perhaps you could add an entry to https://github.com/jruby/jruby/blob/master/USERS.md via a PR? We would like to do a better job of tracking who is using JRuby and how we can better serve you!

PR #8957, opened 2025-08-11
Add a USERS.md file to track JRuby users

headius:
Several open source projects have a list of users or use cases in their repository, so it's easily accessible and included in the source or binary distributions. This is an initial version of that list based on one from the Cilium project: https://github.com/cilium/cilium/blob/main/USERS.md

TODO:

* [x] Include in source distribution
* [x] Include in binary distribution
* [x] Revise and improve text of this form

Once it is in place, we can send a blast out on social media for current users to add themselves.


headius:
I think we'll skip historical users for now and leave that for another document or wiki page.

PR #8965, opened 2025-08-19
Pattern improvements

headius:
Work to improve compilation of pattern matching, with an eye to reduce instruction count and bytecode size.


headius:
Something not quite right in the typeError change... will revisit.

headius:
This is largely performance work so moving to 10.1.

PR #8984, opened 2025-08-29
glob refactoring

enebo:


core/src/main/java/org/jruby/api/Convert.java
```diff
@@ -593,6 +596,18 @@ public static RubyInteger toInteger(ThreadContext context, IRubyObject arg) {
         return arg.convertToInteger();
     }
 
+    /**
+     * Create a Java String from a ByteList with the specified encoding.
+     * @param bytes to be made into a string
+     * @param encoding use encodings charset to make the String
+     * @return a new Java String
+     */
+    public static String asJavaString(ByteList bytes, Encoding encoding) {
```
headius, line 605: ByteList has Encoding, why is a second encoding passed in here?

core/src/main/java/org/jruby/util/Dir.java
```diff
@@ -812,228 +787,161 @@ private static int addToResultIfExists(Ruby runtime, String cwd, byte[] bytes, i
...
-                            int bufBegin = buf.begin();
-                            int bufLen = buf.length();
-                            if (scheme != null) {
-                                bufBytes = prependScheme(scheme, bufBytes, bufBegin, bufBegin + bufLen);
-                                bufBegin = 0;
-                                bufLen = bufBytes.length;
-                            }
-                            resource = JRubyFile.createResource(runtime, cwd, new String(bufBytes, bufBegin, bufLen, enc.getCharset()));
-                            if ( !resource.isSymLink() && resource.isDirectory() && !".".equals(file) && !"..".equals(file) ) {
+                            if ( fnmatch(STAR, 0, 1, fileBytes, 0, fileBytes.length, flags) != 0) continue;
```
headius, line 877: Might as well fix extra spaces anywhere you are modifying lines...

headius (changes requested):
Minor changes but looks good.

core/src/main/java/org/jruby/api/Convert.java
enebo: As it turns out none of the callers have a different encoding so I just make this a single param method.

PR #8988, opened 2025-08-31
Add AsciidoctorJ to USERS.md

robertpanzer:
AsciidoctorJ heavily relies on JRuby to work.
We want to mention this in the USERS file.


headius:
Thank you!

Issue #8987, opened 2025-08-31
InvokeDynamicSupport.findVirtual fails with IllegalAccessException: symbolic reference class is not accessible


rsim:
**Environment Information**
  
* `jruby 10.0.2.0 (3.4.2) 2025-08-07 cba6031bd0 OpenJDK 64-Bit Server VM 21.0.8+9-LTS on 21.0.8+9-LTS +indy +jit [arm64-darwin]`
* Both on local macOS 15.5 and in GitHub CI actions

**Expected Behavior**

https://github.com/rsim/mondrian-olap/ tests are running successfully on JRuby 9.4.x and Java 8, 11, 17, 21.

**Actual Behavior**

When trying to run https://github.com/rsim/mondrian-olap/ tests on JRuby 10.0.2.0 (as well as with previous 10.0.0.0 and 10.0.1.0 versions) and Java 21 then the following line https://github.com/rsim/mondrian-olap/blob/master/lib/mondrian/olap/result.rb#L277 fails with
```
     Failure/Error: Unable to find org.jruby.dist/org.jruby.runtime.invokedynamic.InvokeDynamicSupport.findVirtual(InvokeDynamicSupport.java to read failed line

     Java::JavaLang::RuntimeException:
       java.lang.IllegalAccessException: symbolic reference class is not accessible: class jdk.proxy4.$Proxy49, from class org.jruby.runtime.invokedynamic.InvokeDynamicSupport (module org.jruby.dist)
     # org.jruby.dist/org.jruby.runtime.invokedynamic.InvokeDynamicSupport.findVirtual(InvokeDynamicSupport.java:85)
     # org.jruby.dist/org.jruby.ir.targets.indy.JavaBootstrap.createJavaHandle(JavaBootstrap.java:104)
     # org.jruby.dist/org.jruby.ir.targets.indy.InvokeSite.buildNativeHandle(InvokeSite.java:441)
     # org.jruby.dist/org.jruby.ir.targets.indy.InvokeSite.getHandle(InvokeSite.java:1153)
     # org.jruby.dist/org.jruby.ir.targets.indy.InvokeSite.invoke(InvokeSite.java:781)
     # home.runner.work.mondrian_minus_olap.mondrian_minus_olap.lib.mondrian.olap.result.️❤ def metadata #0(/home/runner/work/mondrian-olap/mondrian-olap/lib/mondrian/olap/result.rb:277)
```

[...]

PR #8989, opened 2025-09-02
Explicitly request access before Lookup.find*

headius:
Edit: a previous version of this patch attempted to switch these lookups to `publicLookup()`, but that broke several tests and internal logic that needed "more than public" access. The new patch was proposed by our friends on the core-libs-dev OpenJDK list.

The Lookup acquired from MethodHandles.lookup appears to produce
IllegalAccessException for cases that publicLookup does not. This
appears to be due to a JDK bug, but we can work around it by'
explicitly requesting read access to the target class's module.

There have been similar bugs relating to the use of
MethodHandles.lookup versus MethodHandles.publicLookup. The most
recent bug was filed in https://github.com/jruby/jruby/issues/8987. Related JRuby bugs and
an OpenJDK bug filed by me are also referenced from there. There's
also a link to an OpenJDK core-libs-dev thread about this issue.

The change here requests read access from the JRuby module to the
module of the target class, avoiding the accessibility issue for
classes that should otherwise be readable. There may be additional
overhead in acquiring the MethodHandle this way.

### DIFF
--- a/core/src/main/java/org/jruby/runtime/invokedynamic/InvokeDynamicSupport.java
+++ b/core/src/main/java/org/jruby/runtime/invokedynamic/InvokeDynamicSupport.java
@@ -28,6 +28,7 @@
 package org.jruby.runtime.invokedynamic;
 
 import java.lang.invoke.MethodHandle;
+import java.lang.invoke.MethodHandles;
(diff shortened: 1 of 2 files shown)


headius:
needs tests

PR #8992, opened 2025-09-04
phase out regression specs to jruby specs

enebo:


headius:
Nice, it will be good to eliminate a separate test suite.

DateTime failure looks like a missing require. Not sure about the EPERM vs EACCES failure though.

PR #9007, opened 2025-09-14
Handle `warn(foo, uplevel: nil`)

Earlopain:
Ruby simply ignores the nil uplevel (it is the default value), JRuby tries to convert it to an integer.

I wanted to write such code for a polyfill in https://github.com/ruby/prism/pull/3647 but had to change it around to make work on jruby.


headius:
Looks good! Thanks!

PR #9010, opened 2025-09-18
Fix "+00:00" and 0 being treated as UTC by Time#localtime and Time.at

### LINKED ISSUE #8998: Time#localtime incorrectly treats zero offset as UTC (regression in v10.0.2.0)
**Environment Information**

```
$ jruby -v
jruby 10.0.2.0 (3.4.2) 2025-08-07 cba6031bd0 OpenJDK 64-Bit Server VM 21.0.8+9-Ubuntu-0ubuntu122.04.1 on 21.0.8+9-Ubuntu-0ubuntu122.04.1 +indy +jit [x86_64-linux]

$ uname -a
Linux linuxdev 5.15.0-153-generic #163-Ubuntu SMP Thu Aug 7 16:37:18 UTC 2025 x86_64 x86_64 x86_64 GNU/Linux
```

**Expected Behaviour**

In MRI (version 3.4.5 and earlier), using `Time#localtime` to set the offset to 0 does not set the time instance to UTC (the offset returned in the `#to_s` string representation is '+0000' and `#utc?` returns `false`):

```
$ ruby -e 't = Time.at(1757243858).localtime(0); puts t; puts t.utc_offset; puts t.utc?'
2025-09-07 11:17:38 +0000
0
false
```

**Actual Behaviour**

JRuby version 10.0.2.0 sets the time instance to UTC (the offset returned in the `#to_s` string representation is 'UTC' instead of '+0000' and `#utc?` returns `true` instead of `false`):

```
$ jruby -e 't = Time.at(1757243858).localtime(0); puts t; puts t.utc_offset; puts t.utc?'
2025-09-07 11:17:38 UTC
0
true
```

This has changed in version 10.0.2.0. JRuby version 10.0.1.0 matches the MRI behaviour:

```
jruby -e 't = Time.at(1757243858).localtime(0); puts t; puts t.utc_offset; puts t.utc?'
2025-09-07 11:17:38 +0000
0
false
```

philr:
This PR changes `Time#localtime` and `Time.at`, so that it no longer treats a zone of `"+00:00"` or `0` as UTC (i.e. `Time#utc?` will return `false`). It resolves #8998.

I've included some additional test cases from ruby/spec#1285.

The fixes to both methods reset `isTzRelative` to `false` and then use `handleUTCDateTimeZone` to determine the time zone instead of `getTimeZoneFromUtcOffset`. `handleUTCDateTimeZone` calls `getTimeZoneFromUtcOffset` and contains the necessary checks to determine if the specified zone should be treated as UTC. It sets `isTzRelative` to `true` if not. The call to `adjustTimeZone` is then made with the newly set `isTzRelative` value.

### DIFF
--- a/core/src/main/java/org/jruby/RubyTime.java
+++ b/core/src/main/java/org/jruby/RubyTime.java
@@ -628,15 +628,14 @@ public class RubyTime extends RubyObject {
 
     @JRubyMethod(name = "localtime")
     public RubyTime localtime(ThreadContext context, IRubyObject arg) {
-        final DateTimeZone zone = getTimeZoneFromUtcOffset(context, arg);
+        setIsTzRelative(false);
+        final DateTimeZone zone = handleUTCDateTimeZone(context, arg);
 
         if (zone == null) {
             throw invalidUTCOffset(context);
-        } else if (zone == DateTimeZone.UTC) {
-            return gmtime(context);
         }
 
-        return adjustTimeZone(context, zone, true);
+        return adjustTimeZone(context, zone, isTzRelative);
     }
 
     private RubyTime adjustTimeZone(ThreadContext context, final DateTimeZone zone, boolean isTzRelative) {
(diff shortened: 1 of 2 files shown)


philr:
The test failure is unrelated to this change.

enebo:
@philr not sure if you tried running mspec without ci options on Time after this fix but there are other UTC/number failures which are tagged out (spec/tags/ruby/core/time/*).

Thanks for the fix.  Time and Date have been tough to address all failures.

enebo:
@philr Just tried and nothing else passed as a result of this which I guess makes sense.

headius:
@philr Nice, thank you for the patch! This turned out easier than expected.

philr:
@enebo thanks for merging this.

I looked through the tagged out tests expecting to see something related, but found it wasn't covered. I hadn't run without the ci option, but wasn't expecting anything else to start passing.

PR #9015, opened 2025-09-30
[test] get more junit test running with `mvn test`

kares:
while working on reproducing a regression I noticed not all JUnit test from under *core/src/test/java* are run 

with the `mvn -Ptest test` target, these changes add back most test cases to run and gets them green...

BEFORE: `Tests run: 440, Failures: 0, Errors: 0, Skipped: 0`
AFTER: `Tests run: 478, Failures: 0, Errors: 0, Skipped: 0`


headius:
Huh I wonder when those tests got excluded. Probably just missed adding them when the tests were reorganized a while back. Good catch!

PR #9023, opened 2025-10-08
Update specs

eregon:


headius:
Thanks Benoit!

headius:
FYI I fixed the typos in `arguments_spec.rb` so that should propagate back next time ("raies" to "raises").

PR #9025, opened 2025-10-08
Improvements for annotation framework use

headius:
Working on some demo applications for a JRuby Spring Boot meta-framework, I've had to deal with some challenges and inconsistencies in how annotations are applied. This PR aims to address those so simplify annotation-heavy use of JRuby's class reification subsystem.

See https://github.com/headius/boing for some of these examples.

* [x] Allow symbols for keys in annotation maps

This allows doing annotations like:

```
add_class_annotations(Entity => {name: "Employee"})
```

and similar for methods and parameters.

* [x] Modify boot sequence to delay requires until after initialization.
* [x] Modify boot sequence to avoid registering global runtime unless explicit.
* [ ] Improve naming and usability, with single-class forms and better use of Ruby 3.4 syntax)
* [ ] Test updates and additions

### DIFF
--- a/core/src/main/java/org/jruby/Ruby.java
+++ b/core/src/main/java/org/jruby/Ruby.java
@@ -802,14 +802,36 @@ public final class Ruby implements Constantizable {
     }
 
     /**
-     * Get the global runtime.
+     * Get the global runtime or initialize it if it has not been created.
      *
      * @return the global runtime
      */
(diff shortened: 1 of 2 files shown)


headius:
See https://github.com/headius/boing/pull/2 for related work on the other side (the new "boing" library).

PR #9027, opened 2025-10-09
Add "since" to bare Deprecated annotations

headius:
The "since" attributes added here reflect the first release after each deprecation was added. We will use this to remove oldest deprecations (by some measure of oldest) in an upcoming release.

The Ruby script used to make this change is provided here:

```ruby
Dir[ARGV[0] + "/**/*.java"].each do |file|
  file_lines = File.readlines(file)
  file_lines.each_with_index do |line, i|
    line.gsub!(/@Deprecated\S*$/) do
      sha = `git blame -L #{i+1},#{i+1} #{file}`.split(" ")[0]
      tag = `git describe --abbrev=0 --tags --contains #{sha}`.split("~")[0]
      "@Deprecated(since = \"#{tag}\")"
    end
  end
  File.write(file, file_lines.join(''))
  puts "process #{file}"
end
```

There may be places where sources moved or the deprecation line does not reflect the true earliest version, but worst case the deprecations stick around a little longer, or require manual investigation.

### DIFF
--- a/core/src/main/java/org/jruby/AbstractRubyMethod.java
+++ b/core/src/main/java/org/jruby/AbstractRubyMethod.java
@@ -88,7 +88,7 @@ public abstract class AbstractRubyMethod extends RubyObject implements DataType 
         return asFixnum(context, method.getSignature().arityValue());
     }
 
-    @Deprecated
+    @Deprecated(since = "10.0.0.0")
     public RubyFixnum arity() {
         return arity(getCurrentContext());
     }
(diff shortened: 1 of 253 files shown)


headius:
I made a few manual changes after the initial push of the automated version:

* jruby-openssl-* tags have been updated to the nearest JRuby release tag.
* All preview and RC tags have been updated to their associated non-prerelease versions.
* One bogus tag "list" was removed from the repo and all "since" lines updated to point at the nearest JRuby release tag.
* The "10.0" deprecations have been updated to "10.0.0.0" so they point at a full, valid tag and release version.
* The "9.4" deprecations added to methods that *should have* been deprecated in 9.4.x were modified to "10.0.0.0" because that is the first release in which they were *actually* deprecated.

PR #9029, opened 2025-10-22
Implement Ractor

headius:
Initial push is very preliminary, but can run the simple script shown below.

* Basic construction and usage are there.
* Port is there but it's just Queue (no shareability detection).
* Some types aren't covered by make_shareable.
* The given proc is checked for sanity.

Script:
```ruby
foo = +"bar"
hash = {foo: foo}
ary = [hash]

puts "shareable before: #{Ractor.shareable?(ary)}"
puts "frozen at all levels?"
puts ary.frozen?
puts ary[0].frozen?
puts ary[0][:foo].frozen?

shary = Ractor.make_shareable(ary)

ractor = Ractor.new(shary) do |a|
  puts "shareable inside: #{Ractor.shareable?(a)}"
  puts "frozen at all levels?"
  puts a.frozen?
  puts a[0].frozen?
  puts a[0][:foo].frozen?
  :ok
end

puts "final value: #{ractor.value}"
```


headius:
I've made a tweak here to move the definition of Ractor into an extension library (`require 'ractor'`).

The current implementation is partial, but because so many libraries use the presence of the `Ractor` constant to change behavior, they fail to boot up and function properly. Gating the `Ractor` constant behind a require avoids this issue but allows me and others to continue experimenting as the API gets filled out.

I will be proposing to ruby-core that Ractor be gated behind a require there as well.

headius:
Recent change adds `ractor` as a feature that can be enabled or disabled (default disabled for now). This breaks a spec that expects all features to be either enabled or disabled by `--all` or omission of that flag.

There's also an alternative implementation in pure Ruby on top of Thread and Queue here that could be shipped as our implementation of Ractor. Since the work here has primarily been to get `make_shareable` working and my implementation of `Ractor` and `Port` largely just use `Thread` and `Queue`, this would be a logical move (a shared implementation for all non-CRuby impls that already have parallel threads). cc @eregon

eregon:
> There's also an alternative implementation in pure Ruby on top of Thread and Queue here that could be shipped as our implementation of Ractor.

Right, that's https://github.com/eregon/ractor-shim and it implements all methods of the Ruby 4.0 Ractor API.

headius:
Land this or ship the shim library in 10.1. We need to have Ractor.make_shareable available, since projects are already moving to use it (see https://github.com/dry-rb/dry-struct/pull/203 for example).

Issue #8924, opened 2025-07-24
const_defined? returns false for External Constants in JRuby 10.0.1.0 and 9.4.14.0


yamam:
When a constant is defined outside of a class, calling const_defined? on that class returns false in JRuby 10.0.1.0. However, in JRuby 10.0.0.1 and CRuby 3.4.5, it returns true.

**Environment Information**

jruby 10.0.1.0 (3.4.2) 2025-07-17 0f10d1dfdf OpenJDK 64-Bit Server VM 21.0.7+6-Ubuntu-0ubuntu124.04 on 21.0.7+6-Ubuntu-0ubuntu124.04 +indy +jit [x86_64-linux]

**Expected Behavior**
```ruby
$ jruby -e 'A=0;p Class.new.const_defined?("A")'                                                                                                                                       
true
```

**Actual Behavior**
```ruby
$ jruby -e 'A=0;p Class.new.const_defined?("A")'                                                                                                                                       
false
```

yamam:
It did not occur up to version 9.4.13.0, but we have confirmed that it also occurs in version 9.4.14.0.

yamam:
As a result of using git bisect, I identified 75ce19e3a9 as the cause.

headius:
Thank you for the report! I missed this one.

headius:
I have a fix in progress in #9031. Could you submit a PR here (in spec/ruby) or at https://github.com/ruby/spec to test this behavior?

PR #9031, opened 2025-10-23
Don't restrict object search to modules

headius:
I'm unsure why this change is here but I could find no equivalent logic in CRuby and this appears to prevent the search detailed in jruby/jruby#8924 from succeeding. Removing this restriction fixes that issue.

Fixes jruby/jruby#8924


headius:
I traced this particular `!isClass` check back as far as 2008. Could be behavior that changed since then in Ruby proper.

headius:
An additional patch here limits the Object search to only modules and classes that descend from Object, excluding BasicObject descendants. Without this, specs for `defined? Kernel` within a BasicObject descendant failed (Kernel should not be found).

https://github.com/jruby/jruby/blob/ff9277f5d5a5a367da94f5078c69e0284f67d002/spec/ruby/core/basicobject/basicobject_spec.rb#L18-L20

PR #9041, opened 2025-10-24
Migrate .java-version contents to .jruby.release

headius:
I think it makes more sense to use .jruby.release as the only file for version-specific details. This change may need to be tweaked to support JRuby 10 releases prior to 10.0.3.0 (assuming it is merged into that release), or it may not be important since we won't be changing the jruby.sh in any released artifacts.


headius:
@enebo @mrnoname1000 I've been doing some cleanup and realized we don't really need both `.jruby.release` and `.java-version` in `bin`. I think it makes sense to lean into the use of `.jruby.release` as the store of all version-specific values, since we already have it in place for the JRuby version and the main class to use.

I had some initial concerns about making this change, since it will not work properly if copied into older 10.x releases that have a `.java-version` file, but perhaps I'm worrying about nothing? We won't be replacing the jruby.sh in those releases now that they're out, and we don't support installing updated jruby.sh from jruby-launcher yet.

Of course, if jruby-launcher *does* gain the ability to update jruby.sh, then we will want to make sure it doesn't break on old releases. Cross that bridge when we come to it?

PR #9043, opened 2025-10-28
Update to ruby/spec@3bc45ba

andrykonchin:


headius:
Thank you @andrykonchin!

Issue #9050, opened 2025-10-31
Marshal fails to load an encoding with an encoding ivar


headius:
I'm not sure how this happens, but in the ri `cache.ri` dump file from CRuby 3.4 (see comments in #9049) there's an Encoding object that has a single encoding ivar. When attempting to read it, the following error results:

```
ArgumentError: UTF-8is not enc_capable
    load at org/jruby/RubyMarshal.java:165
  <main> at -e:1
```

That encoding looks like this in the binary (starting at the `Iu:` sequence):

```
...
00224e0 002 214 036   {  \0   @ 002 223 036   {  \0   :  \r   e   n   c
00224f0   o   d   i   n   g   I   u   :  \r   E   n   c   o   d   i   n
0022500   g  \n   U   T   F   -   8 006   ; 006   F   : 025   i   n   s
0022510   t   a   n   c   e   _   m   e   t   h   o   d   s   { 002   U
...
```

The second `006` there is a reference to the symbol `:E`, and is followed by a character `F`, which combined makes a simple way for Ruby to specify a US-ASCII encoding instance variable. Extracting this to a reproducible example we get:

```
[] jruby $ jruby -e 'Marshal.load("\004\bIu:\rEncoding\nUTF-8\006:\006EF")'
ArgumentError: UTF-8is not enc_capable
    load at org/jruby/RubyMarshal.java:165
  <main> at -e:1
[] jruby $ cx ruby-3.4 ruby -e 'p Marshal.load("\004\bIu:\rEncoding\nUTF-8\006:\006EF")'
#<Encoding:UTF-8>
```

I'm not sure what would cause a dump of an Encoding to have an encoding instance variable, but it seems like this is the case in the `cache.ri` both for this dump and for my local install of Ruby 3.4.7:

```
$ ruby -e 'Marshal.load(File.read("/Users/headius/.rubies/ruby-3.4.7/share/ri/3.4.0/system/cache.ri"))'
ArgumentError: UTF-8is not enc_capable
    load at org/jruby/RubyMarshal.java:165
  <main> at -e:1
```

headius:
Indeed, CRuby appears to dump Encoding with this odd US-ASCII instance variable:

```
$ cx ruby-3.4 ruby -e 'pp Marshal.dump(Encoding::UTF_8)'
"\x04\bIu:\rEncoding\n" + "UTF-8\x06:\x06EF"
```

We do not:

```
[...]

PR #9051, opened 2025-10-31
Ignore encoding when unmarshaling an Encoding

headius:
CRuby appears to marshal all Encoding objects with a special encoding instance variable indicating they have an encoding of US-ASCII, even though Encoding does not actually have an encoding. I've filed an issue about this for clarification below, but for now we ignore this peculiar situation.

https://bugs.ruby-lang.org/issues/21658

Fixes jruby/jruby#9050


headius:
Checked into TruffleRuby impl and they basically ignore encoding instance vars for object types that don't have encoding. CRuby rejects such classes outright, but has an exception for `T_DATA` that is an encoding object, so basically what we have here.

Issue #9049, opened 2025-10-30
ri docs are not available for core


astra1993:
Hello. I recently installed JRuby and tried to use the `ri` tool to check some documentation but it was missing. I ran `jruby -S ri --list-doc-dirs`, which gave me the following output:
```
/usr/share/jruby/share/ri/2.6.0/system
/usr/share/jruby/share/ri/2.6.0/site
/home/astra/.local/share/rdoc
```
There was no `share` folder under `/usr/share/jruby` at all. This is a system installation on Debian. I thought I'd missed the doc package, but there is no `jruby-doc` package. Are there any plans to ship the `ri` documentation for the core?

headius:
Thanks for opening the issue! This is something we have considered in the past but never quite got to the point of including anything.

We certainly could generate documentation during the release build, but most of the core class documentation lives in the original C code. We don't host that content because it would be a nightmare to keep in sync.

The simplest answer would probably be to download pre-generated documentation from a build of standard Ruby, plus maybe add in a little bit for some of the jruby specific gems and libraries we ship.

headius:
@astra1993 maybe you could research a bit and see if there's a canonical place we can download a full set of docs already generated?

astra1993:
There are precompiled documentation available [here](https://ruby-doc.org/downloads/), but it's only the HTML output and not the `ri` database. And I know that Debian (and probably other distros) have a precompiled doc package (like `ruby3.1-doc` on Debian 12). But the most surefire way to get the documentation for the exact version you want, is to generate them from the CRuby source. The more interesting issue is adding the JRuby-specific documentation to the CRuby source, maybe as a group of blank module files with documentation attached to them, and then doing `rdoc --all --ri --op output/` for the generation.
I also think that it'd be easier to cull the `ri` documentation after generation (as I'm seeing that some classes are not shared between the two implementations like RubyVM and Ractor).
Also a very important note: If the Marshal implementation is different between CRuby and JRuby, use JRuby's rdoc to create the documentation, or else JRuby ri will most likely fail to load them (just tested this and I can confirm that JRuby ri fails to load the ri docs generated with CRuby rdoc).

headius:
I think we should look at the doc packages from Debian for now and it can be a longer-term project to generate everything at release time (including JRuby APIs).

headius:
Looks like most recent docs are available in Debian experimental as ruby3.4-doc (and the 3.1 docs are in the stable ruby3.1-doc package you mentioned.

[...]

PR #9052, opened 2025-10-31
Download and include in dist CRuby's rdoc data

headius:
This change does the following:

* Download the CRuby rdoc data published by Debian.
* Unpack and include that data in the build of jruby-dist.

This logic uses antrun and is hardcoded to the specific URL of the ruby3.4-doc package from Debian. It depends on the "ar" and "tar" commands being available on the host system.

Fixes jruby/jruby#9049

### DIFF
--- a/maven/jruby-dist/src/main/assembly/common.xml
+++ b/maven/jruby-dist/src/main/assembly/common.xml
@@ -50,6 +50,42 @@
       <includes>
         <include>share/ri/3.4.0/system/**/*</include>
       </includes>
+      <excludes>
+        <exclude>share/ri/3.4.0/system/contributing/**/*</exclude>
+        <exclude>share/ri/3.4.0/system/contributing</exclude>
+        <exclude>share/ri/3.4.0/system/CoreExtensions/**/*</exclude>
+        <exclude>share/ri/3.4.0/system/CoreExtensions</exclude>
+        <exclude>share/ri/3.4.0/system/fatal/**/*</exclude>
+        <exclude>share/ri/3.4.0/system/fatal</exclude>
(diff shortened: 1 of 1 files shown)


headius:
Note this obviously adds some size to the dist tarball (and a similar amount to the zip):

```
$ ls -l maven/jruby-dist/target/jruby-dist-10.0.3.0-SNAPSHOT-bin.tar.gz
-rw-r--r--  1 headius  staff  36726626 Oct 31 12:23 maven/jruby-dist/target/jruby-dist-10.0.3.0-SNAPSHOT-bin.tar.gz
```
vs
```
$ ls -l maven/jruby-dist/target/jruby-dist-10.0.3.0-SNAPSHOT-bin.tar.gz         
-rw-r--r--  1 headius  staff  34066463 Oct 31 12:29 maven/jruby-dist/target/jruby-dist-10.0.3.0-SNAPSHOT-bin.tar.gz
```

So that's like 2.6MB or so.

This PR also requires #9051 in order for the dumped RI data to be readable by JRuby.

headius:
Updated version uses FTP to download the sole "doc" file from Debian's Ruby 3.4 artifact directory, since this filename may change over time.

It would be nice to remove the dependency on the `ar` and `tar` commands, but there does not appear to be any support for `ar` in any Ant tasks and both unpack commands would require additional Ruby dependencies not published with the jruby-maven-plugin.

mrnoname1000:
`ar` is POSIX, `tar` is ubiquitous, and `tar -C` is available on all the implementations I checked. So those dependencies shouldn't be an issue outside of hyper minimalist systems

headius:
Last change is pushed, to remove CRuby-specific bits of the ri data.

PR #9055, opened 2025-11-04
Array experiments

headius:
This PR will encompass some experiments with the implementation of Array:

* Abstract RubyArray supertype to make it easier to support different backend implementations.
* Concurrency-safe RubyArray backends that can efficiently handle multi-threaded mutation without raising errors.
* Efficient specialized RubyArray backends for fixed sizes, immutability, or specialized data types (e.g. long[] for a homogeneous Array of Fixnums).

### DIFF
--- a/core/src/main/java/org/jruby/Ruby.java
+++ b/core/src/main/java/org/jruby/Ruby.java
@@ -1618,7 +1618,7 @@ public final class Ruby implements Constantizable {
         return null;
     }
 
-    public static final int NIL_PREFILLED_ARRAY_SIZE = RubyArray.ARRAY_DEFAULT_SIZE * 8;
+    public static final int NIL_PREFILLED_ARRAY_SIZE = RubyArrayNative.ARRAY_DEFAULT_SIZE * 8;
     private final IRubyObject nilPrefilledArray[];
 
     public IRubyObject[] getNilPrefilledArray() {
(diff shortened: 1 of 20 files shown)


headius:
The abstract work is proceeding; I have been force-pushing as I find broken ends of my refactoring.

I have tried to preserve as much of the public API of `RubyArray` as possible, even leaving most of the deprecated methods in place (they are apparently used by many of our extensions!) One area that is impossible to maintain is constructors, since they cannot be callable against an abstract class.

I believe only the following forms are public:

* `(Ruby, int)`
* `(Ruby, IRubyObject[], int, int)`
* `(Ruby, RubyClass, IRubyObject[])`
* `(Ruby, RubyClass, IRubyObject[], int, int, boolean)`

These are basically in order of risk, with the sized `(Ruby, int)` form being the most likely to be used in the wild.

The risk here is small, because these are peculiar forms and we have generally discouraged direct construction of core classes, but if some code out there *does* use these constructors there's no way to support that usage. The external code would have to be updated to use one of our factory methods.

Other risks largely revolve around any code that expects `getClass()` to be `org.jruby.RubyArray`, as I found in a few of our own tests, but this is less likely to be found in the wild.

headius:
I believe I have the abstractification green now with a minimal diff (outside of RubyArray.java). The remaining diffs are largely internal cases where we work directly with the "native" implementation from other core classes, such as to construct an uninitialized array for immediate population. These could be replaced with alternative forms.

The risks are outlined in my comment above, and mostly revolve around the irreplaceable constructor forms.

headius:
After discussion with @enebo, we're going to go forward with this. JRuby 10.1 will no longer allow direct construction of `RubyArray` and users will have to move to one of the many factory methods in that class or in `org.jruby.api.Create`.

Full speed ahead!

Issue #9060, opened 2025-11-04
Adopt pure-Ruby Pathname


headius:
Pathname was once pure-Ruby, and then a bunch of it moved to C for reasons, and now it has moved back to Ruby via ruby/pathname#57.

It also has moved to being a built-in core class without any require.

We should trim down or eliminate our extension to match what CRuby still delivers and incorporate the new pure-Ruby Pathname into JRuby's core classes.

headius:
Removed milestone because Pathname moving to core and using this new mostly-Ruby version will not be in a release until Ruby 3.5.

PR #9062, opened 2025-11-05
Move Pathname to core [Ruby 4.0]

headius:
* RubyPathname is stripped down to match the remaining extension in CRuby 4.0 (<=> and sub).
* RubyPathname.path field is bound to @path instance variable.
* pathname.rb comes from ruby/pathname, with undef preamble, extension load, and a few utility methods removed.
* stdlib pathname.rb contains those utility methods.

See ruby/pathname#57 and ruby/pathname#17

Fixes jruby/jruby#9060


headius:
This will require some additional work to support our URI path forms. We may need to continue overriding more of the pure-Ruby logic than I'd hoped.

headius:
Because of the missing URI support, I've temporarily just internalized the existing implementation we already had.

PR #9066, opened 2025-11-05
Move Set to a core class [Ruby 4.0]

headius:
Ruby 4.0 will include Set as a core class, minus some old encumbrances. This PR will do the same for JRuby.

TODO:

* [x] Move existing Set to core
* [x] Delete SortedSet
* [x] Pass new CRuby tests for some updated behaviors.
* [x] Eliminate the `@hash` instance variable
* [ ] The implementation is an efficient, dedicated data structure.
* [x] Pass all remaining test suites.

### DIFF
--- a/test/jruby/test_set.rb
+++ b/test/jruby/test_set.rb
@@ -4,20 +4,6 @@ require 'set.rb'
 # JRuby's Set impl specific or low-level details.
 class TestSet < Test::Unit::TestCase
 
-  class SubSet < Set ; end
-
-  def test_sub_set
-    set = SubSet.new
-    assert_same SubSet, set.class
-    assert set.is_a?(SubSet)
-    assert set.is_a?(Set)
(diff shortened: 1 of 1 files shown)


headius:
Do not merge until we have decided what to do with Ruby 3.5 features.

headius:
This can merge into ruby-4.0 any time, but it would be good to finish the listed tasks (they just aren't critical for functionality).

headius:
Most of this is now implemented, except for the efficient storage format. That will require additional Java hacking to come up with a representation of Set internals that has the following characteristics:

* Preserves insertion order
* Can be switched to compare-by-identity (possibly by replacing the collection)
* More compact than RubyHash (chained bucket impl)
* Concurrency-safe (as part of a general move to make core types safe)

I'll file a separate issue to track that work and we'll hopefully get around to it at some point.

Issue #9061, opened 2025-11-05
Ruby 4.0 support


headius:
Ruby 4.0 will be released soon. This feature checklist is based on the [NEWS](https://github.com/ruby/ruby/blob/master/NEWS.md) file from CRuby's master branch. We'll update this list until the release has been finalized.

Pull requests to implement Ruby 4.0 features should be based on the `ruby-4.0` branch, not `master`.

---

# NEWS for Ruby 4.0

This document is a list of user-visible feature changes
since the **3.4.0** release, except for bug fixes.

Note that each entry is kept to a minimum, see links for details.

## Language changes

* [ ] `*nil` no longer calls `nil.to_a`, similar to how `**nil` does
  not call `nil.to_hash`.  [[Feature #21047]] (https://github.com/jruby/jruby/pull/9074)

* [x] Logical binary operators (`||`, `&&`, `and` and `or`) at the
  beginning of a line continue the previous line, like fluent dot.
  The following two code are equal:

    ```ruby
    if condition1
       && condition2
      ...
    end
    ```

    ```ruby
    if condition1 && condition2
      ...
    end
    ```

    [[Feature #20925]]

## Core classes updates

Note: We're only listing outstanding class updates.

* Array

    * [x] `Array#rfind` has been added as a more efficient alternative to `array.reverse_each.find` [[Feature #21678]]
    * [x] `Array#find` has been added as a more efficient override of `Enumerable#find` [[Feature #21678]]

* Binding

    * [x] `Binding#local_variables` does no longer include numbered parameters.
      Also, `Binding#local_variable_get`, `Binding#local_variable_set`, and
      `Binding#local_variable_defined?` reject to handle numbered parameters.
      [[Bug #21049]]

    * [x] `Binding#implicit_parameters`, `Binding#implicit_parameter_get`, and
      `Binding#implicit_parameter_defined?` have been added to access
      numbered parameters and "it" parameter. [[Bug #21049]]

* Enumerator

    * [x] `Enumerator.produce` now accepts an optional `size` keyword argument
      to specify the size of the enumerator.  It can be an integer,
      `Float::INFINITY`, a callable object (such as a lambda), or `nil` to
      indicate unknown size.  When not specified, the size defaults to
      `Float::INFINITY`.

        ```ruby
        # Infinite enumerator
        enum = Enumerator.produce(1, size: Float::INFINITY, &:succ)
        enum.size  # => Float::INFINITY

        # Finite enumerator with known/computable size
[...]

PR #9069, opened 2025-11-11
Implement Ruby 4.0 support

headius:
This PR will include changes on the ruby-4.0 branch for Ruby 4.0 support.

See #9061 for the feature checklist from CRuby's NEWS and other items.


headius:
Currently still marked as 3.5, since that's what specs and other updates expect currently. Will bump to 4.0 whenever it looks like that's officially propagated through the other libraries and tests.

headius:
Ruby 3.5 was officially promoted to 4.0 last week, so that's what it will be.

PR #9073, opened 2025-11-17
fix: Permit overriding of JSA file location using ENV variable

aleksandrs-ledovskis:
The construct introduced in #8708 to "Allow overriding default JSA file location" doesn't work. 

E.g. running `docker run --rm -e 'JRUBY_JSA=/tmp' jruby:10.0.2.0 ruby --version` fails with `/opt/jruby/bin/ruby: 814: /opt/jruby/bin/jruby.sh: jruby_jsa_file: is read only`.

The fix seems to be trivial.

### DIFF
--- a/bin/jruby.sh
+++ b/bin/jruby.sh
@@ -854,17 +854,18 @@ if $use_modules; then
     process_java_opts "$jruby_module_opts_file"
 fi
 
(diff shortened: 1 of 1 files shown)


headius:
Looks good to me. Waiting for review by @mrnoname1000.

headius:
Thank you!

PR #9081, opened 2025-11-19
Update specs

headius:


headius:
Do not merge this until we can confirm the spec updates in MRI.

headius:
Everything is green except some new CAPI specs that only seem to fail on MRI in CI. Still investigating.

eregon:
It's fine to merge this even if it isn't merged in CRuby yet.
The specs are already upstreamed in ruby/spec before anyway.

PR #9082, opened 2025-11-19
Ignore missing exts on JRuby

headius:
This problem does not seem to be going away, so ignore all missing extensions on JRuby for now.

See https://github.com/ruby/rubygems/issues/3520

### DIFF
--- a/lib/ruby/stdlib/rubygems/defaults/jruby.rb
+++ b/lib/ruby/stdlib/rubygems/defaults/jruby.rb
@@ -57,6 +57,14 @@ end
 
 ## JAR FILES: Allow gem path entries to contain jar files
 class Gem::Specification
+  ##
+  # Ignore missing extensions on JRuby
+  #
+  # See https://github.com/ruby/rubygems/issues/3520
+  def missing_extensions?
+    false
+  end
+
   class << self
     # Replace existing dirs
     def dirs


headius:
@deivid-rodriguez @eregon I'm going ahead with a PR to patch this method in JRuby. I'd prefer not to have to wipe out the extension detection, but it's too specific and usually erroneous in JRuby, which has very few extension ABI issues.

deivid-rodriguez:
Unfortunately I'm no longer involved in the development of RubyGems, so I can't help here. Please check out with the new maintainers team and I appreciate if you don't ping me directly with respect to RubyGems related issues (no big deal, but I'm no longer the right person to ping here).

Good luck!

eregon:
This sounds like the wrong fix to me, see https://github.com/ruby/rubygems/issues/3520#issuecomment-3556683078

eregon:
An example problem with this hack is if there is a gem which does something with `gemspec.extensions =` on JRuby then you would get a pretty late error vs a clearer error that somehow the extension didn't get built, e.g. when reusing a GEM_HOME from another JRuby version or so.
`gemspec.extensions =` is not just for C extensions, it can run an arbitrary Ruby script, and it is used like that for some gems.

headius:
Closing in favor of #9107.

PR #9093, opened 2025-11-21
Sync specs to ruby/spec on push to master

headius:
This PR will eventually set up a job to sync any new spec/ruby changes to the ruby/spec repository after they are pushed to master.

Work in progress.


headius:
Better to do as a branch on jruby/jruby for now.

PR #9094, opened 2025-11-22
Implement GC.config [Ruby 3.4]

trinistr:
This uses a trivial implementation that satisfies MRI's interface. Spec is here: https://github.com/ruby/spec/pull/1306

This PR is to familiarize myself with JRuby, so please, do not hesitate to point out how to do things correctly! A couple things I'm not sure about:
- ~~Is `"java"` string created correctly, considering non-default Encoding? This probably isn't actually needed, a default encoding should be fine, I only uses `newUSASCIIString` for full compatiblity.~~ I've decided to just use the default, there is probably no reason to complicate this.
- Is there a way to create an `ArgumentError` without specifying a message?

~~I haven't actually tested this code, and there don't seem to be any tests for `GC`, so there may need to be further actions on my part.~~ I've successfully built JRuby and tested this manually.

On a separate note, I believe that all "module = true" annotation options on `GC` are actually supposed to be "meta", as `GC` does not use `module_function`. And what does `visibility = PRIVATE` do here?

### DIFF
--- a/core/src/main/java/org/jruby/RubyGC.java
+++ b/core/src/main/java/org/jruby/RubyGC.java
@@ -67,6 +67,30 @@ public class RubyGC {
         return defineModule(context, "GC").defineMethods(context, RubyGC.class);
     }
 
+    @JRubyMethod(module = true, visibility = PRIVATE)
+    public static IRubyObject config(ThreadContext context, IRubyObject recv) {
+        return RubyHash.newHash(
+            context.runtime,
+            RubySymbol.newSymbol(context.runtime, "implementation"),
+            RubyString.newString(context.runtime, "java")
+        );
+    }
+
+    @JRubyMethod(module = true, visibility = PRIVATE)
+    public static IRubyObject config(ThreadContext context, IRubyObject recv, IRubyObject arg) {
+        if (arg.isNil()) {
+            return config(context, recv);
+        } else if (arg instanceof RubyHash) {
+            if (((RubyHash) arg).hasKey(RubySymbol.newSymbol(context.runtime, "implementation"))) {
+                throw context.runtime.newArgumentError("Attempting to set read-only key \"Implementation\"");
+            }
+            return config(context, recv);
+        } else {
+            throw context.runtime.newArgumentError("ArgumentError");
+        }
+    }
+    
+
     @JRubyMethod(module = true, visibility = PRIVATE, optional = 1, checkArity = false)
     public static IRubyObject start(ThreadContext context, IRubyObject recv, IRubyObject[] args) {
         Arity.checkArgumentCount(context, args, 0, 1);


headius:
Very cool! I will review as soon as I can given the US holiday. Thank you for your efforts!

core/src/main/java/org/jruby/RubyGC.java
```diff
@@ -67,6 +67,32 @@ public static RubyModule createGCModule(ThreadContext context) {
         return defineModule(context, "GC").defineMethods(context, RubyGC.class);
     }
 
+    @JRubyMethod(module = true, visibility = PRIVATE)
+    public static IRubyObject config(ThreadContext context, IRubyObject recv) {
+        return RubyHash.newHash(
+            context.runtime,
+            RubySymbol.newSymbol(context.runtime, "implementation"),
+            RubyString.newString(context.runtime, "java")
```
headius: It is possible to get the current GC on JVM but I'm not sure if this is always available (part of the management/monitoring subsystem (example in IRB):

```
irb(main):002> gc_beans = java.lang.management.ManagementFactory.garbage_collector_mx_beans
=> 
#<Java::JavaUtil::ArrayList: [#<Java::ComSunManagementInternal::GarbageCollectorExtImpl:0x6d2e8c15>, #<Java::ComSunManagementInternal::GarbageCollectorExtImpl:0x2ff4ad42>, #<Java::ComSunManagementInternal::...
irb(main):003> gc_beans.each {|bean| puts bean.name}
G1 Young Generation
G1 Concurrent GC
G1 Old Generation
=> #<Java::JavaUtil::ArrayList: [#<Java::ComSunManagementInternal::GarbageCollectorExtImpl:0x6d2e8c15>, #<Java::ComSunManagementInternal::GarbageCollectorExtImpl:0x2ff4ad42>, #<Java::ComSunManagementInternal::GarbageCollectorExtImpl:0x3d7cbcac>]>
irb(main):004> 
```

We also could just get the name of the JVM with system property "java.vm.name".

I'm not sure what's best here, since this is purely informational, but only provides a single string to represent the GC implementation.

core/src/main/java/org/jruby/RubyGC.java
```diff
@@ -67,6 +67,32 @@ public static RubyModule createGCModule(ThreadContext context) {
...
+        );
+    }
+
+    @JRubyMethod(module = true, visibility = PRIVATE)
+    public static IRubyObject config(ThreadContext context, IRubyObject recv, IRubyObject arg) {
+        if (arg.isNil()) {
+            return config(context, recv);
+        }
```
headius: Our style is `} else {` so there's a few places to clean up.

core/src/main/java/org/jruby/RubyGC.java
```diff
@@ -67,6 +67,32 @@ public static RubyModule createGCModule(ThreadContext context) {
...
+    @JRubyMethod(module = true, visibility = PRIVATE)
+    public static IRubyObject config(ThreadContext context, IRubyObject recv) {
+        return RubyHash.newHash(
+            context.runtime,
+            RubySymbol.newSymbol(context.runtime, "implementation"),
+            RubyString.newString(context.runtime, "java")
+        );
+    }
+
+    @JRubyMethod(module = true, visibility = PRIVATE)
```
headius: You are likely right about `meta = true` here. Probably cleaner to do it as a separate PR.

Issue #9088, opened 2025-11-21
Shrink header flags


headius:
Currently we define the `flags` field in `RubyBasicObject` as an int, probably because that was the simplest option at the time and we may have had memory alignment in mind, but in actuality the greated number of bits we currently use is about 10 (for module flags relating to includes and refinements.

Reducing the size of this field to a short would accommodate all of those values easily. Moving the additional flags for objects like Modules and Hashes could allow us to reduce the size of the field further to a byte. If we stop using flags to indicate `nil` and `false`, it could potentially be reduced to a single boolean for the object's frozen status, but future changes to Ruby (such as Ractor's "shareable" flag) would require adding back additional bits.

A reduction to short would immediately reduce the size of all objects by two bytes, and should be examine as a first pass at shrinking our objects. If feasible, moving more type-specific flags to booleans in those objects would allow reducing flags to a byte.

Memory savings could be enormous, similar to reducing small fixnum sizes in #9085.

PR #9095, opened 2025-11-25
Eliminate global flags

headius:
This is an experiment to remove or greatly reduce in size the global RubyBasicObject.flags field.

See #9088.


headius:
StringIO needs changes to support this, since it accesses and mutates RubyBasicObject.flags directly.

headius:
StringIO fixed in ruby/stringio#182 and pending release.

PR #9107, opened 2025-12-03
Separate Java version out of rbconfig 'arch'

headius:
The 'arch' field in RbConfig::CONFIG is used by RubyGems to name a directory where the build artifacts from extension gems list. When a gem has been installed, but its extension does not appear in this directory, RubyGems knows the extension needs to be rebuilt and warns.

JRuby extensions are almost never platform-specific, so we have used a Java-related string for 'arch', with the current Java specification version appended like 'universal-java-21'. However because the version is included, any gems that have an extension build -- even if they don't actually build it on JRuby -- get associated with that exact 'arch' string. Switching to a different version of Java causes RubyGems to mistakenly think an extension needs to be built, and it warns.

See https://github.com/ruby/rubygems/issues/3520 for related discussions.

The RubyGems team originally suggested ignoring the 'arch' version, since a similar situation affected CRuby on macOS (with an 'arch' of 'darwin-<version>'), but ultimately they only ended up patching Bundler to ignore this version. Rather than try to further modify RubyGems to ignore this version component, JRuby will omit it in 'arch' and move it to a separate 'arch_version' field users can query if desired.

### DIFF
--- a/core/src/main/java/org/jruby/ext/rbconfig/RbConfigLibrary.java
+++ b/core/src/main/java/org/jruby/ext/rbconfig/RbConfigLibrary.java
@@ -256,8 +256,10 @@ public class RbConfigLibrary implements Library {
         String javaSpecVersion = System.getProperty("java.specification.version");
         if (javaSpecVersion.equals("1.8")) javaSpecVersion = "8";
 
-        // Rubygems is too specific on host cpu so until we have real need lets default to universal
-        setConfig(context, CONFIG, "arch", "universal-java-" + javaSpecVersion);
+        // Rubygems uses this to indicate extensions have been built, so we separate the Java version.
+        // See https://github.com/ruby/rubygems/issues/3520
+        setConfig(context, CONFIG, "arch", "universal-java");
+        setConfig(context, CONFIG, "arch_version", javaSpecVersion);
 
         // Use property for binDir if available, otherwise fall back to common bin default
         String binDir = SafePropertyAccessor.getProperty("jruby.bindir");


eregon:
:+1: this seems the right fix to me.

Regarding having the Java version available, I'm thinking having `RbConfig::CONFIG['java_version']` would make sense (from system property `java.version`), for both JRuby and TruffleRuby.
`java_specification_version` seems also useful potentially (from property `java.specification.version`). It needs to be a String though because all values of RbConfig::CONFIG are strings and some programs rely on that (notably RVM).

In fact I almost added them in TruffleRuby in https://github.com/truffleruby/truffleruby/pull/4025 but ended up declining that PR because the other versions don't yet have a good way to get them.
Also I guess if nobody needs them there is also no need to add them.

headius:
I would feel better about this change if CRuby did something similar to remove host platform versions from their 'arch' value as well, but I will merge this to head and see how the wider community reacts.

There's a fair justification for turning all of the 'java' prefixed properties into CONFIG values, but I think I would still keep the 'arch_version' to provide a common place where the host platform version can be acquired.

headius:
@eregon I'm on board with putting more "java" properties into CONFIG but it's not exactly relevant to this issue. Happy to discuss in a separate issue or PR.

Merging for 10.0.3.0.

headius:
Note there's a related issue for JRuby builds currently: the resolv-replace gem added in JRuby 10 (to support Ruby 3.4 standard gems) depends on the resolv gem, which means it needs to show up as a default gem. Since it is directly in our stdlib right now, RubyGems always warns about the missing extension.

I'm working to switch over our build to that gem in #9108.

PR #9127, opened 2025-12-09
fix(launcher): correct parsing of java release files under Bash/msys2 etc on Windows

chadlwilson:
- fixes #9122 

The Eclipse Temurin files have CRLFs on Windows which breaks the parsing and unquoting here.

Notes
- After this change any other trailing whitespace still breaks the parsing (at least with bash); this only trims a single trailing CR.
- Parsing `.jruby.java_opts` in the non-module JVM case seems to work fine as-is, because it splits on whitespace anyway (which also means it doesn't work for any args with intentional whitespace in them anyway; quoted or unquoted in the file); but I aligned the `IFS=` to isolate from the environment and do things consistently anyway.
- Made some minor tweaks to the shellcheck rules to be a bit more conservative and make IntelliJ IDEA's integration happen
- Logged the parsed major version and minimum versions to assist debugging with `--environment` in future.
- Corrected typo to use `JAVA_RUNTIME_VERSION` correctly as intended per conversation in https://github.com/jruby/jruby/pull/9127#discussion_r2603455725

Example 1 (Temurin x64 on Windows, with CRLF)
```
IMPLEMENTOR="Eclipse Adoptium"
IMPLEMENTOR_VERSION="Temurin-21.0.9+10"
JAVA_RUNTIME_VERSION="21.0.9+10-LTS"
JAVA_VERSION="21.0.9"
JAVA_VERSION_DATE="2025-10-21"
LIBC="default"
[...]

PR #9128, opened 2025-12-10
[9.4] fix(launcher): correct parsing of java release files under Bash/msys2 etc on Windows 

chadlwilson:
Backport of #9127 for 9.4.x (given other launcher changes for `10.0.x` have been ported across).


headius:
Thank you!

PR #9141, opened 2025-12-13
refactorings and cleanup motivated by error-prone

kares:
setting up `error_prone` compiler plugin (again), no longer ran on master (CI) since it was only activated on Java 11

there's couple useful finds, handed them one by one in separate commits


headius (approved):
Only minor things that may or may not need to be fixed. Seems like generally good changes to me.

PR #9142, opened 2025-12-17
Change GC's methods from module_function to metaclass methods

trinistr:
Follow-up to https://github.com/jruby/jruby/pull/9094#discussion_r2581852307

Methods of `GC` are supposed to be just methods of metaclass (except for `#garbage_collect`), not `module_function`s.

There is a slight possibility of this being a breaking change if someone was `include`ing GC to use its methods, but I don't see why anyone would do this when it's mostly no-ops.


headius:
Thank you!

PR #9145, opened 2025-12-18
Never trust external content for FString

headius:
When creating an FString for the first time, we should never trust that the incoming byte[] or ByteList are now ours to own. The caller might still have a reference to them that they continue to modify, resulting in a zombie FString that can't be found or that has incorrect contents.

This came up while trying to implement the "fake string" optimization for uncached headers in puma/puma#3838. Holding a reference to a RubyString and its ByteList that could be updated in-place and then used to caches new headers led to those cached FStrings being modified directly.


headius:
Pushed a test after merge in https://github.com/jruby/jruby/commit/1cb0236ff61ad5b9d4a0fa064a0991b5040e02fd

PR #9149, opened 2025-12-23
*nil no longer calls nil.to_a, similar to how **nil does not call nil…

enebo:
….to_hash

*nil no longer calls nil.to_a, similar to how **nil does not call nil.to_hash. [Feature #21047]

This could be optimized more but this matches semantics.  The missing optimization is the parser(s) need to see literal `nil` and then set a new flag on build_splat marking that fact.  Then the helper can use a cached empty frozen array.  The bulk of the improvement is not dispatching so this will be a smaller percentage of improvement on a pretty uncommon case.

### DIFF
--- a/core/pom.xml
+++ b/core/pom.xml
@@ -55,9 +55,9 @@ DO NOT MODIFY - GENERATED CODE
     <test.results.dir>${build.dir}/test-results</test.results.dir>
     <tzdata.scope>provided</tzdata.scope>
     <tzdata.version>2019c</tzdata.version>
-    <version.ruby>3.4.5</version.ruby>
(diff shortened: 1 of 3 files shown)


enebo:
Note: I updated Ruby version to 3.5.0 which is not the actual version.  This was to match versioning currently used in ruby/spec to hit the specs updated for Ruby 4.0.0.  I assume ruby/spec will update this soon and we can sync and change to proper version.

headius:
@enebo The version guards for 3.5 should still run if the version is 4.0. I think the specs have been updated but not synced yet.

enebo:
@headius haha of course it should.  I will update it to 4.0

PR #9157, opened 2026-01-07
bump included gems

kares:
bump stringio/strscan (due code-range fixes) and bundler+rubygems to latest 3.x

### DIFF
--- a/lib/pom.rb
+++ b/lib/pom.rb
@@ -21,9 +21,9 @@ end
 default_gems = [
   # treat RGs update special:
   # - we do not want bin/update_rubygems or bin/gem overrides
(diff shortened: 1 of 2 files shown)


headius:
I just did something similar but without RG. Merged!

kares:
thanks, wanted to bump more (patch maybe minor to latest) but there are some failures likely due how e.g. ruby-specs assume `pp` internals.

having latest 3.x Bundler (before the 4.0) is going to be quite useful, so included that at least.

headius:
Most gem updates will need to be accompanied by updates to the MRI tests (see #9108) but if there's specs broken with newer versions of gems we should fix them directly (see my change to a security spec also in #9108).

PR #9161, opened 2026-01-07
Avoid triggering PR actions on jruby/jruby branches

headius:
When we do a PR from a branch on the jruby/jruby repo, it will trigger our CI action to run twice: once for the push and once for the PR. We don't want this wasted effort, so this PR will attempt to fix the triggers to only run for pushes to jruby/jruby branches and not for PRs made from those branches.


headius:
Turns out it's not possible via the trigger events.

Issue #8976, opened 2025-08-26
Regression in keyword argument matching in JRuby v10.0.1.0


floehopper:
The problem was first observed in [this mocha CI build failure](https://app.circleci.com/pipelines/github/freerange/mocha/886/workflows/5086ccc0-7856-4031-ac38-2e11b3b94fe7/jobs/13337):

```
  1) Failure:
LooseKeywordArgumentMatchingTest#test_should_match_hash_parameter_with_keyword_args
[test/acceptance/loose_keyword_argument_matching_test.rb:33]:
Expected nil (NilClass) to respond to #include?.
```

I have reproduced the problem locally by running the following command in [the mocha repo](https://github.com/freerange/mocha) on the ‘main’ branch (097f0addf82c68e7ecdf87ebf5c996d7278fa2a6):

    SEED=21038 bundle exec rake test:acceptance

This particular problem doesn't seem to occur when I run the single test or when running all the acceptnace tests with other seed values, so presumably the problem is related to ordering / statefulness between tests. However, I have seen other similar problems with other seed values:

    SEED=38546 bundle exec rake test:acceptance

```
LooseKeywordArgumentMatchingTest#test_should_match_last_positional_hash_with_keyword_args
[test/acceptance/loose_keyword_argument_matching_test.rb:84]:
Expected nil (NilClass) to respond to #include?.
```

And:

    SEED=32553 bundle exec rake test:acceptance

```
  1) Failure:
StrictKeywordArgumentMatchingTest#test_should_not_match_hash_parameter_with_splatted_keyword_args
[test/acceptance/strict_keyword_argument_matching_test.rb:37]:
Test passed unexpectedly
```

I think the issue pre-dates JRuby v10.0.2.0, because I see the same failures when running against JRuby v10.0.1.0. I do not see these failures when running against CRuby v3.4.5 or v3.5.0preview1.

I’m happy to investigate this further given that I probably have more knowledge of the internals of the Mocha tests, but it would be helpful to have some understanding of what kind of statefulness there might be in relation to keyword arguments.

I’m tracking this issue in https://github.com/freerange/mocha/issues/760.

floehopper:
Any updates on this? Let me know if you need more info.

enebo:
@floehopper I can help fix this but in this test method:

```ruby
  def test_should_match_hash_parameter_with_keyword_args
    execution_point = nil
    test_result = run_as_test do
      mock = mock()
[...]

PR #9192, opened 2026-01-28
Stab at Fixing #8976

enebo:
This effectively fixes mocha by noticing a splatted list with a ruby2keywordhash being passed into a ruby2keyword method.  So it dups and strips the ruby2keyword hash state.  The other half of this was to not dup blindly in a couple of code paths if it is a ruby2keywordhash.


headius:
Aces. Thanks for the help on this one!

PR #9109, opened 2025-12-03
Switch to syslog gem

headius:
This PR will eventually switch us to the syslog gem, once our FFI-based version has been incorporated.

See https://github.com/ruby/syslog/issues/1

PR #9193, opened 2026-01-28
Update specs

eregon:


eregon:
@headius How about enabling `Allow auto-merge` in the Settings + some required CI checks (in branch protection)?
Then PRs like this one could auto-merge automatically as soon as they pass the required CI checks.
Not a big problem, but would be nice and both CRuby and TruffleRuby use that currently.

headius:
@eregon I can discuss it with @enebo and maybe we will enable it. I'm reluctant to make any merge automatic because it could result in incomplete work getting merged (if no guards) or require a third-party review or other conditions (which would prevent me merging my own PRs). Perhaps you can explain how you have it set up for TR?

Thanks for the spec update!

headius:
I've made the following config changes:

* Require passing status checks
* Require PR
* Require one approval
* Require branch to be up to date

But disabled:

* Do not allow these rules to be bypassed

I think that will enable auto-merging without limiting us from manually merging our own branches and PRs.

eregon:
> I'm reluctant to make any merge automatic because it could result in incomplete work getting merged (if no guards) or require a third-party review or other conditions (which would prevent me merging my own PRs).

It still requires someone with write access to the repository to enable auto merge, that remains a manual step.

> Perhaps you can explain how you have it set up for TR?

I have `Allow auto-merge` and only `Require a pull request before merging ` & `Require status checks to pass before merging`.
That way maintainers/people with Write access can still merge their own PR.

I would recommend against `Require branch to be up to date` because that's just annoying and doesn't really gain anything (when the change is merged it will be up-to-date, no point to rerun CI N times just because N pushes to master in between). If contributors use the UI for that it will create an extra merge commit which is unnecessary noise.

`Require one approval` might mean I wouldn't be able to merge PRs like this one I think, but the bypass seems to cancel that requirement because when I look e.g. at https://github.com/jruby/jruby/pull/9109 I do see a green merge button.

So looks OK but would recommend to remove `Require branch to be up to date` to avoid lots of extra CI runs and extra merge commits.

PR #9195, opened 2026-01-29
Refactor keyword arguments logic

headius:
This PR will refactor our keyword arguments logic to accomplish the following:

* Simpler to follow logic
* Unified logic for both interpreter and JIT
* Easing the path toward on-stack keyword arguments handling

This is in response to yet another small keyword arguments bug #8976 brought about by making a small change in JIT logic that did not match the completely different-looking interpreter logic. Having a single, unified, readable path will make many things easier.

This is originally targeted at 10.0 but the risks of merging a large refactor may mean it goes into 10.1.


headius:
This is still based on master so 10.0.4.0 seems like a good target. If we can clean up/simplify/fix this logic further, all the better. At least it's easier to follow now.

Issue #8649, opened 2025-02-21
9.4.8.0 mvn compile fails. used to work before resolv gem latest release


jlahtinen:
**Environment Information**
  
Provide at least:
* JRuby version (`jruby -v`) and command line (flags, JRUBY_OPTS, etc)
* * no jruby binary after `rvm use system`
* Operating system and platform (e.g. `uname -a`)
* * Linux domainname 5.14.0-503.23.2.el9_5.x86_64 #1 SMP PREEMPT_DYNAMIC Fri Feb 7 15:07:19 EST 2025 x86_64 x86_64 x86_64 GNU/Linux

**Expected Behavior**

* Describe your expectation of how JRuby should behave, perhaps by showing how CRuby/MRI behaves.
* Provide an executable Ruby script or a link to an example repository.

9.4.8.0 should compile as it did earlier

**Actual Behavior**

* Describe or show the actual behavior.
* Provide text or screen capture showing the behavior.

maven output
```
[ERROR] Failed to execute goal io.takari.polyglot:polyglot-maven-plugin:0.5.0:execute (install_gems) on project jruby-stdlib: org.jruby.exceptions.RuntimeError: (BuildError) ERROR: Failed to build gem native extension.
[ERROR] 
[ERROR]     current directory: /home/jlahtinen/ws/github/jruby/lib/ruby/gems/shared/gems/resolv-0.6.0/ext/win32/resolv
[ERROR] /home/jlahtinen/ws/github/jruby/bin/jruby -I uri:classloader:/META-INF/jruby.home/lib/ruby/stdlib extconf.rb
[ERROR] Building has failed. See above output for more information on the failure.
[ERROR] extconf failed, exit code 1
[ERROR] 
[ERROR] Gem files will remain installed in /home/jlahtinen/ws/github/jruby/lib/ruby/gems/shared/gems/resolv-0.6.0 for inspection.
[ERROR] Results logged to /home/jlahtinen/ws/github/jruby/lib/ruby/gems/shared/extensions/universal-java-11/3.1.0/resolv-0.6.0/gem_make.out
[ERROR] : (InstallError) extconf failed, exit code 1
[ERROR] -> [Help 1]
org.apache.maven.lifecycle.LifecycleExecutionException: Failed to execute goal io.takari.polyglot:polyglot-maven-plugin:0.5.0:execute (install_gems) on project jruby-stdlib: org.jruby.exceptions.RuntimeError: (BuildError) ERROR: Failed to build gem native extension.

[...]

PR #9197, opened 2026-01-29
Force build-time gem installs to use current JRuby config

headius:
The RubyGems we run at build time runs with JRuby 9.4, which causes it to use properties from that version for things like gem paths and extension directories. This is related to the extension disabling change in jruby/jruby#8415 and causes the build-time extension faking to use the wrong path. For example:

  .../gems/shared/extensions/universal-java-25/3.1.0

when it should be

  .../gems/shared/extensions/universal-java/3.4.0

The change here uses RubyGems' cross-compiling capability by forcing a specific rbconfig.rb to be use, and additionally fakes out the Ruby API version by setting @ruby_api_version on the Gem module.

This was discovered while attempting to switch fully to the syslog gem in jruby/jruby#9109, since as a bundled gem it will not be activated unless we successfully fake-out the extension build.

See also this bug related to the resolv gem installing extensions:

* https://github.com/jruby/jruby/issues/8649

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -29,7 +29,6 @@ failures
 .debug.properties
 .redcar
 /.rbx/
-build
 build.properties
 build_graph.png
 core/src/main/java/org/jruby/runtime/Constants.java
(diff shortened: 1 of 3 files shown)


headius:
I merged #9198 before this so that pulled in the changes here.

Issue #9199, opened 2026-01-29
Output formatting differences between jRuby 10.0.0.1 and Ruby 3.4 when converting Float to JSON


evaniainbrooks:
Tracking https://github.com/ruby/json/issues/817

headius:
Fixed (or at least the fix in json is now possible) by #9200.

PR #9200, opened 2026-01-29
Extract float formatting into a static method to use with to_json

evaniainbrooks:
https://github.com/jruby/jruby/issues/9199
https://github.com/ruby/json/issues/817


headius:
@evaniainbrooks This is great, thank you!

The spec should probably go in spec/ruby if that behavior is not already in specs, and we'll probably want to backport this to 9.4 so the json gem doesn't break for future versions, but this could probably go into 10.0.3.0 to get the json work moving forward.

Requesting review from @enebo.

evaniainbrooks:
I reverted the added spec because this is certainly already covered in `spec/ruby/core/float/shared/to_s.rb`

headius:
@evaniainbrooks Thank you! Let me know if I can do anything else to help.

Issue #9179, opened 2026-01-14
Ruby 4.0.1 fixes


headius:
Ruby 4.0.1 has been released with a few fixes we should confirm on JRuby:

* [ ] [Bug #21812: Kernel#sleep without arguments returns immediately when subprocess exits in another thread (regression in Ruby 4.0) - Ruby - Ruby Issue Tracking System](https://bugs.ruby-lang.org/issues/21812)
* [ ] [Bug #21828: An incorrect warning message related to `benchmark` is shown when using `benchmark-ips` - Ruby - Ruby Issue Tracking System](https://bugs.ruby-lang.org/issues/21828)
* [ ] [Bug #21811: Fix underflow in Array#pack - Ruby - Ruby Issue Tracking System](https://bugs.ruby-lang.org/issues/21811)
* [ ] [Bug #21814: 0.pow(2,-9999999999999999990) should be zero - Ruby - Ruby Issue Tracking System](https://bugs.ruby-lang.org/issues/21814)
* [ ] [Bug #21819: A Data object should be frozen even if it has no members - Ruby - Ruby Issue Tracking System](https://bugs.ruby-lang.org/issues/21819)

PR #9202, opened 2026-01-30
A Data object should be frozen even if it has no members

evaniainbrooks:
issue #9179


headius:
@evaniainbrooks changes for Ruby 4.0 and beyond are currently living on the ruby-4.0 branch. Could you check this behavior there and rebase if necessary? There are already other Data changes there. 

After JRuby 10.0.3.0 I plan to branch off 10.0 and start moving JRuby 10.1 and Ruby 4.0 work to master, which will make these changes easier.

evaniainbrooks:
Rebased to ruby-4.0

enebo:
@evaniainbrooks coolio.  Thanks for the fix.

PR #9205, opened 2026-02-01
Fix encoding fallback error handling and error message formatting

khasinski:
## Summary

Follow-up to #9182 as [suggested by @headius](https://github.com/jruby/jruby/pull/9182#issuecomment-3821300980).

This PR fixes the remaining 12 failing specs in `spec/ruby/core/string/encode_spec.rb`:

1. **Raise `ArgumentError("too big fallback string")`** when a fallback function returns a value that cannot be encoded to the destination encoding. Previously, the invalid character was silently dropped.

2. **Format UTF-8 codepoints as `U+XXXX`** in `UndefinedConversionError` messages instead of showing the raw byte dump (e.g., `U+FFFD` instead of `"\xEF\xBF\xBD"`), matching MRI behavior.

## Test plan

- [x] All 152 specs in `encode_spec.rb` now pass (previously 12 were tagged as failing)
- [x] Removed `spec/tags/ruby/core/string/encode_tags.txt` since all specs now pass


headius:
Nice! I'll review.

headius (approved):
Love seeing those tags disappear!

Issue #8876, opened 2025-06-25
`import_methods` doesn't allow calling another refined method


esotericpig:
**Environment Information**

* `jruby 10.0.0.1 (3.4.2) 2025-05-07 79cf1e4aec OpenJDK 64-Bit Server VM 21.0.4+7-LTS on 21.0.4+7-LTS +indy +jit [x86_64-linux]`
* `Linux x86_64 GNU/Linux`

**Expected Behavior**

The following works on CRuby, but errors on JRuby v9 and v10:

```ruby
# frozen_string_literal: true

module MyRef
  module MyMethods
    def my_ref_method
      print '1, '
      my_ref_method_2
    end

    def my_ref_method_2
      puts '2'
    end
  end

  refine Module do
    import_methods MyMethods
  end
end

module MyMod
  using MyRef

  # Ruby  => [:my_ref_method, :my_ref_method_2]
  # JRuby => [:my_ref_method_2, :my_ref_method]
  owner = Module.instance_method(:my_ref_method).owner
  puts owner
  puts owner.instance_methods(false).inspect

  # Ruby  => 1, 2
  # JRuby => 1, NameError: undefined local variable or method 'my_ref_method_2' for module MyMod
  my_ref_method
end
```

**Actual Behavior**

Can't call `my_ref_method_2` inside of `my_ref_method` in JRuby.

esotericpig:
I forgot to mention it works as expected in JRuby without `import_methods`:

```ruby
  refine Module do
    # import_methods MyMethods

    def my_ref_method
      print '1, '
      my_ref_method_2
    end

    def my_ref_method_2
      puts '2'
    end
  end
```

So a workaround is to not use `import_methods`.

headius:
@esotericpig Esoteric indeed! I didn't even know this pattern was possible.

I assume `import_methods` on its own works correctly?

esotericpig:
> [@esotericpig](https://github.com/esotericpig) Esoteric indeed! I didn't even know this pattern was possible.
> 
> I assume `import_methods` on its own works correctly?

Yes, it works fine as long as you don't call another method. 👍

headius:
Oh I see! Seems like it's not rehoming into the new location and dispatching against the correct class.

esotericpig:
Kind of unrelated, I also see that `used_modules` is always empty. I guess this hasn't been implemented yet in JRuby? I use it in my spec tests to assert that my refinement module was added, but I can just guard it with an if-JRuby check.

[...]

PR #9206, opened 2026-02-02
Fix exception message when calling import_methods without a Module

evaniainbrooks:
issue #8876 

Found some low hanging fruit while looking at this issue.


headius (approved):

headius:
Nice little fix, thank you. Do you have more changes coming?

evaniainbrooks:
@headius Yes, but they can probably come in as a separate PR.

headius:
Thank you!

PR #9209, opened 2026-02-03
Add ruby_bug for spec of Data with no members being frozen

herwinw:
This makes the spec pass on Ruby 4.0.0 and earlies as well. This saves some issues with the next spec sync.


herwinw:
This relates to #9202

headius:
This looks fine to me. Using `ruby_bug` for a teeny version is not common but it is done in a couple of places.

PR #9210, opened 2026-02-03
Remove Process::Status#& and Process::Status#>>

herwinw:
This fixes an issue of #9061.

Related upstream/MRI issue: https://bugs.ruby-lang.org/issues/19868


herwinw:
I'm seeing a lot of CI failures, the few things I've looked at do not appear to be related to this change.

headius:
@herwinw The ruby-4.0 branch is failing anything from 4.0 tests and specs that we have not yet implemented, so don't worry about that for now.

headius:
Thank you!

Issue #9208, opened 2026-02-02
Inconsistency with block parameter default value


katafrakt:
We encountered a bug in dry-types where JRuby does not behave the same way as MRI. I spent some time trying to reproduce it in non-dry-types environment and I think I have a minimal reproducible example now (although still quite convoluted):

```ruby
class Inner
  def call(input)
    result = input.each_with_object([]) { |x, arr| arr << x * 2 }
    yield(result)
  end
end

class Outer
  def initialize(inner) = @inner = inner

  def call(input, &block)
    @inner.call(input) { |output = input| block.call(output) }
  end
end

input = [1, 2, 3]
p Outer.new(Inner.new).call(input) { |x| x }
```

On Ruby 3.4.1 it returns `[2, 4, 6]`, while in JRuby it just returns `2`.

**Environment Information**
  
Provide at least:
* JRuby version: 10.0.2.0
* Operating system and platform : Linux 6.18.6-arch1-1

**Expected Behavior**

* JRuby and MRI behave the same way, returning `[2, 4, 6]`.

**Actual Behavior**

* JRuby returns different data structure.

enebo:
Confirmed.  Thanks for the report.

headius:
Fixed by #9211. Thanks for the report @katafrakt and the fix @evaniainbrooks!

katafrakt:
Thanks a ton @headius and @evaniainbrooks! ❤️ 
I confirmed this works with Dry Types correctly now.

PR #9211, opened 2026-02-03
Fix block argument destructuring with optional parameters

evaniainbrooks:
issue #9208

### DIFF
--- a/core/src/main/java/org/jruby/ir/runtime/IRRuntimeHelpers.java
+++ b/core/src/main/java/org/jruby/ir/runtime/IRRuntimeHelpers.java
@@ -2189,7 +2189,7 @@ public class IRRuntimeHelpers {
 
         // FIXME: For NORMAL/THREAD but it is unclear if we really need any keyword logic in here anymore.
         org.jruby.runtime.Signature sig = block.getBody().getSignature();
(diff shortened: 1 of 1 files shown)


headius:
Thanks for looking into this. That's a peculiar issue.

headius:
The `sequel` and `concurrent-ruby` job fails were due to a bug in #9198 that caused the build to generate broken binstubs. Fixed in f21870a665daf6dcc38d4d354c9218fe618d1a71.

evaniainbrooks:
@headius thanks. I think everything is green now. The single failing test looks like an unrelated time-sensitive issue.

headius:
@evaniainbrooks I think those last failures are due to a similar change being needed in some JIT-related path. That spec run runs the language specs 4 times in the same runtime, so that the JIT can fire and fully bind all calls.

I'll see if I can suss out what's wrong.

headius:
I think I found the right spot in JIT's argument-handling logic to patch. It has diverged for various reasons and needs to be reined in, but if this passes we can proceed for now.

evaniainbrooks:
@headius thank you for the explanation.  That certainly explains some other things that perplexed me while I was stepping through this.

Issue #8210, opened 2024-04-24
Requiring visibility=PRIVATE for native initialize is error-prone


headius:
In #8146 we see a side effect of not properly setting up a native `#initialize` method as private: `DelegateClass` will attempt to also wrap `#initialize`, breaking the delegator's constructor path and leading to `ArgumentError: not delegated` errors.

The issue in that case is that a third-party JRuby extension defines `#initialize` without any change to visibility, which effectively results in it being public. The immediate fix would be to add `visibility = Visibility.PRIVATE` to the `@JRubyMethod` annotation, but this is obviously going to be missed by people unfamiliar with the requirement that `#initialize` be private. I am opening this bug to consider ways to prevent that from happening in the future.

A few ideas:

* Warn at runtime if a native `#initialize` is defined as public. It would be noise on startup, but immediately apparent to any extension writers what they need to do.
* Always defined `#initialize` as private, regardless of the annotation.
  * We could possibly allow an exception for when it is **explicitly** defined as public... but I'm not sure how to detect that. The default in the annotation is already public, so we have no way to know whether it was left unspecified or explicitly set to public.
* Enhance `DelegateClass` to always ignore initialize, since delegating it will almost always be wrong and broken. This would make it possible to fix immediately (with a gem release) and dodge the issue of how to handle users that make public `#initialize` methods.

The workaround is to manually update the visibility of any affected `#initialize` back to private, as shown in #8146.

PR #9213, opened 2026-02-03
Make `refine` method of `Module` private

katafrakt:
In MRI refine method [is defined as private](https://github.com/ruby/ruby/blob/master/eval.c#L2237). But in JRuby it is public leading to inconsistency between the implementations.

```
irb(main):001> [RUBY_ENGINE, RUBY_VERSION]
=> ["jruby", "3.4.2"]
irb(main):002> Module.instance_methods.include?(:refine)
=> true
```

```
irb(main):001> [RUBY_ENGINE, RUBY_VERSION]
=> ["ruby", "3.4.1"]
irb(main):002> Module.instance_methods.include?(:refine)
=> false
```

### DIFF
--- a/core/src/main/java/org/jruby/RubyModule.java
+++ b/core/src/main/java/org/jruby/RubyModule.java
@@ -989,7 +989,7 @@ public class RubyModule extends RubyObject {
         return metaClass.getRealClass().getName(context) + ":0x" + Integer.toHexString(System.identityHashCode(this));
     }
 
-    @JRubyMethod(name = "refine", reads = SCOPE)
+    @JRubyMethod(name = "refine", visibility = PRIVATE, reads = SCOPE)
     public IRubyObject refine(ThreadContext context, IRubyObject klass, Block block) {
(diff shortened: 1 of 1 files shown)


headius:
Yup, good catch. We probably should rework "module" visibility to do this by default, since it's almost always the expected behavior.

headius:
Related issue where we made the default visibility for some methods "private" if unspecified: https://github.com/jruby/jruby/issues/8210

headius:
The sequel and concurrent-ruby jobs seem to be hanging while bundling... I'm not sure what's up with that but it's unlikely to be related to your change.

headius (approved):

headius:
Nevermind what I said about module methods; this is just a normal method that should have had PRIVATE visibility.

headius:
The `sequel` and `concurrent-ruby` job fails were due to a bug in #9198 that caused the build to generate broken binstubs. Fixed in f21870a665daf6dcc38d4d354c9218fe618d1a71.

Issue #9216, opened 2026-02-05
jruby.sh fails with JRUBY_CP: unbound variable


eregon:
**Environment Information**
  
Provide at least:
* JRuby version (`jruby -v`) and command line (flags, JRUBY_OPTS, etc)
`jruby-head`, `jruby 10.0.4.0-SNAPSHOT (3.4.5) 2026-02-05 f21870a665 OpenJDK 64-Bit Server VM 25.0.2+10 on 25.0.2+10 +indy +jit [x86_64-linux]`
* Operating system and platform (e.g. `uname -a`)
`x86_64 GNU/Linux`

Happens both with or without `jruby-launcher`.
I installed jruby using `ruby-build jruby-dev -d ~/.rubies` (needs https://github.com/rbenv/ruby-build/pull/2603).

I noticed that JRuby seems to fail most command_line specs in https://github.com/eregon/rubyspec-stats/actions/runs/21672566977/job/62483988845
I can reproduce when running the same command inside the JRuby repository:
```
$ chruby jruby-dev
$ cd jruby
$ jruby -v
jruby 10.0.4.0-SNAPSHOT (3.4.5) 2026-02-05 f21870a665 OpenJDK 64-Bit Server VM 25.0.2+10 on 25.0.2+10 +indy +jit [x86_64-linux]
$ spec/mspec/bin/mspec ci --timeout 60 --target ruby --marker . --format yaml --output stats.yml spec/ruby/command_line/dash_a_spec.rb
$ ruby -Xbacktrace.style=mri /home/eregon/code/jruby/spec/mspec/bin/mspec-ci --timeout 60 --marker . --format yaml --output stats.yml spec/ruby/command_line/dash_a_spec.rb
/home/eregon/code/jruby/spec/mspec/lib/mspec/commands/mspec.rb:110: warning: unsupported exec option: close_others
jruby 10.0.4.0-SNAPSHOT (3.4.5) 2026-02-05 f21870a665 OpenJDK 64-Bit Server VM 25.0.2+10 on 25.0.2+10 +indy +jit [x86_64-linux]
./home/eregon/code/jruby/bin/jruby.sh: line 990: JRUBY_CP: unbound variable
/home/eregon/code/jruby/bin/jruby.sh: line 990: JRUBY_CP: unbound variable
zsh: exit 1     spec/mspec/bin/mspec ci --timeout 60 --target ruby --marker . --format yaml
```

[...]

PR #9217, opened 2026-02-05
Add an error for running jruby.sh from a broken JRUBY_HOME

headius:
In #9216 @eregon reported that JRuby was failing to run specs with an error about unbound variable `JRUBY_CP`. This variable can only be left unset if the `bin/jruby.sh` being executed is not accompanied by a sibling `lib/jruby.jar`, as might happen when running from a fresh repository clone.

The issue in #9216 was that for some reason, subprocesses launching `ruby` were choosing the bin scripts from a fresh JRuby clone rather than the one in `PATH` associated with a fully-functional JRuby.

This configuration is unsupported; `bin/jruby.sh` uses relative pathing to find its associated `lib/jruby.jar`, and if it is not present there's no way to proceed with execution. But the error message provided is cryptic and does not help a user debug the issue.

This path adds a check for the missing jar and terminates early with the following error message (negotiable):

```
[] jruby $ which ruby
/Users/headius/work/jruby/bin/ruby
[] jruby $ ls /Users/headius/work/jruby/lib/*.jar
zsh: no matches found: /Users/headius/work/jruby/lib/*.jar
[] jruby $ ruby -v
JRUBY_HOME at /Users/headius/work/jruby contains no lib/jruby.jar, exiting.
[] jruby $
```

### DIFF
--- a/bin/jruby.sh
+++ b/bin/jruby.sh
@@ -671,6 +671,11 @@ for j in "$JRUBY_HOME"/lib/jruby.jar "$JRUBY_HOME"/lib/jruby-complete.jar; do
     fi
     jruby_jar="$j"
 done
+
+if [ -z "${jruby_jar-}" ]; then
+    echo "JRUBY_HOME at ${JRUBY_HOME} contains no lib/jruby.jar, exiting."
+    exit 1
+fi
 readonly jruby_jar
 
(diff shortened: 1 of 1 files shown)


headius:
Ping @mrnoname1000 for a review.

PR #9219, opened 2026-02-06
True keyword arguments

headius:
This PR will implement pass-through keyword arguments for simple cases of calling core methods or calling other keyword-receiving methods. It is meant as a starting point to fully support pass-through of as many forms of keyword-based invocation as possible.

## IR changes

- [x] Associate the kwargs hash with the CallInstr directly, so the call has all metadata for both positional and keyword arguments.

## Core changes

- [ ] Introduce annotations and processing to allow specifying keyword-receiving overloads.
- [ ] Identify and modify an existing keyword-argument core method to have a specialized form and redirect generic forms to it. Any methods that accept a single keyword like `exception:` or `nonblock:` would be good candidates for now.

### Caller side

The caller side indy optimizations can be implemented early and provides benefits without the rest due to moving hash construction and flags-setting into the call site.

- [x] Embed keywords and kwargs layout into the invokedynamic call site, passing all elements through without constructing a Hash.
- [x] Implement fallback call site logic that just constructs the hash, sets appropriate flags, and passes it as before.

### Callee side

- [ ] Wire up invokedynamic callee side to discover and bind specialized keyword argument paths, similar to the eager arity-splitting logic already introduced.


headius:
Pushed more fixes that allow language specs to run. Base was prior to many spec fixes and green CI so merged from master.

headius:
Most recent changes have direct argument passing for normal positional and normal keyword arguments wired up to a new MetaCallSite. MetaCallSite will be like a switchboard of sorts, either binding the directly-passed arguments straight through to appropriate target, or falling back on boxing argument lists into arrays and keyword arguments into hashes.

The current logic only has one path: box kwargs and fall back on the old logic that knows how to handle route simple argument lists.

The next missing piece to prove this out is a way to specify native core methods with explicit mapping of parameters to their Ruby type. An idea for what this might look like, based on `IO#read_nonblock`, with one required argument, one optional argument, and one optional keyword argument:

```java
@JRubyMethod(name = "read_nonblock", direct = true, strictKeywords = false)
public IRubyObject read_nonblock(ThreadContext context,
                                 @Req IRubyObject length,
                                 @Opt(NIL) IRubyObject buffer,
                                 @OptKey(name = "exception", defaultValue = NIL) IRubyObject _exception) {
    boolean exception = _exception != context.fals;
    return doReadNonblock(context, length, buffer, exception);
}
```

If we can encode this information into the method table, setting up MetaCallSite to pass these arguments directly through and in the correct order will be a simple matter of method handles.

headius:
Additional cases to consider:

* Forwarded arguments (#9472), both receive side and send side.
* Forwarded blocks

headius:
The downstream work of wiring up MetaCallSite to specialized call paths can happen outside of this PR, so I've updated this branch to ensure it still passes verification and we could merge the base work into 10.1.1.0 to get it out there.

Assuming everything is green I will move the remaining tasks to a subissue (or subissues) and we can continue working toward full call-path specialization in 10.1.2.0 and beyond.

Issue #9231, opened 2026-02-11
Thread safety of require and implications on stdlib string.normalize()


mrckzgl:
Consider the following code:
```
value = value.normalize(:nfc)
```
which runs parallel in multiple threads. With JRuby 9.4.10.0 we just sometimes get this error:
```
/usr/share/rvm/rubies/jruby-9.4.10.0/lib/ruby/stdlib/unicode_normalize.rb:41:in `unicode_normalize': undefined method `normalize' for UnicodeNormalize:Module (NoMethodError)
``` 
presumabely on the first (or second see below) invocation of that code. Later invocations run without the error. It is hard to debug as it is purely undeterministic. Source of unicode_normalize.rb looks like this:
```
  def unicode_normalize(form = :nfc)
    require 'unicode_normalize/normalize.rb' unless defined? UnicodeNormalize
    ## The following line can be uncommented to avoid repeated checking for
    ## UnicodeNormalize. However, tests didn't show any noticeable speedup
    ## when doing this. This comment also applies to the commented out lines
    ## in String#unicode_normalize! and String#unicode_normalized?.
    # String.send(:define_method, :unicode_normalize, ->(form = :nfc) { UnicodeNormalize.normalize(self, form) } )
    UnicodeNormalize.normalize(self, form)
  end
```

It took me a while to get following hypothesis. The [wiki](https://github.com/jruby/jruby/wiki/Concurrency-in-jruby) states:
> You should take care in the following situations: concurrent requires, or lazy requires that may happen at runtime in a parallel thread. If objects are in flight while classes are being modified, or constants/globals/class variables are set before their referrents are completely initialized, other threads could have problems.

Is it possible that this kicks in here? First invocation of normalize on first thread does require. Second thread does that in parallel, but sees UnicodeNormalize to be defined, but not all its instance methods being loaded as first thread is still inside the require call? Is it possible that this is the source of the error message?

If so, I wonder if the standard lib should be modified to not do the lazy require? It is one thing to check your own code for thread safety issues, but I would not ever have consider this kind of concurrency problem in a standard library string method.

headius:
@mrckzgl You hit the nail on the head! 🔨

This is a classic example of Rubyists should *never* rely on the existence of a constant as indication that the constant has been fully loaded. Even in CRuby there's no way to avoid this potentially failing, since it could context-switch to another thread in the middle of loading `unicode_normalize/normalize.rb`.

[...]

PR #9232, opened 2026-02-11
Internalize and threadsafetify String#unicode_normalize

headius:
In CRuby, this code was moved into C in 2017 and the .rb file was removed shortly after (released in Ruby 2.5.0).

This patch does the following:

* Move the script into the JRuby jar so it can be loaded as part of our boot logic.
* Fix the race to load UnicodeNormalize by removing the constant definition check and restoring the method redefinition (see jruby/jruby#9231).

Fixes jruby/jruby#9231 and relates to jruby/jruby#6362.


headius:
CRuby changes to move this into C code:

* https://github.com/ruby/ruby/commit/5fee67c9bade0337b5c2c88a7fa1c1068b477803
* https://github.com/ruby/ruby/commit/90ab1ee023ecdc16913e3aa9ca013eb7781068c1
* https://github.com/ruby/ruby/commit/140560e4eeff4ffbf9d63183db25234f46943ab6
* https://github.com/ruby/ruby/commit/c7baa31c7683e2f8b95950f112a38dfd33e5a2ad

PR #9237, opened 2026-02-14
Fix for import_methods does not allow calling another refined method

evaniainbrooks:
issue #8876


headius:
I'm trying to untangle this a bit so I understand the change. I'd like to better understand and document the conditions where a CallBase is potentiallyRefined but does not have a RefinedCachingCallSite.

headius:
This fix works but only repairs the interpreter. JIT mode needs additional help to clear out already-compiled code when the scope's refinements are changed.

```
[] jruby $ jruby blah.rb
NoMethodError: undefined method 'indent' for an instance of String
  indent_normal at blah.rb:9
         <main> at blah.rb:19
    module_exec at org/jruby/RubyModule.java:4247
     initialize at org/jruby/RubyModule.java:3432
            new at org/jruby/RubyClass.java:1050
         <main> at blah.rb:13
[] jruby $ jruby -X-C blah.rb
"   foo"
```

I'm working on some additional patches on top of this PR.

headius:
Interpreter and JIT can be fixed by having the import clear out old compile code, as @evaniainbrooks did for `fullInterpreterContext` and as we can do by resetting the `callCount` and `actualMethod` in `AbstractIRMethod` and `MixedModeIRMethod`, but methods compiled eagerly at boot (`CompiledIRMethod`) don't have any way to recompile. In order to make them work with this we need to finally make all dynamic call sites in the JIT refinable.

I also observed that the changes in `CallBase` don't appear to trigger for the startup interpreter, likely because the sites passing in unrefined `CallSite` instances are only produced by the full interpreter. I'm going to see if I can clean up those paths so this logic can just be removed (if a call should be refined, we should not be passing in unrefined call sites and having to fix them up in the constructor).

headius:
I've pushed some additional changes as #9248, which fixes the JIT issue and eliminates the call site fix-up logic. The `CompiledIRMethod` issue will take a larger project to fix.

headius:
I will merge this PR and then my additional fixes. I'll open a separate issue for the `CompiledIRMethod` issues with changing refinement status.

Issue #9238, opened 2026-02-14
Cloned class's singleton class has too many classes in hierarchy


headius:
When requesting a singleton class of a class, we appear to be putting too many classes in the ancestors hierarchy.

Example code:
```ruby
a = Class.new
c = a.clone
p a.singleton_class.ancestors
p c.singleton_class.ancestors
```

Output:
```
[] jruby $ jruby blah.rb     
[#<Class:#<Class:0x3b96f8b0>>, #<Class:#<Class:0x2b170932>>, #<Class:#<Class:0x1fc0d9b4>>, Class, Module, Object, Kernel, BasicObject]
[#<Class:#<Class:0x4e4f4092>>, #<Class:#<Class:0x5d829ef0>>, #<Class:#<Class:0x2b170932>>, #<Class:#<Class:0x1fc0d9b4>>, Class, Module, Object, Kernel, BasicObject]
[] jruby $ cx 4 ruby blah.rb
[#<Class:#<Class:0x000000012014e260>>, #<Class:Object>, #<Class:BasicObject>, Class, Module, Object, Kernel, BasicObject]
[#<Class:#<Class:0x000000012014e120>>, #<Class:Object>, #<Class:BasicObject>, Class, Module, Object, Kernel, BasicObject]
```

The cloned class should have the same ancestors for its singleton class, but there's an extra one in there. Somewhere along the way we are treating a cloned class differently.

This leads to failures in Ruby 4.0 CI after other fixes to class and module cloning:

```
 14) Failure:
TestClass#test_class_hierarchy_inside_initialize_clone_bug_21538 [/Users/headius/work/jruby/test/mri/ruby/test_class.rb:297]:
<[#<Class:#<Class:0x27dbaa33>>,
 #<Class:#<Class:0x723ed581>>,
 #<Class:#<Class:0xf59da34>>,
 Class,
 Module,
 Object,
 PP::ObjectMixin,
 Kernel,
 BasicObject]> expected but was
<[#<Class:#<Class:0x27dbaa33>>,
 #<Class:#<Class:0x60fc7f43>>,
 #<Class:#<Class:0x723ed581>>,
 #<Class:#<Class:0xf59da34>>,
 Class,
 Module,
 Object,
 PP::ObjectMixin,
 Kernel,
 BasicObject]>.

[ 396/6636] TestClass#test_class_hierarchy_inside_initialize_dup_bug_21538 = 0.00 s            
 15) Failure:
[...]

PR #9239, opened 2026-02-15
Fix for cloned class singleton class has too many ancestors

evaniainbrooks:
issue #9238


headius:
Looks right. Regressions in JRuby tests are due to an inspect change I made that does not appear to be necessary after this fix. I'll sort it out.

Issue #9120, opened 2025-12-07
`File.path` doesn't raise `Encoding::CompatibilityError` for ASCII-incompatible string


nobu:
**Environment Information**
  
Provide at least:
* JRuby version (`jruby -v`) and command line (flags, JRUBY_OPTS, etc)
  `jruby 10.0.2.0 (3.4.2) 2025-08-07 cba6031bd0 Java HotSpot(TM) 64-Bit Server VM 21.0.8+12-LTS-250 on 21.0.8+12-LTS-250 +indy +jit [arm64-darwin]`

* Operating system and platform (e.g. `uname -a`)
  `Darwin mac.local 25.1.0 Darwin Kernel Version 25.1.0: Mon Oct 20 19:30:01 PDT 2025; root:xnu-12377.41.6~2/RELEASE_ARM64_T6031 arm64 arm Darwin`

Other relevant info you may wish to add:
* Installed or activated gems
* Application/framework version (e.g. Rails, Sinatra)
* Environment variables

**Expected Behavior**

CRuby raises an `Encoding::CompatibilityError`.

```console
$ ruby -e 'p File.path("a".encode("UTF-32BE"))'
ruby 4.0.0dev (2025-12-06T11:11:40Z master 0346206d3e) +PRISM [arm64-darwin25]
-e:1:in 'File.path': path name must be ASCII-compatible (UTF-32BE): "a" (Encoding::CompatibilityError)
	from -e:1:in '<main>'
```

**Actual Behavior**

JRuby doesn't.
```console
$ jruby -v -e 'p File.path("a".encode("UTF-32BE"))'
jruby 10.0.2.0 (3.4.2) 2025-08-07 cba6031bd0 Java HotSpot(TM) 64-Bit Server VM 21.0.8+12-LTS-250 on 21.0.8+12-LTS-250 +indy +jit [arm64-darwin]
"a"
```

headius:
Fixed by #9240. Thanks @evaniainbrooks!

PR #9240, opened 2026-02-16
Fix for File.path doesn't raise Encoding::CompatibilityError for ASCII incompatible string

evaniainbrooks:
issue #9120


enebo:
The sequel thing cannot possibly be related. Looks good to me.  You could have used the newer feature for instanceof to prevent manual cast but it isn't important.  Thanks!

PR #9248, opened 2026-02-17
Additional fixes for import_methods with JIT

headius:
Additional fixes on top of #9237.

That PR modified `import_module` logic to partially clear optimized interpreter state from cloned methods, but did not do the same for JIT-compiled methods. This PR has the additional changes:

* Always create a new `CallSite` in the primary `CallBase` constructor. This avoids the extra logic to fix up non-refined call sites being passed in from cloned call instructions when a refined site is needed. It is better for us to have the call site and call site ID be new for every call instruction anyway.
* Clean up and reduce visibility of all call instruction constructor paths used solely for cloning.
* Add a `reset` method to `MixedModeIRMethod` to also clear any JIT-compiled version of the method.

This does not fix the larger problem of precompiled methods (AOT, at boot or at load, using `CompiledIRMethod`) being unable to recompile after they have been cloned, but 99% of users will run in interpreted or JIT mode, not AOT.

### DIFF
--- a/core/src/main/java/org/jruby/internal/runtime/AbstractIRMethod.java
+++ b/core/src/main/java/org/jruby/internal/runtime/AbstractIRMethod.java
@@ -33,6 +33,7 @@ import org.jruby.RubyModule;
 import org.jruby.compiler.Compilable;
 import org.jruby.internal.runtime.methods.DynamicMethod;
 import org.jruby.internal.runtime.methods.IRMethodArgs;
+import org.jruby.internal.runtime.methods.MixedModeIRMethod;
 import org.jruby.ir.IRFlags;
 import org.jruby.ir.IRMethod;
 import org.jruby.ir.IRScope;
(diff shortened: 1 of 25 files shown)


headius:
@enebo If you get a change to review this after the fact, my only concern is if the `callSiteId` *needs* to be preserved for the profiling/inlining logic. I don't know why that would be the case but I am not as familiar with that logic as you are.

enebo:
@headius That value is used for 2 things: profiling and finding split point when an inline happens.  clone()'ing an instr to a different version in theory is still the same location(callsite) so you could argue it should be the same.  However, there are a few caveats:

 1. If in a speculative optimization it should be treated as a new site.  The statistics gathered for that callsite in a new method could be very different.  so it makes sense to have a new one (note: we do not do this at all today but it is still worth noting)
 2. The callsite effectively changes to something else like normal method becomes another method.  This is a new method which makes it not a candidate at that point since it changed.  So it doesn't really matter that it gets a new callsiteId.

I think though any place we cloneForInlining which is where most of this is occuring it fits into 1 above.  The original problem basically is 2 above.  The spirit of profiling is to find things which do not change.  At worst, giving a new callsiteId if it was done in error would only slow down profiling.  I think this looks good to me from what I see in the PR.

headius:
@enebo ok that's the sort of thing I was thinking of. If we ever get back to speculative profiling and optimizing I'm sure we will revisit this anyway.

PR #9262, opened 2026-02-25
Abstract RubyHash

headius:
This PR is a first pass at making RubyHash abstract to allow new subclass implementations for specific uses. The original implementation, based on pre-direct-addressing CRuby logic, now lives in RubyHashLinkedBuckets.

The impact to consumers of the public RubyHash API should be minimal, except for the now non-functional constructors.

The goal here is to allow RubyHash API-compatible subclasses that either wrap other Hash-like data structures or which implement specific variations like an associative array or a direct- addressing hash. We may also use this to implement thread-safe Hash variants.

At least one external library has already run afoul of the inoperable constructors: jruby-openssl

```diff
diff --git a/src/main/java/org/jruby/ext/openssl/X509Name.java b/src/main/java/org/jruby/ext/openssl/X509Name.java
index 58b6a4e4..9bab2b16 100644
--- a/src/main/java/org/jruby/ext/openssl/X509Name.java
+++ b/src/main/java/org/jruby/ext/openssl/X509Name.java
@@ -123,7 +123,8 @@ public class X509Name extends RubyObject {
         final RubyFixnum IA5_STRING = runtime.newFixnum(BERTags.IA5_STRING);
 
         final ThreadContext context = runtime.getCurrentContext();
-        final RubyHash hash = new RubyHash(runtime, UTF8_STRING);
+        final RubyHash hash = RubyHash.newHash(runtime);
+        hash.default_value_set(UTF8_STRING);
         hash.op_aset(context, newString(runtime, new byte[] { 'C' }), PRINTABLE_STRING);
         final byte[] countryName = { 'c','o','u','n','t','r','y','N','a','m','e' };
         hash.op_aset(context, newString(runtime, countryName), PRINTABLE_STRING);

```

### DIFF
--- a/core/src/test/java/org/jruby/exceptions/TestRaiseException.java
+++ b/core/src/test/java/org/jruby/exceptions/TestRaiseException.java
@@ -318,12 +318,12 @@ public class TestRaiseException extends Base {
 
         boolean hash_default = false;
         for ( Object element : trace ) {
-            if ( element.toString().contains("org.jruby.RubyHash.default")) {
+            if ( element.toString().contains("org.jruby.RubyHashLinkedBuckets.default")) {
                 if ( hash_default ) fail("duplicate " + element + " in : \n" + fullTrace);
                 hash_default = true;
             }
         }
-        assertTrue("missing org.jruby.RubyHash.default ... in : \n" + fullTrace, hash_default);
+        assertTrue("missing org.jruby.RubyHashLinkedBuckets.default ... in : \n" + fullTrace, hash_default);
     }
 
     public void testRubyExceptionUsingEmbedAdapter() {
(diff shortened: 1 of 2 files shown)


headius:
Given the immediate problem of jruby-openssl calling one of the RubyHash constructors, I have devised a way to allow it to still be constructed.

* RubyHash aggregates a single Object state field that can be used in place or by subclasses.
* For RubyHash constructed directly, the state field will hold a delegated RubyHashLinkedBuckets and all formerly abstract methods will call through that delegate.
* For other subclasses like RubyHashLinkedBuckets, the state field will hold some piece of implementation state and be used normally by that implementation's methods.

This allows constructing RubyHash directly (via deprecated constructors) while still allowing all of our internal uses to construct one of the available subclasses.

After some time (perhaps 10.2) and all users of the RubyHash constructors have been fixed, the state field should be moved down into each implementation and the RubyHash class and delegated methods returned to abstract.

Issue #9255, opened 2026-02-21
jruby 10.0.3.0 on Linux 64bit - can not have non-empty directory in paths


rubyFeedback:
I just downloaded the latest binary release for Linux from the github page here.

The version is reported as:

    jruby 10.0.3.0 (3.4.5) 2026-02-02 b0be2ab713 OpenJDK 64-Bit Server VM 25.0.1+8-jvmci-b01 on 25.0.1+8-jvmci-b01 [x86_64-linux]

On the commandline. (I actually thought I had GraalVM in use; not sure right now, need to check this after filing this issue report.)

Oddly enough I get this strange error:

    Error occurred during CDS dumping
    Cannot have non-empty directory in paths

I don't know what this error message means.

But, may I make a suggestion here: the "cannot have non-empty directory in paths"
is not that helpful. Could it be shown which path is meant exactly? I am also uncertain
why non-empty directories are a problem. Is this new? I don't recall this being reported
in the past. Right now I am not sure what to do with this; first I have to check on my local
java installation too. (I typically symlink things, perhaps this is confusing jruby.) Either
way I think the notification is a bit weird, but it could be made better by adding an
additional message such as:

    /tmp/garfield/ is non-empty. # or empty ... or whatever the message is trying to convey

That way I, as the user, could look at that directory and perhaps remove it and have
jruby work. Or perhaps it is just a warning but then why is this given. Is it an important
notification actually?

Edit: So jruby appears to work fine, I just tested hello_world.rb and it works. It only outputs
those two lines afterwards:

    Error occurred during CDS dumping
    Cannot have non-empty directory in paths

So it is probably harmless, but I think older jruby did not show that behaviour. Or at the least I may not have noticed it; I manage to end up breaking my own systems all of the time so my own memory
my fail me here, but I am pretty certain I did not have this issue with older jruby version. I may
test some of them soon.

rubyFeedback:
Ok I did a bit more detective work.

jruby 10.0.2.0 shows another behaviour:

jruby --version

    /home/Programs/Graalvm/25.0.1-64bit/release: line 11: {commit.committer:: command not found
    /home/Programs/Jruby/10.0.2.0/bin/jruby.sh: line 532: [: : integer expected
    /home/Programs/Jruby/10.0.2.0/bin/jruby.sh: line 541: [: : integer expected
    /home/Programs/Jruby/10.0.2.0/bin/jruby.sh: line 552: [: : integer expected
[...]

PR #9272, opened 2026-02-28
Avoid passing bogus -classpath when empty

headius:
If no CLASSPATH env is provided when running jruby.sh, we end up building a bogus classpath flag of `-classpath :`. Other situations may result in forms like `-classpath some.jar:` or `-classpath :some.jar`. The changes here avoid passing -classpath at all if empty, and only insert delimiters when concatenating classpaths if they are non-empty.

Possible fix for https://github.com/jruby/jruby/issues/9255

### DIFF
--- a/bin/jruby.sh
+++ b/bin/jruby.sh
@@ -997,26 +997,22 @@ fi
 
 prepend java_args "$JAVACMD"
 
+classpath_value=$CLASSPATH
 if $NO_BOOTCLASSPATH || $VERIFY_JRUBY; then
     if $use_modules; then
         # Use module path instead of classpath for the jruby libs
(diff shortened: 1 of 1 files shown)


headius:
@mrnoname1000 This appears to work fine but it's a little redundant in places. Maybe you have a suggestion to clean it up?

headius:
Marked for 10.0.4.0 but probably should be backported to 9.4 (at least until we can make it upgradable after release).

mrnoname1000:
@headius I rearranged the logic to be a little simpler, opened a PR into this one here: https://github.com/headius/jruby/pull/7

headius:
@mrnoname1000 Thank you!

PR #9285, opened 2026-03-03
Clear callInfo so it does not taint other calls

headius:
This is the trivial fix for jruby/jruby#9281. A separate PR in jruby/jruby#9284 did not correctly untangle the kwarg handling logic, so we're pushing it to a future release.

Fixes jruby/jruby#9281


headius:
This got half merged during the recent GitHub outage so we are forced to close it but leave it targeted at a release. It has been correctly merged, but the PR state is broken.

PR #9289, opened 2026-03-03
Include gem extension stubs in stdlib artifacts

headius:
The extension stubs installed for JRuby versions of extension gems must be included in the stdlib, dist, and complete artifacts or those libraries will fail to load. This is the cause of issues mentioned in the Docker official-images PR here:

https://github.com/docker-library/official-images/pull/20939#issuecomment-3992984619

The jruby-dist and jruby-complete artifacts get their stdlib content from the jruby-stdlib artifact, so including the files that artifact is necessary before including them in the other two. jruby-complete just packages the contents of jruby-stdlib as-is, but the jruby-dist artifact needs to filter and copy them back into their dist locations.


headius:
Dist verification was improved in #9290 to detect this situation in the future.

PR #9302, opened 2026-03-12
review call-info reset

kares:
trying to get the `keywords = true` aligned with proper `resetCallInfo`, where it's needed (follow-up on https://github.com/jruby/jruby/pull/9285).

on some places this won't trigger an issue unless something changes in the method (would be easy to miss).

but there also  are places where `ArgumentError` reproduces easily (similar to https://github.com/jruby/jruby/issues/9281) e.g. 

```ruby
# On JRuby 10.0.4.0: the 1-arg, 2-arg, and 3-arg forms all leak

class Victim
  def initialize(a, k:)
    @a = a
    @k = k
  end
  def result; [@a, @k]; end
end

puts "Struct#initialize callInfo leak test:"

# Control: no struct creation, no leak
begin
  v = Victim.new(1, k: 2)
  puts "  Control (no Struct):       #{v.result.inspect}"
rescue ArgumentError => e
  puts "  UNEXPECTED: #{e.message}"
end

# Test 1-arg form: S.new(val, **{})
S1 = Struct.new(:x)
begin
  S1.new(1, **{})
  v = Victim.new(1, k: 2)
  puts "  After S1.new(1, **{}):     #{v.result.inspect}"
rescue ArgumentError => e
  puts "  BUG (1-arg form leak):     #{e.message}"
end

# Test 2-arg form: S.new(v1, v2, **{})
S2 = Struct.new(:x, :y)
begin
  S2.new(1, 2, **{})
  v = Victim.new(1, k: 2)
  puts "  After S2.new(1, 2, **{}):  #{v.result.inspect}"
rescue ArgumentError => e
  puts "  BUG (2-arg form leak):     #{e.message}"
end

# Test 3-arg form: S.new(v1, v2, v3, **{})
S3 = Struct.new(:x, :y, :z)
begin
  S3.new(1, 2, 3, **{})
  v = Victim.new(1, k: 2)
  puts "  After S3.new(1,2,3, **{}): #{v.result.inspect}"
rescue ArgumentError => e
  puts "  BUG (3-arg form leak):     #{e.message}"
end
```
~~the above isn't completely fixed due being quite specific and reveals an issue with `IRRuntimeHelper.setCallInfo` which [retains the empty flag](https://github.com/jruby/jruby/blob/jruby-10.0/core/src/main/java/org/jruby/ir/runtime/IRRuntimeHelpers.java#L903)~~:
```java
    @JIT @Interp
    public static void setCallInfo(ThreadContext context, int flags) {
        // FIXME: This may propagate empty more than the current call?   empty might need to be stuff elsewhere to prevent this.
        context.callInfo = (context.callInfo & CALL_KEYWORD_EMPTY) | flags;
    }
```
~~think that should only be used in specific circumstances as the retaining causes issues (such as above)~~.

*~~not 100% why it's kept (esp. the JIT parts) yet and whether simply changing to `context.callInfo = flags` would not cause more harm~~* for now this is not needed, the above bug was resolved.

### DIFF
--- a/test/jruby/test_call_info.rb
+++ b/test/jruby/test_call_info.rb
@@ -115,6 +115,26 @@ class TestCallInfo < Test::Unit::TestCase
     assert_equal [1, 2], [v.a, v.k]
   end
 
+  # ── Regexp#initialize keyword handling ───────────────────────────
+
+  def test_regexp_initialize_with_timeout_keyword
+    # 2-arg form: Regexp.new(source, timeout: val)
+    re = Regexp.new("abc", timeout: 1.0)
+    assert_match re, "abc"
+
+    v = KwargsReceiver.new(1, k: 2)
+    assert_equal [1, 2], [v.a, v.k]
+  end
+
+  def test_regexp_initialize_3arg_with_timeout_keyword
+    # 3-arg form: Regexp.new(source, flags, timeout: val)
+    re = Regexp.new("abc", Regexp::IGNORECASE, timeout: 1.0)
+    assert_match re, "ABC"
+
+    v = KwargsReceiver.new(1, k: 2)
+    assert_equal [1, 2], [v.a, v.k]
+  end
+
   # ── ARGF keyword handling ──────────────────────────────────────────
 
   def test_argf_readlines_chomp_keyword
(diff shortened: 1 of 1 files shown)


headius:
@kares I think you should give it a try (make it just fully overwrite the flags) and we'll check with @enebo as to why he wrote this to preserve the empty flag. I suspect it was either not intentional or served a purpose that no longer exists.

enebo:
As a comment for posterity.  I think passing callInfo as a value in the two places which introduced the extra passing is baking callInfo into the API which was not a goal of callInfo (it was meant to be a temporary workaround for not changing all call code paths).  That said, those two places are private and can be backed out simply when a longer term solution comes along.  So I was ok with this PR but just thought it is worth mentioning somewhere that the goal was to minimally affect API call paths with callInfo as a value.

PR #9303, opened 2026-03-12
build: bump maven from 3.9.11 to 3.9.14 with script-only wrapper

chadlwilson:
Bumps maven and migrates to the script-only variant; which has been the default for quite some time now.

https://github.com/apache/maven/releases/tag/maven-3.9.14
https://github.com/apache/maven/releases/tag/maven-3.9.13
https://github.com/apache/maven/releases/tag/maven-3.9.12

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -70,9 +70,6 @@ jobs:
   repeat-specs:
     runs-on: ubuntu-latest
 
-    strategy:
-      fail-fast: false
-
     name: run specs repeatedly to test optimizations
 
     steps:
@@ -97,9 +94,6 @@ jobs:
 
   jruby-tests-dev:
     runs-on: ubuntu-latest
-
-    strategy:
-      fail-fast: false
 
     env:
       JRUBY_OPTS: '--dev'
@@ -220,9 +214,6 @@ jobs:
   regression-specs-jit:
     runs-on: ubuntu-latest
 
-    strategy:
-      fail-fast: false
-
     env:
       JRUBY_OPTS: '-Xjit.threshold=0'
 
@@ -246,9 +237,6 @@ jobs:
 
   mvn-test-windows:
     runs-on: windows-latest
-
-    strategy:
-      fail-fast: false
 
     name: mvn -Ptest, rake test:jruby (Windows, Java 21)
 
@@ -279,9 +267,6 @@ jobs:
   spec-ruby-language-windows:
     runs-on: windows-latest
 
-    strategy:
-      fail-fast: false
-
     name: language specs (Windows, Java 21)
 
     env:
@@ -310,9 +295,6 @@ jobs:
 
   spec-ruby-core-windows:
     runs-on: windows-latest
-
-    strategy:
-      fail-fast: false
 
     name: core specs (Windows, Java 21)
 
@@ -364,9 +346,6 @@ jobs:
   test-versions:
     runs-on: ubuntu-latest
 
-    strategy:
-      fail-fast: false
-
     name: Verify build artifacts
     steps:
       - name: checkout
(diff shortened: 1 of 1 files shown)


chadlwilson:
@kares thanks for the re-run of the flaky test!

Not sure if you had a look at this “in progress” work, but I can probably split it in two, since I extended to some GHA workflow cleanups/fixes for warnings and potentially insecure outdated actions.

Also, would it better for me to target 10.0 branch and then it gets merged to master through a “normal” regular practice? I’m a bit confused as to the branch policy here (merges, direction of merge, when cherry-picks are used vs not used etc).

The maven wrappers are hideously outdated across jruby repos. I actually question the wisdom of maven wrapper these days, given dependabot still can’t even help update them. But I left it in place here  as it probably needs a wider discussion.

kares:
Hey Chad, this LGTM.

Not sure why the size checks started failing:
```
size of lib/target/jruby-stdlib-10.1.0.0-SNAPSHOT.jar expected smaller than 20*1024*1024 but got 39124856
size of maven/jruby-jars/pkg/jruby-jars-10.1.0.0.SNAPSHOT.gem expected smaller than 35*1024*1024 but got 53663744
size of maven/jruby-jars/lib/jruby-stdlib-10.1.0.0-SNAPSHOT.jar expected smaller than 20*1024*1024 but got 39124856
size of maven/jruby-complete/target/jruby-complete-10.1.0.0-SNAPSHOT.jar expected smaller than 36*1024*1024 but got 56169239
size of maven/jruby-dist/target/jruby-dist-10.1.0.0-SNAPSHOT-bin.tar.gz expected smaller than 45*1024*1024 but got 56109678
size of maven/jruby-dist/target/jruby-dist-10.1.0.0-SNAPSHOT-bin.zip expected smaller than 45*1024*1024 but got 62331789
```
might be due including ruby-maven which adds like 20MB :shrug: 

---

> Also, would it better for me to target 10.0 branch and then it gets merged to master through a “normal” regular practice? I’m a bit confused as to the branch policy here (merges, direction of merge, when cherry-picks are used vs not used etc).

That's probably still the usual flow but I would check on Matrix.
There might be a push back if there's a risk of some CI instability. 

> The maven wrappers are hideously outdated across jruby repos. I actually question the wisdom of maven wrapper these days, given dependabot still can’t even help update them. But I left it in place here as it probably needs a wider discussion.

No idea about that, or dependabot really :sweat_smile: 

I don't think anyone cares about keeping `mvnw` up-to-date until some issues show up (has been the case in the past), so IMHO updating like you did here is a good idea.

chadlwilson:
@kares yeah it’s failing because of ruby-maven - it’s failing on master and was mentioned on Matrix. Can ignore that one.

People really “should” care about mvn/mvnw a bit more than they do, I think - vulnerabilities in build tools are potential supply chain vulnerabilities with something like jruby, and can impact reproducibility. Admittedly it’s less of a concern for maven (than other build tools) if you lock all the plugin versions in pom.xml and use dependabot/renovate to upgrade those. Mvnw helps ensure local and CI are in sync, builds always reproducible over time, and for new folks to bootstrap themselves - but becomes stale/risky without automation. Some jruby projects use versions that have been EOL for 9 years. 

Without the wrapper, you just use `mvn` on GitHub actions and at least you get the latest version the same everywhere across every project at a single point in time - and don’t have any version maintenance overhead. But loss of reproducibility over time is a big loss.

The only thing worse (imho) is doing both, where you lose both reproducibility and add security risk. However there are some special concerns for jruby core because of ruby-maven, maven-tools (and friends) and the embedded maven. I might have a partial blind spot here.

When Maven 4 comes out (likely soon) perhaps there’ll be more enthusiasm for the wrapper again since locking version will be more important for both compatibility and reproducibility reasons, after a very long period of little change within Maven.

Issue #9187, opened 2026-01-27
Memory leak running RubyGems in separate runtimes in a loop


headius:
I'm noticing a lot of memory leaking when running the following looped benchmark of RubyGems booting:

```
jruby -Xparser.summary -Xparser.prism.wasm -e 'loop { t = Time.now; ruby = org.jruby.Ruby.newInstance; ruby.loadService.require("rubygems"); ruby.tearDown; $stderr.puts Time.now - t }'
```

This quickly goes up past 1GB heap on my machine and doesn't seem to slow down or respond to forced GCs.

When I looked into this in the past it seemed like the leak might be related to jruby-openssl and the BouncyCastle provide getting re-registered many times and never cleaned up, but a new analysis is needed.

headius:
The fixes from #9315 (minus the bad AI onces and including my alternatives) appear to fix this issue.

Modified reproduction script:

```ruby
loop {
  t = Time.now
  ruby = org.jruby.Ruby.newInstance
  ruby.loadService.require("rubygems")
  ruby.tearDown
  JRuby.gc
  print "."
}
```

Ten minutes monitoring heap shows the leak appears to be gone:

<img width="494" height="304" alt="Image" src="https://github.com/user-attachments/assets/d6833099-f84a-4abe-8909-6b113ccea160" />

headius:
Fixed by #9359.

PR #9315, opened 2026-03-17
Fix runtime and classloader retention after ScriptingContainer terminate

### LINKED ISSUE #9070: Memory leak in jruby within openhab
Hi team,

I am running some very simple jruby automation scripts within openhab 4.1.3 on RPi 4 and am experiencing what appears to be some memory leaks within jruby as time goes on.  

Linux hapi 5.10.103-v7l+ #1529 SMP Tue Mar 8 12:24:00 GMT 2022 armv7l GNU/Linux

jruby version is 9.4.9.0 

I am not using any gems other than openhab-jrubyscripting to interface with openhab

If left unchecked, the JVM will run out of memory within a week in my RPi setup.   Inspecting heap dumps, several possible leaks are identified which appear to be within jruby.  Details below;

Problem Suspect 1
29,102 instances of org.jruby.RubyClass, loaded by org.openhab.automation.jrubyscripting occupy 195,511,120 (19.70%) bytes.
org.jruby.RubyClass
org.openhab.automation.jrubyscripting

Problem Suspect 2
71,150 instances of org.jruby.ir.IRMethod, loaded by org.openhab.automation.jrubyscripting occupy 148,220,520 (14.93%) bytes.
org.jruby.ir.IRMethod
org.openhab.automation.jrubyscripting

Problem Suspect 3
33 instances of org.jruby.Ruby, loaded by org.openhab.automation.jrubyscripting occupy 118,031,520 (11.89%) bytes.
org.jruby.Ruby
org.openhab.automation.jrubyscripting

I am happy to provide heap dumps privately for any jruby devs that would like to analyse.

Here is the script which I'm using:

<pre>
# automation of HVAC based on astro, temp & solar forecasts
#

ITEM_DEFAULT_VALUES = {
  Hvac_Solar_Power_Threshold_Dusk    => 2000 | 'W',
  Hvac_Solar_Power_Threshold_Dawn    => 3000 | 'W',
  Hvac_Solar_Override_Trigger_Cool   => 9 | '°C',
  Hvac_Solar_Override_Trigger_Heat   => 34 | '°C',
  Hvac_Solar_Override_Normal_Cool    => 13 | '°C',
  Hvac_Solar_Override_Normal_Heat    => 28 | '°C'
}

DAWN_AVG_MINS = 10
DUSK_AVG_MINS = 10

shared_cache.compute_if_absent(:hvacstate) { OFF }
shared_cache.compute_if_absent(:hvacsolaroverride) { OFF }

def hvac_log(msg)
  logger.info "HVAC: UN(#{Airtouch_Unit_Fujitsu_Power.state}) AS(#{shared_cache[:hvacstate]}): #{msg}"
  Airtouch_Unit_Fujitsu_Automation_Message.update(msg)
end

def deploy_scenes(power_scene, setpoint_scene)
  hvac_log sprintf("deploy_scenes(#{power_scene}, #{setpoint_scene})")
  active_scenes = [power_scene, setpoint_scene]
  gAirtouchZones.members.each do |zone_group|
    zone_group.tags.sort.each do |tag|
      if tag =~ /Fujitsu_Scene_(\S+?)\[(\S+)\]\[(\S+)\]$/
        scene, tag, state = $1, $2, $3
        if active_scenes.include? scene
          zone_group.members.flatten.sort_by(&:name).tagged(tag).each do |item| 
            hvac_log sprintf("[Scene_Item_State] %-40s | %-10s | %-10s --> %s", item.name, scene, tag, state)
            item.ensure << state
          end
        end
      end
    end
  end
end

def determine_automation_scenes(mode)
  sunrise_time   = Sunrise_Time.state
  sunset_time    = Sunset_Time.state
  tnow           = Time.now

  # Determine POWER scene
  case tnow.wday
  when 5 # fri
    power_scene = "AllOn"
  when 6, 0 # sat, sun
    power_scene = "Weekend"
  else
    power_scene = "Workday"
  end
  power_scene = 'Night' if  Time.now < (sunrise_time - '1 hours') || Time.now > (sunset_time + '5 hours')

  # Determine SETPOINT scene
  case mode
  when 'HEAT'
    setpoint_scene = 'Winter'
  when 'COOL'
    setpoint_scene = 'Summer'
  end
  #case Season_Name.state
  #when 'WINTER','AUTUMN'
  #  setpoint_scene = 'Winter'
  #when 'SPRING','SUMMER'
  #  setpoint_scene = 'Summer'
  #end

  [power_scene, setpoint_scene]
end

rule "hvac_startup" do
  on_start at_level: 80
  run do
    for item, value in ITEM_DEFAULT_VALUES
      item.state = value if item.state.nil?
    end
  end
end

rule "hvac_scene_changed" do
  changed Airtouch_Unit_Fujitsu_Scene_Power
  changed Airtouch_Unit_Fujitsu_Scene_Setpoint
  run do
    deploy_scenes(Airtouch_Unit_Fujitsu_Scene_Power.state, Airtouch_Unit_Fujitsu_Scene_Setpoint.state)
  end
end

rule "hvac_event_off" do
  changed Airtouch_Unit_Fujitsu_Power, to: OFF
  run do
    shared_cache[:hvacstate] = OFF
    shared_cache[:hvacsolaroverride] = OFF
    logger.info "HVAC: was turned OFF"
  end
end

# Force the HVAC off at sunrise, assuming it was manually turned on beforehand.
rule "hvac_sunrise_force_off" do
	channel "astro:sun:home:rise#event", attach: "sunrise"   
  run do
    hvac_log "Turning unit off due to SUNRISE event"
    Airtouch_Unit_Fujitsu_Power.ensure.off
  end
end
#rule "hvac_morning_force_off" do
#  every :day, at: '07:00'
#  run do
#    hvac_log "Turning unit off due to reached 7am"
#    Airtouch_Unit_Fujitsu_Power.ensure.off
#  end
#end

rule "hvac_automation" do
  every 1.minutes
  run do |event|
    solar_remain   = ForecastSolar_9Hens_Remaining.state
    inlet_temp     = Airtouch_Unit_Fujitsu_Inlet_Temp.state
    max_temp_today = BOM_Day1_MaximumTemperature.state
    cur_temp       = BOM_Day1_AirTemperature.state
    apparent_temp  = BOM_Day1_ApparentTemperature.state

    # Use forecasts and inlet temp to determine if we should HEAT or COOL
    if max_temp_today <= 20 || apparent_temp <= 7 || (inlet_temp <= 12 | '°C')
      mode = 'HEAT'
    elsif max_temp_today >= 27 || apparent_temp >= 28 || (inlet_temp >= 28 | '°C')
      mode = 'COOL'
    else
      mode = nil
    end

    # Overrides for normal solar automation.  Use inlet temp as gauge of general comfort level
    if (inlet_temp <= Hvac_Solar_Override_Trigger_Cool.state | '°C') || (inlet_temp >= Hvac_Solar_Override_Trigger_Heat.state | '°C')
      shared_cache[:hvacsolaroverride] = ON
    end

    if Airtouch_Unit_Fujitsu_Automation_Override.state == ON
      hvac_log "Manual override active"
    else
      if Airtouch_Unit_Fujitsu_Power.state == ON
        if mode.nil?
          hvac_log "Environmentals outside of automation bounds"
          if shared_cache[:hvacstate] == ON
            hvac_log "Temperatures (cur:#{cur_temp} inlet:#{inlet_temp} max:#{max_temp_today} app:#{apparent_temp}) outside of automation bounds"
            hvac_log "turning unit OFF due to automation state being ON"
            Airtouch_Unit_Fujitsu_Power.ensure.off
          end
        else
          hvac_log "Sunset = #{Sunset_Time.state}, Season = #{Season_Name.state}, Mode = #{mode}, SolarRemain = #{solar_remain}"
          if shared_cache[:hvacstate] == ON
            if shared_cache[:hvacsolaroverride] == ON
              if ((mode == 'COOL' && inlet_temp <= Hvac_Solar_Override_Normal_Heat.state | '°C') || 
                  (mode == 'HEAT' && inlet_temp >= Hvac_Solar_Override_Normal_Cool.state | '°C'))
                hvac_log "turn unit off NOW:  SOLAR OVERRIDE IS ACTIVE AND INLET TEMP #{inlet_temp} IS NOW WITHIN COMFORTABLE RANGE"
                Airtouch_Unit_Fujitsu_Power.ensure.off
              end
            else
              sunset_avg_solar_power = Solar_Total_Gen_Power.average_since(DUSK_AVG_MINS.minutes.ago)
              if (solar_remain && (solar_remain < 3 | 'kWh')) || ((sunset_avg_solar_power < Hvac_Solar_Power_Threshold_Dusk.state | 'W') || (Time.now > Sunset_Time.state))
                hvac_log "Solar energy (#{sunset_avg_solar_power} / 10min) too low OR we're past sunset time:  TURN HVAC UNIT OFF NOW"
                Airtouch_Unit_Fujitsu_Power.ensure.off
              end
            end
          end
        end
      else
        if mode.nil?
          hvac_log "Temperatures (cur:#{cur_temp} inlet:#{inlet_temp} max:#{max_temp_today} app:#{apparent_temp}) outside of automation bounds"
        elsif shared_cache[:hvacsolaroverride] == ON || Time.now > Sunrise_Time.state
          sunrise_avg_solar_power = Solar_Total_Gen_Power.average_since(DAWN_AVG_MINS.minutes.ago)
          hvac_log "Sunrise = #{Sunrise_Time.state}, Season = #{Season_Name.state}, Mode = #{mode}, SolarGen(30mAVG) = #{sunrise_avg_solar_power}, SolarRemain = #{solar_remain}, Temp = cur:#{cur_temp} inlet:#{inlet_temp} max:#{max_temp_today} app:#{apparent_temp}"
          if shared_cache[:hvacsolaroverride] == ON || (sunrise_avg_solar_power > Hvac_Solar_Power_Threshold_Dawn.state | 'W') || ((sunrise_avg_solar_power > Hvac_Solar_Power_Threshold_Dawn.state | 'W') && solar_remain && (solar_remain > 20 | 'kWh'))
            if shared_cache[:hvacsolaroverride] == ON
              hvac_log "turn unit on NOW:  SOLAR OVERRIDE IS ACTIVE BECAUSE INLET TEMP #{inlet_temp} IS OUT OF COMFORTABLE RANGE !!!"
            else
              hvac_log "turn unit on NOW:  Average solar generation / solar forecast / temp forecast exceeded thresholds"
            end
            power_scene, setpoint_scene = determine_automation_scenes(mode)
            Airtouch_Unit_Fujitsu_Scene_Power.update(power_scene)
            Airtouch_Unit_Fujitsu_Scene_Setpoint.update(setpoint_scene)
            Airtouch_Unit_Fujitsu_Mode.ensure.command(mode)
            Airtouch_Unit_Fujitsu_Power.ensure.on
            hvac_log "Starting timer to set automation state" 
            sleep 10
            shared_cache[:hvacstate] = ON
            hvac_log "Mode is #{Airtouch_Unit_Fujitsu_Mode.state}, :hvacstate is #{shared_cache[:hvacstate]}, hvacsolaroverride is #{shared_cache[:hvacsolaroverride]}"
          end
        end
      end
    end
  end
end
</pre>

### LINKED ISSUE #9092: ThreadLocal Memory Leak in JRuby 10.0.2.0 with ScriptingContainer
# ThreadLocal Memory Leak in JRuby 10.0.2.0 with ScriptingContainer

## Summary

JRuby 10.0.2.0 has a memory leak when using `ScriptingContainer` with proper cleanup. Despite calling the recommended cleanup sequence (`unregisterCurrentThread()`, `terminate()`, `tearDown(true)`, `clearGlobalRuntime()`), Ruby Runtime objects accumulate indefinitely. This regression from JRuby 9.3.1.0 makes long-running applications that repeatedly create and destroy ScriptingContainer instances infeasible.

## Environment

- **JRuby Version**: 10.0.2.0 (regression from 9.3.1.0)
- **Java Version**: 21.0.8 (OpenJDK)
- **OS**: Linux (also reproducible on other platforms)
- **Scope**: `LocalContextScope.SINGLETHREAD`

## Minimal Reproducible Test Case

A complete standalone test is available that demonstrates the issue:

### Test Files
1. **JRubyMemoryLeakTest.java** - Minimal reproduction (attached below)
2. **test_jruby_leak.sh** - Test runner script (attached below)

### How to Reproduce

```bash
# Test with JRuby 9.3.1.0 (baseline, no leak)
./test_jruby_leak.sh 9 50

# Test with JRuby 10.0.2.0 (shows leak)
./test_jruby_leak.sh 10 50
```

### JRuby 9.3.1.0 (Baseline - NO LEAK)

```
Iterations: 50
Ruby Runtime instances: 5 (STABLE)
Total JRuby objects: 130,443 instances, 6.3 MB
Heap memory: ~37 MB (STABLE after warmup)
```

**Memory profile:**
- Initial: 5 MB
- After 10 iterations: 28 MB
- After 50 iterations: 37 MB (STABLE)

### JRuby 10.0.2.0 (MEMORY LEAK)

```
Iterations: 50
Ruby Runtime instances: 50 (1 PER ITERATION - LEAK!)
Total JRuby objects: 1,206,254 instances, 61.5 MB (9.2x more!)
Heap memory: 99 MB (GROWING LINEARLY)
```

**Memory profile:**
- Initial: 5 MB
- After 10 iterations: 28 MB
- After 50 iterations: 99 MB (GROWING ~2MB per iteration)

**Critical findings:**
1. **Ruby Runtime leak**: 50 instances after 50 iterations (1:1 ratio) vs 5 in JRuby 9.3.1.0
2. **Object accumulation**: 9.2x more JRuby objects retained
3. **Linear memory growth**: ~2 MB per iteration
4. **ScriptingContainer cleanup works**: 0 instances remaining

## Cleanup Code Used

We follow the recommended cleanup pattern, including the critical step of unregistering ThreadContext BEFORE calling tearDown():

```java
Ruby runtime = sc.getProvider().getRuntime();

if (runtime != null) {
    // CRITICAL: Unregister current thread BEFORE tearDown
    // tearDown() replaces ThreadService, must clean up old one first
    ThreadContext currentContext = runtime.getThreadService().getCurrentContext();
    runtime.getThreadService().unregisterCurrentThread(currentContext);
}

// Terminate the container
sc.terminate();

// Force runtime teardown
if (runtime != null) {
    runtime.tearDown(true);
}

// Clear global runtime reference
Ruby.clearGlobalRuntime();
```

This cleanup sequence works perfectly in JRuby 9.3.1.0 but fails to cleanup properly in 10.0.2.0.

## Impact

This regression affects applications that:
- Repeatedly create/destroy ScriptingContainer instances
- Run long-lived server processes
- Use JRuby as an embedded scripting engine

The leak causes eventual heap exhaustion in long-running applications.

## Request

Could the JRuby team please investigate the ThreadLocal cleanup changes between 9.3.1.0 and 10.0.2.0? The minimal test case provided should make it straightforward to reproduce and debug with a memory profiler.

---

## Attachment 1: JRubyMemoryLeakTest.java

```java
import org.jruby.Ruby;
import org.jruby.embed.LocalContextScope;
import org.jruby.embed.ScriptingContainer;

import java.lang.management.ManagementFactory;
import java.lang.management.MemoryMXBean;
import java.lang.management.MemoryUsage;

/**
 * Minimal test demonstrating ThreadLocal memory leak in JRuby 10.0.2.0
 *
 * Creates ScriptingContainer instances with proper cleanup sequence.
 * Despite following recommended cleanup, JRuby 10.0.2.0 leaks Ruby Runtime objects.
 */
public class JRubyMemoryLeakTest {

    static {
        // Disable bundler and RubyGems to isolate from system Ruby
        System.setProperty("jruby.cli.load.gemfile", "false");
        System.setProperty("jruby.cli.rubygems.enable", "false");
    }

    public static void main(String[] args) throws Exception {
        int iterations = Integer.parseInt(args.length > 0 ? args[0] : "10");

        System.out.println("JRuby ThreadLocal Memory Leak Test");
        System.out.println("===================================");
        System.out.println("JRuby Version: " + org.jruby.runtime.Constants.VERSION);
        System.out.println("Java Version: " + System.getProperty("java.version"));
        System.out.println("Iterations: " + iterations);
        System.out.println();

        printMemoryStats("BEFORE");

        for (int i = 1; i <= iterations; i++) {
            runScriptWithProperCleanup();

            if (i % 5 == 0) {
                System.gc();
                Thread.sleep(100);
                printMemoryStats("After iteration " + i);
            }
        }

        System.out.println("\n=== Final Heap Analysis ===");
        printObjectCounts();
    }

    /**
     * Executes Ruby code with proper cleanup sequence:
     * 1. unregisterCurrentThread() BEFORE tearDown
     * 2. terminate()
     * 3. tearDown(true)
     * 4. clearGlobalRuntime()
     */
    private static void runScriptWithProperCleanup() {
        ScriptingContainer sc = null;
        try {
            // Create container with SINGLETHREAD scope to avoid cross-thread issues
            sc = new ScriptingContainer(LocalContextScope.SINGLETHREAD);

            // Isolate from system Ruby environment
            sc.getEnvironment().put("GEM_HOME", "");
            sc.getEnvironment().put("GEM_PATH", "");
            sc.getEnvironment().put("BUNDLE_GEMFILE", "");

            // Execute some Ruby code (simulates real application usage)
            sc.runScriptlet("" +
                "def fibonacci(n)\n" +
                "  return n if n <= 1\n" +
                "  fibonacci(n-1) + fibonacci(n-2)\n" +
                "end\n" +
                "\n" +
                "result = fibonacci(10)\n" +
                "result\n"
            );

            // Proper cleanup sequence
            Ruby runtime = sc.getProvider().getRuntime();

            if (runtime != null) {
                // CRITICAL: Unregister current thread BEFORE tearDown
                // tearDown() replaces ThreadService, must clean up old one first
                ThreadContext currentContext = runtime.getThreadService().getCurrentContext();
                runtime.getThreadService().unregisterCurrentThread(currentContext);
            }

            // Terminate the container
            sc.terminate();

            // Force runtime teardown
            if (runtime != null) {
                runtime.tearDown(true);
            }

            // Clear global runtime reference
            Ruby.clearGlobalRuntime();

        } catch (Exception e) {
            e.printStackTrace();
        } finally {
            sc = null;
        }
    }

    private static void printMemoryStats(String label) {
        MemoryMXBean memoryBean = ManagementFactory.getMemoryMXBean();
        MemoryUsage heapUsage = memoryBean.getHeapMemoryUsage();

        System.out.printf("%s: Heap Used: %.2f MB, Committed: %.2f MB\n",
            label,
            heapUsage.getUsed() / 1024.0 / 1024.0,
            heapUsage.getCommitted() / 1024.0 / 1024.0
        );
    }

    private static void printObjectCounts() {
        try {
            String pid = ManagementFactory.getRuntimeMXBean().getName().split("@")[0];

            System.out.println("\n=== ScriptingContainer Count ===");
            ProcessBuilder pb1 = new ProcessBuilder(
                "bash", "-c",
                "jmap -histo:live " + pid + " | grep 'org.jruby.embed.ScriptingContainer' || echo 'No instances found'"
            );
            pb1.inheritIO();
            pb1.start().waitFor();

            System.out.println("\n=== Ruby Runtime Count ===");
            ProcessBuilder pb2 = new ProcessBuilder(
                "bash", "-c",
                "jmap -histo:live " + pid + " | grep -E 'org.jruby.Ruby[^a-zA-Z]|org.jruby.Ruby$' || echo 'No instances found'"
            );
            pb2.inheritIO();
            pb2.start().waitFor();

            System.out.println("\n=== Total JRuby Object Count ===");
            ProcessBuilder pb3 = new ProcessBuilder(
                "bash", "-c",
                "jmap -histo:live " + pid + " | grep 'org.jruby' | awk '{sum += $2; bytes += $3} END {print \"Total instances: \" sum \", Total bytes: \" bytes}'"
            );
            pb3.inheritIO();
            pb3.start().waitFor();

            System.out.println("\n=== Top 10 JRuby Classes ===");
            ProcessBuilder pb4 = new ProcessBuilder(
                "bash", "-c",
                "jmap -histo:live " + pid + " | grep 'org.jruby' | head -10"
            );
            pb4.inheritIO();
            pb4.start().waitFor();

        } catch (Exception e) {
            e.printStackTrace();
        }
    }
}
```

## Attachment 2: test_jruby_leak.sh

```bash
#!/bin/bash
#
# Test script to demonstrate JRuby ThreadLocal memory leak
# Usage: ./test_jruby_leak.sh [9|10] [iterations]
#

set -e

SCRIPT_DIR="$(cd "$(dirname "${BASH_SOURCE[0]}")" && pwd)"
JRUBY_VERSION=${1:-10}
ITERATIONS=${2:-10}

case $JRUBY_VERSION in
    9)
        JRUBY_JAR="$SCRIPT_DIR/jruby-complete-9.3.1.0.jar"
        ;;
    10)
        JRUBY_JAR="$SCRIPT_DIR/jruby-complete-10.0.2.0.jar"
        ;;
    *)
        echo "Usage: $0 [9|10] [iterations]"
        echo "  9  - Test with JRuby 9.3.1.0"
        echo "  10 - Test with JRuby 10.0.2.0"
        echo "  iterations - Number of ScriptingContainer cycles (default: 10)"
        exit 1
        ;;
esac

if [ ! -f "$JRUBY_JAR" ]; then
    echo "ERROR: JRuby JAR not found: $JRUBY_JAR"
    echo ""
    echo "Please download:"
    echo "  JRuby 9.3.1.0: https://repo1.maven.org/maven2/org/jruby/jruby-complete/9.3.1.0/jruby-complete-9.3.1.0.jar"
    echo "  JRuby 10.0.2.0: https://repo1.maven.org/maven2/org/jruby/jruby-complete/10.0.2.0/jruby-complete-10.0.2.0.jar"
    exit 1
fi

echo "========================================"
echo "JRuby Memory Leak Test"
echo "========================================"
echo "JRuby Version: $JRUBY_VERSION"
echo "JRuby JAR: $JRUBY_JAR"
echo "Iterations: $ITERATIONS"
echo "========================================"
echo

# Compile the test
echo "Compiling JRubyMemoryLeakTest.java..."
javac -cp "$JRUBY_JAR" "$SCRIPT_DIR/JRubyMemoryLeakTest.java"

if [ $? -ne 0 ]; then
    echo "ERROR: Compilation failed"
    exit 1
fi

echo "Compilation successful"
echo

# Run the test with reasonable heap size
echo "Running test with $ITERATIONS iterations..."
echo
java -cp "$JRUBY_JAR:$SCRIPT_DIR" \
    -Xms512m \
    -Xmx2g \
    -XX:+UseG1GC \
    JRubyMemoryLeakTest \
    "$ITERATIONS"

echo
echo "========================================"
echo "Test completed"
echo "========================================"
echo
echo "To compare results:"
echo "  JRuby 9.3.1.0: ./test_jruby_leak.sh 9 50"
echo "  JRuby 10.0.2.0: ./test_jruby_leak.sh 10 50"
```

---

**Note:** The test is self-contained with no external dependencies beyond JRuby JAR files. Results are reproducible across different systems and Java versions.

jimtng:
Clear lingering ThreadContext soft references during teardown, drop the embed LocalContext runtime reference, and avoid re-adopting the current thread while Ruby is shutting down.

Also switch Java integration class caches to weak values and avoid recreating JavaSupport during teardown so terminated runtimes and their JRubyClassLoaders can be collected promptly.

Add regression coverage for same-thread and worker-thread container termination.

Fixes #9092, Fixes #9070, Fixes openhab/openhab-jruby#520

---

Note: I asked an AI tool to help identify this issue. I verified the result on a patched JRuby 10.0.4.0 running inside openHAB, where the memory leak had previously been observed. After applying the fix, the GC consistently returned the heap to its original size, even after repeated script‑engine creation and teardown cycles.

---

Note: In order to get the JVM to aggressively perform GC, during testing I've started it with:

```
 -XX:+UseG1GC -XX:G1PeriodicGCInterval=1000 -XX:-G1PeriodicGCInvokesConcurrent -XX:MinHeapFreeRatio=5 -XX:MaxHeapFreeRatio=10 -XX:SoftRefLRUPolicyMSPerMB=0
```

### DIFF
--- a/core/src/main/java/org/jruby/util/collections/StableClassValue.java
+++ b/core/src/main/java/org/jruby/util/collections/StableClassValue.java
@@ -49,11 +49,6 @@ final class StableClassValue<T> extends ClassValue<T> {
 
                 result = this.calculator.apply(input);
                 this.result = result;
-                // Null out the calculator so that StableValue no longer holds a reference
-                // chain back to the ClassValue identity (its own weak key in Class.classValueMap).
-                // Without this, the hard value in ClassValueMap would transitively retain the
-                // weak key, making the entry immortal and leaking the Ruby runtime. (GH-9092)
-                this.calculator = null;
             }
 
             return result;


headius:
There's clearly some issues here but I'd like to figure out how to land a fix.

It seems that the weak references to the class values is too weak, and I'm not sure we should be using weak references at this point anyway. If they're weak and a JavaClass goes unreferenced for a couple of GC cycles, it will end up getting collected and a future access will cause it to be recreated. That's the source of all the reinitialized constant warnings and the reason the test/jruby tests fail here.

Even a SoftReference is probably too weak, since it just delays the time until the JavaClass gets collected.

Is that part of the patch necessary for this to work? Is clearing the reference to the JavaSupport class not enough?

jimtng:
> Is that part of the patch necessary for this to work?

Yes it's necessary.

Reverting only the JavaSupport/ClassValue part causes all three GH-9092 regression tests in ScriptingContainerClassLeakTest to fail, and restoring it makes that test class pass again. So that part is necessary for the leak fix, not for normal runtime cache behavior.

If you have a cleaner lifecycle-based alternative, I’m happy to switch to that.

headius:
My question was really if we need *both* changes. Clearing the reference to the JavaSupport instance seems like it would be sufficient, and making the Java class references weak leads to other problems quickly.

I can give it a try myself when I'm in the office.

headius:
My question was really if we need *both* changes. Clearing the reference to the JavaSupport instance seems like it would be sufficient, and making the Java class references weak leads to other problems quickly.

I can give it a try myself when I'm in the office.

jimtng:
I've removed the weak reference and instead, explicitly clear it at teardown. It seems that most tests are passing except that test:mri:core:jit which seem flaky - sometimes java21 failed but java25 passed, and another time java21 passed, but java25 failed. Is this a known issue?

headius:
That particular test does seem to be flaky on and off so I don't think it's related to your changes. I will take another look at that but I think we are probably okay with this set of changes.

core/src/main/java/org/jruby/Ruby.java
```diff
@@ -3384,7 +3385,13 @@ private void systemTeardown(final ThreadContext context) {
...
         constantNameInvalidators.clear();
         symbolTable.clear();
-        javaSupport = loadJavaSupport();
+        if (javaSupport != null) javaSupport.tearDown();
+        javaSupport = null;
+
+        // Shut down and replace thread service after all other hooks, finalizers, and cleanup
+        // steps that touch the thread service have run.
+        threadService.teardown();
+        threadService = new ThreadService(this);
```
kares, line 3394: why the discrepancy between now `nul`l-ing `javaSupport` yet doing a `new ThreadService` here?

core/src/main/java/org/jruby/internal/runtime/ThreadService.java
```diff
@@ -131,6 +133,12 @@ public class ThreadService extends ThreadLocal<SoftReference<ThreadContext>> {
 
     private final AtomicLong threadCount = new AtomicLong(0);
 
+    /**
+     * All SoftReferences ever stored into this ThreadLocal (one per thread that has entered Ruby
+     * space on this runtime).  Tracked so that teardown() can clear their referents even on
+     * threads we cannot schedule work on (adopted / pool threads).
+     */
+    private final List<SoftReference<ThreadContext>> trackedRefs = new CopyOnWriteArrayList<>();
```
kares, line 141: you probably should manually dispose Ruby threads that were adopted, this feels like quite an invasive work-around for a very specific issue? I mean there's usually other approaches reference clearing is done in general with queues. 🤷

core/src/main/java/org/jruby/javasupport/JavaSupport.java
```diff
@@ -66,8 +67,8 @@ public abstract class JavaSupport {
     protected final Ruby runtime;
 
     @Deprecated(since = "9.4.3.0")
-    private final ClassValue<JavaClass> javaClassCache;
-    private final ClassValue<RubyModule> proxyClassCache;
+    private final MapBasedClassValue<JavaClass> javaClassCache;
+    private final MapBasedClassValue<RubyModule> proxyClassCache;
```
kares: now there's no way to switch back to proper `ClassValue`, AFAIU the `MapBasedClassValue` was really a work-around for Java `ClassValue` bugs in some releases.

core/src/main/java/org/jruby/Ruby.java
headius: I understand the confusion here. Usually when I have had teardown reset a particular field (new Whatever) rather than clear it (set to null) it was in service of reusing the overall runtime (in a server-like background JRuby daemon setup) or to allow still-running threads to finish gracefully (which really should not be allowed to happen).

In general, no threads should be accessing this state after (or during) termination, but setting them to null is a pretty severe kill switch. Not sure what we should aim for here given the two possible post-teardown situations.

core/src/main/java/org/jruby/internal/runtime/ThreadService.java
headius: Adopted threads are troublesome... we don't control their lifecycle but we also don't want them to be accessing JRuby runtime state after teardown. My understanding of this code is that it simply releases our references to those threads (which we can control) without trying to wait for them to gracefully terminate (which we cannot control).

core/src/main/java/org/jruby/javasupport/JavaSupport.java
headius: I need to revisit this behavior distinction. I'm not sure where we ended up in discussions with JDK core-libs folks about the weakness (or strength) of the ClassValue references.

This comes back to my concerns about the original super-weak change to the ClassValue that made it too weak and caused classes to get re-proxied repeatedly. If the hard reference is the concern, then clearing the old JavaSupport reference should be enough and further changes should be unnecessary. Perhaps I'm missing some subtlety here.

headius:
I think we're largely here but I echo @kares concerns about a few of these changes. It feels to me (begging pardon if this is not the case) that there's a smattering of related attempts to fix this issue that all got balled up in one PR, when the simple fix of fully clearing the JavaSupport reference might solve all of them at once (and in a similar way to other `tearDown` cleanup).

Let's try to finish this discussion and merge for 10.0.5.0, due out this week.

jimtng:
I've asked copilot to write this response to explain the various changes:

The PR actually started with that simpler approach — the first commit used `javaSupport = null` and weak-valued caches (`ClassValue.newWeakInstance`) — but the three leaking reference chains all need independent fixes, so the changes accumulated.

Here's why `javaSupport = null` alone isn't sufficient:

**Path 1 — `LocalContext` strong reference** (the `LocalContext.java` change)

```
ScriptingContainer → SingleThreadLocalContextProvider → LocalContext → Ruby (strong field)
```

`LocalContext.remove()` never cleared `runtime`. This keeps the entire runtime strongly reachable from the container itself, regardless of what happens to `javaSupport`. Fixed by nulling `runtime` in `LocalContext.remove()`.

**Path 2 — ThreadService `SoftReference` chain** (the `ThreadService.java` change)

```
Thread (GC root) → ThreadLocalMap → SoftReference<ThreadContext> → ThreadContext → Ruby
```

`SoftReference`s are only cleared under heap pressure, which may never occur in a well-resourced long-running server. This specifically affects adopted/pool threads that call into JRuby and then go idle — their `ThreadLocalMap` entry silently anchors the runtime forever. `javaSupport = null` doesn't touch this path; the explicit `trackedRefs` clearing during `teardown()` is what severs it.

**Path 3 — JDK `ClassValue` external root** (the `JavaSupport.java` change)

```
Class<X> (JDK GC root) → java.lang.ClassValue table → computed value → lambda capture → Ruby
```

`ClassValue.newInstance()` delegates to `StableClassValue`, which wraps `java.lang.ClassValue`. That stores computed values *inside the `Class` object itself* via JDK internals — not inside ``JavaSupport``. So setting `javaSupport = null` makes our `ClassValue` instance unreachable from JRuby's side, but the JDK's own table entries (rooted at `Class` objects) can only be cleared by the GC's weak-key processing. The switch to `MapBasedClassValue` eliminates this external root: the entire map is owned through `javaSupport`, so `javaSupport = null` (after `tearDown()`) cuts the chain immediately and deterministically.

---

The `MapBasedClassValue` + explicit `tearDown()` approach (rather than weak values) was chosen specifically in response to earlier reviewer feedback: weak values caused live-runtime cache instability and "already initialized constant" warnings on heavily-threaded reuse.

core/src/main/java/org/jruby/internal/runtime/ThreadService.java
jimtng: We can't manually dispose adopted threads, since we don't own them.

A `ReferenceQueue` won't help here because it delivers notifications *after* GC clears a reference. At teardown time the soft references are still live, that's the whole problem. We need to clear them proactively so the runtime can be collected immediately after `terminate()`, not wait for GC to eventually decide to clear them under heap pressure.

`ref.clear()` is callable from any thread since it just nulls the referent field on a normal heap object, so iterating `trackedRefs` at teardown is the minimal thing that actually works.

kares: okay thanks, still think it's just a partial solution for the adopted problem. but likely worth trying what you have here. :clap: 

the queue was off on my part, did not check out the full code.

a related question, why `SoftReference`  (context is soft wrapped elsewhere and there's a valid reason for that), here weak seems like a better fit not keeping them alive longer; or did I miss anything regarding that as well?

jimtng: Wrapping them in a WeakReference seems better. I've made made the change

jimtng:
I found some more small leaks - patched. 

Now, after 20+ reloads I don't see any more heap growth, other than those caused by soft references, which takes 1000s to be released by the jvm by default.

core/src/main/java/org/jruby/internal/runtime/ThreadService.java
headius: WeakReference is problematic and SoftReference has been used in other places to avoid a given thread suddenly losing identity if it becomes dereferenced by JRuby for a short time. If a JVM thread is being actively used with JRuby, there will generally not be enough time for SoftReference to vacate. WeakReference could vacate immediately, causing a still-active thread to gain a new identity and trigger re-initialization repeatedly.

I think SoftReference is the safer choice.

headius:
@jimtng Thanks for iterating on this.

Regarding the justifications for the additional changes, most seem to make general sense.

I believe the extra effort to clear references throughout the system is appropriate at teardown, and anywhere we are doing so we should be skeptical of decreasing the strength of those references. In other words, if we can actively clear those references and that is sufficient to eventually dereference and collect the entire runtime, no change should be necessary in the strength of those references (Weak or SoftReference additions should not be necessary).

Put yet another way, we should only be using Weak or SoftReference in situations where a still-running JRuby instance does not vacate otherwise unreferenced objects OR if it is not possible at teardown to actively clear those references. This PR is about making teardown clean, so we should focus primarily on that aspect.

This line from Copilot is also not correct (or perhaps it is correct but as a *general statement* about the behavior of GC:

> SoftReferences are only cleared under heap pressure,

This really applies to *all* objects in a GCed system. If there's no pressure to clear the GC, most GCs will not actively move to vacate the heap. You can see this effect by having an app essentially pause all operation and then forcing a GC some time later. In the absence of heap pressure due to ongoing allocation, objects will not be cleared.

SoftReference is essentially just a WeakReference with a time delay from the last traversal. This is typically specified in time versus heap size, like milliseconds per megabyte. It can be tuned up or down, but the intent is that infrequently-accessed objects don't *immediately* get collected like you may see from WeakReference (and we have had bad tests come in from CRuby or ruby/spec that fail to consider this possibility).

> Now, after 20+ reloads I don't see any more heap growth, other than those caused by soft references, which takes 1000s to be released by the jvm by default.

And it appears that all tests are passing, which is a pretty good testament. I will review the latest round of changes and see if I'm comfortable including this in JRuby 10.0.5.0 (which has been delayed until Monday to consider this and other issues).

core/src/main/java/org/jruby/embed/internal/LocalContext.java
```diff
@@ -104,6 +104,12 @@ public void remove() {
         if (varMap != null) {
             synchronized(this) { varMap.clear(); }
         }
+        // Allow the Ruby runtime to be GC-eligible immediately after terminate().
```
headius, line 107: Perhaps it is out of scope for this PR, but if we are fully terminating a container, we probably should have hard errors for any further attempt to use it. That would mean not just nulling out the runtime but also marking the container as dead in a formal way.

That can come in a separate PR, though, since the teardown of the runtime already basically renders the container unusable.

core/src/main/java/org/jruby/internal/runtime/ThreadService.java
```diff
@@ -131,6 +134,16 @@ public class ThreadService extends ThreadLocal<SoftReference<ThreadContext>> {
 
     private final AtomicLong threadCount = new AtomicLong(0);
 
+    /**
+     * Weak references to every SoftReference ever stored into this ThreadLocal (one per thread
```
headius: I don't believe this is necessary. If the SoftReferences are held by ThreadService elsewhere, wrapping them in a WeakReference does not aid their being collected, and in the case where those SoftReferences do actually become dereferenced, we have just shifted the problem to a list of WeakReferences. Please explain if there's something I'm missing here. I've never seen a SoftReference wrapped in a WeakReference before.

core/src/main/java/org/jruby/internal/runtime/ThreadService.java
```diff
@@ -167,6 +180,33 @@ public void teardown() {
 
         // clear thread map
         rubyThreadMap.clear();
+
+        // Clear every SoftReference<ThreadContext> that was ever stored into this ThreadLocal.
```
headius, line 179: If the goal here is to avoid this list accumulating a bunch of dead references, we should use an existing weak map or set. Doing it this way just means we'll slowly accumulate a list of WeakReferences instead of a list of SoftReferences.

core/src/main/java/org/jruby/javasupport/JavaSupport.java
```diff
@@ -112,9 +113,9 @@ public IRubyObject allocateProxy(Object javaObject, RubyClass clazz) {
     public JavaSupport(final Ruby runtime) {
         this.runtime = runtime;
 
-        this.javaClassCache = ClassValue.newInstance(klass -> new JavaClass(runtime, getJavaClassClass(), klass));
+        this.javaClassCache = new MapBasedClassValue<>(klass -> new JavaClass(runtime, getJavaClassClass(), klass));
```
headius: Is this change just so that we can go clear those entries? Again I wonder why dereferencing the JavaSupport is not enough to dereference the ClassValue and all values it refers to.

headius:
I think this is the part that's confusing me:

> ClassValue.newInstance() delegates to StableClassValue, which wraps java.lang.ClassValue. That stores computed values inside the Class object itself via JDK internals — not inside JavaSupport. So setting javaSupport = null makes our ClassValue instance unreachable from JRuby's side, but the JDK's own table entries (rooted at Class objects) can only be cleared by the GC's weak-key processing.

ClassValue does store the values in a WeakHashMap with the ClassValue's "identity" object used as the weak key. The references to values are hard, but dereferencing the ClassValue should allow these WeakHashMaps to dereference the ClassValue identities and eventually clear that hard reference.

So I believe I'm correct in thinking that clearing the JavaSupport reference is at least *sufficient* to ensure the proxy objects are vacated, but I recognize that may not happen as rapidly as we'd like (especially if there are hundreds or thousands of references from that ClassValue identity to our JavaClass proxies).

Using MapBasedClassValue gives us the ability to actively clear those references, since we control the underlying map, albeit at the cost of a much longer teardown process than simply walking awaay from them.

I'd ike to see what the heap curve looks like for each case. I suspect the regular ClassValue version will climb higher before clearing (due to WeakReference delays) but I would not expect it to continue "leaking" over time.

So I propose a compromise: ship the logic from your patch that actively clears MapBasedClassValue caches, but do not use MapBasedClassValue by default. This will allow us the benefit of ClassValue's JDK optimizations for non-embedded use cases (faster access and less teardown overhead for command-line apps) while allowing embedded use cases a way to opt into active ClassValue clearing with MapBasedClassValue.

It also minimizes the changes, since MapBasedClassValue is already configurable.

What do you think?

core/src/main/java/org/jruby/internal/runtime/ThreadService.java
jimtng: I can revert the WeakReference wrapping. It isn't required for the leak fix.

jimtng:
> So I believe I'm correct in thinking that clearing the JavaSupport reference is at least _sufficient_ to ensure the proxy objects are vacated, but I recognize that may not happen as rapidly as we'd like (especially if there are hundreds or thousands of references from that ClassValue identity to our JavaClass proxies).

Based on my test, no it is not sufficient.

I checked out jruby-10.0 and then only changed a single line in Ruby.java:
```
--- a/core/src/main/java/org/jruby/Ruby.java
+++ b/core/src/main/java/org/jruby/Ruby.java
@@ -3384,7 +3384,8 @@ public final class Ruby implements Constantizable {
         allModules.clear();
         constantNameInvalidators.clear();
         symbolTable.clear();
-        javaSupport = loadJavaSupport();
+        javaSupport = null;
```

Then I ran openhab with this JVM options:
```
-Xmx150m -XX:+UseG1GC -XX:G1PeriodicGCInterval=1000 -XX:-G1PeriodicGCInvokesConcurrent -XX:MinHeapFreeRatio=5 -XX:MaxHeapFreeRatio=10 -XX:SoftRefLRUPolicyMSPerMB=0
```

Then each time a new jruby engine is re-loaded I ran 5 GC calls to ensure that GC is performed.
It shows the same leak as the unmodified jruby-10.0.4.0 - in my case 9MB of heap is never reclaimed each time the engine reloads. 

Given the small Xmx setting, on the 5th reload I eventually hit OOM:

```
java.lang.OutOfMemoryError: Java heap space
        at java.util.concurrent.ConcurrentHashMap$ValuesView.iterator(ConcurrentHashMap.java:4743) ~[?:?]
        at org.openhab.core.thing.internal.ThingManagerImpl.allEnabledThingsAreInitialized(ThingManagerImpl.java:1177) ~[?:?]
        at org.openhab.core.thing.internal.ThingManagerImpl.lambda$21(ThingManagerImpl.java:1208) ~[?:?]
        at org.openhab.core.thing.internal.ThingManagerImpl$$Lambda/0x0000000100e00228.run(Unknown Source) ~[?:?]
        at java.util.concurrent.Executors$RunnableAdapter.call(Executors.java:572) ~[?:?]
        at java.util.concurrent.FutureTask.runAndReset(FutureTask.java:358) ~[?:?]
        at java.util.concurrent.ScheduledThreadPoolExecutor$ScheduledFutureTask.run(ScheduledThreadPoolExecutor.java:305) ~[?:?]
        at java.util.concurrent.ThreadPoolExecutor.runWorker(ThreadPoolExecutor.java:1144) [?:?]
        at java.util.concurrent.ThreadPoolExecutor$Worker.run(ThreadPoolExecutor.java:642) [?:?]
        at java.lang.Thread.runWith(Thread.java:1596) [?:?]
        at java.lang.Thread.run(Thread.java:1583) [?:?]
```

This seems to me that there is a genuine leak, and not merely SoftReferences waiting to be claimed.

headius:
@jimtng I think you misunderstood me. I meant that the ClassValue change should not be necessary if we are clearing the reference to JavaSupport, since that should in turn cause the ClassValues to all be dereferenced.

The other changes make sense.

Edit: DEreferenced

jimtng:
> So I propose a compromise: ship the logic from your patch that actively clears MapBasedClassValue caches, but do not use MapBasedClassValue by default. This will allow us the benefit of ClassValue's JDK optimizations for non-embedded use cases (faster access and less teardown overhead for command-line apps) while allowing embedded use cases a way to opt into active ClassValue clearing with MapBasedClassValue.


I tried reverting all the MapBasedClassValue back to ClassValue as they were before this PR. The result is OOM on the 5th reload in the same set up as above. This shows me that we do need to use MapBasedClassValue and manually clear them.


This is way beyond my Java knowledge, so I asked copilot:


The answer is right there in the [`Java7ClassValue` comment](https://github.com/jruby/jruby/blob/f2ff4258e0d54b37331ac62fddfc134c27ee1b3b/core/src/main/java/org/jruby/util/collections/Java7ClassValue.java#L23-L41) — **this is the same known bug from #3228**:

```java
// If we put any objects that reference an org.jruby.Ruby runtime
// (like RubyClass instances) in here we run into a circular
// reference situation that GC isn't handling where they will
// always be strongly referenced and never garbage collected.
// This appears to be a bug in OpenJDK. See jruby/jruby#3228.
```

### The exact cycle in `StableClassValue` (the current default)

`StableClassValue` stores `StableValue` objects in `Class.classValueMap`. `StableValue` is a **non-static inner class** of `StableClassValue`, so every instance carries an implicit `StableClassValue.this` reference:

```
String.class → classValueMap (WeakHashMap)
    weak key  ──────────────────────────────────────────── proxy.identity
    hard value → StableValue (inner class instance)
                   ↓ implicit outer ref
                   StableClassValue.this
                     → proxy (java.lang.ClassValue field)
                         → proxy.identity  ◄── SAME OBJECT as weak key
                   ↓ .result
                   RubyModule → Ruby_old → (javaSupport = null)
```

The hard value in the `WeakHashMap` strongly references `proxy.identity`, which IS the weak key. GC will never collect the weak key — it is permanently strongly reachable through the hard value. `expungeStaleEntries()` can run a million times and find nothing to remove. It doesn't matter how aggressive your GC flags are — this isn't a timing issue, it's a **permanent strong reference to the weak key from within its own value**.

Your GC flags, `-SoftRefLRUPolicyMSPerMB=0`, periodic full GC — none of them can help because the object the GC needs to collect is not actually weakly reachable. It has a strong reference to itself via the cycle.

### How `Java7ClassValue` worked around this

It wrapped the stored value in `WeakReference<T>` to break the cycle:
```java
private final java.lang.ClassValue<WeakReference<T>> proxy = ...
```
Now `classValueMap` stores a `WeakReference<RubyModule>` — that doesn't hold `StableClassValue.this`, so no cycle. The strong retention was moved into a `List<T> computedValues` inside `proxy` itself — which goes away when `proxy` goes away.

### Why `MapBasedClassValue` is the right fix

`MapBasedClassValue` never writes into `Class.classValueMap` at all. There's no `proxy.identity`, no `WeakHashMap` with this pathological cycle, no JDK bug involved. `cache.clear()` is a direct `ConcurrentHashMap.clear()`.

The bug is explicitly documented in JRuby's own `Java7ClassValue` as an **OpenJDK bug** (#3228), and `StableClassValue` (the current default) recreates the same cycle. `MapBasedClassValue` is not just an "eager clearing" preference — it's the only implementation that doesn't reproduce the known GC cycle bug.

headius:
Ok, give me a day to verify this and try to fix the ClassValue cycle.

jimtng:
> I tried reverting all the MapBasedClassValue back to ClassValue as they were before this PR.

To clarify, I started with the current PR https://github.com/jruby/jruby/pull/9315/changes/7a17756dc8266331038dfc40bab149832457de32 and only reverted the changes that involved the use of MapBasedClassValue back to their pre-PR ClassValue, but leaving all the other changes in this PR.

That didn't work, confirming that the changes to MapBasedClassValue are necessary.

headius:
I believe I've fixed the problem.

I'm using this script to reproduce, so hopefully I'm getting the same leaks as you are with OpenHAB:

```ruby
loop {
  10.times {
    ruby = org.jruby.Ruby.newInstance
    ruby.evalScriptlet("java.lang.String; java.util.ArrayList; java.io.InputStream")
    ruby.tearDown
  }
  3.times { JRuby.gc }
  print "."
}
```

This script spins up 10 JRuby instances, accesses some Java classes to make sure they have been proxied, tears the instances down, and then does a few GCs.

Here's your PR as it stands with all changes in place, including MapBasedClassValue (about 4 minutes):

Screenshot 2026-04-01 at 23 02 32

It's noisy, but level. Compare that to this graph, with the MapBasedClassValue changes reverted to ClassValue (a bit over 5 minutes):

Screenshot 2026-04-01 at 23 07 36

The leak is obvious.

Copilot was right about one thing: the cycle is due to the `StableClassValue.this` from `StableValue`. But we can just patch StableClassValue without throwing it out.

Here's the patch:
```diff
diff --git a/core/src/main/java/org/jruby/util/collections/StableClassValue.java b/core/src/main/java/org/jruby/util/collections/StableClassValue.java
index 7d2a8999cf..2c78e0909a 100644
--- a/core/src/main/java/org/jruby/util/collections/StableClassValue.java
+++ b/core/src/main/java/org/jruby/util/collections/StableClassValue.java
@@ -7,6 +7,7 @@ import java.util.function.Function;
  * occurs at most once.
  */
 final class StableClassValue<T> extends ClassValue<T> {
+    final Object lock = new Object();
 
     public StableClassValue(ClassValueCalculator<T> calculator) {
         super(calculator);
@@ -27,11 +28,13 @@ final class StableClassValue<T> extends ClassValue<T> {
      * @param <Input> input of the computation
      * @param <Result> result of the computation
      */
-    private class StableValue<Input, Result> {
+    private static class StableValue<Input, Result> {
         private final Function<Input, Result> calculator;
         private volatile Result result;
-        StableValue(Function<Input, Result> calculator) {
+        private final Object lock;
+        StableValue(Object lock, Function<Input, Result> calculator) {
             this.calculator = calculator;
+            this.lock = lock;
         }
         Result get(Input input) {
             Result result = this.result;
@@ -39,7 +42,7 @@ final class StableClassValue<T> extends ClassValue<T> {
             if (result != null) return result;
 
             // lock on the StableClassValue so there are not multiple locks potentially in different orders
-            synchronized (StableClassValue.this) {
+            synchronized (lock) {
                 result = this.result;
 
                 if (result != null) return result;
@@ -55,7 +58,7 @@ final class StableClassValue<T> extends ClassValue<T> {
     private final java.lang.ClassValue<StableValue<Class<?>, T>> proxy = new java.lang.ClassValue<StableValue<Class<?>, T>>() {
         @Override
         protected StableValue<Class<?>, T> computeValue(Class<?> type) {
-            return new StableValue<>(calculator);
+            return new StableValue<>(lock, calculator);
         }
     };
 }
```

And here's the graph (20 minutes to be sure):
Screenshot 2026-04-01 at 23 24 59

Noisy, but no worse than the Copilot version and no leak.

This is not exactly the patch I'd used; a proper Lock would be a better choice. But if you can confirm this fix with the `MapBasedClassValue` changes reverted, I'll do a cleaner version after we merge.

jimtng:
Wow nice find! I'll test in a few hours and report back.

jimtng:
> Here's the patch:
> 
> ```diff
> diff --git a/core/src/main/java/org/jruby/util/collections/StableClassValue.java b/core/src/main/java/org/jruby/util/collections/StableClassValue.java
> index 7d2a8999cf..2c78e0909a 100644
> --- a/core/src/main/java/org/jruby/util/collections/StableClassValue.java
> +++ b/core/src/main/java/org/jruby/util/collections/StableClassValue.java
> @@ -7,6 +7,7 @@ import java.util.function.Function;
>   * occurs at most once.
>   */
>  final class StableClassValue<T> extends ClassValue<T> {
> +    final Object lock = new Object();
>  
>      public StableClassValue(ClassValueCalculator<T> calculator) {
>          super(calculator);
> @@ -27,11 +28,13 @@ final class StableClassValue<T> extends ClassValue<T> {
>       * @param <Input> input of the computation
>       * @param <Result> result of the computation
>       */
> -    private class StableValue<Input, Result> {
> +    private static class StableValue<Input, Result> {
>          private final Function<Input, Result> calculator;
>          private volatile Result result;
> -        StableValue(Function<Input, Result> calculator) {
> +        private final Object lock;
> +        StableValue(Object lock, Function<Input, Result> calculator) {
>              this.calculator = calculator;
> +            this.lock = lock;
>          }
>          Result get(Input input) {
>              Result result = this.result;
> @@ -39,7 +42,7 @@ final class StableClassValue<T> extends ClassValue<T> {
>              if (result != null) return result;
>  
>              // lock on the StableClassValue so there are not multiple locks potentially in different orders
> -            synchronized (StableClassValue.this) {
> +            synchronized (lock) {
>                  result = this.result;
>  
>                  if (result != null) return result;
> @@ -55,7 +58,7 @@ final class StableClassValue<T> extends ClassValue<T> {
>      private final java.lang.ClassValue<StableValue<Class<?>, T>> proxy = new java.lang.ClassValue<StableValue<Class<?>, T>>() {
>          @Override
>          protected StableValue<Class<?>, T> computeValue(Class<?> type) {
> -            return new StableValue<>(calculator);
> +            return new StableValue<>(lock, calculator);
>          }
>      };
>  }
> ```

Unfortunately this didn't work for me. 
- I reverted the MapBasedClassValue changes in JavaSupport.java and JavaSupportImpl.java
- I applied your diff to StableClassValue.java

Under my test, it also caused OOM crash after 5 reloads. I confirmed that the modified StableValue constructor was called by inserting a log call.

jimtng:
I tried asking copilot to go down this path and it tried setting calculator = null after it was used, but that too didn't fix the leak. This is what it's saying, FWIW:

The `calculator = null` fix was not sufficient. The remaining chain is even simpler and **completely immune to GC** — no amount of GC calls can break it.

### The permanent strong chain

```
String.class  ← permanent GC root (bootstrap classloader, NEVER collected)
  └─► String.class.classValueMap (WeakHashMap)
           └─► backing array entry (hard value)
                 └─► StableValue_N
                       └─► .result  (hard field, volatile Result)
                             └─► RubyModule_N
                                   └─► .runtime
                                         └─► Ruby_N  ← NEVER collected
```

`StableValue_N.result` is a **hard (strong) reference**. `String.class` is a permanent GC root. Every class your script accesses (`java.lang.String`, `java.util.ArrayList`, etc.) creates one such chain. `Ruby_N` is permanently strongly reachable. GC cannot collect an object that is strongly reachable — that is literally the definition of "live object". Your 5 GC calls, `G1PeriodicGCInterval`, `SoftRefLRUPolicyMSPerMB=0` — none of them have any effect on this.

### Why headius's test hides the leak

His script only touches 3 JI classes (`java.lang.String`, `java.util.ArrayList`, `java.io.InputStream`). With 10 iterations, he has 30 `StableValue.result` entries pinning 10 `Ruby` instances. If each minimal Ruby instance is ~3-4MB, that's ~30-40MB retained — fits in 150MB. In a real application accessing hundreds or thousands of Java classes, each reload retains proportionally more.

### Why `Java7ClassValue` actually worked

Worth noting — `Java7ClassValue` (the deprecated one) solved this correctly by storing `WeakReference<T>` in classValueMap rather than `T` directly:

```java
// classValueMap hard value: WeakReference<RubyModule>  ← does NOT strongly hold Ruby_N
// strong reference to RubyModule_N lives in: proxy.computedValues (a List<T>)
//   but proxy is in the island {JavaSupport_N, proxy_N} which IS GC-eligible
//   after javaSupport = null → island collected → list dropped → RubyModule_N released
```

That approach actually worked. It was deprecated for performance/other reasons, not correctness.

### Conclusion

`StableClassValue` with a hard `result` field simply cannot be used for runtime-scoped values that need to be released at teardown. `MapBasedClassValue` is the correct solution — it never writes into `Class.classValueMap` at all, so `clear()` is a direct `ConcurrentHashMap.clear()` with no involvement of permanent GC roots or `java.lang.ClassValue` mechanics.

headius:
This analysis is still wrong, or at least deeply misleading. My patience is wearing thing, but let's play this game a bit longer so we can get it right for 10.0.5.0.

# AI is often confidently incorrect

> His script only touches 3 JI classes (java.lang.String, java.util.ArrayList, java.io.InputStream). With 10 iterations, he has 30 StableValue.result entries pinning 10 Ruby instances.

For some reason, Copilot ignored completely that my script ran in a **loop**, generating and terminating 10 runtime instances and then running GC. Each iteration takes around 1-2 seconds on my system once warmed up.

After 5 minutes without my fix, it was consuming almost 500MB. The leak was obvious. Let's use that as a baseline.

Five minutes is 600 seconds. We'll be generous and say that 50MB out of that 500 was always going to be occupied, leaving 450MB leaking. That works out to something like 0.75MB leaking per second.

Now let's look at the fixed version that ran for **20 minutes**. If it were still broken and every second leaked 0.75MB, that's 0.75MB * (1200s) = **900GB** of heap memory that should have been permanently stuck.

You saw both graphs... where 5 minutes without the fix climbed up to 500MB, the fixed version stayed within the 100-200MB range. The total available heap size never even got past 500MB.

A leak was definitely fixed. It just wasn't the leak you saw when you tried to confirm it.

It's shocking that Copilot would be **so wrong** about this. It's confidently giving you bad information. Beware.

# Java7ClassValue vs StableClassValue

I said above:

> ...with the MapBasedClassValue changes reverted to ClassValue

The changes I reverted dd not completely reset JavaSupport and JavaSupportImpl. I just removed the direct construction of `MapBasedClassValue`, leaving in place your earlier change to use the weakly-referenced `Java7ClassValue` via `ClassValue.newWeakInstance`.

Here's the diff I applied to revert those changes:

```diff
commit d88d692332ac4a1137f4532f6f627f84f128ba55
Author: Charles Oliver Nutter <headius@headius.com>
Date:   Thu Apr 2 12:03:49 2026 -0500

    Revert MapBasedClassValue changes

diff --git a/core/src/main/java/org/jruby/javasupport/JavaSupport.java b/core/src/main/java/org/jruby/javasupport/JavaSupport.java
index 75089d1d6b..6bf9499314 100644
--- a/core/src/main/java/org/jruby/javasupport/JavaSupport.java
+++ b/core/src/main/java/org/jruby/javasupport/JavaSupport.java
@@ -67,8 +67,8 @@ public abstract class JavaSupport {
     protected final Ruby runtime;
 
     @Deprecated(since = "9.4.3.0")
-    private final MapBasedClassValue<JavaClass> javaClassCache;
-    private final MapBasedClassValue<RubyModule> proxyClassCache;
+    private final ClassValue<JavaClass> javaClassCache;
+    private final ClassValue<RubyModule> proxyClassCache;
 
     static final class UnfinishedProxy extends ReentrantLock {
         final RubyModule proxy;
@@ -113,9 +113,9 @@ public abstract class JavaSupport {
     public JavaSupport(final Ruby runtime) {
         this.runtime = runtime;
 
-        this.javaClassCache = new MapBasedClassValue<>(klass -> new JavaClass(runtime, getJavaClassClass(), klass));
+        this.javaClassCache = ClassValue.newWeakInstance(klass -> new JavaClass(runtime, getJavaClassClass(), klass));
 
-        this.proxyClassCache = new MapBasedClassValue<>(this::computeProxyClass);
+        this.proxyClassCache = ClassValue.newWeakInstance(this::computeProxyClass);
 
         // Proxy creation is synchronized (see above) so a HashMap is fine for recursion detection.
         this.unfinishedProxies = new ConcurrentHashMap<>(8, 0.75f, 1);
@@ -397,8 +397,8 @@ public abstract class JavaSupport {
     }
 
     public void tearDown() {
-        javaClassCache.clear();
-        proxyClassCache.clear();
+//        javaClassCache.clear();
+//        proxyClassCache.clear();
         unfinishedProxies.clear();
         objectProxyCache.clear();
     }
diff --git a/core/src/main/java/org/jruby/javasupport/JavaSupportImpl.java b/core/src/main/java/org/jruby/javasupport/JavaSupportImpl.java
index 21a81ae1b5..18738b1e54 100644
--- a/core/src/main/java/org/jruby/javasupport/JavaSupportImpl.java
+++ b/core/src/main/java/org/jruby/javasupport/JavaSupportImpl.java
@@ -53,10 +53,10 @@ import org.jruby.javasupport.proxy.JavaProxyClass;
  */
 public class JavaSupportImpl extends JavaSupport {
 
-    private final MapBasedClassValue<Map<String, AssignedName>> staticAssignedNames =
-            new MapBasedClassValue<>(JavaSupportImpl::newAssignedNames);
-    private final MapBasedClassValue<Map<String, AssignedName>> instanceAssignedNames =
-            new MapBasedClassValue<>(JavaSupportImpl::newAssignedNames);
+    private final ClassValue<Map<String, AssignedName>> staticAssignedNames =
+            ClassValue.newInstance(JavaSupportImpl::newAssignedNames);
+    private final ClassValue<Map<String, AssignedName>> instanceAssignedNames =
+            ClassValue.newInstance(JavaSupportImpl::newAssignedNames);
 
     public JavaSupportImpl(final Ruby runtime) {
         super(runtime);
@@ -138,8 +138,8 @@ public class JavaSupportImpl extends JavaSupport {
     public void tearDown() {
         super.tearDown();
 
-        staticAssignedNames.clear();
-        instanceAssignedNames.clear();
+//        staticAssignedNames.clear();
+//        instanceAssignedNames.clear();
 
         synchronized (javaProxyClasses) {
             javaProxyClasses.clear();
diff --git a/core/src/main/java/org/jruby/util/collections/ClassValue.java b/core/src/main/java/org/jruby/util/collections/ClassValue.java
index 6641aea029..170a1ae543 100644
--- a/core/src/main/java/org/jruby/util/collections/ClassValue.java
+++ b/core/src/main/java/org/jruby/util/collections/ClassValue.java
@@ -54,6 +54,17 @@ public abstract class ClassValue<T> {
         return Options.JI_CLASS_VALUES.load().function.apply(calculator);
     }
 
+    /**
+     * Create a weak-valued ClassValue for caches whose computed values can retain a Ruby runtime.
+     *
+     * java.lang.ClassValue has historically caused retention cycles in OpenJDK when values are
+     * strongly held and those values reference JRuby runtime state. Using weak values here breaks
+     * that cycle while preserving class-keyed lookup semantics.
+     */
+    public static <T> ClassValue<T> newWeakInstance(ClassValueCalculator<T> calculator) {
+        return new Java7ClassValue<>(calculator);
+    }
+
     @Deprecated(since = "9.4.13.0")
     private static <T> ClassValue<T> newJava7Instance(ClassValueCalculator<T> calculator) {
         return new Java7ClassValue<>(calculator);
```

I suspect this is the real difference between your most recent run and mine. Perhaps I should have been more specific and said "after reverting the MapBasedClassValue changes but retaining the newWeakInstance change" or simply provided my entire diff. In any case, you were testing something different than I was. We'll mark that down as a miscommunication aggravated by bad AI analysis.

It's notable that even though my run was using your weak proxy cache change, it **still leaked** 0.75MB per second. That's because the other two ClassValue in JavaSupportImpl went back to being StableClassValue (by default config) and that's likely where the retained memory came from.

Copilot was correct that the `StableValue.this` reference was a problem. **My fix corrected that problem.**. It seems to have ignored this and moved the goalposts to try to "help" you figure out why your test still failed.

The reference chain explanations... are correct. With all caches restored to StableClassValue, there is indeed a cycle from the classValueMap back to the StableClassValue.result, keeping it alive. That leaks much more quickly because it does actually retain a whole JRuby runtime (which is almost certainly much more than 3-4MB).

# Fixing the right problem

There's still a memory issue when using StableClassValue to caching the proxy instances, that much we can agree on. The fix may be to return to weak references, or perhaps we can avoid some additional hidden references that are not cleared when the runtime is torn down. Throwing out the class altogether is not the answer I'm looking for.

In my last review, I asked of your MapBasedClassValue changes:

> Is this change just so that we can go clear those entries? Again I wonder why dereferencing the JavaSupport is not enough to dereference the ClassValue and all values it refers to.

Let's drill down on this particular change.

If I go back to stock JavaSupport/Impl (from 10.0.4.0, with the `newWeakInstance` change also reverted) and run my example, it leaks very quickly because of StableClassValue retaining runtimes. This is after just one minute:

Screenshot 2026-04-02 at 12 35 27

We can configure it to use MapBasedClassValue instead of StableValue by specifying JVM property `ji.class.values=HARD_MAP`.

Running the same script as above, but with `jruby -Xji.class.values=HARD_MAP`, memory usage remains stable. Copilot is correct that MapBasedClassValue avoids a GC root holding references to the JRuby runtimes. That's a big reason why it still exists.

Here's a 10-minute run just to be sure:

Screenshot 2026-04-02 at 12 58 43

But here's the real kicker: my reverted JavaSupport/Impl **does not include your MapBasedClassValue clearing logic**. My local JavaSupport/Impl diff against 10.0.4.0 is **empty**. We have configured away the proxy leak **solely by switching to MapBasedClassValue**.

So the answer to "is this just so we can clear those entries?" appears to be "no", because MapBasedClassValue also avoids retaining runtime references from the bootstrap classloader GC root. **That's why it is configurable.**

# More history

The use of ClassValue has gone through many iterations.

Originally, we tried to just use the JDK ClassValue directly, guarded by a configuration property (since we still supported earlier Java versions at the time). We made Java7ClassValue opt-in and provided MapBasedClassValue to abstract the API away. 

Later, once we required Java 8 in JRuby 9k, we switched the default to Java7ClassValue. We quickly ran into the hard-referencing problem and a contributor helped us make Java7ClassValue hold weak references (#3228). 

The current "breaking" change may have arrived last year when I started switching the default back to a regular JDK ClassValue. (5ac95eac7a9c9d4911c225134ce22198651077db in #8844). The torture tests I ran against this did not appear to leak.

Next, I discovered that ClassValue calculation was not idempotent and could execute more than once so I added StableClassValue and a configuration flag to select it (remaining changes in #8844).

At this point there were still no obvious leaks, as proved by a 12 hour run of my torture test. There clearly is a leak now.

So either my torture test was insufficient to show the leaks we are now seeing, or I later introduced another change that is causing the current leak.

**That's the fix I currently seek.**

headius:
The `StableClassValue.this` leak was introduced in 8288cac12305cb9c02a54f4824703870b1e3bb85.

headius:
Well, how about that... the **actual** leak we seek to fix was also introduced in https://github.com/jruby/jruby/commit/8288cac12305cb9c02a54f4824703870b1e3bb85.

Here's the fix (the StableClassValue lock fix is also needed):

```diff
diff --git a/core/src/main/java/org/jruby/javasupport/JavaSupport.java b/core/src/main/java/org/jruby/javasupport/JavaSupport.java
index e9c31e8778..150ee77ebb 100644
--- a/core/src/main/java/org/jruby/javasupport/JavaSupport.java
+++ b/core/src/main/java/org/jruby/javasupport/JavaSupport.java
@@ -114,21 +114,28 @@ public abstract class JavaSupport {
 
         this.javaClassCache = ClassValue.newInstance(klass -> new JavaClass(runtime, getJavaClassClass(), klass));
 
-        this.proxyClassCache = ClassValue.newInstance(this::computeProxyClass);
+        this.proxyClassCache = ClassValue.newInstance(new ProxyComputer(runtime)::computeProxyClass);
 
         // Proxy creation is synchronized (see above) so a HashMap is fine for recursion detection.
         this.unfinishedProxies = new ConcurrentHashMap<>(8, 0.75f, 1);
     }
 
-    /**
-     * Because of the complexity of processing a given class and all its dependencies,
-     * we opt to synchronize this logic. Creation of all proxies goes through here,
-     * allowing us to skip some threading work downstream.
-     */
-    private synchronized RubyModule computeProxyClass(Class<?> klass) {
-        RubyModule proxyKlass = Java.createProxyClassForClass(runtime, klass);
-        JavaExtensions.define(runtime, klass, proxyKlass); // (lazy) load extensions
-        return proxyKlass;
+    private static class ProxyComputer {
+        private final Ruby runtime;
+
+        ProxyComputer(Ruby runtime) {
+            this.runtime = runtime;
+        }
+        /**
+         * Because of the complexity of processing a given class and all its dependencies,
+         * we opt to synchronize this logic. Creation of all proxies goes through here,
+         * allowing us to skip some threading work downstream.
+         */
+        private synchronized RubyModule computeProxyClass(Class<?> klass) {
+            RubyModule proxyKlass = Java.createProxyClassForClass(runtime, klass);
+            JavaExtensions.define(runtime, klass, proxyKlass); // (lazy) load extensions
+            return proxyKlass;
+        }
     }
```

The problem was actually really simple: `this::computeProxyValue` is a lambda with a reference to JavaSupport, which references the ClassValue instances. Once I started looking for a long-lived reference from the ClassValue to itself, it was obvious.

Five minutes of monitoring seemed sufficient to prove it's fixed:

Screenshot 2026-04-02 at 16 11 48

This is again not an ideal fix; the synchronization is now against the (singleton) ProxyComputer rather than against JavaSupport, so it won't disallow other JavaSupport synchronizers from running at the same time. It should be replaced with a shared lock that all such synchronizers use.

But it's probably good enough for now, and I'll clean it up with the rest of StableClassValue after we merge your other fixes.

@jimtng I don't want you to think I was impatient with you.

I know you were just trying to find a solution to a complex problem, and AI was the only tool you had to do it. Unfortunately in this case, it obscured the actual problem and led both of us around in circles.

I've lost a couple of days dealing with this misleading analysis, when I probably should have just tried to reproduce your case and investigated it myself. My impatience was with an AI telling me I was wrong when I wasn't and backing it up with falsehoods.

It's not your fault... you were just the middle man, and important information was getting lost along the way.

If you find that there's still memory being leaked, just let me know. Leave Copilot out of it.

headius:
Also, I take back what I said earlier. This analysis of the reference chain keeping everything alive was just wrong:

```
String.class  ← permanent GC root (bootstrap classloader, NEVER collected)
  └─► String.class.classValueMap (WeakHashMap)
           └─► backing array entry (hard value)
                 └─► StableValue_N
                       └─► .result  (hard field, volatile Result)
                             └─► RubyModule_N
                                   └─► .runtime
                                         └─► Ruby_N  ← NEVER collected
```

The bug I finally fixed involved a reference from ClassValue to the compute function to JavaSupport to the ClassValue. It was a simple sequence and the AI description of it was completely wrong.

In fact I'm pretty sure the only things that Copilot got right were:

* StableClassValue.this reference was keeping StableClassValue alive.
* There was still a leak after fixing that.

The rest was noise, or worse than noise since it convinced you that it had found the actual problem and made me doubt my own analysis and understanding of the system.

headius:
Fixes #9187 with my additions in place.

kares:
> the synchronization is now against the (singleton) ProxyComputer rather than against JavaSupport, so it won't disallow other JavaSupport synchronizers from running at the same time. It should be replaced with a shared lock that all such synchronizers use.

> But it's probably good enough for now, and I'll clean it up with the rest of StableClassValue after we merge your other fixes.

is it though? recall there was a reason to sync against concurrent proxy initialization due issues when multiple threads initializing the same proxy.

headius:
I'm going back over all the commits now because I no longer trust any of them. The most recent changes, which are clean and leak-free on my local system, fails a test that Copilot wrote. The first commit in the PR makes claims about a reference path that again seem to be incorrect. We won't be merging this as-is.

headius:
> is it though? recall there was a reason to sync against concurrent proxy initialization due issues when multiple threads initializing the same proxy.

The synchronization against a global lock is necessary, but I'm not sure if that global lock needs to be the JavaSupport instance. In any case, I am going to redo this whole thing and make all places synchronizing this way use a common Lock instance.

core/src/main/java/org/jruby/embed/internal/LocalContext.java
```diff
@@ -104,6 +104,12 @@ public void remove() {
...
+        // Without this, the strong reference chain
+        //   ScriptingContainer -> SingleThreadLocalContextProvider -> LocalContext -> Ruby
+        // keeps the runtime (and its JRubyClassLoader) alive until this LocalContext
+        // is itself collected, which may be delayed by ScriptingContainer.finalize().
+        runtime = null;
```
headius, line 112: This change is accepted. It may  not be necessary, but it's reasonable to clear a reference to the runtime. I'm adding a termination flag to ensure the context is not able to be re-vivified.

See #9359.

core/src/main/java/org/jruby/internal/runtime/ThreadService.java
headius, line 141: This change is rejected.

* Despite the Copilot analysis below, the reference chain from the ThreadLocal (ThreadService) to the runtime has at least one more weak link: the ThreadLocalMap's Entry. When the ThreadService is cleared at teardown, the ThreadLocal no longer has any GC roots, and will be collected. This will in turn clear out the SoftReference to the ThreadContext and the rest of the graph.
* Accumulating SoftReference here introduces a new leak because as threads terminate during runtime there's nothing to clean their references out of this list.

core/src/main/java/org/jruby/javasupport/JavaSupport.java
```diff
@@ -112,9 +112,9 @@ public IRubyObject allocateProxy(Object javaObject, RubyClass clazz) {
     public JavaSupport(final Ruby runtime) {
         this.runtime = runtime;
 
-        this.javaClassCache = ClassValue.newInstance(klass -> new JavaClass(runtime, getJavaClassClass(), klass));
+        this.javaClassCache = ClassValue.newWeakInstance(klass -> new JavaClass(runtime, getJavaClassClass(), klass));
 
-        this.proxyClassCache = ClassValue.newInstance(this::computeProxyClass);
+        this.proxyClassCache = ClassValue.newWeakInstance(this::computeProxyClass);
```
headius, line 117: These changes are rejected. The problem here is not in the use of StableClassValue (ignoring a separate bug there that keeps it alive). The problem is that `this::computeProxyClass` introduces a reference chain from the ClassValue to the compute lambda to the JavaSupport to the ClassValue. We will fix this by removing the JavaSupport reference from that chain, allowing the ClassValue to clear properly when dereferenced.

See #9359

core/src/main/java/org/jruby/util/collections/ClassValue.java
```diff
@@ -54,6 +54,17 @@ public static <T> ClassValue<T> newInstance(ClassValueCalculator<T> calculator)
         return Options.JI_CLASS_VALUES.load().function.apply(calculator);
     }
 
+    /**
+     * Create a weak-valued ClassValue for caches whose computed values can retain a Ruby runtime.
+     *
+     * java.lang.ClassValue has historically caused retention cycles in OpenJDK when values are
+     * strongly held and those values reference JRuby runtime state. Using weak values here breaks
+     * that cycle while preserving class-keyed lookup semantics.
+     */
```
headius, line 63: Rejected; incorrect fix.

core/src/main/java/org/jruby/Ruby.java
```diff
@@ -3369,11 +3369,12 @@ private void systemTeardown(final ThreadContext context) {
             }
         }
 
-        // Shut down and replace thread service after all other hooks and finalizers have run
-        threadService.teardown();
-        threadService = new ThreadService(this);
-
-        // Release classloader resources
+        // Release classloader resources before tearing down the thread service.
```
headius, line 3372: This change is rejected, except for reordering the teardown sequence. LoadService does indeed call getCurrentContext, but it can simply be moved above the ThreadService teardown above. The order of the ThreadService and classloader termination is unimportant, because neither of them depends upon the other.

See #9359

jimtng:
Superseded by #9359

Issue #9319, opened 2026-03-20
[Java 26 Leyden] AOT Cache breaks dynamic Ruby proxies for classes loaded from external JARs


gillesbergerp:
### Expected Behavior

When using Java 25 or Java 26 ``-XX:AOTCacheOutput``, JRuby should retain or rebuild dynamic Ruby proxy methods (like ``generate_salt``) for Java classes loaded from external ``.jar`` files.

### Actual Behavior

While standard JDK classes (like ``java.util.logging.Logger``) survive AOT caching perfectly, **classes loaded from external JARs lose their Ruby proxy methods**.

Upon loading the AOT cache, JRuby successfully instantiates the Java object from the JAR, but attempting to call a Ruby-fied method (like ``generate_salt``) throws a ``NoMethodError``. Using ``java_send(:generate_salt, ...)`` works, proving the class is in the cache but JRuby's dynamic proxy generation was bypassed.

This completely breaks gems like ``bcrypt`` in production when utilizing Leyden.

### Minimal Reproducible Example

```
# reproduce.rb
require 'bcrypt'

puts "Salt: #{BCrypt::Engine.generate_salt}"
```

#### Step 1: The Training Run (Works)

```
$ jruby --nocache -J-XX:AOTCacheOutput=test.aot reproduce.rb

Salt: $2a$12$5JAd646raDMwuCpJ8Id0vu
Temporary AOTConfiguration recorded: test.aot.config
Launching child process /usr/lib/jvm/java-26-amazon-corretto/bin/java to assemble AOT cache test.aot using configuration test.aot.config
Picked up JAVA_TOOL_OPTIONS: -Djava.class.path=: --add-opens=java.base/java.io=org.jruby.dist --add-opens=java.base/java.nio.channels=org.jruby.dist --add-opens=java.base/sun.nio.ch=org.jruby.dist --add-opens=java.management/sun.management=org.jruby.dist -Xss2048k -Djffi.boot.library.path=/opt/jruby/lib/jni -Djava.security.egd=file:/dev/urandom --enable-native-access=org.jruby.dist --sun-misc-unsafe-memory-access=allow --module-path=/opt/jruby/lib/jruby.jar -Djruby.home=/opt/jruby -Djruby.lib=/opt/jruby/lib -Djruby.script=jruby -Djruby.shell=/bin/sh -XX:AOTCacheOutput=test.aot -XX:AOTConfiguration=test.aot.config -XX:AOTMode=create
[...]

PR #9332, opened 2026-03-26
Support Java 26

headius:
I think we need a few tweaks to fully support Java 26, such as for #9319.


headius:
Only failures were unrelated to Java 26, but I fixed those too.

PR #9353, opened 2026-03-31
Update JNR dependencies

headius:
jffi -> 1.3.15
jnr-ffi -> 2.2.19
jnr-posix -> 3.1.22
jnr-enxio -> 0.32.20
jnr-unixsocket -> 0.38.25


headius:
The build will be green once the deploy of jnr-unixsocket completes. Maven Central seems a bit backed up today.

PR #9359, opened 2026-04-03
Leak fixes redux

headius:
This PR will reconsider and apply appropriate fixes from #9315.

The fixes and analysis in that PR were performed by Copilot, and after several of them turned out to be unnecessary or incorrect, we lost faith that any of the fixes were correct or based on sound analysis. This PR will attempt to re-visit that analysis and apply only necessary, correct fixes.


jimtng:
Thanks for the fix, @headius! My openhab test setup is no longer leaking with this.

PR #9364, opened 2026-04-05
[fix] TCPServer#initialize: close channel on bind/init failure

kares:
noticed during some (heavy JOSSL) `TCPServer` testing that the native channel wasn't being always closed.


headius (approved):
Looks good to me.

PR #9366, opened 2026-04-06
Update specs

eregon:


eregon:
@headius Following up on https://github.com/jruby/jruby/pull/9193#issuecomment-3817254795
I cannot merge this PR myself, it says needs an approval:
Screenshot From 2026-04-06 21-00-04

I think it would be good to remove that requirement since it's frequent that PRs are merged to JRuby without a review or approval.
At least for ruby/spec updates it would be convenient if they can merge without needing someone else to review it, as soon as the CI is green.

PR #9367, opened 2026-04-07
review (wrong) asserts

kares:
tripped over 2 of these running with `-ea` e.g. `ByteList#setEncoding` which works with `null`


headius (approved):
LGTM

PR #9368, opened 2026-04-07
[refactor] remove duplicate fallback in CallableSelector

kares:
tried AssignableOrDuckable twice in a row before falling through


headius (approved):
lol

PR #9369, opened 2026-04-07
Abstract RubyString

headius:
This is the final of three planned abstractions for JRuby 10.1. Abstracting RubyString will allow us to more easily introduce subtypes that compactly represent single bytes or characters, efficiently wrap Java String and CharSequence, and explore alternative backing stores.


headius:
I don't believe this can happen for 10.1 given the exposure of the RubyString constructors. I will push a different PR that deprecates all of them for removal and we will hope to do this some time later.

PR #9376, opened 2026-04-16
Fix destructuring of block argument with default value

### LINKED ISSUE #9375: `instance_exec` block argument destructuring when default value passed
**Environment Information**

* JRuby 10.0.5.0
* OS: MacOS (arm)

**Expected Behavior**

Same as with CRuby:

```
irb(main):001> RUBY_VERSION
=> "4.0.1"
irb(main):002> Object.new.instance_exec([1]) { |x = :unset| x }
=> [1]
irb(main):003> Object.new.instance_exec([1]) { |x| x }
=> [1]
```

**Actual Behavior**

```
irb(main):001> Object.new.instance_exec([1]) { |x = :unset| x }
=> 1
irb(main):002> Object.new.instance_exec([1]) { |x| x }
=> [1]
```

When there is and array and a default value, only params first value is returned. It also happens for the array with more than one element:

```
irb(main):003> Object.new.instance_exec([1, 2]) { |x = :unset| x }
=> 1
```

I looks similar to #9208

katafrakt:
It fixes #9375 - incorrect behaviour found in `instance_exec`, `module_exec` and `class_exec`:

```
Object.new.instance_exec([1, 2]) { |x = :unset| x } 
=> 1
```

The fix is similar to #9211 

The alternative I considered is to maybe put that logic (`sig.opt() + sig.required() == 1 && !sig.hasRest()`) but I'm not sure it would be correct. Sounds like it, but it's not how the fix was applied in the other PR. Will be happy to change.

### DIFF
--- a/core/src/main/java/org/jruby/runtime/Block.java
+++ b/core/src/main/java/org/jruby/runtime/Block.java
@@ -227,7 +227,9 @@ public class Block implements FunctionOneOrTwoOrThree<ThreadContext, IRubyObject
     // This should be the only argument massaging in yield.  This handles all argument conversion logic except
     // for the generic Block#yield(IRubyObject value).
     private static IRubyObject[] maybeSpreadArgs(ThreadContext context, IRubyObject[] args, Block block) {
-        return block.type != Type.LAMBDA && args.length == 1 && block.getSignature().isSpreadable() ?
+        Signature sig = block.getSignature();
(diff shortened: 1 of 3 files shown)


headius (approved):
Seems simple enough for now, just narrowing the cases where we expand that argument. Needs a long-term cleanup at some point.

PR #9379, opened 2026-04-19
Make fixnum packing configurable and disable ints

headius:
During benchmarking I realized two things about the split fixnums:

* IntFixnum and LongFixnum are the same size, due to object alignment (i.e. having Int gains us nothing)
* Having all three types (Short, Int, and Long) can lead to wide-ranging Integer code becoming tri-morphic, which may interefere with JVM optimizations.

I created this patch to allow enabling and disabling all *four* widths of fixnum for experimentation, but having only short and long enabled by default.

I this is a WIP and I don't like how it conditionally enables the different widths. The code is far too complex. A better option would be to have fully specialized paths for each combination (possibly generated) and choose one at boot that will never again check a condition.

will land a separate patch for 10.1.0.0 that just hard disables IntFixnum to keep it minimal.

### DIFF
--- a/core/src/main/java/org/jruby/RubyFixnum.java
+++ b/core/src/main/java/org/jruby/RubyFixnum.java
@@ -303,61 +303,239 @@ public abstract class RubyFixnum extends RubyInteger implements Constantizable, 
     private static final boolean USE_SHORT_FIXNUMS = Options.USE_SHORT_FIXNUMS.load();
     private static final boolean USE_BYTE_FIXNUMS = Options.USE_BYTE_FIXNUMS.load();
 
-    public static RubyFixnum newFixnum(Ruby runtime, long value) {
-        if (!USE_COMPACT_FIXNUMS || !USE_INT_FIXNUMS || value > Integer.MAX_VALUE || value < Integer.MIN_VALUE) {
-            // long is never in cache range
+    private static final FixnumFactory FIXNUM_FACTORY;
+    static {
+        if (USE_INT_FIXNUMS || USE_BYTE_FIXNUMS) {
+            // fallback handles all configurations at runtime
+            FIXNUM_FACTORY = new FallbackFixnumFactory();
+        } else if (USE_COMPACT_FIXNUMS && USE_SHORT_FIXNUMS) {
+            if (USE_CACHE) {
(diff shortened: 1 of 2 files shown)


headius:
This is complete, but after some extensive benchmarking on MacOS AArch64 the new ShortFixnum may introduce more instruction overhead than the size is worth. The bimorphic nature of having both LongFixnum and ShortFixnum does not appear to impact things much, but the ShortFixnum constantly widening and narrowing its values for reads and initialization of new instances adds enough overhead to a small loop to reduce performance. It also appears to interfere with escape analysis, both on HotSpot and Graal JIT.

Some loop benchmarks follow.

All shorts:
```ruby
def foo
  i = 0
  while i < 32_000
    i+=1
    j = 0
    while j < 32_000
      j+=1
    end
  end
end
10.times {
  t = Time.now
  foo
  puts Time.now - t
}
```

Long and short with caching, HotSpot 25:
```
3.0774749999999997
2.830375
2.738893
2.7603999999999997
2.81643
2.75495
2.75328
2.7494810000000003
2.771321
2.801205
```

Long only with caching, HotSpot 25 (`-Xfixnum.compact=false`):
```
2.679109
2.154972
2.1185169999999998
2.0998430000000003
2.0842650000000003
2.088898
2.155581
2.0820969999999996
2.125191
2.12845
```

Long only without caching, HotSpot 25 (`-Xfixnum.compact=false -Xfixnum.cache=false`):
```
2.613342
2.065781
1.877827
1.8706520000000002
1.856989
1.899999
1.882702
1.8590119999999999
1.9631930000000002
1.873456
```

Same on HotSpot 26 (escape analysis seems to have improved? Enabling cache degrades this back to 25 speeds.):
```
3.782794
2.172116
1.229315
1.2248510000000001
1.231879
1.302273
1.234393
1.220123
1.233962
1.229031
```

Same on GraalVM 25 (escape analysis eliminating most objects):
```
0.17663399999999999
0.16367500000000001
0.14729299999999998
0.137191
0.142957
0.13831700000000002
0.143365
0.13711500000000001
0.14526699999999998
0.13666
```

On the other hand, an array of 1M short-ranged values goes from 37,391,992 bytes to 29,391,992 bytes, almost 22% less memory. That savings only applies to short-ranged integers but it still makes this a tough call since most loops will also be in short range.

PR #9381, opened 2026-04-19
Shrink RubyHashLinkedBuckets

headius:
This started out as an experiment to remove the "generation" count that appeared to only be used for some rehashing heuristics, and expanded to an overall reduction of Hash size (based on the RubyHashLinkedBuckets implementation).

Changes
---------

* generation count (4 bytes) is removed, as it did not appear to be doing anything useful that other fields didn't handle. See previous discussion below.
* threshold (4 bytes) was removed, as it was only used by the "Javasoft" hash strategy that has been disabled forever.
* iteratorCount (4 bytes) is reduced to 2 bytes, as there's no reason to be able to accommodate 2^32 concurrent iterators.
* The "head" entry is reduced by 32 bytes by introducing a base entry link type, since it only needed to track next and previous added entries.

These changes reduce the retained size of an empty hash from 160 bytes (in 10.1.0.0) to 136 bytes.

On generation count
---------------------

This generation count did not appear to actually be doing anything useful, only being incremented on rehash and only used to follow a slightly different path for clearing the Hash instance. Removing it does not appear to affect functionality, and it was a 32-bit integer increasing the size of every linked-bucket Hash object in the system.

It is a relic of old ported code from CRuby and I'm not sure we need it now (or perhaps ever needed it).

### DIFF
--- a/core/src/main/java/org/jruby/RubyHash.java
+++ b/core/src/main/java/org/jruby/RubyHash.java
@@ -351,16 +351,10 @@ public class RubyHash extends RubyObject implements Map {
         public Object getKey() {
             return key;
         }
-        public Object getJavaifiedKey(){
-            return key.toJava(Object.class);
-        }
 
         @Override
         public Object getValue() {
             return value;
-        }
-        public Object getJavaifiedValue() {
-            return value.toJava(Object.class);
         }
 
         @Override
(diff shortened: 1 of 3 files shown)


headius:
Two additional changes were attempted but did not produce expected results:

* Combining the flags and iteratorCount into a single 16-bit value would allow Hotspot to shrink the base object size another 8 bytes (eliminating internal and external padding) but led to much more complicated code for managing these values. It also reduces the safe number of concurrent iterators to 2^13, 8192, which gets closer to theoretically possible with many fibers and threads.
* The `prevAdded` and `nextAdded` on the `head` link could be moved directly into `RubyHashLinkedBuckets`, eliminating that object altogether. In practice, making the access of those fields bimorphic increased the execution time of `Hash.new` by 2x.

At this point the remaining tricks  to shrink the implementation are less productive than introducing a new space-efficient implementation (see #9394).

PR #9387, opened 2026-04-21
[9.4] deps(stdlib): bump erb from 2.2.3 to 4.0.4.1 to resolve CVE-2026-41316

chadlwilson:
2.2.3 is EOL from CRuby perspective, so they will not backport the fix to a maintained gem (I [asked](https://github.com/ruby/erb/pull/114#issuecomment-4293182677)).

This thus deviates from the CRuby 3.1 stdlib version and bumps to 4.0.4.1 (as bundled with Ruby 3.4.x, maintained at https://github.com/ruby/erb/tree/ruby-3.4).

This still has Ruby 2.7 compatibility and the two relevant "breaking" changes seem innocuous: https://github.com/ruby/erb/blob/ruby-3.4/NEWS.md (Drop deprecated -S option from command / Bump `required-ruby_version` to Ruby 2.7+)

My only open question with this is whether the gem extension stub introduced in erb `4.0.x` is a problem, since this did not exist with `2.2.3`. Is something like https://github.com/jruby/jruby/pull/9289 needed for `9.4` - even though it's not there currently?

### DIFF
--- a/lib/pom.rb
+++ b/lib/pom.rb
@@ -41,7 +41,7 @@ default_gems = [
   ['digest', '3.1.0'],
   ['drb', '2.1.0'],
   ['english', '0.7.1'],
-  ['erb', '2.2.3'],
+  ['erb', '4.0.4.1'],
   ['error_highlight', '0.3.0'],
   # https://github.com/ruby/etc/issues/19
   # ['etc', '1.3.0'],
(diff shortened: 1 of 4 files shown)


chadlwilson:
Is the extensions thing OK @headius ? 2.2.3 didn't have an extension (or the java stub).

PR #9401, opened 2026-04-28
improve performance for Java "block-to-interface" conversion

### LINKED ISSUE #1401: Java interface using closure implementaion is slower than mixin implementation

### LINKED ISSUE #5023: excessive loading of same InterfaceImpl

### LINKED ISSUE #9424: proc to interface inconsistency with multiple args
Normally when JRuby encounters a functional interface with (the abstract method having) multiple parameters, the block receives all of them or the ones from the left e.g.

```
> { one: 1 }.to_java.forEach { |arg, val| puts arg }
one
=> nil
> { one: 1 }.to_java.forEach { |arg| puts arg }
one
=> nil
```

This behavior isn't consistent with a functional interface that includes an overload for the abstract interface method e.g.

```java
  @FunctionalInterface
  public interface TwoArgWriter {

      void write(String label, java.io.File file);

      default void write(java.io.File file) {
          write("default-label", file);
      }
}
```

in this case the current behavior (in 9.4 - 10.0.5.0) is that the generated implementation ends up re-defining the default method as the implementation, the **single `{ |arg| ... }` form will end up receiving the (second) `File` argument** as the sole argument. 

AI generated full reproducer: https://gist.github.com/kares/19b7e9aa8781eb4abf572d203642c980

kares:
Passing a Ruby block to a Java method that expects a SAM (single-abstract-method "functional interface") type has evolved into having very bad performance despite being one of the very useful JI features available in JRuby, current behavior:

1. allocates a fresh `RubyProc` (to wrap the `block`)
2. materialises a singleton class (`MetaClass`) for the proc object
3. `singletonClass.include(<interfaceModule>)` - `synchronized(runtime.hierarchyLock)`.
4. `extended` callback re-invokes `singletonClass.include(<interfaceModule>)`
   a no-op, but `includeModule` still calls `invalidateCacheDescendants` unconditionally, taking the same lock again
5. `addMethod("method_missing")` - `synchronized(hierarchyLock)`
6. `addMethod("<sam-method>")` - `synchronized(hierarchyLock)`

**4 acquisitions per task on a single monitor**
Under load with multiple threads, the lack of caching becomes very noticeable and prevents proper concurrent execution: 

- **per-call latency** is dominated by singleton-class setup, even when the call site is otherwise trivial
- **multi-thread throughput** is capped - adding more workers doesn't make things "faster"

---

The fix introduces a **specialised, lock-free conversion path** for blocks that originate at the Java-integration layer and are *known* to be on their way into just being used for *block-to-interface* execution, existing semantics for user-land `RubyProc`s are preserved.

### Performance

Numbers for the included micro benchmark: https://gist.github.com/kares/c4616956570fd58515ad6f0ffd822f8f

Improvement seems to be at the **order of 20-30x** and with **multi-threaded execution far beyond** (no more contention between threads).

### DIFF
--- a/core/src/main/java/org/jruby/java/codegen/BlockInterfaceGenerator.java
+++ b/core/src/main/java/org/jruby/java/codegen/BlockInterfaceGenerator.java
@@ -4,6 +4,7 @@ import java.lang.reflect.Constructor;
 import java.lang.reflect.Method;
 import java.lang.reflect.Modifier;
 import java.util.ArrayList;
+import java.util.concurrent.ConcurrentMap;
 
 import org.jruby.Ruby;
 import org.jruby.RubyProc;
@@ -11,6 +12,9 @@ import org.jruby.compiler.impl.SkinnyMethodAdapter;
 import org.jruby.java.proxies.BlockInterfaceTemplate;
 import org.jruby.javasupport.Java;
 import org.jruby.util.ASM;
+import org.jruby.util.ClassDefiningClassLoader;
+import org.jruby.util.JRubyClassLoader;
+import org.jruby.util.collections.ConcurrentWeakHashMap;
 import org.objectweb.asm.ClassWriter;
 import org.objectweb.asm.Type;
 import org.objectweb.asm.commons.GeneratorAdapter;
@@ -54,36 +58,49 @@ import static org.objectweb.asm.Opcodes.V11;
 public final class BlockInterfaceGenerator {
 
     /**
+     * Cache of generated {@link BlockInterfaceTemplate} subclass constructors, weakly keyed by the interface.
+     *
+     * <p>
+     * The generated class lives in the {@link JRubyClassLoader}, so the value never roots a classloader outside of
+     * JRuby itself; the only strong root held by a cache entry is the constructor (class pair) for the interface.
+     */
+    static final ConcurrentMap<Class<?>, Constructor<? extends BlockInterfaceTemplate>> CACHE = new ConcurrentWeakHashMap<>();
+
+    public static Constructor<? extends BlockInterfaceTemplate> fromCache(final Class<?> interfaceType) {
+        return CACHE.get(interfaceType);
+    }
+
+    /**
      * @return the constructor, which is usable with any Ruby block targeting the same interface
      */
     @SuppressWarnings("unchecked")
     public static Constructor<? extends BlockInterfaceTemplate> getConstructor(final Ruby runtime, final Class<?> interfaceType)
         throws ReflectiveOperationException {
+
+        var constructor = CACHE.get(interfaceType);
+        if (constructor != null) return constructor;
+
         assert interfaceType.isInterface();
 
         final String implClassName = makeImplClassName(interfaceType);
-        final ClassLoader loader = runtime.getJRubyClassLoader();
+        final JRubyClassLoader loader = runtime.getJRubyClassLoader();
 
-        Class<?> implClass;
         synchronized (loader) {
+            constructor = CACHE.get(interfaceType);
+            if (constructor != null) return constructor;
+
+            Class<?> implClass;
             try {
                 implClass = Class.forName(implClassName, true, loader);
             } catch (ClassNotFoundException e) {
+                assert Java.getFunctionalInterfaceMethod(interfaceType) != null : "not a functional-interface: " + interfaceType;
                 implClass = defineImplClass(loader, interfaceType, implClassName);
             }
-        }
 
-        return (Constructor<? extends BlockInterfaceTemplate>) implClass.getConstructor(RubyProc.class);
-    }
-
-    private static ArrayList<Method> getAbstractMethods(final Class<?> interfaceType) {
-        final var result = new ArrayList<Method>();
-        for (Method method : interfaceType.getMethods()) {
-            if (!Modifier.isAbstract(method.getModifiers())) continue;
-            if (method.getDeclaringClass() == Object.class) continue;
-            result.add(method);
+            constructor = (Constructor<? extends BlockInterfaceTemplate>) implClass.getConstructor(RubyProc.class);
+            CACHE.put(interfaceType, constructor);
+            return constructor;
         }
-        return result;
     }
 
     private static String makeImplClassName(final Class<?> interfaceType) {
@@ -99,10 +116,10 @@ public final class BlockInterfaceGenerator {
      *           implements <interfaceType> { ... }
      * }</pre>
      */
-    private static Class<?> defineImplClass(final ClassLoader loader,
+    private static Class<?> defineImplClass(final ClassDefiningClassLoader loader,
                                             final Class<?> interfaceType,
                                             final String implClassName) {
-        final ClassWriter cw = ASM.newClassWriter(loader);
+        final ClassWriter cw = ASM.newClassWriter((ClassLoader) loader);
         final String pathName = implClassName.replace('.', '/');
 
         cw.visit(V11,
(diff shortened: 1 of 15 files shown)


test/jruby/test_instantiating_interfaces.rb
```diff
@@ -51,6 +50,222 @@ def cs.char_at; nil end
     assert_equal ' ', cs.charAt(2)
   end
 
+  # --- arity 0 -----------------------------------------------------------------
+
+  def test_proc_to_java_functional_interface
```
kares, line 55: just extra tests, there aren't changed (given these are use-originated `Proc` objects -> they use the same old path as before)

core/src/main/java/org/jruby/java/codegen/BlockInterfaceGenerator.java
```diff
@@ -0,0 +1,310 @@
...
+
+            Class<?> implClass;
+            try {
+                implClass = Class.forName(implClassName, true, loader);
+            } catch (ClassNotFoundException e) {
+                assert Java.getFunctionalInterfaceMethod(interfaceType) != null : "not a functional-interface: " + interfaceType;
+                implClass = defineImplClass(loader, interfaceType, implClassName);
+            }
+
+            constructor = (Constructor<? extends BlockInterfaceTemplate>) implClass.getConstructor(RubyProc.class);
```
headius, line 98: Of course this stuff uses java.lang.reflect all over the place (both before and after this change) but we could be using MethodHandles for all of these and skip the overhead of a reflective invocation.

core/src/main/java/org/jruby/java/codegen/BlockInterfaceGenerator.java
```diff
@@ -0,0 +1,310 @@
...
+        }
+
+        emitReturn(mv, returnType);
+        mv.end();
+    }
+
+    /**
+     * Loads {@code argIndex} from the Java arg slots and boxes it if primitive.
+     */
+    private static void loadBoxedArg(GeneratorAdapter ga, int argIndex, Class<?> paramType) {
```
headius: These three utility methods already exist in various forms throughout JRuby. Look at RealClassGenerator for some examples. We don't need to reimplement primitive boxing and unboxing and class literal loading again.

headius (commented):
This is basically ok but there's a lot of duplicated code from elsewhere and we could consider using MethodHandle instead of reflection objects in some of these places. The logic seems sound; generate a class for Interface### to Proc dispatch and reuse it. I have some concerns about how these are being cached, for situations where many classes are encountered rarely or only once.

Overall I approve but we can chat through some improvements before merging.

I'd also like to see the benchmark code somewhere so we can continue to audit and profile those cases.

core/src/main/java/org/jruby/java/proxies/BlockInterfaceTemplate.java
```diff
@@ -0,0 +1,45 @@
...
+    public BlockInterfaceTemplate(final RubyProc proc) {
+        assert proc != null;
+        this.runtime = proc.getRuntime();
+        this.block = proc.getBlock();
+    }
+
+    @JIT
+    @SuppressWarnings("unused")
+    protected final IRubyObject __ruby_call(final Class<?> returnType) {
+        return block.call(runtime.getCurrentContext());
```
kares, line 25: while this seem a little abandoned (could be part of the generated class); at occasions I wanted to be able to set a break point around the "block-to-interface" execution, in Java, having the "template" super-class allows for that. no hard feelings if preference is to simply get rid of this.

core/src/main/java/org/jruby/java/codegen/BlockInterfaceGenerator.java
kares: tried unreflecting the constructor into a handle, no difference with the benchmarks mentioned in the PR.

core/src/main/java/org/jruby/java/proxies/BlockInterfaceTemplate.java
headius: I get the desire and it's nice to be able to fall back on something debuggable. I wish there were a way to specify `__FILE__` and `__LINE__` in Java so you could provide the original generated code lines in the stack trace.

core/src/main/java/org/jruby/java/codegen/BlockInterfaceGenerator.java
headius: The handle needs a static root of some kind, either a static final field (not possible here) or by using LambdaMetaFactory to generate an interface implementation. Without those it usually is not much faster than reflection, which also uses unrooted handles internally.

kares:
For the record, did a re-run of the snapshot build from this on a private (large JRuby scripting) codebase and everything was :green_circle: apart from one occurance of the inconsistency "bug" : https://github.com/jruby/jruby/issues/9424 (which is resolved here in the PR).

core/src/main/java/org/jruby/java/codegen/BlockInterfaceGenerator.java
headius: I'll open a separate issue to explore places we're using reflection and try replacing them with LambdaMetaFactory conversions for perf.

headius (approved):
Approved. Patch is much cleaner now that duplicate code has been eliminated and additional cleanup by hand. Merge at your discretion @kares.

headius:
Opened https://github.com/jruby/jruby/issues/9455 to explore lambda conversion for this and other places we still use reflection.

PR #9403, opened 2026-04-29
add keyword arg support to callMethod directly

drzaiusx11:
started a discussion here: https://github.com/jruby/jruby/discussions/9402, copying verbatim here for reference:

I've been using JRuby to wrap some pre-existing Gems to be called by Java using the ScriptingContainer interface provided by JRuby 10.x, however, whenever a Gem uses keyword arguments there doesn't seem to be a clean way to force a Map to be treated as the keyword arguments for that Ruby API. I see theres a ThreadContext.CALL_KEYWORD flag I can use, but it seems more an internal implementation detail than something that should be exposed to scripting users. Afaict this works:

```java
  // Manually build a RubyHash with symbol keys
  RubyHash kwargs = RubyHash.newHash(runtime);
  kwargs.fastASet(runtime.newSymbol("name"), JavaUtil.convertJavaToUsableRubyObject(runtime, "World"));

  // Set the flag right before callMethod
  context.callInfo = ThreadContext.CALL_KEYWORD;
  container.callMethod(receiver, "greet", new Object[]{kwargs});
```

What I'm proposing is adding further callMethod method overloads to include a last argument of Map<String, Object> or similar that will become those keyword arguments (convert strings to symbols automatically, call convertJavaToUsableRubyObject() on values, etc. My guess on why this hasn't come up before is that any scripting use cases predated Ruby's 3.2 order argument changes around auto-conversion of the last hash into keyword args. Since 3.2+ that behavior is disallowed so some other mechanism needs to exist going forward. I think what I propose makes sense here, but i could be missing something important.

This is my attempt at implementing it


enebo:
I like this and I also like the PR only pushes this into embedding API itself.  We have a long running task to remove callInfo and to make kwargs more first class and not based on some out-of-band field set per thread.

My only thing I wonder about is that Map<String, Object> would somehow be nicer if it could be String or RubySymbol for people who already know they are directly interacting with Ruby from an embedded context.  Not sure I am recommending another set of overloads here or whether that case is solved by our larger effort to rewrite kwargs but it seems like a reasonable use case (also it probably could be added as a later PR).

headius:
Yes, this seems like a reasonable addition.

@enebo To your point, we could use Map<CharSequence, Object> since RubySymbol and RubyString also implement CharSequence. Dunno if it's worth it or not.

I'll review.

core/src/main/java/org/jruby/embed/internal/EmbedRubyObjectAdapterImpl.java
```diff
@@ -281,4 +285,78 @@ private static IRubyObject[] convertArgs(final Ruby runtime, final Object[] args
         return rubyArgs;
     }
 
+    // --- Keyword argument support ---
+
+    public Object callMethod(Object receiver, String methodName, Map<String, Object> kwargs) {
```
headius: I have some vague concerns about this signature and the `Object...` signature above co-existing. There's some cases that might end up ambiguous in existing code (even though these are a little weird)

```java
Object[] ary = ...
Map<String, Object> map = ...
callMethod(recv, name, ary, map); # ambiguous call
```

So I'm wondering if it might be better to give these methods different names to ensure it's clear you're calling with explicit keywords. We could potentially even keep the `Object...` for positional args:

```java
callKeywordMethod(Object receiver, String methodName, Map<String, Object> kwargs, Object... posArgs)
```

core/src/main/java/org/jruby/embed/internal/EmbedRubyObjectAdapterImpl.java
```diff
@@ -281,4 +285,78 @@ private static IRubyObject[] convertArgs(final Ruby runtime, final Object[] args
...
+            IRubyObject[] rubyArgs = (args != null && args.length > 0) ? convertArgs(runtime, args) : IRubyObject.NULL_ARRAY;
+
+            RubyHash kwargsHash = convertKwargsToRubyHash(runtime, kwargs);
+
+            IRubyObject[] finalArgs;
+            if (kwargsHash != null) {
+                finalArgs = new IRubyObject[rubyArgs.length + 1];
+                System.arraycopy(rubyArgs, 0, finalArgs, 0, rubyArgs.length);
+                finalArgs[rubyArgs.length] = kwargsHash;
+                context.callInfo = ThreadContext.CALL_KEYWORD;
```
headius, line 319: I think we should also defensively clear context.callInfo on the way out so it doesn't get stuck if we call a method that behaves badly. Perhaps also save existing callInfo and restore it, but I don't think we should go too far in trying to preserve callInfo for Java embedding API consumers.

headius (commented):
Generally on the right track. I think it's worth discussing whether it's safer/cleaner to just add new method names for the kwargs calls rather than more complicated overloads that might be ambiguous in some cases.

core/src/main/java/org/jruby/embed/internal/EmbedRubyObjectAdapterImpl.java
drzaiusx11: Yeah probably in a finally block. I can try and get to this sometime this week

core/src/main/java/org/jruby/embed/internal/EmbedRubyObjectAdapterImpl.java
drzaiusx11: After mulling this over a bit, I'm in agreement that a new name would be better to avoid confusion and properly disambiguate when Object or Object.. are involved in the callMethod sig.  That does mean we'd likely want to (re-) implement similar arg permutations of the preexisting API but with keyword args (hopefully with an additional goal of reducing ambiguity.)

Personally I'm not a huge fan of Object and 'Object...' in user targeting API signatures as it leads to disambiguation problems like this one. The worst culprit for ambiguity in this particular case is likely the generics form where the last arg is the Class Type token as you point out. Those can be interpreted as one of the Object... elements, since java.lang.Class also extends Object iirc. I'm honestly not sure why it compiles and runs now that you point it out. It's certainly not intuitive regarding which one "wins" (at least to me, and I wrote the new APIs 😅)


Maybe we just start with List<Object> for ordered arguments and a Map<CharSequence, Object> for keyword args, ending with the return Type for for a T like:

<T> T callMethodWithKeywords(Object receiver, String methodName, List<Object> orderedArgs, Map<String, Object> keywordArgs, Class<T> type)

If folks find it useful we may want to add more permutations in later iterations.

Random thought: keeping the "callMethod" method name prefix would probably be more auto complete friendly / discoverable to users familiar with the existing APIs than callKeywordMethod, but I defer to you on naming here.

drzaiusx11: An alternative to the ... Vararg support we could provide 1..N arg permutation override methods and have the last arg always be the type for consistency across the new API signature-wise. N can be sufficiently small to cover 95pct use and outliers just use the array arg version which the N permutation APIs would support >N outlier cases. Like N=5 or so:

N=1
T callMethodWithKeywords(Object receiver, String methodName, RubyObject ro1, Map<CharSequence, Object> keywordArgs, Class type)
...

N=5
T callMethodWithKeywords(Object receiver, String methodName, RubyObject ro1, RubyObject ro2, RubyObject ro3, RubyObject ro4, RubyObject ro5, Map<CharSequence, Object> keywordArgs, Class type)

N>5 user just use the list one:
T callMethodWithKeywords(Object receiver, String methodName, List orderedArgs, Map<CharSequence, Object> keywordArgs, Class type)

drzaiusx11: An alternative with a much larger deviation from the current API would be to support a first class scripting Parameters class that could be constructed with a fluent API or similar...

drzaiusx11:
ok pushed the updates

- renamed keyword-specific callMethod(...) overloads to a new API: callMethodWIthKeywordArgs(...)
- restores context.callInfo to original value in finally block

I'm also purposefully avoiding using 'Object...' for now in the new APIs, opting for arrays to avoid the inevitable disambiguation issues discussed.

We can add N-arg positional overloads for N...4 or something sane if thats more ergonomic but I think arrays isn't a terrible first pass

core/src/main/java/org/jruby/embed/internal/EmbedRubyObjectAdapterImpl.java
drzaiusx11: added finally

core/src/main/java/org/jruby/embed/internal/EmbedRubyObjectAdapterImpl.java
drzaiusx11: went with new API: `callMethodWithKeywordArgs(...)` which aligns with preexisting JRuby API naming schemes

core/src/main/java/org/jruby/embed/EmbedRubyObjectAdapter.java
```diff
@@ -184,4 +186,108 @@ public interface EmbedRubyObjectAdapter extends RubyObjectAdapter {
...
+     * @param kwargs is a map of keyword argument names to values. Keys must be
+     *              {@code String}, {@code RubyString}, or {@code RubySymbol}.
+     *              May be {@code null} or empty to indicate no keyword arguments.
+     * @return an instance automatically converted from Ruby to Java
+     * @throws IllegalArgumentException if a key is not a supported type
+     * @throws InvokeFailedException if the underlying Ruby call raises and the
+     *              implementation is configured to wrap exceptions.
+     * @since JRuby 10.1.1.0
+     */
+    Object callMethodWithKeywordArgs(Object receiver, String methodName, Map<?, Object> kwargs);
```
drzaiusx11, line 213: question for @headius (or anyone that is more familiar): is the reason the other (pre-existing) `callMethod(...)` APIs return Object instead of IRubyObjects to support non-IRubyObject receivers? I followed the pattern but seems odd to return an Object here instead of IRubyObject when not passing a Type final arg.  Let me know if IRubyObject return type is preferred here.

drzaiusx11:
Ready for re-review @enebo @headius

headius:
Hey I missed the re-review but I'll take a look tonight.

core/src/main/java/org/jruby/embed/EmbedRubyObjectAdapter.java
```diff
@@ -184,4 +186,108 @@ public interface EmbedRubyObjectAdapter extends RubyObjectAdapter {
...
+     * @param receiver is an instance that will receive this method call
+     * @param methodName is a method name to be called
+     * @param kwargs is a map of keyword argument names to values. Keys must be
+     *              {@code String}, {@code RubyString}, or {@code RubySymbol}.
+     *              May be {@code null} or empty to indicate no keyword arguments.
+     * @return an instance automatically converted from Ruby to Java
+     * @throws IllegalArgumentException if a key is not a supported type
+     * @throws InvokeFailedException if the underlying Ruby call raises and the
+     *              implementation is configured to wrap exceptions.
+     * @since JRuby 10.1.1.0
```
headius: Minor nitpick: JRuby 10.1.1.0 was released since the last update here, so these should say 10.1.2.0 now.

headius (changes requested):
Only one small change (same change in several places) needed and this can merge for 10.1.2.0.

core/src/main/java/org/jruby/embed/EmbedRubyObjectAdapter.java
drzaiusx11: updated

headius (approved):

headius:
Thank you for your help on this!

I am not sure if we've ever chatted about your use of JRuby, but I wanted to mention this in case it works for your use of JRuby:

My work on JRuby is now 100% funded by support contracts from [Headius Enterprises](https://headius.com). We offer customizable packages for JRuby users, with the middle-tier Expert Support being the most popular. Expert Support gives you a couple of dedicated hours of my time per month, prioritized bug fixes, security SLAs, and a whole lot of peace-of-mind that your application will keep running and getting better. We also offer a baseline Pro Support level on a per-seat basis and higher levels of support for codebases under NDA, performance profiling, debugging help, and more.

If you have a mission-critical use of JRuby, I'd recommend the Expert Support level, and I'd love to jump on a call to talk about how we can support you. Please consider partnering with us to keep your app running well and the JRuby community moving forward!

[JRuby Support Services by Headius Enterprises](https://headius.com/services)

PR #9404, opened 2026-05-01
Warn on class/const assign in condition

sampokuokkanen:
Extend checkAssignmentInCondition to treat ClassVarAsgnNode and ConstDeclNode as assignable nodes so assignments like class vars or constant declarations in conditionals will trigger the existing "assignment in conditional" warning. Also re-enable the test_assign_in_conditional test by removing its exclude in test/mri/excludes/TestParse.rb.

So code like following:
```ruby
if @@hello = true
  pp "hello"
end
```

and

```ruby
if BASE_HELLO = true
  pp "hello"
end
```

will now have the `found '= literal' in conditional, should be ==` warning. 

Sorry about the Copilot review, I have now disabled it so it will no longer auto review my PRs from now on.

### DIFF
--- a/core/src/main/java/org/jruby/parser/RubyParserBase.java
+++ b/core/src/main/java/org/jruby/parser/RubyParserBase.java
@@ -973,13 +973,13 @@ public abstract class RubyParserBase {
      * assign_in_cond
 	 **/
     private boolean checkAssignmentInCondition(Node node) {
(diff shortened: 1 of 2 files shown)


enebo:
@sampokuokkanen sorry I missed this one.  I think it looks good but it required a merge so I will merge this once that completes.  Looks like the right change and we love to see tests untagged!

PR #9406, opened 2026-05-01
Fix Set marshal object link registration

### LINKED ISSUE #9405: Marshal.load corrupts repeated object reference after Set appears earlier in payload
**Environment Information**

JRuby:

```sh
$ jruby -v
jruby 10.1.0.0 (4.0.0) 2026-04-20 32f988b78c OpenJDK 64-Bit Server VM 26+35 on 26+35 +indy +jit [x86_64-linux]
```

MRI comparison:

```
ruby 4.0.1 (2026-01-13 revision e04267a14b) +PRISM [x86_64-linux]
```

**Expected Behavior**

`Marshal.load(Marshal.dump(obj))` should preserve repeated object references correctly. In the script below, `loaded[2]` should be `"good"`.

Repro:

```ruby
require 'set'

shared = 'good'
payload = [
  Set.new,
  ['bad', shared],
  shared,
]

loaded = Marshal.load(Marshal.dump(payload))

puts RUBY_DESCRIPTION
puts "expected: good"
puts "actual:   #{loaded[2].inspect}"
puts "ok: #{loaded[2] == 'good'}"
```

MRI output:

```
ruby 4.0.1 (2026-01-13 revision e04267a14b) +PRISM [x86_64-linux]
expected: good
actual:   "good"
ok: true
```

**Actual Behavior**

JRuby 10.1.0.0 output:

```
jruby 10.1.0.0 (4.0.0) 2026-04-20 32f988b78c OpenJDK 64-Bit Server VM 26+35 on 26+35 +indy +jit [x86_64-linux]
expected: good
actual:   "bad"
ok: false
```

The third element of the top-level array becomes `"bad"` after the Marshal roundtrip. It should remain the same string object/value as shared, i.e. `"good"`.

Some narrowing observations:

-    No threading is involved.
-    The issue reproduces with plain Marshal.dump / Marshal.load.
-    Replacing `Set.new` with `[]` or `{}` avoids the problem.
-    Using shared.dup for the third payload element avoids the problem.
-    The failure seems to require a Set appearing before a nested array containing a shared string object that is referenced again later.

This was found while debugging Opal's compiler cache on JRuby 10.1.0.0 (a regression since JRuby 10.0.2.0), where a cached compiler's generated JavaScript string was replaced by a source-map mappings string after a Marshal roundtrip.

Ref: https://github.com/opal/opal/pull/2787

hmdne:
**Note that this PR is agent generated and I don't really understand the logic behind it. I'm publishing it just to help you handle the issue.**

JRuby's Set custom marshal compatibility path serialized the Set through a dummy Ruby object containing the legacy @hash instance variable, but it never registered the real Set object with the marshal object link table. On load, the dummy object was registered instead of the Set that was returned to Ruby code. That shifted later object links and could make a subsequent repeated object reference resolve to the wrong earlier object.

Register the real Set with both MarshalDumper and MarshalLoader while keeping the dummy object only as the compatibility carrier for the legacy @hash payload. This preserves the old wire shape for Set marshaling while keeping marshal links aligned with the objects that are actually dumped and returned.

Fixes: https://github.com/jruby/jruby/issues/9405
Assisted-by: gpt-5.5

### DIFF
--- a/core/src/main/java/org/jruby/ext/set/RubySet.java
+++ b/core/src/main/java/org/jruby/ext/set/RubySet.java
@@ -109,6 +109,8 @@ public class RubySet extends RubyObject implements Set {
         public void marshalTo(ThreadContext context, RubyOutputStream out, Object obj, RubyClass type, MarshalDumper marshalStream) {
             RubySet set = (RubySet) obj;
 
+            marshalStream.registerObject(set);
+
             // create dummy object with extra @hash ivar and dump variables from that
             RubyObject dummy = new RubyObject(context.runtime, type);
             dummy.setInstanceVariable("@hash", set.hash);
@@ -128,8 +130,10 @@ public class RubySet extends RubyObject implements Set {
         public Object unmarshalFrom(ThreadContext context, RubyInputStream in, RubyClass type, MarshalLoader loader) {
             RubySet set = new RubySet(context.runtime, type, false);
 
+            loader.entry(set);
+
             // unmarshal as dummy object and extra @hash ivar
-            RubyObject dummy = (RubyObject) loader.entry(new RubyObject(context.runtime, type));
+            RubyObject dummy = new RubyObject(context.runtime, type);
             loader.ivar(context, in, null, dummy, null);
 
             set.hash = (RubyHash) dummy.getInstanceVariables().removeInstanceVariable("@hash");
--- a/test/jruby/test_set.rb
+++ b/test/jruby/test_set.rb
@@ -39,6 +39,20 @@ class TestSet < Test::Unit::TestCase
     assert_equal Set.new([1, 2]), set
   end
 
+  def test_marshal_object_links
+    shared = 'good'
+    loaded = Marshal.load Marshal.dump([Set.new, ['bad', shared], shared])
+
+    assert_equal 'good', loaded[2]
+    assert_same loaded[1][1], loaded[2]
+
+    set = Set.new
+    loaded = Marshal.load Marshal.dump([set, set])
+
+    assert_same Set, loaded[0].class
+    assert_same loaded[0], loaded[1]
+  end
+
   def test_yaml_dump; require 'yaml'
     str = YAML.dump(Set.new)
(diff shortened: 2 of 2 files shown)


headius:
Going forward we will have a policy of rejecting any agent-generated PRs where the submitter cannot explain the changes, but as this is fairly small I'll have a look at it this time.

The tests are not in the correct place, as usual, and I'm not sure they're sufficient.

hmdne:
@headius Thank you. I have corrected the test location (unless you want it to go to regressions or rubyspec, but it seemed to me that you intended this location, as above it there's a test touching both subsystems), also adding one additional check to ensure links are preserved.

As for the agent assistance, I was seeing a CI failure and asked an agent to isolate it. I could have stopped it at the point of reporting the issue, but decided it could be of help if I also asked it to also generate a fix - I marked it this way, so that you could give it additional scrutiny, as I am not familiar with JRuby codebase and I can't guarantee it doesn't break things that aren't covered, even though I made efforts to manually review all the steps and outputs. By the way, if there are recurrent issues from agents, it may help to nudge them in a particular direction using AGENTS.md and CLAUDE.md (ie. by pointing them to a canonical guidelines document or mentioning the pressing issues) - or more conservatively, by strengthening the language in README.md.

headius:
We do intend to improve our documentation such that agents know better where to put tests. Unfortunately this is sometimes subjective.

If the change is related to normal Ruby behavior, the right place is almost always under ruby/spec. That gets the tests back into the community so that other implementations (including CRuby) can run them.

But if the test is especially peculiar to JRuby's implementation, that's not the right place. And of course if it's truly specific to JRuby it should go in one of our spec or test suites (we have too many and have been slowly trying to condense them).

headius:
I've confirmed the fix is appropriate but cleaned it up and documented why it needs to be this way in #9418.

The test has no home right now; it should go in ruby/spec, but the specs there for marshal are really messy. We'll discuss with other contributors (ruby/spec#1355) and try to clean it up (your assistance would also be welcome!)

Issue #9398, opened 2026-04-27
JRuby 10.1 `circular causes` failures within nested rescue (e.g rspec's tests)


chadlwilson:
There are failures within the rspec test suite that might expose some intentional or unintentional differences between JRuby 10.1 and CRuby 4.0.

https://github.com/chadlwilson/rspec/actions/runs/24822251102

**Environment Information**
  
`jruby 10.1.0.0 (4.0.0) 2026-04-20 32f988b78c OpenJDK 64-Bit Server VM 21.0.10+7-LTS on 21.0.10+7-LTS +indy +jit [x86_64-linux]`

**Expected Behavior**

* Tests should be passing; or any known differences understood so we can mark the tests as `Pending` or similar, and linked to upstream issues here.

**Actual Behavior**

**circular causes**
- [x] **rspec-support** [RSpec::Support.method_handle_for(object, method_name) fails with `NameError` when an undefined method is fetched from an object that has overridden `method` to raise an Exception](https://github.com/rspec/rspec/blob/76e6b1af10b43def8dab5a47ff8d8a3ad0977c05/rspec-support/spec/rspec/support_spec.rb#L54-L61)
   <details>
     <summary>stack</summary>

  ```
    1) RSpec::Support.method_handle_for(object, method_name) fails with `NameError` when an undefined method is fetched from an object that has overridden `method` to raise an Exception
       Failure/Error: DEFAULT_FAILURE_NOTIFIER = lambda { |failure, _opts| raise failure }
  
         expected NameError, got #<ArgumentError: circular causes> with backtrace:
           # ./lib/rspec/support.rb:55:in 'method_handle_for'
           # ./spec/rspec/support_spec.rb:59:in 'block in RSpec'
           # /home/runner/work/rspec/rspec/rspec-expectations/lib/rspec/matchers/built_in/raise_error.rb:63:in 'matches?'
           # /home/runner/work/rspec/rspec/rspec-expectations/lib/rspec/expectations/handler.rb:40:in 'block in handle_matcher'
           # /home/runner/work/rspec/rspec/rspec-expectations/lib/rspec/expectations/handler.rb:18:in 'with_matcher'
           # /home/runner/work/rspec/rspec/rspec-expectations/lib/rspec/expectations/handler.rb:39:in 'handle_matcher'
           # /home/runner/work/rspec/rspec/rspec-expectations/lib/rspec/expectations/expectation_target.rb:67:in 'to'
[...]

PR #9407, opened 2026-05-01
Refactor Kernel#raise and fix remaining specs

headius:
In response to #9398 I finally did a major refactor of Kernel#raise. This PR splits up the different arity paths rather than running them all together and moves most of the logic to common utility methods. The last two rubyspecs related to cause have also been fixed.


headius:
Merged! This does not do the additional work to improve pass rate and coverage of `Thread#raise` and `Fiber#raise` but it lays the groundwork to make that easy.

I also audited the CRuby test_exception failures, fixed several, and added good descriptions for the remaining ones.

PR #9409, opened 2026-05-01
Ensure core exceptions get raised with $! as cause

headius:
Many places where we raise exceptions from JRuby core code, we do not attach $! as a cause. This leads to several issues, including part of the issue reported in #9398. This PR adds specs and fixes known cases.

### DIFF
--- a/spec/mspec/lib/mspec/matchers/raise_error.rb
+++ b/spec/mspec/lib/mspec/matchers/raise_error.rb
@@ -1,11 +1,18 @@
 class RaiseErrorMatcher
   FAILURE_MESSAGE_FOR_EXCEPTION = {}.compare_by_identity
+  UNDEF_CAUSE = Object.new
 
   attr_writer :block
 
-  def initialize(exception, message, &block)
+  def initialize(exception, message = nil, options = nil, &block)
+    if message.is_a? Hash
+      @message = nil
+      options = message
+    else
+      @message = message
+    end
+    @cause = options ? options.fetch(:cause, UNDEF_CAUSE) : UNDEF_CAUSE
     @exception = exception
-    @message = message
     @block = block
     @actual = nil
   end
@@ -45,24 +52,45 @@ class RaiseErrorMatcher
     end
   end
 
-  def matching_exception?(exc)
-    matching_class?(exc) and matching_message?(exc)
-  end
-
-  def exception_class_and_message(exception_class, message)
-    if message
-      "#{exception_class} (#{message})"
+  def matching_cause?(exc)
+    case @cause
+    when UNDEF_CAUSE
+      true
     else
-      "#{exception_class}"
+      @cause == exc.cause
     end
   end
 
+  def matching_exception?(exc)
+    matching_class?(exc) and matching_message?(exc) and matching_cause?(exc)
+  end
+
+  def exception_class_and_message_and_cause(exception_class, message, cause)
+    string = "#{exception_class}"
+    prefixed = false
+    prefix = -> { prefixed ? ", " : prefixed = "(" }
+
+    if message != nil
+      string << "#{prefix.()}#{message.inspect}"
+    end
+
+    if cause != UNDEF_CAUSE
+      string << "#{prefix.()}cause: #{cause.inspect}"
+    end
+
+    string << ")" if prefixed
+
+    string
+  end
+
   def format_expected_exception
-    exception_class_and_message(@exception, @message)
+    exception_class_and_message_and_cause(@exception, @message, @cause)
   end
 
   def format_exception(exception)
-    exception_class_and_message(exception.class, exception.message)
+    exception_class_and_message_and_cause(exception.class,
+                                          @message == nil ? nil : exception.message,
+                                          @cause == UNDEF_CAUSE ? UNDEF_CAUSE : exception.cause)
   end
 
   def failure_message
@@ -87,18 +115,18 @@ class RaiseErrorMatcher
 end
 
 module MSpecMatchers
-  private def raise_error(exception = Exception, message = nil, &block)
-    RaiseErrorMatcher.new(exception, message, &block)
+  private def raise_error(exception = Exception, message = nil, options = nil, &block)
+    RaiseErrorMatcher.new(exception, message, options, &block)
   end
 
   # CRuby < 4.1 has inconsistent coercion errors:
   # https://bugs.ruby-lang.org/issues/21864
   # This matcher ignores the message on CRuby < 4.1
   # and checks the message for all other cases, including other Rubies
-  private def raise_consistent_error(exception = Exception, message = nil, &block)
+  private def raise_consistent_error(exception = Exception, message = nil, options = nil, &block)
     if RUBY_ENGINE == "ruby" and ruby_version_is ""..."4.1"
       message = nil
     end
-    RaiseErrorMatcher.new(exception, message, &block)
+    RaiseErrorMatcher.new(exception, message, options, &block)
   end
 end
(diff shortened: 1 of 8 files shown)


headius:
@eregon There are some modifications to mspec's `raise_error` to support matching causes, modifications to existing specs to use that functionality, and some additional cause-checking specs here, if you want to review.

headius:
Failures appear to be GHA issues or bad timing in specs, unrelated to these changes.

eregon:
Thank you for adding more `cause` specs.

My initial thought here is whether making `raise_error` more complicated for this is worth it:
```ruby
-> { 1 / 0 }.should raise_error(ZeroDivisionError, cause:)
# vs
-> { 1 / 0 }.should raise_error(ZeroDivisionError) { |e| e.cause.should == cause }
```
It looks nice and it's expressive but OTOH it's not used in that many files.

The main concern is MSpec tries to be as simple as possible so it can be used by early Ruby implementations.
That means so far we don't use keyword arguments in MSpec (`git grep -F '**' lib` has none).
That's easily solved though, see https://github.com/ruby/mspec/blob/31e8f11c0bf18cf85fa3d3643ea8fc5a395fc3fe/lib/mspec/matchers/complain.rb#L4-L7 for example.

If you want to modify `raise_error`, could you make a PR https://github.com/ruby/mspec and add spec for it in https://github.com/ruby/mspec/blob/master/spec/matchers/raise_error_spec.rb? Or you could also do it in this PR and run the mspec specs locally.

spec/mspec/lib/mspec/matchers/raise_error.rb
```diff
@@ -45,24 +47,43 @@ def matching_message?(exc)
...
+  end
+
   def matching_exception?(exc)
-    matching_class?(exc) and matching_message?(exc)
+    matching_class?(exc) and matching_message?(exc) and matching_cause?(exc)
   end
 
-  def exception_class_and_message(exception_class, message)
+  def exception_class_and_message_and_cause(exception_class, message, cause)
+    string = +"#{exception_class}"
```
eregon:
```suggestion
    string = "#{exception_class}"
```
(always mutable when interpolated)

spec/mspec/lib/mspec/matchers/raise_error.rb
```diff
@@ -45,24 +47,43 @@ def matching_message?(exc)
...
+    prefix = false
+    prefix = ->{ prefix ? ", " : prefix = "("}
```
eregon:
```suggestion
    prefix = -> { prefix ? ", " : prefix = "("}
```
In specs (and MSpec too) we always use `->` space `{` space ... space `}` for consistency.
This is checked by RuboCop in ruby/spec, so better to have it consistent now than requiring whoever syncs specs to fix it later.

Same for specs below using `->{`.

spec/mspec/lib/mspec/matchers/raise_error.rb
headius: IDE tried to tell me it wouldn't be mutable but I didn't confirm. I'll repair in situ.

spec/mspec/lib/mspec/matchers/raise_error.rb
headius: Will fix.

headius:
> It looks nice and it's expressive but OTOH it's not used in that many files.

I think it will be used in many more as we continue to specify language-level exceptions that should have causes. I've only added a few cases here.

And even without those additional cases, there were many places doing the block form of `raise_error` just to check causes. Those are almost all one-liners now.

headius:
@eregon I cleaned up the commit that adds the `cause:` functionality, added specs for it, and updated others to support the expanded error output. Also fixed all style issues.

spec/ruby/core/exception/cause_spec.rb
```diff
@@ -4,53 +4,34 @@
...
-      rescue RuntimeError => e
-        e.should be_an_instance_of(RuntimeError)
-        e.message.should == "the consequence"
-
-        e.cause.should be_an_instance_of(Exception)
-        e.cause.message.should == "the cause"
-      end
+    rescue Exception => cause
+      -> { raise RuntimeError, "the consequence" }
+        .should raise_error(RuntimeError, "the consequence", cause:)
```
eregon:
```suggestion
      -> {
        raise RuntimeError, "the consequence"
      }.should raise_error(RuntimeError, "the consequence", cause:)
```
This is more consistent with ruby/spec style

spec/mspec/lib/mspec/matchers/raise_error.rb
```diff
@@ -45,24 +53,45 @@ def matching_message?(exc)
...
+  def exception_class_and_message_and_cause(exception_class, message, cause)
+    string = "#{exception_class}"
+    prefixed = false
+    prefix = -> { prefixed ? ", " : prefixed = "(" }
+
     if message
-      "#{exception_class} (#{message})"
-    else
-      "#{exception_class}"
+      string << "#{prefix.()}message: #{message.inspect}"
```
eregon:
```suggestion
      string << "#{prefix.()}#{message.inspect}"
```
A `message: ` prefix every time seems quite noisy, and it's a clear a String is the message and not the cause, so let's omit that.
The `#{message.inspect}` sounds fine.

spec/mspec/lib/mspec/matchers/raise_error.rb
```diff
@@ -1,11 +1,19 @@
 class RaiseErrorMatcher
   FAILURE_MESSAGE_FOR_EXCEPTION = {}.compare_by_identity
+  UNDEF_CAUSE = Object.new
 
   attr_writer :block
 
-  def initialize(exception, message, &block)
+  def initialize(exception, message = nil, options = { }, &block)
```
eregon:
```suggestion
  def initialize(exception, message = nil, options = nil, &block)
```
This is consistent with what `raise_error` below passes.
Also it avoids allocating an extra Hash when it's not used in most cases.

Below it becomes
```ruby
@cause = options ? options.fetch(:cause, UNDEF_CAUSE) : UNDEF_CAUSE
```
which seems simple enough to avoid the extra Hash allocation.

spec/mspec/spec/matchers/raise_error_spec.rb
```diff
@@ -177,7 +177,47 @@ def result.pretty_inspect
...
+    end
+
+    matcher = RaiseErrorMatcher.new(RuntimeError, "foo", cause:)
+    begin
+      matcher.matches?(proc)
+    rescue RuntimeError
+      expect(matcher.failure_message).to eq(
+        ["Expected RuntimeError(message: \"foo\", cause: #<RuntimeError: bar>)", "but got: RuntimeError(message: \"foo\", cause: nil)"]
+      )
+    end
```
eregon, line 232: It would be good to test the case with `cause` but not message too

spec/mspec/spec/matchers/raise_error_spec.rb
```diff
@@ -84,10 +84,10 @@ class UnexpectedException < Exception; end
       matcher.matches?(Proc.new { raise exc })
     rescue UnexpectedException => e
       expect(matcher.failure_message).to eq(
-        ["Expected ExpectedException (message)", "but got: UnexpectedException (message)"]
+        ["Expected ExpectedException(message: \"message\")", "but got: UnexpectedException(message: \"message\")"]
```
eregon: Minor: these would read nicer (and close to what gets printed) if using single-quote string now that we have `"` inside.

eregon (approved):
Thank you, looks good, just needs a few more tweaks and it's good to go

spec/mspec/lib/mspec/matchers/raise_error.rb
headius: I prefer leaving it to be clear, but I don't care that much. Would have been good input before I changed all the specs.

headius:
All requested changes have been made, merging.

spec/mspec/lib/mspec/matchers/raise_error.rb
eregon: I wasn't sure before if fine or not, but looking at the specs made it clear for me, thanks for changing it.

PR #9410, opened 2026-05-02
Fix Hash#inspect of Symbol keys with non-ASCII encoding

sampokuokkanen:
isSymbolName was classifying names via a Java String built by RubyEncoding.decodeRaw (byte-as-Latin-1), losing multi-byte character grouping. Use the encoding-aware SymbolNameType when the string is not ASCII instead.

Re-enables TestHash#test_inspect; also drops a stale no-op exclude in TestHashOnly (no test_inspect method exists in that class).

### DIFF
--- a/core/src/main/java/org/jruby/RubySymbol.java
+++ b/core/src/main/java/org/jruby/RubySymbol.java
@@ -491,8 +491,15 @@ public class RubySymbol extends RubyObject implements MarshalEncoding, EncodingC
         ByteList bytes = fstring.getByteList();
         Encoding enc = bytes.getEncoding();
 
+        boolean nameInvalid;
+        if (str.isAsciiOnly()) {
+            nameInvalid = !isSymbolName(symbol);
+        } else {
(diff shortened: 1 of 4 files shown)


headius:
Nice, thank you! Will wait to see how the tests shake out.

sampokuokkanen:
Thanks! All the tests pass. 

The challenging part was the differences between `isSymbolName(String)` and `SymbolNameType`. `isSymbolName(String)` has checks for global symbols like `:$-w` and other weird Ruby-stuff, but `SymbolNameType` (in `IdUtil.determineSymbolNameType`) sets these as OTHER. However SymbolNameType does work with non-ASCII characters. 

So I guess a follow up PR would be to look more into SymbolNameType possibly miscategorizing symbols.

headius:
Those methods are intended to mimic the same symbol classification logic in cruby but they have never been exactly the same and have diverged frequently. Every so often I run into a case where we need them to work correctly for some optimization or error check. This is a good step in the right direction.

headius (approved):

PR #9411, opened 2026-05-03
Fix Enumerator#each(*args) inheriting parent size

sampokuokkanen:
`Enumerator#each(*args)` returns a new `Enumerator` with the args bound to the underlying iteration. CRuby drops the parent's size block on this new `Enumerator` because the parameterization may change iteration shape. JRuby was passing the parent's `size` and `sizeFn` through to the new `Enumerator`, causing `#size` to lie about parameterized iteration.

Fix: pass `null` for `size` and `sizeFn` when constructing the new `Enumerator` in `RubyEnumerator#each(IRubyObject[], Block)`.

Reproducer:

```ruby
class Repeater
  def each(times = 1)
    times.times { yield }
  end
end

e = Repeater.new.to_enum { 1 }
e.size                # => 1
e.each(5).size        # CRuby: nil   JRuby before fix: 1
e.each(5).to_a.size   # both runtimes: 5
```

CRuby side:
https://github.com/ruby/ruby/blob/db26919bd033dde6a07d18572c621c6959de300f/enumerator.c#L640
```c
e->size = Qnil;
e->size_fn = 0;

```
Re-enables `TestEnumerator#test_each_arg`.

### DIFF
--- a/core/src/main/java/org/jruby/RubyEnumerator.java
+++ b/core/src/main/java/org/jruby/RubyEnumerator.java
@@ -356,7 +356,7 @@ public class RubyEnumerator extends RubyObject implements java.util.Iterator<Obj
         ArraySupport.copy(args, newArgs, mlen, args.length);
 
         return new RubyEnumerator(context.runtime, getType(), object, asSymbol(context, method), newArgs,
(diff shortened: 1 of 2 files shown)


headius (approved):
Yeah good catch and simple fix. Approved!

PR #9412, opened 2026-05-03
Truncate Hash#fetch KeyError message for long keys

sampokuokkanen:
Hash#fetch was including the full inspected key in its KeyError message, regardless of length. CRuby ellipsizes the key at 65 characters (62 chars + "...") via `rb_str_ellipsize` in `hash.c#rb_hash_fetch_m`. JRuby was producing arbitrarily long error messages for large keys.

Re-enables `TestHash#test_fetch_error`.

Behavior after this fix: 
```
irb(main):001> h = {}
=> {}
irb(main):002> h.fetch("hiya" * 200)
org/jruby/RubyHashLinkedBuckets.java:1370:in 'fetch': key not found: 
"hiyahiyahiyahiyahiyahiyahiyahiyahiyahiyahiyahiyahiyahiyahiyah... (KeyError)
```

### DIFF
--- a/core/src/main/java/org/jruby/RubyHashLinkedBuckets.java
+++ b/core/src/main/java/org/jruby/RubyHashLinkedBuckets.java
@@ -1362,7 +1362,12 @@ public class RubyHashLinkedBuckets extends RubyHash {
         if (value != null) return value;
         if (block.isGiven()) return block.yield(context, key);
 
-        throw context.runtime.newKeyError("key not found: " + key.inspect(context), this, key);
+        RubyString desc = (RubyString) key.inspect(context);
+        if (desc.size() > 65) {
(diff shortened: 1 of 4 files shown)


kares:
interesting, 
did not realize MRI has decided to make ~~inspect~~ less useful at some point.

UPDATE: it's only for the error raising, got confused at first.

sampokuokkanen:
Thanks for taking a look.

https://github.com/ruby/ruby/commit/46142e472fd713023766aa2943158d7fd50011ae
The 65 char limit was added here. No other places in CRuby seem to limit the error at 65 characters though. 

Not really sure if this is actually needed. So the alternative could be:
```ruby
exclude :test_fetch_error, "MRI ellipsizes long KeyError keys at 65 chars, we don't"
```

headius (approved):
Looks good!

PR #9419, opened 2026-05-06
Add FreeBSD and OpenBSD runs of ruby-spec

headius:
This is an attempt to add FreeBSD and OpenBSD runs of ruby/spec to JRuby CI.

These use QEMU-based VM actions:

* FreeBSD: https://github.com/vmactions/freebsd-vm
* OpenBSD https://github.com/vmactions/openbsd-vm

Current status:

* The OpenBSD image used by the action claims to be 7.8 but appears to be connected to an older set of ports. The newest JDK offered is 17. I attempted to point it at a different ports location but still failed to find JDK 21.
* Core specs fail on FreeBSD with various OS-level and time-related failures. These specs may never have been tested on FreeBSD, so it might not indicate a problem with JRuby.

cc @jeremyevans for assistance figuring out OpenBSD JDK 21 and which of these spec failure are expected.

cc @eregon for any interest in getting ruby/spec running on *BSD. It is trivial to do with JRuby but running on other Ruby implementations will require being able to build or install them on these platforms.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -407,6 +407,7 @@ jobs:
             pkg install -y openjdk21
           run: |
             export JAVA_HOME="/usr/local/openjdk21"
+            export TZ="America/Chicago"
             bin/jruby spec/mspec/bin/mspec ci :${{ matrix.rubyspec-target }}
 
   spec-ruby-openbsd:
@@ -439,6 +440,7 @@ jobs:
             pkg_add jdk--%21
           run: |
             export JAVA_HOME="/usr/local/jdk-21"
+            export TZ="America/Chicago"
             bin/jruby spec/mspec/bin/mspec ci :${{ matrix.rubyspec-target }}
 
   dependency-check:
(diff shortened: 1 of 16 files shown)


eregon:
> cc @eregon for any interest in getting ruby/spec running on *BSD. It is trivial to do with JRuby but running on other Ruby implementations will require being able to build or install them on these platforms.

Adding that in ruby/spec in CI is probably not gonna happen until there are free *BSD GitHub-hosted runners (no emulation).
We can't just use JRuby instead either because it doesn't pass all the specs and we don't maintain a set of tags in ruby/spec.

eregon:
https://rubyci.org/ does run on BSD though, so effectively ruby/spec is already passing on BSD.

headius:
> We can't just use JRuby

I never suggested that. I just said it would be eaasy with JRuby since it doesn't require a rebuild for each platform.

Obviously if we were adding a *BSD job to ruby-spec it would run with CRuby, but that would require building it from scratch.

> effectively ruby/spec is already passing on BSD

Only the version of ruby/spec that has been committed to ruby/ruby runs in that CI. We should include equivalent BSD runs on the ruby/spec repository to catch breakage early.

Issue #9257, opened 2026-02-22
Bugs in `Method`/`UnboundMethod#original_name`


trinistr:
**Environment Information**
  
Provide at least:
* `jruby 10.0.4.0-SNAPSHOT (3.4.5) 2026-02-22 ffffffffff OpenJDK 64-Bit Server VM 21.0.10+7 on 21.0.10+7 +indy +jit [x86_64-linux]` (`master`)
* `Linux meow4 6.18.4-1-MANJARO #1 SMP PREEMPT_DYNAMIC Thu, 08 Jan 2026 12:46:08 +0000 x86_64 GNU/Linux`

# Issue 1: transplanted and/or aliased methods display wrong `#original_name`
**Expected Behavior**

Given this file:
```ruby
class A < BasicObject
  define_method(:method, ::Kernel.instance_method(:method))

  define_method(:is_a?, ::Kernel.instance_method(:is_a?))
  alias is_stuff? is_a?
  define_method(:kind_of?, ::Kernel.instance_method(:kind_of?))
  alias kind_of_stuff? kind_of?

  define_method(:my_inspect, ::Kernel.instance_method(:inspect))
end

puts A.instance_method(:is_a?).original_name
puts A.new.method(:is_a?).original_name
puts A.instance_method(:is_stuff?).original_name
puts A.new.method(:is_stuff?).original_name

puts A.instance_method(:kind_of?).original_name
puts A.new.method(:kind_of?).original_name
puts A.instance_method(:kind_of_stuff?).original_name
puts A.new.method(:kind_of_stuff?).original_name

puts A.instance_method(:my_inspect).original_name
puts A.new.method(:my_inspect).original_name
```

Running it, I expect to see:
```
is_a?
is_a?
is_a?
is_a?
kind_of?
kind_of?
kind_of?
kind_of?
inspect
inspect
```
(on MRI `ruby 3.3.9 (2025-07-24 revision f5c772fc7c) [x86_64-linux]`)

**Actual Behavior**

```
kind_of?
is_a?
kind_of?
kind_of?
kind_of?
kind_of?
kind_of?
kind_of?
inspect
my_inspect
```
1. `is_a?` appears only for unaliased `Method`, but it shouldn't be considered an alias of `kind_of?`
2. `my_inspect` appears only for `Method`, but it shouldn't

# Issue 2: `attr_accessor` and co are named by ivar

**Expected Behavior**

Given this file:
```ruby
class A
  attr_accessor :var
  alias var_r var
[...]

PR #9420, opened 2026-05-07
Fix Method/UnboundMethod#original_name deep alias chains

sampokuokkanen:
Use getRealMethod() instead of entry.method, which only unwrapped one level. Fixes Method#/UnboundMethod#original_name for chains 3+ deep. Also drops the stale RubyMethod#original_name override left over from when the implementation moved to AbstractRubyMethod. 

Fixes the first part of https://github.com/jruby/jruby/issues/9257. (except for the usage of `is_a?`)

I added tests here: https://github.com/ruby/spec/pull/1353 and here: https://github.com/ruby/spec/pull/1354. 

With this the sample from the example in the issue becomes the same as MRI. I'm not sure about the breaking up of `kind_of?` and `is_a?` to different Java methods, but this seems to be the way things are done for methods like `map`/`collect`. 

Sample code: 

```ruby
class Test
  def hello = "hello"
  alias hallo hello
  alias moi hallo
  alias tere moi
  alias hej tere
end

Test.new.method(:hej).original_name
=> :hello # MRI
=> :moi # JRuby 10.1.0.0
=> :hello # JRuby on this branch
```

### DIFF
--- a/core/src/main/java/org/jruby/internal/runtime/methods/AliasMethod.java
+++ b/core/src/main/java/org/jruby/internal/runtime/methods/AliasMethod.java
@@ -158,9 +158,9 @@ public class AliasMethod extends DynamicMethod {
 
 
     public String getOldName() {
-        return entry.method.getName();
+        return getRealMethod().getName();
     }
-    
+
     @Override
     public DynamicMethod getRealMethod() {
         return entry.method.getRealMethod();
--- a/core/src/main/java/org/jruby/RubyMethod.java
+++ b/core/src/main/java/org/jruby/RubyMethod.java
@@ -299,11 +299,6 @@ public class RubyMethod extends AbstractRubyMethod {
         return super_method(context, receiver, superClass);
     }
 
-    @JRubyMethod
-    public IRubyObject original_name(ThreadContext context) {
-        return method instanceof AliasMethod ? asSymbol(context, ((AliasMethod)method).getOldName()) : name(context);
-    }
-
     public IRubyObject getReceiver() {
         return receiver;
     }


core/src/main/java/org/jruby/internal/runtime/methods/AliasMethod.java
```diff
@@ -158,9 +158,9 @@ public Signature getSignature() {
 
 
     public String getOldName() {
-        return entry.method.getName();
+        return getRealMethod().getName();
```
sampokuokkanen, line 161:
```Java
if (entry.method instanceof AliasMethod innerMethod) {
    return innerMethod.getOldName();
}

return entry.method.getName();
```

Would also work.

headius: Using the override is cleaner, so let's leave it that way until there's a compelling reason to switch to the branch.

core/src/main/java/org/jruby/RubyKernel.java
```diff
@@ -2466,11 +2466,16 @@ public static IRubyObject itself(IRubyObject self) {
         return self;
     }
 
-    @JRubyMethod(name = {"kind_of?", "is_a?"})
```
headius, line 2469: I don't think there's any need to split these up. CRuby also defines the exact same code for `kind_of?` and `is_a?`, even down to the optimized JIT versions. There's also no benefit for us on JVM; the two calls inline and optimize the same.

Other methods that have the same function are usually split up to support `super` using the correct name, since without splitting it's trickier for us to know which name was called (for now... improvements are coming).

Let's just leave these as one method for now.

headius (changes requested):
Let's put `is_a?` back the way it was. The rest is good. 👍

core/src/main/java/org/jruby/RubyKernel.java
sampokuokkanen: In CRuby:
```C
rb_define_method(rb_mKernel, "kind_of?", rb_obj_is_kind_of, 1);
rb_define_method(rb_mKernel, "is_a?", rb_obj_is_kind_of, 1);
```
So they are defined as methods on Kernel, not aliases. 

In JRuby, but there is a difference when it comes to `original_name` since `is_a?` is an alias: 
```ruby
# CRuby
Class.new.method(:is_a?)
=> #<Method: #<Class:#<Class:0x0000000126cb1548>>(Kernel)#is_a?(_)>
Class.new.method(:is_a?).original_name
=> :is_a?

#JRuby
Class.new.method(:is_a?)
=> #<Method: #<Class:#<Class:0x6114902c>>(Kernel)#is_a?(kind_of?)(_)>
Class.new.method(:is_a?).original_name
=> :kind_of?
```

I reverted my changes in `RubyKernel.java`. 
I would guess a cleaner way to handle this would be to have a clearer rb_define_method and rb_define_alias separation like CRuby. Looking forward to the improvements you mentioned. 👍

sampokuokkanen:
Thank you for the review. Removed the change from `RubyKernel.java`.

core/src/main/java/org/jruby/RubyKernel.java
headius: In the `@JRubyMethod` annotation, there's two ways to define a single method with multiple names:

* Using `alias` is equivalent to `rb_define_alias`. CRuby almost never uses this in core code.
* Using multiple names in `name` is equivalent to `rb_define_method`, which is what CRuby usually does.

So the code as written is equivalent to having two `rb_define_method` calls, and you can see this in the generated `POPULATOR` class for `Kernel`:

```java
public class org$jruby$RubyKernel$POPULATOR extends TypePopulator {
    public void populate(ThreadContext context, RubyModule cls, Class clazz) {
...
        javaMethod = new org.jruby.RubyKernel$INVOKER$s$1$0$kind_of_p(cls, Visibility.PUBLIC, "kind_of?");
        populateMethod(javaMethod, 1, "kind_of_p", true, false, core, org.jruby.RubyKernel.class, "kind_of_p", org.jruby.RubyBoolean.class, CONTEXT_ARG2);
        aliasedMethod = cls.putMethod(context, "kind_of?", javaMethod);
        cls.putMethod(context, "is_a?", javaMethod);
...
```

Both methods are defined using `putMethod` rather than `putAlias`.

headius (approved):

PR #9421, opened 2026-05-07
Fix missing frozen check in 1 and 2-element Array sort!/reverse!

sampokuokkanen:
Call modifyCheck(context) in RubyArrayTwoObject to enforce frozen/modify checks during in-place operations. Added checks in reverse_bang and in the sort/compare paths (after block yield and inside compareOthers) so attempts to freeze or modify the array during reverse! or sort! will raise FrozenError. Also added unit tests to exercise the two-element specialization: reverse! on a frozen array, freezing the array inside a sort! block, and freezing during <=> comparator.

JRuby 10.1.0.0:
```ruby
[1, 2].freeze.reverse!
=> [2, 1]
```

This branch: 
```ruby
[1, 2].freeze.reverse!
org/jruby/specialized/RubyArrayTwoObject.java:229:in 'reverse!': can't modify frozen Array: [1, 2] (FrozenError)
        from (irb):1:in 'evaluate'
        from org/jruby/RubyKernel.java:1448:in 'eval'
        from org/jruby/RubyKernel.java:1850:in 'loop'
        from org/jruby/RubyKernel.java:1587:in 'catch'
        from /Users/sampokuokkanen/jruby/lib/ruby/gems/shared/gems/irb-1.16.0/exe/irb:9:in '<main>'
        from org/jruby/RubyKernel.java:1402:in 'load'
        from ./bin/irb:25:in '<main>'
```

The reason why I noticed this is I was looking into https://github.com/jruby/jruby/blob/master/test/mri/ruby/test_array.rb#L1827 and when playing around with it I noticed that the third test case failed, but it was masked by the `(count += 1) == 6` check failing. 

I also created a PR agains MRI:
https://github.com/ruby/ruby/pull/16885
If they accept this PR we can enable `test_freeze_inside_sort!`. 

RubyArrayOneObject was also missing the check in `.reverse!` Added a test case and confirmed that `[1].freeze.reverse!` will raise a FrozenError.

### DIFF
--- a/test/jruby/test_array.rb
+++ b/test/jruby/test_array.rb
@@ -183,4 +183,22 @@ class TestArray < Test::Unit::TestCase
     assert_equal([], rules)
   end
 
+  # 2-element arrays use the RubyArrayTwoObject specialization.
+  def test_reverse_bang_two_element_frozen
+    assert_raise(FrozenError) { [1, 2].freeze.reverse! }
+  end
+
+  def test_sort_bang_two_element_freeze_in_block
+    array = [1, 2]
+    assert_raise(FrozenError) do
+      array.sort! { |_a, _b| array.freeze; -1 }
+    end
+  end
+
+  def test_sort_bang_two_element_freeze_in_cmp
+    o1, o2 = Object.new, Object.new
+    array = [o1, o2]
+    o1.define_singleton_method(:<=>) { |_| array.freeze; -1 }
+    assert_raise(FrozenError) { array.sort! }
+  end
 end
--- a/test/jruby/test_array.rb
+++ b/test/jruby/test_array.rb
@@ -183,8 +183,9 @@ class TestArray < Test::Unit::TestCase
     assert_equal([], rules)
   end
 
-  # 2-element arrays use the RubyArrayTwoObject specialization.
+  # 1- and 2-element arrays use the RubyArrayOneObject and RubyArrayTwoObject specializations.
   def test_reverse_bang_two_element_frozen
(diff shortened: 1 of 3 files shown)


headius:
Good catch! There's probably more we could do to make those specialized Array subtypes share more logic, so we don't have one-off bugs like this.

Also thanks for pushing the MRI change, I see it was merged already.

headius:
This should probably be rebased on the `jruby-10.0` branch given that it's a pretty obvious bug for a frozen 2-element array.

sampokuokkanen:
Thanks for reviewing this. 

Rebased on the `jruby-10.0` branch.

sampokuokkanen:
@headius 
Sorry for not noticing it sooner, but `RubyArrayOneObject` also had the same bug in `.reverse!`

I added a commit fixing it. It is only a one line fix so I hope it's ok to include it in this PR.

headius:
@sampokuokkanen Yes that's just fine.

test/jruby/test_array.rb
```diff
@@ -183,4 +183,23 @@ def test_delete_if_with_modification
     assert_equal([], rules)
   end
 
+  # 1- and 2-element arrays use the RubyArrayOneObject and RubyArrayTwoObject specializations.
```
headius, line 186: Normally we'd want behavioral tests to go in ruby-spec but since these cases are pretty JRuby-specific I think this is fine for now.

headius (approved):

headius:
@sampokuokkanen Thank you!

PR #9423, opened 2026-05-07
Update specs

eregon:


headius:
@eregon Are you unable to merge still? I thought I gave you permission to skip the review.

eregon:
@headius Indeed, I can't merge:
Screenshot From 2026-05-08 20-33-25

I think the fix is either:
* Don't require an approval on the PR (anyway only people with write access can merge PRs), or
* Add me to the bypass list

headius:
I disabled the required reviews for now.

PR #9425, opened 2026-05-08
ivarSet missing return after megamorphic fallback

kares:
observed with debug (using log.binding)

before the fix, a megamorphic site logged"failed (polymorphic)" followed by a "bound directly" for the same type.

seems to me like this isn't intentional esp. given the `ivarGet` logic ends up returning after  `invokedynamic.maxpoly` is reached...


headius (approved):
You are correct. This logic could be cleaned up (and hopefully will be soon).

Issue #9296, opened 2026-03-09
NullPointerException in CacheEntry.typeOk during DefineModuleInstr on JRuby 10.0.4.0


brometeo:
My JRuby client application is very sophisticated and complex. It works with JRuby 9.4.8.0, but updating from that version brings some varied bugs.

trying JRuby 10.0.4.0 I get an error of type Stack trace: DefineModuleInstr.interpret → CachingCallSite.call → CacheEntry.typeOk → NPE "incomingType is null".

My system is Debian Linux testing with Java 21+, and applications uses Swing with AWT Event Dispatch Thread. Perhaps there are some problems with EDT and module and class loading, but I don't see the origin in stack trace, so it is a frustrating situation.

I need some help.

headius:
NPEs, especially within interpreter code, is nearly always a JRuby bug. There were probably new behaviors here in 10 that don't properly null check all paths.

brometeo:
El lunes 09 de marzo, Charles Oliver Nutter escribió:
> NPEs, especially within interpreter code, is nearly always a JRuby bug. There
> were probably new behaviors here in 10 that don't properly null check all
> paths.

I can give the error backtrace if needed.

-- 
David

headius:
@brometeo Yes please! I can eyeball the `DefineModuleInstr` code but the NPE could be coming from somewhere else.

brometeo:
El lunes 09 de marzo, Charles Oliver Nutter escribió:
> @brometeo Yes please! I can eyeball the `DefineModuleInstr` code but the NPE
> could be coming from somewhere else.

Here is the backtrace:

GRAVE: Error loading appliation: Cannot invoke "org.jruby.RubyClass.getGeneration()" because "incomingType" is null
org.jruby.dist/org.jruby.runtime.callsite.CacheEntry.typeOk(CacheEntry.java:27)
org.jruby.dist/org.jruby.runtime.callsite.CachingCallSite.call(CachingCallSite.java:192)
org.jruby.dist/org.jruby.ir.interpreter.InterpreterEngine.processCall(InterpreterEngine.java:352)
org.jruby.dist/org.jruby.ir.interpreter.StartupInterpreterEngine.interpret(StartupInterpreterEngine.java:66)
org.jruby.dist/org.jruby.ir.interpreter.Interpreter.interpretFrameScope(Interpreter.java:174)
[...]

PR #9426, opened 2026-05-08
prevent UNDEF from leaking through constant resolution

kares:
`iterateConstantNoConstMissing` with `loadConstant=false`, the `getConstantSkipAutoload` call path (used by `InheritanceSearchConst`), could return `UNDEF` instead of` null` when an autoload entry existed for the constant;

could than lead to a method call receiver causing a NPE in CacheEntry.typeOk (due null meta-class for the UNDEF)

---

had this around for a while, was looking at potential issues with AI tooling when looking at #9296 it's hard to tell whether that is the issue (`autoload` usage ended up unconfirmed) but regardless the fix here seems legit, as there indeed was a way to leak-out `UNDEF` used as a marked for autoloads...

### DIFF
--- a/core/src/main/java/org/jruby/RubyModule.java
+++ b/core/src/main/java/org/jruby/RubyModule.java
@@ -5441,8 +5441,13 @@ public class RubyModule extends RubyObject {
                     mod.getConstantWithAutoload(context, name, null, true) :
                     mod.fetchConstant(context, name, true);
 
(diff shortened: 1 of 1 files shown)


headius:
The issue appears to be real but the fix isn't happy.

kares:
actually, this no longer makes sense

PR #9441, opened 2026-05-12
Fix original_name when two methods share a wrapper (like is_a? kind_of?)

### LINKED ISSUE #9257: Bugs in `Method`/`UnboundMethod#original_name`

sampokuokkanen:
Preserve a method's source name when methods share a single underlying wrapper, so `Method#original_name` and `#inspect` report the right name instead of whichever name happened to live on the shared wrapper. The dup'd method is wrapped in a per-binding `RenamedDynamicMethod` (a `DelegatingDynamicMethod` subclass) holding the source name in a final field. `Method#original_name` consults `DynamicMethod#getOldName`, which the wrapper overrides and `AliasMethod` walks through.

Covered by new specs in ruby/spec#1356 (https://github.com/ruby/spec/pull/1356).

Sample (the case this PR fixes):

```ruby
class A
  define_method(:my_is_a?, Kernel.instance_method(:is_a?))
  define_method(:my_kind_of?, Kernel.instance_method(:kind_of?))
end

# master:    both report the same name (last writer wins on the shared wrapper)
# this PR:
A.new.method(:my_is_a?).original_name    # => :is_a?
A.new.method(:my_kind_of?).original_name # => :kind_of?
```

Known limitation: direct lookup of multi-name `@JRubyMethod` methods still leaks the underlying name (`Class.new.method(:is_a?).inspect` still shows `(kind_of?)`). The specs in ruby/spec#1356 for the inspect case for like `Class.new.method(:id_a?)` will fail, but I think that would be better tackled in a follow-up PR. 

This PR fixes the first part of https://github.com/jruby/jruby/issues/9257.

@headius
Thanks for explaining to me how method lookup works in JRuby in https://github.com/jruby/jruby/pull/9420.

### DIFF
--- a/core/src/main/java/org/jruby/RubyModule.java
+++ b/core/src/main/java/org/jruby/RubyModule.java
@@ -92,6 +92,7 @@ import org.jruby.internal.runtime.methods.PartialDelegatingMethod;
 import org.jruby.internal.runtime.methods.ProcMethod;
 import org.jruby.internal.runtime.methods.RefinedMarker;
 import org.jruby.internal.runtime.methods.RefinedWrapper;
+import org.jruby.internal.runtime.methods.RenamedDynamicMethod;
 import org.jruby.internal.runtime.methods.SynchronizedDynamicMethod;
 import org.jruby.internal.runtime.methods.UndefinedMethod;
 import org.jruby.ir.IRClosure;
@@ -2991,8 +2992,7 @@ public class RubyModule extends RubyObject {
 
             checkValidBindTargetFrom(context, (RubyModule) method.owner(context), false);
 
-            newMethod = method.getMethod().dup();
-            newMethod.setSourceName(method.getMethodName());
+            newMethod = new RenamedDynamicMethod(method.getMethod().dup(), method.getMethodName());
             newMethod.setImplementationClass(this);
             newMethod.setVisibility(visibility);
         } else {
--- a/core/src/main/java/org/jruby/RubyModule.java
+++ b/core/src/main/java/org/jruby/RubyModule.java
@@ -2992,6 +2992,7 @@ public class RubyModule extends RubyObject {
             checkValidBindTargetFrom(context, (RubyModule) method.owner(context), false);
 
             newMethod = method.getMethod().dup();
+            newMethod.setSourceName(method.getMethodName());
             newMethod.setImplementationClass(this);
             newMethod.setVisibility(visibility);
         } else {
--- a/core/src/main/java/org/jruby/internal/runtime/methods/AliasMethod.java
+++ b/core/src/main/java/org/jruby/internal/runtime/methods/AliasMethod.java
@@ -158,7 +158,12 @@ public class AliasMethod extends DynamicMethod {
 
 
     public String getOldName() {
-        return getRealMethod().getName();
+        DynamicMethod terminal = entry.method;
+        while (terminal instanceof AliasMethod) {
+            terminal = ((AliasMethod) terminal).entry.method;
+        }
+        String source = terminal.getSourceName();
+        return source != null ? source : terminal.getName();
     }
 
     @Override
--- a/core/src/main/java/org/jruby/internal/runtime/methods/AliasMethod.java
+++ b/core/src/main/java/org/jruby/internal/runtime/methods/AliasMethod.java
@@ -157,13 +157,13 @@ public class AliasMethod extends DynamicMethod {
     }
 
 
+    @Override
     public String getOldName() {
         DynamicMethod terminal = entry.method;
-        while (terminal instanceof AliasMethod) {
-            terminal = ((AliasMethod) terminal).entry.method;
+        while (terminal instanceof AliasMethod alias) {
+            terminal = alias.entry.method;
         }
-        String source = terminal.getSourceName();
-        return source != null ? source : terminal.getName();
+        return terminal.getOldName();
     }
 
     @Override
--- a/core/src/main/java/org/jruby/internal/runtime/methods/DynamicMethod.java
+++ b/core/src/main/java/org/jruby/internal/runtime/methods/DynamicMethod.java
@@ -77,6 +77,9 @@ public abstract class DynamicMethod {
     protected byte flags;
     /** The simple, base name this method was defined under. May be null.*/
     protected final String name;
+    /** The original source name when this method was bound under a different name
+     *  via define_method(name, Method/UnboundMethod); null otherwise. */
+    protected String sourceName;
     /** An arbitrarily-typed "method handle" for use by compilers and call sites */
     protected Object handle;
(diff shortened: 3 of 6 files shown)


core/src/main/java/org/jruby/internal/runtime/methods/AliasMethod.java
```diff
@@ -158,7 +158,12 @@ public Signature getSignature() {
 
 
     public String getOldName() {
-        return getRealMethod().getName();
+        DynamicMethod terminal = entry.method;
+        while (terminal instanceof AliasMethod) {
+            terminal = ((AliasMethod) terminal).entry.method;
+        }
+        String source = terminal.getSourceName();
+        return source != null ? source : terminal.getName();
```
headius: Perhaps `getSourceName` should do this null check instead? We don't want to be null-checking everywhere.

Also wondering if we can avoid introducing a new method name and lift `getOldName` up to `DynamicMethod` doing basically what `getSourceName` does in your PR. I'm reluctant to add yet another way to get a method's name.

core/src/main/java/org/jruby/RubyModule.java
```diff
@@ -2992,6 +2992,7 @@ public IRubyObject defineMethodFromCallable(ThreadContext context, IRubyObject a
             checkValidBindTargetFrom(context, (RubyModule) method.owner(context), false);
 
             newMethod = method.getMethod().dup();
+            newMethod.setSourceName(method.getMethodName());
```
headius: If we made the source name final we could modify this logic to `dupWithSourceName` or similar.

core/src/main/java/org/jruby/internal/runtime/methods/DynamicMethod.java
```diff
@@ -77,6 +77,9 @@ public abstract class DynamicMethod {
     protected byte flags;
     /** The simple, base name this method was defined under. May be null.*/
     protected final String name;
+    /** The original source name when this method was bound under a different name
+     *  via define_method(name, Method/UnboundMethod); null otherwise. */
+    protected String sourceName;
```
headius: Wondering if we can make `sourceName` final since it's only set for dup'ed methods passed to `define_method`.

headius (changes requested):
I think we can distill this down a bit:

* We don't need both `getOldName` and `getSourceName`. We should either lift the former up or deprecate it and use the latter, I think.
* AliasMethod should override whichever method is the standard "get original method name" going forward, so we're not checking for AliasMethod and null values all over.
* The `sourceName` stored in `DynamicMethod` can be final, I think. It is only initialized during method duping, which could call an internal constructor.

core/src/main/java/org/jruby/internal/runtime/methods/AliasMethod.java
sampokuokkanen: I ended up dropping the null checks entirely and removing getSourceName along the way (had to introduce a `RenamedDynamicMethod` class for this special case). Lifted getOldName onto DynamicMethod as suggested and this way it's much cleaner.

core/src/main/java/org/jruby/RubyModule.java
sampokuokkanen: The only simple way I could make final work was by introducing `RenamedDynamicMethod` .

core/src/main/java/org/jruby/internal/runtime/methods/DynamicMethod.java
sampokuokkanen: Had problems with `Object.clone()` if I tried making it final, so ended up doing it by introducing `RenamedDynamicMethod` for this case.

sampokuokkanen:
Refactored: for `define_method(name, Method/UnboundMethod)`, the dup'd method is now wrapped in a per-binding `RenamedDynamicMethod` that holds sourceName as a final field. No more mutable field or setter on `DynamicMethod`.

headius (approved):
Much cleaner after updates. Sorry for the delay... approved!

PR #9444, opened 2026-05-14
Allow bytecode version to be open-ended

### LINKED ISSUE #9443: JRuby 9.4 does not generate bytecode for newer JDK versions
The version guard is still maxed out at Java 18 but should be modified to support newer releases. At the very least, update to support bytecode versions up to 25 and if it can safely be made open-ended, do that (test that with newer JDK than the shipped version of ASM supports).

headius:
The ASM library hard codes a set of known bytecode versions, which frequently forces us to select an earlier version than the JVM we are running on. This change makes that byte code version open-ended, decoupling us from the ASM supported bytecode versions.

This is somewhat experimental, since we do not know how ASM will react if the JDK version is much newer than it's known byte code versions. If this patch does not work, we will fall back on updating ASM to latest and dealing with and updating our version calculation in the future.

Fixes #9443 

Update: as suspected, ASM will not permit us to use bytecode versions newer than it supports, so I have made an additional change to limit the open-endedness of this to just the values the current ASM version supports.

### DIFF
--- a/core/src/main/java/org/jruby/RubyInstanceConfig.java
+++ b/core/src/main/java/org/jruby/RubyInstanceConfig.java
@@ -55,6 +55,7 @@ import static org.jruby.util.StringSupport.EMPTY_STRING_ARRAY;
 import org.objectweb.asm.Opcodes;
 
 import java.io.*;
+import java.lang.reflect.Field;
 import java.nio.charset.StandardCharsets;
 import java.util.ArrayList;
 import java.util.Collection;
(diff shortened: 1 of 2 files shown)


headius:
As I feared, if ASM doesn't know about the bytecode version, it will not allow it. I will investigate if there is a way to programmatically determine the highest version it supports because I really want to eliminate this manual code.

headius:
Latest changes seem to work but I want to add a test to verify that it is picking the latest version.

Issue #7787, opened 2023-05-10
nil.singleton_method does not find method defined on nil


headius:
Investigating this with @jeremyevans I discovered JRuby has some broken behavior around methods defined on nil:

```
[] jruby $ jruby -e 'def nil.foo; end; p nil.singleton_method(:foo)'
NameError: undefined method `foo' for `nil'
  singleton_method at org/jruby/RubyKernel.java:2319
            <main> at -e:1
[] jruby $ (chruby 3.2 ; ruby -e 'def nil.foo; end; p nil.singleton_method(:foo)')
#<Method: NilClass#foo() -e:1>
```

This behavior does not match CRuby, but CRuby also is not consistent here in that `singleton_methods` does **not** see the method:

https://bugs.ruby-lang.org/issues/11064

I believe it should consistently see the method or consistently not see the method, but a decision would have to be made on the CRuby side. In the short term, we have never had this filed as an issue, so I'm inclined to leave it alone until such time as ruby-core decides it should be visible through these methods.

sampokuokkanen:
Seems like CRuby actually now agrees with JRuby and we can close this issue! The tests added to ruby/spec also pass on JRuby as is and there are no exclusions for them. 
https://github.com/ruby/ruby/pull/7973

headius:
OMG I didn't think to retest CRuby. Indeed, it errors.

I'd love to get @jeremyevans input here but I agree, we can close this.

jeremyevans:
Yep we can close this. This is one of the bugs we investigated during the RubyKaigi 2023 Rubyists on Rails train between Tokyo and Matsumoto. Bug was originally reported back in 2015 (https://bugs.ruby-lang.org/issues/11064). :)

PR #9446, opened 2026-05-19
Fix metaclass dispatch missing meta-metaclass methods

### LINKED ISSUE #287: Class does not inherit singleton_class's class methods

### LINKED ISSUE #6119: shopify-api gem reporting NoMethodError: undefined method `threadsafe_attribute'
shopify-api gem does not work.  Early versions used to work (a few years back).

**Environment Information**
  JRuby  9.2.11.0
  Windows 10

Easy way to reproduce:
```
gem install shopify_api
gem install shopify_api_console
shopify-api console
```

```
NoMethodError: undefined method `threadsafe_attribute' for #<Class:ShopifyAPI::Base>
       method_missing at org/jruby/RubyBasicObject.java:1708
      singleton class at C:/jruby-9.2.11.0/lib/ruby/gems/shared/gems/shopify_api-9.0.2/lib/shopify_api/resources/base.rb:34
         <class:Base> at C:/jruby-9.2.11.0/lib/ruby/gems/shared/gems/shopify_api-9.0.2/lib/shopify_api/resources/base.rb:33
  <module:ShopifyAPI> at C:/jruby-9.2.11.0/lib/ruby/gems/shared/gems/shopify_api-9.0.2/lib/shopify_api/resources/base.rb:4
               <main> at C:/jruby-9.2.11.0/lib/ruby/gems/shared/gems/shopify_api-9.0.2/lib/shopify_api/resources/base.rb:3
              require at org/jruby/RubyKernel.java:974
              require at C:/jruby-9.2.11.0/lib/ruby/stdlib/rubygems/core_ext/kernel_require.rb:54
               <main> at C:/jruby-9.2.11.0/lib/ruby/gems/shared/gems/shopify_api-9.0.2/lib/shopify_api/resources.rb:1
              require at org/jruby/RubyKernel.java:974
              require at C:/jruby-9.2.11.0/lib/ruby/stdlib/rubygems/core_ext/kernel_require.rb:54
               <main> at C:/jruby-9.2.11.0/lib/ruby/gems/shared/gems/shopify_api-9.0.2/lib/shopify_api.rb:21
              require at org/jruby/RubyKernel.java:974
              require at C:/jruby-9.2.11.0/lib/ruby/stdlib/rubygems/core_ext/kernel_require.rb:54
               <main> at C:/jruby-9.2.11.0/lib/ruby/gems/shared/gems/shopify_api_console-2.0.0/lib/shopify_api_console/console.rb:4
              require at org/jruby/RubyKernel.java:974
              require at C:/jruby-9.2.11.0/lib/ruby/stdlib/rubygems/core_ext/kernel_require.rb:54
               <main> at C:/jruby-9.2.11.0/lib/ruby/gems/shared/gems/shopify_api_console-2.0.0/bin/shopify-api:3
                 load at org/jruby/RubyKernel.java:1009
               <main> at C:\jruby-9.2.11.0\bin\shopify-api:23
```

Works fine with normal ruby-2.5.7-1-x64 from https://www.ruby-lang.org

sampokuokkanen:
The sample code from https://github.com/jruby/jruby/issues/287 still errors on JRuby master: 

```ruby
class Foo
  class << self
    class << self
      def foo; end
    end
  end
end
class Bar < Foo; end
Bar.singleton_class.foo
(irb):11:in 'evaluate': undefined method 'foo' for class #<Class:Bar> (NoMethodError)
```

This is because currently in master the code is stripping away singleton classes and not creating a singleton class in case one is missing. 

So skip the wrapper with `isIncluded()` and use `singletonClass(context)` to get the correct eigenclass. 
This lets us drop `spec/tags/ruby/language/metaclass_tags.txt`, but the MRI tests are still failing since they go one step further (meta-meta-meta tier, blows up on line 160), and with meta-meta classes there is still an issue there with how they are handled (see MetaClass.java:109, can't just fix it as it leads to a stack overflow with the same shape of fix as I did on line 542)

### DIFF
--- a/core/src/main/java/org/jruby/RubyBasicObject.java
+++ b/core/src/main/java/org/jruby/RubyBasicObject.java
@@ -538,7 +538,9 @@ public class RubyBasicObject implements Cloneable, IRubyObject, Serializable, Co
         MetaClass klass = new MetaClass(context.runtime, superClass, this); // rb_class_boot
         setMetaClass(klass);
 
-        klass.setMetaClass(superClass.getRealClass().metaClass);
+        // MRI: SET_METACLASS_OF(metaclass, ENSURE_EIGENCLASS(tmp))
+        RubyClass effectiveSuper = superClass.isIncluded() ? superClass.getRealClass() : superClass;
+        klass.setMetaClass(effectiveSuper.singletonClass(context));
 
         return klass;
     }
--- a/spec/tags/ruby/language/metaclass_tags.txt
+++ b/spec/tags/ruby/language/metaclass_tags.txt
@@ -1,1 +1,0 @@
-fails:calling methods on the metaclass calls a method defined on the metaclass of the metaclass


core/src/main/java/org/jruby/RubyBasicObject.java
```diff
@@ -538,7 +538,9 @@ public RubyClass makeMetaClass(ThreadContext context, RubyClass superClass) {
         MetaClass klass = new MetaClass(context.runtime, superClass, this); // rb_class_boot
         setMetaClass(klass);
 
-        klass.setMetaClass(superClass.getRealClass().metaClass);
+        // MRI: SET_METACLASS_OF(metaclass, ENSURE_EIGENCLASS(tmp))
+        RubyClass effectiveSuper = superClass.isIncluded() ? superClass.getRealClass() : superClass;
```
kares, line 542: not sure I follow, is the included check really needed?

sampokuokkanen: While writing this I first thought we wouldn't need the check, but the code didn't work without it. Turns out we can get a BlankSlateWrapper here instead of the actual module (from JavaPackage.createJavaPackageClass).

Should I add `// unwrap IncludedModuleWrapper (e.g. BlankSlateWrapper from JavaPackage)` as an inline comment?

sampokuokkanen: Without it on boot: 
`java.lang.UnsupportedOperationException: An included class is only a wrapper for a module`

headius (approved):
Aha, I think I understand after looking at the code.

An object's singleton class's metaclass should be the singleton class of the object's class's superclass. When the object's class has included a module, the module's singleton class should be used.

The old logic simply used the superclass's (or supermodule's) metaclass without acquiring its singleton class. This resulted in future method lookups traversing the metaclass hierarchy rather than the singleton class hierarchy.

It's a complicated system but this seems more correct. I'm sure there's other tags and specs passing now because of this.

headius:
I hoped it might fix https://github.com/jruby/jruby/issues/7787 but it does not. Perhaps another challenge?

sampokuokkanen:
Thanks, I'll look into https://github.com/jruby/jruby/issues/7787.

headius:
Appears to also fix #6119.

headius:
My searching did not find any other issues, but I did not audit all tagged tests/specs (it's a hard thing to search for when some of the failures are due to methods being missing on singleton classes!) and I still feel like there have been other issues related to incorrect singleton class hierarchies.

Nevertheless I think we'll proceed with this since it fixes at least two known issues, one of which was silently closed long ago.

PR #9447, opened 2026-05-20
Improve alias redefinition warnings

headius:
These warnings are firing a lot in Rails when they should not. The key piece is setting the aliased method's aliased bit so aliasing back doesn't warn, based on the hack in https://github.com/rails/rails/pull/29233. This PR makes tweaks to handle that case and clean up related code.


headius:
Example case:

```ruby
class Foo
  def foo end
  alias_method :bar, :foo
  define_method(:foo) {}
  alias_method :foo, :foo
  alias_method :foo, :bar
end
```

The self-aliasing here adds an alias bit to the `foo` method so that the subsequent `alias_method` call does not warn.

PR #9448, opened 2026-05-20
Use implementation class for protected check

sampokuokkanen:
Replace metaClass.getRealClass().isInstance(self) with method.getImplementationClass().isInstance(self) in Helpers.java so the protected-method visibility check inspects the method's implementation class and not the receiver's class.

Removes `test_defined_protected_method` exclusion from `test/mri/excludes/TestDefined.rb`.

### Reproduction

```ruby
class Foo
  def foo; end
  protected :foo

  def bar(f)
    defined?(f.foo)
  end
end
class Sub < Foo; end

p Foo.new.bar(Foo.new) # MRI: "method", JRuby master: "method"
p Foo.new.bar(Sub.new) # MRI: "method", JRuby master: nil ← bug
p Foo.new.bar(Class.new(Foo).new) # MRI: "method", JRuby master: nil ← same bug
```


sampokuokkanen:
Interestingly, while working on this I noticed an MRI bug which I fixed here: https://github.com/ruby/ruby/pull/17069

```ruby
module Mix
  def secret; 42; end
  protected :secret
end

class A; include Mix; end

class B
  include Mix
  def call_it(other)    = other.secret
  def defined_it(other) = defined?(other.secret)
end

p B.new.call_it(A.new)      # => 42
p B.new.defined_it(A.new)   # => nil on MRI before my fix was merged
```

JRuby on this branch passes the new MRI tests I added in the above PR.

headius (approved):
Yup, makes sense!

PR #9460, opened 2026-05-25
Check argument definedness in defined?(receiver.method(arg))

sampokuokkanen:
Mirrors the FCALLNODE path. Fixes test_defined_undefined_argument.

Sample code of the problem: 

```ruby
class Test
  def hello x
    puts "Hello #{x}"
  end
end

p defined? Test.new.hello(hello)
```
Before this fix: `"method"`, now `nil`.


headius:
Looks good, thank you!

PR #9461, opened 2026-05-25
[ji] support java.lang.Throwable#backtrace_locations

### LINKED ISSUE #9458: Support backtrace_locations with Java exceptions
JRuby had supported duck-typing of `java.lang.Throwable` hierarchy, to act like Ruby's `Exception`, for a long time.

There's `java.lang.Throwable#backtrace` which maps the Java stack-trace into an Array. 

The `backtrace_locations` method are a later Ruby language feature and so wasn't considered for addiiton. 
It would be desired to support `backtrace_location` with JRuby's Java integration as well, esp. since frameworks such as Rails might do `rescue Exception` (which catched Java throwables) and assume `backtrace_locations` exists.

kares:
the one thing left, in terms of Ruby exception compatibility, would be `Exception#exception(str)` which might be a bit tricky to fully support.

### DIFF
--- a/core/src/main/java/org/jruby/RubyThread.java
+++ b/core/src/main/java/org/jruby/RubyThread.java
@@ -567,9 +567,10 @@ public class RubyThread extends RubyObject implements ExecutionContext {
         public static RubyArray newLocationArray(Ruby runtime, RubyStackTraceElement[] elements,
             final int offset, final int length) {
 
+            final RubyClass locationClass = runtime.getLocation();
             IRubyObject[] ary = new IRubyObject[length];
             for ( int i = 0; i < length; i++ ) {
-                ary[i] = newLocation(runtime, elements[i + offset]);
+                ary[i] = new Location(runtime, locationClass, elements[i + offset]);
             }
 
             return RubyArray.newArrayNoCopy(runtime, ary);
--- a/core/src/main/java/org/jruby/javasupport/ext/JavaLang.java
+++ b/core/src/main/java/org/jruby/javasupport/ext/JavaLang.java
@@ -44,6 +44,7 @@ import org.jruby.runtime.Helpers;
 import org.jruby.runtime.JavaInternalBlockBody;
 import org.jruby.runtime.Signature;
 import org.jruby.runtime.ThreadContext;
+import org.jruby.runtime.backtrace.RubyStackTraceElement;
 import org.jruby.runtime.backtrace.TraceType;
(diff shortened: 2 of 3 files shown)


kares:
Based on Rails usage, as mentioned in https://github.com/jruby/jruby/issues/9458 would likely make sense to re-target this to 10.0.x

core/src/main/java/org/jruby/javasupport/ext/JavaLang.java
```diff
@@ -235,6 +235,20 @@ public static IRubyObject backtrace(final ThreadContext context, final IRubyObje
             return RubyArray.newArrayMayCopy(runtime, backtrace);
         }
 
+        @JRubyMethod
+        public static IRubyObject backtrace_locations(final ThreadContext context, final IRubyObject self) {
```
headius, line 239: This and `backtrace` are very similar. Maybe share the common parts and just construct the elements differently.

core/src/main/java/org/jruby/RubyThread.java
```diff
@@ -579,9 +579,10 @@ public static RubyArray newLocationArray(Ruby runtime, RubyStackTraceElement[] e
         public static RubyArray newLocationArray(Ruby runtime, RubyStackTraceElement[] elements,
             final int offset, final int length) {
 
+            final RubyClass locationClass = runtime.getLocation();
             IRubyObject[] ary = new IRubyObject[length];
             for ( int i = 0; i < length; i++ ) {
-                ary[i] = newLocation(runtime, elements[i + offset]);
+                ary[i] = new Location(runtime, locationClass, elements[i + offset]);
```
headius, line 573: Would a new overload of `newLocation` make sense here?

headius (commented):
Minor suggestions to clean it up a bit but generally looks fine. In the future we might want to also use our JRuby-aware stack trace processing to rewrite JIT and interpreter frames but that's out of scope here.

core/src/main/java/org/jruby/RubyThread.java
kares: not sure, literally would be the same as doing `new Location()` directly.

core/src/main/java/org/jruby/javasupport/ext/JavaLang.java
kares: yeah the preamble, but even if we dry out the loop mapping with a function you still need to return a different array.

PR #9463, opened 2026-05-26
attempt to suppress alias warning ~ MRI

### LINKED ISSUE #9450: Reverse aliasing warns when it should not
A form of reverse aliasing (returning a method name back to its original target method after previously aliasing it) found in Rails 8.0's redefine_method.rb produces warnings when apparently it should not.

```
/Users/headius/work/rails/activesupport/lib/active_support/testing/time_helpers.rb:62: warning: method redefined; discarding old new
/Users/headius/work/rails/activesupport/lib/active_support/testing/time_helpers.rb:181: warning: previous definition of new was here
/Users/headius/work/rails/activesupport/lib/active_support/testing/time_helpers.rb:62: warning: method redefined; discarding old today
/Users/headius/work/rails/activesupport/lib/active_support/testing/time_helpers.rb:190: warning: previous definition of today was here
/Users/headius/work/rails/activesupport/lib/active_support/testing/time_helpers.rb:62: warning: method redefined; discarding old now
```

This seems to have something to do with a hack from https://github.com/rails/rails/pull/29233 that re-aliases the same alias so it can go back without a warning. However this does not appear to work properly in JRuby.

I attempted to fix this in #9447 but it has not yet solved the issue.

Can be reproduced by running the Rails ActiveRecord suite for ARJDBC as described in #9449.

kares:
these boolean fields could maybe just end up as additional flags, wasn't sure if that would be preferable

### DIFF
--- a/core/src/main/java/org/jruby/RubyModule.java
+++ b/core/src/main/java/org/jruby/RubyModule.java
@@ -5702,15 +5702,22 @@ public class RubyModule extends RubyObject {
 
         DynamicMethod method = getMethods().get(name);
         if (method != null && entry.method.getRealMethod() != method.getRealMethod() && !method.isUndefined()) {
-            if (method.getRealMethod().getAliasCount() == 0) warning(context, "method redefined; discarding old " + name);
+            if (method.getRealMethod().getAliasCount() == 0) {
+                warning(context, "method redefined; discarding old " + name);
 
-            if (method instanceof PositionAware posAware) {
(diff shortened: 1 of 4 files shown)


headius:
I also considered putting these in a flag, but removing the boolean field doesn't seem to have much impact on object size. It would be good to reevaluate all of the fields we have in DynamicMethod to see if we can shrink those, because a typical app has thousands of them.

kares:
closing due https://github.com/jruby/jruby/pull/9447/ would be nice to get the changes back-ported as the warnings are a bit annoying on Rails 8.x

PR #9466, opened 2026-05-28
Fix Enumerator::Lazy value packing of source yields

sampokuokkanen:
Lazy.make_proc was an arity-faithful passthrough (`yield yielder, *x`) instead of applying CRuby's rb_enum_values_pack rule (enum.c): 0 args -> nil, 1 arg -> the value, N args -> Array. Pipeline stages downstream expect the packed form, so a zero-argument source yield surfaced as `[]` and a multi-argument yield surfaced as separate block args, both diverging from MRI.

Specced via Enumerable#take / Enumerator::Lazy#take shared examples in https://github.com/ruby/spec/pull/1363.

### DIFF
--- a/core/src/main/ruby/jruby/kernel/enumerator.rb
+++ b/core/src/main/ruby/jruby/kernel/enumerator.rb
@@ -146,7 +146,9 @@ class Enumerator
       proc do |yielder, *args|
         catch yielder do
           obj.each(*args) do |*x|
-            yield yielder, *x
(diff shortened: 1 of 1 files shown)


sampokuokkanen:
This needs a bigger Java-based refactor with something like LazyYielder. Will turn the PR into a draft for now.

sampokuokkanen:
Wrong place for the fix so closing this for now.

PR #9470, opened 2026-06-02
Clean up more divergent keyword logic

### LINKED ISSUE #9462: Regression in JRuby 10.1 with Prawn
**Environment Information**

This affects:

```
jruby 10.1.0.0 (4.0.0) 2026-04-20 32f988b78c OpenJDK 64-Bit Server VM 21.0.10+7-1 on 21.0.10+7-1 +indy +jit [x86_64-OpenBSD]
OpenBSD 7.9 amd64
```

as well as an Ubuntu 24.04 GitHub Runner (https://github.com/jeremyevans/tilt/actions/runs/26411921259/job/77748035355#step:5:563)

The gem showing incorrect behavior with JRuby 10.1 is Prawn.

**Expected Behavior**

Prawn on JRuby 10.1 should work as it does on JRuby 10.0 or CRuby 4.0.

**Actual Behavior**

Prawn on JRuby 10.1 produces incorrect PDFs after the 10th iteration. This doesn't happen on JRuby 10.0 or CRuby 4.0. Example tilt CI run: https://github.com/jeremyevans/tilt/actions/runs/26411921259/job/77748035355#step:5:563

While the exception occurs in pdf-reader, the actual problem is in Prawn, and is shown by this code:

```ruby
require 'prawn'

11.times do
  pdf = ::Prawn::Document.new
  pdf.text "Hello PDF!"
  p pdf.render.to_str.bytesize
end
```

For the first 10 iterations, you get `911`. For the last iteration, you get `899`. The difference in the PDF produced in the 11th iteration:

```diff
--- good.pdf    Mon May 25 11:36:29 2026
+++ bad.pdf     Mon May 25 11:36:29 2026
@@ -39,8 +39,7 @@
 /CropBox [0 0 612 792]
 /MediaBox [0 0 612 792]
 /Parent 3 0 R
-/Resources << /Font << /F1.0 6 0 R
->>
+/Resources << /Font << >>
 /ProcSet [/PDF /Text /ImageB /ImageC /ImageI]
 >>
 /TrimBox [0 0 612 792]
@@ -62,12 +61,12 @@
 0000000158 00000 n
 0000000215 00000 n
 0000000333 00000 n
-0000000599 00000 n
+0000000587 00000 n
 trailer
 << /Info 1 0 R
 /Root 2 0 R
 /Size 7
 >>
 startxref
-696
+684
 %%EOF
```

I worked around this in tilt by modifying the tests to not run more than 10 iterations, and that made CI pass: https://github.com/jeremyevans/tilt/actions/runs/26414918691

Maybe there is some internal misoptimization that runs after 10 iterations?

headius:
There's still enough divergent keyword argument logic to cause #9462 due to a JIT path failing to check for empty kwargs. That fix and others will be part of this PR to unify more of the two sides.

### DIFF
--- a/core/src/main/java/org/jruby/ir/runtime/IRRuntimeHelpers.java
+++ b/core/src/main/java/org/jruby/ir/runtime/IRRuntimeHelpers.java
@@ -721,7 +721,8 @@ public class IRRuntimeHelpers {
         return last;
     }
 
-    public static KwargsAction kwargsActionJIT(IRubyObject last, boolean ruby2Keywords, int callInfo) {
+    @JIT
+    private static KwargsAction kwargsActionJIT(IRubyObject last, boolean ruby2Keywords, int callInfo) {
         boolean isKwarg = hasKeywords(callInfo);
         if (ruby2Keywords) {
             // ruby2_keywords only get unmarked if it enters a method which accepts keywords.
@@ -792,6 +793,7 @@ public class IRRuntimeHelpers {
      * @param callInfo the callInfo metadata for the call
      * @return true if keywords should be processed, false otherwise
      */
+    @Interp
     private static boolean shouldHandleKwargs(IRubyObject[] args, int callInfo) {
         if (args.length > 0) {
             return shouldHandleKwargs(args[args.length - 1], callInfo);
@@ -809,6 +811,7 @@ public class IRRuntimeHelpers {
      * @param callInfo the callInfo metadata for the call
      * @return true if keywords should be processed, false otherwise
      */
+    @JIT
     private static boolean shouldHandleKwargs(IRubyObject lastArg, int callInfo) {
         return !keywordsEmpty(callInfo) && lastArg instanceof RubyHash;
     }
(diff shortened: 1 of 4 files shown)


headius:
The first fix here is valid, but in the original investigation of #9462 I found that disabling JIT for the *target* method fixed the issue. I believe there's more happening here.

The example given in that issue fails, but this non-delegated version does not. It's possible that delegation is setting more keyword flags in `callInfo` than it should. Logically, it seems as though it should pass the flags through exactly as they came in, but that may not be exactly correct given how it captures and restructures the argument list into an IRubyObject[].

headius:
I've filed #9472 to look into reworking how argument forwarding is done. I believe it was implemented to decompose and recompose the argument list as a convenience but this introduces a lot of overhead and potentially alters the structure of the arguments or their keyword metadata.

This will probably come along with work on #9219 to make all dispatch paths fully specified and adaptable.

headius:
I did not find much else to clean up that wouldn't be better waiting until after a rework of how keyword arguments are passed.

PR #9475, opened 2026-06-04
Backport extended dist verification to 10.0

headius:
This backports the release verification workflow from master to 10.0 along with installer verification.


headius:
Closing to push as a local branch, so installers will be tested.

PR #9477, opened 2026-06-04
Dist verification 9.4

headius:
This backports release artifact verification to 9.4.


headius:
A few diffs from 10.0+ to work properly on 9.4:

* Use Java 8 instead of 21
* Use Install4J 9 since 12 does not work with Java 8

headius:
Additional tweak:

* Use zulu JDK instead of temurin (temurin no longer has builds of 8 for macos)

PR #9481, opened 2026-06-09
Implement Proc#with_refinements

shugo:
Add Proc#with_refinements(mod1, ...), which returns a new Proc whose body runs with the given modules' refinements active. The receiver is unchanged and the captured closure environment is shared; only the refinement scope differs.

In JRuby whether a call resolves refinements is decided at IR build time: CallInstr.create reads scope.maybeUsingRefinements() and, when true, builds a RefinedCachingCallSite that captures the static scope. So a refinement-aware copy of the receiver proc's IR is required (the analog of CRuby recursively deep-copying the block iseq); merely swapping an overlay at runtime is not enough.

Implementation:
- RubyProc#with_refinements validates args (>=1, all Modules), rejects non-IR-block / method-backed procs, then deep-copies the block's IRClosure and wraps it in a new Block that shares a clone of the original binding.
- IRClosure#cloneForRefinements clones the closure tree via cloneForInlining with a new refinementsClone flag and activates the requested refinements on the top clone's overlay module.
- CloneInfo/SimpleCloneInfo gain isRefinementsClone(); when set each clone gets its own duplicated StaticScope (enclosing re-pointed to the parent clone) and maybeUsingRefinements, WrappedIRClosure deep-clones nested closures, and every call-bearing instruction's clone() rebuilds as a refined call site bound to the clone's scope. Fixnum/Float fast-path calls downgrade to generic refined calls. New IRScope/IRClosure copy-constructor variants accept a StaticScope since IRScope.staticScope is final.

When isRefinementsClone() is false (all other clone paths) behavior is unchanged. Adds test/jruby/test_proc_with_refinements.rb.

### DIFF
--- a/core/src/main/java/org/jruby/ir/IRClosure.java
+++ b/core/src/main/java/org/jruby/ir/IRClosure.java
@@ -5,6 +5,7 @@ import java.util.EnumSet;
 import java.util.List;
 import java.util.function.Supplier;
 
+import org.jruby.RubyModule;
 import org.jruby.RubySymbol;
 import org.jruby.ast.DefNode;
 import org.jruby.ast.IterNode;
@@ -26,6 +27,7 @@ import org.jruby.runtime.IRBlockBody;
 import org.jruby.runtime.MixedModeIRBlockBody;
 import org.jruby.runtime.InterpretedIRBlockBody;
 import org.jruby.runtime.Signature;
+import org.jruby.runtime.ThreadContext;
 import org.jruby.util.ByteList;
 
 // Closures are contexts/scopes for the purpose of IR building.  They are self-contained and accumulate instructions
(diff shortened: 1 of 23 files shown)


shugo:
I accidentally opened the pull request in the wrong repository. Sorry.

headius:
Looks pretty good at a glance. 😀

shugo:
> Looks pretty good at a glance. 😀

Thank you!

I've created an issue at https://bugs.ruby-lang.org/issues/22097

core/src/main/java/org/jruby/ir/IRClosure.java
```diff
@@ -399,7 +445,7 @@ public IRClosure cloneForInlining(CloneInfo ii) {
      * @param modules the refinement modules to activate
      * @return a new IRClosure whose body runs with the given refinements active
      */
-    public IRClosure cloneForRefinements(ThreadContext context, List<RubyModule> modules) {
+    public IRClosure cloneForRefinements(ThreadContext context, RubyModule[] modules) {
```
enebo, line 448: This changes a public API but I am not sure if we actually care as these are only public from a Java perspective.  We could easily add a deprecation if this is an issue later.

enebo: @shugo sorry I missed you have closed this and moved to PR on your fork.  The comment still applies there if it switched from List<RubyModule> to RubyModule[].

shugo: @enebo 
Thanks for your comment.  cloneForRefinements() was introduced in this feature branch and has not been released yet, so I believe there is no compatibility issue with this change.

PR #9485, opened 2026-06-09
ASM update

headius:
Update to latest ObjectWeb ASM library.

This brings in support for JVM bytecode versions newer than Java 23.


headius:
Need to upgrade in jnr-* as well so this is bumped to 10.0.7.0.

headius:
Already done other than the test update, which I'll do separately.

headius:
See #9601 for the update.

PR #9486, opened 2026-06-10
Implement Proc#refined

shugo:
Add Proc#refined(mod1, ...), which returns a new Proc whose body runs
with the given modules' refinements active. The receiver is unchanged
and the captured closure environment is shared; only the refinement
scope differs.

In JRuby whether a call resolves refinements is decided at IR build
time: CallInstr.create reads scope.maybeUsingRefinements() and, when
true, builds a RefinedCachingCallSite that captures the static scope.
So a refinement-aware copy of the receiver proc's IR is required (the
analog of CRuby recursively deep-copying the block iseq); merely
swapping an overlay at runtime is not enough.

## Cost for code that never uses Proc#refined

Branch `proc-with-refinements-ruby-4.1` (feature commit `12c2bdea1c`), baseline = parent
commit `e66555aa43`.

### Memory

- **`RubyProc` / `IRScope`**: no new fields — zero overhead per proc or per scope.
- **`IRClosure`**: two `boolean` flags (`inRefinementsCloneTree`, `refinementsClone`).
  Verified with JOL 0.17 that they fall into existing alignment padding after
  `closureId`/`isEND`: instance size is unchanged with or without them
  (120 bytes with compressed oops, 192 bytes without) — exactly zero bytes per closure.

  ```
   96   4  int     IRClosure.closureId
  100   1  boolean IRClosure.isEND
  101   1  boolean IRClosure.inRefinementsCloneTree   <- added
  102   1  boolean IRClosure.refinementsClone         <- added
  103   1          (alignment/padding gap)
  104   4  ref     IRClosure.signature
  Instance size: 120 bytes
  ```

- **`IRManager`**: one empty memo map (`refinementsCloneCache`, a synchronized
  `IdentityHashMap`) per runtime, ~300 bytes. This is the only fixed cost for a runtime
  that never uses `Proc#refined`.
- The map is intentionally strong- and identity-keyed:
  - Weak keys could never clear because the memo value (the clone) reaches its own key
    through the shared lexical parent (tried and reverted in presquash history:
    `c9c955434a` -> `fd06ace73b`). CRuby's iseq-keyed memo has the same clone->source reach.
  - Identity keying avoids `IRScope.equals`' scopeId-wraparound collisions (`3b259619f5`).
  - Entries are bounded to one slot per source closure, so growth is bounded by the number
    of distinct closures ever passed to `refined`.
- The map is read only inside `Proc#refined` itself (`IRClosure#refinementsClone`); no other
  runtime path touches it.

### Speed

Methodology: microbenchmarks per touched path; each sample = min of 10 timed passes after
5 warmup passes in a fresh process; binaries alternated per round, and which side runs
first alternates per round to cancel drift. Values are medians of per-process samples.
Machine noise floor: identical-binary samples varied up to 13-19% (e.g. base fixnum_ops
0.538-0.609). Startup is measured on the bare jar (`java -Djruby.home=... -jar jruby.jar
-e 0`); comparing via the `bin/jruby` launchers was confounded by unrelated differences in
the two working trees' state.

| bench       | path exercised                            | n | base   | feature | delta  |
|-------------|-------------------------------------------|---|--------|---------|--------|
| yield       | block dispatch                            | 8 | 0.2956 | 0.3041  | +2.9%  |
| proc_call   | Proc#call                                 | 5 | 0.2279 | 0.2242  | -1.6%  |
| proc_new    | Block/BlockBody instantiation             | 5 | 0.1193 | 0.1201  | +0.7%  |
| sym_to_proc | baked `&:sym` (SymbolProc)                | 5 | 0.0329 | 0.0322  | -2.1%  |
| fixnum_ops  | specialized `+`/`<` call instrs           | 5 | 0.5508 | 0.5447  | -1.1%  |
| float_ops   | specialized float `+` call instr          | 5 | 0.4527 | 0.4435  | -2.0%  |
| aref        | ArrayDerefInstr (was modified)            | 8 | 0.5204 | 0.5028  | -3.4%  |
| ir_build    | parse + IR build (incl. ensure cloning)   | 5 | 0.0109 | 0.0089  | -18%*  |
| startup     | bare-jar `-e 0` wall time                 | 8 | 1.525  | 1.495   | -2.0%  |

\* millisecond-scale absolute values; dominated by noise.

Verdict: no measurable slowdown. Deltas scatter in both directions within the noise floor,
and the leaning benches are not stable across sessions: aref was +2.4% in an earlier run
and -3.4% here; yield's per-round pairs split 3 feature wins / 5 losses (median +2.0%).
Mechanistically, the changes on these paths (`ArrayDerefInstr`, `CallInstr`, specialized
instrs, `Define*Instr`) are confined to `clone()`/`create()` — build time only; no
runtime-executed code or runtime-read field changed.

Bench scripts and raw results are in: https://github.com/shugo/jruby-refined-bench

### DIFF
--- a/core/src/main/java/org/jruby/RubyModule.java
+++ b/core/src/main/java/org/jruby/RubyModule.java
@@ -1134,6 +1134,9 @@ public class RubyModule extends RubyObject {
         if (context.getCurrentStaticScope().isWithinMethod()) {
             throw runtimeError(context, "Module#using is not permitted in methods");
         }
+        if (context.getCurrentStaticScope().isWithinRefinementsProc()) {
+            throw runtimeError(context, "Module#using is not permitted in a proc with refinements");
+        }
 
         // I pass the cref even though I don't need to so that the concept is simpler to read
         StaticScope staticScope = context.getCurrentStaticScope();
@@ -1164,7 +1167,12 @@ public class RubyModule extends RubyObject {
         RefinementStore refinementStore = module.refinementStore;
         if (refinementStore == null) return; // No refinements registered for this module
 
-        Map<RubyModule, RubyModule> refinements = refinementStore.refinements;
+        // Snapshot under the map's own lock: Proc#refined activates refinements from arbitrary threads
+        // while Module#refine on another thread may be mutating this map.
+        Map<RubyModule, RubyModule> refinements;
+        synchronized (refinementStore.refinements) {
+            refinements = new IdentityHashMap<>(refinementStore.refinements);
+        }
         for (Map.Entry<RubyModule, RubyModule> entry: refinements.entrySet()) {
             usingRefinement(context, cref, entry.getKey(), entry.getValue());
         }
@@ -2938,9 +2946,21 @@ public class RubyModule extends RubyObject {
         return context.getCurrentFrame().getSelf() == this ? context.getCurrentVisibility() : PUBLIC;
     }
 
+    private static void checkNotRefinementsProc(ThreadContext context, Block block) {
+        // A method defined from a Proc#refined proc would silently drop its refinements (a method is invoked
+        // against its method entry, not the proc's refinement scope), so reject it.
+        if (block.getBody() instanceof IRBlockBody body) {
+            IRClosure closure = body.getScope();
+            if (closure != null && closure.isRefinementsClone()) {
+                throw argumentError(context, "can't define a method from a Proc with refinements");
+            }
+        }
+    }
+
     public IRubyObject defineMethodFromBlock(ThreadContext context, IRubyObject arg0, Block block, Visibility visibility) {
         RubySymbol name = checkID(context, arg0);
         if (!block.isGiven()) throw argumentError(context, "tried to create Proc object without a block");
+        checkNotRefinementsProc(context, block);
 
         if ("initialize".equals(name.idString())) visibility = PRIVATE;
 
@@ -2984,6 +3004,7 @@ public class RubyModule extends RubyObject {
         if (context.runtime.getProc().isInstance(arg1)) {
             // double-testing args.length here, but it avoids duplicating the proc-setup code in two places
             RubyProc proc = (RubyProc)arg1;
+            checkNotRefinementsProc(context, proc.getBlock());
 
             newMethod = createProcMethod(context.runtime, name.idString(), visibility, proc.getBlock());
         } else if (arg1 instanceof AbstractRubyMethod) {
--- a/core/src/main/java/org/jruby/RubyProc.java
+++ b/core/src/main/java/org/jruby/RubyProc.java
@@ -39,6 +39,8 @@ import org.jruby.anno.JRubyClass;
 import org.jruby.anno.JRubyMethod;
 
 import org.jruby.ast.util.ArgsUtil;
+import org.jruby.ir.IRClosure;
+import org.jruby.ir.interpreter.Interpreter;
 import org.jruby.ir.runtime.IRRuntimeHelpers;
 import org.jruby.java.codegen.BlockInterfaceGenerator;
 import org.jruby.javasupport.Java;
@@ -58,6 +60,7 @@ import org.jruby.runtime.marshal.DataType;
 
 import static org.jruby.api.Convert.asBoolean;
 import static org.jruby.api.Convert.asFixnum;
+import static org.jruby.api.Convert.castAsModule;
 import static org.jruby.api.Create.*;
 import static org.jruby.api.Define.defineClass;
 import static org.jruby.api.Error.argumentError;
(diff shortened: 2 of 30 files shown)


headius:
@shugo Thank you for iterating on this! Let me know when you consider this ready to merge.

shugo:
@headius 
Thanks! I've finished the changes I had in mind, so from my side this is ready.
I'd appreciate a review when you have time.

Two points I'd especially like your opinion on:

1. The memo moved from IRClosure to an external table.
    Following ko1's review of the CRuby implementation, CRuby moved the memo of refined proc copies off the iseq into an external table, so I matched that here: the single-slot memo now lives in a per-runtime map on `IRManager` (`refinementsCloneCache`) instead of a field on `IRClosure`. Like CRuby's table it is not weakly keyed — weak keys would never clear anyway, because the memoized clone reaches its source closure through the shared lexical parent. In realistic use cases I expect it to be rare for the closure behind a refined block to become garbage anyway.

2. The `clone()` restructuring of the CallInstr specializations.
   The specialized subclasses (`OneFixnumArgNoBlockCallInstr` etc.) no longer override `clone()`; `CallInstr.clone()` now goes through `create()`, which re-picks the specialization from the cloned operands and the effective refinement flag, so a clone into a refined scope drops the fixnum/float/aref fast paths and gets a refined call site instead. I'd like to be sure this doesn't break any assumptions in the other cloning paths (inlining, ensure-region cloning).

shugo:
I've added Co-Authored-By: to the commit message and have force-pushed, but no code is changed.

headius:
> Like CRuby's table it is not weakly keyed

Of course this immediately worries me, but there's plenty of other non-weak IR and structural data being stored. I'll look at the code and make an assessment.

> The clone() restructuring of the CallInstr specializations.

Instruction cloning is largely related to inlining, which we have never fully hooked up. Maybe @enebo can offer some thoughts here about the restructuring, but it sounds ok to me.

headius (approved):
I did not see @ko1's reason for moving the cache off of the closure and into a global cache but I think it's fine to merge as-is for now. I'd like to understand why the global location was preferred in CRuby.

The other changes look ok to me and remind me that we also use cloning for the optimization passes. I believe what you have here does not break any expectations in those subclasses.

We need to get a performance CI set up so we can catch perf regressions, but I appreciate that you verified some related performance with this PR.

headius:
@shugo Thank you for your efforts here, this is a great help to us in keeping up with 4.1 changes! If you created any LLM notes/markdown/whatever that might help future contributors, perhaps you can help us find a good place to put those in the repository.

**The only thing I'd like to see adjusted** would be to move the tests into ruby/spec so they're not in the JRuby-specific suite. I'll update our documentation everywhere to make this recommendation more explicit for LLM users. I've merged as-is for now, but if you can turn these into specs I think it would be best for everyone.

shugo:
@headius Thanks for merging this!

> I did not see @ko1's reason for moving the cache off of the closure and into a global cache but I think it's fine to merge as-is for now. I'd like to understand why the global location was preferred in CRuby.

From what I understand, he didn't want an iseq to carry such a cache, because an iseq is (ideally) an immutable, Ractor-shareable value.

> @shugo Thank you for your efforts here, this is a great help to us in keeping up with 4.1 changes! If you created any LLM notes/markdown/whatever that might help future contributors, perhaps you can help us find a good place to put those in the repository.

My CLAUDE.md is pretty simple and may not be so useful: https://gist.github.com/shugo/11ccb81abd7e571f0e8c16182cb8e419

If you'd like, I'll create a pull request removing the parts specific to my environment.

> **The only thing I'd like to see adjusted** would be to move the tests into ruby/spec so they're not in the JRuby-specific suite. I'll update our documentation everywhere to make this recommendation more explicit for LLM users. I've merged as-is for now, but if you can turn these into specs I think it would be best for everyone.

I've added specs to ruby/ruby (https://github.com/ruby/ruby/pull/17964), so they will be synced to ruby/spec, but I forgot to reduce the tests in test/jruby/test_proc_refined.rb.  Sorry.
I've created another pull request: https://github.com/jruby/jruby/pull/9535

Note that RUBY_VERSION is still 4.0.0 on the ruby-4.1 branch, so specs guarded by `ruby_version_is "4.1"` are skipped until it is bumped.  Could you bump it at an appropriate time?

headius:
> From what I understand, he didn't want an iseq to carry such a cache, because an iseq is (ideally) an immutable, Ractor-shareable value.

I understand. 

> If you'd like, I'll create a pull request removing the parts specific to my environment.

That's OK, I can learn from this just fine. 

> I've created another pull request

I saw, thank you!

> Note that RUBY_VERSION is still 4.0.0 on the ruby-4.1 branch,

I have another pr that also updates the version and will merge that soon. 

Thanks again!

PR #9499, opened 2026-06-18
Use Integer for JRuby VM messages

### LINKED ISSUE #9498: jruby/vm integer messages fail on JRuby 10 due to Fixnum reference
`jruby/vm` still references `Fixnum` in `JRuby::VM#send`:

```ruby
def send(obj)
  case obj
  when String
    @queue.put obj
  when Fixnum
    @queue.put obj
  end
end
```

On JRuby 10 / Ruby 3.4 semantics, `Fixnum` is undefined, so sending integer messages fails.

Minimal repro:

```sh
jruby -rjruby/vm -e 'JRuby::VM.send_message(JRuby::VM_ID, 1)'
```

Actual:

```text
NameError: uninitialized constant JRuby::VM::Fixnum
```

This also breaks the documented/sample sub-VM pattern:

```ruby
vm << JRuby::VM_ID
```

because `JRuby::VM_ID` is an integer.

Expected: integer messages should work.

Likely fix:

```diff
- when Fixnum
+ when Integer
```

lloeki:
`JRuby::VM#send` checks `Fixnum` when accepting integer messages. On JRuby 10 / Ruby 3.4 semantics, `Fixnum` is undefined, so sending integer messages raises `NameError` before enqueuing.

Use `Integer` instead. This keeps integer messages working on JRuby 10 and remains compatible with older supported JRuby versions.

This fixes the sub-VM sample pattern:

```ruby
vm << JRuby::VM_ID
```

Fixes #9498.

## Testing

- JRuby 10:

```sh
docker run --rm \
  -v /home/ll/src/github.com/jruby/jruby:/workspace \
  -w /workspace \
  ghcr.io/datadog/images-rb/engines/jruby:10.0 \
  jruby -I /workspace/lib/ruby/stdlib -rjruby/vm \
    -e "JRuby::VM.send_message(JRuby::VM_ID, 1); puts JRuby::VM.get_message"
```

- JRuby 9.4:

```sh
docker run --rm \
  -v /home/ll/src/github.com/jruby/jruby:/workspace \
  -w /workspace \
  ghcr.io/datadog/images-rb/engines/jruby:9.4 \
  jruby -I /workspace/lib/ruby/stdlib -rjruby/vm \
    -e "JRuby::VM.send_message(JRuby::VM_ID, 1); puts JRuby::VM.get_message"
```

### DIFF
--- a/lib/ruby/stdlib/jruby/vm.rb
+++ b/lib/ruby/stdlib/jruby/vm.rb
@@ -69,7 +69,7 @@ module JRuby
       case obj
       when String
         @queue.put obj
-      when Fixnum
+      when Integer
         @queue.put obj
       end
     end
--- a/test/jruby.index
+++ b/test/jruby.index
@@ -117,3 +117,4 @@ jruby/test_load_compiled_ruby
 jruby/test_load_compiled_ruby_class_from_classpath
 
 jruby/test_date_parse_limit
+jruby/test_jruby_vm
--- a/test/jruby/test_jruby_vm.rb
+++ b/test/jruby/test_jruby_vm.rb
@@ -1,0 +1,70 @@
+require 'test/unit'
+require 'jruby/vm'
+
+class TestJRubyVM < Test::Unit::TestCase
+  MESSAGE_TIMEOUT = 5
+  THREAD_JOIN_TIMEOUT_MS = 5_000
+
+  def setup
+    drain_messages
+  end
+
+  def teardown
+    drain_messages
+  end
+
+  def test_send_message_accepts_string_and_integer
+    ['message', 1].each do |message|
+      JRuby::VM.send_message(JRuby::VM_ID, message)
+
+      assert_equal message, wait_for_message
+    end
+  end
+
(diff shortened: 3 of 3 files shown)


lloeki:
I could not find tests for sub-vm machinery, are these desirable to add in this PR?

If so I would appreciate guidance as to the best spot/approach to follow.

headius:
@lloeki Tests would be very welcome! Clearly this needs more coverage. This utility library was largely created as a one-off experiment but I'd like to formalize it better with tests and documentation.

Since this does not seem like a time-critical fix for you perhaps we can add a few tests here and merge before the next release of 10.0.x?

lloeki:
@headius I've added some tests.

Note that there's a code path so that the test doesn't block when run against a JRuby version that doesn't carry the fix. LMK if you'd rather have that removed.

headius:
@lloeki Yes, please remove it. We expect new tests to only be run on fixed JRuby versions going forward.

headius (approved):

Issue #9472, opened 2026-06-02
Argument forwarding does too much work


headius:
The current implementation of argument forwarding still works by decomposing the incoming argument list into a set of argument groups and then re-composing those groups back into a varargs-like call. This introduces significant overhead, due to the extra processing overhead and creation of argument lists, but it also seems to be introducing some variability into argument forwarding that can impact behavior.

An example case is from #9462:

```
def foo(...)
  bar(...)
end
def bar(a)
  p a.object_id
end
x = {a: 1}
p x.object_id
foo(x)
```

This case failed when JIT compiled because the `x` hash gets duplicated before the innermost `object_id` call. But the duplication is happening in the `bar` method, not as part of the delegated call.

I believe this is because the process of decomposing and recomposing the argument list during delegation is altering the structure of the arguments or the associated `callInfo` flags in such a way that triggered the JIT bug. That JIT bug will be fixed by #9470, but it is a symptom of the delegation not being a "pure" pass-through.

With the old AST, a delegated method gets parsed as follows:

```
$ jruby -S ast -e 'def foo(...); bar(...); end' 
AST:
RootNode line: 0
  DefnNode*:foo line: 0
    ArgsNode line: 0
      ListNode line: 0
      ListNode line: 0
      UnnamedRestArgNode:* line: 0
      ListNode line: 0
      ListNode line: 0
      KeywordRestArgNode:** line: 0
      ForwardingBlockArgNode:& line: 0
    FCallNode*:bar line: 0
      ArgsPushNode line: 0
        SplatNode line: 0
          LocalVarNode:* line: 0
        HashNode line: 0, onlykwrest, null
          LocalVarNode:** line: 0
      BlockPassNode line: 0
        ArgsPushNode line: 0
          SplatNode line: 0
            LocalVarNode:* line: 0
          HashNode line: 0, onlykwrest, null
            LocalVarNode:** line: 0
        LocalVarNode:& line: 0
```

Notice the first thing the method does is process all possible forms of arguments, which then get re-assembled for the delegated call.

The Prism parser provides forwarding as a special AST node:

```
$ jruby -Xparser.prism -S ast -e 'def foo(...); bar(...); end'
AST:ProgramNode
  locals: 
  statements: StatementsNode
    body: 
      DefNode[Li]
        name: "foo"
        receiver: null
        parameters: ParametersNode
          requireds: 
          optionals: 
          rest: null
          posts: 
          keywords: 
          keyword_rest: ForwardingParameterNode
[...]

PR #9501, opened 2026-06-21
Fix CALL_KEYWORD_EMPTY leaking from (...) forwards to native methods

segiddins:
## Problem

A literal keyword argument is silently dropped — passed as a trailing positional Hash — when the immediately-preceding statement is a `(...)` argument-forwarding call to a native method, raising a bogus `ArgumentError: wrong number of arguments (given N, expected N-1)`.

Found via sigstore-ruby on JRuby 10.x: a `bundle.expected_tlog_entry(hashed_input, verifier_pem:)` call raised `ArgumentError` only because the preceding statement forwarded `(...)` to a native method (`OpenSSL::X509::Certificate#to_pem`).

Minimal reproduction (gem-free; `openssl` ships with JRuby):

```ruby
require "openssl"
cert = OpenSSL::X509::Certificate.new   # any signed cert
def forward(o, ...) = o.to_pem(...)     # `(...)` forward to a native method
def kwm(a, k: nil) = [a, k]
v = forward(cert)
kwm("HI", k: v)   # => ArgumentError: wrong number of arguments (given 2, expected 1)
```

Reproduces under `-X-C` as well, so it is an IR-runtime bug, not JIT-specific.

## Root cause

`ThreadContext#callInfo` is transient state passed from a caller's argument build to its callee. `CALL_KEYWORD_EMPTY` (`1<<3`) is set dynamically while building a call's arguments (`Helpers#argsPush`, `IRRuntimeHelpers#isHashEmpty`/`#irSplat`) when a keyword-rest or splat turns out empty, and must survive into the call so the callee treats the kwargs as explicitly empty.

`IRRuntimeHelpers#setCallInfo` carried that bit forward unconditionally:

```java
context.callInfo = (context.callInfo & CALL_KEYWORD_EMPTY) | flags;  // had a FIXME
```

`callInfo` is normally cleared on the callee side in `#receiveKeywords` via `resetCallInfo`, but a native callee (a generated `$INVOKER` with no `recv_kw` step) does not clear it — and the precompiled jruby-openssl invoker predates and bypasses the `InvocationMethodFactory` wipe. So the empty bit set by the `(...)` forward survives, and the next keyword call inherits it and skips keyword processing.

## Fix

A legitimate `CALL_KEYWORD_EMPTY` is only ever produced for a call that itself passes a keyword-rest or splat. So carry it forward only when this call's flags indicate `CALL_KEYWORD_REST` or `CALL_SPLATS`; otherwise a leftover bit from a previous call is stale and is dropped. This preserves correct empty-`**`/splat delegation — which setting `callInfo = flags` outright would regress — while stopping the leak into unrelated keyword calls.

## Commits

1. **Add failing test** — `test_call_info.rb#test_forward_dots_to_native_method_does_not_leak`, documenting the bug (fails on the base).
2. **Fix** — the targeted `setCallInfo` change above.

## Verification

- `test/jruby/test_call_info.rb`: 34/34 pass under both JIT and `-X-C`.
- `spec/ruby/language` keyword/method/block specs: no regressions vs. a rebuilt baseline (notably the empty-`**` delegation specs still pass).

### DIFF
--- a/core/src/main/java/org/jruby/ir/runtime/IRRuntimeHelpers.java
+++ b/core/src/main/java/org/jruby/ir/runtime/IRRuntimeHelpers.java
@@ -936,8 +936,19 @@ public class IRRuntimeHelpers {
 
     @JIT @Interp
     public static void setCallInfo(ThreadContext context, int flags) {
-        // FIXME: This may propagate empty more than the current call?   empty might need to be stuff elsewhere to prevent this.
-        context.callInfo = (context.callInfo & CALL_KEYWORD_EMPTY) | flags;
+        // CALL_KEYWORD_EMPTY is set dynamically while building this call's arguments (argsPush,
+        // isHashEmpty, irSplat) when a keyword-rest or splat turns out to be empty, and it must
+        // survive into the call so the callee treats the kwargs as explicitly empty. It is only
+        // ever set for a call that passes a keyword-rest or splat, so we carry it forward only
+        // when this call is such a call. Otherwise a leftover CALL_KEYWORD_EMPTY from a previous
+        // call whose (native) callee never reset callInfo would leak into an unrelated call (e.g.
+        // a plain keyword call), causing the keyword argument to be mishandled. See
+        // test/jruby/test_call_info.rb#test_forward_dots_to_native_method_does_not_leak.
+        if ((flags & (CALL_KEYWORD_REST | CALL_SPLATS)) != 0) {
+            context.callInfo = (context.callInfo & CALL_KEYWORD_EMPTY) | flags;
+        } else {
+            context.callInfo = flags;
+        }
     }
 
     public static void checkForExtraUnwantedKeywordArgs(ThreadContext context, final StaticScope scope, RubyHash keywordArgs) {
--- a/test/jruby/test_call_info.rb
+++ b/test/jruby/test_call_info.rb
@@ -1,6 +1,7 @@
 require 'test/unit'
 require 'tempfile'
 require 'stringio'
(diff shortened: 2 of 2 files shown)


headius:
Good find, I think I ran into this while investigating #9472. I or @enebo will review and merge.

headius:
@enebo If you get a chance to review I would appreciate extra eyes. It seems right to me.

headius:
@segiddins I'm wondering if there's a more generic test in MRI we could enable/un-exclude or a test we could add to ruby/spec that doesn't require openssl specifically.

segiddins:
There probably is! I just really struggled to get another native-implemented method to reproduce -- the reduction already took me (and my french friend) a while to narrow in on. I'm happy for either the implementation of the test to change, I just figured it would be more helpful to share both than just an issue that required running a whole other test suite

enebo:
This seems good to me.  We went through and audited setCallInfo a while back due to us using that method incorrectly someplace.  The documentation along in this PR is also good as it explains the odd nature of empty. 

@headius this probably should be cherry-picked back to 10.0.

@segiddins If you do find a more generic (or we do)  then we can eliminate this test and move another test into ruby/spec.

kares:
> Found via sigstore-ruby on JRuby 10.x: a bundle.expected_tlog_entry(hashed_input, verifier_pem:) call raised ArgumentError only because the preceding statement forwarded (...) to a native method (OpenSSL::X509::Certificate#to_pem)

think this should go towards **jruby-10.0** branch as well, there were similar CALL_INFO leak fixes in 10.0.5.0

PR #9502, opened 2026-06-22
avoid setting error-info on RaiseException's init

kares:
The `$!` semantics are handled by IR protocol, the  `RaiseException`'s filling of `$!` has been in-place for a very long time but is not the right place to set error-info.


kares:
seems like something is broken around the branch, there should be more changes showing up than 1 commit...

PR #9503, opened 2026-06-22
avoid setting $! from RaiseException <init>

kares:
The `$!` semantics are handled by IR protocol, specifically the `rescue`/`ensure` parts (and also from `raise` internally), `new RaiseException()`'s filling of `$!` has been in-place for a very long time but is not the right place to set error-info.

Maybe we can get rid of this once and for all... :crossed_fingers:

### DIFF
--- a/core/src/main/java/org/jruby/exceptions/RaiseException.java
+++ b/core/src/main/java/org/jruby/exceptions/RaiseException.java
@@ -212,8 +212,6 @@ public class RaiseException extends JumpException {
         context.runtime.incrementExceptionCount();
         if (RubyInstanceConfig.LOG_EXCEPTIONS) TraceType.logException(exception);
 
-        doSetLastError(context);
-
         if (backtrace == null) {
             if (capture) { // only false to support legacy RaiseException construction (not setting trace)
                 if (requiresBacktrace(context)) exception.captureBacktrace(context);
@@ -272,12 +270,8 @@ public class RaiseException extends JumpException {
         if (newTrace != curTrace) setStackTrace(newTrace);
     }
 
-    private void doSetLastError(final ThreadContext context) {
-        context.setErrorInfo(exception); // $!
-    }
-
     /**
-     * Sets the exception
+      * Sets the exception
      * @param newException The exception to set
      */
     protected final void setException(RubyException newException) {
--- a/core/src/main/java/org/jruby/exceptions/RaiseException.java
+++ b/core/src/main/java/org/jruby/exceptions/RaiseException.java
@@ -35,6 +35,8 @@
 
 package org.jruby.exceptions;
 
+import java.util.Arrays;
+
 import org.jruby.Ruby;
 import org.jruby.RubyArray;
 import org.jruby.RubyClass;
@@ -49,9 +51,6 @@ import org.jruby.runtime.ThreadContext;
 import org.jruby.runtime.backtrace.RubyStackTraceElement;
 import org.jruby.runtime.backtrace.TraceType;
 import org.jruby.runtime.builtin.IRubyObject;
-
-import java.lang.reflect.Member;
-import java.util.Arrays;
 
 import static org.jruby.api.Access.objectClass;
 import static org.jruby.api.Error.nameError;
@@ -263,15 +262,8 @@ public class RaiseException extends JumpException {
         return trace;
     }
 
-    private void fillInStackTraceSkipPreRaise() {
-        originalFillInStackTrace(); // (fillInStackTraceSkipPreRaise) originalFillInStackTrace, preRaise
-        StackTraceElement[] curTrace = getStackTrace();
-        StackTraceElement[] newTrace = skipFillInStackTracePart(curTrace);
-        if (newTrace != curTrace) setStackTrace(newTrace);
-    }
-
     /**
-      * Sets the exception
+     * Sets the exception
      * @param newException The exception to set
      */
     protected final void setException(RubyException newException) {
--- a/core/src/main/java/org/jruby/exceptions/RaiseException.java
+++ b/core/src/main/java/org/jruby/exceptions/RaiseException.java
@@ -223,8 +223,6 @@ public class RaiseException extends JumpException {
             }
             setStackTraceFromException();
         }
-
-        IRRuntimeHelpers.traceRaise(context);
     }
 
     private void setStackTraceFromException() {
--- a/core/src/main/java/org/jruby/internal/runtime/GlobalVariable.java
+++ b/core/src/main/java/org/jruby/internal/runtime/GlobalVariable.java
@@ -31,6 +31,7 @@
 package org.jruby.internal.runtime;
 
 import java.util.ArrayList;
+import java.util.List;
 
 import org.jruby.Ruby;
 import org.jruby.RubyProc;
@@ -68,13 +69,13 @@ public final class GlobalVariable {
         return scope;
     }
 
-    public ArrayList getTraces() {
+    public List<IRubyObject> getTraces() {
         return traces;
     }
 
     public void addTrace(RubyProc command) {
         if (traces == null) {
-            traces = new ArrayList<IRubyObject>(1);
+            traces = new ArrayList<>(2);
         }
         traces.add(command);
     }
--- a/core/src/main/java/org/jruby/internal/runtime/GlobalVariable.java
+++ b/core/src/main/java/org/jruby/internal/runtime/GlobalVariable.java
@@ -69,7 +69,7 @@ public final class GlobalVariable {
         return scope;
     }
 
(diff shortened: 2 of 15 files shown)


headius:
This seems like an appropriate time to make this change. My immediate concern was about not properly setting $! every time we raise a Ruby exception from Java, but such cases should be using `Kernel#raise` logic in general.

I'll do a review with more specific notes. This could potentially go into 10.1.1.0 if you are happy with it, @kares.

core/src/main/java/org/jruby/exceptions/RaiseException.java
```diff
@@ -272,12 +270,8 @@ private void fillInStackTraceSkipPreRaise() {
         if (newTrace != curTrace) setStackTrace(newTrace);
     }
 
-    private void doSetLastError(final ThreadContext context) {
-        context.setErrorInfo(exception); // $!
-    }
-
     /**
-     * Sets the exception
+      * Sets the exception
```
headius: Remove whitespace change.

core/src/main/java/org/jruby/internal/runtime/GlobalVariable.java
```diff
@@ -68,13 +69,13 @@ public Scope getScope() {
         return scope;
     }
 
-    public ArrayList getTraces() {
+    public List<IRubyObject> getTraces() {
```
headius: This is technically a public API but I'm not sure how likely it is to be a problem.

core/src/main/java/org/jruby/ir/builder/EnsureBlockInfo.java
```diff
@@ -6,7 +6,7 @@
 import org.jruby.ir.instructions.ExceptionRegionStartMarkerInstr;
 import org.jruby.ir.instructions.Instr;
 import org.jruby.ir.instructions.LabelInstr;
-import org.jruby.ir.instructions.RuntimeHelperCall;
+import org.jruby.ir.instructions.defined.RestoreErrorInfoInstr;
```
headius, line 9: Most of the rest of PR is related to this IR change and associated downstream changes for JIT. This all seems generally fine and I'm glad to get rid of another RuntimeHelperCall usage.

core/src/main/java/org/jruby/ir/instructions/ThrowExceptionInstr.java
```diff
@@ -47,14 +47,14 @@ public Object interpret(ThreadContext context, StaticScope currScope, DynamicSco
...
 
-        if (excObj instanceof IRubyObject exc) {
-            RubyKernel.raise(context, kernelModule(context), new IRubyObject[] {exc}, Block.NULL_BLOCK);
-        } else if (excObj instanceof Throwable exc) { // java exception -- avoid having to add 'throws' clause everywhere!
-            Helpers.throwException(exc);
+        if (excObj instanceof Throwable) {
+            excObj = Helpers.wrapJavaException(context.runtime, (Throwable) excObj); // IRubyObject
         }
 
+        RubyKernel.raise(context, kernelModule(context), new IRubyObject[] {(IRubyObject) excObj}, Block.NULL_BLOCK);
```
headius: Given that we are expecting all Ruby-visible exception throws to go through `Kernel#raise` logic, we should finally make this easy. This signature requires an argument array, reference to a self object, and a block reference. Add some signatures (maybe in RubyKernel, or perhaps in RubyException) to make this easy for JRuby's Java code to call it.

headius (changes requested):
Minor changes requested but overall it looks fine. Much of the changes are to support using purpose-built IR instructions and cleaning up messy code. A bit more cleanup and clarification and this can merge.

core/src/main/java/org/jruby/ir/instructions/ThrowExceptionInstr.java
kares: have dried out some of the `RubyKernel.raise` exception logic and added a `RubyException.raise` helper https://github.com/jruby/jruby/pull/9503/changes/7274cc54cb7c20ca305401ac753589156aabde88 it's problematic to add to RubyKernel due the existing overloads.

core/src/main/java/org/jruby/internal/runtime/GlobalVariable.java
kares: should be internal, but just in case restored the signature

headius:
Rails verifications failed due to a `raise` event hook suddenly getting nil values. This resulted in the server not starting, and the verification script is not smart enough to fail so it ran forever. I've killed it but that definitely seems like it's related to this PR.

Perhaps we should let this one bake and merge in 10.1.2.0.

headius:
Here's a patch that might be appropriate. It would be nice to try to match up the exact sequence of events and annotate the calls we make with the equivalent functions in CRuby, because our `setup_exception` is currently `preRaise` and it gets called regardless of whether `raise` is called. CRuby appears to defer exception setup longer, until it's actually about to be raised.

```diff
diff --git a/core/src/main/java/org/jruby/RubyKernel.java b/core/src/main/java/org/jruby/RubyKernel.java
index 33c1bcd6b9..0267466cd0 100644
--- a/core/src/main/java/org/jruby/RubyKernel.java
+++ b/core/src/main/java/org/jruby/RubyKernel.java
@@ -1110,7 +1110,12 @@ public class RubyKernel {
     private static IRubyObject raiseException(ThreadContext context, RubyException exception) {
         printDebugException(context, exception);
 
-        context.setErrorInfo(exception); // set $! as part of the raise flow (like MRI's setup_exception)
+        // set $! as part of the raise flow (like MRI's setup_exception)
+        context.setErrorInfo(exception);
+
+        // trace the raise immediately before throwing
+        IRRuntimeHelpers.traceRaise(context);
+
         return Helpers.throwExceptionT(exception.toThrowable());
     }
 
diff --git a/core/src/main/java/org/jruby/exceptions/RaiseException.java b/core/src/main/java/org/jruby/exceptions/RaiseException.java
index 095ebb9107..ce31f76291 100644
--- a/core/src/main/java/org/jruby/exceptions/RaiseException.java
+++ b/core/src/main/java/org/jruby/exceptions/RaiseException.java
@@ -223,8 +223,6 @@ public class RaiseException extends JumpException {
             }
             setStackTraceFromException();
         }
-
-        IRRuntimeHelpers.traceRaise(context);
     }
 
     private void setStackTraceFromException() {
```

I also discovered an old bug in our trace logic: we are not clearing $! during the hook body as CRuby does. The following spec tests this, and fails on all versions of JRuby I tested.

```diff
diff --git a/spec/ruby/core/tracepoint/trace_spec.rb b/spec/ruby/core/tracepoint/trace_spec.rb
index 167f594bb9..c1bf364002 100644
--- a/spec/ruby/core/tracepoint/trace_spec.rb
+++ b/spec/ruby/core/tracepoint/trace_spec.rb
@@ -7,4 +7,18 @@ describe 'TracePoint.trace' do
     trace.should.enabled?
     trace.disable
   end
+
+  it 'clears $! when invoking the trace block' do
+    exception_in_trace = nil
+    trace = TracePoint.trace(:raise) do |tp|
+      exception_in_trace = $!
+    end
+    begin
+      raise
+    rescue => e
+      exception_in_trace.should == nil
+      $!.should == e
+    end
+    trace.disable
+  end
 end
```

This does not appear to be tested in any Ruby spec. I did not investigate CRuby tests.

headius:
@eregon Unsure where to put the spec above. It is testing behavior of the hook block passed to `TracePoint.new` or `TracePoint.trace` but it is not *really* a behavior of either of those methods. I went with `TracePoint.trace` specs for simplicity, but the behavior applies to all forms of tracing.

eregon:
That's good. Especially TracePoint tends to cause some overhead so it's good to keep specs using TracePoint under `core/tracepoint`.

> applies to all forms of tracing.

BTW what other forms are there? `set_trace_func`? Anything else?

headius:
> BTW what other forms are there? set_trace_func? Anything else?

@eregon Those are the two Ruby-facing APIs. I believe both should probably have this specification. Even though they both are implemented with the same event hook logic in CRuby, that might not be the case on all implementations (for example, JRuby might in the future use Java debug APIs to hook up some of these events, like raising).

The C-facing APIs probably should also have specs, but I am not equipped to write or test them.

core/src/main/java/org/jruby/ir/instructions/ThrowExceptionInstr.java
headius: Yeah I think that's what held me back from arity-splitting `Kernel#raise` in the past. This is a good time to do it and I agree with having the actual utility code elsewhere (I suggested RubyException, but we might want to move it to one of the org.jruby.api utilities like Error; that can happen after I merge in a moment).

headius (approved):

PR #9511, opened 2026-06-29
Update jnr:* and asm to latest

headius:


headius:
Dependency convergence failing due to https://github.com/roastedroot/chicory-redline/pull/20 not being finished and merged yet. Looking into that. cc @andreaTP

headius:
The aforementioned chicory-redline PR looks good now and ideally will be merged and released soon. Once that happens we can plumb that dependency through the Prism libraries and get this convergence resolved.

headius:
Draft until snapshot dep of jruby-prism is resolved.

PR #9512, opened 2026-06-29
Begin deprecating jruby-complete and jruby-core

headius:
After jruby-dist, which is necessary to release JRuby for Ruby users, jruby-complete is by far the largest artifact published for JRuby, combining all Java dependencies and all Ruby dependencies into a single jar file. This combination can be represented by the org.jruby:jruby artifact, which depends upon jruby-base (itself depending on all Java dependencies) and jruby-stdlib (containing all Ruby dependencies. The only utility of jruby-complete is to publish a single uber-jar.

Similarly, the jruby-core artifact is essentially just a version of jruby-base that shades all Java dependencies. It cannot run on its own due to the missing Ruby dependencies, but it duplicates the logical content of jruby-base.

With Maven Central moving toward publishing limits, we would like to start reducing the number and size of artifacts we push. This commit starts the process of deprecating these two artifacts.

* jruby-complete is deprecated in favor of org.jruby:jruby. Users needing an uber-jar can use other mechanisms to merge all transitive dependencies of that artifact into a single library.
* jruby-core is deprecated in favor of org.jruby:jruby-base, which provides the same logical contents. Users needing a shaded version of jruby-base can use other mechanisms to achieve it.

Removing these two artifacts would reduce the size of a JRuby release to Maven Central by about 71MB based on the 10.1.0.0 release (~30MB for jruby-core and ~41MB for jruby-complete).

### DIFF
--- a/maven/jruby-complete/pom.rb
+++ b/maven/jruby-complete/pom.rb
@@ -11,6 +11,11 @@ project 'JRuby Complete' do
   inherit "org.jruby:jruby-artifacts:#{version}"
   packaging 'bundle'
 
+  distribution_management do
+    relocation(groupId: 'org.jruby', artifactId: 'jruby',
+               message: 'The jruby-complete artifact is deprecated in favor of the jruby artifact.')
+  end
+
   extension 'org.jruby.maven:mavengem-wagon:2.0.2'
 
   plugin_repository id: :mavengems, url: 'mavengem:https://rubygems.org'
--- a/shaded/pom.rb
+++ b/shaded/pom.rb
@@ -9,6 +9,11 @@ project 'JRuby Core' do
   inherit 'org.jruby:jruby-parent', version
   id 'org.jruby:jruby-core'
 
+  distribution_management do
+    relocation(groupId: 'org.jruby', artifactId: 'jruby-base',
+               message: 'The jruby-core artifact is deprecated in favor of the jruby-base artifact.')
+  end
+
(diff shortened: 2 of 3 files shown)


headius:
Example project to test these deprecation messages is here:

https://github.com/headius/jruby-complete-deprecation-test

Output at the top of the build:

```
$ mvn clean package
[INFO] Scanning for projects...
[INFO] 
[INFO] ------------< com.headius:jruby-complete-deprecation-test >-------------
[INFO] Building jruby-complete-deprecation-test 1.0-SNAPSHOT
[INFO]   from pom.xml
[INFO] --------------------------------[ jar ]---------------------------------
[WARNING] The artifact org.jruby:jruby-complete:jar:10.1.1.0-SNAPSHOT has been relocated to org.jruby:jruby:jar:10.1.1.0-SNAPSHOT: The jruby-complete artifact is deprecated in favor of the jruby artifact.
[WARNING] The artifact org.jruby:jruby-core:jar:10.1.1.0-SNAPSHOT has been relocated to org.jruby:jruby-base:jar:10.1.1.0-SNAPSHOT: The jruby-core artifact is deprecated in favor of the jruby-base artifact.
[INFO] 
```

headius:
We'll push this forward and see if we hear from users that they absolutely need one of these two artifacts. I suspect most jruby-core users have switched to the jruby or jruby-base artifacts, but jruby-complete is a harder sell (it's our only "uber-jar" that can be run directly).

PR #9515, opened 2026-06-30
Switch to jar-deps based on mima

### LINKED ISSUE #9301: Find an alternative to shipping ruby-maven
For https://github.com/ruby/rubygems/issues/9386 I was forced to add ruby-maven as a hard dependency of jar-dependencies, so that it would get fully installed before the jar-dependencies post-install hooks execute.

I believe bundler/rubygems could do a better job here and defer post-install hooks until *after* all gems install, but until that is fixed I could find no other workaround.

The problem is that shipping ruby-maven and ruby-maven-libs with JRuby increases the size of our artifacts by at least 20MB, due to the large size of the Maven libraries and dependencies.

We need to consider other options to work around the issue above.

cc @mkristian for ideas.

headius:
We have a working version of jar-dependencies now using the mima stripped-down Maven dependency resolver. This PR will switch us to jar-dependencies 0.6.0 to eliminate issues with post-install hooks trying to install ruby-maven and hitting concurrency challenges.

This is based on finalization changes in https://github.com/jruby/jar-dependencies/pull/107.

### DIFF
--- a/lib/pom.rb
+++ b/lib/pom.rb
@@ -48,7 +48,7 @@ default_gems = [
   # ['io-nonblock', '0.3.2'],
   ['io-wait', '0.4.0'],
   ['ipaddr', '1.2.8'],
-  ['jar-dependencies', '0.5.7'],
+  ['jar-dependencies', '0.6.0.pre2'],
   ['jruby-readline', '1.3.7'],
   ['jruby-openssl', '0.16.0'],
   ['json', '2.18.0'],
--- a/lib/pom.rb
+++ b/lib/pom.rb
@@ -48,7 +48,7 @@ default_gems = [
   # ['io-nonblock', '0.3.2'],
   ['io-wait', '0.4.0'],
   ['ipaddr', '1.2.8'],
-  ['jar-dependencies', '0.5.7'],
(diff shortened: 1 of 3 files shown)


headius:
Premature to try to merger this into 10.1.1.0 so we'll punt for one more release.

kares:
we should make sure to have these included in the release:
- https://github.com/jruby/jar-dependencies/pull/108
- https://github.com/jruby/jar-dependencies/pull/107
- https://github.com/jruby/jar-dependencies/pull/109

headius:
All the PRs that @kares pointed out have been merged and released in pre2, along with:

* https://github.com/jruby/jar-dependencies/pull/110

A separate ping for @mkristian here... we are about to move forward with this.

Once jar-dependencies 0.6.0 is released we'll also update JRuby 10.1 dev builds and see if it impacts anyone. 🤞

headius:
Also need to include the update of mima in 0.6.0 final:

* https://github.com/jruby/jar-dependencies/pull/111

mkristian:
Coolio. Happy to see this going in. I am sure it removes a lot old quirks

Issue #9493, opened 2026-06-12
Use of mvnw on CI seems flaky


chadlwilson:
Since I changed the workflows to use mvnw consistently, I've noticed builds seem to flake reasonably frequently, which is a waste of people's time.

Not sure exactly why, possibly due to newer rate limiting, maybe compounded by lack of caching in the workflow, but maven/Apache don't seem to publish their non-latest builds to a decent CDN for some reason.

I should probably look at it and find an improvement if no-one else gets to it.

headius:
I have noticed it failing occasionally as well. Taking five minutes to research it.

headius:
We might have better luck using the main Maven Central URL to download, which I believe is mirrored and federated across a number of servers:

https://repo1.maven.org/maven2/org/apache/maven/apache-maven/3.9.16/apache-maven-3.9.16-bin.zip

We might also be able to add caching for the wrappers to our builds (using a simple cache key based on the version, perhaps), or get the setup-java action to add wrapper caching by default into the `cache: maven` handling.

chadlwilson:
Oh < if they publish dists to maven central I wonder why they don't use that, given issues on their own infra.... That'd be a quick win.

The wrappers are just using scripts now, so nothing binary to cache there now, plus setup-java already caches the distributions if enabled with `maven` style: https://github.com/actions/setup-java#caching-packages-dependencies

I didn't actually check the workflows, but I imagined the problem with naively adding caching everywhere may be bad cache keys if anything polyglot is downloading stuff into the Maven cache. But didn't spend any time yet.

headius:
> nothing binary to cache there now

The wrapper fetches maven to `~/.m2/wrappers` so that would be the additional location to cache across runs. Since only downloaded Maven release go there the cache key could be trivial.

I'll make a change on 10.0 branch to use the Maven Central URL and we'll see if the download failures go away.

headius:
Ugh... the PR to move to using a Maven Central URL immediately had an mvnw download failure: https://github.com/jruby/jruby/actions/runs/28970521797/job/85964762963?pr=9518

headius:
> setup-java already caches the distributions

I missed you mentioning this. I'm not sure whether this is cached separately or as part of the overall maven cache; if the latter, then it would be invalidated across PRs, branches, commits potentially (unsure what granularity) which would gain us less than the former model (since it could be cached almost indefinitely so long as the config has not changed).

[...]

PR #9518, opened 2026-07-08
Use Maven Central URL to fetch Maven for mvnw

headius:
The original URL here, the default provided by the wrapper logic from Apache, uses an Apache URL that apears to have some rate limiting or download issues that sometimes fails in CI. This patch changes to using the Maven Central download URL which should be mirrored and federated and avoid these issues.

Relates to jruby/jruby#9493


headius:
Immediately failed to download maven for one job, so this fix doesn't seem to help. Copilot's explanation did not provide any solution, but claimed that repo.maven.apache.org is actually the better canonical address to use.

I don't think this is our fault at this point. Perhaps we can turn on better logging to know why the download fails.

Closing this as it does not appear to be the right approach.

PR #9519, opened 2026-07-08
Also set up the current module when in a for body

### LINKED ISSUE #9491: defined?(...) in eval'd for loop fails with TypeError
When executing this code:

```ruby
class ClassB
  def foo
    code = eval( %Q{lambda { |*params|
for r in [1, 2, 3]
  puts defined?(Array)
end
}
} )

      self.define_singleton_method( :bar, &code )

self
  end
end

ClassB.new.foo.bar(1, 2, 3)
```

the following error happens:

```
(eval):3:in `block in foo':  is not a type/class (TypeError)
	from org/jruby/RubyArray.java:2018:in `each'
	from (eval):1:in `block in foo'
	from (irb):62:in `evaluate'
	from org/jruby/RubyKernel.java:1277:in `eval'
	from org/jruby/RubyKernel.java:1725:in `loop'
	from org/jruby/RubyKernel.java:1426:in `catch'
	from /Users/gerhard/.rbenv/versions/jruby-9.4.14.0/lib/ruby/gems/shared/gems/irb-1.15.3/exe/irb:9:in `<main>'
	from org/jruby/RubyKernel.java:1223:in `load'
	from /Users/gerhard/.rbenv/versions/jruby-9.4.14.0/bin/irb:25:in `<main>'
```

**Environment Information**

Tested with Ruby 9.4.14.0 on Java 11.0.31:

```
JRUBY_VERSION
=> "9.4.14.0"

Java::java.lang.System.get_property("java.version")
=> "11.0.31"
```

Operating system:

```bash
uname -a
Darwin MacBook-Pro-7.local 25.5.0 Darwin Kernel Version 25.5.0: Mon Apr 27 20:39:42 PDT 2026; root:xnu-12377.121.6~2/RELEASE_ARM64_T6031 arm64
```

**Expected Behavior**

When executing the Script with CRuby/MRI, I get the following:

```
ruby /tmp/script.rb 
constant
constant
constant
```

**Actual Behavior**

* The error from above

headius:
Regardless of where the current module comes from, we need this setup to ensure that the temp variable %current_module has been initialized, or else we may end up in a situation like #9491 where we expect it to be available and it is not.

It's not clear to me why this logic was gated on being a for loop body.

Fixes #9491

### DIFF
--- a/core/src/main/java/org/jruby/ir/builder/IRBuilder.java
+++ b/core/src/main/java/org/jruby/ir/builder/IRBuilder.java
@@ -1619,7 +1619,7 @@ public abstract class IRBuilder<U, V, W, X, Y, Z> {
         // can be U_NIL if the node is an if node with returns in both branches.
         if (closureRetVal != U_NIL) addInstr(new ReturnInstr(closureRetVal));
 
-        prependUsedClosureImplicitState(forNode);
+        prependUsedClosureImplicitState();
 
         // Add break/return handling in case it is a lambda (we cannot know at parse time what it is).
         // SSS FIXME: At a later time, see if we can optimize this and do this on demand.
@@ -1655,7 +1655,7 @@ public abstract class IRBuilder<U, V, W, X, Y, Z> {
         // can be U_NIL if the node is an if node with returns in both branches.
         if (closureRetVal != U_NIL) addInstr(new ReturnInstr(closureRetVal));
 
-        prependUsedClosureImplicitState(false);
+        prependUsedClosureImplicitState();
 
         handleBreakAndReturnsInLambdas();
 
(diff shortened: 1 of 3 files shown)


headius:
Needs a test or non-excluded spec.

headius:
A bit more spelunking here leads to this code:

https://github.com/jruby/jruby/blob/af10e264436022a27d4ac9a4207b25f791159d4b/core/src/main/java/org/jruby/ir/IRScope.java#L353-L357

When preparing the "current module" operand for things like `defined?` this logic will return -1 for any eval scope, which causes it to fall back on using the `%current_module` variable to use for constant searches. It does this because eval-scoped bodies can get re-homed into different hierarchies, so we can't rely on the containing hierarchy remaining the same.

This logic combined with the patch in this PR ultimately caused #9491.

The fix I've made is somewhat form-fitted but I can think of no cases where the temporary variable should *not* be set up in this way; if it's left uninitialized, we may attempt to do scoped or inherited module access against nil, which is never correct.

The risk of this fix is that we might not be setting up this current module correctly for a `for` in an `eval`, but the situation is made no worse by merging this patch.

PR #9523, opened 2026-07-11
Don't copy the singleton class in Proc#dup

shugo:
procDup() passed getMetaClass() to the copy, which is the receiver's singleton class when the proc has singleton methods, so the dup unexpectedly responded to the receiver's singleton methods; MRI drops the singleton class on dup:

    pr = proc { }
    def pr.tag; end
    pr.dup.respond_to?(:tag)  # JRuby: true, MRI: false

Pass the real class instead.  Proc#clone is unaffected because cloneSetup already replaces the copy's metaclass with a clone of the receiver's singleton class, and Proc subclasses are preserved since the real class of a subclass instance is the subclass itself.

### DIFF
--- a/core/src/main/java/org/jruby/RubyProc.java
+++ b/core/src/main/java/org/jruby/RubyProc.java
@@ -259,7 +259,7 @@ public class RubyProc extends RubyObject implements DataType {
     }
 
     private RubyProc procDup() {
-        return newProc(getRuntime(), getMetaClass(), block, type, file, line);
+        return newProc(getRuntime(), getMetaClass().getRealClass(), block, type, file, line);
     }
 
     @Override


core/src/main/java/org/jruby/RubyProc.java
```diff
@@ -259,7 +259,12 @@ public IRubyObject dup(ThreadContext context) {
     }
 
     private RubyProc procDup() {
-        return newProc(getRuntime(), getMetaClass(), block, type, file, line);
+        // The real class, not the receiver's metaclass: dup must not carry the receiver's
+        // singleton class over to the copy (MRI drops it), and dupSetup does not reset the
```
kares: this is fairly apparent from the code itself, no need for LLM generated 5 line comment (also okay if you just compress and keep these 2 lines).

spec/regression/proc_dup_does_not_copy_singleton_class_spec.rb
```diff
@@ -0,0 +1,31 @@
+require 'rspec'
+
+# Proc#dup went through the receiver's metaclass, so a proc with singleton
+# methods handed its singleton state to the dup (MRI drops it; only clone
+# keeps a copy of the singleton class).
+describe 'Proc#dup' do
```
kares: the regression spec suite is a bit deprecated, these would make sense as ruby/spec (there might be one already, maybe it's just excluded)

kares:
:+1: good catch

shugo:
@kares Thanks for your review.

I've removed obvious comments, and have removed the regression spec.
I'll check ruby/spec later.

headius (approved):

headius:
Thanks @shugo!

PR #9527, opened 2026-07-16
Use setup-java with fixed maven wrapper caching

### LINKED ISSUE #9526: Confirm setup-java is latest to get Maven wrapper caching
Maven wrapper caching in setup-java was using the same key as the build dependencies from pom.xml, which meant it invalidated much more easily than it should. This has been fixed after my bug report https://github.com/actions/setup-java/issues/1095 so we should make sure all our setup-java are using an appropriate version (may already be gaining the improved functionality via "v5").

headius:
Fixes #9526

See https://github.com/actions/setup-java/issues/1095

Update to release version before merging.


headius:
This has been verified and the associated fix in setup-java has been released as v5.6. Our use of `setup-java@v5` should pick that change up automatically, so there's nothing to merge.

PR #9535, opened 2026-07-18
Reduce Proc#refined tests to JRuby-specific ones

shugo:
Implementation-agnostic behavior is covered by
spec/ruby/core/proc/refined_spec.rb (synced from ruby/spec) and by CRuby's test/ruby/test_proc.rb (synced as test/mri), so keep only the tests tied to JRuby internals: clone retention in the IR lexical parent, JIT compilation of the clone, concurrent refine/refined, backtrace construction through a top-level clone, and GC survival.

Requested in https://github.com/jruby/jruby/pull/9486


headius:
Thank you @shugo!

I have started updating the ruby-4.1 branch with tests from CRuby so they'll come along eventually.

The specs get synced to master periodically and master gets merged to ruby-4.1.

headius:
@shugo I also merged my PR that sets Ruby version to 4.1. Should be easier to validate specs now.

PR #9548, opened 2026-07-22
[fix] re-raising exception with nil backtrace

### LINKED ISSUE #9528: raise should re-populate exception trace set to nil
Spec came in recently and lack of this feature breaks several tests in net-imap.

Spec is here, and applies to Kernel, Fiber, and Thread#raise:

https://github.com/jruby/jruby/blob/576ea48c5de0311b82392fa7d46bab2e9670c457/spec/ruby/shared/kernel/raise.rb#L144

Discussion in net-imap: https://github.com/ruby/net-imap/discussions/720

cc @nevans

kares:
tried to keep the change-set minimal, will likely cause conflicts when merging over to master (10.1)

### DIFF
--- a/core/src/main/java/org/jruby/RubyException.java
+++ b/core/src/main/java/org/jruby/RubyException.java
@@ -543,7 +543,7 @@ public class RubyException extends RubyObject {
      *
      * MRI {@code setup_exception} which fills the backtrace only when it is <code>nil</code>
      */
-    void prepareBacktrace(ThreadContext context) {
+    public void prepareBacktrace(ThreadContext context) {
         if (getBacktrace().isNil()) {
             clearBacktraceObject();
             captureBacktrace(context);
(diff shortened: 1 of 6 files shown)


headius (approved):

headius:
I'll handle the merge.

headius:
I believe I got the merge basically correct, but a newer spec appeared to have more assertions this PR did not pass.

1. I had to force the backtrace to be generated eagerly, since otherwise it might not reflect the correct call stack when it is generated later lazily: https://github.com/jruby/jruby/blob/7ab39ec0d466fb3c59c29049ba6da0bf907a5cc2/core/src/main/java/org/jruby/RubyException.java#L585
2. I had to modify the newer spec to not expect the class name in the backtrace (we do not support this yet): https://github.com/jruby/jruby/commit/7ab39ec0d466fb3c59c29049ba6da0bf907a5cc2#diff-58136503b89bf15c23c03cc117cde181e07bc5e2b49db8bbc44909f335856688L175

There are still outstanding tags for both Thread and Fiber#raise that we could work to get rid of.

PR #9555, opened 2026-07-29
Allow no-arg and chained Proc#refined and copy the block lazily

shugo:
Since prc.refined(*ms).refined(*ns) is behaviorally equivalent to prc.refined(*ms, *ns), a chained call re-clones from the original closure with the merged module list and the memoized copy is shared by these behaviorally equivalent Procs.  The copy of the block's IR is deferred until the first call, so a Proc that is never called is no longer copied at all.
(https://bugs.ruby-lang.org/issues/22213, ruby/ruby d4a6282429)


headius:
@shugo Let me know if you need assistance here.

shugo:
> @shugo Let me know if you need assistance here.

Thank you!

[Feature #22213] has been accepted by Matz, so I'd like to squash the commits in this branch and make this pull request ready for review.
However, #9556 doesn't seem to have been merged into the ruby-4.1 branch yet.
Could you merge it?

headius:
@shugo I have merged master into ruby-4.1.

headius (approved):

PR #9556, opened 2026-07-29
Implement Proc#hash consistently with Proc#eql?

shugo:
Proc#hash was missing, so procs that are eql? (e.g. a proc and its dup) hashed differently.


headius (approved):
Looks simple enough and only a slight (null-checking) modification to the already-existing hashCode method.

Issue #9558, opened 2026-08-03
`Kernel#sprintf` calls `Hash#default_proc` with incorrect args


nevans:
Some new ruby specs were recently added for invoking `Hash#default_proc` from `Kernel#sprintf`:
* https://github.com/ruby/spec/commit/d750b85f
* https://github.com/truffleruby/truffleruby/issues/4155
* https://github.com/truffleruby/truffleruby/pull/4156
 
Unfortunately, those specs did not specify that `default_proc` is called with the _correct args!_

Here's some simple code to demonstrate the issue:
```ruby
h = Hash.new {|*args| "args=%p" % [args] }
pp h[:foo]       # normal
pp "%{foo}" % h  # sprintf
```

**Expected Behavior**
For CRuby (and TruffleRuby), we get the expected results:
```
"args=[{}, :foo]"
"args=[{}, :foo]"
```

**Actual Behavior**
For JRuby 10.0.5.0, we see that only the key is passed:
```
"args=[{}, :foo]"
"args=[:foo]"
```

headius:
Fix was simple and is in #9567. I've included spec updates separately in #9569.

PR #9569, opened 2026-08-06
Expand Kernel#sprintf Hash default to include proc behavior

headius:
Old version checked that the proc was called but not that it was called with appropriate arguments.

See jruby/jruby#9558

Draft until fix merges in from 10.0 and this is green.


headius:
Fix is coming in via 10.0 from #9567.

Issue #9559, opened 2026-08-03
dup of frozen class can't be extended


HoneyryderChuck:
**Environment Information**
  
Provide at least:
* JRuby version: jruby 10.1.1.0 (4.0.0) 2026-07-22 aaa95d57e1 OpenJDK 64-Bit Server VM 24.0.1+9-30 on 24.0.1+9-30 +indy +jit [x86_64-darwin]

**Expected Behavior**

the following script works with no issues in cruby:

```ruby
class A; end
module B; end

A.freeze
# A.extend B fails expectedly

dupped = A.dup
dupped.frozen? #=> false

dupped.extend B
```

**Actual Behavior**

In jruby, the last line of the script above fails with the error:

```
org/jruby/RubyModule.java:1741:in 'include': can't modify frozen module: #<Class:C> (FrozenError)
        from org/jruby/RubyModule.java:3780:in 'extend_object'
        from org/jruby/RubyKernel.java:2564:in 'extend'

headius:
Fixed by #9570 for JRuby 10.1.2.0.

PR #9570, opened 2026-08-06
fix dup for frozen class

evaniainbrooks:
issue #9559


headius (approved):

headius:
Looks good! I'll let CI settle down and then merge. Thanks @evaniainbrooks!

headius:
GHA seems to be having troubles today and failed to run several jobs. I'll give them a second attempt but this is probably fine to merge.

PR #9574, opened 2026-08-07
Update to ruby/spec@0f44fd5

andrykonchin:
There are some additional fixes made during specs synchronization.


headius:
Thanks as always, @andrykonchin!

PR #9576, opened 2026-08-07
Update jnr dependencies for BSD aliases

### LINKED ISSUE #9553: FFI warning on when running jruby on OpenBSD/arm64, possible $$ issue as well
Since JRuby 9.0.0.0, OpenBSD has only built a JRuby package on amd64 (dropped i386 due to #3194). Looking at the 10.1.1.0 release notes, OpenBSD/arm64 should be supported now, so I asked an OpenBSD developer with access to arm64 to test it. It appears to work for basic testing, but it results in this annoying warning on start:

```
Jul 26, 2026 5:39:47 PM jnr.ffi.provider.jffi.NativeRuntime buildNativeTypeAliases
SEVERE: failed to load type aliases: java.lang.ClassNotFoundException: jnr.ffi.provider.jffi.platform.aarch64.openbsd.TypeAliases
```

When testing the FFI code with this (which I got from a Google AI result):

```ruby
require 'ffi' 

module Libc 
extend FFI::Library 
ffi_lib FFI::Library::LIBC 
attach_function :getpid, [], :int 
end 

puts "Process ID from C libc: #{Libc.getpid}" 
puts "Process ID from Ruby ($$): #{$sprintf || $$}" 
```

On OpenBSD/amd64, this has expected results:

```
Process ID from C libc: 15145
Process ID from Ruby ($$): 15145
```

On OpenBSD/arm64, it does not:

```
# Warning as shown above
Process ID from C libc: 25308
Process ID from Ruby ($$): 222624801
```

The weird thing here is that the FFI result looks correct (OpenBSD process ids are 1-5 digits), but `$$` looks wrong. So it is possible these are two separate issues, but I'm not sure.

I assume the warning is because I'm doing something wrong in the OpenBSD JRuby port, but I'm not sure what it is. For some more details, see https://marc.info/?l=openbsd-ports&m=178510257247969&w=2 .

Note that the OpenBSD JRuby port builds jffi from scratch (OpenBSD keeps a stable API, not a stable ABI). This warning occurs both with jffi 1.3.15 and with jffi 1.4.0. When building on both amd64 and arm64, we get warnings at the start of the jffi build:

```
-build-native-library: 
     [exec] Configuring libffi for x86_64-openbsd 
     [exec] cc: error: unsupported option '-print-multi-os-directory' 
     [exec] cc: error: no input files 
```
 
However, later in the same `-build-native-library` section, it does compile, link, and strip: 
 
```
     [exec] libtool: compile:  cc -DHAVE_CONFIG_H -I. ... /usr/obj/ports/jruby-10.1.1.0/jruby-10.1.1.0/jffi/jni/libffi/src/prep_cif.c -fPIC -DPIC -o src/.libs/prep_cif.o
     ... 
     [exec] cc -o /usr/obj/ports/jruby-10.1.1.0/jruby-10.1.1.0/jffi/build/jni/libjffi-1.2.so -pthread -shared -static-libgcc /usr/obj/ports/jruby-10.1.1.0/jruby-10.1.1.0/jffi/build/jni/jffi/Array.o ...
     [exec] strip -S /usr/obj/ports/jruby-10.1.1.0/jruby-10.1.1.0/jffi/build/jni/libjffi-1.2.so
```

Since this occurs both on amd64 (where there is no jffi warning at start) and on arm64 (where there is), I would guess that this isn't a compilation issue, but I'm not sure.  Both OpenBSD amd64 and arm64 install the file at `/usr/local/jruby/lib/jni/${JFFI_ARCH}/libjffi-1.2.so` (OpenBSD's JRuby port installs into `/usr/local/jruby`). `JFFI_ARCH` should be `x86_64` on amd64, and `aarch64` on arm64.

The OpenBSD JRuby port can be found at https://github.com/openbsd/ports/tree/master/lang/jruby . This includes the port Makefile as well as the local patches.

Any ideas on what could be the underlying cause or solution?  Is there additional information that would be helpful to provide?

headius:
Updates FreeBSD type aliases and adds OpenBSD type aliases.

Fixes jruby/jruby#9553

Draft until verified and jnr-ffi and its downstream JNR dependencies have been released.

### DIFF
--- a/core/pom.rb
+++ b/core/pom.rb
@@ -49,7 +49,7 @@ project 'JRuby Base' do
   jar 'com.github.jnr:jnr-unixsocket:0.39.1', exclusions: ['com.github.jnr:jnr-ffi']
   jar 'com.github.jnr:jnr-posix:3.2.1', exclusions: ['com.github.jnr:jnr-ffi']
   jar 'com.github.jnr:jnr-constants:0.11.0', exclusions: ['com.github.jnr:jnr-ffi']
-  jar 'com.github.jnr:jnr-ffi:2.3.0'
+  jar 'com.github.jnr:jnr-ffi:2.3.1-SNAPSHOT'
   jar 'com.github.jnr:jffi:${jffi.version}'
   jar 'com.github.jnr:jffi:${jffi.version}:native'
 
--- a/core/pom.xml
+++ b/core/pom.xml
@@ -133,7 +133,7 @@ DO NOT MODIFY - GENERATED CODE
     <dependency>
       <groupId>com.github.jnr</groupId>
       <artifactId>jnr-ffi</artifactId>
-      <version>2.3.0</version>
+      <version>2.3.1-SNAPSHOT</version>
     </dependency>
     <dependency>
       <groupId>com.github.jnr</groupId>


headius:
Closed in favor of #9601.

PR #9588, opened 2026-08-14
Accept `source_offset` in `IO::Buffer#copy` and `#set_string`

### LINKED ISSUE #9587: `IO::Buffer#copy` and `#set_string` reject the documented source_offset fourth argument
## Environment Information

- JRuby 10.0.3.0 and 10.1.1.0 (the same arity checks are present on master, f57d6a54)
- Oracle JDK 25.0.4, macOS 15 (arm64)

## Expected Behavior

CRuby has accepted up to four arguments for both methods since IO::Buffer was introduced:
`io_buffer_copy` and `io_buffer_set_string` both do `rb_error_arity(argc, 1, 4)` in Ruby 3.1.0.
On CRuby 3.4.5 / 4.0.4:

```ruby
b = IO::Buffer.new(64)
b.set_string("XXhelloXX", 8, 5, 2)  # writes "hello" (5 bytes of the source starting at 2)
b.copy(b, 20, 5, 8)                 # copies 5 bytes from offset 8 to offset 20
```

## Actual Behavior

```
ArgumentError: wrong number of arguments (given 4, expected 1..3)
  set_string at org/jruby/RubyIOBuffer.java:1481
```

Both `#copy` and `#set_string` fail the same way.

## Analysis

The 4-argument support is already almost entirely present in `RubyIOBuffer.java`.
The catch-all overloads are declared `@JRubyMethod(name = "copy", required = 1, optional = 3, checkArity = false)`, their `switch` has a `case 4:` branch, and the 4-argument overloads and the `sourceOffset`-aware internals all exist, but the explicit `Arity.checkArgumentCount(context, args, 1, 3)` runs first, so `case 4:` is unreachable.

This also explains two entries in `test/mri/excludes/TestIOBuffer.rb`:
`test_copy_overlapped_bwd` and `test_set_string_overlapped_bwd` are excluded as `"needs investigation"`, and they are exactly the only cases in `TestIOBuffer` that use the 4-argument form (added in ruby/ruby#11640 for [Bug #20745](https://bugs.ruby-lang.org/issues/20745)).

Raising the arity limit exposes one more latent bug on the previously unreachable path: the internal guard compares `sourceOffset > length`, while MRI’s `io_buffer_copy_from` compares the source offset against the source size.
So `b.copy(src, 20, 5, 8)` (a 5-byte copy from source offset 8) raises “The given source offset is bigger than the source itself!” even when the source is large enough.

makenowjust:
Fix #9587

The catch-all overloads for `copy` and `set_string` declare `optional = 3` and have a `case 4:` branch backed by complete 4-argument implementations, but the explicit `Arity.checkArgumentCount(context, args, 1, 3)` made that branch unreachable, so the fourth argument (`source_offset`, accepted by CRuby since 3.1.0) raised `ArgumentError`.

Raising the checks to 4 exposed a latent bug on the previously unreachable path: the internal guards compared the source offset against the copy length, where MRI's `io_buffer_copy_from` compares it against the source size.

This re-enables `test_copy_overlapped_bwd` and `test_set_string_overlapped_bwd` (excluded as `"needs investigation"`): they are the only `TestIOBuffer` test cases that use the 4-argument form.
With excludes applied, `ruby/test_io_buffer.rb` runs 57 tests with 0 failures; without excludes, the failure set is unchanged apart from the two re-enabled tests.


test/mri/excludes/TestIOBuffer.rb
```diff
@@ -13,6 +12,5 @@
 exclude :test_read_with_length_and_offset, "read of zero is supposed to read as much as it can"
 exclude :test_resize_zero_external, "missing AccessError"
 exclude :test_set_string_null_destination, "needs investigation"
```
makenowjust, line 14: I guess this message `"needs investigation"` should be `"no support for null buffer"` instead, matching `:test_copy_null_destination` and `:test_copy_null_source` cases.

However, it is unrelated to the issue, so the change is not included in this pull request.

headius: Yes, we use "needs investigation" when we don't have an explanation, but it should be replaced when we know the actual cause. I can make that update separately. Thanks for pointing it out!

headius (approved):
Fix looks correct, thank you.

PR #9592, opened 2026-08-17
Scan the code range when folding a compound string

### LINKED ISSUE #9591: Compiler gives a constant-folded ASCII String the code range CR_VALID, which breaks the `eql?` / `hash` contract
The compiler folds some compound strings into a single literal. When it does,
it assigns the code range `CR_VALID` without a scan. If the text holds only
ASCII, the correct code range is `CR_7BIT`.

`String#hash` mixes the encoding into the digest when a String is not 7-bit.
Two byte-identical Strings with the same encoding then compare equal but hash
differently. `Array#uniq`, `Hash` and `Set` treat them as two values.

The interpreter builds the same string at run time and scans it, so it gets
`CR_7BIT`. The two execution modes therefore disagree. The bug disappears with
`-Xcompile.mode=OFF`.

**Environment Information**

* JRuby version:

  ```
  jruby 10.0.6.0 (3.4.5) 2026-06-11 716ad51f54 OpenJDK 64-Bit Server VM 21.0.11+10-LTS on 21.0.11+10-LTS +indy +jit [arm64-darwin]
  ```

  Also reproduced on:

  ```
  jruby 10.1.1.0 (4.0.0) 2026-07-22 aaa95d57e1 OpenJDK 64-Bit Server VM 21.0.11+10-LTS on 21.0.11+10-LTS +indy +jit [arm64-darwin]
  ```

* Command line: `jruby issue_repro.rb`, with no flags and no `JRUBY_OPTS` set.
* Operating system and platform:

  ```
  Darwin 25.6.0 Darwin Kernel Version 25.6.0: Sat Jul 11 15:27:26 PDT 2026; root:xnu-12377.161.13~4/RELEASE_ARM64_T6031 arm64
  ```

* JDK: Temurin OpenJDK 21.0.11+10-LTS.
* Gems: none. The scripts use only core Ruby.
* The scripts were run on macOS arm64. The downstream failure that led to this
  report also occurs on Linux x86_64 in GitHub Actions, with JRuby 10.0.6.0 on
  JDK 21.0.11 and 21.0.12. See "Origin" below.

**Expected Behavior**

Each pair below holds the same bytes and the same encoding, and every byte is
ASCII. Both members must report `ascii_only? == true`. `String#hash` must agree
with `String#eql?`, so `uniq` must return one element.

Script (`issue_repro.rb`):

```ruby
# frozen_string_literal: true

HEREDOC = <<~EOS.chomp
  type Widget {
    id: ID
  }
EOS

LITERAL = "type Widget {\n  id: ID\n}"

FOLDED = "abc#{1}def"
PLAIN = "abc1def"

puts RUBY_DESCRIPTION
puts

def report(label, made, expected)
  puts "#{label}:"
  puts "  ascii_only?  #{made.ascii_only?}  (expected #{expected.ascii_only?})"
  puts "  same bytes?  #{made.bytes == expected.bytes}"
  puts "  ==           #{made == expected}"
  puts "  eql?         #{made.eql?(expected)}"
  puts "  hash equal?  #{made.hash == expected.hash}"
  puts "  uniq.size    #{[made, expected].uniq.size}"
  puts
end

report("heredoc vs literal", HEREDOC, LITERAL)
report('"abc#{1}def" vs "abc1def"', FOLDED, PLAIN)
```

CRuby gives the expected result:

```
$ ruby issue_repro.rb
ruby 4.0.0 (2025-12-25 revision 553f1675f3) +PRISM [arm64-darwin25]

heredoc vs literal:
  ascii_only?  true  (expected true)
  same bytes?  true
  ==           true
  eql?         true
  hash equal?  true
  uniq.size    1

"abc#{1}def" vs "abc1def":
  ascii_only?  true  (expected true)
  same bytes?  true
  ==           true
  eql?         true
  hash equal?  true
  uniq.size    1
```

JRuby with the compiler disabled also gives the expected result:

```
$ JRUBY_OPTS=-Xcompile.mode=OFF jruby issue_repro.rb
jruby 10.0.6.0 (3.4.5) ... +indy -jit [arm64-darwin]

heredoc vs literal:
  ascii_only?  true  (expected true)
  ...
  uniq.size    1

"abc#{1}def" vs "abc1def":
  ascii_only?  true  (expected true)
  ...
  uniq.size    1
```

**Actual Behavior**

JRuby 10.0.6.0, default settings:

```
$ jruby issue_repro.rb
jruby 10.0.6.0 (3.4.5) 2026-06-11 716ad51f54 OpenJDK 64-Bit Server VM 21.0.11+10-LTS on 21.0.11+10-LTS +indy +jit [arm64-darwin]

heredoc vs literal:
  ascii_only?  false  (expected true)
  same bytes?  true
  ==           true
  eql?         true
  hash equal?  false
  uniq.size    2

"abc#{1}def" vs "abc1def":
  ascii_only?  false  (expected true)
  same bytes?  true
  ==           true
  eql?         true
  hash equal?  false
  uniq.size    2
```

JRuby 10.1.1.0, default settings, gives the same output.

Two problems appear:

1. `ascii_only?` is `false`, although every byte is below 128.
2. `eql?` is `true` but `hash` differs. `uniq` therefore returns 2.

`Hash` and `Set` behave the same way. A two-key Hash and a two-element Set are
built from Strings that are `eql?`.

**Which forms are affected**

The trigger is a compound string that the compiler folds into one piece, in a
mutable or chilled context. The table shows the code range of each result.
`-Xcompile.mode=FORCE` makes the compiler run for every method.

| Expression | interpreter | compiler |
|---|---|---|
| `<<~` heredoc, body indented | CR_7BIT | **CR_VALID** |
| `"abc#{1}def"` | CR_7BIT | **CR_VALID** |
| `"abc#{:sym}def"` | CR_7BIT | **CR_VALID** |
| `"abc#{1.5}def"` | CR_7BIT | **CR_VALID** |
| `"abc#{"X"}def"` | CR_7BIT | CR_7BIT |
| `"abc" "def"` | CR_7BIT | CR_7BIT |
| `"abc#{[1].size}def"` | CR_7BIT | CR_7BIT |
| `"abcdef"` | CR_7BIT | CR_7BIT |
| `<<-` heredoc, body indented | CR_7BIT | CR_7BIT |
| `<<~` heredoc, body at column 0 | CR_7BIT | CR_7BIT |

Notes on the table:

* A squiggly heredoc is affected only when the dedent removes characters. The
  dedent makes the parser produce a compound string.
* An interpolated value that is an immutable literal but not a String is
  affected. A Symbol, an Integer or a Float all fold this way.
* An interpolated String is not affected, because the parser merges the pieces
  earlier and rescans the result.
* `# frozen_string_literal: true` is not required. Interpolated strings are
  never frozen by the magic comment, so they always take the affected branch.

**Extra: the result can change during one process**

With the default compile mode the first calls run in the interpreter and give
the correct code range. The result then flips when the method is compiled:

```
$ jruby jit_flip.rb 3000
jruby 10.0.6.0 (3.4.5) ... +indy +jit [arm64-darwin]
calls: 3000

  7bit x237 -> NOT7bit x2763
```

The flip point follows `-Xjit.threshold`. It also varies between runs, because
compilation is asynchronous. Two runs of 600 calls each:

| `-Xjit.threshold` | first wrong call |
|---|---|
| 10 | 138, 127 |
| 50 (default) | 309, 215 |
| 200 | none in 600 calls, 531 |
| 500 | none in 600 calls, none in 600 calls |

One process can therefore hold both a correct and an incorrect copy of the same
text. Any hash-based comparison across that boundary fails.

<details>
<summary>Source of <code>jit_flip.rb</code> and <code>doc_builder.rb</code></summary>

The heredoc must sit in a required file to show the flip.

`doc_builder.rb`:

```ruby
# frozen_string_literal: true

module DocBuilder
  module_function

  def with_field
    yield self
  end

  def documented_field
    with_field do |field|
      field.format_documentation(<<~EOS)
        Matches records where any of the provided sub-filters evaluate to true.
        This works just like an OR operator in SQL.

        When `null` is passed, matches all documents.
      EOS
    end
  end

  def format_documentation(comment)
    %("""\n#{comment.chomp}\n"""\n)
  end
end
```

`jit_flip.rb`:

```ruby
# frozen_string_literal: true

require_relative "doc_builder"

N = Integer(ARGV[0] || 3_000)

states = Array.new(N) { DocBuilder.documented_field.ascii_only? ? "7bit" : "NOT7bit" }
runs = states.chunk_while { |a, b| a == b }.map { |chunk| "#{chunk.first} x#{chunk.size}" }

puts RUBY_DESCRIPTION
puts "JRUBY_OPTS=#{ENV["JRUBY_OPTS"].inspect}"
puts "calls: #{N}"
puts
puts "  #{runs.join(" -> ")}"

first_wrong = states.index("NOT7bit")
puts
if first_wrong
  puts "RESULT: bug reproduced -- code range flipped at call #{first_wrong}"
else
  puts "RESULT: no bug on this runtime"
end
exit(first_wrong ? 1 : 0)
```

</details>

**Root cause**

`BuildCompoundStringInstr.simplifyInstr` folds a compound string into a single
literal. Two places then hardcode `CR_VALID`.

`copy` is the one that causes the reports above:

https://github.com/jruby/jruby/blob/10.0.6.0/core/src/main/java/org/jruby/ir/instructions/BuildCompoundStringInstr.java#L240-L248

```java
private Instr copy(FrozenString string) {
    Operand value = switch (stringStyle) {
        case Frozen -> string;
        case Mutable -> new MutableString(string.bytelist, CR_VALID, file, line);
        default -> new ChilledString(string.bytelist, CR_VALID, file, line);
    };

    return new CopyInstr(result, value);
}
```

The `Frozen` branch is correct, because it reuses the operand and its code
range. The other two branches replace a known code range with `CR_VALID`.

The empty-string branch has the same problem. An empty String is 7-bit:

https://github.com/jruby/jruby/blob/10.0.6.0/core/src/main/java/org/jruby/ir/instructions/BuildCompoundStringInstr.java#L153-L161

The interpreter does not use this path. `interpret` appends each piece with
`RubyString#cat`, which keeps the code range correct. That is why the two modes
disagree.

`asOperand` in the same file already uses `CR_UNKNOWN`, which is safe, because
a later scan then computes the true value.

**Origin**

This causes a rare CI failure in
[block/elasticgraph](https://github.com/block/elasticgraph). A spec builds the
same GraphQL schema six times and asserts that all six strings are identical:

```
Failure/Error: expect(uniq_results_for_each_ordering.size).to eq 1
  expected: 1
       got: 2
```

The six strings are byte-identical. The JIT threshold falls inside the example,
so the first three builds get a correct code range and the last three do not.
`Array#join` then reports `US-ASCII` for the first three and `UTF-8` for the
last three, and `Array#uniq` keeps both. The failure appeared three times in
recent weeks, on three different RSpec seeds, and only in the JRuby CI job.

jwils:
Fixes #9591.

`BuildCompoundStringInstr#simplifyInstr` folds a compound string into a single
literal. Two places assigned `CR_VALID` without a scan of the bytes. A string
that holds only ASCII must be `CR_7BIT`.

`String#hash` mixes the encoding into the digest when a string is not 7-bit.
Two byte-identical strings with the same encoding therefore compared equal but
hashed differently, and `Array#uniq`, `Hash` and `Set` treated them as two
values.

The interpreter is not affected. `interpret` appends each piece with
`RubyString#cat`, which keeps the code range correct. The two modes disagreed,
so one process could hold both a correct and an incorrect copy of the same
text once the JIT threshold was crossed part way through a run.

## The change

* `copy`: keep the code range of the operand that is folded. The `Frozen`
  branch was already correct, because it reuses the operand and its code range.
* the empty-string branch: use `CR_7BIT`. An empty string is 7-bit.

`asOperand` in the same file already uses `CR_UNKNOWN`, which is safe, because
a later scan computes the true value.

## Which expressions were affected

Code range of the result, with `-Xcompile.mode=FORCE`:

| Expression | before | after |
|---|---|---|
| `"abc#{1}def"` | CR_VALID | CR_7BIT |
| `"abc#{:sym}def"` | CR_VALID | CR_7BIT |
| `"abc#{1.5}def"` | CR_VALID | CR_7BIT |
| `<<~` heredoc with an indented body | CR_VALID | CR_7BIT |
| `"café#{1}!"` | CR_VALID | CR_VALID |

The dedent of a squiggly heredoc makes the parser produce a compound string,
which is why only the indented form was affected. An interpolated String was
not affected, because the parser merges those pieces earlier and rescans.

## Tests

`test/jruby/test_compound_string_coderange.rb` is new. The bug only appears
once a method is compiled, so each case runs in a child process with
`jruby.compile.mode=FORCE`. Without this change 5 of the 6 tests fail. The
sixth checks that real non-ASCII text still gets `CR_VALID`, and it passes
either way.

I also ran `spec/ruby/core/string`, `spec/ruby/core/symbol`,
`spec/ruby/core/encoding`, `spec/ruby/language/string_spec.rb` and
`spec/ruby/language/heredoc_spec.rb` with `-Xcompile.mode=FORCE`. The result is
the same before and after: 4862 examples, 50 failures, 8 errors. Those failures
are already present on master.

### DIFF
--- a/core/src/main/java/org/jruby/ir/instructions/BuildCompoundStringInstr.java
+++ b/core/src/main/java/org/jruby/ir/instructions/BuildCompoundStringInstr.java
@@ -154,9 +154,9 @@ public class BuildCompoundStringInstr extends NOperandResultBaseInstr {
             var newByteList = new ByteList(0);
             newByteList.setEncoding(encoding);
             Operand string = switch (stringStyle) {
-                case Frozen -> new FrozenString(newByteList, CR_VALID, file, line);
-                case Mutable -> new MutableString(newByteList, CR_VALID, file, line);
-                default -> new ChilledString(newByteList, CR_VALID, file, line);
+                case Frozen -> new FrozenString(newByteList, CR_7BIT, file, line);
+                case Mutable -> new MutableString(newByteList, CR_7BIT, file, line);
+                default -> new ChilledString(newByteList, CR_7BIT, file, line);
             };
             return new CopyInstr(getResult(), string);
         } else if (piecesArray.length == 1) { // not sure we can have a compound string with only one piece AND a non-string.
@@ -240,8 +240,8 @@ public class BuildCompoundStringInstr extends NOperandResultBaseInstr {
     private Instr copy(FrozenString string) {
         Operand value = switch (stringStyle) {
             case Frozen -> string;
-            case Mutable -> new MutableString(string.bytelist, CR_VALID, file, line);
-            default -> new ChilledString(string.bytelist, CR_VALID, file, line);
+            case Mutable -> new MutableString(string.bytelist, string.getCodeRange(), file, line);
+            default -> new ChilledString(string.bytelist, string.getCodeRange(), file, line);
         };
 
         return new CopyInstr(result, value);
--- a/test/jruby.index
+++ b/test/jruby.index
@@ -17,6 +17,7 @@ jruby/test_case
 jruby/test_class
 jruby/test_class_subclasses
 jruby/test_comparable
+jruby/test_compound_string_coderange
 jruby/test_core_arities
 jruby/test_coverage
 jruby/test_custom_enumerable
--- a/test/jruby/test_compound_string_coderange.rb
+++ b/test/jruby/test_compound_string_coderange.rb
@@ -1,0 +1,59 @@
+# encoding: UTF-8
+require 'test/unit'
+require 'test/jruby/test_helper'
+
+# GH-9591
+#
+# BuildCompoundStringInstr#simplifyInstr folds a compound string into a single
+# literal. It used to assign CR_VALID without scanning the bytes. An ASCII-only
+# string must be CR_7BIT, otherwise String#hash disagrees with String#eql? and
+# uniq/Hash/Set split byte-identical strings.
+#
+# The interpreter builds these strings at run time and scans them, so the bug
+# only appeared once a method was compiled. Each test therefore runs in a child
+# process with the compiler forced on.
+class TestCompoundStringCodeRange < Test::Unit::TestCase
+  include TestHelper
+
+  FORCE_COMPILE = { "jruby.compile.mode" => "FORCE" }
+
+  def test_folded_integer_interpolation_is_ascii_only
+    assert_equal 'true', run_compiled('print "abc#{1}def".ascii_only?')
+  end
+
+  def test_folded_symbol_interpolation_is_ascii_only
+    assert_equal 'true', run_compiled('print "abc#{:sym}def".ascii_only?')
+  end
+
+  def test_folded_float_interpolation_is_ascii_only
+    assert_equal 'true', run_compiled('print "abc#{1.5}def".ascii_only?')
+  end
+
+  def test_dedented_heredoc_is_ascii_only
+    script = 'value = <<~EOS' "\n" \
+             '  hello world' "\n" \
+             'EOS' "\n" \
+             'print value.ascii_only?'
+    assert_equal 'true', run_compiled(script)
+  end
+
+  def test_folded_string_hash_agrees_with_eql
+    script = 'folded = "abc#{1}def"' "\n" \
+             'plain = "abc1def"' "\n" \
+             'print [folded.eql?(plain), folded.hash == plain.hash, [folded, plain].uniq.size].inspect'
+    assert_equal '[true, true, 1]', run_compiled(script)
+  end
+
+  # The fix must not mark real non-ASCII text as 7-bit.
+  def test_folded_non_ascii_string_is_not_ascii_only
+    script = 'value = "café#{1}!"' "\n" \
+             'print [value.ascii_only?, value.valid_encoding?].inspect'
+    assert_equal '[false, true]', run_compiled(script)
+  end
+
+  private
+
+  def run_compiled(script)
+    with_temp_script(script) { |f| jruby(f.path, FORCE_COMPILE) }
+  end
+end


headius (approved):
The fix makes sense. Normally I'd also prefer that the tests go into the RubySpec suite (spec/ruby), but I don't believe there's a good way to observe the code range from pure-Ruby.

We can go with this as-is and leave the test/jruby test in place.

headius:
I approved this PR, but it probably should be applied to JRuby 10.0 instead of master (10.1) as that is the current LTS and this is clearly an observable behavior difference from standard Ruby.

@jwils Could you rebase this fix atop the `jruby-10.0` branch?

jwils:
Thanks for the quick review. Rebased.

headius (approved):
Approved again. Might revisit the tests later to see if they make sense for rubyspec.

PR #9594, opened 2026-08-17
More accurate TTY checking for IO

### LINKED ISSUE #9590: `IO#tty?` silently returns `true` for redirected stdio in embedded runtimes when `--add-opens` are missing
## Summary

Currently on POSIX (not sure about Windows), when `java.io` and `sun.nio.ch` are not opened, JRuby appears to assume that everything is a tty.

This causes some issues, particularly with Bundler on Java 21+ where there is specific handling of failures when `tty?` is true to actually do `exit`/`System.exit()`, [e.g here](https://github.com/ruby/rubygems/blob/b463ced1459e03bf871116fff3a09c5d287806c2/bundler/lib/bundler/setup.rb#L17-L30). This then obscures the actual bundler error that caused it, due to the `tty?` mis-detection.

In an **embedded** JRuby runtime (e.g. `ScriptingContainer` as used by something like jruby-rack inside a servlet container), `STDOUT.tty?` / `STDERR.tty?` return `true` even when the process' stdio is redirected to a file or `/dev/null`, unless the JVM was started with:

```
--add-opens java.base/java.io=ALL-UNNAMED --add-opens java.base/sun.nio.ch=ALL-UNNAMED
```

Both opens are required. No warning seems to be logged when the reflective path fails — JRuby silently falls back to "assume standard stdio is a tty" (although there may be reasons to avoid logging and putting more "stuff" on the output...)

The JRuby launcher passes these opens automatically (via `bin/.jruby.module_opts`), so plain CLI usage is unaffected. Embedding containers (Tomcat, Jetty, …) do not pass them. Tomcat's stock `catalina.sh` opens `java.base/java.io` (for its own code) but **not** `java.base/sun.nio.ch`, so JRuby webapps on Tomcat get the wrong answer by default.

## Environment

- JRuby 10.0.6.0
- OpenJDK Temurin 21 and 26, macOS 15 (arm64)
- Observed in the wild under Apache Tomcat 11.0.24 via jruby-rack, with stdout/stderr redirected to `/dev/null` (brew services / launchd)

## Reproduction (plain JVM, no launcher — mirrors a servlet container)

```sh
JRUBY_JAR="$(mise where ruby@jruby-10.0)/lib/jruby.jar"   # or $JRUBY_HOME

SNIPPET='sc = org.jruby.embed.ScriptingContainer.new(org.jruby.embed.LocalContextScope::THREADSAFE)
p sc.run_scriptlet("STDERR.tty?")'

java -cp "$JRUBY_JAR" org.jruby.Main -e "$SNIPPET" 2>/dev/null
# => true   (WRONG - stderr NOT a terminal -> expected false)

java --add-opens java.base/java.io=ALL-UNNAMED --add-opens java.base/sun.nio.ch=ALL-UNNAMED \
     -cp "$JRUBY_JAR" org.jruby.Main -e "$SNIPPET" 2>/dev/null
# => false  (correct)
```

Alternatively, you can reproduce with the jruby launcher, by temporarily editing or removing `bin/.jruby.module_opts`:

```sh
mv "$JRUBY_HOME/bin/.jruby.module_opts" "$JRUBY_HOME/bin/.jruby.module_opts.bak"
ruby -e "$SNIPPET" 2>/dev/null
mv "$JRUBY_HOME/bin/.jruby.module_opts.bak" "$JRUBY_HOME/bin/.jruby.module_opts"
```

## Claude Analysis

Note: the naive one-liner `ruby -e 'p STDERR.tty?' 2>/dev/null` does **not** reproduce, even with the opens stripped. The top-level CLI runtime wires stdio to native device channels with hardcoded filenos 0/1/2, so no reflection is needed. Only *embedded* runtimes wrap the `System.out`/`System.err` `PrintStream`s in channels whose real fd must be recovered reflectively — hence the `ScriptingContainer` (fresh runtime, `LocalContextScope::THREADSAFE`) in the repro. The default `ScriptingContainer.new` (singleton scope) also doesn't reproduce from the CLI, because it reuses the CLI's already-wired runtime.

## Mechanism

1. For an embedded runtime, stdio channels are built from the configured `System.out`/`System.err` streams; the real file descriptor is recovered via reflection (`FilenoUtil.getFilenoUsingReflection` → `sun.nio.ch.*` internals / `java.io.FileDescriptor.fd`).
2. Without the two opens, `setAccessible` fails, `ChannelFD.realFileno` stays `-1`, and `IO#tty?` cannot use native `isatty(3)`.
3. The fallback (jnr-posix `JavaPOSIX` path) reports the standard descriptors as ttys unconditionally — even when `System.console()` is `null` and fd 1/2 point at `/dev/null` (verified: `-Djruby.native.enabled=false` flips a redirected `STDOUT.tty?` from `false` to `true`).
4. Nothing is logged, so the misdetection is invisible until something behaves differently on a "tty" — e.g. Bundler 2.7's `exit` above.

## Suggestions

Claude thinks any of these might help:

- When the real fileno for a stdio stream is unknown, default `tty?` to **false** rather than true — "unknown" is far more likely to be a redirected/embedded stream than an interactive terminal (CRuby also returns false when in doubt).
- Log a warning (once) when stdio fd reflection fails due to module access, naming the required `--add-opens`, as is already done for other JPMS-restricted paths.

headius:
This introduces some cleanup and improvements in how we determine whether an IO is a TTY (`IO#tty?` method):

* If the IO has a proper native file descriptor and we can call `isatty`, use that result.
* Otherwise, if the IO was set up as user-facing stdio AND the current JRuby was started via `Main`, use the JDK's `System.console()` and `Console.isTerminal()`.
* Otherwise, treat the IO as a non-terminal.

This should address situations where JRuby has been embedded as part of a larger application with user-facing stdio streams connected to non-terminal channels, such as in #9590 (deployed in Tomcat using jruby-rack).

Draft until appropriate tests can be added.

### DIFF
--- a/core/src/main/java/org/jruby/RubyGlobal.java
+++ b/core/src/main/java/org/jruby/RubyGlobal.java
@@ -330,7 +330,7 @@ public class RubyGlobal {
         // If we're the main for the process and native stdio is enabled, use default descriptors
         if (!Platform.IS_WINDOWS && // Windows does not do native IO yet
                 runtime.getPosix().isNative() &&
-                runtime.getInstanceConfig().isHardExit() && // main JRuby only
+                runtime.getInstanceConfig().isMain() && // main JRuby only
                 Options.NATIVE_STDIO.load()) {
             stdin = RubyIO.prepStdio(
                     runtime, runtime.getIn(), new NativeDeviceChannel(0), OpenFile.READABLE, runtime.getIO(), "<STDIN>");
--- a/core/src/main/java/org/jruby/RubyIO.java
+++ b/core/src/main/java/org/jruby/RubyIO.java
@@ -66,6 +66,7 @@ import org.jcodings.Encoding;
 import org.jcodings.specific.ASCIIEncoding;
 import org.jcodings.transcode.EConvFlags;
 import org.jruby.api.API;
+import org.jruby.api.Convert;
 import org.jruby.ast.util.ArgsUtil;
 import org.jruby.anno.JRubyMethod;
 import org.jruby.anno.JRubyClass;
@@ -89,6 +90,7 @@ import org.jruby.runtime.callsite.CachingCallSite;
 import org.jruby.runtime.encoding.EncodingService;
 import org.jruby.util.ShellLauncher.POpenProcess;
 import org.jruby.util.*;
+import org.jruby.util.cli.JVMConsole;
 import org.jruby.util.cli.Options;
 import org.jruby.util.io.ChannelFD;
 import org.jruby.util.io.EncodingUtils;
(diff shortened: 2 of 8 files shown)


headius:
See also https://github.com/jruby/jruby/issues/8934 for previous attempts to deal with JDK 22 always returning non-null `System.console()`.

PR #9603, opened 2026-08-21
Implement :fiber_switch trace event

### LINKED ISSUE #9600: TracePoint does not support `fiber_switch` (Ruby 2.3+ compatibility)
**Environment Information**

```console
$ mise x ruby@jruby -- ruby -v 
jruby 10.1.0.0 (4.0.0) 2026-04-20 32f988b78c OpenJDK 64-Bit Server VM 26.0.2+10-55 on 26.0.2+10-55 +indy +jit [arm64-darwin]
$ uname -a
Darwin Konstantins-MBP.localdomain 25.5.0 Darwin Kernel Version 25.5.0: Tue Jun  9 22:28:34 PDT 2026; root:xnu-12377.121.10~1/RELEASE_ARM64_T6041 arm64
```

No other dependencies are needed to run into this issue.

I found this by tracing which thread a fiber scheduler fiber belongs to.

**Expected Behavior**

Should trace `fiber_switch` events (introduced in Ruby 2.3.0)

```ruby
triggered = false

TracePoint.trace(:fiber_switch) do |tp|
  triggered = true
end

Fiber.new { p triggered }.resume
```

Should print `true`

**Actual Behavior**

Raises an exception:

```
ArgumentError: unknown event: fiber_switch
  initialize at org/jruby/ext/tracepoint/TracePoint.java:65
         new at org/jruby/RubyClass.java:1058
       trace at org/jruby/ext/tracepoint/TracePoint.java:282
      <main> at play/trace.rb:3
```

headius:
This is mostly working, but does not yet pass CRuby's test_fiber_switch in test_setttracefunc.rb for a few reasons:

* The current thread check there is incompatible with JRuby's Thread.current being different for each Fiber.
* Raise events are not being triggered from Fiber.raise or Fiber errors.

As usual, for accurate file and line numbers, it should be run with --debug.

Might be enough for jruby/jruby#9600.

### DIFF
--- a/core/src/main/java/org/jruby/ext/fiber/ThreadFiber.java
+++ b/core/src/main/java/org/jruby/ext/fiber/ThreadFiber.java
@@ -23,6 +23,7 @@ import org.jruby.javasupport.JavaUtil;
 import org.jruby.runtime.Block;
 import org.jruby.runtime.ExecutionContext;
 import org.jruby.runtime.Helpers;
+import org.jruby.runtime.RubyEvent;
 import org.jruby.runtime.ThreadContext;
 import org.jruby.runtime.Visibility;
 import org.jruby.runtime.builtin.IRubyObject;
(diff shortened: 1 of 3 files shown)


headius:
Rebased and still does not pass the relevant test due to `Thread.current` behavior, but I think this would be fine to merge any time.

PR #9604, opened 2026-08-22
Fix Fiber.schedule inside a non-blocking Fiber

sampokuokkanen:
Fiber.schedule looked up the scheduler on the Fiber's carrier thread rather than the Thread that owns it, so calling it from inside a non-blocking Fiber raised "No scheduler is available!". Use getFiberCurrentThread(), like Fiber.scheduler and MRI's rb_fiber_s_schedule.

Specs come from ruby/spec#1388 (merged) and three of the six fail without this. One is tagged for a separate blocking-count bug (passes when run in isolation but fails when run with other fiber tests).

On master the following tests fail with  `RuntimeError: No scheduler is available!`:

- Fiber.schedule when a scheduler is set can be called from inside a non-blocking Fiber
- Fiber.schedule when a scheduler is set uses the scheduler of the Thread owning the Fiber it is called from
- Fiber.schedule when a scheduler is set runs the block in a Fiber which sees the scheduler as its current scheduler

(The last one is the one that fails on this branch too when run with other fiber tests but passes when you just run `spec/ruby/core/fiber/schedule_spec.rb`)

### DIFF
--- a/core/src/main/java/org/jruby/ext/fiber/ThreadFiber.java
+++ b/core/src/main/java/org/jruby/ext/fiber/ThreadFiber.java
@@ -810,7 +810,7 @@ public class ThreadFiber extends RubyObject implements ExecutionContext {
         // MRI: rb_fiber_s_schedule_kw and rb_fiber_s_schedule, kw passes on context
         @JRubyMethod(name = "schedule", meta = true, rest = true, keywords = true)
         public static IRubyObject schedule(ThreadContext context, IRubyObject self, IRubyObject[] args, Block block) {
-            IRubyObject scheduler = context.getThread().getScheduler();
+            IRubyObject scheduler = context.getFiberCurrentThread().getScheduler();
             if (scheduler.isNil()) throw runtimeError(context, "No scheduler is available!");
             return scheduler.callMethod(context, "fiber", args, block);
         }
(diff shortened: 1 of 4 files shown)


headius:
Nice simple fix and I appreciate the extra work getting the specs in place! The specs are periodically merged so I'd like to hear from @eregon or @andrykonchin that it's okay for us to also pull them in this way. 

The failures are not your fault and are due to a bad merge of an unrelated fix.

headius:
The bad spec has been fixed on master; you can rebase if you get a chance, but it will be fine to merge either way (assuming the spec updates are ok).

sampokuokkanen:
Thanks, I see that you already merged master, do you still want me to rebase? 
@eregon told me that this (on truffleruby side [here](https://github.com/truffleruby/truffleruby/pull/4272#issuecomment-4568206979)) is the correct way of doing things, since if the test files are identical the git merge will not conflict and we get the tests passing right away. 

Most of the tests are passing now too, great! Thanks about that too.

eregon:
Yep, just to confirm it's fine to copy files over if they don't conflict, i.e. if they are copied as-is in a commit.
There can even be further tweaks to the specs here, just needs to be a different commit so it applies cleanly.

headius:
@sampokuokkanen Nope we're all good here! 

@eregon Thanks for clarifying!

Issue #9609, opened 2026-08-28
No validation of unsetenv_others for Process.exec causes specs to execve() abruptly


eregon:
For this spec:
```ruby
-> { Process.exec("true", unsetenv_others: 1) }.should.raise(ArgumentError, /expected true or false/)
```
JRuby doesn't raise and therefore execve()s to `true`, ending the spec process successfully.

There are two parts:
* That spec should be improved to handle that better: https://github.com/ruby/spec/commit/8301a5880149cb529a61207bb262a82a8d77a4b4#r198061669
* JRuby should raise (this issue), 

Also see https://github.com/eregon/rubyspec-stats/issues/6

This likely cause only a subset of specs to run in JRuby CI, so feels quite urgent to fix.

eregon:
The part about execve()'ing in the middle of specs is fixed by https://github.com/jruby/jruby/pull/9615.
So it's just the compatibility issue now, which is pretty minor, closing this.

headius:
The compatibility issue should be dealt with anyway.

headius:
Behavior appears to have changed in CRuby in 2019, for 2.7:

https://github.com/ruby/ruby/commit/d8e23a67c3869c70ef82b4337499367632ce911a

headius:
Part of work completed in #9626 but the actual case represented here is via Process.exec, which is currently not hooked up to the ported process logic. I'm working on a separate PR for that now.

headius:
This turned out to be a larger piece of work than I expected, since neither `Kernel.exec` nor `Process.exec` were hooked up to the ported CRuby popen logic, but it should be resolved shortly by #9627.

PR #9615, opened 2026-08-28
Fix :core specs run to not execve() in the middle

eregon:
* This avoids the spec process execve()'ing in the middle, as described in https://github.com/jruby/jruby/issues/9609


eregon:
Should be green now with the tags

PR #9617, opened 2026-08-30
Increase timeout affordance for thread tests

jcharaoui:
This sets the delay for two timeout-based tests from 10 to 250 milliseconds to avoid spurious test failures on slow/busy hosts


core/src/test/java/org/jruby/test/TestRubyThread.java
```diff
@@ -47,7 +47,7 @@ public void testJavaErrorDoesPropagate() throws InterruptedException {
 
         context.runtime.evalScriptlet("$run_thread = true");
 
-        Thread.sleep(100);
+        Thread.sleep(2500);
```
chadlwilson:
```suggestion
        Thread.sleep(250);
```

2.5s is quite a long time when it's a forced sleep. Would .25 be enough?

jcharaoui: My mistake, I was aiming to 250ms not 2.5s. I'll run another batch of test to check if 250ms is sufficient.

PR #9557, opened 2026-08-03
refactor and improve reification of java subclasses

kares:
The work here is an attempt to improve performance for Java sub-classing in Ruby.
Since 9.3 such sub-classes always generate a Java class upon the constructor invocation.

The [performance degradation](https://github.com/jruby/jruby/issues/7763) comes from the need to support the constructor "split-state" as there might be Ruby code before and after the `super` (Java) constructor invocation, while this is possible the common use-case is usually simple `super(...)` - there's an attempt to detect these and skip extra work when possible (**improving performance ~5x**).

A custom `JavaCtorSuperInst` instruction is added, which simplifies the verbosity a bit. 
The Java `super` call is cached as an indy call-site.

The `ExitableInterpreterContext` tries to cache some common forms of Ruby `initialize` (on a Java class) that don't require the full interpreter context.

There's even the *weird* Java `ctor` configuration support (which I am not sure is really used or even documented beyond just the jrubyfx), which would be nice to get rid of, for now it's left as is.

### Numbers (i/s, higher is better)

| workload (`.new`)                         | 9.4.14  | 10.0.6  |  10.1.0 | 10.1 PR | |
|-------------------------------------------|--------:|--------:|--------:|--------:|-------:|
| baseline `ArrayList.new` (no subclass)    | 9404476 |11069136 | 4183418 | 4266039 | 1.02x  |
| subclass, no `initialize`                 | 6207511 | 6281261 | 2670414 | 5335156 | 2.00x  |
| terminal literal `super(10)`              | 1439885 | 1578454 |  834816 | 4878712 | 5.84x  |
| direct `super()` no args                  | 1735299 | 1965477 |  966431 | 4773751 | 4.94x  |
| direct forward fixed `super(n)`           | 1467508 | 1573772 |  791602 | 3753935 | 4.74x  |
| direct forward rest `super(*args)`        | 1069462 | 1373900 |  718076 | 3779693 | 5.26x  |
| zsuper `(first, *rest)`                   | 1054923 | 1323442 |  677280 | 3771679 | 5.57x  |
| arbitrary pre-super work                  | 1180063 | 1435396 |  743467 | 1324021 | 1.78x  |
| post-super continuation                   | 1176290 | 1224408 |  705805 |  984608 | 1.40x  |
| Ruby < Ruby < Java chain                  |  821912 |  895147 |  526642 | 2283123 | 4.34x  |

---

*p.s. LLM assisted, test coverage is generated, (non-test) code is mixed mode and hand reviewed.*

PR #9619, opened 2026-08-30
refactor and improve reification of java subclasses (refactor only)

kares:
the *refactor only* parts from https://github.com/jruby/jruby/pull/9557 includes the addition of Java costructor super instruction, can be followed commit-by-commit if needed.

p.s. remaining part (performance improvement) look like https://github.com/kares/jruby/pull/7

### DIFF
--- a/core/src/main/java/org/jruby/RubyClass.java
+++ b/core/src/main/java/org/jruby/RubyClass.java
@@ -2466,7 +2466,7 @@ public class RubyClass extends RubyModule {
                 final int mod = constructor.getModifiers();
                 if (!Modifier.isPublic(mod) && !Modifier.isProtected(mod)) continue;
                 candidates.add(constructor);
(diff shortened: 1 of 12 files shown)


enebo:
@kares I think the superinstr is pretty clever way of handling this.  We would never inline a java ctor but it is possible to see it now.  Also no perf penalty in interpreter since is in a special interpreter.

kares:
yeah, felt like a simplification worth doing. 
existing code is hard to navigate, the idea around the Java super + interpreter split is understandable and simple but figuring out the bits from code not that much. :crying_cat_face: 

hopefully, we can get this going and also consider https://github.com/jruby/jruby/pull/9557 which should give up a slightly better base-line for experiments... :crossed_fingers:

headius (approved):
All looks pretty benign and cleans up some ugly stuff.

PR #9622, opened 2026-09-01
Defer Kernel#sleep to the Fiber scheduler

sampokuokkanen:
Kernel#sleep never consulted Fiber.scheduler, so a non-blocking Fiber that slept parked its thread instead of yielding. Follow MRI's rb_f_sleep: look up the current scheduler first and hand it the arguments untouched, nil included, before any interval conversion. FiberScheduler.kernelSleep was already written, this just wires it up.

Closing the scheduler at exit is also fixed. MRI's rb_ec_teardown sets the thread's scheduler to nil before at-exit handlers, which runs out the fibers still waiting on it; JRuby only did that for spawned threads, through RubyThread#dispose. Harmless until now, since nothing parked a fiber on the main thread. But with sleep deferring, a fiber sleeping there would never be resumed, skipping the rest of its body and its ensure blocks.

Passes the tests in https://github.com/ruby/spec/pull/1391, while master fails the new "closes the scheduler at exit" spec. 
Follow-up needed: wire `Thread::Queue#pop` and `Thread::Mutex#lock` to `FiberScheduler.block/unblock` and delete the test exclusions added in this PR.

### DIFF
--- a/core/src/main/java/org/jruby/FiberScheduler.java
+++ b/core/src/main/java/org/jruby/FiberScheduler.java
@@ -4,6 +4,7 @@ import jnr.constants.platform.Errno;
 import org.jruby.runtime.Helpers;
 import org.jruby.runtime.ThreadContext;
 import org.jruby.runtime.builtin.IRubyObject;
+import org.jruby.util.cli.Options;
 import org.jruby.util.io.OpenFile;
 
 import java.nio.ByteBuffer;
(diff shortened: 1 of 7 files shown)


headius (approved):
Yup, good catch. Since I originally implemented the thread scheduler stuff, more blocking operations have steadily been added. Good to get `sleep` checked off the list.

PR #9624, opened 2026-09-01
Fix out-of-bounds read building Time.new parse errors

### LINKED ISSUE #9623: Time.new(String) raises java.lang.ArrayIndexOutOfBoundsException when the string ends where a ':' is expected
**Environment Information**

```
jruby 10.0.6.0 (3.4.5) 2026-06-11 716ad51f54 OpenJDK 64-Bit Server VM 25.0.4+7-LTS on 25.0.4+7-LTS +indy +jit [arm64-darwin]
```

```
Darwin 25.5.0 arm64, Amazon Corretto 25.0.4+7-LTS
```

No gems, no framework, no `JRUBY_OPTS`. Plain `jruby -e`.

Also present by source inspection in 10.0.5.0, the `jruby-10.0` head (856ca57679), 10.1.x and
`master`: `core/src/main/java/org/jruby/util/RubyTimeParser.java` is identical across them apart
from `new RubyFixnum(...)` → `asFixnum(...)` at line 251.

**Expected Behavior**

`ArgumentError`, as CRuby 3.4 raises for the same strings:

```
Time.new("2020-12-25 00:56")   # ArgumentError: missing sec part: 00:56
Time.new("2020-12-25 00")      # ArgumentError: missing min part: 00
```

**Actual Behavior**

A raw Java exception escapes into Ruby:

```
$ jruby -e 'Time.new("2020-12-25 00:56")'
Java::JavaLang::ArrayIndexOutOfBoundsException: arraycopy: last source index 17 out of bounds for byte[16]
    java.base/java.lang.System.arraycopy(Native Method)
    org.jruby.dist/org.jruby.util.ByteList.<init>(ByteList.java:228)
    org.jruby.dist/org.jruby.util.ByteList.<init>(ByteList.java:207)
    org.jruby.dist/org.jruby.util.RubyTimeParser.substr(RubyTimeParser.java:162)
    org.jruby.dist/org.jruby.util.RubyTimeParser.needColon(RubyTimeParser.java:150)
    org.jruby.dist/org.jruby.util.RubyTimeParser.parse(RubyTimeParser.java:72)
    org.jruby.dist/org.jruby.RubyTime.timeInitParse(RubyTime.java:1910)
    org.jruby.dist/org.jruby.RubyTime.initialize(RubyTime.java:1888)
```

Three inputs reproduce it. The trigger is exactly "the string ends at the byte where a `:` was
expected":

| input | bytes | JRuby 10.0.6.0 | CRuby 3.4 |
|---|---|---|---|
| `"2020-12-25 00"` | 13 | **`ArrayIndexOutOfBoundsException`** | `ArgumentError: missing min part: 00` |
| `"2020-12-25 00:56"` | 16 | **`ArrayIndexOutOfBoundsException`** | `ArgumentError: missing sec part: 00:56` |
| `"2020-12-25T00:56"` | 16 | **`ArrayIndexOutOfBoundsException`** | `ArgumentError: missing sec part: 00:56` |
| `"2020-12-25 00:5"` | 15 | `ArgumentError: two digits min is expected after ':': :5` | same |
| `"2020-12-25 00:56Z"` | 17 | `ArgumentError: missing sec part: 00:56Z` | same |
| `"2020-12-25 00:56:17"` | 19 | parses | parses |

One byte shorter and the two-digit check fires first; one byte longer and there is a byte to read.

**Cause**

`RubyTimeParser.substr` builds the message fragment for `needColon` and `noFraction`:

```java
// RubyTimeParser.java:161-163
private IRubyObject substr(ThreadContext context, int from) {
    return context.runtime.newString(new ByteList(bytes, from, ptr - from + 1, true));
}
```

The `+ 1` is correct in intent and matches CRuby's `time_init_parse`, which formats the same region
as `%.*s` with a length of `(int)(ptr + 1 - time_part)` — the consumed region **plus the byte that
failed the check**, which is what makes `missing sec part: 00:56 ` show the offending space.

What CRuby has and JRuby does not is a terminator. `%.*s` treats its precision as a maximum and
stops at a NUL (`vsnprintf.c`, `case 's'`: `memchr(cp, 0, prec)`), and a Ruby String's buffer is
NUL-terminated at `RSTRING_END`. So when the string ends at the failing position CRuby writes the
consumed region and stops. JRuby copies `ptr - from + 1` bytes out of `bytes` unconditionally and
runs off the end.

Walking `"2020-12-25 00:56"` (`beg = 0`, `end = 16`):

| step | `ptr` |
|---|---|
| year | 4 |
| `-`, mon | 7 |
| `-`, mday | 10 |
| `' '`, `timePart = 11` | 11 |
| `expectTwoDigits("hour")` | 13 |
| `needColon("min")`, `expectTwoDigits("min")` | 16 = `end` |
| `noFraction("min")` — `peek()` is EOS-guarded, returns 0 | 16 |
| `needColon("sec", start = 11)` — `peek()` returns 0, not `':'` | 16 |

`substr(context, 11)` then asks for `new ByteList(bytes, 11, 16 - 11 + 1 = 6, true)`, i.e.
`System.arraycopy(bytes, 11, dst, 0, 6)` against a 16-byte array.

`peek()` and `isEOS()` are correctly guarded; this length computation is the only unguarded read
in the class.

**Suggested fix**

Clamp the fragment to the end of the string, which reproduces what the NUL terminator does for
CRuby:

```java
private IRubyObject substr(ThreadContext context, int from) {
    return context.runtime.newString(new ByteList(bytes, from, Math.min(ptr + 1, end) - from, true));
}
```

Dropping the `+ 1` outright also stops the crash, but it removes the offending byte from every
message that does not end at the failing position, which contradicts CRuby and turns two ruby/spec
examples red (`spec/ruby/core/time/new_spec.rb:566` expects `missing sec part: 00:56 ` with the
trailing space, and `:635` expects `fraction min is not supported: 00:56.` with the trailing dot).

**Why this shows up in practice**

`ActiveModel::Type::Time#cast_value` manufactures exactly this shape:

```ruby
# activemodel/lib/active_model/type/time.rb
dummy_time_value = value.sub(/\A\d{4}-\d\d-\d\d(?:T|\s)|/, "2000-01-01 ")
```

so any `time`-typed attribute assigned an `"HH:MM"` string — what a time picker posts — becomes
`"2000-01-01 06:00"`, 16 bytes, and hits this. Rails' `fast_string_to_time` rescues `ArgumentError`
only (`active_model/type/helpers/time_value.rb`), so a `java.lang.*` exception propagates out as a
500 instead of becoming a validation error. Rails selects that code path on
`RUBY_VERSION >= "3.2"`, so it is invisible on JRuby 9.4 and appears on the move to JRuby 10.

Not the same as #9531 / #9544, which is a keyword-flag problem in the invokedynamic binding.

A PR against `jruby-10.0` follows.

aminmansuri:
Fixes #9623.

`Time.new(String)` threw `java.lang.ArrayIndexOutOfBoundsException` instead of raising
`ArgumentError` whenever the string ended exactly at the byte where the parser expects a `':'`:

```ruby
Time.new("2020-12-25 00:56")   # 16 bytes, missing sec
Time.new("2020-12-25T00:56")   # 16 bytes, ISO separator
Time.new("2020-12-25 00")      # 13 bytes, missing min
```

```
Java::JavaLang::ArrayIndexOutOfBoundsException: arraycopy: last source index 17 out of bounds for byte[16]
    org.jruby.util.RubyTimeParser.substr(RubyTimeParser.java:162)
    org.jruby.util.RubyTimeParser.needColon(RubyTimeParser.java:150)
    org.jruby.util.RubyTimeParser.parse(RubyTimeParser.java:72)
```

**Cause.** `substr` builds the fragment for the `needColon` and `noFraction` messages out of the
consumed region *plus the byte that failed the check* — deliberately, because that is what CRuby's
`time_init_parse` does with `%.*s` and a length of `(int)(ptr + 1 - time_part)`, and it is what
makes `missing sec part: 00:56 ` show the offending space. The difference is that CRuby's `%.*s`
treats the precision as a maximum and stops at the string's NUL terminator (`vsnprintf.c`,
`case 's'`: `memchr(cp, 0, prec)`), so at end of input it truncates by itself; JRuby copies
`ptr - from + 1` bytes out of the backing array with nothing to stop it.

**Fix.** Clamp the fragment to `end`, which is what the NUL terminator does for CRuby. Messages
for input that does not end at the failing byte are byte-for-byte unchanged.

Note that simply dropping the `+ 1` is not equivalent — it also removes the offending byte from
the messages that currently carry it, and turns `spec/ruby/core/time/new_spec.rb:566`
(`missing sec part: 00:56 `, trailing space) and `:635` (`fraction min is not supported: 00:56.`,
trailing dot) red.

**Test.** `spec/regression/GH-9623_time_new_string_ends_at_colon_spec.rb`. Three examples for the
three crashing inputs; a fourth pins the messages for input that does not end at the failing byte,
so the trailing byte cannot be removed by a later change. Put in `spec/regression` rather than
`spec/ruby` because a Java exception escaping is JRuby-specific and `spec/ruby` is synced from
ruby/spec, following `spec/regression/GH-8854_time_new_with_empty_keywords_spec.rb`.

**Verified.** Built on Corretto 25 / macOS arm64.

* the new spec: 3 failures before, 0 after
* `rake spec:regression`: 401 examples, 4 failures before, 1 after (the remaining one is a
  pre-existing environmental `File#utime` failure on this box)
* `rake spec:ruby:fast`: 32674 examples, identical failure set before and after
* `mspec ci spec/ruby/core/time`: 778 examples, 0 failures, 0 errors, unchanged
* `test/mri/ruby/test_time.rb`: unchanged

The MRI suite does not currently catch this — `TestTime#test_new_from_string` is in
`test/mri/excludes/TestTime.rb` for an unrelated pre-existing reason.

**Where it shows up.** `ActiveModel::Type::Time#cast_value` rewrites any value to
`"2000-01-01 <value>"`, so a `time` column given an `"HH:MM"` string becomes a 16-byte string and
hits this. Rails rescues `ArgumentError` only, so the Java exception propagates out as a 500 rather
than a validation error. Rails picks that path on `RUBY_VERSION >= "3.2"`, so it is invisible on
JRuby 9.4 and appears on the move to JRuby 10.


sampokuokkanen:
The tests could go to ruby/spec: https://github.com/ruby/spec/blob/620a912d97cf321a47eb5c754f37a060a94326e3/core/time/new_spec.rb#L563.  Looks like every error case there ends with a trailing `+09:00`, same in `TestTime#test_new_from_string`.

aminmansuri:
Good suggestion — submitted the boundary cases to ruby/spec as well as ruby/spec#1393.

headius (changes requested):
This looks good, but the specs should be added to `spec/ruby/core/time/new_spec.rb` or similar since they describe standard Ruby behavior.

aminmansuri:
Changes made: test cases are now in spec/ruby/core/time/new_spec.rb.  I removed the spec/regression file. The case is identical to ruby/spec#1393 so if ruby spec accepts the change then there will be no conflict.

headius (approved):
Thank you for moving the specs! Looks good.

PR #9625, opened 2026-09-01
Read unnamed modules of the loaders JRuby loads Java classes from

### LINKED ISSUE #9319: [Java 26 Leyden] AOT Cache breaks dynamic Ruby proxies for classes loaded from external JARs

gillesbergerp:
Fixes #9319.

org.jruby.dist is an automatic module, and automatic modules get their read edge to unnamed modules from Module.defineModules at boot. An AOT-archived boot layer skips that, and the edge lives in Module.ReflectionData.reads, which isn't archived -- while resolved 
eads sets survive inside the archived Module objects. Hence JDK classes keep binding and gem-jar classes stop. This asks for the edge instead of assuming it.

Verified on JDK 26.0.1+8-34: with -XX:AOTCache, canRead false -> true and the methods come back.

### DIFF
--- a/core/src/main/java/org/jruby/Ruby.java
+++ b/core/src/main/java/org/jruby/Ruby.java
@@ -634,11 +634,19 @@ public final class Ruby implements Constantizable {
             else {
                 jrubyClassLoader = new SelfFirstJRubyClassLoader(config.getLoader());
             }
+            addReadsUnnamed(jrubyClassLoader);
+            addReadsUnnamed(config.getLoader());
         }
         else {
             jrubyClassLoader = null; // a NullClassLoader object would be better ...
         }
         return jrubyClassLoader;
+    }
+
+    // An automatic module's read edge to unnamed modules comes from Module.defineModules, which
+    // a CDS/AOT archived boot layer skips, so ask for it rather than assume it (jruby/jruby#9319).
+    private static void addReadsUnnamed(ClassLoader loader) {
+        if (loader != null && JRUBY_MODULE.isNamed()) JRUBY_MODULE.addReads(loader.getUnnamedModule());
     }
 
     private void initDefaultEncodings(ThreadContext context) {
--- a/rakelib/rspec.rake
+++ b/rakelib/rspec.rake
@@ -59,7 +59,7 @@ namespace :spec do
     t.pattern = 'spec/grammar/**/*_spec.rb'
   end
 
-  permute_specs "regression", compile_flags do |t|
+  permute_specs "regression", compile_flags, "test:compile" do |t|
     t.rspec_opts ||= []
     t.rspec_opts << '--color --format documentation '
     t.pattern = 'spec/regression/**/*_spec.rb'
--- a/spec/regression/GH-9319_aot_cache_binds_runtime_loaded_java_classes_spec.rb
+++ b/spec/regression/GH-9319_aot_cache_binds_runtime_loaded_java_classes_spec.rb
@@ -1,0 +1,37 @@
+# -*- encoding: utf-8 -*-
+
+# https://github.com/jruby/jruby/issues/9319
+
+require 'tmpdir'
+
+# An AOT-restored boot layer skips Module.defineModules, dropping org.jruby.dist's read
+# edge to the unnamed module JRubyClassLoader loads gem classes into.
+SRC = <<~RUBY
+  $CLASSPATH << 'test/target/test-classes'
+  obj = Java::Java_integrationFixtures::ClassWithSimpleMethod.new
+  puts obj.foo('bar')
+RUBY
+
+def aot_cache_supported?
+  ENV_JAVA['java.specification.version'].to_i >= 25
+end
+
+describe 'a Java class loaded from the JRuby class loader' do
+  before { skip 'requires a JVM with AOT cache support' unless aot_cache_supported? }
+
+  it 'keeps its bound methods when the boot layer comes from an AOT cache' do
+    # --nocache avoids -XX:SharedArchiveFile, which the JVM rejects alongside AOT cache flags.
+    jruby = "#{ENV_JAVA['jruby.home']}/bin/jruby --nocache"
+
+    Dir.mktmpdir('jruby-9319') do |dir|
+      cache = File.join(dir, 'test.aot')
+      script = File.join(dir, 'probe.rb')
+      File.write(script, SRC)
+
+      expect(`#{jruby} -J-XX:AOTCacheOutput=#{cache} #{script} 2>&1`).to include('foobar')
+      expect(File).to exist(cache)
+
+      expect(`#{jruby} -J-XX:AOTCache=#{cache} #{script} 2>&1`).to include('foobar')
+    end
+  end
+end


headius (approved):
Approve, and a good find. I'll merge this but I'd like to find some clarification about the behavior change in AOT mode: why does it change?

headius:
@gillesbergerp One small item: could you rebase this to merge into `jruby-10.0` branch?

gillesbergerp:
Rebased onto `jruby-10.0` and retargeted the PR base. 

1. Normally, JRuby's automatic module gets a read edge to the unnamed module of each classloader via `Module.defineModules` at boot.
2. That edge is stored in `Module.ReflectionData.reads` — a field that isn't archived.
3. An AOT-cached boot layer restores already-resolved `Module` objects, but skips the step that adds the edge — so `canRead` comes back `false`.
4. This PR just calls `addReads` explicitly after creating each classloader, instead of relying on that boot step.

headius:
> skips the step that adds the edge

This is what I want to investigate separately. Is it intentional that AOT mode skips this, or is it a JDK bug?

Your regression tests do not seem to pass. Take another look?

gillesbergerp:
Thank you for making me aware. The spec was failing because `bin/jruby` enables AppCDS by default and appends
`-XX:SharedArchiveFile=...`, which the JVM won't accept alongside `-XX:AOTCacheOutput`:

gillesbergerp:
I dug into this a bit more and I think I found and confirmed it against JDK 25.

`org.jruby.dist` is an automatic module ([shaded/pom.xml#L60](https://github.com/jruby/jruby/blob/jruby-10.0/shaded/pom.xml#L60)) and only gets its read edge to unnamed modules from one place: [`Module.defineModules`](https://github.com/openjdk/jdk/blob/jdk-25-ga/src/java.base/share/classes/java/lang/Module.java#L1293-L1295), which calls `implAddReads(ALL_UNNAMED_MODULE, true)`. That edge is stored in a plain static map (`ReflectionData.reads`) that never gets archived as [`CDS.defineArchivedModules`](https://github.com/openjdk/jdk/blob/jdk-25-ga/src/java.base/share/classes/jdk/internal/misc/CDS.java#L130) only restores the native `Module` objects, and `ArchivedBootLayer` never touches `reads` at all. Since `defineModules` is skipped entirely when booting from an archive ([`ModuleBootstrap.boot()`](https://github.com/openjdk/jdk/blob/jdk-25-ga/src/java.base/share/classes/jdk/internal/module/ModuleBootstrap.java#L154-L169)), that edge just never gets added and [`Module.canRead`](https://github.com/openjdk/jdk/blob/jdk-25-ga/src/java.base/share/classes/java/lang/Module.java#L438) comes back `false` for an unnamed module — which is exactly why nothing gets bound at [`MethodGatherer.java#L173`](https://github.com/jruby/jruby/blob/jruby-10.0/core/src/main/java/org/jruby/javasupport/binding/MethodGatherer.java#L173).

The interesting part is why this is new in 25/26: JDK 21 flatly refuses to use the archived boot layer if `--module-path` or `--add-opens` are set. JDK 25 [`canUseArchivedBootLayer()`](https://github.com/openjdk/jdk/blob/jdk-25-ga/src/java.base/share/classes/jdk/internal/module/ModuleBootstrap.java#L141-L144) dropped that restriction down to [just three checks](https://github.com/openjdk/jdk/commit/efdbb6afce4116140c066641128264ab42697912#diff-4cf2686849263d0c29f4b88df52cfdd3c2be66db49a7188cafe45ec470c7ec3fL145).

Related, the two sibling checks, [`canUse`](https://github.com/openjdk/jdk/blob/jdk-25-ga/src/java.base/share/classes/java/lang/Module.java#L1132) and [`implIsExportedOrOpen`](https://github.com/openjdk/jdk/blob/jdk-25-ga/src/java.base/share/classes/java/lang/Module.java#L695), both derive their "automatic modules get X for free" behavior from the descriptor on the fly, so they survive archiving just fine. Readability is the odd one out, computed once and thrown away which feels like an oversight rather than something intentional.

This also connects back to #8987/#8996. The [`canRead` gate](https://github.com/jruby/jruby/blob/jruby-10.0/core/src/main/java/org/jruby/javasupport/binding/MethodGatherer.java#L173) added there (alongside [`JRUBY_MODULE`](https://github.com/jruby/jruby/blob/jruby-10.0/core/src/main/java/org/jruby/Ruby.java#L5408)) is exactly what turns the missing edge into "no methods bound" instead of some quieter failure.

gillesbergerp:
Sorry for the test failure. This is the first regression with a compiled Java fixture so I had to add `"test:compile"`

headius:
Ok I follow the behavior of JDK 25 now. I'm not sure I entirely agree with it, but that's a debate to be had with the AOT folks perhaps.

Issue #9507, opened 2026-06-25
IO::Buffer#read and #write treat length=0 as zero bytes instead of remaining buffer capacity


samuel-williams-shopify:
## Environment

```text
jruby 10.1.0.0 (4.0.0) 2026-04-20 32f988b78c OpenJDK 64-Bit Server VM 21.0.11 on 21.0.11 +indy +jit [arm64-darwin]
```

Compared with:

```text
ruby 4.0.5 (2026-05-20 revision 64336ffd0e) +PRISM [arm64-darwin24]
```

## Reproduction: `IO::Buffer#read`

```ruby
# frozen_string_literal: true

input, output = IO.pipe
buffer = IO::Buffer.new(5)

begin
  output.write("Hello")

  result = buffer.read(input, 0, 0)

  puts RUBY_DESCRIPTION
  puts "read result: #{result.inspect}"
  puts "buffer: #{buffer.get_string.inspect}"
ensure
  input.close
  output.close
end
```

## Reproduction: `IO::Buffer#write`

```ruby
# frozen_string_literal: true

input, output = IO.pipe
buffer = IO::Buffer.for("Hello")

begin
  result = buffer.write(output, 0, 0)
  output.close

  puts RUBY_DESCRIPTION
  puts "write result: #{result.inspect}"
  puts "pipe: #{input.read.inspect}"
ensure
  input.close unless input.closed?
  output.close unless output.closed?
end
```

## Expected behavior

For both `IO::Buffer#read(io, length, offset)` and `IO::Buffer#write(io, length, offset)`, `length = 0` means to transfer as much as can fit in the buffer from the given offset, i.e. `buffer.size - offset` bytes.

CRuby `IO::Buffer#read` output:

```text
ruby 4.0.5 (2026-05-20 revision 64336ffd0e) +PRISM [arm64-darwin24]
read result: 5
buffer: "Hello"
```

CRuby `IO::Buffer#write` output:

```text
ruby 4.0.5 (2026-05-20 revision 64336ffd0e) +PRISM [arm64-darwin24]
write result: 5
pipe: "Hello"
```

## Actual behavior

JRuby treats `length = 0` as an explicit zero-byte transfer.

JRuby `IO::Buffer#read` output:

```text
jruby 10.1.0.0 (4.0.0) 2026-04-20 32f988b78c OpenJDK 64-Bit Server VM 21.0.11 on 21.0.11 +indy +jit [arm64-darwin]
read result: 0
[...]

Issue #9614, opened 2026-08-28
Unreliable net-http tests


jcharaoui:
The HTTP tests in `test/mri/net/http/test_http.rb` seem to be unreliable, at least in 10.1.1.0.

On Debian unstable every time I run these tests at least on random test case fails with an `EBADF: Bad file descriptor` error.

Example:

```
root@05ef37c32f9d:/jruby# ./bin/jruby test/mri/runner.rb --excludes=test/mri/excludes:test/mri/excludes_wip test/mri/net/http/test_http.rb
Run options: 
  --seed=49126
  --excludes=test/mri/excludes:test/mri/excludes_wip

# Running tests:

[59/84] TestNetHTTP_v1_2#test_send_requestwarning: thread "Ruby-0-Thread-45: /jruby/test/mri/net/http/utils.rb:27" terminated with exception (report_on_exception is true):Errno::EBADF: Bad file descriptor - No message available
  close at org/jruby/RubyIO.java:2352
  start at /jruby/test/mri/net/http/utils.rb:33
   loop at org/jruby/RubyKernel.java:1861
  start at /jruby/test/mri/net/http/utils.rb:28
 = 0.02 s
  1) Error:
TestNetHTTP_v1_2#test_send_request:
Errno::EBADF: Bad file descriptor - No message available
    org/jruby/RubyIO.java:2352:in 'close'
    /jruby/test/mri/net/http/utils.rb:33:in 'block in start'
    org/jruby/RubyKernel.java:1861:in 'loop'
    /jruby/test/mri/net/http/utils.rb:28:in 'block in start'

Finished tests in 9.457121s, 8.6707 tests/s, 65.5591 assertions/s.                                           
82 tests, 620 assertions, 0 failures, 1 errors, 0 skips

ruby -v: jruby 10.1.1.0 (4.0.0) 2026-08-28 ffffffffff OpenJDK 64-Bit Server VM 25.0.4.1+1-1-Debian on 25.0.4.1+1-1-Debian +indy +jit [x86_64-linux]
```

I'm able to reproduce these errors on a git build of the latest HEAD.

jcharaoui:
I applied this patch to check if it would help but it didn't: https://github.com/ruby/net-http/commit/69f6f6dedd5648290e224956c6b3f57ab516f537

headius:
Yeah this has been a known problem with this series of tests for a while. I'm not sure why they fail intermittently.

headius:
[...]

PR #9628, opened 2026-09-02
Fix IO::Buffer#read and #write, treat length=0 as remaining buffer capacity

evaniainbrooks:
issue #9507


evaniainbrooks:
Failing test TestNetHTTP_v1_2_chunked#test_s_post_form seems to be time sensitive and unrelated to these changes.

headius:
@evaniainbrooks Yeah I haven't figured out why those [net/http tests are so flaky](https://github.com/jruby/jruby/issues/9614) but the rest of this looks fine.

headius (approved):

Issue #9276, opened 2026-03-01
OpenURI tests' with_http fails due to dead thread


headius:
There's some issue with the `with_http` method used by the Ruby 4.0 OpenURI tests resulting in a number of failures in `test:mri:stdlib` that all look the same. It's unclear if this is a problem with this test code or something broken in OpenURI itself.

The code:

https://github.com/jruby/jruby/blob/f3a452590546173e04267e3858ed2c337f1638f0/test/mri/open-uri/utils.rb#L318-L341

Example error:

```
TestOpenURI#test_redirect_auth_success [/home/runner/work/jruby/jruby/test/mri/open-uri/utils.rb:340]:
exceptions on 1 threads:
#<Thread:0x534df4b /home/runner/work/jruby/jruby/test/mri/open-uri/utils.rb:324 dead>:
org/jruby/ext/socket/RubyTCPServer.java:143:in 'accept': stream closed in another thread (IOError)
	from /home/runner/work/jruby/jruby/test/mri/open-uri/utils.rb:22:in 'block in start'
	from org/jruby/RubyKernel.java:1687:in 'loop'
	from /home/runner/work/jruby/jruby/test/mri/open-uri/utils.rb:21:in 'block in start'
```

The tests will be excluded for now.

headius:
This appears to affect the bulk of tests in test_open-uri.rb.

headius:
Fixed by #9629.

PR #9629, opened 2026-09-02
Fix thread-kill/close race in open-uri test server and unexclude tests

evaniainbrooks:
Move @server.close into the threads own outer ensure, the same fix already applied to the analogous net/http test server in f07fc25

issue #9276


headius (approved):
Good catch! We should probably do another mass update of CRuby tests (from appropriate branches) but this is great to have resolved.

PR #9634, opened 2026-09-02
Report the errno that native exec actually returned

sampokuokkanen:
This should fix flaky openBSD test runs like this one: https://github.com/jruby/jruby/actions/runs/33608150521/job/100176806273#step:5:248

The exec logic threw from a finally block using the last POSIX errno, so a command not found in PATH reported whatever the previous syscall left behind: ENOENT on Linux by accident, EACCES on OpenBSD, NOERROR on macOS.

The PATH search also diverged from `dln_find_1`: an empty name matched the PATH entries themselves, a non-executable candidate aborted the search, and names with a slash were not passed to execv unresolved. Untags two specs.

### DIFF
--- a/core/src/main/java/org/jruby/RubyProcess.java
+++ b/core/src/main/java/org/jruby/RubyProcess.java
@@ -1606,7 +1606,10 @@ public class RubyProcess {
         if (!Platform.IS_WINDOWS && Options.NATIVE_EXEC.load() && context.runtime.getPosix().isNative()) {
             int errno = PopenExecutor.exec(context, args, context.nil);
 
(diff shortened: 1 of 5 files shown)


headius (approved):
These look like good fixes, but I'm also tracking down why CRuby always raises ENOENT for an empty exec string even though most systems would raise EACCES:

```
$ cx ruby-4.0 ruby -e 'exec ""'
-e:1:in 'Kernel#exec': No such file or directory -  (Errno::ENOENT)
	from -e:1:in '<main>'
$ sh -c 'exec ""'
sh: 1: exec: : Permission denied
```

Since your fixes seem appropriate and resolve the issue we'll go with it for now.

PR #9635, opened 2026-09-02
Run error-prone on one CI job instead of all of them (reduce 429 errors on Java 21)

sampokuokkanen:
The profile activated on any JDK 21 build with CI set, so all 41 Java 21 jobs resolved error_prone_core from Central and ran the check. Activation now needs JRUBY_ERROR_PRONE, set by one new job.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -45,7 +45,7 @@ jobs:
     name: error-prone (Java 21)
 
     env:
-      JRUBY_ERROR_PRONE: true
+      ERROR_PRONE: true
 
     steps:
       - name: checkout
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -39,6 +39,26 @@ jobs:
       - name: rake ${{ matrix.target }}
         run: bin/jruby -S rake ${{ matrix.target }}
 
+  error-prone:
+    runs-on: ubuntu-latest
+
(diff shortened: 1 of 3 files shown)


kares (approved):
LGTM, since locally error-prone isn't enabled due longer `mvn` build times.

Might be worth to revisit later, as Maven has been adding some caching (similar to Gradle). 

Hopefully the job won't get stale or disabled/forgotten - happened previously and some error-prone offenses ended up being introduced.

p.s. The `JRUBY_ERROR_PRONE` sounds like smt custom is being done specifically for JRuby, otherwise this being a `JRUBY` build is implicit so maybe just `ERROR_PRONE` or a different prefix/suffix (`ERROR_PRONE_CHECKS`).

sampokuokkanen:
Thanks, renamed to `ERROR_PRONE`.

sampokuokkanen:
I think we could have error-prone run first and then have the other jobs gated on it. This could also warm the cache so this could actually be not that much slower than the current workflow. Might be worth a try.

headius:
@sampokuokkanen We used to have a gated build that built JRuby and performed some basic checks, but since the `ruby` organization allowed us to start using their subscription, all jobs start at the same time and a pre-build just slows them down.

This is a simple enough change so I'll approve and merge.

headius (approved):

PR #9636, opened 2026-09-02
Raise ENOENT eagerly for blank exec string

headius:
Found by OpenBSD tests, which raise EACCES for a blank exec prog.


headius:
May be fixed by other changes in #9634 so we'll close this for now and revisit if the problem shows up again.

Issue #9597, opened 2026-08-18
JRUBY_OPTS does not propagate into some rake tasks


headius:
Discovered by an agent in #9596...

Currently we have a few spec runs that are intended to execute with invokedynamic optimizations turned off. They do this by setting the `JRUBY_OPTS` environment variable before running rake, adding flags like `-Xcompile.invokedynamic=false`.

However, because some of the rake tasks in question build their own `JRUBY_OPTS` environment variables, the value set at the command line gets clobbered. As a result the CI jobs are not actually disabling invokedynamic.

I believe we want the specific tasks like `spec:ruby:fast:jit` to always run the same way, but I recognize we need a better way to add more options (and clearly when we have a job that's supposed to run without indy, it should run without indy). Permuting every runtime config option is clearly not feasible.

Passing arguments to the tasks may be the best way, and I'm researching to figure out what is standard these days for doing so. If we had a single configurable task that took multiple runtime arguments, it could be called like

`rake spec:ruby:fast[jit = true, threshold = 0, indy = false]`

while still keeping some easy short-cut targets in place.

PR #9643, opened 2026-09-03
Propagate JRUBY_OPTS from parent env when running mspec

headius:
We need to be able to manipulate the mspec jobs by setting JRUBY_OPTS, as described in jruby/jruby#9597, but the old logic here overwrote that environment variable while setting up the subprocess launch. This change builds the new JRUBY_OPTS starting with the current environment, so additional flags can be passed through.

This is not ideal; a rogue JRUBY_OPTS in the parent env could cause these jobs to run in ways that break their verification. Unfortunately rake provides few ways to parameterize tasks, so we are forced to permute and generate all tasks we wish to see, and ENV is the cleanest way to pass in additional tweaks.

Fixes jruby/jruby#9597

### DIFF
--- a/rakelib/commands.rake
+++ b/rakelib/commands.rake
@@ -79,7 +79,7 @@ def mspec(mspec_options = {}, java_options = {}, &code)
     sysproperty :key => "emma.verbosity.level", :value=> "silent"
 
     env :key => "JAVA_OPTS", :value => "-Demma.verbosity.level=silent"
-    env :key => "JRUBY_OPTS", :value => ms[:jruby_opts] || ""
+    env :key => "JRUBY_OPTS", :value => "#{ENV['JRUBY_OPTS']} #{ms[:jruby_opts]}"
     # launch in the same mode we're testing, since config is loaded by top process
 
     # add . to load path so mspec config is found


headius:
I'll work through the regressions in non-indy mode before merging.

headius:
First failure I'm looking at:

```
Integer#dup returns self for large integers FAILED
Expected 4611686018427387905 to be identical to 4611686018427387905
```

This appears to be due to the IR compiler propagating some literal values into their use sites.

The IR looks like this right before JIT for `i = ... ; p i.object_id; p i.dup.object_id`:

```
BB [2:LBL_1:-1] -EXIT->3
	%self = recv_self()
	%v_0 = recv_kw()
	check_arity(%v_0 ;req: 0, opt: 0, *r: false)
	line_num(;n: 0, coverage: false, oneshot: false)
	i = copy(Fixnum:4611686018427387905)
	%v_2 = call_0o(Fixnum:4611686018427387905 ;n:object_id, t:NO, cl:false)
	%v_1 = call_1o(%self, %v_2 ;n:p, t:FU, cl:false)
	%v_4 = call_0o(i ;n:dup, t:NO, cl:false)
	%v_5 = call_0o(%v_4 ;n:object_id, t:NO, cl:false)
	%v_3 = call_1o(%self, %v_5 ;n:p, t:FU, cl:false)
	return(%v_3)
```

Note that the first call to `object_id` has inlined the integer value.

headius:
Odd thing about this is that the indy compiler also shows different object_id, so it is not normalizing all propagated values to the same object either.

PR #9645, opened 2026-09-04
Fix the Fiber scheduler IO path: negated errno, zero length, buffer size

sampokuokkanen:
Three small bug fixes on the Fiber scheduler's IO path:

`FiberScheduler.result` returned a positive errno, so a scheduler comparing against `-Errno::EAGAIN::Errno` read it as a byte count instead of a wait. `IO::Buffer.new(64).read(io, 64, 0)` on a drained non-blocking socket gives `35` on master, `-35` here and on CRuby.

A length of `0` means the remaining capacity, but `IO::Buffer#pread` and `#pwrite` transferred nothing. #9628 fixed `#read` and `#write`.

`OpenFile` passed `ioReadMemory`/`ioWriteMemory` the array offset as the buffer size and the byte count as the minimum transfer length, so `io_read` got a zero-size buffer and a minimum it could never reach.

I added the new specs under `spec/ruby/core/io/buffer/` like #9628. Removes 2 excludes from `test/mri/excludes/TestFiberIOBuffer.rb`.

### DIFF
--- a/core/src/main/java/org/jruby/util/io/OpenFile.java
+++ b/core/src/main/java/org/jruby/util/io/OpenFile.java
@@ -1514,7 +1514,7 @@ public class OpenFile implements Finalizable {
         if (fptr.fiberScheduler) {
             IRubyObject scheduler = context.getFiberCurrentThread().getSchedulerCurrent();
             if (!scheduler.isNil()) {
-                IRubyObject result = FiberScheduler.ioReadMemory(context, scheduler, fptr.io, buffer, buf, count);
+                IRubyObject result = FiberScheduler.ioReadMemory(context, scheduler, fptr.io, buffer, count, 0);
 
                 if (result != null) {
                     return FiberScheduler.resultApply(context, result);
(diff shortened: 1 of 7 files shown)


core/src/main/java/org/jruby/util/io/OpenFile.java
```diff
@@ -1514,7 +1514,7 @@ public static int readInternal(ThreadContext context, OpenFile fptr, ChannelFD f
         if (fptr.fiberScheduler) {
             IRubyObject scheduler = context.getFiberCurrentThread().getSchedulerCurrent();
             if (!scheduler.isNil()) {
-                IRubyObject result = FiberScheduler.ioReadMemory(context, scheduler, fptr.io, buffer, buf, count);
+                IRubyObject result = FiberScheduler.ioReadMemory(context, scheduler, fptr.io, buffer, count, 0);
```
sampokuokkanen, line 1517: See https://github.com/ruby/ruby/blob/ruby_4_0/io.c#L1297

headius: Yeah looks like a bad read of the original code on my part.

headius (approved):
Looks good. Bonus points might be to reduce duplication in this logic, but it's fine to merge now.

PR #9646, opened 2026-09-04
Consider encoding in ByteList equality

headius:
Two ByteLists may have exactly the same bytes, but not the same encoding. Such ByteLists have been treated as equals() but we have frequently run into bugs from this behavior. This PR finally adds encoding to the conditions required for ByteList equality.

### DIFF
--- a/core/src/main/java/org/jruby/util/ByteList.java
+++ b/core/src/main/java/org/jruby/util/ByteList.java
@@ -959,7 +959,7 @@ public class ByteList implements Comparable, CharSequence, Serializable {
     @Override
     public boolean equals(Object other) {
         if (other == this) return true;
-        if (other instanceof ByteList) return equal((ByteList)other);
+        if (other instanceof ByteList bl) return equal(bl);
         return false;
     }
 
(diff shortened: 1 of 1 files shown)


headius:
Note that the caching in #9644 depends on true ByteList equality, so this will probably be required to ship that PR as it is currently implemented.

headius:
A good experiment but we appear to have many places that depend on ByteList equality only considering the contents of the byte array.

It fails at startup, most likely because two versions of the ByteList for the `import_classes` variable in the code below get confused, and we don't see that it's a variable when compiling the access.

core/src/main/ruby/jruby/java/core_ext/object.rb:
```
# @see Module.java_import
def java_import(*import_classes, &block)
  Object.send :java_import, *import_classes, &block
end
```

Resulting error:

```
$ jruby -e 1
NameError: undefined local variable or method 'import_classes' for main
  <main> at uri:classloader:/jruby/java/core_ext/object.rb:2
    load at org/jruby/RubyKernel.java:1403
  <main> at uri:classloader:/jruby/java/core_ext.rb:4
    load at org/jruby/RubyKernel.java:1403
  <main> at file:///Users/headius/work/jruby/lib/jruby.jar!/jruby/java.rb:37
```

Likely source of the problem:

https://github.com/jruby/jruby/blob/6cdb065506647052bd38695cc4a7aea76482d1e7/core/src/main/java/org/jruby/parser/RubyParserBase.java#L1332-L1340

The symbol table is keyed on ByteList, and the code parsed here may be producing two bytelists with the same content but different encodings (US-ASCII vs UTF-8 for example).

PR #9649, opened 2026-09-07
Mutex Deadlock fix: Track the fiber that owns a Mutex

sampokuokkanen:
The `Mutex` deadlock check asked the owning thread's `ThreadContext` which fiber it was currently running. That answer is only right while the root fiber holds the lock. The reverse case is a fiber holding the lock while the root fiber tries to take it. There the check never fires, and `Mutex#lock` blocks forever instead of raising ThreadError. Store the owning fiber on the mutex instead.

Unexcludes `TestFiberMutex#test_mutex_deadlock`.

Sample code: 
```ruby
m = Mutex.new
f = Fiber.new { m.lock; Fiber.yield }
f.resume
m.lock
```
Master: hangs
This branch: `ThreadError: deadlock; lock already owned by another fiber belonging to the same thread`

### DIFF
--- a/core/src/main/java/org/jruby/ext/thread/Mutex.java
+++ b/core/src/main/java/org/jruby/ext/thread/Mutex.java
@@ -57,6 +57,8 @@ public class Mutex extends RubyObject implements DataType {
      * thread.
      */
     volatile RubyThread lockingThread;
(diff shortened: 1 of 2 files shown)


headius (approved):
Yup, makes sense.

Issue #9275, opened 2026-02-28
Add additional Fiber scheduler hooks


headius:
Ruby 4.0 has expanded the callbacks for the Fiber scheduler, and we are missing some. At the very least, the following tests fail, probably because read and write are not fully hooked up to the scheduler:

```
  6) Failure:
TestFiberScheduler#test_io_read_error [/home/runner/work/jruby/jruby/test/mri/fiber/test_scheduler.rb:339]:
Expected #<SystemCallError: Unknown error (SystemCallError) - /tmp/ruby_test_io_read_error_60db71be21beca0fb478a5f9aca41e06> to be a kind of Errno::EBADF, not SystemCallError.

  7) Failure:
TestFiberScheduler#test_io_write_error [/home/runner/work/jruby/jruby/test/mri/fiber/test_scheduler.rb:360]:
Expected #<SystemCallError: Unknown error (SystemCallError) - /tmp/ruby_test_io_write_error_4c68306821923f764cae5d5dbee9cc2a> to be a kind of Errno::EINVAL, not SystemCallError.

  8) Failure:
TestFiberScheduler#test_io_write_on_flush [/home/runner/work/jruby/jruby/test/mri/fiber/test_scheduler.rb:312]:
<[[:io_write, 55, "foo"], [:io_write, 55, "bar"]]> expected but was
<[]>.

  9) Failure:
TestFiberScheduler#test_io_write_flush_error [/home/runner/work/jruby/jruby/test/mri/fiber/test_scheduler.rb:381]:
Expected nil to be a kind of Errno::EINVAL, not NilClass.
```

These need to be hooked up and made to pass.

PR #9656, opened 2026-09-09
Fix IO#wait arguments and return values, and defer it to Fiber.scheduler

### LINKED ISSUE #8743: IO#wait updates for Ruby 3.2-3.4
There's a few updates to the IO#wait API since it was merged back into core (from an ext gem). This leads to a few new spec failures for JRuby 10:

```
IO#wait [events, timeout] passed returns events mask when the READABLE event is ready during the timeout ERROR
IO#wait [events, timeout] passed returns events mask when the WRITABLE event is ready during the timeout ERROR
IO#wait [events, timeout] passed raises ArgumentError when events is not positive ERROR
IO#wait [timeout, mode] passed accepts timeout and mode in any order FAILED
IO#wait [timeout, mode] passed raises ArgumentError when several Integer arguments passed ERROR
```

They will be tagged for JRuby 10 and addressed in 10.x.

sampokuokkanen:
`IO#wait` only understood `io.wait(timeout, mode)`. MRI takes a timeout in any position among any number of mode symbols, or an events mask plus a timeout, and answers that second form with the ready mask. It now follows `io_wait`/`io_wait_event`/`rb_io_wait`, honors `IO#timeout`, and returns `false` rather than `nil` on a mask miss. Reporting a mask needed `RubyThread.selectReadyOps` and `OpenFile.readyOps`.

None of the three wait methods ever called `Fiber.scheduler`'s `io_wait`, so a non-blocking Fiber parked its thread instead of yielding. Fixed now, part of #9275.

Also removed the shadowed static copies of IO's wait methods.

Added specs are from ruby/spec#1395 and ruby/spec#1396.

Still shadowed: `require 'io/wait'` loads a jar that redefines all three with the pre-10.0 code. ruby/io-wait#76 empties it out the way MRI's `Init_wait` already is. Once that releases the last two `TestIOWait` exclusions can go.

This closes #8743.


sampokuokkanen:
drb requires `io/wait`, that seems to make the tests fail. So will need to wait until https://github.com/ruby/io-wait/pull/76 is merged and released.

core/src/main/java/org/jruby/RubyIO.java
```diff
@@ -4129,109 +4129,140 @@ public static IRubyObject select(ThreadContext context, IRubyObject recv, IRubyO
...
-        long tv = prepareTimeout(context, argv);
-
-        return doWait(context, fptr, tv, SelectionKey.OP_CONNECT | SelectionKey.OP_WRITE);
+        return ioWaitEvent(context, IOEvent.IO_WRITABLE.value, argv.length == 1 ? argv[0] : context.nil, true);
     }
 
-    @JRubyMethod(optional = 2)
+    // MRI: io_wait
+    @JRubyMethod(rest = true)
     public IRubyObject wait(ThreadContext context, IRubyObject[] argv) {
```
headius, line 4156: Now that this is a primary part of Ruby IO, we'll probably want to split this up by arity as well, which may simplify the argument logic here.

headius (approved):
Yeah good catch-up changes. I think io/wait kept changing and I just lost track as it evolved along with fiber scheduling.

headius:
Well that's unexpected... seems like something here is triggering a JVM crash across multiple versions... likely some native call for IO given the subject.

The concurrent-ruby failures also appear to be a regression.

headius:
Ok I'm catching up... I see your ruby/io-wait patch to avoid overwriting the JRuby versions with bad versions from the extension. I'll approve that and arrange a release.

headius:
We may want to consider a backport to 10.0 (Ruby 3.4).

headius:
@sampokuokkanen FYI, I am hoping to do a 10.1.2.0 release before Euruko next week (ideally some time on 10 Sept or 14 Sept at the latest). Let me know if you have more patches coming you think should be included.

sampokuokkanen:
@headius 
Thanks, too bad I can't make it to Euruko this year. Hopefully next one, I've only been to Euruko once. 

I'm having Samuel look at https://github.com/ruby/spec/pull/1397 now, but I'll do a PR for Fiber#transfer today.

headius:
I filed ruby/io-wait#77 to get a release out and ruby/io-wait#78 and ruby/io-wait#79 to fully zero out and deprecate the gem. Waiting on feedback from @nobu or @hsbt.

headius:
Almost there! After taking another look, @nobu and I agreed to finally hollow out and deprecate the gem, so there will be no extension code in there at all anymore. Once ruby/io-wait#80 is merged, a final version of the gem will be released and the repository archived. We can then switch to that empty gem in this PR and everything should be green.

@sampokuokkanen if you have some free cycles, you can confirm that the empty gem works properly while we wait for the release.

sampokuokkanen:
@headius 
Seems to work well. 
But now there are warnings when loading `io-wait` and it is loaded from gems like drb. Which I guess is also nice since it will force those gems to update their code.

headius:
@sampokuokkanen Yeah that is unfortunate but they should not be depending on this gem anymore.

io-wait 1.0.0, the final empty release, is now available. You can update this PR to switch to it and things should largely be green.

sampokuokkanen:
Tests green on the new and improved (= emptied) `io-wait`!

headius:
I'm working my way through other libraries that still require io/wait but the changes are propagating.

PR #9657, opened 2026-09-09
Do not NPE when a fiber terminates without a resumer

sampokuokkanen:
A fiber entered via `Fiber#transfer` can reach the end of its block with `FiberData.prev` still null, and the fiber thread body dereferenced it unconditionally when handing the result back:

```java
data.prev.data.queue.push(ctxt, result);
```

Follow-up in a second PR makes `returnFiber()` actually match MRI's semantics, which is what un-excludes `test_terminate_transferred_fiber`.


sampokuokkanen:
Closing, still under work.

PR #9658, opened 2026-09-09
Implement MRI return_fiber semantics for Fiber#transfer

sampokuokkanen:
### Problem

`Fiber#transfer` stole `prev` from the transferring fiber and handed it to the target:

```java
if (currentFiberData.prev != null) {
    data.prev = currentFiberData.prev;
    currentFiberData.prev = null;
    currentFiberData.transferred = true;
}
```

That conflates the transfer chain with the resume chain. A fiber terminating after being transferred to returned control to whoever had resumed further up, skipping the fiber actually parked waiting for it. `test_terminate_transferred_fiber` logged `[:fa2, :fa1, :root]` instead of `[:fa2, :fa1, :r1, :root]`.

### Fix

Follow MRI: `transfer` does not touch `prev` on either fiber, and the resume relationship is tracked from both ends. `FiberData` gains `resumingFiber` (MRI: `fiber->resuming_fiber`), set only by `Fiber#resume`, and `returnFiber()` walks that chain from the root fiber exactly as MRI's `return_fiber` does.

`Fiber.yield` from a fiber that was never resumed is now a plain `FiberError` with MRI's wording instead of a "report this bug" message, and the two error cases in `Fiber#transfer` now say which one was hit.


sampokuokkanen:
Closing, still under work.

PR #9659, opened 2026-09-10
[ji] resolve Java super <init> through included modules

### LINKED ISSUE #9655: NoSuchMethodError when a Ruby subclass of a Java class includes a module
## Environment

- JRuby: `jruby 10.1.2.0-SNAPSHOT (4.0.0) 2026-09-08 d6eb6862bb`
- Java: OpenJDK 25.0.4
- OS: Ubuntu Linux x86_64

## Reproduction

Save the following as `repro.rb`:

```ruby
class ReproChild < java.lang.Object
  include Enumerable

  def initialize
    super
  end
end

ReproChild.new
```

Run:

```bash
jruby repro.rb
```

## Actual behavior

JRuby raises `NoSuchMethodError` while creating `ReproChild`:

```
Unhandled Java exception: java.lang.NoSuchMethodError: 'org.jruby.exceptions.RaiseException org.jruby.Ruby.newNoMethodError(java.lang.String, java.lang.String, org.jruby.runtime.builtin.IRubyObject[])'
java.lang.NoSuchMethodError: 'org.jruby.exceptions.RaiseException org.jruby.Ruby.newNoMethodError(java.lang.String, java.lang.String, org.jruby.runtime.builtin.IRubyObject[])'
               <init> at rubyobj/repro.rb:3
               <init> at rubyobj/repro.rb:3
                 call at org/jruby/java/proxies/ConcreteJavaProxy.java:312
                 call at org/jruby/java/proxies/ConcreteJavaProxy.java:170
  performIndirectCall at org/jruby/ir/targets/indy/InvokeSite.java:893
               invoke at org/jruby/ir/targets/indy/InvokeSite.java:797
            ❤ script at repro.rb:7
                  run at repro.rb:-1
  invokeWithArguments at java/lang/invoke/MethodHandle.java:735
                 load at org/jruby/ir/Compiler.java:114
            runScript at org/jruby/Ruby.java:1227
          runNormally at org/jruby/Ruby.java:1139
          runFromMain at org/jruby/Ruby.java:984
          internalRun at org/jruby/main/Main.java:289
                  run at org/jruby/main/Main.java:239
                 main at org/jruby/main/Main.java:211

```

## Expected behavior

`ReproChild.new` should complete without an exception.

## Additional information

The error appears to come from a generated Java constructor attempting to call `Ruby.newNoMethodError(String, String, IRubyObject[])`, although the available overload has a different third-argument type. This may be related to the Java subclass reification changes in commit e5a8dc10d4 (#9557).

kares:
the second commit - fallback branch in the generated code is hard to test, since it's being handled up to stack. just poked manually (revering the first commit) and got:

```
NoMethodError: No available java superconstructors match that type signature
  <main> at repro.rb:9
```

### DIFF
--- a/spec/java_integration/methods/inheritance_variants_spec.rb
+++ b/spec/java_integration/methods/inheritance_variants_spec.rb
@@ -288,6 +288,71 @@ describe "Java subclassing - constructor argument variants" do
     end
   end
 
+  describe "module inclusion (Ruby subclass includes a module)" do
+    it "constructs when an included module does not define initialize (GH-9655)" do
+      cls = Class.new(MultiCtorBase) do
+        include Enumerable
+        def initialize; super(7); end
+      end
+      obj = cls.new
+      expect(obj.ctor).to eq("(int)")
+      expect(obj.trace.to_a).to eq(["Java (int) ctor with 7"])
+    end
+
+    it "constructs with a no-arg super through an included module without initialize" do
+      mod = Module.new { def size; 42; end }
+      cls = Class.new(MultiCtorBase) do
+        include mod
+        def initialize; super(); end
+      end
+      expect(cls.new.ctor).to eq("()")
+    end
+
+    # NOTE: older JRuby silently skipped an included module's initialize here
+    it "runs an included module's initialize between the class and the Java ctor" do
+      mod = Module.new do
+        attr_reader :mod_marker
+        def initialize(s)
+          super(s)
+          @mod_marker = "mod-#{s}"
+        end
+      end
+      cls = Class.new(MultiCtorBase) do
+        include mod
+        attr_reader :cls_marker
+        def initialize(s)
+          super(s)
+          @cls_marker = "cls-#{s}"
+        end
+      end
+
+      obj = cls.new("x")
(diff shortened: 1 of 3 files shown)


spec/java_integration/methods/inheritance_variants_spec.rb
```diff
@@ -288,6 +288,71 @@ def initialize(s, n)
...
+      mod = Module.new { def size; 42; end }
+      cls = Class.new(MultiCtorBase) do
+        include mod
+        def initialize; super(); end
+      end
+      expect(cls.new.ctor).to eq("()")
+    end
+
+    # NOTE: older JRuby silently skipped an included module's initialize here
+    it "runs an included module's initialize between the class and the Java ctor" do
```
kares, line 312: this has been a "regression" for a long time - since the Java auto-reification got introduced.

there's some old issues reported around this, will need to review those, believe the updated logic (considering `initialize` in included modules) is correct and should be the behavior specified.

headius (approved):
The logic is still complex but the consideration of included and prepended modules looks correct.

PR #9660, opened 2026-09-10
Fix where Fiber#transfer hands control back

sampokuokkanen:
`Fiber#transfer` took `prev` from the transferring fiber and gave it to the target, mixing the transfer chain up with the resume chain. A fiber terminating after a transfer handed control back up the resume chain instead of to the fiber waiting for it. With no resumer at all it hit a null `prev`.

So do as MRI does. Transfer leaves `prev` alone, resume and raise track the link from both ends through a new `resumingFiber`, and `returnFiber` walks that chain from the root fiber like `return_fiber`. A difference with MRI: we hand control over through queues rather than by swapping coroutine stacks, so a fiber can die with nobody waiting for it, and `returnFiber` only hands over to a fiber genuinely parked in a transfer.

Also fixed `to_s` with fibers so debugging will be easier. 

This drops 6 `test_fiber.rb` exclusions and 12 ruby/spec tags. It adds 5 for imported specs that describe MRI behaviour we still lack, most of which need `Fiber#kill` to become synchronous (ruby/spec#1397).


headius:
Very nice! I meant to circle back to this and figure out the proper sequence of events for transfer but I will review and get this merged today.

sampokuokkanen:
Thanks, seems like we have to exclude one more `Fiber#kill` spec as kill can return before the fiber unwinds.

headius:
Making `Fiber#kill` or anything else synchronous will take some careful work, due to the ability of fibers to potentially run in parallel for short bursts on JVM. On CRuby, they are by design and by necessity always synchronous with an explicit hand-off.

Someone (perhaps me) also needs to catch up with the latest work on Loom, including custom schedulers (speed up the hand-off) and pinning (explicitly give up the current native thread to a target virtual thread in one shot).

headius (approved):
Looks like a good alignment with the CRuby fiber internal structure and fixes a bunch of specs. Ideally unrelated fixes would be in different PRs, but I understand they are pretty intertwined here. Thanks for your continued work here! I'm excited to look into more fiber performance once we get some scheduler and async use cases working well enough to benchmark.

sampokuokkanen:
Could we use Loom continuations? Would be quite a big change I imagine though. 

I'll try to keep the changes in different PR's next time, could've at least had the `to_s` fix in a different one.

PR #9661, opened 2026-09-10
Report a failed non-blocking connect through select and SO_ERROR (#8786)

### LINKED ISSUE #8786: `Redis::ConnectionError` with JRuby 10
Hi, the following code raises a `Redis::ConnectionError: Socket is not connected - No message available` error with JRuby 10 (but not JRuby 9.4.12.0).

```ruby
require "redis"

client = Redis.new
p client.get("hello")
```

**Environment Information**
  
* jruby 10.0.0.0 (3.4.2) 2025-04-13 6ed59bc847 OpenJDK 64-Bit Server VM 21.0.6 on 21.0.6 +indy +jit [arm64-darwin]
* macOS 15.4.1 arm64

**Expected Behavior**

No error

**Actual Behavior**

```text
Redis::ConnectionError: Socket is not connected - No message available (redis://localhost:6379)
  translate_error! at /Users/user/.rbenv/versions/jruby-10.0.0.0/lib/ruby/gems/shared/gems/redis-5.4.0/lib/redis/client.rb:32
            call_v at /Users/user/.rbenv/versions/jruby-10.0.0.0/lib/ruby/gems/shared/gems/redis-5.4.0/lib/redis/client.rb:92
      send_command at /Users/user/.rbenv/versions/jruby-10.0.0.0/lib/ruby/gems/shared/gems/redis-5.4.0/lib/redis.rb:152
       synchronize at org/jruby/ext/monitor/Monitor.java:85
      send_command at /Users/user/.rbenv/versions/jruby-10.0.0.0/lib/ruby/gems/shared/gems/redis-5.4.0/lib/redis.rb:151
               get at /Users/user/.rbenv/versions/jruby-10.0.0.0/lib/ruby/gems/shared/gems/redis-5.4.0/lib/redis/commands/strings.rb:191
            <main> at repro.rb:4
RedisClient::ConnectionError: Socket is not connected - No message available (redis://localhost:6379)
             write at /Users/user/.rbenv/versions/jruby-10.0.0.0/lib/ruby/gems/shared/gems/redis-client-0.24.0/lib/redis_client/ruby_connection.rb:78
              call at /Users/user/.rbenv/versions/jruby-10.0.0.0/lib/ruby/gems/shared/gems/redis-client-0.24.0/lib/redis_client/connection_mixin.rb:30
            call_v at /Users/user/.rbenv/versions/jruby-10.0.0.0/lib/ruby/gems/shared/gems/redis-client-0.24.0/lib/redis_client.rb:302
              call at /Users/user/.rbenv/versions/jruby-10.0.0.0/lib/ruby/gems/shared/gems/redis-client-0.24.0/lib/redis_client/middlewares.rb:16
            call_v at /Users/user/.rbenv/versions/jruby-10.0.0.0/lib/ruby/gems/shared/gems/redis-client-0.24.0/lib/redis_client.rb:301
  ensure_connected at /Users/user/.rbenv/versions/jruby-10.0.0.0/lib/ruby/gems/shared/gems/redis-client-0.24.0/lib/redis_client.rb:707
            call_v at /Users/user/.rbenv/versions/jruby-10.0.0.0/lib/ruby/gems/shared/gems/redis-client-0.24.0/lib/redis_client.rb:300
            call_v at /Users/user/.rbenv/versions/jruby-10.0.0.0/lib/ruby/gems/shared/gems/redis-5.4.0/lib/redis/client.rb:90
      send_command at /Users/user/.rbenv/versions/jruby-10.0.0.0/lib/ruby/gems/shared/gems/redis-5.4.0/lib/redis.rb:152
       synchronize at org/jruby/ext/monitor/Monitor.java:85
      send_command at /Users/user/.rbenv/versions/jruby-10.0.0.0/lib/ruby/gems/shared/gems/redis-5.4.0/lib/redis.rb:151
               get at /Users/user/.rbenv/versions/jruby-10.0.0.0/lib/ruby/gems/shared/gems/redis-5.4.0/lib/redis/commands/strings.rb:191
            <main> at repro.rb:4
Errno::ENOTCONN: Socket is not connected - No message available
    write_nonblock at org/jruby/RubyIO.java:1496
             write at /Users/user/.rbenv/versions/jruby-10.0.0.0/lib/ruby/gems/shared/gems/redis-client-0.24.0/lib/redis_client/ruby_connection/buffered_io.rb:130
              loop at org/jruby/RubyKernel.java:1658
             write at /Users/user/.rbenv/versions/jruby-10.0.0.0/lib/ruby/gems/shared/gems/redis-client-0.24.0/lib/redis_client/ruby_connection/buffered_io.rb:129
             write at /Users/user/.rbenv/versions/jruby-10.0.0.0/lib/ruby/gems/shared/gems/redis-client-0.24.0/lib/redis_client/ruby_connection.rb:76
              call at /Users/user/.rbenv/versions/jruby-10.0.0.0/lib/ruby/gems/shared/gems/redis-client-0.24.0/lib/redis_client/connection_mixin.rb:30
            call_v at /Users/user/.rbenv/versions/jruby-10.0.0.0/lib/ruby/gems/shared/gems/redis-client-0.24.0/lib/redis_client.rb:302
              call at /Users/user/.rbenv/versions/jruby-10.0.0.0/lib/ruby/gems/shared/gems/redis-client-0.24.0/lib/redis_client/middlewares.rb:16
            call_v at /Users/user/.rbenv/versions/jruby-10.0.0.0/lib/ruby/gems/shared/gems/redis-client-0.24.0/lib/redis_client.rb:301
  ensure_connected at /Users/user/.rbenv/versions/jruby-10.0.0.0/lib/ruby/gems/shared/gems/redis-client-0.24.0/lib/redis_client.rb:707
            call_v at /Users/user/.rbenv/versions/jruby-10.0.0.0/lib/ruby/gems/shared/gems/redis-client-0.24.0/lib/redis_client.rb:300
            call_v at /Users/user/.rbenv/versions/jruby-10.0.0.0/lib/ruby/gems/shared/gems/redis-5.4.0/lib/redis/client.rb:90
      send_command at /Users/user/.rbenv/versions/jruby-10.0.0.0/lib/ruby/gems/shared/gems/redis-5.4.0/lib/redis.rb:152
       synchronize at org/jruby/ext/monitor/Monitor.java:85
      send_command at /Users/user/.rbenv/versions/jruby-10.0.0.0/lib/ruby/gems/shared/gems/redis-5.4.0/lib/redis.rb:151
               get at /Users/user/.rbenv/versions/jruby-10.0.0.0/lib/ruby/gems/shared/gems/redis-5.4.0/lib/redis/commands/strings.rb:191
            <main> at repro.rb:4
```

aminmansuri:
Fixes #8786

It looks like it was incorrect error handling.

It happens because:

1. JRuby 10, because it ships Ruby 3.4's socket library with the Happy Eyeballs connect loop. JRuby 9.4 has the old sequential loop and is not affected.
2. A call to Socket.tcp with a hostname. An IP literal takes a different, older path. Libraries that use TCPSocket.new, such as redis 4.4.0 and Net::HTTP, never enter this loop.
3. A hostname that resolves to more than one address, in practice both an IPv6 and an IPv4 one. localhost is the everyday case: most systems map it to ::1 and 127.0.0.1. A dual-stack DNS name does the same.
4. One of those connects is refused. The everyday case again: the server listens on IPv4 only, which is Redis's default bind 127.0.0.1, so the attempt on ::1 is refused. The loop races both addresses; when select sees the refused one, JRuby raises instead of reporting it, and the whole call fails even though the IPv4 connect would have worked.

It does not happen when the server listens on both families, when the connect fails by timeout rather than refusal, when the JVM has IPv6 disabled with the IPv4 flag, or on an IP literal.

**Right now:** select tries to complete the refused connect, the JDK throws and closes that channel, JRuby turns the exception into a plain IOError, and the loop dies right there: it never learns which socket failed, never sees the IPv4 socket finish, and its cleanup then trips over the channel the JDK already closed. The caller gets an error and no connection.

**The Fix:** select catches the JDK's exception, keeps it on the refused socket's descriptor, and returns both sockets as writable, which is what the operating system would do. The loop asks each one SO_ERROR. The refused socket now answers "connection refused" instead of zero, so the loop drops it, closes it, which no longer fails, and asks the IPv4 socket, which answers zero. That socket is handed back and the first write succeeds. If the refused address had been the only one, the loop would have reported "connection refused" with the right errno instead of a generic IO error.


aminmansuri:
was able to remove test_tcp_socket_hostname_resolution_failed_after_connection_failure exclusion.

headius:
The one failing job looked unrelated so I ran it again.

test/jruby/test_socket.rb
```diff
@@ -763,4 +763,56 @@ def test_read_zero_never_blocks
     server.close rescue nil
     client.close rescue nil
   end if RUBY_VERSION >= '1.9'
+
```
headius: This is very specific behavior, but probably should still go into rubyspec (spec/ruby/library/socket) to be tested on other implementations as well. None of the behavior tested here is specific to JRuby.

core/src/main/java/org/jruby/util/io/SelectExecutor.java
```diff
@@ -187,8 +187,20 @@ IRubyObject selectInternal(ThreadContext context) throws IOException {
...
+                                sock.finishConnect();
+                            } catch (IOException ioe) {
+                                // select(2) reports a non-blocking connect that failed as writable
+                                // and leaves the error for getsockopt(SO_ERROR) or the next
+                                // operation on the socket; it does not fail the select itself.
+                                // finishConnect() throws (and closes the channel) instead, so stash
+                                // the error and carry on. Socket.tcp's Happy Eyeballs v2 logic
+                                // depends on this to fall back to the other address family.
+                                // See jruby/jruby#8786.
+                                channelFD.connectError = ioe;
```
headius: I hate that we have to emulate this behavior, but it's not a huge price to pay to be able to leverage the same socket.rb logic as CRuby.

core/src/main/java/org/jruby/util/io/ChannelFD.java
```diff
@@ -207,6 +209,16 @@ private void initChannelTypes() {
...
+    /**
+     * The error from a non-blocking connect that failed, if any.
+     *
+     * POSIX leaves the error of a failed connect on the descriptor, for
+     * getsockopt(SOL_SOCKET, SO_ERROR) to report, and the descriptor stays valid until
+     * the application closes it. {@link SocketChannel#finishConnect()} instead throws
+     * and closes the channel, so we stash the error here and reproduce the POSIX
+     * behaviour for both getsockopt and close. See jruby/jruby#8786.
+     */
+    public volatile IOException connectError;
```
headius: Is this socket-specific enough to move to RubyBasicSocket instead of including it on all IO objects?

headius (changes requested):
This is a good find and closes the loop (pun intended) with the new Happy Eyeballs code and JRuby's emulation of BSD socket behaviors. A few questions and a few changes and we can move forward with this.

test/jruby/test_socket.rb
aminmansuri: Moved to spec/ruby/library/socket in e647790f.

core/src/main/java/org/jruby/util/io/ChannelFD.java
aminmansuri: Yes. That sounds more correct. Moved to RubyBasicSocket in dcf281b2.

headius (approved):
Thanks for the updates!

PR #9662, opened 2026-09-11
Fix custom respond_to? handling: include_all, defined? lookup, deprecated signature

sampokuokkanen:
I was only going to fix `test_respond_to` and the deprecation warning it needs, but got kind of carried away. I can split this into two PRs for the two commits too.  It's in one for now because they touch the same files, and `test_respond_to` is actually the only test in MRI that covers a redefined `respond_to?`.

`defined?` called `respond_to_missing?` directly. MRI reaches it through `rb_check_funcall`, which asks `respond_to?` first, so an object that overrides `respond_to?` to say no can stop the lookup. 

```ruby
class Blocker
  def respond_to?(name, include_all = false) = false
  def respond_to_missing?(name, priv) = true
end

defined?(Blocker.new.anything)   # MRI: nil, JRuby: "method"
```

Also the include_all bug:

```ruby
class Priv
  def respond_to?(name, include_all = false) = super
  private  def to_str = "converted"
end

p String(Priv.new)
```

```
MRI 3.4.7:     "converted"
JRuby 10.1.1.0:  "#<Priv:0x44c73c26>" (only wrong if you override)
```

I'll try to upstream the specs later. 
Should this be against the jruby-10.0 branch? https://github.com/jruby/jruby/pull/9448 is not on jruby-10.0 yet and this touches the same file and lines in it. 

<details>

<summary>Interesting note: </summary>

```ruby
private

def to_ary = nil

def respond_to?(method, include_all = false)
  super || _remote_specification.respond_to?(method, include_all)
end
public :respond_to?
```
https://github.com/ruby/rubygems/blob/master/lib/bundler/remote_specification.rb#L124

This is an extra network call on JRuby master(fixed here). 
</details>


headius (approved):
Looks good. Largely pulling over equivalent logic from CRuby but there's some good cleanup to alias handling.

PR #9673, opened 2026-09-12
Bump the amount of memory for baseline CI runs

headius:
1GB should be enough for most jobs, but as more dependencies have crept in and test suites have gotten larger, it sometimes fails. Bump the default max heap size up to 1GB.

Note that most test jobs specify their own sizes. This limit has only been noted to affect some larger bundles and third-party suites.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -3,7 +3,7 @@ name: JRuby CI
 on: [push, pull_request]
 
 env:
-  JAVA_OPTS: '-XX:+TieredCompilation -XX:TieredStopAtLevel=1 -Xms60M -Xmx1G -XX:InitialCodeCacheSize=40M -XX:ReservedCodeCacheSize=120M'
+  JAVA_OPTS: '-XX:+TieredCompilation -XX:TieredStopAtLevel=1 -Xms60M -Xmx2g -XX:InitialCodeCacheSize=40M -XX:ReservedCodeCacheSize=120M'
 
 permissions:
   contents: read
--- a/.github/workflows/dist-verification-ci.yml
+++ b/.github/workflows/dist-verification-ci.yml
@@ -10,7 +10,7 @@ on:
 
 
 env:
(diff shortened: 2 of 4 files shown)


headius:
It does concern me that this frequently affects bundling with our own `Gemfile`, which contains only a handful of dependencies. I cannot reproduce locally even limiting the JVM to 128MB of memory:

```
$ jruby -J-Xmx128M -S bundle install --force
Installing diff-lcs 1.6.2
Installing rspec-support 3.12.2
Installing rspec-core 3.12.3
Installing rspec-expectations 3.12.4
Installing rspec-mocks 3.12.7
Installing rspec 3.12.0
Bundle complete! 1 Gemfile dependency, 6 gems now installed.
Use `bundle info [gemname]` to see where a bundled gem is installed.
```

This only appears to happen on GHA.

headius:
From GHA:

```
Run bin/jruby --dev -S bundle install
  bin/jruby --dev -S bundle install
  shell: /usr/bin/bash -e {0}
  env:
    JAVA_OPTS: -XX:+TieredCompilation -XX:TieredStopAtLevel=1 -Xms60M -Xmx1G -XX:InitialCodeCacheSize=40M -XX:ReservedCodeCacheSize=120M
    JAVA_HOME: /opt/hostedtoolcache/Java_Zulu_jdk/25.0.4-7/x64
    JAVA_HOME_25_X64: /opt/hostedtoolcache/Java_Zulu_jdk/25.0.4-7/x64
    MAVEN_ARGS: -ntp
Fetching gem metadata from https://rubygems.org/...


Your JVM has run out of memory, and Bundler cannot continue. You can decrease the amount of memory Bundler needs by removing gems from your Gemfile, especially large gems. (Gems can be as large as hundreds of megabytes, and Bundler has to read those files!). Alternatively, you can increase the amount of memory the JVM is able to use by running Bundler with jruby -J-Xmx1024m -S bundle (JRuby defaults to 500MB).
```

Issue #8118, opened 2024-02-22
Flaky OpenSSL "noverify" test


jcharaoui:
Hello,

I'm seeing intermittent `mri-stdlib` test failures with JRuby 9.4.5.0 on Debian testing/sid with OpenJDK 17.0.10.

The failures are related to one specific test: `TestOpenURISSL#test_validation_noverify`.

When I loop this command, usually after anywhere between one and forty runs, it always ends up failing:

`while true; do echo "attempt #${n} ..."; jruby test/mri/runner.rb --excludes=test/mri/excludes:test/mri/excludes_wip test/mri/open-uri/test_ssl.rb; sleep 1; n=$(( n+1 )); done`

When it fails, this is the stack trace:

```
Run options: 
  --seed=61134
  --excludes=test/mri/excludes:test/mri/excludes_wip

# Running tests:

warning: thread "Ruby-0-Thread-22: /tmp/autopkgtest-lxc.6djqypt5/downtmp/autopkgtest_tmp/test/mri/open-uri/test_ssl.rb:45" terminated with exception (report_on_exception is true):
EOFError: end of file reached
                 rbuf_fill at /usr/share/jruby/lib/ruby/stdlib/net/protocol.rb:227
                 readuntil at /usr/share/jruby/lib/ruby/stdlib/net/protocol.rb:193
                  readline at /usr/share/jruby/lib/ruby/stdlib/net/protocol.rb:203
          read_status_line at /usr/share/jruby/lib/ruby/stdlib/net/http/response.rb:42
                  read_new at /usr/share/jruby/lib/ruby/stdlib/net/http/response.rb:31
         transport_request at /usr/share/jruby/lib/ruby/stdlib/net/http.rb:1609
                     catch at org/jruby/RubyKernel.java:1310
         transport_request at /usr/share/jruby/lib/ruby/stdlib/net/http.rb:1600
                   request at /usr/share/jruby/lib/ruby/stdlib/net/http.rb:1573
                 open_http at /usr/share/jruby/lib/ruby/stdlib/open-uri.rb:333
                     start at /usr/share/jruby/lib/ruby/stdlib/net/http.rb:985
                 open_http at /usr/share/jruby/lib/ruby/stdlib/open-uri.rb:327
               buffer_open at /usr/share/jruby/lib/ruby/stdlib/open-uri.rb:760
                 open_loop at /usr/share/jruby/lib/ruby/stdlib/open-uri.rb:214
[...]

PR #9675, opened 2026-09-12
Release the autoload claim when Autoload#load has nothing to load

### LINKED ISSUE #9674: Constant declared with `autoload` and then defined by a direct `require` keeps its UNDEF table slot; every later lookup re-enters LoadService (very slow inside a WAR)
**Environment Information**

* jruby 10.0.6.0 (3.4.5) 2026-06-11 716ad51f54 OpenJDK 64-Bit Server VM 25.0.4+7-LTS on 25.0.4+7-LTS -indy +jit [arm64-darwin]; also reproduced on x86_64 Linux (Ubuntu 22.04) under Apache Tomcat 10.1.57 with jruby-rack 1.2.7
* No JRUBY_OPTS; `-Djruby.compile.invokedynamic=false` (same result with it on)
* Rails 7.2.3.2, activerecord-jdbcpostgresql-adapter 72.1

**Expected Behavior**

After

```ruby
module Foo
  autoload :Bar, "foo/bar"
end
require "foo/bar"
```

`Foo::Bar` is a defined constant and a lookup of it is a constant-table hit, as on CRuby. `Foo.autoload?(:Bar)` is `nil` and `Foo.const_defined?(:Bar)` is `true`, which JRuby reports correctly.

**Actual Behavior**

The constant-table slot for `Foo::Bar` still holds the autoload marker (UNDEF) after the `require` satisfied it: `JRuby.reference(Foo).getConstantMap.get("Bar").value` raises a NullPointerException, whereas a constant defined by a plain `require` with no prior `autoload` holds its value. Every subsequent lookup of `Foo::Bar` from a thread other than the one that ran the `require` goes through `RubyModule$Autoload#load` → `LoadService#featureAlreadyLoaded`, which walks `$LOAD_PATH`, instead of returning the constant directly. (`Autoload#load` short-circuits with `isSelf` for the requiring thread, so the effect is invisible on that thread.)

In a command-line process the walk is cheap and the effect is only visible in the table. Inside a WAR it is not: every `$LOAD_PATH` entry is a `uri:classloader:` resource, and each expansion is a `ClassLoader#getResource` plus `URLEncoder.encode` plus `new java.net.URI`, about 450 µs per lookup measured under Tomcat 10.1.57. Rails declares its classes with `ActiveSupport::Autoload` and then requires them during eager loading, so on a request thread `Marshal.load` of an ActiveRecord object resolves several such constants per embedded object:

* `Marshal.load` of one cached 6 KB record: 0.34 ms on the boot thread, 23.1 ms on a request thread, same process, same bytes.
* A page that reads 89 cached records: 1,229 ms in deserialisation; 36 ms after the constants are re-set (see below). JFR attributed 82.6 % of execution samples of that page to `Marshal.load` → autoload → `LoadService#featureAlreadyLoaded` → `WebappClassLoaderBase#findResource`.

Re-reading each affected constant on the boot thread and `const_set`-ing the same object back replaces the slot and removes the cost entirely, which is the workaround we deploy; it also shows that nothing else about the constant is wrong.

**Reproduction**

`undef_slot_repro.rb` attached (with `lib/foo/bar.rb` and `lib/foo/baz.rb` each defining an empty class): `jruby -Ilib undef_slot_repro.rb` prints the table state for both constants. The timing block in it is near zero on the command line by design, for the reason above; the cost needs a classloader-backed `$LOAD_PATH`.

Possibly related: #2510 (explicit require of an autoloaded class), #8923 / #8995 (autoload slots and `private_constant`, 10.0.1–10.0.3), #9426 (UNDEF leaking through constant resolution).

[example.zip](https://github.com/user-attachments/files/32141106/example.zip)

aminmansuri:
Fixes #9674.

`Autoload#load` records the claim (`ctx`) before it checks `featureAlreadyLoaded`. When the feature is already loaded or still loading there is nothing to load, but the claim is kept. Any constant that file then defines is treated as this autoload's own result and parked on the `Autoload`; the constant table keeps UNDEF, `autoload?` and `const_defined?` report correctly, and every later lookup from another thread (`Marshal.load`, `const_get`, class reopening) goes through `LoadService#featureAlreadyLoaded` again instead of reading the table.

The feature is "already loaded or loading" in three common shapes: a direct `require` of the autoload's file, several constants declared with `autoload` against one file (`ActiveSupport::Autoload`'s `autoload_at`, the usual Rails layout) where the first one loaded parks its siblings, and a circular `require` between two autoloaded files.

On the command line that walk is cheap. Under a classloader-backed `$LOAD_PATH` (a WAR) each lookup costs about 450 µs; a page deserialising 89 cached objects spent 1,229 ms in `Marshal.load`, 36 ms once the slots held their values.

The fix releases the claim in the already-loaded branch. The specs pin what Ruby can observe (a `Marshal.load` and a `$LOAD_PATH` change from another thread no longer depend on the loader; `remove_const` returns the value; every constant of a shared file becomes a regular constant); the `test/jruby` cases pin the constant-table slot itself, including the sibling and circular shapes.

### DIFF
--- a/core/src/main/java/org/jruby/RubyModule.java
+++ b/core/src/main/java/org/jruby/RubyModule.java
@@ -6689,11 +6689,14 @@ public class RubyModule extends RubyObject {
                 // This method needs to be synchronized for removing Autoload
                 // from autoloadMap when it's loaded.
                 LoadService loadService = loadService(context);
-                if (!loadService.featureAlreadyLoaded(path.asJavaString())) {
-                    if (loadService.autoloadRequire(path)) {
-                        // Do not finish autoloading by cyclic autoload
-                        finishAutoload(context, symbol);
-                    }
+                if (loadService.featureAlreadyLoaded(path.asJavaString())) {
+                    // Nothing to load here: the feature is loaded, or a direct require of it is in
+                    // progress. Keeping the claim would make that require's definition of the
+                    // constant look like this autoload's own and leave UNDEF in the constant table.
+                    this.ctx = null;
+                } else if (loadService.autoloadRequire(path)) {
+                    // Do not finish autoloading by cyclic autoload
+                    finishAutoload(context, symbol);
                 }
             } catch (LoadError | RuntimeError lre) {
                 // reset ctx to null for a future attempt to load
--- a/spec/ruby/core/kernel/autoload_spec.rb
+++ b/spec/ruby/core/kernel/autoload_spec.rb
@@ -164,6 +164,39 @@ describe "Kernel.autoload" do
       KernelSpecs::AutoloadMethod2::AutoloadFromIncludedModule2.loaded.should == :autoload_from_included_module2
     end
   end
+
+  describe "after the autoload was satisfied by requiring the file directly" do
+    before :each do
+      @path = fixture(__FILE__, "autoload_satisfied_by_require.rb")
+      Kernel.autoload :KSAutoloadSatisfiedByRequire, @path
+      ScratchPad.record []
+    end
+
+    after :each do
+      if Object.const_defined?(:KSAutoloadSatisfiedByRequire, false)
+        Object.send(:remove_const, :KSAutoloadSatisfiedByRequire)
+      end
+    end
+
+    it "defines a regular constant that no longer depends on $LOADED_FEATURES" do
+      require @path
+      ScratchPad.recorded.should == [:loaded]
+      Kernel.autoload?(:KSAutoloadSatisfiedByRequire).should be_nil
+      klass = KSAutoloadSatisfiedByRequire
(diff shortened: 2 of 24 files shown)


headius (approved):
I would like to get this merged since it's a significant impact to startup when running in an embedded environment. The specs added will probably want a bit of review from @eregon but I will merge as is for now. The tests are extensive, likely a lot more than this particular fix demanded, but I appreciate the additional coverage. @eregon may have ideas for making them more general to include in ruby/spec.

headius:
The two failures are known-flaky tests (see #8118) and unrelated to the content of this PR.

I've pinged @eregon in my review, to offer suggestions for improving the ruby/spec specs, but we'll get this merged in for JRuby 10.0.7.0 and later releases.

Issue #5147, opened 2018-04-23
Coverage options support (2.5)


headius:
To move forward with JRuby 9.2 I have tagged off the new Coverage options tests (lines, branches, methods). We still need to implement this for 2.5 compatibility, but it is being reduced in priority.

The excludes:
```
exclude :test_branch_coverage_for_case_statement, "no support for coverage options yet"
exclude :test_branch_coverage_for_if_statement, "no support for coverage options yet"
exclude :test_branch_coverage_for_safe_method_invocation, "no support for coverage options yet"
exclude :test_branch_coverage_for_while_statement, "no support for coverage options yet"
exclude :test_line_coverage_for_multiple_lines, "no support for coverage options yet"
exclude :test_method_coverage, "no support for coverage options yet"
exclude :test_method_coverage_for_alias, "no support for coverage options yet"
exclude :test_method_coverage_for_define_method, "no support for coverage options yet"
exclude :test_method_coverage_for_singleton_class, "no support for coverage options yet"
```

PragTob:
:wave: 

Just chiming in that I'm finally adding branch coverage support to simplecov. No pressure though has/is taking us long enough and honesty I don't think it's that much of a killer feature but people want it :)

headius:
I'm not sure what's required for us to support it. @enebo you have looked at event hook stuff recently, did you happen to see how the new coverage options work?

headius:
Nobody is jumping on this or other tracing-related issues so we are untargeting. Help wanted!

enebo:
Obviously this is ancient but since I am reviewing old issues marked with parser label I will mention the basics on this.  Branching combine with traceinstr needs to mark branch followed.  I have not looked at how MRI does this but we definitely know where we are branching as ruby semantic branching vs our internal needs of branching in our instruction set.  I envision setting a field in each branch path hwich is then seen when the trace logic happens which adds some new attributes to the trace instr.

PR #9676, opened 2026-09-13
Implement method coverage

sferik:
This adds the `methods` mode of the `Coverage` library. Results have the same shape, keys and counts as CRuby:

```ruby
Coverage.start(methods: true)
require "foo"
Coverage.result
# => { "/path/foo.rb" => { methods: { [Foo, :bar, 3, 2, 5, 5] => 2, ... } } }
```

Keys are `[owner, name, start_line, start_column, end_line, end_column]`, exactly as CRuby reports them (1-based lines, 0-based byte columns, `def` through `end`, or the block/lambda passed to `define_method`).

Addresses the method half of https://github.com/jruby/jruby/issues/5147. Once this is merged, I intend to work on branch coverage, which I think would be substantially trickier.

`MethodCoverage` is the JRuby analogue of the per-method-entry counters CRuby keeps (its `me2counter` hash). It is created in `RubyModule#putMethod`, the one place every method entry lands, whenever methods are being measured and the entry's source file is being tracked, and it is attached to the `DynamicMethod`. As in CRuby, entries that only forward to another entry get no counter of their own: aliases, `private :inherited_method` in a subclass, `define_method(:x, instance_method(:aliased))`. Calls through them count toward the original. `module_function`, `Class#dup` and `define_method(name, method_object)` create fresh entries with the new owner, again as in CRuby.

CRuby counts on the `CALL` event, which fires after the arguments have been bound, so a call that fails with an `ArgumentError` (wrong arity, missing keyword) or a raising default value is not counted. To match that, the count happens inside the body rather than in `DynamicMethod#call`:
* every Ruby-level call path (`InterpretedIRMethod`, `MixedModeIRMethod`, `CompiledIRMethod`, `ProcMethod`) hands the entry's counter to the body through a slot on `ThreadContext`
* `ReceiveMethodCoverageInstr`, the first instruction of the body, takes the counter into a temp before any argument is received, so nested calls made while receiving arguments (default value expressions, `to_ary` conversions) cannot take it
* `CoverMethodInstr`, emitted right after argument receipt, increments it. The counter records which scope it belongs to, so a stale hand-off can never be attributed to another method.
* a `define_method` method whose lambda arity check fails (that check runs in Java, before the body) clears the counter it left pending, so a later plain call of the same proc is not counted as a call of the method
* the `initialize` of a Java subclass, which `ConcreteJavaProxy` runs through the split-constructor machinery rather than `DynamicMethod#call` (or skips entirely when it is a plain forwarding `super`), gets the hand-off in `startSplitSuperCall` and is counted directly when its body is skipped

Blocks get the same two probes because any block can become a method via `define_method`. When called as plain blocks they see no counter and the probes are a null check. Invokedynamic call sites no longer bind directly to the compiled handle of a method that is being counted, so the hand-off is never bypassed.

Counting is a single lock-free atomic add (`VarHandle#getAndAdd`), so threads calling the same method neither serialize nor lose increments. A new test verifies 8 threads × 5000 calls produce exactly 40000. Registration and result conversion are synchronized on `CoverageData` as the existing line coverage code already is.

The grammar now records byte columns and end positions for `def`, blocks and lambdas (`DefNode`/`IterNode` gained `getStartColumn`/`getEndColumn`, and `MethodDefNode#getEndLine` is now the line of the `end` keyword rather than the end of the body). Lambda spans follow CRuby: from the start of the parameter list, or just past `->` when there is none, through the end of the body. Both generated parsers were regenerated from `RubyParser.y` with jay; the toolchain was verified to reproduce the checked-in sources byte-for-byte before the grammar was touched. The spans are carried into `IRScope` and survive the block-to-method conversion used by `define_method`.

`CoverageData` now keeps a `FileCoverage` per file (lines, methods, and room for branches) instead of a bare `IntList`. Results contain only the requested modes, `branches` reports `{}` for now, files are listed in parse order like CRuby, and `Coverage.supported?(:methods)` is true.

Two other notes:
* `IRClosure` silently dropped the `coverageMode` passed to its constructor, so blocks converted into methods for `define_method` were built without any coverage instrumentation.
* `oneshot_lines` probes disarmed themselves when hit while coverage was suspended, losing the line for good. They now stay armed until the line is actually counted.

Known remaining differences, both pre-existing and unrelated to counting: JRuby keys loaded files by their real path, and `Class#dup` keeps aliases as aliases (CRuby materializes them into real entries). Prism builds do not supply columns yet, so keys would carry `-1` columns there.

Following @enebo's note in https://github.com/jruby/jruby/issues/5147#issuecomment-4189391581, branches will be marked at IR build time where the builder knows it is compiling a Ruby-semantic branch, rather than recovered from the IR's own jumps. The per-file container, the mode-driven result conversion and the parser source spans this PR adds are the groundwork for that.


core/src/main/java/org/jruby/internal/runtime/methods/CompiledIRMethod.java
```diff
@@ -134,6 +134,7 @@ protected void printMethodIR() {
 
     @Override
     public IRubyObject call(ThreadContext context, IRubyObject self, RubyModule clazz, String name, IRubyObject[] args, Block block) {
+        prepareMethodCoverage(context);
```
kares, line 137: is this approach the best we can do, having to invoke `prepareMethodCoverage` from all places?

not sure how much thought went into the PR but would be nice to consider other options...

sferik: I did weigh alternatives but open to suggestions I may not have considered.

The counter has to be per method *entry* (CRuby keys results by `[owner, name, location]`, so `Class.new { def x; end }` in a loop or `%w[a b].each { |n| define_method(n) { } }` produce distinct counts from one scope), and the count has to happen after arguments are bound (CRuby counts on the `CALL` event, so a call that dies on arity or a missing keyword is not counted). So something has to get the entry's counter from the `DynamicMethod` into the body, and the body has to do the increment.

Here are the options I considered:
1. Wrap the entry in a delegating `DynamicMethod` at registration, the way `ProfilingDynamicMethod`/`MethodEnhancer` do it, so the hand-off lives in one class. The problem is that a wrapper changes the concrete class of the entry and several places key real behavior off that: `InvokeSite` decides keyword handling by `entry.method instanceof AbstractIRMethod`, refinement `import_methods` raises for non-`AbstractIRMethod` entries, `Method#==` for `define_method` methods checks `instanceof ProcMethod`, `ruby2_keywords` reads the static scope through the same check, and the Java proxy code detects user-defined `initialize`/`new` with it. Each would need an unwrap. That is more invasive and more fragile than one line per call variant, and it only matters while coverage is on.
2. I didn’t count in `call` without a body probe because it’s the same number of touch points and wrong for calls that fail on their arguments.
3. Hook the trace `CALL` event won’t work because it’s only emitted in full-trace mode and before arguments are received.
4. I didn’t fold the hand-off into the compiled `MethodHandle`s because it covers `CompiledIRMethod` only. Interpreted and mixed-mode paths still need explicit hooks.
5. I didn’t pass the entry into the body because it touches the interpreter, JIT, and indy signatures.

The one-liner per arity variant follows what these classes already do for every per-call concern: `doDebug()` and the `callCount`/`tryJit` check are repeated in each `call` of `MixedModeIRMethod` and `InterpretedIRMethod`. `prepareMethodCoverage` sits next to those. With coverage off it is a field load plus a predictable null check, which is cheaper than the `callCount` check beside it.

If you would rather see fewer sites, `InterpretedIRMethod`'s ten variants could route through the five `Interpreter.INTERPRET_METHOD` helpers by passing the method instead of its scope, but that trades ten trivial lines for a signature change on a public helper. Happy to do that if you think it is worth it!

core/src/main/java/org/jruby/internal/runtime/methods/ProcMethod.java
```diff
@@ -59,6 +59,7 @@ public ProcMethod(RubyModule implementationClass, RubyProc proc, Visibility visi
     }
 
     public IRubyObject call(ThreadContext context, IRubyObject self, RubyModule klazz, String name, IRubyObject[] args, Block block) {
+        prepareMethodCoverage(context);
```
kares, line 67: this won't cut compatibility for procs the same way as MRI: e.g. counter is prepared **before the proc performs (lambda) arity checking**; if the call raises `ArgumentError` the counter isn't consumed.

reproduces under both interpreter and synchronous JIT

core/src/main/java/org/jruby/internal/runtime/methods/MixedModeIRMethod.java
```diff
@@ -95,6 +95,7 @@ protected void printMethodIR() {
     @Override
     public IRubyObject call(ThreadContext context, IRubyObject self, RubyModule clazz, String name, IRubyObject[] args,
             Block block) {
+        prepareMethodCoverage(context);
```
kares, line 98: another( minor) concern, I am aware of (due working) on Java sub-classing in Ruby, is that the so called "execute split-super constructor paths" don't end up calling `prepareMethodCoverage` (likely same with `AbstractIRMethod`)

that is: `initialize` on a Java subclass reports zero calls after repeated `JavaSubClass.new`

core/src/main/java/org/jruby/ir/runtime/IRRuntimeHelpers.java
```diff
@@ -686,13 +703,26 @@ public static boolean shouldPrintScope(IRScope scope) {
...
     public static void updateCoverage(ThreadContext context, String filename, int line) {
+        coverLine(context, filename, line);
+    }
+
+    /**
+     * Update coverage data for the given file and zero-based line number.
+     *
+     * @return whether the line was counted (coverage is running); a oneshot_lines probe must stay armed until then
+     */
+    public static boolean coverLine(ThreadContext context, String filename, int line) {
```
kares, line 714: minor performance concern: `isCoverageEnabled` is mode-independent, meaning `coverLine` (and `growCoverageLines`) allocate and populate parser line arrays even for `Coverage.start(methods: true)`

core/src/main/java/org/jruby/ext/coverage/FileCoverage.java
```diff
@@ -0,0 +1,60 @@
...
+ * <ul>
+ * <li>{@link #getLines()}: execution count per line (-1 for lines that hold no code), present only while
+ * lines are being measured;</li>
+ * <li>{@link #getMethods()}: one {@link MethodCoverage} for every method entry defined from this file since
+ * coverage was set up, in definition order.</li>
+ * </ul>
+ */
+public final class FileCoverage {
+    private IntList lines;
+    private final List<MethodCoverage> methods = new CopyOnWriteArrayList<>();
```
kares: is this the best collection, if we're under a lock already (which seems like the case here) than every method definition does not need to copy the entire array...

headius:
@sferik thank you for jumping in here and taking a shot at this! I agree with your analysis of how CRuby does this and I'm sure we can work through refining your patch to minimize performance impact. I'm going to mark this as 10.1.x so we can take a little more time to make it perfect.

These missing coverage features have been bothering me more lately so I'm excited to get some help implementing them!

sferik:
@headius Thanks Charlie! I’m happy to help move these coverage features forward however I can, whether or not you actually use any of this code. I’ve added a bunch of features to SimpleCov in the past few releases, and I’d like for JRuby users to benefit from those changes. More selfishly, I’d like to remove the `unless RUBY_ENGINE == "jruby"` conditions from that codebase.

My experience with Java and the JVM is limited, so I’ll happily take whatever advice you have. I’m working through @kares feedback now.

Targeting 10.1 sounds good to me. These features have been missing from JRuby for a while, so there’s no particular urgency and I think it’s more important that this is done right than done quickly. I’m particularly sensitive to any performance impact it might have on production code. If those impacts can’t be mitigated, I’m open to gating them behind `--debug` (or a similar flag), since users typically only use these features in test environments. That said, [SimpleCov 1.3 introduced a production mode for finding dead code](https://github.com/simplecov-ruby/simplecov/blob/main/docs/Production.md), so I’d want to exclude oneshot line coverage from that flag, if we end up going that route. Hopefully we can optimize everything enough that we don’t have to!

sferik:
@headius I just tried rebasing this branch from `10.1-dev` but that looks like it hasn’t been updated since February and is an ancestor of `master`, so I just rebased from `master` to resolve a conflict. Please advise if there’s something you’d like me to do differently.

core/src/main/java/org/jruby/internal/runtime/methods/ProcMethod.java
sferik: Good catch. `ProcMethod#call` now clears the pending counter in a `finally` (only on the path where the entry actually has a counter, so uncovered calls are unchanged) and there is a test for exactly that sequence. The def paths are not affected because their arity checks run inside the body after `ReceiveMethodCoverageInstr` has already taken the counter.

core/src/main/java/org/jruby/internal/runtime/methods/MixedModeIRMethod.java
sferik: The hand-off now lives in `AbstractIRMethod#startSplitSuperCall` with the compiled-or-interpreted choice split out into `splitSuperCall` so `MixedModeIRMethod` can still delegate to its jitted method without losing the counter. For the two `ConcreteJavaProxy` shortcuts that skip the Ruby body altogether, the call is counted directly via `DynamicMethod#coverElidedCall`.

core/src/main/java/org/jruby/ir/runtime/IRRuntimeHelpers.java
sferik: Right, the runtime `coverLine` was already gated on `LineNumberInstr.coverage` (lines mode only) but the parser side was not. It built the per-line array for every parse while coverage was set up. `RubyParserBase` now only grows and fills the array when lines are being counted and skips it for `oneshot_lines` as well since `prepareCoverage` throws it away in that mode. The file is still registered in every mode so it shows up in `Coverage.result`.

core/src/main/java/org/jruby/ext/coverage/FileCoverage.java
sferik: Changed this to a plain `ArrayList`.

sferik:
@kares Thank you for the thoughtful review. I believe I have addressed all of your initial feedback. Please let me know if there are any additional changes you’d like me to make.

headius:
Master is 10.1 and that old branch should just be deleted. You did the right thing! I will be in flight soon and working most of the time so I'll try to have a look at these changes then.

headius:
I have removed the 10.1-dev branch to avoid future confusion.

kares:
> I believe I have addressed all of your initial feedback. Please let me know if there are any additional changes you’d like me to make.

Took a look, some of the changed are a bit invasive but don't really have an alternative moving forward atm.
Thing worth investigating is how MRI implements the coverage approach, they seem to have it more centralized.

If it's okay with @headius to ship like this and potentially do more refactoring later I think it's good enough...
esp. if you're Java/JVM experience is low, green field features are IMO a good starting point - great job. :heart_decoration: 

---

p.s. Had a bit of a struggle reading some of the technical comment text (seems like my comment about that got lost with GH), try to instruct AI to compact/clarify further, maybe to use simplified technical English.

headius:
I'll give it a look once I'm settled at my hotel, tonight or tomorrow. Thanks again for the effort @sferik!

sferik:
@kares Thanks again for your review.

I compared this against MRI (`ext/coverage`, `thread.c`, `vm_method.c`). MRI is centralized because its VM stores the method entry in every frame: the `CALL` hook reads `cfp->me` and bumps a hash counter, so no call path knows about coverage. JRuby frames carry only the implementation class and name, not the entry, so the hand-off has to start in the entry’s own `call` methods. Everything else already mirrors MRI: registration on `putMethod` matches `method_added` recording into `me_set`, forwarding entries resolve to the original, and results are built per file from the same keys.

I did reduce the number of call sites: `InterpretedIRMethod` and `MixedModeIRMethod` now do the hand-off from a single `prepareCall`, which also holds the debug/JIT preamble those overloads already repeated, instead of one line per arity overload. `CompiledIRMethod` keeps its per-overload line because each `invokeExact` has its own exact signature. I also rewrote all the comments this PR adds in shorter, plainer English.

sferik:
I found a few more bugs and pushed some fixes (and tests) to make this implementation consistent with CRuby. Let me know if you want me to squash all of these down into one commit.

@headius I believe this should be in pretty good shape for you to review. Much appreciated!

headius:
This is a big piece of work I think should delay until 10.1.3.0 (I'm hoping to release 10.1.2.0 shortly and this will not have enough bake time). I'm marking it as such and will review for merge soon after the release.

sferik:
@headius Congrats on [shipping 10.1.2.0](https://www.jruby.org/2026/09/21/jruby-10-1-2-0.html) on Monday. Are you able to review this patch now? Could you also please take a look at https://github.com/jruby/jruby/pull/9682?

PR #9678, opened 2026-09-14
Stop CGI::EscapeExt landing in CGI's singleton twice

### LINKED ISSUE #9251: Alias chaining does not properly restore old methods
In `test_cgi_escape.rb` there's a section that attempts to undefine the extension methods to force a test of the pure-Ruby versions. The logic looks like this:

```
class CGIEscapePureRubyTest < Test::Unit::TestCase
  def setup
    CGI::EscapeExt.module_eval do
      alias _escapeHTML escapeHTML
      remove_method :escapeHTML
      alias _unescapeHTML unescapeHTML
      remove_method :unescapeHTML
    end if defined?(CGI::EscapeExt) and CGI::EscapeExt.method_defined?(:escapeHTML)
  end

  def teardown
    CGI::EscapeExt.module_eval do
      alias escapeHTML _escapeHTML
      remove_method :_escapeHTML
      alias unescapeHTML _unescapeHTML
      remove_method :_unescapeHTML
    end if defined?(CGI::EscapeExt) and CGI::EscapeExt.method_defined?(:_escapeHTML)
  end

  include CGIEscapeTest::UnescapeHTMLTests
...
```

Unfortunately there appears to be two problems in JRuby with aliasing and re-aliasing the original method:

* The alias should not continue to re-wrap the old alias. Instead, it should use the original alias's contents as its payload.
* The re-aliasing of the original method does not appear to restore either the name or the class to use for super, leading to one of two errors:

Original name fails to be superable after aliasing:
```
CGIEscapeTest#test_cgi_escapeHTML:IBM037:
NoMethodError: super: no superclass method 'escapeHTML' for class CGI
    org/jruby/ext/cgi/escape/CGIEscape.java:402:in 'escapeHTML'
    /Users/headius/work/jruby/test/mri/cgi/test_cgi_escape.rb:208:in 'block in CGIEscapeTest'
```

Stack overflow due to superclass search not progressing (e.g. alias causes a lower class to be the current class, so super just keeps going back to the same class):
```
   org.jruby.dist/org.jruby.runtime.Helpers.invokeSuper(Helpers.java:955)
    org.jruby.dist/org.jruby.ext.cgi.escape.CGIEscape.cgiesc_unescape_html(CGIEscape.java:421)
    org.jruby.dist/org.jruby.ext.cgi.escape.CGIEscape$INVOKER$s$1$0$cgiesc_unescape_html.call(CGIEscape$INVOKER$s$1$0$cgiesc_unescape_html.gen)
    org.jruby.dist/org.jruby.internal.runtime.methods.JavaMethod$JavaMethodOneOrN.call(JavaMethod.java:966)
    org.jruby.dist/org.jruby.internal.runtime.methods.AliasMethod.call(AliasMethod.java:131)
    org.jruby.dist/org.jruby.internal.runtime.methods.AliasMethod.call(AliasMethod.java:131)
    org.jruby.dist/org.jruby.internal.runtime.methods.AliasMethod.call(AliasMethod.java:131)
    org.jruby.dist/org.jruby.internal.runtime.methods.AliasMethod.call(AliasMethod.java:131)
    org.jruby.dist/org.jruby.internal.runtime.methods.AliasMethod.call(AliasMethod.java:131)
    org.jruby.dist/org.jruby.internal.runtime.methods.AliasMethod.call(AliasMethod.java:131)
    org.jruby.dist/org.jruby.internal.runtime.methods.AliasMethod.call(AliasMethod.java:131)
    org.jruby.dist/org.jruby.internal.runtime.methods.AliasMethod.call(AliasMethod.java:131)
    org.jruby.dist/org.jruby.internal.runtime.methods.AliasMethod.call(AliasMethod.java:131)
    org.jruby.dist/org.jruby.internal.runtime.methods.AliasMethod.call(AliasMethod.java:131)
    org.jruby.dist/org.jruby.runtime.Helpers.invokeSuper(Helpers.java:955)
    org.jruby.dist/org.jruby.ext.cgi.escape.CGIEscape.cgiesc_unescape_html(CGIEscape.java:421)
```

These issues cause all of the tests to fail if the pure-Ruby tests run first due to the damage it causes to those aliased methods. I am unsure if this is a recent regression, since our older version of CGI::Escape avoided making these super calls.

sampokuokkanen:
cgi/escape.rb has already run "extend Escape" by the time it reaches load_ext, so extending CGI with EscapeExt and then prepending it to Escape left the module in the chain twice, and unescapeHTML's super could end up finding itself.

Closes #9251 (the alias handling from that deserves a seperate PR though).


headius (commented):
This might be ok but the explanation confuses me. 

In the C ext I see two module operations:

* EscapeExt is prepended into the Escape module
* CGI extends EscapeExt (i.e. its singleton includes it)

https://github.com/ruby/ruby/blob/e7551c08831c26cdfca6c9ca4bface71e621c93b/ext/cgi/escape/escape.c#L485

This matches what the JRuby ext was doing before this change.

The Ruby code in escape.rb matches JRuby:

* CGI includes Escape
* CGI extends Escape

https://github.com/ruby/ruby/blob/e7551c08831c26cdfca6c9ca4bface71e621c93b/lib/cgi/escape.rb#L15

So I believe the code as written should work correctly and I'm reluctant to patch around that without understanding why it doesn't behave correctly.

I believe the intent is that the Escape module, which contains the pure Ruby methods, always has EscapeExt prepended, so that if the extension is loaded it will immediately shadow the pure Ruby versions.

I'm on mobile so I can't test this, but it's also possible that recent improvements to prepend logic have fixed part of this.

headius:
Weirdly the CGI tests all seem to **pass** if I run with default settings (JIT, threshold 50 etc):

```
$ jruby test/mri/runner.rb test/mri/cgi 
Run options: 
  --seed=41502

# Running tests:

Finished tests in 3.825824s, 110.0417 tests/s, 342.4099 assertions/s.                                
421 tests, 1310 assertions, 0 failures, 0 errors, 0 skips
```

If I force it to interpreter-only or JIT-only (-X-C or -X+C), I get a bunch of stack overflows:

```
  1) Error:
CGIEscapeTest#test_cgi_unescapeHTML:ISO-2022-JP:
Java::JavaLang::StackOverflowError: 
    java.base/java.lang.String.lastIndexOf(String.java:2573)
    java.base/java.lang.String.lastIndexOf(String.java:2535)
    org.jruby.dist/org.jruby.runtime.Helpers.getSuperNameFromCompositeName(Helpers.java:1100)
    org.jruby.dist/org.jruby.runtime.ThreadContext.getFrameName(ThreadContext.java:734)
    org.jruby.dist/org.jruby.runtime.Helpers.checkSuperDisabledOrOutOfMethod(Helpers.java:1439)
    org.jruby.dist/org.jruby.runtime.Helpers.invokeSuper(Helpers.java:1159)
    org.jruby.dist/org.jruby.ext.cgi.escape.CGIEscape.cgiesc_unescape_html(CGIEscape.java:421)
    org.jruby.dist/org.jruby.ext.cgi.escape.CGIEscape$INVOKER$s$1$0$cgiesc_unescape_html.call(CGIEscape$INVOKER$s$1$0$cgiesc_unescape_html.gen)
    org.jruby.dist/org.jruby.internal.runtime.methods.JavaMethod$JavaMethodOneOrN.call(JavaMethod.java:828)
    org.jruby.dist/org.jruby.internal.runtime.methods.AliasMethod.call(AliasMethod.java:131)
    org.jruby.dist/org.jruby.runtime.Helpers.invokeSuper(Helpers.java:1171)
    org.jruby.dist/org.jruby.ext.cgi.escape.CGIEscape.cgiesc_unescape_html(CGIEscape.java:421)
    org.jruby.dist/org.jruby.ext.cgi.escape.CGIEscape$INVOKER$s$1$0$cgiesc_unescape_html.call(CGIEscape$INVOKER$s$1$0$cgiesc_unescape_html.gen)
    org.jruby.dist/org.jruby.internal.runtime.methods.JavaMethod$JavaMethodOneOrN.call(JavaMethod.java:828)
    org.jruby.dist/org.jruby.internal.runtime.methods.AliasMethod.call(AliasMethod.java:131)
    org.jruby.dist/org.jruby.runtime.Helpers.invokeSuper(Helpers.java:1171)
    org.jruby.dist/org.jruby.ext.cgi.escape.CGIEscape.cgiesc_unescape_html(CGIEscape.java:421)
    org.jruby.dist/org.jruby.ext.cgi.escape.CGIEscape$INVOKER$s$1$0$cgiesc_unescape_html.call(CGIEscape$INVOKER$s$1$0$cgiesc_unescape_html.gen)
...
```

Your patch allows all modes to pass, but there's something else happening here I can't explain.

headius:
@sampokuokkanen Could it have simply been a matter of bad ordering in the extension's extend and prepend?

https://github.com/jruby/jruby/pull/9685

sampokuokkanen:
This fails:
```ruby
module A; def m = 1; end
module B; def m = super; alias x m; alias m x; end
o = Object.new.extend(B, A)
A.prepend B
p o.m
```

Seems like it is a problem when calling an aliased method. It requires a module to appear twice in the ancestor chain and aliasing (with super), but leads to a `StackOverflowError`.  So the reordering here just sidesteps the bug. Might not be the most urgent bug to fix though.

headius:
The secondary bug remains and should be looked at, but I'm ok with "sidestepping" the issue by fixing the ordering in #9685 (and that aligns our code better with CRuby).

PR #9679, opened 2026-09-14
Don't wrap an alias in another alias

sampokuokkanen:
Aliasing an alias pointed the new one at the old alias instead of at what the old one pointed to, so every link stayed on the call path. Alias a method away and back, the way setup and teardown do, and the chain grew two links a round, never shrank, and eventually ran the stack out.

putAlias unwraps first now. It still marks the alias it skips as aliased, since defineAlias reads that flag before warning about a redefinition, and getOldName has no chain left to walk.

The idea and much of the code is by @headius and comes from here: https://github.com/jruby/jruby/issues/9251#issuecomment-3919047965


headius (approved):
Looks good, and thanks for revisiting my one-off patch.

PR #9682, opened 2026-09-14
Fix line coverage of statements inside string interpolation and heredocs

sferik:
Coverage counted a statement holding a heredoc twice per execution when the heredoc's interpolation sat on a later line, and marked the interpolation's line as code although nothing ever counted there. A statement's coverage event is now emitted once, when the statement starts. The call site only refreshes the line for backtraces (`determineIfWeNeedLineNumberForCall`). Also, a lone statement inside `#{}` is not a line event. The string it is part of is, so the parser no longer leaves that line marked as code (`uncoverLine`). Several statements in one interpolation each remain a line event.


core/src/main/java/org/jruby/parser/RubyParser.y
```diff
@@ -4027,7 +4027,12 @@ string_content  : tSTRING_CONTENT {
                    p.setHeredocLineIndent(-1);
 
                    /*%%%*/
-                   if ($6 != null) $6.unsetNewline();
+                   if ($6 != null) {
+                       // A lone statement in an interpolation is not a line event of its own (MRI); the
+                       // string it is part of is. Several statements in one interpolation each remain one.
+                       $6.unsetNewline();
+                       p.uncoverLine(@6.start());
```
headius: I cannot find equivalent logic in CRuby for this; the lines related only unset newline and have no additional logic:

https://github.com/ruby/ruby/blob/b2ec82d4ea84f5526b3759923159a039bc5c1d3b/parse.y#L6088-L6089

I'd like to understand why we need this and CRuby does not appear to do the same.

headius (commented):
Calling all @enebo's for a secondary review. This might be fine but it diverges from parse.y and I want to understand why.

headius:
Let's get this in with the other coverage work in 10.1.3.0.

core/src/main/java/org/jruby/parser/RubyParser.y
sferik: CRuby gets the same result from `nd_unset_fl_newline` alone because it decides which lines are coverable at compile time, from the newline flag. `iseq_compile_each` adds `RUBY_EVENT_COVERAGE_LINE` only to nodes that still have the flag:

https://github.com/ruby/ruby/blob/b2ec82d4ea84f5526b3759923159a039bc5c1d3b/compile.c#L11045-L11052

and `iseq_set_sequence` sets a line's entry from `nil` to `0` only when that event lands on an instruction:

https://github.com/ruby/ruby/blob/b2ec82d4ea84f5526b3759923159a039bc5c1d3b/compile.c#L2713-L2720

So once the parser clears the flag, the line was never coverable.

JRuby marks coverable lines earlier, in the parser. `newline_node` sets the flag and marks the line at the same time. It has to happen there because method bodies are built lazily, so the IR can't mark lines for methods that never run. By the time this action clears the flag, the line is already marked `0`, and it would show up as code that never ran where CRuby shows `nil`. The extra call undoes what `newline_node` did.

Your question also led me to a bug in that call. It set the line back to `nil` without checking anything, which wiped out marks left by other statements on the same line. It now only undoes the mark the lone statement added itself, and only if nothing else had marked that line (db35bd179c).

CRuby also counts a statement on the line of the first instruction after its line event, not on the line where the statement starts. For example, `x =\n  y.size` counts on line 2, and `x = <<~EOS` counts on the line of its first interpolation. JRuby treated this differently in several ways, so I pushed a few more commits:

* dbf912ccd2: a statement nested in another (such as one of several statements in `#{}`) no longer replaces the outer statement's pending line event.
* 30039a8818: a call built before a statement's first instruction no longer moves the statement's count to an unmarked line. Most multi-line statements used to read `[1, 0, 0]`. Backtrace lines don't change.
* 814037bd26: multi-line array, hash, string, regexp and symbol literals get the line they start on.
* 92a5b705d5: a new `LineEvents` helper finds each statement's first instruction. The parser, `Coverage.line_stub` and the IR builder all use it, so marked lines and counted lines agree.
* a76bc297c7: the coverage array is no longer one line short when a heredoc ends the file.

sferik:
> Let's get this in with the other coverage work in 10.1.3.0.

Sounds good! I also have a branch for `branch-coverage` that is based on the `method-coverage` branch, as well as some fixes to oneshot and multiline line coverage. Once all these are merged, we should be able to completely delete `test/mri/excludes/TestCoverage.rb`.

If we can ship all of that in 10.1.3.0, I think it will be the biggest change to JRuby’s Coverage support since it was introduced!

sferik:
I pushed two more commits:

- 741d5e483f: an `undef` statement now takes the line of its `undef` keyword, not the line of the first name it undefines. Before, `undef\n  a` was counted on the second line. In `x = 1; undef\n  a` it was also counted as a separate line event, where CRuby merges it with the other statement on the first line.
- 6f7ba4dab7: each arm of a ternary is now a line event, as in CRuby. In `x = y.size ?\n  1 : 2`, the second line used to read `nil`.

PR #9683, opened 2026-09-14
Fix bin/* patterns in .gitignore

jcharaoui:


jcharaoui:
Some context here: I wanted to leverage the `.gitignore` file in Debian package builds, where the source package doesn't include the upstream `.git`, to cleanup the build directory post-build. To do this I initialize a bare git repo with no history and run `git clean -Xdf` which unexpectedly removes the entire `bin` directory. Moving the `bin/*` pattern above the `!bin/<file>` negations fixed the problem, while revealing that a couple of removed scripts like `bin/rake` where still being referenced in it.

headius:
Yeah I think this makes sense, and I have missed legitimate edits to bin files because of this blanket exclusion. I'll review.

headius (approved):

PR #9685, opened 2026-09-14
Fix setup sequence for cgi/escape

### LINKED ISSUE #9251: Alias chaining does not properly restore old methods

headius:
The order is important: the `Escape` module must prepend `EscapeExt` before `EscapeExt` gets extended into the CGI class if they happen in the reverse order, the search logic will not see it and ends up searching twice and improperly re-starting the super chain from the lower include.

This is possibly a bug in how we handle prepend and include when the modules being incorporated also have their own includes or prepends, but this fix allows all cgi/escape tests to pass and matches the CRuby order during setup.

This combined with aliasing fixes in jruby/jruby#9679 resolves jruby/jruby#9251.

Fixes jruby/jruby#9251


headius:
If backported, this PR likely also needs #9679 to go with it.

PR #9687, opened 2026-09-15
Hold ObjectSpace::WeakMap keys weakly, by identity

### LINKED ISSUE #8456: ObjectSpace::WeakMap is double-weak in CRuby
It appears that at some point the contract of ObjectSpace::WeakMap changed so that both the values *and* the keys are weak. CRuby may have had this behavior all along by accident, due to their using object pointers for keys (in such a way that they won't be prevented from being collected.

The documentation for WeakMap in CRuby was changed in 2023 to indicate that all object references (keys and values) are weak. Previous documentation did not make that clear.

https://github.com/ruby/ruby/commit/39c072d6f7846a99f64e52090a320576715325c5#diff-14642ca05e367c9201974b1641f91730c15bde48eebe09f9494f5b7ecb42b2fbR983

At least one location in Rails assumes this weak-keyed behavior and falls back on an `object_id`-based map when running on JRuby:

https://github.com/rails/rails/blob/bc979c5cf4e29e2aba95b7b13ca53e7a02fddee2/activesupport/lib/active_support/descendants_tracker.rb#L23-L33

This code is not in the same place in recent versions, but may have been relocated.

We should emulate the CRuby behavior with a weak-keyed, weak-valued implementation of WeakMap.

aminmansuri:
Fixes #8456.

`ObjectSpace::WeakMap` held its keys strongly, so any object used as a key lived as long as the map; every `WeakRef` ever created stayed alive, since `weakref.rb` keys its map by the `WeakRef` itself. The documented contract is that keys and values are both weak and that keys compare by identity.

The identity-keyed half of `WeakMap` now uses a new `WeakKeyValueIdentityMap`: a weak key and a weak value per entry, identity hash, concurrent backing map, cleared references pruned on access and during iteration. `WeakKeyMap` is untouched.

What a caller can notice:

- Integer and Float keys stay strong and compare by value, as immediates are never collected; keying a map by `object_id` keeps working.
- A key nothing else references no longer keeps its entry.
- Iterating while another thread inserts no longer raises.
- On very large maps, inserts and lookups cost more.
- The numeric half no longer evicts a value put right after its predecessor was collected.

Tests: `test/jruby/test_weakmap.rb` (explicit JVM collection) and a JUnit test for the eviction fix.


headius:
I could have sworn I fixed this but I guess not. I'll review. 

Probably didn't strike me as a high priority since it was an internal API for so long.

aminmansuri:
Note for completeness: I did some performance and size tests on local (MacOS) machine. The new ConcurrentHashMap based + WeakRef solution increases the size of the WeakMap a bit. It also can be slightly slower in some circumstances but it's much faster in shared circumstances.

I considered using just a lock based solution but the gain was minimal, and it was thought that these WeakMaps tend to be small anyway so the overall overhead is negligible.

headius:
I'm not super concerned about the size or performance of the weak maps at this point. I still believe it was the wrong approach to expose this rather than just implementing a proper WeakRef + ReferenceQueue like the JVM does (you can implement everything else on top of that).

If WeakRef becomes a performance issue because of the weak map implementations, it's just more justification for us to go back to using java.lang.ref.WeakReference instead of this hacked-together alternative.

test/jruby/test_weakmap.rb
```diff
@@ -0,0 +1,54 @@
...
+    ref = make_unreferenced_weakref(LIVE)
+    assert gc_until { ref.refersTo(nil) }, "WeakRef instance stayed alive"
+  end
+
+  def test_immediate_keys_stay
+    map = ObjectSpace::WeakMap.new
+    map[LIVE.object_id] = LIVE
+    map[:sym] = LIVE
+    map[nil] = LIVE
+    3.times { java.lang.System.gc }
```
headius: shouldn't this use `gc_until` from below? The "until" would be "until enough time or iterations have passed that we're sure the reference isn't getting cleared".

test/jruby/test_weakmap.rb
```diff
@@ -0,0 +1,54 @@
...
+  end
+
+  def test_weakref_instance_is_collected
+    ref = make_unreferenced_weakref(LIVE)
+    assert gc_until { ref.refersTo(nil) }, "WeakRef instance stayed alive"
+  end
+
+  def test_immediate_keys_stay
+    map = ObjectSpace::WeakMap.new
+    map[LIVE.object_id] = LIVE
```
headius: This seems to be the only Integer test... I'd like to see more proving of Integer and Float keys to ensure they truly work correctly.

headius (commented):
Pretty close to ready. The implementation is pretty textbook, even though the use of ConcurrentHashMap makes it a little fat. You might consider just using a synchronized map again; CHM is really heavy for small collections (which hopefully we won't see many of during a given run).

A few comments on testing but otherwise ok. 👍

aminmansuri:
I did look at the weight and perf tradeoffs.. but I believe the ConcurrentHashMap is the better approach.

test/jruby/test_weakmap.rb
aminmansuri: I improved the test in [3988d97]

test/jruby/test_weakmap.rb
aminmansuri: Strengthened here [4c1e181]

aminmansuri:
I added some further improvements to the tests to make them catch more cases.

New edits ensure that integers and floats find same object by comparing by value. So 1 is the same as 1+0 to preserve CRuby behavior.

headius (approved):
A purpose-built collection that is itself a concurrent map would be lighter-weight, but this is good for now to get the behavior fixed.

PR #9691, opened 2026-09-15
Support zero-size IO::Buffers

sampokuokkanen:
`IO::Buffer.new(0)` is now a null buffer like in CRuby, and null buffers act as empty instead of raising AllocationError or a NullPointerException. `copy`, `set_string` and `slice(offset)` also check their ranges like CRuby.

The second commit: fixes `read` and `pread` under a Fiber scheduler without `io_read`/`io_pread`, which returned a Java null instead of reading.

Enables three excluded MRI tests and removes seven spec tags.


core/src/main/java/org/jruby/RubyIOBuffer.java
```diff
@@ -679,6 +679,9 @@ public IRubyObject slice(ThreadContext context, IRubyObject _offset, IRubyObject
     public IRubyObject slice(ThreadContext context, int offset, int length) {
         validateRange(context, offset, length);
 
+        // only a zero-length slice gets past validateRange on a null buffer, and like CRuby it is null too
+        if (base == null) return new RubyIOBuffer(context.runtime, context.runtime.getIOBuffer());
```
headius: Could this just return `this`? Or is the contract that slice always returns a new Buffer? The zero-length buffer is basically immutable, but perhaps there's other side effects that could be visible.

core/src/main/java/org/jruby/RubyIOBuffer.java
```diff
@@ -779,6 +791,8 @@ private ByteBuffer getBufferForWriting(ThreadContext context) {
             return base;
         }
 
+        if (size == 0) return ByteBuffer.allocate(0);
```
headius: Again I wonder if we could cache a single zero-length ByteBuffer and reuse it.

headius (commented):
Probably would be fine to merge as-is but also seems like we could reduce some allocation for slicing zero lengths.

core/src/main/java/org/jruby/RubyIOBuffer.java
sampokuokkanen: `b.slice.resize(4)` would then resize the root instead of just the slice, so it has to be a new buffer, but it now carries READONLY like CRuby does.

core/src/main/java/org/jruby/RubyIOBuffer.java
sampokuokkanen: Fixed!

core/src/main/java/org/jruby/RubyIOBuffer.java
headius: Ahh resize... ok, makes sense then.

maven/jruby-jars/bin/lib/jruby-jars.rb
```diff
@@ -0,0 +1,13 @@
+require 'jruby-jars/version'
```
headius: Most recent commit appears to have copied a bunch of files into maven/jruby-jars/bin.

headius (changes requested):
Base PR is fine but what copied all those files? That obviously needs to be removed.

maven/jruby-jars/bin/lib/jruby-jars.rb
sampokuokkanen: Sorry, committed them by accident. I added them to my .git/info/exclude so will not happen again.

headius: I'm confused how they got there though... it looks like the entire contents of the jruby-jars maven project got recopied into its bin directory. That's certainly not something we do in any of our builds.

headius (approved):

headius:
Thanks again... but do figure out how those files got copied around. I'd hate for it to be something wrong with our builds.

maven/jruby-jars/bin/lib/jruby-jars.rb
sampokuokkanen: I'll look into it a bit, they've been showing up for me most of the time. Maybe I have something set up wrong.

Issue #9667, opened 2026-09-11
Allow specifying AppCDS cache location and use safer default


headius:
JRuby is frequently installed via a package manager that makes the install location read-only. This prevents the AppCDS cache (and any future caches stored in `lib`) from being writable, leading to warnings or errors at startup.

We need to provide a standard way to specify the location of the AppCDS `jsa` cache file. Additionally, this cache should probably default to somewhere under `$HOME/.cache` as is now common on most *nix platforms.

headius:
I like JRUBY_CACHE_HOME as an env var after looking [this Debian patch](https://sources.debian.org/src/jruby/9.4.14.0%2Bds-1/debian/patches/0013-fix-default-appcds-cache-path.patch?pow_referer=https%3A%2F%2Fgithub.com%2Fjruby%2Fjruby%2Fissues%2F9413) that mentions XDG_CACHE_HOME.

I'm also coming around to defaulting to a subdirectory of `$HOME/.cache` which seems to be a standard these days. We would create directories as necessary in that location to separate JRuby AppCDS caches (and other JRuby caches in the future, like Prism parser dumps) from other non-JRuby caches. We would also need to continue separating those caches by the location of jruby.jar and the JDK used.

Note that the Debian patch above also defaults to a location under `$HOME/.cache`.

Additional debate: how much of this should ship with JRuby 10.0.x given that it is a stable branch. The config variable would be purely additive, but the new default cache location would be a surprise.

headius:
Since we only use the cache location for JSA right now, I have gone with `JRUBY_JSA_HOME` in #9698. That would not conflict with a future general "cache home" variable, and may eventually be defunct if we start using the new AOT features of Java 25+.

We'll call this one resolved with #9698 and if there's a need for a more general cache location config in the future we'll cross that bridge.

headius:
Fixed by #9698.

PR #9692, opened 2026-09-15
Move default JSA path to $HOME/.cache

headius:
In order to avoid writing into the installed JRuby directory, we
move the default location to the .cache directory in the user's
$HOME, where most programs store such caches these days.

Relates to the new JSA home config from jruby/jruby#9698


headius:
See #9667 which this is intended to fix.

headius:
I'm going to hold off on the default location change for now and merge the config for both 10.0 and 10.1.

I like the new default but it's a pretty visible change that should probably bake in `jruby-head` for a while (and I don't think it should be backported to 10.0).

headius:
This now contains just the default location change (to $HOME/.cache), with the configurable JSA home merged from #9698. Will merge after 10.1.2.0 and let it bake in `jruby-head` for a bit.

jcharaoui:
I wonder if it's best to assume `$HOME/.cache` exists here. Other platforms like Mac have a different path for such data, and depending on the host, the operator might have defined it otherwise using the `XDG_CACHE_HOME` environment variable.

headius:
Well, that is a fair point, but is it worse than assuming the JRuby install location is writable? I'm trying to find a better option that allows users to get the benefit of the JSA file without requiring that.

Presumably this is also why the Debian package defaults to .cache. I want to reduce the amount of third-party patching required to maintain JRuby packages.

headius:
FWIW on my MacOS machine I do have a .cache directory, containing several Ruby-related items and a few from packages I've installed with Homebrew:

```
$ ls -1 ~/.cache                    
bashbrew
codex-runtimes
gem
install4j
javasharedresources
jruby
opencode
rbs
rubocop_cache
ruby-install
uv
vllm
```

Interestingly, the `javasharedresources` directory is put there by IBM's OpenJ9 JVM (Semeru) as a place to store that JVM's equivalent of AppCDS JSA files:

https://github.com/eclipse-openj9/openj9/pull/2862

PR #9696, opened 2026-09-16
Set the aliased flag on the real method, not the wrapper

### LINKED ISSUE #9695: require "json" warns about redefined to_hash with -w in JRuby 10.1.1.0
## Description

JRuby 10.1.1.0 emits a method redefinition warning when requiring `json`
with verbose warnings enabled. The same code does not emit the warning on
JRuby 10.1.0.0.

This appears to be a regression related to the alias redefinition warning
changes in 63a0ede02b.

## Minimal reproduction

```sh
jruby -w -e 'require "json"
```

On JRuby 10.1.1.0:

```
.../json/ext/generator/state.rb:72: warning: method redefined; discarding old to_hash
```

Without `-w`, no warning is emitted.

## Expected behavior
Requiring `json` should not emit a method redefinition warning.

## Workaround
Suppressing verbosity only while requiring `json` avoids the warning:

```ruby
verbose = $VERBOSE
$VERBOSE = nil
begin
  require "json"
ensure
  $VERBOSE = verbose
end
```

sampokuokkanen:
[63a0ede02b](https://github.com/jruby/jruby/commit/63a0ede02b) started marking `entry.method` and checking `method.isAliased()`. An `AliasMethod` wrapper is never itself marked, so aliasing over an existing alias warned when it shouldn't. CRuby keeps the flag on the definition, shared by the original and its aliases; `getRealMethod()` is the equivalent.

Closes https://github.com/jruby/jruby/issues/9695.


headius (approved):
Arguably these methods should delegate to the wrapped method, but perhaps being explicit here is better.

PR #9700, opened 2026-09-16
Ignore generated build artifacts

sampokuokkanen:
I mistakenly committed build output when working on https://github.com/jruby/jruby/pull/9691, this is an attempt to fix that happening to anyone else. It seems to have been caused by the Java VS Code extension, which imports maven/jruby-jars as a Maven project.


headius:
This is a very weird side effect of the VS Code plugin. Why would it copy the entire contents of `maven/jruby-jars` into `maven/jruby-jars/bin`? That doesn't make any sense.

headius:
If this is a problem with the jruby-jars maven project we should figure out how to fix that rather than patching around it by ignoring copied files.

sampokuokkanen:
Got it, I'll look into jruby-jars maven project and VS Code Java extension tomorrow.

PR #9693, opened 2026-09-15
Pass unions by value with the C ABI's register classes

aminmansuri:
Fixes #9329.

Makes JRuby pass C unions by value the way a C compiler does. 

Before, every 8-byte union went in an integer register, so a union of floats came back as garbage and a callback receiving one crashed the JVM. 

Now the union's members decide the register class, on x86_64 and ARM64.

PR #9702, opened 2026-09-16
Pass unions by value with the C ABI's register classes

### LINKED ISSUE #9329: Wrong union-by-value argument passing in JRuby
As part of https://github.com/ffi/ffi/pull/1178 we added some more tests of union-by-value function arguments. There are some new specific measurements to fix ABI conventions on x86_64 and ARM64. See the description and code of the PR.

We also added tests to verify more union-by-value situations. Unfortunately they fail on JRuby, at different test cases on x86_64 and ARM64.

[specs (ubuntu-latest, jruby-head)](https://github.com/ffi/ffi/actions/runs/23118355929/job/67147777645?pr=1178#logs)
```

Failures:

  1) Union by_value should correctly align/write a float value to union_float_test_t by value
     Failure/Error: expect((@u[name] - ByValueLibTest.send("#{uni}_val_align_#{k}", @u)).abs).to be < 0.00001

       expected: < 1.0e-05
            got:   2.71598704641059e+36
     # ./spec/ffi/union_spec.rb:141:in 'block in UnionSpec'

  2) Union by_value should correctly align/write a double value to union_float_test_t by value
     Failure/Error: expect((@u[name] - ByValueLibTest.send("#{uni}_val_align_#{k}", @u)).abs).to be < 0.00001

       expected: < 1.0e-05
            got:   2.0000000000001
     # ./spec/ffi/union_spec.rb:141:in 'block in UnionSpec'

  3) Union by_value should read a float value from memory of union_float_test_t by value
     Failure/Error: expect((@u[name] - val).abs).to be < 0.00001

       expected: < 1.0e-05
            got:   3.444
     # ./spec/ffi/union_spec.rb:152:in 'block in UnionSpec'

  4) Union by_value should read a double value from memory of union_float_test_t by value
     Failure/Error: expect((@u[name] - val).abs).to be < 0.00001

       expected: < 1.0e-05
            got:   2.0000000000001
     # ./spec/ffi/union_spec.rb:152:in 'block in UnionSpec'
```

[specs (ubuntu-24.04-arm, jruby-head)](https://github.com/ffi/ffi/actions/runs/23118355929/job/67147777735?pr=1178#logs)
```
Failures:

  1) Union by_value should return a union of doubles by value
     Failure/Error: expect(u[:coord][:x]).to eq(1.5)

       expected: 1.5
            got: 1.4978809957e-313

       (compared using ==)
     # ./spec/ffi/union_spec.rb:162:in 'block in UnionSpec'

  2) Union by_value should pass a union of doubles by value
     Failure/Error: expect(ByValueLibTest.union_double_get_x(u)).to eq(10.0)

       expected: 10.0
            got: 1.4979119926e-313

       (compared using ==)
     # ./spec/ffi/union_spec.rb:170:in 'block in UnionSpec'

  3) Union by_value should pass and return a union of doubles by value
     Failure/Error: expect(r[:coord][:x]).to eq(1.5)

       expected: 1.5
            got: 1.49791841507e-313

       (compared using ==)
     # ./spec/ffi/union_spec.rb:180:in 'block in UnionSpec'

  4) Union by_value should access union doubles via array field
     Failure/Error: expect(u[:v].to_a).to eq([1.5, 2.5, 3.5, 4.5])

       expected: [1.5, 2.5, 3.5, 4.5]
            got: [1.4979237569e-313, 1.38697476291614e-309, 1.78e-321, 1.386978271489887e-309]

       (compared using ==)
     # ./spec/ffi/union_spec.rb:188:in 'block in UnionSpec'
```

For now we'll disable the specs in the FFI project https://github.com/ffi/ffi

**Environment Information**
  
Provide at least:
  * jruby 10.1.0.0-SNAPSHOT (4.0.0) 2026-03-15 6fbd5bc91d OpenJDK 64-Bit Server VM 21.0.10+7-LTS on 21.0.10+7-LTS +indy +jit [x86_64-linux]
  * jruby 10.1.0.0-SNAPSHOT (4.0.0) 2026-03-15 6fbd5bc91d OpenJDK 64-Bit Server VM 21.0.10+7-LTS on 21.0.10+7-LTS +indy +jit [aarch64-linux]

**Expected Behavior**

* FFI should pass the union as value the same way as a C compiler
  * `rake compile spec` in the FFI project https://github.com/ffi/ffi

**Actual Behavior**

* The union by value is passed in some other way than the C compiler. The ABI seems to be different.

aminmansuri:
Fixes #9329.

Makes JRuby pass C unions by value the way a C compiler does. 

Before, every 8-byte union went in an integer register, so a union of floats came back as garbage and a callback receiving one crashed the JVM. 

Now the union's members decide the register class, on x86_64 and ARM64. 

Same commits as #9693, rebased onto master as requested.


larskanis:
Thank you @aminmansuri for fixing this issue! It looks similar complex as https://github.com/ffi/ffi/pull/1178 , so there seems to be no easier way to implement these unions. Libffi should directly support unions, IMHO. 

While implementing this, did you notice anything that should be ported back to or improved in the ffi project?

aminmansuri:
Yeah. I agree that union support would simplify things.

I think what would be most useful is:

1. **Per-cell classification on x86_64.** Replace "every leaf is floating, so use a float filler sized by alignment" with a decision per 4-byte cell, floating only where every field overlapping that cell is floating. That fixes the mixed unions and the union-at-offset-4 case that the gem currently gets wrong.
2. **Restrict the float filler to the platforms whose ABI floats unions.** Apply the homogeneous-float rule only on aarch64, arm, ppc64le and x86_64, and keep the integer filler on riscv64, loongarch64 and s390x, where unions always go in integer registers.

headius:
Yeah all the platform-specific bits are clearly something that should live in libffi.

In our case, I'm thinking most of that should probably move into jnr-ffi or jffi, but I haven't thought through it much yet.

headius:
Here's the class in Java's FFM that handles union layouts:

https://docs.oracle.com/en/java/javase/25/docs/api/java.base/java/lang/foreign/UnionLayout.html

Eventually this will be moot once we finally land jnr-ffi support for FFM, but we're not there yet.

@aminmansuri @larskanis if you're interested in the FFM-based jnr-ffi check out the branch here: https://github.com/jnr/jnr-ffi/tree/panama-22-unsquashed

aminmansuri:
Would you take this in JRuby now and move it to jnr-ffi later, or should it go to jnr-ffi directly?

PR #9706, opened 2026-09-17
Advertise Matrix chat room instead of Libera

jcharaoui:
The #jruby channel on Libera hasn't been effectively bridged with Matrix in a long time, so it might be time to advertise a chat room where the JRuby team members are actually reachable.


jcharaoui:
Alternatively: fix the IRC bridge! :sweat_smile:

headius:
I would love to fix the IRC bridge, and I probably should get my account active in that room anyway.

Matrix is definitely the preferred chat though for development. We're working on Discord and Slack options for folks who prefer those.

PR #9716, opened 2026-09-19
Raise IOError, not NullPointerException, when an IO is closed during accept

### LINKED ISSUE #9714: IO#accept raises NullPointerException instead of IOError when the socket is closed concurrently
`OpenFile#channel()` reads the non-volatile field `fd` twice: `checkClosed()`
first, then `return fd.ch`. Another thread closing the IO sets `fd = null`
(OpenFile.java:953) between the two reads, so the caller gets a Java NPE
rather than the IOError the check was there to raise.

    Java::JavaLang::NullPointerException:
      Cannot read field "ch" because "this.fd" is null
      org/jruby/util/io/OpenFile.java:2584 in channel
      org/jruby/RubyIO.java:481 in getChannel
      org/jruby/ext/socket/RubyTCPServer.java:292 in getServerSocketChannel
      org/jruby/ext/socket/RubyTCPServer.java:159 in accept

Reproduced on jruby-10.0 at 6ab9bcc325, JDK 25: the suite's own
`test_accept_on_closed_server_raises_ioerror` (test/jruby/test_io.rb) fails
about 1 run in 50, and 8 in 10 with a `Thread.yield()` inserted between the
check and the dereference. It shows up as an intermittent CI failure on
`rake test:jruby:jit`.

The window appeared in b84d7839c2 (the fix for #9324), which replaced
`assert(fd != null)` with `checkClosed()`; the assert had the same two-read
shape but was disabled in normal runs.

`ready()` dereferences `fd` with no check at all, and `setBlocking()` has the
same pattern.

Fix: read the field once. PR to follow.

aminmansuri:
Fixes #9714

`OpenFile.channel()` called `checkClosed()` and then read `fd.ch`: two separate reads of a field a concurrent `close` nulls, so `TCPServer#accept` raises `NullPointerException` instead of the `IOError` callers rescue. The same was in `readChannel()`…`socketChannel()`, in `ready()` (which dereferenced `fd` with no check at all) and in `setBlocking()`. All of them now take a single read, through a new private `checkedFD()`.

Measured on `jruby-10.0` at 6ab9bcc, running `test/jruby/test_io.rb -n test_accept_on_closed_server_raises_ioerror`:

- **1 failure in 50 runs before the change, 0 in 200 after.**
- With a `Thread.yield()` inserted between the check and the dereference to widen the window: **2 of 10 runs passed before, 10 of 10 after.**

`test/jruby/test_io.rb` (48 tests) and `test/jruby/test_socket.rb` (44 tests, multicast excluded — it blocks in `DatagramSocket.receive` on this machine) give identical results either side of the change.

Edit: made `fd` volatile to garantee narrow CRuby guarantee of being able to close it from another thread. 

**History of this bug**

Present in 9.1.17.0:

```Ruby
public Channel channel() {
    assert(fd != null);
    return fd.ch;
}
```

- Until March 2026, channel() had only a disabled assertion, so a concurrent close gave a raw null dereference with no check at all. Existed in 9.1.17.0 from 2018, and it dates from the 9.x rewrite of the IO layer; an NPE at the very same place was being patched around as far back as 2015.
- March 2026: @kares added a real closed check, which fixed the common case, and left a narrow window between the check and the use.
- Now: this change reads the field once and makes it volatile, closing that window and the visibility hole, and does the same for two other sites.

**KUDOS:** to the test suite for finding this obscure bug.


aminmansuri:
TestOpenURISSL#test_bad_ssl_version
  exceptions on 1 threads:
  RubyTCPServer.java:152 in 'accept': stream closed in another thread (IOError)
    from openssl/ssl.rb:585 in 'accept'
    from test/mri/open-uri/utils.rb:280 in 'block in start'

The test's helper runs a server thread looping on accept and then closes the server from the main thread. With the change, that close now surfaces as IOError out of accept, and the helper does not rescue it, so the test reports an exception on a thread.

That is the change working as designed, one layer too widely: we intended the NPE case to become IOError, and we also converted cases that previously did not raise at all, most likely through ready() or setBlocking(), which had no check before.

Fixing.

aminmansuri:
there still seems to be a flake there.. I'll investigate that separately.

aminmansuri:
test_accept_on_closed_server_raises_ioerror ran 250 times on an unmodified build without a single failure.

To reproduce I had to change the following inside OpenFile#channel():

public Channel channel() {
checkClosed(); // reads fd, sees it non-null
// <- here
return fd.ch; // reads fd again, now null -> NPE
}

Thread.sleep(2) added: 6 of 6 runs raise the NPE, 0 of 6 with the
field read once.
Thread.yield() added: 9 of 40 against 0 of 40.

aminmansuri:
Some multithreaded flakiness here. 

`fd` must be volatile since it's read by different threads.

Also we need to take care to read it once so that we don't get a different one when several threads are involved.

aminmansuri:
ALL GREEN!

headius (approved):
Yeah this makes sense. Simple thread race and simple fix (get fd into local var before checking null and using it).

A bonus improvement might be to make all accesses of this variable be atomic, but that might also be more overhead than it is really worth.

aminmansuri:
Technically the read of fd. Was already "atomic" (since the JDK doesn't permit a half read of a reference).

The issue here was that the code was reading fd twice, so if someone changed the fd pointer then you'd get different fd references causing a NullPointerException.

The point of volatile here was to impose visibility and ordering  so the JIT doesn't keep the value in a register. 

What AtomicReference would give us here is a CAS operation. But we don't need a CAS here. We don't have a read-modify-write situation for fd so there's nothing here for it to protect. 

So volatile is sufficient and preferable for this situation.

Issue #9715, opened 2026-09-19
CI jobs fail on transient network errors during bootstrap and bundle install


aminmansuri:
A download glitch makes the whole job fail. Happened in 35284429020:

* `bundle install` — `Could not reach host index.rubygems.org`, exit 17.
  Bundler's own retry has no backoff, so four attempts are spent in
  milliseconds.
* `bootstrap` — `Could not transfer artifact rubygems:bundler:gem:2.7.2`,
  `SocketException: Connection reset`, from the mavengem wagon. The maven
  side of #9493.

Not PR-specific: three plain pushes to `jruby-10.0` failed the same way —
runs 33791374721, 34624974473, 34646215480.

Suggested fix: wrap the two network-bound steps in a retry with backoff
(3 attempts, 15s then 30s) instead of relying on each tool's own. 

PR to follow.

aminmansuri:
These flaky tests were driving me nuts.

Issue #9713, opened 2026-09-18
Address flaky concurrent-ruby tests


headius:
These could be bad tests or unreliable behavior in JRuby (or something flaky with GHA) but we should investigate them. This issue will be updated for a while with any more concurrent-ruby failures, because I've seen a few different ones that come and go.

Here's one to start:

```
  1) Concurrent::Actor termination terminates with all its children
     Failure/Error: expect(subject.ask!(:terminated?)).to be_truthy

       expected: truthy value
            got: false
     # /home/runner/work/jruby/jruby/lib/ruby/gems/shared/gems/rspec-support-3.13.7/lib/rspec/support.rb:110:in 'block in Support'
     # /home/runner/work/jruby/jruby/lib/ruby/gems/shared/gems/rspec-support-3.13.7/lib/rspec/support.rb:119:in 'notify_failure'
     # /home/runner/work/jruby/jruby/lib/ruby/gems/shared/gems/rspec-expectations-3.13.5/lib/rspec/expectations/fail_with.rb:35:in 'fail_with'
     # /home/runner/work/jruby/jruby/lib/ruby/gems/shared/gems/rspec-expectations-3.13.5/lib/rspec/expectations/handler.rb:39:in 'handle_failure'
     # /home/runner/work/jruby/jruby/lib/ruby/gems/shared/gems/rspec-expectations-3.13.5/lib/rspec/expectations/handler.rb:55:in 'block in handle_matcher'
     # /home/runner/work/jruby/jruby/lib/ruby/gems/shared/gems/rspec-expectations-3.13.5/lib/rspec/expectations/handler.rb:26:in 'with_matcher'
     # /home/runner/work/jruby/jruby/lib/ruby/gems/shared/gems/rspec-expectations-3.13.5/lib/rspec/expectations/handler.rb:47:in 'handle_matcher'
     # /home/runner/work/jruby/jruby/lib/ruby/gems/shared/gems/rspec-expectations-3.13.5/lib/rspec/expectations/expectation_target.rb:65:in 'to'
     # /home/runner/work/jruby/jruby/lib/ruby/gems/shared/gems/rspec-expectations-3.13.5/lib/rspec/expectations/expectation_target.rb:101:in 'to'
     # ./spec/concurrent/actor_spec.rb:155:in 'block in Actor'
```

PR #9717, opened 2026-09-19
Retry the network-dependent CI steps

aminmansuri:
Fixes #9715.

Two steps fail on network errors, not on anything under test:

* `bundle install` — `Could not reach host index.rubygems.org` (exit 17);
  bundler's own retry has no backoff.
* `bootstrap` — `Could not transfer artifact rubygems:bundler:gem:2.7.2`,
  `Connection reset`, from the mavengem wagon.

`tool/ci-retry.sh` retries both, 3 attempts, 15s then 30s apart. Windows jobs
untouched (pwsh). Transparent on success: every wrapped step green in run
35290624118.

#9713 tracks a different class of flake.


aminmansuri:
Run 35408906308 failed the same way in a step this missed: the five Maven
builds inside `test/check_versions.sh` could not resolve two plugins, so
nothing was built and every artefact read as "not found" (same cause as
#9493). Those builds now retry and the script stops on a build failure; the
name and size checks never retry. Also wrapped: the two dist builds with
their `bundle install`, and the shellcheck install.

Left unwrapped on purpose: anything that runs tests, including the sequel and
concurrent-ruby scripts, and the snapshot deploy.

aminmansuri:
ok. I believe this is ready.. and ALL GREEN!

headius (commented):
I'm on the fence about this fix. Obviously I'd prefer if the original commands were reliable, but simply retrying seems like a blunt instrument. We are also patching around some combination of GitHub Actions flakiness and remote service flakiness, which is both super annoying and gross.

Perhaps there's some additional config we can set up to make these commands retry failed network access?

See for example the `BUNDLE_RETRY` env var listed on this doc page: https://guides.rubygems.org/command-reference/bundle-config/

I already added some Maven properties to try to make artifact downloads retry, but I'm not sure the property is having any effect on mavengem fetches. 🤔

aminmansuri:
Yeah.. I see the mistake here, this could maybe hid a legit reason for failure. I'll try to narrow this down

BUNDLE_RETRY is already on at its default of 3, but it cannot help here: Bundler retries without any backoff, and "Could not reach host" is a NetworkDownError, which its downloader lists as non-retryable.

What may help is a retry that waits, which neither bundler nor mavengem wagon will do.

aminmansuri:
@headius what do you think about it now? It narrows down retries to some very specific cases.

I don't love that it triggers on text. But the codes are not rich enough to get the info I needed.

aminmansuri:
The two remaining red checks here, test:mri:stdlib on Java 25 and 26, are not the retry change: both are Errno::EBADF from the socket&.close in test/mri/net/http/utils.rb:35, which is #9722, fixed by #9723. 

The script already passed the test you worried about: a real bug turned up here and still showed red, because retries only cover network steps, not tests.

Issue #9722, opened 2026-09-20
IO#close raises Errno::EBADF after Thread#kill closed the channel underneath it


aminmansuri:
Thread#kill on a thread blocked in IO unblocks it by interrupting the native thread. The JDK responds by closing the interruptible channel. The Ruby stream is untouched — closed? is still false, the ChannelFD is still set — so the ensure that follows calls IO#close, and that close raises Errno::EBADF for a descriptor Ruby never closed. CRuby cannot produce this for the same program.

Deterministic reproduction

```ruby
require 'jruby'
read, write = IO.pipe
JRuby.reference(read).open_file.fd.ch.close   # what the JDK does on interrupt
read.close                                    # Errno::EBADF: Bad file descriptor
```

Realistic reproduction

The MRI test helpers: a thread looping over accept with socket&.close in an ensure, torn down with kill then join. Over 800 rounds across kill delays from 0 to 50 ms:

- CRuby 4.0.7: 400 rounds, 0 failures
- JRuby 10.0.7.0-SNAPSHOT, Java 21: 14 failures in 800
- JRuby 10.0.7.0-SNAPSHOT, Java 25: 12 failures in 800

Every failure is Errno::EBADF from org/jruby/RubyIO.java in close.

Where it comes from

RubyThread.interrupt() calls the unblock function, which ends at nativeThread.interrupt() in OpenFile.READ_TASK/WRITE_TASK.wakeup — OpenFile.java already carries // FIXME: NO! This will kill many native channels 
WRITE_TASK.wakeup, the FIXME at L1466:
https://github.com/jruby/jruby/blob/6718ce1fcb75d4af96dd2b9be1b439e107ffc6a3/core/src/main/java/org/jruby/util/io/OpenFile.java#L1464-L1468

The JDK's AbstractInterruptibleChannel then closes the channel. The subsequent IO#close reaches ChannelFD.finish(), which throws ClosedChannelException when the channel is not open, and Helpers.errnoFromException maps that to Errno::EBADF.

IO#close itself is not the divergence: closing the same object twice, and closing an IO whose descriptor was genuinely closed elsewhere, behave identically on both engines. The divergence is that JRuby invalidates a descriptor the program never closed.

READ_TASK.wakeup, same interrupt, no comment:
https://github.com/jruby/jruby/blob/6718ce1fcb75d4af96dd2b9be1b439e107ffc6a3/core/src/main/java/org/jruby/util/io/OpenFile.java#L1431-L1434

[...]

PR #9723, opened 2026-09-21
Do not report a channel closed underneath us as EBADF on close

aminmansuri:
Fixes #9722.

`Thread#kill` unblocks IO by interrupting the native thread, and the JDK closes the interruptible channel in response. Ruby never closed that stream, but the `ensure` that follows calls `IO#close`, and `ChannelFD.finish()` raised `ClosedChannelException` — surfacing as `Errno::EBADF`, which `Thread#join` then re-raises into the caller.

The close is now skipped when the channel has already gone. Two details:

- `isOpen()` guards `ch.close()` only, not the branch, so the fileno bookkeeping still runs. It cannot be dropped either: `FileDescriptorByteChannel.close()` throws on a second close, and `IOChannel.close()` would call a user object twice.
- `unregisterWrapper` now takes the wrapper and removes only if the fileno still maps to it. By key alone, a stale close evicts a live wrapper on a recycled descriptor and a later close closes an unrelated file.

The one-argument `unregisterWrapper(int)` is left in place, now unused and unsafe unless the caller owns the mapping — deprecating it seemed an API call to leave to you.

### Evidence

Forcing test red before, green after. 800 reproduction rounds: 14 failures on Java 21 and 12 on Java 25 before, 0 after; CRuby 0 in 400. MRI `net/http` reproduces the defect on base at about 1 run in 10, absent across 10 runs after. `test_io.rb` 49, MRI `ruby/test_io.rb` 200, stdlib `net socket io fiber` 362 — all green; io, file and socket specs identical.

### Limits

Only `close` is fixed. After the interrupt the stream still reports `closed? == false` and still fails `write` and `select`. And `io.to_channel.close; io.close` no longer raises `Errno::EBADF`, which CRuby does — a lost diagnostic.

The cause is the interrupt itself, as the `FIXME` at `OpenFile.java:1466` says. Giving `writeInternal` the pre-select `readInternal` has would fix all four, at the cost of a select per write. Happy to take that on if you want it.


aminmansuri:
My interest in the fix is primarily eradicating the build errors.
But the main motivation is that it is an inconsistency with CRuby's behavior. So fixing this makes CRuby compatibility better.

test/jruby/test_io.rb
```diff
@@ -698,4 +698,18 @@ def test_accept_on_closed_server_raises_ioerror
...
 
+  # The JDK closes an interruptible channel when the thread blocked on it is interrupted,
+  # which is how Thread#kill unblocks IO. Ruby never closed the stream, so the close that
+  # follows must succeed rather than report the vanished channel as Errno::EBADF.
+  def test_close_when_channel_closed_underneath
+    require 'jruby'
+
+    read, write = IO.pipe
+    @to_close << write
+    JRuby.reference(read).open_file.fd.ch.close
```
kares: the bug report mentions the MRI test-helpers being reliable source of reproducing.

would be nice to have a higher level test-case here instead of asserting on the JRuby internals.

aminmansuri: Good point. That test asserts on internals because my higher-level reproduction was a race — but it does not have to be. A thread blocked in a write loses its channel to Thread#kill every time, since writeInternal has no pre-select. readInternal does, which is exactly why a blocked read or accept survives.

Measured on 6718ce1 and on this branch, 200 rounds each:

- pipe blocked write: 200/200 Errno::EBADF on base, 0/200 after
- tcp blocked write: the same
- pipe blocked read, tcp blocked read, tcp blocked accept: 0/200 on both

So this replaces the internals test:

```ruby
  # Thread#kill unblocks a write by interrupting the thread, and the JDK closes the
  # interruptible channel in response. Ruby never closed the stream, so the close in
  # the ensure must succeed rather than report the vanished channel as Errno::EBADF.
  def test_close_after_kill_during_blocking_write
    read, write = IO.pipe
    @to_close << read
    @to_close << write

    started = Queue.new
    thread = Thread.new do
      begin
        started << true   # put something in queue
        write.write('x' * 1_000_000) # larger than the pipe buffer, so it blocks
      ensure
        write.close
      end
    end

    started.pop  # <-- wait till something in queue
    sleep 0.2 # let the write fill the buffer and block
    thread.kill

    assert_nothing_raised { thread.join }
  end      
```
                                                                                                         
Red on 6718ce1, green here.

The started queue and the sleep guarantee the kill lands while the thread is already blocked in the write.

Thread#kill unblocks IO by interrupting the thread, which only works if it is already blocked. Kill any earlier and there is nothing to interrupt: the thread is flagged for death, enters the write, and blocks on a pipe nobody reads. The kill is never redelivered, so join never returns.

Polling thread.status == 'sleep' instead killed too early and hung 2 runs in 200. Unrelated to this patch — that thread never reaches the close.

